PR #13907, opened 2021-09-27
fix: wrap itemcounter comp. changes in forVersion (#13828)

### LINKED ISSUE #13828: Cx-item-counter width and forVersion
The forVersion needs to be added to the styling updates in _item-counter.scss and the width that was added needs to be removed.

BenjaminEmiliani:
Closes #13828


projects/storefrontstyles/scss/components/cart/_item-counter.scss
```diff
@@ -1,5 +1,7 @@
 %cx-item-counter {
-  width: 146px;
+  @include forVersion(4.2) {
+    width: 146px;
```
morganm58, line 3: could you completely remove this width. It's not needed.

morganm58 (approved):

PR #13921, opened 2021-09-29
fix: Prevent query initial double load

### LINKED ISSUE #13894: Double requests when using queries
**Describe the bug**
Query call endpoints 2 times instead of once.

**To Reproduce**
1. Go to the login page
2. Reload (to get rid of any potential cache)
3. Navigate to register page
4. Notice double XHR request for titles (first cancelled)

**Expected behaviour**
Subscription to query should invoke underlying request only once.

marlass:
- [x] double loader factory execution fix
- [x] unit tests to cover query service
- [x] incorrect order of emissions after query reset fix
- [x] fixed reload/resets after error throw
- [x] fixed reload trigger behaviour when it happens during query inactive time (0 subscribers)

Closes #13894


projects/core/src/util/command-query/query.service.ts
```diff
@@ -65,19 +65,25 @@ export class QueryService implements OnDestroy {
     );
```
znikola: can we improve the jsdoc comment above?
from `if query will be unsubscribed when data is loaded, we will end up with loading flag set to true`
to `if the query will be unsubscribed from while the data is being loaded, we will end up with the loading flag set to true`

projects/core/src/util/command-query/query.service.ts
```diff
@@ -65,19 +65,25 @@ export class QueryService implements OnDestroy {
...
 
     const loadTrigger$ = this.getTriggersStream([
+      retryInterruptedLoad$,
```
znikola: let's also add a comment here e.g. `we need to evaluate retryInterruptedLoad$ before other triggers in order in order to not emit multiple times`

projects/core/src/util/command-query/query.service.ts
```diff
@@ -65,19 +65,25 @@ export class QueryService implements OnDestroy {
...
       state$.pipe(
         filter(
           ({ data, loading, error }) => data === undefined && !loading && !error
         )
       ),
       ...(options?.reloadOn ?? []),
-      retryInterruptedLoad$,
+      ...(options?.resetOn ?? []),
```
znikola, line 68: Unless I'm missing something, we shouldn't reload the data when the reset trigger fires

projects/core/src/util/command-query/query.service.ts
```diff
@@ -101,7 +107,7 @@ export class QueryService implements OnDestroy {
             state$.next({
               data: undefined,
               error: false,
-              loading: false,
+              loading: true,
```
znikola: If we're just resetting the state, the `loading` flag should be false by default

projects/core/src/util/command-query/query.service.spec.ts
```diff
@@ -24,30 +34,172 @@ describe('QueryService', () => {
...
     });
 
     it('should not load if not accessed', () => {
       const data = query.get();
       expect(data).toBeDefined();
-      expect(loadingStreamAccessed).toBeFalse();
+      expect(loadingStreamAccessed).toBeFalsy();
     });
 
     it('should load if first accessed', () => {
```
znikola: Maybe these two tests should be under `describe('get')`

projects/core/src/util/command-query/query.service.spec.ts
```diff
@@ -24,30 +34,172 @@ describe('QueryService', () => {
...
+      loaderFactoryCalls = 0;
+      query = service.create(
+        () =>
+          defer(() => {
+            loadingStreamAccessed = true;
+            loaderFactoryCalls++;
+            return loadingStream$.pipe(take(1));
+          }),
+        {
+          reloadOn: [TestingEvent],
```
znikola: Let's rename the `TestingEvent` to `ReloadEvent`. It improves the readability of this test

projects/core/src/util/command-query/query.service.spec.ts
```diff
@@ -24,30 +34,172 @@ describe('QueryService', () => {
...
+
+      it('should emit new value after reload trigger', (done) => {
+        const data = query.getState();
+        const states: QueryState<string>[] = [];
+        data.pipe(take(4), skip(2)).subscribe((state) => {
+          states.push(state);
+          if (states.length === 2) {
+            expect(states[0]).toEqual(
+              jasmine.objectContaining({
+                loading: true,
```
znikola, line 292: Just to understand, this one is testing the case when the reloadOn trigger fired?
And that's why the loading is true, but the data is the old one.

znikola (changes requested):

projects/core/src/util/command-query/query.service.spec.ts
marlass: It kinda make sense to have them outside particular method, because they test the general create of the query not get or getState methods.

projects/core/src/util/command-query/query.service.ts
```diff
@@ -87,13 +81,24 @@ export class QueryService implements OnDestroy {
       tap((data) => {
         state$.next({ loading: false, error: false, data });
       }),
-      catchError((error) => {
+      catchError((error, retryStream$) => {
         state$.next({ loading: false, error, data: undefined });
-        return EMPTY;
+        return retryStream$;
```
znikola, line 86: We should discuss this. It can lead to some infinite loops

znikola (changes requested):

znikola (approved):

Issue #13711, opened 2021-09-14
Update schematics to use app-routing-module.ts from storefrontlib 


bgambocjaviniar:
In our schematics, we create the app-routing-module, and put it in the shellapp, which then gets imported. 

We will need to remove this behavior as it will be part of our library, but update it to import the 'new' app routing module found in our storefrontlib

bgambocjaviniar:
PR: GH-13928

PR #13928, opened 2021-09-29
feat: Update schematics to use app-routing-module.ts from storefrontlib

bgambocjaviniar:
closes GH-13711


projects/schematics/src/add-spartacus/index.ts
```diff
@@ -290,8 +301,6 @@ export function addSpartacus(options: SpartacusOptions): Rule {
         module: 'app',
```
LTiger14: This whole section `ensureModuleExists` will generate an empty module file `app-routing.module.ts` and import it in the `app.module.ts` so you should remove it.

projects/schematics/src/add-spartacus/index.ts
```diff
@@ -244,6 +247,14 @@ function updateAppModule(project: string): Rule {
             },
             content: 'HttpClientModule',
           });
+          addModuleImport(sourceFile, {
```
LTiger14, line 243: Should you add a unit test for this?

LTiger14 (changes requested):

projects/schematics/src/add-spartacus/index.ts
bgambocjaviniar: nice catch

LTiger14 (approved):

PR #13932, opened 2021-09-30
feat: add url parsing service to public api

### LINKED ISSUE #13495: Don't find service in the specified "import" path
I try to import the "UrlParsingService" in my customized service but the path shows "@spartacus/core/src/routing/configurable-routes/url-translation/url-parsing.service"; and the console shows "Module not found: Error: Can't resolve '@spartacus/core/src/routing/configurable-routes/url-translation/url-parsing.service' in 'xxx/protected-routes'

Screen Shot 2021-08-18 at 2 28 21 PM

**Desktop
 - OS: [MacOs]
 - Browser [chrome]
 - Version [22]

LTiger14:
closes #13495


Platonn (approved):
It's a low level util service I didn't mean to expose in public API.
But it looks like we didn't need to change it for quite long time. So it seems stable and now I have nothing against exposing.

PR #13946, opened 2021-10-01
fix: add global message on successful payment deletion (#13917)

### LINKED ISSUE #13917: Missing successful global message when user deletes a payment method
**Describe the bug**
When a payment method is deleted successfully, there is no global message issued.

4.1

**To Reproduce**
1. Login
2. Add a payment method through checkout
3. Go to payment details
4. Delete payment

**Expected behavior**
A global message notifies users that payment method is deleted successfully

BenjaminEmiliani:
Closes #13917


projects/core/src/user/store/effects/payment-methods.effect.ts
```diff
@@ -75,8 +80,27 @@ export class UserPaymentMethodsEffects {
...
+    ofType(UserActions.DELETE_USER_PAYMENT_METHOD_SUCCESS),
+    tap(() => {
+      this.showGlobalMessage('paymentCard.deletePaymentSuccess');
+    })
+  );
+
+  /**
+   * Show global confirmation message with provided text
+   */
+  private showGlobalMessage(text: string) {
```
saad-mo: move this below constructor

projects/core/src/user/store/effects/payment-methods.effect.ts
```diff
@@ -75,8 +80,27 @@ export class UserPaymentMethodsEffects {
     })
   );
 
+  @Effect({ dispatch: false })
```
saad-mo: need a unit test for this. I also noticed that there is no spec file for `payment-method.effects` so would have to create a whole new file and add specs for all other effects as well.

saad-mo (dismissed):
missing specs

projects/core/src/user/store/effects/payment-methods.effect.ts
BenjaminEmiliani: Created spec file.

dydome (dismissed):
please provide migration doc

docs/migration/5_0.md
```diff
@@ -40,3 +40,8 @@ under the section `headerDisplayOnly`.
 #### Translations (i18n) changes
 
 - `configurator.addToCart.buttonDisplayOnly` was added to `configurator-common.ts`
+
+### PaymentMethodsEffects
+
+- One new required constructor parameter `globalMessageService: GlobalMessageService`
```
dydome:
```suggestion
- `GlobalMessageService` is a new, required constructor dependency.
```

znikola:
> please provide migration doc

As we are not exposing effects in our public API, there's no need for migration docs or migrations at all :)

docs/migration/5_0.md
```diff
@@ -41,6 +41,11 @@ under the section `headerDisplayOnly`.
 
 - `configurator.addToCart.buttonDisplayOnly` was added to `configurator-common.ts`
 
+### PaymentMethodsEffects
+
+- `GlobalMessageService` is a new, required constructor dependency.
+- `deleteUserPaymentMethod$` effect modified to trigger a global message
+
```
znikola: Please remove, as we don't expose effects

projects/core/src/user/store/effects/payment-methods.effect.spec.ts
```diff
@@ -0,0 +1,108 @@
+import { TestBed } from '@angular/core/testing';
```
znikola, line 1: Thanks for adding a test for the whole class!

projects/core/src/user/store/effects/payment-methods.effect.spec.ts
```diff
@@ -0,0 +1,108 @@
...
+import { provideMockActions } from '@ngrx/effects/testing';
+import { cold, hot } from 'jasmine-marbles';
+import { Observable, of } from 'rxjs';
+import {
+  GlobalMessageService,
+  GlobalMessageType,
+} from '../../../global-message/index';
```
znikola: Please import directly from the file

docs/migration/5_0.md
```diff
@@ -41,13 +41,6 @@ under the section `headerDisplayOnly`.
 
 - `configurator.addToCart.buttonDisplayOnly` was added to `configurator-common.ts`
 
-### PaymentMethodsEffects
-
-- `GlobalMessageService` is a new, required constructor dependency.
-- `deleteUserPaymentMethod$` effect modified to trigger a global message
-
-#### CardComponent
```
znikola: Not sure if this was intentional?
I see this section empty?

znikola (approved):

docs/migration/5_0.md
BenjaminEmiliani: @znikola I am not sure what you are referring to in this question. I think you mean line 49, the removal of #### CardComponent? 

If that is what you are referring to I removed because it was duplicate header and empty

PR #13971, opened 2021-10-05
add abstract services

Pucek9:


feature-libs/cart/import-export/components/import-export/import-export.component.ts
```diff
@@ -20,9 +24,20 @@ export class ImportExportComponent {
...
+  };
+
   constructor(
     protected cmsComponent: CmsComponentData<CmsImportExportComponent>,
-    protected routingService: RoutingService
+    protected routingService: RoutingService,
+    protected activeCartService: ImportExportActiveCartService,
+    protected newSavedCartService: ImportExportNewSavedCartService,
+    protected savedCartService: ImportExportSavedCartService,
+    protected quickOrderService: ImportExportQuickOrderService
```
Platonn: Side issue: how to avoid instatntiating services for features if they are not used
(lower prio)

feature-libs/cart/import-export/components/import-to-cart/import-entries-dialog/import-entries-dialog.component.ts
```diff
@@ -40,20 +41,23 @@ export class ImportEntriesDialogComponent {
...
-  CartTypes = CartTypes;
+  service$ = this.launchDialogService.data$;
 
   constructor(
     protected launchDialogService: LaunchDialogService,
     protected importToCartService: ImportToCartService
   ) {}
 
+  isNewCartForm(service: AbstractImportExportService) {
+    return service instanceof ImportExportNewSavedCartService;
```
Platonn: Customer can have different impl. Then instanceof won't work. Let's introduce a property `readonly cartType` inside each service for runtime inspection

feature-libs/cart/import-export/components/import-to-cart/import-entries-dialog/import-entries-dialog.component.ts
```diff
@@ -40,20 +41,23 @@ export class ImportEntriesDialogComponent {
     errorMessages: [],
   });
 
-  cartType$ = this.launchDialogService.data$;
-  CartTypes = CartTypes;
+  service$ = this.launchDialogService.data$;
```
Platonn: `data$ = service$` looks odd. Let's wrap this service in an object. and the property name should have some meaningful name

feature-libs/cart/import-export/components/export-entries/export-entries.service.ts
```diff
@@ -67,46 +60,19 @@ export class ExportEntriesService {
...
-
-  protected getQuickOrderEntries(): Observable<OrderEntry[]> {
-    return this.quickOrderFacade.getEntries();
-  }
-
-  protected getSavedCartEntries(): Observable<OrderEntry[]> {
-    return this.savedCartDetailsService
-      .getCartDetails()
+  protected getResolvedValues(
+    service: AbstractImportExportService
```
Platonn: Let's pass here `entries` instead of the service. Instead, invoke `service.getEntries()` one level higher.

Note: similar in import-to-cart.service regarding `addEntries`


Btw. then we can consider renaming this file service to e.g.: `ExportProductsToCsvService`, `ImportProductsFromCsvService`

PR #13985, opened 2021-10-06
feature/GH-13918: expose function isGuestCart() in ActiveCartFacade

### LINKED ISSUE #13918: expose function `isGuestCart()` in `ActiveCartFacade`
Facade only can expose functions which return `void` or `observable`. We will make function `isGuestCart` returns `Observable<boolean>`, so that this function can be exposed.

WeizhengSap:
close #13918


feature-libs/cart/main/core/facade/active-cart.service.ts
```diff
@@ -474,21 +477,31 @@ export class ActiveCartService implements OnDestroy {
...
+  isGuestCart(cart?: Cart): Observable<boolean> {
+    if (cart) {
+      const cartUser = cart?.user;
+      return of(
+        Boolean(
+          cartUser &&
+            (cartUser.name === OCC_USER_ID_GUEST ||
+              this.isEmail(cartUser.uid?.split('|').slice(1).join('|')))
+        )
+      );
```
plabadie: These lines are repeated twice within `isGuestCart()`, is there a way we can avoid this duplication?

WeizhengSap: we can create a private function

feature-libs/checkout/components/guards/checkout-auth.guard.ts
```diff
@@ -51,10 +51,7 @@ export class CheckoutAuthGuard implements CanActivate {
     );
   }
 
-  protected handleAnonymousUser(cartUser?: User): boolean | UrlTree {
-    if (this.activeCartService.isGuestCart()) {
-      return !!cartUser;
-    }
```
plabadie, line 57: Why we remove this?

WeizhengSap: we don't get `this.activeCartService.getAssignedUser()`, but get `this.activeCartFacade.isGuestCart()` directly. So no need to check `this.activeCartService.isGuestCart()` again.

plabadie (approved):

PR #13996, opened 2021-10-07
Bugfix/occ 6119

kinga-k:
Fix for duplicate request to strategy service from merchandising carousel component.

Also fixes https://github.com/SAP/spartacus/issues/14105.


integration-libs/cds/src/merchandising/facade/cds-merchandising-user-context.service.ts
```diff
@@ -30,7 +33,9 @@ export class CdsMerchandisingUserContextService {
     private routingService: RoutingService,
     private productSearchService: ProductSearchService,
     private converterService: ConverterService,
-    private profileTagEventService: ProfileTagEventService
+    private profileTagEventService: ProfileTagEventService,
+    private config: CdsConfig,
```
bgambocjaviniar: This is a breaking change, therefore you would need to deprecate this constructor and create other constructors through constructor overload.

An example would be this https://github.com/SAP/spartacus/blob/develop/projects/storefrontlib/cms-components/cart/add-to-cart/add-to-cart.component.ts#L71

---
One note, instead of having these two dependencies, we can depend maybe on only `ProfileTagLifecycleService`.consentChanged, which returns an object that shows if the consent is granted or not.

integration-libs/cds/src/merchandising/facade/cds-merchandising-user-context.service.ts
```diff
@@ -61,11 +66,24 @@ export class CdsMerchandisingUserContextService {
   }
 
   private getConsentReferenceContext(): Observable<MerchandisingUserContext> {
-    return this.profileTagEventService.getConsentReference().pipe(
-      startWith(''),
-      distinctUntilChanged(),
-      map((consentReference) => ({ consentReference }))
-    );
+    return this.consentService
```
bgambocjaviniar: As mentioned above, you can use the one dependency which would be ProfileTagLifecycleService.

**However, I do have a question.**

I believe this logic should be done in the profile tag part, where the `getConsentReference` would return empty string is consent is not given or undefined right?

integration-libs/cds/src/merchandising/facade/cds-merchandising-user-context.service.ts
```diff
@@ -61,11 +66,24 @@ export class CdsMerchandisingUserContextService {
...
+      .pipe(
+        switchMap((profileConsent) => {
+          if (this.consentService.isConsentGiven(profileConsent)) {
+            return this.profileTagEventService.getConsentReference().pipe(
+              distinctUntilChanged(),
+              map((consentReference) => ({ consentReference }))
+            );
+          } else {
+            return this.profileTagEventService.getConsentReference().pipe(
+              startWith(''),
```
bgambocjaviniar: Can you remind me of the case for merchandising, where the consent is not given. Should it request something to the strategy endpoint if it is not given? 

I believe that would be the case as if we take a look at https://jsapps.cltq4tr3nt-contextdr1-d66-public.model-t.cc.commerce.ondemand.com/apparel-uk-spa/en/GBP/?profileTagDebug=true 
when we remove the consent, it does not send a request when it does not have a consent reference.

I believe anything related to CDS (profiletag / merchandising) can only get triggered only if the consent is given for CDS right? Therefore, keeping what it had before and removing startWith('') would solve the issue at hand.

bgambocjaviniar: In conclusion,

1 - if logic holds for with and without consentReference to send a request to strategy, then I believe the logic, where the `getConsentReference` from profile tag service should be changed to emit an empty string or null if no consent is given.

2 - From my point of view and context, merchandising shouldn't request any data from strategies until a consent is given

3 - have to not create breaking changes as customers who uses the old versions will have their application broken once they upgrade. Make sure to create an issue in the spartacus repo to keep track of the deprecation if it needs to be done.

4 - there's a current bug with this implementation as they're no longer distinct when toggling. Make sure consent is enabled. Toggle it off and back on, and see the there's a duplicate request.

5 - merge latest integration-libs to this branch to get the e2es that Shankari merged in so that it can run against this PR as well.

bgambocjaviniar (changes requested):

integration-libs/cds/src/merchandising/facade/cds-merchandising-user-context.service.ts
bgambocjaviniar: refer to https://github.com/SAP/spartacus/pull/13996/files#r728368238 before starting to fix/debug

integration-libs/cds/src/merchandising/facade/cds-merchandising-user-context.service.ts
bgambocjaviniar: refer to https://github.com/SAP/spartacus/pull/13996/files#r728368238 before starting to fix/debug

integration-libs/cds/src/merchandising/facade/cds-merchandising-user-context.service.ts
kinga-k: 1 - I didn't want to change code from profile tag to not break that functionality
2 - Merchandising Carousel should be displayed on the storefront even if consent are not given
3 - Ok
4 - You are right. I didn't noticed this problem.
5 - I have merged it

bgambocjaviniar: 1 - the reason why it's best to keep it in profile tag this logic is because getConsentReference should be emitting a default and we wouldn't be repeating ourselves to say if a consent is given or not. It would help alleviate the issue on adding dependencies for breaking changes here as well, whereas the functionality fix on profile tag will be considered a bugfix instead of a breaking change.

PR #14002, opened 2021-10-07
fix: add aria-label to store finder search input (#13439)

### LINKED ISSUE #13439: Find a Store search box needs label other than the placeholder
**Describe the bug**

- Automated test detects missing label.
- Placeholders shouldn't be used as labels.

- Library version: 4.0.0

**Expected behavior**

- Associate a label, or use aria-label.
- Use aria-describedby if more instructions are needed. (e.g. enter a promo code here if you have one)
- Screen Reader vocalizes field and required action.

**Screenshots**
Screen Shot 2021-08-05 at 6 33 01 PM

BenjaminEmiliani:
Closes #13439


feature-libs/storefinder/components/store-finder-search/store-finder-search.component.html
```diff
@@ -8,6 +8,7 @@
           (keyup)="onKey($event)"
           type="text"
           class="form-control"
+          [attr.aria-label]="'storeFinder.searchBox' | cxTranslate"
```
bit172: We shouldn't use the placeholder text for the aria labels. It should be text to describe the element instead of providing instructions.

bit172 (approved):

PR #14007, opened 2021-10-08
Move cart code to cart lib, step 4

### LINKED ISSUE #13909: Move cart code to cart lib, step 4: leftover
There are some leftover cart codes still in `core` and `storefrontlib`.

plabadie:


feature-libs/cart/saved-cart/core/events/saved-cart-event.builder.ts
```diff
@@ -1,6 +1,7 @@
 import { Injectable, Type } from '@angular/core';
 import { ofType } from '@ngrx/effects';
 import { ActionsSubject } from '@ngrx/store';
+import { CartActions } from '@spartacus/cart/main/core';
```
plabadie, line 4: Saved cart depends on "main" cart.
Does this mean automatically that we need to configure a lazy loading dependency?
Lets discuss

projects/storefrontapp/src/app/spartacus/features/saved-cart-feature.module.ts
```diff
@@ -17,6 +18,7 @@ import { CmsConfig, I18nConfig, provideConfig } from '@spartacus/core';
         [CART_SAVED_CART_FEATURE]: {
           module: () =>
             import('@spartacus/cart/saved-cart').then((m) => m.SavedCartModule),
+          dependencies: [CART_FEATURE],
```
plabadie: Lets discuss about where we configure `dependencies: [CART_FEATURE]`
-In the feature module?
-In the default configs?
We can then align other modules.

feature-libs/cart/main/components/cart/cart-shared/cart-item/cart-item.component.ts
```diff
@@ -1,15 +1,14 @@
 import { Component, Input, OnChanges, SimpleChanges } from '@angular/core';
 import { FormControl } from '@angular/forms';
-import { CartOutlets, PromotionLocation } from '@spartacus/cart/main/root';
+import {
+  CartItemComponentOptions,
+  CartItemContext,
+  CartItemContextSource,
```
WeizhengSap: `CartItemContext` can be in `root`, but `CartItemContextSource` has to stay with `CartItemComponent`, because it is the component's provider.

plabadie: I put CartItemContextSource back where it was in components.

feature-libs/cart/saved-cart/core/events/saved-cart-event.builder.ts
WeizhengSap: If we just `import ...` from cart lib, no need add `cart` to lazy loading dependency.

projects/storefrontapp/src/app/spartacus/features/product-configurator-rulebased-feature.module.ts
```diff
@@ -22,6 +23,7 @@ import { VariantConfiguratorOverviewLayoutModule } from '@spartacus/product-conf
             import('@spartacus/product-configurator/rulebased').then(
               (m) => m.RulebasedConfiguratorModule
             ),
+          dependencies: [CART_FEATURE],
```
WeizhengSap: if just `import`, no need add this.

feature-libs/checkout/occ/adapters/occ-checkout-cost-center.adapter.ts
```diff
@@ -1,12 +1,9 @@
 import { HttpClient } from '@angular/common/http';
 import { Injectable } from '@angular/core';
+import { CART_NORMALIZER } from '@spartacus/cart/main/core';
```
WeizhengSap: We can put tokens in `root`. Most tokens do not have the corresponding `useExising` classes. But, any other libs which need use the tokens, can put `cart` into their lazy loading dependencies.

plabadie: All normalizer tokens moved to root

projects/storefrontapp/src/app/spartacus/features/product-configurator-rulebased-feature.module.ts
plabadie: It's imports + CartItemContext, but it is in carrt/main/root.  So indeed, I don't think it's needed.

WeizhengSap (approved):

PR #14032, opened 2021-10-13
fix: add SRs to checkboxs for new payment during checkout (#13733)

### LINKED ISSUE #13733: During checkout: when SR users adds new Payment Method more context is needed.
**Describe the bug**
- When adding a new Payment Method during checkout, the checkbox "Set as default" needs more context. E.g. "Set this payment method as default"
- If user tabs to checkbox "Same as shipping address" only **hears** exactly that. Without the missing context of "billing address is the same as shipping address

- Library version: 4.0.0

**To Reproduce**

1. Steps to reproduce the behavior:
2. Log in as a registered user
3. Add items to cart and start checkout process.
4. Open any Screen Reader (VO, NVDA, JAWS).
5. Go to the Payment Details step.
6. Hit the Add New Payment button
7. Tab through the form, focus on checkboxes
8. Listen to screen reader feedback

**Expected behavior**
- Use `aria-label` to extend screen reader vocalization, as per screenshots below.
- - The `aria-label` is not needed if designers agree to change the text for mouse users as well, please validate with @Manurod2018 and @vekterdude 

**Screenshots**
Suggested vocalization:

Screen Shot 2021-09-15 at 11 33 05 AM
Screen Shot 2021-09-15 at 11 33 44 AM

BenjaminEmiliani:
Closes #13733 

As mentioned in PR, #14053,  Manuel agreed that changing the text of the checkbox (set as default) would be the best solution.


projects/assets/src/translations/en/payment.ts
```diff
@@ -12,9 +12,11 @@ export const payment = {
     securityCode: 'Security code (CVV)',
     securityCodeTitle: 'Card Verification Value',
     saveAsDefault: 'Save as default',
-    setAsDefault: 'Set as default',
+    setAsDefault: 'Set as default payment method',
     billingAddress: 'Billing address',
     sameAsShippingAddress: 'Same as shipping address',
```
RadhepS, line 17: I could be wrong but I don't think we need both "sameAsShippingAddress and billingAddressSameAsShipping".

It seems to only be used in one place. We can either delete sameAsShippingAddress or change the text in it to contain the text from billingAddressSameAsShipping.

BenjaminEmiliani: Maybe, but the intention is to have 'sameAsShippingAddress' as the text of the checkbox, and the more detailed 'billingAddressSameAsShipping' to be read by a Screen Reader to add more context for a11y users. 

Any thoughts @developpeurweb?

RadhepS: Yeah the original ticket mentions that we can avoid using aria-labels. I'm not sure if it applies to both or only the set as default text. We can see what Miguel thinks. (Albeit we would want to re-arrange how the view so it makes sense.)

image

developpeurweb: @BenjaminEmiliani @RadhepS 

SR-wise it LGTM as is. Context is solid.

However, to be sure, ask Manu for the second checkbox as you did for the first one. Either change the text or use aria-label. Be aware that, it seems we can't avoid being repetitive for either, sighted and SR users. Having a "Billing Address" title over the checkbox will be repetitive for sighted users if changing with the long text. But also, in this case, the aria-label doesn't override but adds up because the `<label>` is wrapping both texts instead of associating them with `for/id`, so both texts will get vocalized for SR users (not a big deal IMO as it reinforces context).

The important part SR users need to hear is that the `Billing address is the same as shipping address` since they can't see the title over the checkbox.

Having said all that, a decision for balance is needed. **Please have Manu decide** which option he prefers, both work well with Screen Readers.

Option 1
Screen Shot 2021-10-26 at 7 53 12 PM

Option 2
Screen Shot 2021-10-26 at 5 47 37 PM

BenjaminEmiliani: Will do, thanks @developpeurweb

BenjaminEmiliani: Spoke with Manu, he prefers option 2.

developpeurweb (approved):

PR #14035, opened 2021-10-14
fix: add information about opening link in a new window

### LINKED ISSUE #13736: Find a Store: No vocalization warning the link is external and will open in a new window
**Describe the bug**
- There is no verbal indication to screen users that the "Get directions" button will open on a different website and different window, as opposed to the store title that opens more details for a given store on the same website.

- Library version: 4.0.0

**To Reproduce**

4. Select any country
5. Tab to any location's "Get Directions" button
6. Hear screen reader feedback, missing the important detail that the link is external and will open in a new window.

**Expected behavior**
- On focusing the link, the screen reader vocalizes link purpose and the additional detail that it will open in a new window.
- Use `aria-label` to extend screen reader vocalization, as per screenshot.
- Suggested `aria-label`: _"Get Directions, this external link opens in a new window"_ please validate phrasing with @Manurod2018 and @vekterdude

**Screenshots**
Suggested markup:

Screen Shot 2021-09-15 at 1 45 00 PM

Matejk00:
Closes: #13736

### DIFF
--- a/feature-libs/storefinder/assets/translations/en/store-finder.ts
+++ b/feature-libs/storefinder/assets/translations/en/store-finder.ts
@@ -5,6 +5,7 @@ export const storeFinder = {
     back: 'Back',
     call: 'Call',
     getDirections: 'Get Directions',
+    ariaLabelGetDirections: 'this link opens in a new window',
     listView: 'List View',
     mapView: 'Map View',
     noStoresFound: 'No Stores Found.',
--- a/feature-libs/storefinder/assets/translations/en/store-finder.ts
+++ b/feature-libs/storefinder/assets/translations/en/store-finder.ts
@@ -5,7 +5,7 @@ export const storeFinder = {
     back: 'Back',
     call: 'Call',
     getDirections: 'Get Directions',
-    ariaLabelGetDirections: 'this link opens in a new window',
+    ariaLabelGetDirections: 'Get Directions, this link opens in a new window',
     listView: 'List View',
     mapView: 'Map View',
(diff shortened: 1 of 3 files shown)


feature-libs/storefinder/assets/translations/en/store-finder.ts
```diff
@@ -5,6 +5,7 @@ export const storeFinder = {
     back: 'Back',
     call: 'Call',
     getDirections: 'Get Directions',
+    ariaLabelGetDirections: 'this link opens in a new window',
```
BenjaminEmiliani: The labels work as intended on VO. Te only suggestion I have is to modify this label to:
ariaLabelGetDirections: 'Get Directions, this link opens in a new window',

I am not sure if on JAWS or NVDA the text of the link itself, 'Get Directions', is being read as well, but at least on VO, when the link gains focus, it only reads the 'ariaLabelGetDirections' label, so some context is lost

developpeurweb: Same for JAWS/NVDA, this has to do with `aria-label` being very aggressive in that regard: it will override everything else. Even existing text. SRs will pick up only the aria-label. Consequently, the "Get Directions" text needs to be part of the label.

Matejk00: thanks, I updated the translation

BenjaminEmiliani (approved):
Good on VO!

developpeurweb (approved):
Works well with JAWS and NVDA.

PR #14036, opened 2021-10-14
fix: Domain handling in SSR for render key

znikola:
This PR takes the whole domain when resolving the rendering key in SSR.
Without it, the sites which embed the site info in their domain won't render properly (e.g. my-site.rs, my-site.ca, my-site.pl, etc.).


core-libs/setup/ssr/optimized-engine/optimized-ssr-engine.spec.ts
```diff
@@ -637,18 +710,19 @@ describe('OptimizedSsrEngine', () => {
 
     describe('when disabled', () => {
       it('should fallback to CSR for parallel subsequent requests for the same rendering key', fakeAsync(() => {
+        const renderingKey = `http://localhost:4200/${requestUrl}`;
```
Platonn: I'm not sure it's justified to pollute all the unit tests below with the `http://localhost:4200/`. Won't it suffice the rendering will be just an abstract `a` or `b`? IMHO the tests shouldn't break if the don't pass here domain and port.

core-libs/setup/ssr/optimized-engine/optimized-ssr-engine.spec.ts
```diff
@@ -48,21 +53,40 @@ class TestEngineRunner {
...
+    url: string,
+    params?: {
+      /** headers */
+      httpHeaders?: IncomingHttpHeaders;
+    }
+  ): TestEngineRunner {
     const response: { [key: string]: string } = {};
-    const headers = new Headers(params?.httpHeaders ?? {});
+    const headers = params?.httpHeaders ?? { host };
+    /** used when resolving SERVER_REQUEST_URL and SERVER_REQUEST_ORIGIN */
```
Platonn:
```suggestion
    /** used when resolving getRequestUrl() and getRequestOrigin() */
```

Platonn (approved):

PR #14053, opened 2021-10-15
improvement: when SR users adds new Shipping Address more context is needed

### LINKED ISSUE #13734: During checkout: when SR users adds new Shipping Address more context is needed
**Describe the bug**
- When adding a new Payment Method during checkout, the checkbox "Set as default" needs more context. E.g. "Set this payment method as default"
- If user tabs to checkbox "Same as shipping address" only **hears** exactly that. Without the missing context of "billing address is the same as shipping address

- Library version: 4.0.0

**To Reproduce**

1. Steps to reproduce the behavior:
2. Log in as a registered user
3. Add items to cart and start checkout process.
4. Open any Screen Reader (VO, NVDA, JAWS).
5. Go to the Shipping Address step.
6. Hit the Add New Address button
7. Tab through the form, focus on last checkboxe.
8. Listen to screen reader feedback

**Expected behavior**
- Use `aria-label` to extend screen reader vocalization, as per screenshot.
- The `aria-label` is not needed if designers agree to change the text for mouse users as well, please validate with @Manurod2018 and @vekterdude 

**Screenshots**
Suggested vocalization:

Screen Shot 2021-09-15 at 11 36 22 AM
 below.

RadhepS:
closes: https://github.com/SAP/spartacus/issues/13734

As mentioned in that ticket. I spoke with Manuel and he agreed that changing the text would be the best solution


BenjaminEmiliani (dismissed):
Good on Voice Over!

projects/assets/src/translations/en/address.ts
```diff
@@ -31,7 +31,7 @@ export const address = {
     streetAddress: 'Street Address',
     aptSuite: 'Apt, Suite',
     selectOne: 'Select One...',
-    setAsDefault: 'Set as default',
+    setAsDefault: 'Set as default shipping address',
```
BenjaminEmiliani, line 34: I do wonder if this is a breaking change. In the Spartacus Docs it mentions:
`Changing or removing translation keys or values. Note: Adding new translation keys and values is not considered a breaking change.`

So maybe create a new key with the value 'Set as default shipping address'? I'm not sure

RadhepS: Yes you are right!

I think if we create a new translation item and use it, it would still be a breaking change because we changed the view for the client.

I think we can go with the approach of using an aria-label now to avoid breaking changes. Good find!

RadhepS: After thinking it over, I think a better solution would just be to push this ticket back to the 5.0 release. This is a cleaner/better solution so I rather not use the aria-label implementation because we would want to change it in the future.

Awesome find again :)

BenjaminEmiliani: Okay, I agree and I will do the same for a very similar issue #13733

BenjaminEmiliani, line 34: Change needs to be documented in 5_0.md

RadhepS: Added

docs/migration/5_0.md
```diff
@@ -37,6 +37,7 @@ under the section `headerDisplayOnly`.
 - `leaveConfigurationOverview(): void `
 - `goToOrderDetails(owner: CommonConfigurator.Owner): void `
 
-#### Translations (i18n) changes
+#### Translation changes
 
 - `configurator.addToCart.buttonDisplayOnly` was added to `configurator-common.ts`
+- `setAsDefault` text was changed in `address.ts`
```
BenjaminEmiliani: I don't think you want to add this change here. I think these are translation changes related to `configurator add to cart`.

I would suggest:

```
#### ShippingAddress Related Translations (i18n) changes
- Content of `address.addressForm.setAsDefault` changed from `Set as default` to `Set as default shipping address`
```

(I did not verify spelling/punctuation)

RadhepS: Changed!

BenjaminEmiliani (dismissed):

BenjaminEmiliani (approved):

PR #14056, opened 2021-10-15
feature/GH-14055: make CMS component trigger multiple libs loading

### LINKED ISSUE #14055: [POC] make CMS component trigger multiple libs loading

WeizhengSap:
close #14055


projects/storefrontlib/cms-structure/services/cms-features.service.ts
```diff
@@ -62,7 +62,9 @@ export class CmsFeaturesService {
             featureConfig?.cmsComponents?.length
           ) {
             for (const component of featureConfig.cmsComponents) {
-              this.componentFeatureMap.set(component, featureName);
+              const existing = this.componentFeatureMap.get(component) || [];
+              const newValue = existing.concat([featureName]);
```
plabadie, line 70: `existing` value might not exist yet, right?  Perhaps check if `existing` is defined before calling concat on it.

plabadie: nevermind, I just saw the `|| []`

projects/storefrontlib/cms-structure/services/cms-features.service.ts
```diff
@@ -83,16 +85,22 @@ export class CmsFeaturesService {
...
+    const arr$ = features.map((feature) =>
+      this.resolveFeatureInstance(feature)
+    );
+    return zip(...arr$).pipe(
+      map((featureInstances) => {
+        const foundInstance = featureInstances.find(
+          (instance) => instance.componentsMappings?.[componentType]
+        );
+        return foundInstance?.componentsMappings?.[componentType];
+      })
```
plabadie, line 107: This will return only one of the cms mappings, while ini reality a component can have many.  
We should make sure this will have no side effects.
Also, if ever we return just one cms mapping arbitrarily, could we just pick the first one from the feature map?

projects/storefrontlib/cms-structure/services/cms-features.service.ts
```diff
@@ -102,21 +110,34 @@ export class CmsFeaturesService {
...
+          return featureInstances.find(
+            (instance) => instance.componentsMappings?.[componentType]
+          );
+        })
+      )
+      .subscribe((featureInstance) => {
+        module = featureInstance?.moduleRef;
       })
       .unsubscribe();
+
```
plabadie, line 143: This will return only one of the feature modules, while ini reality a component can have many.  
We should make sure this will have no side effects.
Also, if ever we return just one cms mapping arbitrarily, could we just pick the first one from the feature map?

projects/storefrontlib/cms-structure/services/cms-features.service.ts
WeizhengSap: There is only one cms mapping. Even one cms component can trigger multiple libs, but its mapping can only be inside one lib.

projects/storefrontlib/cms-structure/services/cms-features.service.ts
WeizhengSap: It only get the module which has the cms mapping.

projects/storefrontlib/cms-structure/services/cms-features.service.spec.ts
```diff
@@ -88,10 +88,11 @@ describe('CmsFeaturesService', () => {
       });
     });
 
-    it('should call FeatureModulesService.resolveFeature', (done) => {
+    it('should call FeatureModulesService.resolveFeature for each mapped feature', (done) => {
       spyOn(featureModules, 'resolveFeature').and.callThrough();
       service.getCmsMapping('component1').subscribe(() => {
         expect(featureModules.resolveFeature).toHaveBeenCalledWith('feature1');
+        expect(featureModules.resolveFeature).toHaveBeenCalledWith('feature2');
```
WeizhengSap, line 95: I added 'component1' to `feature2`'s config. The only effect is: when '`getCmsMapping(component1)` is called, `featureModules.resolveFeature` call 'feature2' as well, which leads to the `feature2` module will be loaded. There are no other side effects.

plabadie (approved):

PR #14060, opened 2021-10-18
fix: add global message for payment method selection (#13729)

### LINKED ISSUE #13729: During checkout: missing successful global message when user selects different Payment Method
**Describe the bug**
- Screen reader users wouldn't be able to tell if they have switched/selected an additional Payment Method during checkout.

- Library version: 4.0.0

**To Reproduce**

4. Switch Payment Method from default to an additional one.

**Expected behavior**
- Screen reader users will hear a successful confirmation message saying they have changed/selected a different Payment Method. E.g _"Payment Method selection successfully changed"._

**Screenshots**

pm-change-2021-09-15_10-22-53

BenjaminEmiliani:
Closes #13729


feature-libs/checkout/core/store/effects/checkout.effect.ts
```diff
@@ -289,6 +292,17 @@ export class CheckoutEffects {
     withdrawOn(this.contextChange$)
   );
 
+  @Effect({ dispatch: false })
+  showGlobalMessageOnSetPaymentDetailsSuccess$ = this.actions$.pipe(
+    ofType(CheckoutActions.SET_PAYMENT_DETAILS_SUCCESS),
+    tap(() => {
+      this.addGlobalMessage(
```
saad-mo: why not call  `globalMessageService` directly here instead of creating a new private function (we'll have to write to write a test as well in this case) 

see [`saved-cart.effect.ts`](https://github.com/SAP/spartacus/blob/180409f98dbacdac5ba48e645d0da19ee0bf8b61/feature-libs/cart/saved-cart/core/store/effects/saved-cart.effect.ts#L114)

feature-libs/checkout/core/store/effects/checkout.effect.ts
```diff
@@ -456,6 +470,16 @@ export class CheckoutEffects {
...
+    private checkoutConnector: CheckoutConnector,
+    private messageService?: GlobalMessageService
   ) {}
+
+  /**
+   * Add Global message with provided text
+   * @param text
+   * @param type
+   */
+  private addGlobalMessage(text: string, type: GlobalMessageType): void {
```
saad-mo: may not need this if we call the `globalMessageService` directly in the effect.

feature-libs/checkout/core/store/effects/checkout.effect.ts
```diff
@@ -456,6 +470,16 @@ export class CheckoutEffects {
     private checkoutDeliveryConnector: CheckoutDeliveryConnector,
     private checkoutPaymentConnector: CheckoutPaymentConnector,
     private checkoutCostCenterConnector: CheckoutCostCenterConnector,
-    private checkoutConnector: CheckoutConnector
+    private checkoutConnector: CheckoutConnector,
+    private messageService?: GlobalMessageService
```
saad-mo: instance name should match the import for consistency `private globalMessageService: GlobalMessageService`

developpeurweb (dismissed):

dydome (approved):

PR #14061, opened 2021-10-18
Adding vocalization for cancel order page when no items are added

0xfaustian:
I added the role="status" as there seems to be problems with aria-live on NVDA and JAWS


BenjaminEmiliani:
If 'role="status' does not fix the issue NVDA and JAWS, then maybe replace all three:
'aria-live=assertive' role=status' 'aria-atomic=true'

With 'role="alert"'

feature-libs/order/components/amend-order/cancellations/cancel-order/cancel-order.component.html
```diff
@@ -4,7 +4,13 @@
   <cx-amend-order-items *ngIf="entries$ | async as entries" [entries]="entries">
   </cx-amend-order-items>
 
-  <cx-form-errors [control]="form.get('entries')"></cx-form-errors>
+  <cx-form-errors
+    [control]="form.get('entries')"
+    aria-live="assertive"
+    aria-atomic="true"
+    role="status"
```
developpeurweb, line 11: Unfortunately, this doesn't work for JAWS and NVDA. I tested with role="alert" as well, and nothing. This is strange, I'll do some research and get back to you.

saad-mo: @developpeurweb There is a general UX issue with this page as well. When you have a longer list of products, the `Select at least one item` error notice is only displayed at the bottom of the page and the user does not know if there is an error unless they scroll to the bottom of the page. I know this is a question for UX, but what would be the impact on SR if the error notice is displayed both on the top and bottom of the product list? 

Screen Shot 2021-10-18 at 2 48 01 PM

developpeurweb: @saad-mo Only one instance of the error should be picked up by the SR. I agree with your UX observation, it should be on top, or more visible. Also, we already know `<cx-global-message>` is working well with VO/JAWS/NVDA, so if we manage to place the alert there, that might solve the issue. If the previous is possible, pls validate with designers. Otherwise, pls let me know to keep researching other alternatives.

BenjaminEmiliani:
Closing PR, member has left the team. Fix will be done in new branch/PR.

PR #14063, opened 2021-10-18
fix: add region landmark to Search Results page (#13517)

### LINKED ISSUE #13517: Add region landmarks to Search Results page
**Describe the bug**

- Screen Reader users must be able to navigate by landmarks, currently, there aren't many nor labelled, therefore relevant sections on the page should be identified with meaningful landmarks names.

- Library version: 4.0.0.

**To Reproduce**
2. Bring up the Rotor or Element List Menu.
3. Choose Landmarks view.
4. See absence of relevant landmarks.

**Expected behavior**

- Use only predefined landmarks https://wiki.wdf.sap.corp/wiki/display/spar/Landmarks
- `<cx-product-facet-navigation role="region" aria-label="Filter Results by Facets">`
- `<cx-product-list role="region" aria-label="Product Results List">`

BenjaminEmiliani:
Closes #13517


developpeurweb (dismissed):
Minor issue with VO's focus management to be retested before 5.0 release. After refactorings take place on that area.

developpeurweb (dismissed):
Still works well with VO, JAWS and NVDA.
Minor issue with VO's focus management to be retested before 5.0 release.

projects/storefrontlib/cms-components/product/product-list/product-facet-navigation/facet-list/facet-list.component.html
```diff
@@ -1,4 +1,4 @@
-<div
+<section
   class="inner"
```
developpeurweb, line 2: @BenjaminEmiliani  Can you please move the `aria-label` to this new `<section>` and then remove the `role="region"` from the parent tag.
 If we equalize both `<section>`s to have the same attributes it works pretty well. Also, it seems that by adding `tabindex="-1"` to the other section labelled `Product results List` the focus issue we had for the VO rotor gets fixed.
Can you please check if it's possible to make it look like these:

Screen Shot 2021-10-28 at 6 53 45 PM
Screen Shot 2021-10-28 at 6 51 56 PM

developpeurweb (changes requested):

developpeurweb (dismissed):

developpeurweb (approved):

PR #14085, opened 2021-10-21
fix: Allow user to backspace with one character to enter bigger number

### LINKED ISSUE #13630: Quick Order cx-item-counter input of value >= 200
**Describe the bug**
End user can't enter the item count manually for values past 199. So if a customers wants 201 units of an item, they need to enter 199, then click on the plus icon twice. I think this is because the cx input has an attribute 'min="1"'.

**Note**: this issue does not occur with the item counters in the active cart.

Since I am new to the team I don't know if this issue was previously discussed and concluded.

https://b2bspastore.cg79x9wuu9-eccommerc1-p5-public.model-t.myhybris.cloud/powertools-spa/en/USD/

**To Reproduce**
1. Go to Quick Order
2. Enter a valid SKU and hit ENTER
3. Try to enter 201 in the QTY input
4. See the behaviour described above

**Expected Behaviour**
An intuitive way of entering item count  >= 200.

 - OS: Mac OS
 - safari
 - Version 14.1.2

dydome:
close #13630

### DIFF
--- a/feature-libs/cart/quick-order/components/cart-quick-order-form/cart-quick-order-form.component.html
+++ b/feature-libs/cart/quick-order/components/cart-quick-order-form/cart-quick-order-form.component.html
@@ -26,7 +26,6 @@
           </label>
           <input
             aria-required="true"
-            [min]="min"
             class="form-control input-quantity"
             formControlName="quantity"
             type="number"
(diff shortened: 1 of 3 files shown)


BenjaminEmiliani (dismissed):
Works good Michal! Only thing I have to note is that the Quick Order component inside the shopping cart still has the bug (no backspace with one char)

BenjaminEmiliani (approved):
Looks good!

PR #14088, opened 2021-10-21
fix: Alt text is missing or not meaningful in describing the image (#11319)

### LINKED ISSUE #11319: Alt text is missing or not meaningful in describing the image
Current Narration

Not Meaningful
Screen Shot 2020-09-28 at 5 20 15 PM

Absent
Screen Shot 2020-09-28 at 5 20 35 PM

Expectation

- Alternative text should describe the image in no more than 125 characters.
- If a description is longer than 125 characters needed, then the image should have a "longdesc" attribute, otherwise, it will linger even if the user moves to the next item.
- A good rule of thumb to consider is to include what you might relay over the phone.
- The word "image" is vocalized by default at the end of the alt text so no need to include it.

bit172:
closes #11319


RadhepS (dismissed):
Tested using Voice Over for Mac and it works as expected

projects/storefrontlib/cms-components/product/product-list/product-list-item/product-list-item.component.html
```diff
@@ -9,7 +9,7 @@
         class="cx-product-image"
         [container]="product.images?.PRIMARY"
         format="product"
-        [alt]="product.summary"
+        [alt]="getMediaAltText()"
```
BenjaminEmiliani: Not completely sure if this would work, but to avoid adding logic to the component, maybe try:

[alt]="{{ product.summary ?? product.name }}"

dydome: I will also go this way -> [alt]="{{ product.summary ?? product.name }}"

dydome (changes requested):

Matejk00:
Before the merge please consider changes proposed by @BenjaminEmiliani and @dydome

dydome (dismissed):

BenjaminEmiliani (approved):
Works as expected

PR #6751, opened 2020-03-02
fix: adapt a11y for storefinder

Matejk00:
Closes: #6676

PR #14102, opened 2021-10-25
fix: store name link with rotor links navigation for screen reader #13359

### LINKED ISSUE #13359: [VoiceOver/Chrome-Safari] Store name link is vocalized but doesn't work
**Describe the bug**

- While navigating page content using Rotor> Headings, and once a specific store name has been focused, the user is unable to follow the link because it seems code order is inversed.
- Instead of `<a><h2></h2></a>` as it is in most places, there we have `<h2><a></a></h2>`, so both the heading and the link gets vocalized but VoiceOver is unable to follow the link with this combination.

- Library version: 4.0.0

**To Reproduce**
1. Go to Find a Store > View All Stores > Any country
2. Open VoiceOver
3. Open Rotor in VoiceOver and select "Headings" (ctrl+opt+U).
4. Select a heading for a store name so it gets focused.
5. Try to follow the link by pressing Enter key.

**Expected behavior**
The user should be able to focus heading using the rotor and follow the link.

**Screenshots**
Screen Shot 2021-07-28 at 11 38 40 AM

OS: BigSur 11.4

- Browser Chrome Version 92.0.4515.107
- Browser Safari Version 14.1.1 (16611.2.7.1.4)

**Additional context**
- Not an issue for JAWS/Chrome (Win)
- Not an issue for NVDA/Chrome (Win)

saad-mo:
closes #13359


saad-mo:
@Matejk00 I tagged you on this one because the markup had your fingerprints on it from some changes a while back :slightly_smiling_face:

I simplified the markup for `cx-store-name` h2 element. I see that were is a button rendered as well as a link. I traced the button makeup to a PR that was addressing some a11y e2e tests (#6751). 

This PR address a problem with the link was not accessible via the SR  rotor links menu. The problem seems to be with the nesting of the anchor tag with the heading tag. The fix was to reverse the nesting so that the heading is now wrapped with the anchor tag. I think this will also fix the original being addressed in #6751, and we don't need to render a button. I could be completely wrong about this assumption so need feedback if this is the case.

Matejk00:
> @Matejk00 I tagged you on this one because the markup had your fingerprints on it from some changes a while back 🙂
> 
> I simplified the markup for `cx-store-name` h2 element. I see that were is a button rendered as well as a link. I traced the button makeup to a PR that was addressing some a11y e2e tests (#6751).
> 
> This PR address a problem with the link was not accessible via the SR rotor links menu. The problem seems to be with the nesting of the anchor tag with the heading tag. The fix was to reverse the nesting so that the heading is now wrapped with the anchor tag. I think this will also fix the original being addressed in #6751, and we don't need to render a button. I could be completely wrong about this assumption so need feedback if this is the case.

Hey @saad-mo :) 

I did a code review after your markup change and found two small issues:
1. When you go to the search results by using for example following link: http://localhost:4200/electronics-spa/en/USD/store-finder/find?query=%20
and try to open any store (by clicking on store name) empty page appears. Probably there is some problem with correct routing path.
2. If you use tab, and tabbing through store items elements, there is no indication on currently focused store name. Would be nice to add default blue outline as is it working for other focusable elements.

Tested on Chrome, let me know if you have any questions.

saad-mo:
> Hey @saad-mo :)
> 
> I did a code review after your markup change and found two small issues:
> 
> 1. When you go to the search results by using for example following link: http://localhost:4200/electronics-spa/en/USD/store-finder/find?query=%20
>    and try to open any store (by clicking on store name) empty page appears. Probably there is some problem with correct routing path.
> 2. If you use tab, and tabbing through store items elements, there is no indication on currently focused store name. Would be nice to add default blue outline as is it working for other focusable elements.
> 
> Tested on Chrome, let me know if you have any questions.

@Matejk00 restored store search results page link functionality. I guess we need both chunks of codes to cover the two scenarios. 

@developpeurweb I simplified the markup further by just including `a > store name` instead of `a > h2 > store name` do we need a `h2` here? it was causing styling issues with focus state and not rendering the outline properly.

Matejk00 (dismissed):

Matejk00:
> @Matejk00 restored store search results page link functionality. I guess we need both chunks of codes to cover the two scenarios.

Tested, now it looks very good, thanks!

developpeurweb:
> 

@saad-mo those links show up well organized on the "links menu" so we can do without the `h2`.

developpeurweb (dismissed):

developpeurweb (approved):

PR #14103, opened 2021-10-25
fix: add global message on sign out (#13700)

### LINKED ISSUE #13700: Missing successful global message when users Sign Out
**Describe the bug**

- Screen reader users wouldn't be able to tell if they have in fact signed out.

- Library version: 4.0.0

**To Reproduce**
2. Sign in as a user.
3. Got My Account > Sign Out
4. Hear the absence of a confirmation message the user has signed out.

**Expected behavior**
- Screen reader users will hear a successful confirmation message as they sign out. E.g _"You have successfully signed out"._

**Screenshots**
Screen Shot 2021-09-13 at 7 50 25 PM

BenjaminEmiliani:
Closes #13700


projects/assets/src/translations/en/user.ts
```diff
@@ -29,4 +29,7 @@ export const user = {
     emailIsRequired: 'Invalid email format',
     emailsMustMatch: 'Email does not match',
   },
+  messages: {
```
marlass: I would rename this translation chunk to `logout`. `messages` are to generic

projects/core/src/auth/user-auth/facade/auth.service.ts
```diff
@@ -92,6 +94,10 @@ export class AuthService {
     return new Promise((resolve) => {
       this.oAuthLibWrapperService.revokeAndLogout().finally(() => {
         this.store.dispatch(new AuthActions.Logout());
+        this.globalMessageService?.add(
+          { key: 'messages.signedOutSuccessfully' },
+          GlobalMessageType.MSG_TYPE_CONFIRMATION
+        );
```
marlass: I would add a parameter to the `coreLogout` method, so with each call, we can decide if we want to show this message or not.

For example. I would not like to show this message if we logout user, because his access token expired.

marlass (changes requested):

projects/assets/src/translations/en/user.ts
BenjaminEmiliani: @marlass  I see what you mean, but in the future I was thinking of adding the successful login message to that chunk as well. So maybe something else like `sessionStatus` or `sessionMessage`?

marlass: `authMessages` could work for me

BenjaminEmiliani: Changed to authMessages

projects/core/src/auth/user-auth/facade/auth.service.ts
BenjaminEmiliani: Added optional parameter.

projects/core/src/auth/user-auth/facade/auth.service.ts
```diff
@@ -10,6 +10,7 @@ import { AuthStorageService } from '../services/auth-storage.service';
 import { OAuthLibWrapperService } from '../services/oauth-lib-wrapper.service';
 import { AuthActions } from '../store/actions/index';
 import { UserIdService } from './user-id.service';
+import { GlobalMessageService, GlobalMessageType } from '@spartacus/core';
```
marlass: These import paths should be relative

projects/core/src/auth/user-auth/facade/auth.service.ts
```diff
@@ -85,13 +87,20 @@ export class AuthService {
   /**
    * Revokes tokens and clears state for logged user (tokens, userId).
    * To perform logout it is best to use `logout` method. Use this method with caution.
+   * @param showGlobalMsg show a successful global message upon sign out.
    */
-  coreLogout(): Promise<void> {
+  coreLogout(showGlobalMsg?: boolean): Promise<void> {
```
marlass:
```suggestion
  coreLogout(showGlobalMsg = true): Promise<void> {
```

Can we provide `true` as a default?

And in this case can you pass `false` value where we call it in `AuthHttpHeaderService` and in `UpdateEmailComponentService`.

For these 2 cases, it doesn't make sense to show this message.

marlass (changes requested):

marlass (dismissed):

dydome (changes requested):
please provide migration doc

docs/migration/5_0.md
```diff
@@ -40,3 +40,12 @@ under the section `headerDisplayOnly`.
...
 
 - `configurator.addToCart.buttonDisplayOnly` was added to `configurator-common.ts`
+
+### AsmAuthService
+
+- `coreLogout()` method was modified to trigger a global message by default. `showGlobalMsg: false` parameter can be passed to opt-out of the default global message.
+
+### AuthService
+
+- One new required constructor parameter `globalMessageService: GlobalMessageService`
```
psmul:
```suggestion
- Added `GlobalMessageService` to constructor.
```

BenjaminEmiliani: @psmul Is this okay:

`GlobalMessageService` is a new, required constructor dependency.

psmul (dismissed):

dydome (dismissed):

hackergil (approved):

PR #14123, opened 2021-10-26
fix: add more SR context to Payment-Details (#13728)

### LINKED ISSUE #13728: During checkout SR users are unable to tell which Payment Method is default, additional or selected
**Describe the bug**
- Screen reader users are unable to tell which Payment Method is default and which one is an additional Payment Method.
- Screen reader users are unable to tell which Payment Method is selected
- Action links are being described as links, because they are semantically links. They should be `<button>`s styled as links, they don't have an `href` to be followed (fixed in #14073).

- Library version: 4.0.0

**To Reproduce**
2. Add items to cart and start checkout process.
4. Tab to Payment Method cards and listen to screen reader feedback.
5. Hear absence of differentiation between Payment Methods.

**Expected behavior**
- Screen reader users can hear which one is the default Payment Method as soon as they tab into it.
- Screen reader users can hear a differentiator for non-default Payment Method.
- Screen reader users can hear which Payment Method is selected, either the default or an additional Payment Method
- Use as per screenshot below `role="region" aria-label="Default Payment Method" tabindex="0"` and 
`role="region" aria-label="Additional Payment Method" tabindex="0"`
- Change all `<a tabindex="0" class="cx-card-link card-link btn-link cx-action-link">...</a>` to `<button class="cx-card-link cx-action-link"> ... </button>`
- As per screenshot below, use `aria-describedby` and an ID to associate the `<button>` with the "Selected" text state. E.g. `<div class="card-header" id="selectedTxt"> Selected </div>` ... ... `<button aria-describedby="selectedTxt" class="cx-card-link cx-action-link">Use this payment</button>`

**Additional context**
- See screenshots for ticket #13724. Similar issue for Shipping Addresses during checkout. So, similar screen reader vocalization is expected. 

**Screenshots**

Screen Shot 2021-09-15 at 10 25 47 AM

BenjaminEmiliani:
Closes #13728


projects/storefrontlib/shared/components/card/card.component.html
```diff
@@ -90,6 +94,7 @@ <h3 *ngIf="content.title" class="cx-card-title">
             (click)="send()"
             (keydown.enter)="send()"
             tabindex="0"
+            [attr.aria-describedby]="index ? 'content-header-' + index : ''"
             >{{ action.name }}</a
           >
           <a
```
developpeurweb: Please make this a `<button>` instead of an anchor. Also, as it is natively focusable, there's no need for `tabindex`

developpeurweb: Works well with VO, JAWS and NVDA.

BenjaminEmiliani: @developpeurweb yes the tab index will be removed and the anchor will be changed to a <button> in the PR  #14335, for 5.0

saad-mo (approved):
LGTM 👍

Issue #13933, opened 2021-09-30
Epic Release: SAP EPD Visualization integration


sam-garland:
Add an integration library that will allow for visual picking of spare parts from a 2D or 3D model.

This will include:
- A Spartacus component that acts as a wrapper around a UI5 `sap.ui.vk.Viewport` that will load 2D/3D data from the SAP EPD Visualization services
- A spare part list component
- A component that provides a responsive layout for the spare part list and viewer
- Services for mapping between commerce cloud product codes and visualizations/scene nodes in SAP EPD Visualization data

## Checklist

The following is a checklist for new epics or features acceptance for Spartacus, based on [our Definition of Done document](https://sap.github.io/spartacus-docs/definition-of-done/)

## General

- [x] Epic has been tested on a production-like server using Spartacus installation script
- [ ] Epic has been tested in all supported browsers and devices (Check list on DoD page)
- [x] Epic has sufficient end to end (cypress) test coverage
- [x] Schematics code deprecations have been provided and tested
- [ ] Epic complies with [Security best practices](https://sap.github.io/spartacus-docs/security-best-practices/)
- [ ] Epic complies with [Accessibility best practices](https://sap.github.io/spartacus-docs/a11y-best-practices/)
- [ ] Epic CSS supports directionality (if required)
- [ ] Epic code scans have been conducted (sonar, whitesource, checkmarx)
- [ ] Epic renders successfully in SSR

## Audits/reviews

- [ ] Architecture review of the epic (with @tobi-or-not-tobi)
- [ ] Security and threat modeling audit/review from the Security team
- [ ] Accessibility audit/review from the UX team

## New Library

If your epic will be introduced or introduces a new library:

- [ ] Library provides schematics to install the library
- [ ] Library modules can be lazy loaded appropriately
- [ ] Library has been tested for a customer using Spartacus installation script

## Sample data

If the new feature requires new or updated sample data for a specific Commerce backend version:

- [ ] Sample data has been added to the sampledata repo and merged to the next version to be released
- [ ] New sample data has been loaded and tested on a dev server
- [ ] New sample data has been loaded and tested in our CI server

## Documentation

- [ ] Documentation input is provided for it to be published on the doc site.

PR #14126, opened 2021-10-27
feat: integration with EPD Visualization (#13933)

sam-garland:
Commit for initial review.
Provides components and services for visual spare parts picking.

At least the following tasks remain:
- Increase unit test coverage
- Add e3e tests
- Add automated accessibility testing
- Update to handle 5 tabs nicely
- Add sample data
- Some UI updates
- Update UI5 version when 1.96.0 released

image


angular.json
```diff
@@ -603,6 +603,43 @@
         }
       }
     },
+    "epd-visualization": {
+      "projectType": "library",
+      "root": "integration-libs/epd-visualization",
+      "sourceRoot": "integration-libs/epd-visualization/src",
+      "prefix": "lib",
```
wangzixi-diablo: I am thinking about the better value for prefix. The term "lib" is too generic, isn't it?

sam-garland: It looks like `cx` is used elsewhere. I will update it.

integration-libs/epd-visualization/src/components/visual-picking/visual-picking-product-filter/visual-picking-product-filter.component.html
```diff
@@ -0,0 +1,33 @@
+<div class="form-group search-wrapper">
+  <input
+    #queryInput
```
wangzixi-diablo: it seems this id is not used later in this file. Do we really need it?

sam-garland: It is not needed. I will remove it.

integration-libs/epd-visualization/src/components/visual-picking/visual-picking-product-filter/visual-picking-product-filter.component.html
```diff
@@ -0,0 +1,33 @@
...
+    [hidden]="filter.length > 0"
+  ></cx-icon>
+
+  <cx-icon
+    [type]="iconTypes.RESET"
+    [attr.aria-label]="
+      'visualPicking.visualPickingProductFilter.resetButton.label' | cxTranslate
+    "
+    (mousedown)="filter = ''"
+    (keydown.enter)="filter = ''"
```
wangzixi-diablo, line 27: here we use mousedown and keydown.enter event to clear filter. How about in mobile device? Can these two events still be triggered when user taps in mobile?

wangzixi-diablo, line 27: here we use mousedown and keydown.enter event to clear filter. How about in mobile device? Can these two events still be triggered when user taps in mobile?

integration-libs/epd-visualization/src/components/visual-picking/visual-picking-product-filter/visual-picking-product-filter.service.ts
```diff
@@ -0,0 +1,67 @@
...
+  /**
+   * Returns an Observable that produces a ProductReference[] each time the filter is updated or the set of product references to filter changes.
+   * @param unfilteredProductReferences$ An Observable that returns the unfiltered ProductReference[] to apply filtering to.
+   * @returns An Observable that produces a ProductReference[] each time the filter is updated or the set of product references to filter changes.
+   */
+  public getFilteredProducts$(
+    unfilteredProductReferences$: Observable<ProductReference[]>
+  ): Observable<ProductReference[]> {
+    return combineLatest([
+      concat(of(''), this.filter$),
```
wangzixi-diablo, line 43: why do we need this dummy "" observable here?

integration-libs/epd-visualization/src/components/visual-picking/visual-picking-product-filter/visual-picking-product-filter.component.html
sam-garland: I have been using the Chrome devtools device toolbar for simulation touch events (and I have used this for testing touch gestures for the animation slider). A tap does seem to get converted into mouse events, so the mousedown handler has been getting invoked for a touch tap in this simulated mode.

sam-garland: I think the mousedown event will be triggered. I have not tested with a real device, but the Chrome devtools touch simulation has been pretty good.

integration-libs/epd-visualization/src/components/visual-picking/visual-picking-product-filter/visual-picking-product-filter.service.ts
sam-garland: combineLatest will only return a value when both of its arguments have produced at least one value.
this.filter$ isn't producing anything until the filter value changes, but we want to get results as soon as there are product references available.

I have tried other approaches because this approach bugs me, but the alternatives I have tried have had issues (although I could be confusing this case with another similar one). I would like to eliminate this if possible.

sam-garland: I have tried this change:

```text
diff --git a/integration-libs/epd-visualization/src/components/visual-picking/visual-picking-product-filter/visual-picking-product-filter.service.ts b/integration-libs/epd-visualization/src/components/visual-picking/visual-picking-product-filter/visual-picking-product-filter.service.ts
index 2011aa729c..61fc33af11 100644
--- a/integration-libs/epd-visualization/src/components/visual-picking/visual-picking-product-filter/visual-picking-product-filter.service.ts
+++ b/integration-libs/epd-visualization/src/components/visual-picking/visual-picking-product-filter/visual-picking-product-filter.service.ts
@@ -1,13 +1,15 @@
 import { EventEmitter, Injectable } from '@angular/core';
 import { Product, ProductReference } from '@spartacus/core';
-import { combineLatest, Observable, of, concat } from 'rxjs';
+import { combineLatest, Observable } from 'rxjs';
 import { filter, map } from 'rxjs/operators';
 
 @Injectable({
   providedIn: 'root',
 })
 export class VisualPickingProductFilterService {
-  constructor() {}
+  constructor() {
+    this.filter$.emit(this.filter);
+  }
 
   /**
    * The current filter value.
@@ -39,10 +41,7 @@ export class VisualPickingProductFilterService {
   public getFilteredProducts$(
     unfilteredProductReferences$: Observable<ProductReference[]>
   ): Observable<ProductReference[]> {
-    return combineLatest([
-      concat(of(''), this.filter$),
-      unfilteredProductReferences$,
-    ]).pipe(
+    return combineLatest([this.filter$, unfilteredProductReferences$]).pipe(
       filter(
         (valuePair: [string, ProductReference[]]) =>
           valuePair[0] !== undefined && valuePair[1] !== undefined
```

But it gives the wrong results. The list that is rendered ends up being initially empty, rather than initially unfiltered. For now I think I will keep the existing code.

sam-garland:
I will need to merge this PR so that other can more easily work in this branch. Thanks for the comments.

PR #14132, opened 2021-10-27
fix: During checkout: missing successful global message when user selects different Shipping Address (#13730)

### LINKED ISSUE #13730: During checkout: missing successful global message when user selects different Shipping Address
**Describe the bug**
- Screen reader users wouldn't be able to tell if they have switched/selected an additional Shipping Address during checkout.

- Library version: 4.0.0

**To Reproduce**

4. Switch Shipping Address from default to an additional one.

**Expected behavior**
- Screen reader users will hear a successful confirmation message saying they have changed/selected a different Shipping Address. E.g _"Shipping Address selection successfully changed"._

**Screenshots**

change2021-09-15_10-18-21

bit172:
closes #13730 

needs migration docs


saad-mo:
The global message works great when you change the shipping address, however, I
Screen Shot 2022-01-13 at 3 08 09 PM
 think we can improve the UX by depressing the "Ship to this address" button for the selected address.

feature-libs/checkout/components/components/shipping-address/shipping-address.component.html
```diff
@@ -30,14 +30,14 @@ <h2 class="cx-checkout-title d-none d-lg-block d-xl-block">
         >
           <div
             class="cx-shipping-address-card-inner"
-            (click)="selectAddress(card.address)"
+            (click)="onAddressCardSelect(card.address)"
```
saad-mo, line 33: just curious to know why a new function was added instead of modifying the existing `selectAddress` to render a global message? 

if the new function is needed, it will also require a test in the in the spec file.

feature-libs/checkout/components/components/shipping-address/shipping-address.component.html
```diff
@@ -30,14 +30,14 @@ <h2 class="cx-checkout-title d-none d-lg-block d-xl-block">
...
           >
             <cx-card
               [border]="true"
               [index]="i"
               [fitToContainer]="true"
               [content]="card.card"
-              (sendCard)="selectAddress(card.address)"
+              (sendCard)="onAddressCardSelect(card.address)"
```
saad-mo, line 40: same as above

feature-libs/checkout/components/components/shipping-address/shipping-address.component.html
bit172: The original function gets called onInit so I wrote a new one so that the message only appears if the button was pressed

saad-mo (approved):
LGTM 👍

RadhepS (approved):

PR #14137, opened 2021-10-27
feature: add AnchorPipe

### LINKED ISSUE #14108: ParagraphComponent does not make anchor link with #id work as expected
How to reproduce:
1. go to `FAQ` page by clicking `Help` in the header
2. click any links (such as "Returns and refunds") will redirect to homepage

The content of the `ParagraphComponent` contains the anchor link, e.g.:
`<a href="#b00">Delivery information</a>` and `<h2><a id="b00"></a>Delivery information</h2>`

But, in html, `[innerHTML]="data.content"` remove the id.

Pucek9:
Closes: #14108


projects/storefrontlib/cms-components/content/paragraph/anchor.pipe.ts
```diff
@@ -0,0 +1,30 @@
+import { Pipe, PipeTransform, Renderer2 } from '@angular/core';
+import { DomSanitizer, SafeHtml } from '@angular/platform-browser';
+import { RoutingService } from '@spartacus/core';
+
+@Pipe({ name: 'cxAnchor' })
+export class AnchorPipe implements PipeTransform {
```
Platonn: Let's add JSDocs describing what it pipe does.

projects/storefrontlib/cms-components/content/paragraph/anchor.pipe.ts
```diff
@@ -0,0 +1,30 @@
...
+      const href = link.getAttribute('href');
+      if (href?.indexOf('#') === 0) {
+        this.renderer.setProperty(
+          link,
+          'href',
+          `${this.routingService.getUrl({})}${href}`
+        );
+      }
+    });
+    return this.sanitizer.bypassSecurityTrustHtml(template.innerHTML);
```
Platonn: Why do we need to call here `this.sanitizer.bypassSecurityTrustHtml()`? It looks suspicious.
If it's mandatory, let's add a // code comment explaining why we need it, as it's not obvious on the first sight.

What about instead just returning the plain string template.innerHTML?

projects/storefrontlib/cms-components/content/paragraph/anchor.pipe.ts
```diff
@@ -0,0 +1,30 @@
...
+    template.innerHTML = html.trim();
+    const linkNodes: NodeList = template.content.querySelectorAll('a');
+
+    Array.from(linkNodes).forEach((link: HTMLAnchorElement) => {
+      const href = link.getAttribute('href');
+      if (href?.indexOf('#') === 0) {
+        this.renderer.setProperty(
+          link,
+          'href',
+          `${this.routingService.getUrl({})}${href}`
```
Platonn: This will strip the query parameters from the current URL, if any present. 
It would be good to preserve them
e.g. test your solution on URL: http://localhost:4200/electronics-spa/en/USD/faq?query=param&and=other

Pucek9: Good catch! Thanks

projects/storefrontlib/cms-components/content/paragraph/anchor.pipe.ts
```diff
@@ -0,0 +1,29 @@
...
+
+/* Enriches anchor links with the current location (path and query params).
+ * Useful for cms-provided content injected via innerHTML.
+ */
+@Pipe({ name: 'cxAnchor' })
+export class AnchorPipe implements PipeTransform {
+  constructor(protected renderer: Renderer2, protected winRef: WindowRef) {}
+
+  protected getPath(anchorId: string): string {
+    const { pathname, search } = this.winRef.location;
```
Platonn: In SSR, the winRef.location has only those 2 properties:
https://github.com/SAP/spartacus/blob/c347ac28bed48a1a3540bf3e043650ef936bee4b/projects/core/src/window/window-ref.ts#L78-L81
What about:
```
const currentUrlWithoutFragment = this.winRef.document.location.href.replace(/#.*$/, '');
return `${currentUrlWithoutFragment}${anchorId}`
```

projects/storefrontlib/cms-components/content/paragraph/paragraph.component.html
```diff
@@ -1 +1,4 @@
-<p *ngIf="component.data$ | async as data" [innerHTML]="data.content"></p>
+<p
+  *ngIf="component.data$ | async as data"
+  [innerHTML]="data.content | cxAnchor"
```
Platonn: The word 'anchor' means a _link_ in general.
The goal of our pipe is very specific - to prepend the current URL to links containing only a hash fragment in the `href` attribute (or so called [URI/URL fragment](https://en.wikipedia.org/wiki/URI_fragment)).

IMHO It would be good to have more precise name. What do you think about the following?
```
[innerHTML]="data.content | cxHashAnchors
export class HashAnchorsPipe {}


[innerHTML]="data.content | cxSupplementHashAnchors
export class SupplementHashAnchorsPipe {}


[innerHTML]="data.content | cxPrefixFragmentUrls
[innerHTML]="data.content | cxEchanceHashAnchors
[innerHTML]="data.content | cxEnrichHashAnchors
[innerHTML]="data.content | cxBuildHashAnchors
[innerHTML]="data.content | cxFixHashAnchors
[innerHTML]="data.content | cxPrefixHashAnchors
[innerHTML]="data.content | cxCurrentUrlHashAnchors
```

projects/storefrontlib/cms-components/content/paragraph/anchor.pipe.ts
```diff
@@ -0,0 +1,29 @@
+import { Pipe, PipeTransform, Renderer2 } from '@angular/core';
+import { WindowRef } from '@spartacus/core';
+
+/* Enriches anchor links with the current location (path and query params).
+ * Useful for cms-provided content injected via innerHTML.
```
Platonn:
```suggestion
/* 
 * Supplements the anchor links that contain only the hash fragment in the `href` attribute,
 * (e.g. `<a href="#someId">`), by prepending the current location (path and query params),
 * so it becomes a link to a full url
 * e.g. `<a href="https://domain.com/current/path?and=query-params#someId">`.
 *
 * This helps to avoid the undesirable navigation to the homepage URL (`/#someId`)
 * when clicking the original link.
 *
 * It's useful for example for cms-provided content passed to the [innerHTML] directive.
```

projects/storefrontlib/cms-components/content/paragraph/paragraph.component.html
Platonn: Decided on 
```
[innerHTML]="data.content | cxSupplementHashAnchors export class SupplementHashAnchorsPipe {}
```

Platonn (approved):

Issue #13433, opened 2021-08-06
Review ng-select labeling and attributes


developpeurweb:
**Describe the bug**

- Automated testing not passing, although this gets vocalized by all Screen Readers.

**Tell us the version of Spartacus**
- Library version: 4.0.0

**Expected behavior**
- Test should pass
- Alternative: Move the label to a closer level from the ng-select.

**Screenshots**
![Screen Shot 2021-08-05 at 6 02 17 PM](https://user-images.githubusercontent.com/15113219/128565751-ddb86242-409a-4e86-a692-d6e21a527390.png)
![Screen Shot 2021-08-05 at 6 01 30 PM](https://user-images.githubusercontent.com/15113219/128565754-1804506b-303f-47e7-bcdc-76dc420f44b7.png)
![Screen Shot 2021-08-05 at 6 00 54 PM](https://user-images.githubusercontent.com/15113219/128565755-4f4a6292-a96c-49b8-9fd9-c55ba0293ce6.png)
![Screen Shot 2021-08-05 at 7 33 10 PM](https://user-images.githubusercontent.com/15113219/128571567-799bbe16-212a-40eb-80e1-6146975508b9.png)

RadhepS:
Will be resolved by tabbing changes

PR #14151, opened 2021-10-29
[BLOCKED #14989] test: Product search a11y tabbing test migration

Zeyber:


Zeyber:
Product list page is not keyboard accessible because the sort button cannot be interacted with. Also, there is an autocomplete value that is generated randomly causing a11y test to fail. Problems with `ng-select`?

image

Issue could relate to https://github.com/SAP/spartacus/issues/13433

Zeyber:
Blocked by https://github.com/SAP/spartacus/issues/14989 due to disruptive changes to happen with domSnapshot()(eg. changing snapshot formats).

PR #14196, opened 2021-11-02
fix: Import Products description in Quick Order is misleading

### LINKED ISSUE #14180: Import Products description in Quick Order is misleading
**Describe the bug**
When using the Import feature in Quick Order component, the description tells user that by importing a CSV file, a new saved cart is created when it is actually creating a new Quick Order list. Bug is only textual (the feature is properly importing products to the quick order list). Found during 4.2 exploratory testing.

- Library version: 4.2 RC - CX 2105 B2C 

**To Reproduce**
1. Go to Quick Order
2. Click on 'Import Products'
3. Read text in bold

**Expected behavior**
Description should be appropriate:
Add products by importing a .CSV file and creating a new Quick Order list.

**Screenshots**
Screen Shot 2021-11-01 at 10 23 19 AM

Pucek9:
Closes #14180

### DIFF
--- a/feature-libs/cart/import-export/assets/translations/en/import-export.i18n.ts
+++ b/feature-libs/cart/import-export/assets/translations/en/import-export.i18n.ts
@@ -16,7 +16,8 @@ export const importEntries = {
 
 export const importEntriesDialog = {
   importProducts: 'Import Products',
-  importProductsSubtitle:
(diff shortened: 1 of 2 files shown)


BenjaminEmiliani (approved):
Looks good to me!

PR #14205, opened 2021-11-02
Move wishlist out of cart main

### LINKED ISSUE #14076: Move Wishlist out of main
Wishlist needs to ultimately depend on saved cart

As a first step, we could 
- Move wishlist the `saved-cart` entry point
- Move wishlist in a new `cart/wishlist` entry point that depends on `save-cart`

WeizhengSap:
close #14076


projects/storefrontapp/src/app/spartacus/features/cart-feature.module.ts
```diff
@@ -24,6 +32,22 @@ import { I18nConfig, provideConfig } from '@spartacus/core';
         fallbackLang: 'en',
       },
     }),
+    provideConfig({
```
plabadie: Perhaps we should give wish list it's own `wishlist-feature.module.ts` file like other features like saved cart?

feature-libs/cart/main/core/store/multi-cart-state.ts
```diff
@@ -18,5 +18,7 @@ export interface StateWithMultiCart {
 export interface MultiCartState {
   carts: StateUtils.EntityProcessesLoaderState<Cart | undefined>;
   active: string | null;
-  wishList: string;
+  index: {
+    [cartType: string]: string;
+  };
```
plabadie, line 23: What is the purpose of the index with a CartType of Active and Wishlist?  The `wishlist` and `active` first level fields seem to already fill this function.

WeizhengSap: `wishlist` is moved to `index`. `active` still in first level, but will be moved in index in #14238

plabadie (approved):

PR #14207, opened 2021-11-02
fix: Intermittent scroll page navigation for iOS

### LINKED ISSUE #14206: Scroll position restoration with intermittent back to top restoration on page routes for iOS
**Describe the bug**
On iphone using chrome, when we click a hamburger menu item, and click the last product on the product listing page, it sometimes does not go back to the top. (no browser controls)

**To Reproduce**
1. Go to hamburger menu and click any category
2. scroll to the last product and click it
3. See pdp scroll restoration not going back to the top

**Expected behavior**
For any navigation (click to other routes / pages), it should always go back to the top of the page

bgambocjaviniar:
Issue only seen in iphone chrome when we navigate in pages from hamburger menu to plp page and to pdp. Pdp page on navigation view gets placed near middle or near bottom 'at times'. (not using browser controls)

**Solution is to wait for the page to finish repainting before navigating to the top of the page + additional 0.1 s for the image load on pdp page that pushes the view (only in ios chrome)**

closes GH-14206

### DIFF
--- a/projects/storefrontapp/src/environments/environment.ts
+++ b/projects/storefrontapp/src/environments/environment.ts
@@ -15,7 +15,7 @@ import { Environment } from './models/environment.model';
 
 export const environment: Environment = {
   production: false,
-  occBaseUrl: buildProcess.env.CX_BASE_URL,
+  occBaseUrl: 'https://40.76.109.9:9002',
   occApiPrefix: '/occ/v2/',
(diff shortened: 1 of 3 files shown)


RadhepS:
Tested on iPhone 13 with latest IOS. It's working as expected now with the latest changes. Consistently loads to the top of the page.

RadhepS (approved):

PR #14211, opened 2021-11-03
GH-14048: Add ghost data 

### LINKED ISSUE #14048: Change Facade layer, that it returns the ghost object.
A ghost configuration object needs to be defined that is returned by the facade layer during the time where the actual load of the data from the backend is still in progress. This can be achieved by using the rxjs 'startWith' operator. See https://github.com/SAP/spartacus/tree/poc/chhiGhostDataConfigurator as an inspiration (ghost data should not be part of service, that is just for conciseness reasons).

Ghost data should

Contain 5 attributes (two simple attribute types should be enough, e.g. radio-button and dropdown listbox)
Contain 10 groups

### LINKED ISSUE #14050: Adapt components to use ghost object, check with config ID.
Outcome of https://github.com/SAP/spartacus/issues/14049.

Adapt components:

- [ ] Form

- [ ] Overview Form
- [ ] Group Menu
- [ ] Group Title
- [ ] Product Title
- [ ] Previous Next Button (should not use a ghost design, it should not be displayed during initial load, check via config ID)
- [ ] Price Summary (should not use a ghost design, it should not be displayed during initial load, check via config ID)

### LINKED ISSUE #14051: Adapt component styling
Adapt styling of all components.
Reuse already defined ghost animation classes from SPA core.

- [ ] Form
- [ ] Overview Form
- [ ] Product Title
- [ ] Group Title
- [x] Group Menu

example: https://codepen.io/schuppel/pen/eYpZKJm
Wires: https://4prsat.axshare.com/#id=lje6is&p=d_configuration_placeholders&g=1
https://4prsat.axshare.com/#id=rhw9kx&p=d_summary_placeholders&g=1 (ignore filter area)

schuppell:
Closes #14048
Closes #14051 
Closes #14050


schuppell:
This PR will be used to do the other changes for Ghost Design as well. Therefor do only merge when all Ghost design tasks are done.

feature-libs/product-configurator/rulebased/core/model/configurator.ghostdata.ts
```diff
@@ -0,0 +1,104 @@
...
+  ghostOverviewGroups.push({
+    id: i + '',
+    groupDescription: 'Ghost group ' + i,
+    attributes: ghostOverviewAttributes,
+  });
+}
+
+export const ghostConfigurationId: string = 'GHOST_CONFIGURATION';
+
+export const ghostConfiguration: Configurator.Configuration = {
```
ChristophHi: We need to expose the ghost configuration to the public API. Means: Introduction of an index.ts, progagation in upper index.ts.
Do not expose configurator.model.ts, because this is exposed differently for augmentation reasons

ChristophHi: I wonder if 'ghost' configuration is a good name for that. Should we use one that refers to 'loading animation'?

feature-libs/product-configurator/rulebased/core/model/configurator.ghostdata.ts
```diff
@@ -0,0 +1,104 @@
+import { Configurator } from './configurator.model';
+import { CommonConfigurator } from '@spartacus/product-configurator/common';
+
+const ghostAttributes: Configurator.Attribute[] = [];
+for (let i = 0; i < 3; i++) {
```
ChristophHi: Why do we need those attributes, I thought we would display the loading animation for the entire config form (so the attributes wouldn't matter at all?)

feature-libs/product-configurator/rulebased/core/model/configurator.ghostdata.ts
```diff
@@ -0,0 +1,104 @@
...
+
+export const ghostConfigurationId: string = 'GHOST_CONFIGURATION';
+
+export const ghostConfiguration: Configurator.Configuration = {
+  configId: ghostConfigurationId,
+  groups: ghostGroups,
+  flatGroups: ghostGroups,
+  interactionState: {},
+  overview: { configId: ghostConfigurationId, groups: ghostOverviewGroups },
+  owner: {
```
ChristophHi: Besser: owner: ConfiguratorModelUtils.createInitialOwner(),

feature-libs/product-configurator/rulebased/core/facade/configurator-commons.service.ts
```diff
@@ -71,7 +83,8 @@ export class ConfiguratorCommonsService {
       select(ConfiguratorSelectors.getConfigurationFactory(owner.key)),
       filter((configuration) =>
         this.configuratorUtils.isConfigurationCreated(configuration)
-      )
+      ),
+      startWith(ghostConfiguration)
```
ChristophHi: There is no test that actually checks on the first emission of ghostConfiguration, I think we could add that. In the tests, you only filter out the ghostConfig in order to let the existing tests pass

ChristophHi (changes requested):

feature-libs/product-configurator/rulebased/core/model/configurator.ghostdata.ts
schuppell: Done

feature-libs/product-configurator/rulebased/core/model/configurator.ghostdata.ts
schuppell: I understood it differently, that we specify the number of attributes and use the attribute HTML DOM and just change the styling to a loading animation

feature-libs/product-configurator/rulebased/core/model/configurator.ghostdata.ts
schuppell: The list calls it ghostData as well. But we could also change the naming

schuppell:
```
/**
   * The ghost data contains an empty list of objects that is used in the UI
   * to render the HTML elements.
   *
   * This list contains 10 items, so that the ghost will show 10 rows by default.
   */
  protected ghostData = { values: new Array(10) } as EntitiesModel<T>;
```

ChristophHi: Ok, if other areas call it like that, we can stay with that

ChristophHi (changes requested):
Group navigation does not work any longer. Clicking on navigation buttons simply doesn't navigate, clicking on group menu leads to exceptions
```
core.js:6456 ERROR TypeError: Cannot read properties of undefined (reading 'id')
    at SafeSubscriber._next (configurator-groups.service.ts:208)
```
Error on addToCart:
```
core.js:6456 ERROR TypeError: Cannot read properties of undefined (reading 'id')
    at ConfiguratorGroupStatusService.setGroupStatusVisited (configurator-group-status.service.ts:78)
```

ChristophHi (dismissed):

ChristophHi (approved):

PR #14216, opened 2021-11-03
fix: SSR - expose getDefaultResolverKey function

znikola:
This PR exposes `getDefaultResolverKey` function which is used by Spartacus only if the `renderKeyResolver` SSR option is not provided.

### DIFF
--- a/core-libs/setup/ssr/optimized-engine/optimized-ssr-engine.ts
+++ b/core-libs/setup/ssr/optimized-engine/optimized-ssr-engine.ts
@@ -12,10 +12,9 @@ import {
 } from './ssr-optimization-options';
 
 /**
- * Returns the default resolver key.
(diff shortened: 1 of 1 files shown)


core-libs/setup/ssr/optimized-engine/optimized-ssr-engine.ts
```diff
@@ -11,6 +11,12 @@ import {
   SsrOptimizationOptions,
 } from './ssr-optimization-options';
 
+/**
+ * Returns the default resolver key.
+ * Used only if `renderKeyResolver` option is not provided.
+ */
```
Platonn, line 17:
```suggestion
/**
 * Returns the full url for the given SSR Request.
 */
```

Platonn (approved):

PR #14219, opened 2021-11-03
feat: Add accessibility pattern to quick order (#14187,#13654)

### LINKED ISSUE #13654: Quick Order: Accessibility Pattern revision
**Describe the bug**

### LINKED ISSUE #14187: Search box on Quick Order page doesn't support keyboard navigation
**Describe the bug**
- Tabbing away from the suggestions listbox leaves it open.
- X button is not keyboard accessible, therefore search string can't be cleared.
- Suggestions listbox should not receive focus as a group.
- While the Suggestions listbox is open and focused, and trying to close it with ESC, at some point, the focus jumps up to the skippers

- Library version: 4.2.RC0

**To Reproduce**
1. Go to the Quick Order page and start tabbing to reach the search box.
2. Once in the search box type something to trigger suggestions. E.g. "cameras"
3. Tab again to reach the "X" button, to try and clear the search string.
4. Notice the suggestion listbox getting focused as a group.
4. Tab again to move out of the suggestions box.
5. See both errors.

**Expected behavior**
- Tabbing away from the suggestions listbox should close it.
- X button should be keyboard accessible, in order to clear the search string.
- **Same behaviour as in Main Search.**

**Screenshots**

- Suggestions listbox receveing focus as group, skipping the "X" button:

keyboard-nav
- Suggestions listbox stays open while tabbing away:

stays-open
hidden-focus-ring

- Hitting ESC while focusing the suggestions listbox makes the focus jump to Skippers.

Screen Shot 2021-11-01 at 8 03 32 PM

dydome:
close #14187


developpeurweb (dismissed):
LGTM.
@dydome However, please note this only fixes `GH-14187`, but **not** `GH-13654` which is a Screen Reader ticket.

developpeurweb (dismissed):
@dydome Pls don't forget to remove `GH-13654` from this PR which is a Screen Reader ticket that needs to be cross-screen-reader tested.

developpeurweb (dismissed):

PR #14231, opened 2021-11-04
feat: Add loading strategy input to media component

### LINKED ISSUE #13172: Image loading strategy should be possible to set as an instance property
I am using Spartacus version 3.4, and see it is possible to set loading strategy by setting it via the MediaService. Though helpful, it should be possible to pass down through instance property (`@Input`), as some images may require to be eager loaded, while maintaining lazy loading images for the rest.

https://github.com/SAP/spartacus/blob/release/4.0.x/projects/storefrontlib/src/shared/components/media/media.component.ts

An approach akin to just doing:

```
// Use passed down value, or defer to mediaService as default
@Input() loadingStrategy: ImageLoadingStrategy = this.mediaService.loadingStrategy === ImageLoadingStrategy.LAZY
            ? 'lazy'
            : null;
```

Zeyber:
Closes: #13172

### DIFF
--- a/docs/migration/5_0.md
+++ b/docs/migration/5_0.md
@@ -1,0 +1,4 @@
+# Technical Changes in Spartacus 5.0
+
+#### MediaComponent
+- `get loadingStrategy()` changed return type from `string` to `ImageLoadingStrategy | null`.
--- a/docs/migration/5_0.md
+++ b/docs/migration/5_0.md
@@ -1,4 +1,0 @@
-# Technical Changes in Spartacus 5.0
-
-#### MediaComponent
-- `get loadingStrategy()` changed return type from `string` to `ImageLoadingStrategy | null`.
--- a/projects/storefrontlib/shared/components/media/media.component.html
+++ b/projects/storefrontlib/shared/components/media/media.component.html
@@ -4,7 +4,7 @@
   [attr.srcset]="media.srcset"
   [attr.alt]="media.alt"
   [attr.role]="media.role"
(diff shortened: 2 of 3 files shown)


Zeyber:
QA Steps: 
1) Set loading strategy to lazy in `app.module.ts`.
```
provideConfig({
  imageLoadingStrategy: ImageLoadingStrategy.LAZY
} as MediaConfig)
```
2) Set `[loading]` input to `eager` in some media component.
3) Check attribute `loading` is `eager` in image in DOM.

docs/migration/5_0.md
```diff
@@ -0,0 +1,4 @@
+# Technical Changes in Spartacus 5.0
+
+#### MediaComponent
+- `get loadingStrategy()` changed return type from `string` to `ImageLoadingStrategy | null`.
```
Platonn: IMHO we should add a note in migration doc only in the PR which acutally introduces the breaking change. Here we don't.

Btw. it's nicely handled here to avoid breaking changes 😉 

PS. Perhaps you might want to create a ticket to introduce the breaking change and remove deprecated code.

Zeyber: I created a PR for the deprecations as an experiment to see if we can avoid making tickets to explain the fixes for breaking changes. The main risk I see here is the branch becoming stale. Are there other issues you can think of with this approach?
https://github.com/SAP/spartacus/pull/14236/files

Platonn (approved):

PR #14233, opened 2021-11-04
fix: Improve SSR logging related to maxRenderTime (#14198)

znikola:
This PR enables a message log about a hanging render finishing after the maxRenderTime expires (previously it was logged only on the debug level).


znikola:
no longer needed to back-port to rc0

PR #14234, opened 2021-11-04
fix: Domain handling in SSR for render key (#14036)

znikola:
This PR takes the whole domain when resolving the rendering key in SSR.
Without it, the sites which embed the site info in their domain won't render properly (e.g. my-site.rs, my-site.ca, my-site.pl, etc.).


znikola:
no longer needed to back-port to rc0

PR #14244, opened 2021-11-04
fix: focus expand button on dialog close (#14186)

### LINKED ISSUE #14186: After closing the Zoom dialog the focus doesn't return to the triggering button
**Describe the bug**
When using keyboard-only navigation to open the Zoom dialog, and then closing it, the focus does not return to the "Expand image" button that triggered the event. Focus parks somewhere in-between buttons. From that parking spot, the user can tab backwards and get back to the triggering button. 

- Library version: 4.2.RC0

**To Reproduce**
1. Go to the "Expand Image" button using the keyboard
2. Hit Enter or Spacebar to open the Zoom dialog.
3. Hit Enter or Spacebar again to close the Zoom dialog.
4. Focus ring disappears and goes to a place between the "Expand image" and the "Show reviews" buttons.
5. After closing the Zoom dialog Tabbing-back gets the user back to the "Expand image" button.
6. After closing the Zoom dialog Tabbing-forward gets the user forward to the "Show reviews" button.

**Expected behavior**
- Focus should go back to the "Expand Image" button when closing the dialog. Similar to how the Add to Cart button works after closing the dialog.

**Screenshots**
focus-entry
focus-exit

BenjaminEmiliani:
Closes #14186

### DIFF
--- a/feature-libs/product/image-zoom/components/product-image-zoom/product-image-zoom-trigger/product-image-zoom-trigger.component.html
+++ b/feature-libs/product/image-zoom/components/product-image-zoom/product-image-zoom-trigger/product-image-zoom-trigger.component.html
@@ -1,4 +1,8 @@
-<button #expandBtn class="btn btn-link cx-action-link" (click)="triggerZoom()">
+<button
+  #expandButton
+  class="btn btn-link cx-action-link"
+  (click)="triggerZoom()"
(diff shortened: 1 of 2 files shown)


developpeurweb (approved):
keyboard interaction works well and as expected.

PR #14265, opened 2021-11-05
fix: login flaky e2e

### LINKED ISSUE #13153: Fix flaky login e2e test

hackergil:
Fixes #13153


projects/storefrontapp-e2e-cypress/cypress/integration/regression/login.e2e-spec.ts
```diff
@@ -1,22 +1,14 @@
-import { clickHamburger } from '../../helpers/homepage';
 import * as login from '../../helpers/login';
 import { viewportContext } from '../../helpers/viewport-context';
 
 describe('Login', () => {
-  viewportContext(['mobile', 'desktop'], () => {
```
RadhepS, line 6: Ran the tests locally 20 times without any failures so looks good!

So I can better understand, why did we remove the desktop version of the test exactly?

hackergil: There are no substancial differences from a UI standpoint between desktop and mobile for this test. Therefore, running it in both viewports just adds execution time but it doesn't bring value.

RadhepS: Makes sense!

RadhepS (approved):

PR #14266, opened 2021-11-05
test: asm e2e flaky test

### LINKED ISSUE #13091: Flaky e2e asm.e2e-spec.ts
Last 7 days on `asm.e2e-spec.ts` [has a high failure rate on develop](https://dashboard.cypress.io/projects/k3nmep/analytics/top-failures/9b5ca8b2-6650-eb29-babc-a0c9d89cbe95-6bd8135b-5cd3-f405-49f7-8ec924b39d61?branches=%5B%7B%22label%22%3A%22develop%22%2C%22suggested%22%3Afalse%2C%22value%22%3A%22develop%22%7D%5D&browsers=%5B%5D&chartRangeMostCommonErrors=%5B%5D&chartRangeSlowestTests=%5B%5D&chartRangeTopFailures=%5B%5D&committers=%5B%5D&cypressVersions=%5B%5D&flaky=%5B%5D&operatingSystems=%5B%5D&runGroups=%5B%5D&specFiles=%5B%5D&status=%5B%7B%22label%22%3A%22Passed%22%2C%22value%22%3A%22PASSED%22%7D%2C%7B%22label%22%3A%22Failed%22%2C%22value%22%3A%22FAILED%22%7D%5D&tags=%5B%5D&timeInterval=WEEK&timeRange=%7B%22startDate%22%3A%222021-07-07%22%2C%22endDate%22%3A%222021-07-14%22%7D&viewBy=TEST_CASE). Let's mark it as flaky and then fix it.

image

hackergil:
Fixes #13091


dydome (dismissed):
I have run it about 20 times locally and all passed test. In this case, I think I can approve it :)

projects/storefrontapp-e2e-cypress/cypress/integration/regression/asm.e2e-spec.ts
```diff
@@ -0,0 +1,239 @@
...
+      cy.get('cx-asm-main-ui').should('not.exist');
+
+      cy.log('--> Agent login');
+      checkout.visitHomePage('asm=true');
+      cy.get('cx-asm-main-ui').should('exist');
+      cy.get('cx-asm-main-ui').should('be.visible');
+
+      agentLogin();
+
+      startCustomerEmulation();
```
plabadie, line 43: Suggestion: starting a customer emulation could use a cy.log

projects/storefrontapp-e2e-cypress/cypress/integration/regression/asm.e2e-spec.ts
```diff
@@ -0,0 +1,239 @@
...
+
+      cy.log('--> Update personal details');
+      cy.selectUserMenuOption({
+        option: 'Personal Details',
+      });
```
plabadie: In the past, I have seen `cy.selectUserMenuOption` being the cause of some flakiness.
If ever we have doubts that it could be the case, we could use `cy.visit` instead since the goal of this test is not to test the menu navigation

projects/storefrontapp-e2e-cypress/cypress/integration/regression/asm.e2e-spec.ts
```diff
@@ -0,0 +1,239 @@
...
+    it.skip('agent logout when user was logged and emulated should restore the session', () => {});
+  });
+});
+
+function listenForAuthenticationRequest(): string {
+  return interceptPost(
+    'csAgentAuthentication',
+    '/authorizationserver/oauth/token'
+  );
+}
```
plabadie, line 153: Why not verify the response code anymore?

projects/storefrontapp-e2e-cypress/cypress/integration/regression/asm.e2e-spec.ts
```diff
@@ -0,0 +1,239 @@
...
+    cy.get('cx-csagent-login-form').should('exist');
+    cy.get('cx-customer-selection').should('not.exist');
+    cy.get('cx-csagent-login-form form').within(() => {
+      cy.get('[formcontrolname="userId"]').type('asagent');
+      cy.get('[formcontrolname="password"]').type('pw4all');
+      cy.get('button[type="submit"]').click();
+    });
+  });
+
+  cy.wait(authRequest);
```
plabadie, line 177: Why not verify the response code 200 anymore when when there are waits?

projects/storefrontapp-e2e-cypress/cypress/support/utils/intercept.ts
```diff
@@ -0,0 +1,67 @@
...
+): string {
+  return interceptAndAlias('DELETE', alias, path, cmsPrefix);
+}
+
+/**
+ * Intercepts a request and creates an alias for it
+ * @param method  http method
+ * @param alias intercept alias
+ * @param path  xhr relative path
+ * @param cmsPrefix true if the call is for a CMS OCC API
```
plabadie: Two comments about the `cmsPrefix` in all the functions in this file:

- The param does not seem related to cms.  Perhaps we use a name that does not reference the cms? (like sitePrefix, baseSitePrefix or soomethinf else).
- Could a default value of `true` improve the design?  I presume the majority of calls need the site-prefix.

projects/storefrontapp-e2e-cypress/cypress/integration/regression/asm.e2e-spec.ts
hackergil: It's implicit. If you don't get a 200, the test will fail. Unless you get an equivalent code (20x for example), where your response will still be valid. We should only validate status codes when it's relevant or important to the test

projects/storefrontapp-e2e-cypress/cypress/integration/regression/asm.e2e-spec.ts
hackergil: Good one. Done.

projects/storefrontapp-e2e-cypress/cypress/support/utils/intercept.ts
hackergil: Renamed to baseSitePrefix.
I marked it as `optional` so that you don't have to provide it every time. To add a default value of `true`, it can't be optional. And if I change the logic, then it will be more confusing I think. Thoughts?

plabadie: A parameter with a default value isn't optional in the the function declaration, but when you call the function, you can omit the argument and the parameter will be assigned the default value, like this example:

```
function calculateTax(price: number, taxRate: number = 0.05): number {
    return price * (1 + taxRate);
}

console.log(calculateTax(100)); // 105
```

hackergil: Got it. Addressed. Thanks for the explanation

plabadie (approved):

PR #14269, opened 2021-11-08
Bugfix/occ 6119

### LINKED ISSUE #14105: CDS sends requests with old consent-reference when consent is withdrawn after being granted
**Describe the bug**
Profile-tag part of the CDS integration adds X-Consent-Reference header to the requests sent to OCC url. When a user grants and revokes the consent, it is expected that this header is not sent. However, the header is sent as of now with the old CR which is wrong.

- Library version: 4.1.x

**To Reproduce**
1. In the latest spartacus site, when a user visits for the first time, requests to back-end (ones that start with OCC_BASE_URL) don't contain the X-Consent-Reference header which is expected.
2. Once the user grants consent, a new CR is generated. From this moment on, all requests to back-end contain `X-Consent-Reference:CR` header.
3. When the user revokes consent a tad later, we would see that the old CR is still being sent as a header to the back-end.

**Expected behavior**

**Screenshots**
Screenshot 2021-10-25 at 16 37 09

 - OS: MacOs
 - Browser: chrome
 - Vers: latest

**Additional context**
Found while addressing duplicate requests to strategy from merchandising service.

kinga-k:
Also fixes [Issue-14105](https://github.com/SAP/spartacus/issues/14105)


bgambocjaviniar (commented):
🟢 

but as mentioned in our private channel, we can't move this until further notice.

bgambocjaviniar (approved):

PR #14272, opened 2021-11-08
[BLOCKED #14989] test: Storefinder a11y tabbing test migration

Zeyber:


Zeyber:
Blocked by #14989 due to disruptive changes to happen with `domSnapshot()`(eg. changing snapshot formats).

PR #14310, opened 2021-11-10
fix: Fix installer occ path generation on windows

### LINKED ISSUE #14255: Windows installation creates incorrect OCC path
**Describe the bug**
When installing with `run.sh` on a windows machine, the OCC path gets the absolute url of c drive and appends it to the OCC url defined in `config.sh` or `config.default.sh`. We need to remove the appended path to correct the OCC url after installing spartacus.

4.2

**To Reproduce**
1. Run `scripts/install/run.sh`.
2. Check `spartacus-configuration.module.ts` in CSR or SSR maps for the OCC value.

**Expected behavior**
OCC url should be as defined in `config.sh`

**Screenshots**

image

 - OS: Windows 10

Zeyber:
Closes: #14255


Zeyber:
QA Steps (should test install on windows and linux machines with git bash and regular terminal to check nothing broke):
1) In gitbash `cd scripts/install`
2) `./run.sh install`
3) Check OCC path is `/occ/v2` in generated `spartacus-configuration.module.ts`

PR #14312, opened 2021-11-10
feature/GH-14276: fix the issue that Cart main feature chunk is broken

### LINKED ISSUE #14276: Cart main feature chunk is broken.
Issue from branch `feature/GH-14172` on PR #14246

There should be a little over 300kB of lazy loaded chunks for the cart main feature.  We have much less than that.

```
default-feature-libs_cart_main_occ_public_api_ts.js                  | -                    |   39.33 kB
feature-libs_cart_main_public_api_ts.js                              | -                    |    3.44 kB
```

WeizhengSap:
close #14276


projects/storefrontlib/context/routing-context.service.ts
```diff
@@ -47,7 +47,7 @@ export class RoutingContextService {
    */
   get<T>(contextToken: ContextToken): Observable<T | undefined> {
     return this.contextTokenMapping$.pipe(
-      map((contextMapping) => {
+      switchMap((contextMapping) => {
```
plabadie, line 50: was this a bug in the original implementation?

WeizhengSap: no, `UnifiedInjector`'s get return `Observable`, but `Injector.get` not.

plabadie: Thank! I didn't know that.

plabadie (approved):

PR #14316, opened 2021-11-11
Feature/spike/gh 14251 extensibility attribute types

### LINKED ISSUE #14251: Provide component for attribute type without the need to copy configurator form
As of 4.2, only the entire configuration form can be exchanged via CMS. It's not possible to directly address the component representing a single attribute type.
That would ease the extension of existing attribute types, also it would be easier to provide a component for a custom attribute type.
(Received that as feedback from a customer project also)

ChristophHi:


feature-libs/product-configurator/rulebased/components/form/configurator-form.component.html
```diff
@@ -27,13 +27,13 @@
...
-          [ownerKey]="configuration.owner.key"
-          (selectionChange)="updateConfiguration($event)"
-        ></cx-configurator-attribute-radio-button>
+        <div
+          [cxInnerComponentsHost]="{
+            attribute: attribute,
+            configuration: configuration,
+            group: currentGroup
+          }"
+        ></div>
```
Platonn: 1. I guess it was also your intention and you took only one component by example just for PoC purpose. But let me clarify so we're on the same page: To my understanding and vision, the `configurator-form.component.html` should contain almost only this element with `cxInnerComponentsHost` directive (+/- surrounding things, if any, e.g. for styling). All concrete attribute components should be listed in the `inner` components configuration. 

2. Ideally, I'd prefer the context-provider mechanism to be a generic and independent mechanism. Independent from specific extensibility mechanisms (outlet, innerComponents, whatever....).
In particular, I'd not pass the context via the `@input` of specific extenisbility directive, like here.

   I described the idea roughly in the ticket: https://github.com/SAP/spartacus/issues/14332, however it needs some work to PoC it, then overcome possible obstacles found and finally polish the API. This doesn't sound like a low hanging fruit. However a very juicy one we definitely want to have soon. 
It's a matter of other competing priorities and team capacity, when we start doing it.

feature-libs/product-configurator/rulebased/components/attribute/types/radio-button/configurator-attribute-radio-button.component.ts
```diff
@@ -12,10 +15,41 @@ export class ConfiguratorAttributeRadioButtonComponent
...
+
+    //this.selectionChange.emit(event);
+    this.configuratorCommonsService.updateConfiguration(
+      this.ownerKey,
+      {
+        ...this.attribute,
+        selectedSingleValue: value,
+      },
+      Configurator.UpdateType.ATTRIBUTE
+    );
```
Platonn: Let me know if I understand correctly the communication model here:
- any input data for this component are passed via the `context`
- but output of this component (e.g. side-effect commands dispatched to services) are communicated directly to the specific services.

This doesn't look bad. At the same time I'm wondering about encapsulating input and output in the context itself (e.g. outputs would be methods of the context). For example the context class `ConfiguratorFormContext`would have a method `updateConfiguration()` (which I guess all those inner components need).

I'm talking here, regardless the _way of passing_ the context class (regardless its a class provided in DI tree, or context passed via directive). I'm focusing here on the idea of the context that encapsulates both input and output. What do you think about it? Does it simplify the architecture/usage/brings benefits or is just an unnecessary abstraction to you?

feature-libs/product-configurator/rulebased/components/form/configurator-form.component.html
ChristophHi: > I guess it was also your intention and you took only one component by example just for PoC purpose. But let me clarify so we're on the same page: To my understanding and vision, the configurator-form.component.html should contain almost only this element with cxInnerComponentsHost directive (+/- surrounding things, if any, e.g. for styling). All concrete attribute components should be listed in the inner components configuration.

Yes, completely agree. I only added the single attribute component in order to try things out

feature-libs/product-configurator/rulebased/components/attribute/types/radio-button/configurator-attribute-radio-button.component.ts
ChristophHi: Yes, your understanding of the communication model is correct.
Not easy to answer your question. 
First for me it felt like we _should_ let the context also handle the output, but TBH I didn't get it to work and just wanted to assess the overall feasibility. 
Now, when I think about it, I am not 100% sure we need that. Our case seems to suggest it, especially because the update was handled by the form before, but calling the facade service from the sub components even reduces the complexity. And when I think of other use cases, I assume that parent/child components are normally more loosely coupled than in our, quite specific case, meaning less need for such a callback option.

So I tend to think that it's not needed, but might fundamentally miss other options to use that concept.

PR #14318, opened 2021-11-11
Feature/gh 14152

Changsuwan:
Add "done" button for read only mode. "Add to cart" button changed to "done" when users take a look at the configuration through order history.


feature-libs/product-configurator/rulebased/components/add-to-cart-button/configurator-add-to-cart-button.component.ts
```diff
@@ -2,16 +2,19 @@ import { ChangeDetectionStrategy, Component } from '@angular/core';
 import {
   GlobalMessageService,
   GlobalMessageType,
+  Order,
   RoutingService,
 } from '@spartacus/core';
+import { OrderFacade } from '@spartacus/order/root';
```
ChristophHi, line 8: We need to state this dependency to the order library also in package.json (also see config check)

feature-libs/product-configurator/rulebased/components/add-to-cart-button/configurator-add-to-cart-button.component.html
```diff
@@ -1,5 +1,5 @@
 <ng-container *ngIf="container$ | async as container">
-  <ng-container *ngIf="!container.routerData.displayOnly">
+  <ng-container *ngIf="!container.routerData.displayOnly; else done">
```
ChristophHi, line 2: Better name 'displayOnly' instead of 'done' (as done is just the description we are using for the button)

feature-libs/product-configurator/rulebased/components/add-to-cart-button/configurator-add-to-cart-button.component.html
```diff
@@ -12,4 +12,16 @@
       </button>
     </div>
   </ng-container>
+  <ng-template #done>
```
ChristophHi, line 15: See above ('displayOnly').

feature-libs/product-configurator/rulebased/components/add-to-cart-button/configurator-add-to-cart-button.component.html
```diff
@@ -12,4 +12,16 @@
...
+    <ng-container *ngIf="order$ | async as order">
```
ChristophHi, line 16: This does not cover the checkout. In this case, we also run this page in displayOnly mode and want to show the done button. Here, the button needs to trigger a browser back.

ChristophHi (changes requested):

PR #14319, opened 2021-11-11
fix: add regions in PDP (#13518)

### LINKED ISSUE #13518: Add region landmarks to Products Details Page
**Describe the bug**

- Screen Reader users must be able to navigate by landmarks, currently, there aren't many nor labelled, therefore relevant sections on the page should be identified with meaningful landmarks names.

- Library version: 4.0.0.

**To Reproduce**
2. Bring up the Rotor or Element List Menu.
3. Choose Landmarks view.
4. See absence of relevant landmarks.

**Expected behavior**

- Use only predefined landmarks https://wiki.wdf.sap.corp/wiki/display/spar/Landmarks
- `<cx-carousel role="region" aria-label="Carousel, Images of product name">`
- `<cx-tab-paragraph-container role="region" aria-label="Tab group with more product details">`

BenjaminEmiliani:
Closes #13518


developpeurweb (dismissed):

feature-libs/product/image-zoom/components/product-image-zoom/product-image-zoom-product-images/product-image-zoom-product-images.component.ts
```diff
@@ -13,11 +13,22 @@ import {
 export class ProductImageZoomProductImagesComponent extends ProductImagesComponent {
   expandImage = new BehaviorSubject(false);
   selectedIndex: number | undefined;
+  productName: string | undefined;
 
   constructor(protected currentProductService: CurrentProductService) {
     super(currentProductService);
   }
 
+  getProductName(): string | undefined {
```
Zeyber: Better to get the entire product for easier extensibility. We should also return an observable and fetch in the template using an `async` pipe.

projects/storefrontlib/cms-components/content/tab-paragraph-container/tab-paragraph-container.component.ts
```diff
@@ -34,10 +40,13 @@ export class TabParagraphContainerComponent
 
   tabTitleParams: (Observable<any> | null)[] = [];
 
+  tabPanelContainerRegion: string;
+
   // TODO: it is not used any more, so can be removed in 5.0
   subscription: Subscription;
 
   constructor(
+    protected translationService: TranslationService,
```
Zeyber: This is a breaking change. Service should be optional via `?` and added last as a param.

projects/storefrontlib/cms-components/content/tab-paragraph-container/tab-paragraph-container.component.ts
```diff
@@ -23,6 +24,11 @@ import { BREAKPOINT } from '../../../layout/config/layout-config';
   selector: 'cx-tab-paragraph-container',
   templateUrl: './tab-paragraph-container.component.html',
   changeDetection: ChangeDetectionStrategy.OnPush,
+  host: {
```
Zeyber: This seems like a lot of complexity to have aria attributes. Can we move and clean up this logic in the template?

projects/storefrontlib/cms-components/content/tab-paragraph-container/tab-paragraph-container.component.ts
```diff
@@ -105,6 +115,15 @@ export class TabParagraphContainerComponent
     }
   }
 
+  getRegionLabel(): void {
```
Zeyber: We can use an async pipe if we can move this to the template somehow which would be a lot cleaner.

feature-libs/product/image-zoom/components/product-image-zoom/product-image-zoom-product-images/product-image-zoom-product-images.component.html
```diff
@@ -9,6 +9,11 @@
 
 <ng-container *ngIf="thumbs$ | async as thumbs">
   <cx-carousel
+    role="region"
+    [attr.aria-label]="
+      'carousel.carouselForProduct' | cxTranslate: { product: getProductName() }
+    "
+    tabindex="-1"
```
Zeyber: The `cx-carousel` itself shouldn't be accessible by tab key. Is there an attribute that makes it accessible?

Zeyber (changes requested):

projects/storefrontlib/cms-components/content/tab-paragraph-container/tab-paragraph-container.component.ts
BenjaminEmiliani: I tried but adding screen reader properties does not seem to work with ng-container

feature-libs/product/image-zoom/components/product-image-zoom/product-image-zoom-product-images/product-image-zoom-product-images.component.html
BenjaminEmiliani: Well the `tabindex="-1"` means that the element should be focusable, but should not be reachable via sequential keyboard navigation. The purpose of this attribute is so that focus is kept in the region when user navigates to the region via Landmark Navigation menu 

Does this answer the question?

projects/storefrontlib/cms-components/content/tab-paragraph-container/tab-paragraph-container.component.ts
BenjaminEmiliani: I have not been able to successfully add SR context inside an ng-container due to its nature

feature-libs/product/image-zoom/components/product-image-zoom/product-image-zoom-product-images/product-image-zoom-product-images.component.ts
```diff
@@ -18,6 +20,10 @@ export class ProductImageZoomProductImagesComponent extends ProductImagesCompone
     super(currentProductService);
   }
 
+  getProduct(): Observable<Product> {
```
Zeyber: We don't need this method. `product$` is already exposed thanks to extending `ProductImagesComponent`. We just need to make it public for the template like `product$: Observable<Product> = this.product$;`.

feature-libs/product/image-zoom/components/product-image-zoom/product-image-zoom-product-images/product-image-zoom-product-images.component.html
```diff
@@ -8,14 +8,21 @@
...
-    *ngIf="thumbs.length"
-    class="thumbs"
-    [items]="thumbs"
-    itemWidth="120px"
-    [hideIndicators]="false"
-    [template]="thumb"
-  ></cx-carousel>
+  <ng-container *ngIf="getProduct() | async as product">
+    <cx-carousel
+      role="region"
```
Zeyber, line 13: This is another instance where aria attributes shouldn't really be on the component element tag but rather within it, like the TabParagraphContainer.

Zeyber (changes requested):

feature-libs/product/image-zoom/components/product-image-zoom/product-image-zoom-product-images/product-image-zoom-product-images.component.ts
BenjaminEmiliani: Done

feature-libs/product/image-zoom/components/product-image-zoom/product-image-zoom-product-images/product-image-zoom-product-images.component.html
BenjaminEmiliani: I see what you mean, however, I don't think we want to make every single `cx-carousel` a region. And the accessibility label is also specific to a product: 'Carousel, Images for {{product}}'

So these a11y attributes would not fit the carousels in homepage for example

developpeurweb (dismissed):
Works well with VO, JAWS and NVDA

feature-libs/product/image-zoom/components/product-image-zoom/product-image-zoom-product-images/product-image-zoom-product-images.component.html
```diff
@@ -8,14 +8,21 @@
...
 <ng-container *ngIf="thumbs$ | async as thumbs">
-  <cx-carousel
-    *ngIf="thumbs.length"
-    class="thumbs"
-    [items]="thumbs"
-    itemWidth="120px"
-    [hideIndicators]="false"
-    [template]="thumb"
-  ></cx-carousel>
+  <ng-container *ngIf="product$ | async as product">
```
Zeyber, line 11: Just for safety, could be good to mention the addition of this wrapper and newly exposed `product$` observable.

Zeyber (changes requested):

Zeyber (approved):

PR #14331, opened 2021-11-15
chore: Fix SSR build

znikola:
This PR fixes the build of the SSR setup lib by removing a cyclic dependency.


core-libs/setup/ssr/providers/ssr-providers.ts
```diff
@@ -0,0 +1,47 @@
...
+      deps: [REQUEST],
+    },
+  ];
+}
+
+export function getRequestUrl(req: Request): string {
+  return getRequestOrigin(req) + req.originalUrl;
+}
+
+export function getRequestOrigin(req: Request): string {
```
Platonn: those 2 functions might be used (and one of them is already used) in other contexts than the SSR providers.
Let's move them to a separate file, e.g. ssr/utils/get-request-url.ts, ssr/utils/get-request-origin.ts

Platonn (approved):

PR #14336, opened 2021-11-15
fix: add SR default context to Address Book (#13717)

### LINKED ISSUE #13717: Address card doesn't vocalize the fact that it contains the "default" address.
**Describe the bug**
- Screen reader users are unable to tell which address is default and which one is an additional address.
- Action links are being described as links, because they are semantically links. They should be `<button>`s styled as links, they don't have an `href` to be followed.

- Library version: 4.0.0

**To Reproduce**
1. Go to My Account > Address Book
2. Add 2 different addresses or more.
4. Tab to address cards and listen to screen reader feedback.
5. Hear absence of differentiation between addresses.

**Expected behavior**
- Screen reader users can hear which one is the default address as soon as they tab into it.
- Screen reader users can hear a differentiator for non-default addresses.
- Use as per screenshot below `role="region" aria-label="Default shipping address" tabindex="0"` and 
`role="region" aria-label="Additional shipping address" tabindex="0"`
- Change all `<a tabindex="0" class="cx-card-link card-link btn-link cx-action-link">...</a>` to `<button class="cx-action-link"> ... </button>` (this change should be resolved by PR #14073)

**Screenshots**
Expected screen reader vocalization:

s2021-09-14_18-39-36

s2021-09-14_18-42-24

BenjaminEmiliani:
Closes #13717


saad-mo (commented):
Small UI bug in maybe open a new ticket for it
Screen Shot 2021-11-16 at 8 57 00 AM

saad-mo (commented):
The PR looks good and adds the required changes as described in the ticket. 

However, I am wondering how an SR user would know which address they would be modifying?

For example, if you have 4 addresses saved. One is default and narrated as "default shipping address" when you land in the region. The other addresses are narrated as "additional shipping address". There is no indication to the SR user what address they are on as the address itself is never narrated. Tabbing further only goes through the action buttons (set as default, edit, delete) 

I am wondering if we need some additional aria attributes such as aria-describedby to improve this? or there are ways SR users can initiate narration within a region? 

@developpeurweb any thoughts?

BenjaminEmiliani:
@saad-mo @developpeurweb 
Any thoughts on appending the address line to the region as follows
Screen Shot 2021-11-16 at 2 04 10 PM

developpeurweb:
@saad-mo @BenjaminEmiliani 
Sorry for the late reply.

There's no need to extend the description, once on the region, users will look for details with `VO+Right-arrow`, or just `Down-arrow` for JAWS/NVDA.

A bit simpler for Windows Screen Reader users, they can navigate/skip landmarks, by just pressing the `R` key for JAWS and `D` key for NVDA. So once skipped, and at the wanted region, then it's `Down-arrow` to inspect/read what's inside.

My only suggestion would be to add a number, to differentiate the landmark. E.g. `Additional Address 1`, `Additional Address 2`, otherwise it will be too verbose.

projects/storefrontlib/cms-components/myaccount/address-book/address-book.component.html
```diff
@@ -23,6 +23,9 @@
       <div
         *ngFor="let address of addresses$ | async"
         class="col-md-6 cx-address-card"
+        role="region"
+        tabindex="0"
+        [attr.aria-label]="getRegionLabel(address) | cxTranslate"
```
saad-mo: just need to append a number to an additional shipping address label so that it renders "Additional Shipping address 1", "Additional Shipping address 2", etc...

saad-mo (dismissed):

projects/storefrontlib/cms-components/myaccount/address-book/address-book.component.html
BenjaminEmiliani: Hey @saad-mo,

Yes, I discussed with @bit172 a generic approach to this so that the region and label change is done to the Card component itself and is reusable in any template that uses <cx-card>. This change is being made in PR, #14335.

Will make a few changes to this PR once the other one is merged.

bit172 (approved):
Default and additional addresses get vocalized on VO

PR #14344, opened 2021-11-16
fix: Use useExisting in CmsOccModule

znikola:
This PR switches from using `useClass` to `useExisting` in `CmsOccModule`.


Platonn (approved):
LGTM

Note: for `useExisting` to work, the referenced class need to be provided somewhere, either in the module, or via decorator `@Injectable({providedIn:'root'})`. The latter is the case for `OccCmsPageAdapter` and `OccCmsComponentAdapter`, so it should work.

PR #14354, opened 2021-11-17
chore: Prefix classes with checkout

znikola:
This PR should rename all the classes from /base, /b2b and /scheduled-replenishment to have a "Checkout" prefix.
This will help to differentiate between the new and old class names.


feature-libs/checkout/scheduled-replenishment/components/checkout-place-order/checkout-place-order.component.ts
```diff
@@ -17,15 +17,15 @@ import {
...
 
 @Component({
   selector: 'cx-place-order',
-  templateUrl: './place-order.component.html',
+  templateUrl: './checkout-place-order.component.html',
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
-export class ScheduledReplenishmentPlaceOrderComponent
-  extends PlaceOrderComponent
+export class CheckoutScheduledReplenishmentPlaceOrderComponent
```
bgambocjaviniar, line 27: Just want to make sure it's okay with you too, but the filename for like scheduled thank you and place order is fine to keep right? 

Don't need to prefix scheduled-replenishment and stuff right :d?

bgambocjaviniar (changes requested):

feature-libs/checkout/scheduled-replenishment/components/checkout-place-order/checkout-place-order.component.ts
znikola: I was debating with myself, and I decided to keep it short-ish. If it's located in the repl folder, it should be clear what the components is used for.
Also, not bad to have it like, if someone needs to update a template in the base, maybe they'll notice this cmp and update it as well (🤞🏻)

PR #14369, opened 2021-11-18
fix/quick order order entries context improvement

Pucek9:
Due to low coverage in this file and problems with doing any changes in cart lib without 100% coverage, and difficulties with the testing previous version of code, the following changes have been implemented:

- removed unnecessary subscription and subject
- replaced forkJoin by merge - now we can see real progress when uploading
- changed order of calling methods `this.productConnector.get`, `this.quickOrderService.canAdd` to vice versa
- tests cover all cases now


feature-libs/cart/quick-order/components/page-context/quick-order-order-entries.context.spec.ts
```diff
@@ -76,37 +104,172 @@ describe('QuickOrderOrderEntriesContext', () => {
...
+          productCode: unableToAddProductsData[0].productCode,
+          statusCode: ProductImportStatus.UNKNOWN_IDENTIFIER,
+        },
+      ]);
     });
 
-    it('should not add entries due to limit', () => {
-      canAdd$.next(false);
-      service.addEntries(mockProductData).subscribe();
+    it('should catch unwnown errors', () => {
```
dydome:
```suggestion
    it('should catch unknown errors', () => {
```

dydome (changes requested):

dydome (approved):

PR #14377, opened 2021-11-18
GH-14359: Enable cart and wishlist in the installer

### LINKED ISSUE #14359: Make sure cart/main and cart/wiish-list are installed by schematics
Add what needs to be added.

Install the features with the installation script to max sure int works

plabadie:


projects/schematics/src/shared/constants.ts
```diff
@@ -1081,6 +1081,10 @@ export const CLI_CART_SAVED_CART_FEATURE = 'Saved-Cart';
 
 export const CLI_CART_QUICK_ORDER_FEATURE = 'Quick-Order';
 
+export const CLI_CART_MAIN_FEATURE = 'Cart';
```
WeizhengSap, line 1084: can we call it `main cart` instead of `cart`?

feature-libs/cart/schematics/add-cart/schema.json
```diff
@@ -20,14 +20,34 @@
...
+        "Saved-Cart",
+        "Quick-Order",
+        "Import-Export"
+      ],
       "x-prompt": {
         "message": "Which Cart features would you like to set up?",
         "type": "list",
         "items": [
+          {
+            "value": "Cart",
```
WeizhengSap, line 44: main-cart?

plabadie: This is displayed in the installer "menu".  I'd leave it to cart for now.

projects/schematics/src/shared/constants.ts
plabadie: Has to match the value passed by the installer menu

WeizhengSap (approved):

PR #14400, opened 2021-11-23
feat: Use backoff mechanism in the adapter layer and add tests

bgambocjaviniar:
This PR moves the back-off mechanism from the query service to the adapter layer.
It also adds unit tests as well as an e2e test.


feature-libs/checkout/b2b/core/facade/checkout-payment-type.service.spec.ts
```diff
@@ -118,12 +123,28 @@ describe(`CheckoutPaymentTypeService`, () => {
           done();
         });
     });
+
+    it(`should return an empty array if query's data is falsy`, (done) => {
+      spyOn(service, 'getPaymentTypesState').and.returnValue(
```
bgambocjaviniar, line 129: an example of 2. from DM

left spyOn the actual service, which is not one of the mocked dependency that gets injected to this service.

feature-libs/checkout/b2b/root/checkout-b2b-root.module.ts
```diff
@@ -11,7 +12,7 @@ import {
 } from './feature-name';
 
 export function defaultCheckoutComponentsConfig() {
-  const config = {
+  const config: CmsConfig = {
```
bgambocjaviniar, line 15: REMINDER:

we want to put the dependency in here instead of using schematics to generate the 'x'.feature.module for ll.

feature-libs/checkout/base/components/guards/cart-not-empty.guard.spec.ts
```diff
@@ -1,31 +1,26 @@
...
-  get(a: string) {
-    return `/${a}`;
-  }
+  get = createSpy().and.returnValue('/home');
 }
 
 describe('CartNotEmptyGuard', () => {
   let cartNotEmptyGuard: CartNotEmptyGuard;
-  let activeCartService: ActiveCartServiceStub;
+  let activeCartService: ActiveCartService;
```
znikola, line 23: Nice!

znikola (approved):

feature-libs/checkout/b2b/root/checkout-b2b-root.module.ts
znikola: Added a TODO, thanks!

PR #14407, opened 2021-11-23
chore: Revert setup core lib and old checkout

znikola:
This PR reverts the core-libs' setup changes (adds back the default b2b configs) and it also reverts changes to the old checkout.


core-libs/setup/recipes/b2b/config/default-b2b-occ-config.ts
```diff
@@ -1,6 +1,5 @@
 // We need this import for augmentation of OccEndpoints to pick up
-// TODO#checkout: type CheckoutOccEndpoints breaks build (might need to move this elsewhere for ll?)
-// import { CheckoutOccEndpoints } from '@spartacus/checkout/occ';
+import { CheckoutOccEndpoints } from '@spartacus/checkout/occ';
```
bgambocjaviniar, line 2: Brian to doublecheck

PR #14418, opened 2021-11-23
fix an issue in express checkout

### LINKED ISSUE #14396: Get checkout details only when checkout-related actions are not loading
In `CheckoutDetailsService`, there `loadCheckoutDetails`. We need to check whether the "set delivery address" or "set delivery mode" or "set payment details" is running. `loadCheckoutDetails` only can be run when the loading flags of these processes are `false`.

WeizhengSap:
close #14396


feature-libs/checkout/components/services/express-checkout.service.ts
```diff
@@ -283,17 +299,16 @@ export class ExpressCheckoutService {
...
 
   public trySetDefaultCheckoutDetails(): Observable<boolean> {
     this.clearCheckoutService.resetCheckoutProcesses();
 
-    return combineLatest([this.deliveryModeSet$, this.paymentMethodSet$]).pipe(
-      map(([deliveryModeSet, paymentMethodSet]) =>
-        Boolean(deliveryModeSet && paymentMethodSet)
-      )
+    return this.paymentMethodSet$.pipe(
+      map((paymentMethodSet) => Boolean(paymentMethodSet))
```
plabadie, line 311: Why do we remove the deliveryMode?

plabadie: Ah, I think I see.  In the stream for `this.paymentMethodSet$`, deliveryModeSet is checked.

plabadie (approved):

Issue #14371, opened 2021-11-18
Create a Service for Dynamic Vocalized Feedback (SR improvement)


BenjaminEmiliani:
Discussion with @saad-mo @developpeurweb led to the following proposition:

There are a few outstanding Screen Reader tickets in our aim for 5.0 that require inserting context dynamically for accessibility end users upon an action (place an order, cancel a deletion, etc).

**Issue**
The issue we are facing with such tickets is that when using Window's compatible assistive technologies, JAWS and NVDA, a dynamically added "live-region" (regions exposed to Screen Reader API to make it announce any changes within the region) will NOT be picked up by the API unless it is preloaded in the page and then the content changes.

For this reason, we had to insert the live-region for Global Messages where it is pre-loaded (storefront.component.html) instead of the dynamic component itself (`<div *ngIf="messages$ | async as messages">`). Link to this PR --> #13781.

**Suggestion**
To avoid using Global Messages every time we would like to add more Screen Reader context to specific use cases, we wonder if it would be possible to create a new service similar to Global Message in the sense that we can pre-load it with a live region in the main template and pass the desired message to be read by the Screen Reader. In this service we would not inject a banner as Global Message service does, but instead the message can be contained in a cx-visually-hidden div so no visual changes occur and then remove it after a delay.

Example of selector name: `cx-at-message` (at -> assistive technology)

This service would enable adding more context for visually impaired users much easier for current and future features.

We are open to further discussions and suggestions on this topic.

KateChuen:
Please also tackle the issue mentioned about the payment details from ticket https://github.com/SAP/spartacus/issues/13676

`Payment Details page during checkout. If you input an invalid expiry date (in the past), it does not voice the error message.`

KateChuen:
Tested on my local, using develop branch. 

Talked with Miguel, he said this is already in place, and one scenario to check this is to hear the message when we remove an item from cart.

Confirmed this action is vocalized:
<img width="920" alt="Screen Shot 2022-05-19 at 3 33 54 PM" src="https://user-images.githubusercontent.com/45409515/169388227-e9cfcd61-4cc8-47f4-ad29-6970d4e10555.png">

Issue #13556, opened 2021-08-26
When the user changes the page this doesn't trigger Screen Reader narration


developpeurweb:
**Describe the bug**

- When the user changes the page this doesn't trigger Screen Reader narration, detailing **where the user has landed** and elements on the new page.
- Basically, the focus stays trapped on the last read element of the previous page and there's silence as the user lands on a new page.
- Most Screen Readers start "read mode" on landing, this doesn't happen because the focus is trapped.

**Tell us the version of Spartacus**
- Library version: 4.0.0

**To Reproduce**
1. Open any Screen Reader (VO, NVDA, JAWS)
2. Tab to Film Cameras Menu option, press Enter to follow the link.
3. Notice how even if the user has changed to a new page (Category Listings) the focus remains on the navigation link.
4. In other cases, it seems the focus moves automatically to the <main> area, instead of having their <title> read on landing.

**Pages having this issue (non-exhaustive list):**

- Homepage
- Category Listings
- Product Details Page
- Order Confirmation
- Order History
- Order Details
- Cancel Items
- Cancel Order
- Cancel Order Confirmation
- My Account > All pages

**Expected behavior**
- When landing on a Spartacus page, the page title gets narrated by the Screen Reader.

**Screenshots**
![2021-08-25_20-16-51](https://user-images.githubusercontent.com/15113219/130881048-a259e4a1-ad14-4ce5-8f1c-26bbd8a59d11.jpg)

saad-mo:
@Zeyber any idea who to fix this? can set keyboard focus on page title when the view changes?

KateChuen:
Tested on develop branch, using yarn:
2 Pages where I found it doesn't narrate are: @developpeurweb can you confirm please?
- Login page
- Homepage 

These pages were ok:
- Register Page
- Product listing page 
- Product Details Page
- Cart page
- Checkout page
- Order Confirmation Page
- Order History Page
- Order Details Page
- Cancel Order Page
- Cancel Order Confirmation Page
- Return Order Page
- Return Order Confirmation Page
- Return Order Details Page
- Wishlist page
- Saved carts page
- Quick Order page
- Find a store page
- Contact Us page
- Help Page
- My Account pages (Address Book, Payment Details, Personal Details, Update Password, Email Address)
- Consent Management Page
- My Interests Page
- Notification Preference Page
- My Coupons page
- Close Account Page

developpeurweb:
@KateChuen 
[...]

Issue #13539, opened 2021-08-23
Page title for checkout doesn't change along with checkout steps


developpeurweb:
**Describe the bug**

- Page title stays short and generic throughout the checkout process.
- Automated test detects this

**Tell us the version of Spartacus**
- Library version: 4.0.0

**To Reproduce**
1. Add product to cart
2. Start the checkout process
3. Notice the page title not changing along with checkout steps.

**Expected behavior**
- Page title for each step is meaningful. E.g. Checkout: Choose a payment method

**Screenshots**
![Screen Shot 2021-08-05 at 7 13 31 PM](https://user-images.githubusercontent.com/15113219/130532270-9837c4c3-65ce-44c8-8784-7fea893a9cea.png)

saad-mo:
`pageMetaResolver` in `common.ts` only has one generic title for checkout. This need to be expanded to cover all checkout steps

saad-mo:
@developpeurweb The title of the page and page heading (h1 tag) are currently coupled together. Updating the page title for checkout will also update the page title. I am assuming this won't be an issue? 

<img width="1491" alt="Screen Shot 2021-10-28 at 2 11 53 PM" src="https://user-images.githubusercontent.com/15215400/139320615-b6c45f90-a6f9-4e4f-8480-ad639b44db74.png">

also, it would be helpful if you can validate the titles below for each checkout step

- Checkout: Enter the Shipping Address
- Checkout: Choose a Delivery Mode 
- Checkout: Choose a Payment Method
- Checkout: Review Order Details

developpeurweb:
@saad-mo  Accessibility-wise this is NOT an issue but in fact good practice. However, let's notify designers @Manurod2018 and @vekterdude about this change, as this may have an impact on wires/visuals. Also, in case they want to adjust presentation (size, capitalization, separators, etc.). I have also seen this kind of arrangement using dashes and vertical bars.
- Checkout - Enter the Shipping Address
- Checkout | Enter the Shipping Address

Manurod2018:
If it's not an accessibility issue, I am not sure to understand why we are repeating the name of the steps? It is unnecessary and isn't part of any convention.
FYI, for the Santorini theme, we also have Checkout as Heading level 1 and the step title as Heading level 2. This approach follows the standards of each e-commerce site.

[...]

PR #14425, opened 2021-11-24
feat: Add role alert to store heading

### LINKED ISSUE #13740: Find a Store: On the Store Details page, there is no vocalized feedback of context change when landing
**Describe the bug**
- When landing on a Store Details page, there's no change of context feedback as to where the user has landed or what happened on the screen.

- Library version: 4.0.0.

**To Reproduce**

4. Click on any location having stores
5. When landing on the Store Details page, there is no vocalization this is a new page.

**Expected behavior**

- The best way to achieve this is by unlocking the focus before landing on the Store Details page, and adding the store name to the page title.
- An alternative method, if the previous isn't possible, is to add `role="alert"` to the store name, so it gets vocalized on landing. See example below.

**Screenshots**
Screen Shot 2021-09-15 at 3 48 38 PM

Matejk00:
Closes: #13740


feature-libs/storefinder/components/store-finder-store-description/store-finder-store-description.component.html
```diff
@@ -2,7 +2,7 @@
   <div class="container">
     <div class="row">
       <article class="cx-store col-md-4">
-        <h2>{{ location.displayName || location.name }}</h2>
+        <h2 role="alert">{{ location.displayName || location.name }}</h2>
```
BenjaminEmiliani, line 5: @Matejk00 I assume you have tried using aria-live="assertive" but it did not work on JAWS/NVDA? Because role is very intrusive

Matejk00: Yes, I tried to use `aria-live="assertive"` solution, but had an issue on JAWS. The problem was that title has been not vocalised. 

With `role="alert"` approach vocalising is working on each screen reader.

BenjaminEmiliani: Yeah I am aware of that issue with JAWS. An alternative would be to add a hidden div that contains the text to be read by the screen reader. Try removing 'role="alert"' and adding this element outside the ng-container:

```
<div aria-live="polite" class="cx-visually-hidden">
  {{ location.displayName || location.name }}
</div>
```

Matejk00: Thanks, I tried to use the solution you proposed, but still JAWS cannot read the content in "hidden" div container.
Is it working for you?

Also another thing to consider is: do we need to create special markup in components for a11y purpose only? I noticed that in storefinder we adjust this approach in some areas. Of course if it's working better than `role="alert"` it's totally fine (because no additional alert prefix during vocalisation), but on the other hand it's more code to maintenance in the future and to be honest `cx-visually-hidden` styling is little bit "hacky" :)

Just wanted to know what's your opinion.

BenjaminEmiliani: @Matejk00 
1) Yes the code I proposed does work for me but on Voice Over.

2) I agree that this approach is better for the user, but not ideal from a software design prospective. It is for this reason and this kind of situation that we discussed and proposed #14371. This service would allow us to inject dynamically the text we want the Screen Reader to read in a common spot instead of having hidden divs in each component

Matejk00: > Yes the code I proposed does work for me but on Voice Over.

@BenjaminEmiliani would you mind to test your solution (https://github.com/SAP/spartacus/pull/14425#discussion_r758555957) on JAWS? For me is not working even if I put this near to the top of component's markup.

Not sure about requirements here, if it should work only on Voice Over that's fine, but I thought that we need to cover all Screen Readers: Voice Over, JAWS, NVDA. So for compatibility `role="alert"` approach works better imo.

BenjaminEmiliani: @Matejk00 My issue is that I am only able to test screen readers on VO as I only own a Mac machine. And yes, I do believe that all Screen Reader implementations should be compatible on Screen Reader/machine.

BenjaminEmiliani: When you tried the solution on JAWS, did you add the hidden div element outside the `<ng-container>`?

Matejk00: Yes, I'm trying it outside the `<ng-container>` and with VO works fine, but not working on JAWS and NVDA.
I don't have any other ideas how to resolve this issue to work properly with all Screen Readers, so should we stay with `role="alert"`solution?

Let's also add @developpeurweb into the loop.

developpeurweb: @Matejk00 @BenjaminEmiliani 
- **Solution 1:** More and more I'm convinced the issue for the whole Find a Store section is **focus management**, it's less evident in VO, but very clear in JAWS/NVDA, because on every UI change the **skipper gets vocalized**, so the first tab on every subpage gets the user to the skipper instead of wherever they were. In this case, it would be great to have the focus reset and the `<title>` to update as the user moves along. E.g.: 
  - Find a Store
  - Find a Store - 24 results found
  - Find a Store - Poland, 2 results found
  - Find a Store - SAP Labs Polska
- **Solution 2:** If the above is not possible, then I see `aria-live` not working mainly because `cx-store-finder-grid` and `cx-store-finder-stores-count` are not preloaded, they get created and destroyed on demand. Not even the `aria-live` number of results get vocalized on JAWS/NVDA. So, as Ben said, a dedicated service preloaded and constantly present, at the top or the bottom of the page will help vocalize messages we can't push otherwise. Same as how `cx-global-message` is preloaded, this one works for all SR.
- **Solution 3:** there are tickets addressing similar and previous issues, we could **park this ticket** until we know if any of these tickets provide a solution.
  - GH-13556
  - GH-13539
  - GH-14371
- **Solution 4:** if any of the aforementioned fixes are **NOT possible**. Then, reluctantly, we will have to use `role="alert"` to tell users they have moved to a different page.

PR #14453, opened 2021-11-26
fix: i18n back-end config for http client

### LINKED ISSUE #14372: JSON translations don't work in SSR Spartacus 4.x
Hello,

I am opening a bug after creating a topic [here](https://stackoverflow.com/questions/69977308/backend-translations-dont-work-in-ssr-spartacus).

I have implemented the Lazy Loading of the translations as in this [guide](https://sap.github.io/spartacus-docs/i18n/#lazy-loading). If I run the application in SSR mode it always falls back in CSR, even if I increase the timeout to over 1 minute.
If I remove the backend configuration and restore the static files, the SSR works fine.
This issue occurs on both version 4.1 and 4.2 of Spartacus.

Steps to reproduce:
```
ng new electronics-spa --style=scss --package-manager yarn
cd electronics-spa/
ng add @spartacus/schematics@latest --baseUrl https://spartacus-demo.eastus.cloudapp.azure.com:8443 --baseSite=electronics-spa --currency=USD --language=en --ssr --skip-confirmation
cp ./node_modules/@spartacus/assets/i18n-assets ./src/assets -r
```

(I am using of course mine SAP Cloud baseUrl)

Change the file `src/app/spartacus/spartacus-configuration.module.ts` in this way (as per the documentation):

```
providers: [
  provideConfig({
    i18n: {
      backend: {
        loadPath: 'assets/i18n-assets/{{lng}}/{{ns}}.json',
      },
    },
  }),
];
```

Then finally run
```
yarn build:ssr
yarn serve:ssr
```

And when you open the application in the browser it will fall back in the CSR mode without any error on server side.
If you revert the spartacus-configuration.module.ts and re-build and re-run the application, SSR works fine.

I have created a simple repo that reproduces the error: https://github.com/fgrecchi-aswatson/spartacus-ssr-issue

Thanks
Filippo Grecchi

### LINKED ISSUE #14458: Fix i18n back-end configuration for http client
In 4.0 we've switched from `i18next-xhr-backend` to `i18next-http-backend`, but it seems the config passed to the new library doesn't match the old one.

znikola:
closes #14458

### DIFF
--- a/projects/core/src/i18n/i18next/i18next-init.ts
+++ b/projects/core/src/i18n/i18next/i18next-init.ts
@@ -1,7 +1,10 @@
 import { HttpClient } from '@angular/common/http';
 import { Injectable, OnDestroy } from '@angular/core';
 import { i18n, InitOptions } from 'i18next';
-import i18nextHttpBackend from 'i18next-http-backend';
+import i18nextHttpBackend, {
+  BackendOptions,
+  RequestCallback,
+} from 'i18next-http-backend';
 import { Subscription } from 'rxjs';
 import { tap } from 'rxjs/operators';
 import { ConfigInitializerService } from '../../config/config-initializer/config-initializer.service';
@@ -23,22 +26,22 @@ export function i18nextInit(
         tap((config) => {
           let i18nextConfig: InitOptions = {
             ns: [], // don't preload any namespaces
-            fallbackLng: config.i18n.fallbackLang,
-            debug: config.i18n.debug,
+            fallbackLng: config.i18n?.fallbackLang,
+            debug: config.i18n?.debug,
             interpolation: {
               escapeValue: false,
             },
           };
 
-          if (config.i18n.backend) {
-            i18next.use(i18nextHttpBackend);
+          if (config.i18n?.backend?.loadPath) {
+            i18next = i18next.use(i18nextHttpBackend);
             const loadPath = getLoadPath(
               config.i18n.backend.loadPath,
               serverRequestOrigin
             );
-            const backend = {
+            const backend: BackendOptions = {
               loadPath,
-              ajax: i18nextGetHttpClient(httpClient),
+              request: i18nextGetHttpClient(httpClient),
             };
             i18nextConfig = { ...i18nextConfig, backend };
           }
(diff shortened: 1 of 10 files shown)


projects/core/src/i18n/i18next/i18next-init.ts
```diff
@@ -23,30 +26,30 @@ export function i18nextInit(
...
+            debug: config.i18n?.debug,
             interpolation: {
               escapeValue: false,
             },
           };
 
-          if (config.i18n.backend) {
-            i18next.use(i18nextHttpBackend);
+          if (config.i18n?.backend?.loadPath) {
+            i18next = i18next.use(i18nextHttpBackend);
```
Platonn, line 37: Why do we need this assignment?

projects/core/src/i18n/i18next/i18next-init.ts
```diff
@@ -96,11 +99,21 @@ export class SiteContextI18nextSynchronizer implements OnDestroy {
...
+    _options: BackendOptions,
+    url: string,
+    _payload: object | string,
+    callback: RequestCallback
+  ) => {
     httpClient.get(url, { responseType: 'text' }).subscribe(
-      (data) => callback(data, { status: 200 }),
-      (error) => callback(null, { status: error.status })
+      (data) => callback(null, { status: 200, data }),
+      (error) => callback(error, { data: error, status: error.status })
```
Platonn: Why do we pass the error as the `data`? This property is meant to be valid translation object.
See the original source code of i18next-http-backend:
https://github.com/i18next/i18next-http-backend/blob/217f0306da941e93c6e904974a2c66eae62921ab/lib/index.js#L67-L88

At the same time, I understand it's a typing issue in `i18next-http-backend`, that `data` is a required field in `RequestCallback`. However, I'm in favor of forcing `data` being undefined here.

projects/core/src/i18n/i18next/i18next-init.ts
znikola: Tbh, I didn't check how it behaves without it, but I wanted to align it with their docs: https://github.com/i18next/i18next-http-backend#getting-started

projects/core/src/i18n/i18next/i18next-init.ts
znikola: Good catch.
I'll just pass `null`, as this is what their docs say: 

```text
'err' should be an error
  //            'res' should be an object with a 'status' property and a 'data' property containing a stringified object instance beeing the key:value translation pairs for the
  //            requested language and namespace, or null in case of an error.

```

projects/core/src/i18n/i18next/i18next-init.ts
Platonn: The assignment seems to be not needed.
`i18next.use()` returns `i18next`, just for chaining purposes (ergonomics of the API).

Reference: source code of `i18next.use()` https://github.com/i18next/i18next/blob/cb9bd97e26734b4f0089fd69a110149e769a54e0/src/i18next.js#L265-L298

Platonn (approved):

Issue #13360, opened 2021-08-03
[JAWS-NVDA/Chrome-FF] Bug while browsing search box suggestions


developpeurweb:
**Describe the bug**
While using the Screen Readers detailed below and browsing the search box suggestions with the down-arrow key it will kick the user out of the search box suggestions menu. Typing suggestions are accessible but the user is unable to reach the product suggestions area and the focus moves to the next component (Account).

Bug detected for:
- JAWS/Chrome (Win)
- NVDA/Chrome (Win)
- NVDA/Firefox (Win)

No issue for:
- _JAWS/Firefox (Win)_
- VoiceOver/Safari (Mac)
- VoiceOver/Chrome (Mac)

**Tell us the version of Spartacus**
- Library version: 4.0.0

**To Reproduce**
1. Open any of the Screen Readers having an issue.
2. Tab to search box
3. Type the word "cam" to trigger suggestions.
4. Use down-arrow to move throughout suggestions.
5. After the last typing suggestion and before the first product suggestion the focus moves to the Account component.

**Expected behavior**
Users should be able to move using up/down arrow keys without leaving the suggestions menu, only clicking the reset icon, ESC key or tabbing away should close the menu and move focus to the next/previous component. While using arrows, the focus should remain "trapped".

**Screenshots**
<img width="960" alt="jaws-searchbox" src="https://user-images.githubusercontent.com/15113219/128056624-c45ec504-2882-4930-a953-92406cf2bcc0.png">
<img width="960" alt="nvda-saerchbox" src="https://user-images.githubusercontent.com/15113219/128056629-555d540f-6f46-4532-b332-bd159290dd8a.png">

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser Chrome Version 92.0.4515.107
 - Browser Firefox Version 90.0.2
 - SR: JAWS Professional Edition Version 2018.1811.30
 - SR: NVDA 2021.1

developpeurweb:
Similar issue found in Quick Order search box suggestions, as it is based on Main Search behaviour.

PR #14457, opened 2021-11-29
Feature/gh 13654 v2

### LINKED ISSUE #13654: Quick Order: Accessibility Pattern revision

dydome:
close #13654


developpeurweb (approved):
This works well for VO and NVDA. However, as it is based on the Main Search behaviour, there is a known bug for JAWS. We will address that specific bug in ticket GH-13360. Other than that, it's fine. Approving.

PR #14471, opened 2021-11-29
GH-13737: Lazy Loading Optimization - mini cart

### LINKED ISSUE #13737: Lazy Loading Optimization - mini cart.
Some cart related components display on the home page or very early in the navigation
* mini cart

We need to delay the moment we load all of the cart lib.  We need to spike separating out the mini cart somehow.

Create another lib/entry point (name tbd, `minicart` ?  ).   See if it's worth it.

plabadie:


feature-libs/cart/main/core/services/multi-cart-state-persistence.service.ts
```diff
@@ -60,6 +69,19 @@ export class MultiCartStatePersistenceService implements OnDestroy {
     } else {
       this.store.dispatch(new CartActions.SetActiveCartId(''));
     }
+    this.dispatchActiveCartBrowserStorageChangeEvent(state);
+  }
+
+  protected onPersist(state: { active: string } | undefined) {
```
WeizhengSap, line 75: When `onPersist` is called and event is created, the cart-lib has been already loaded. So, we don't care about whether `ActiveCartBrowserStorageChange`, we care about whether the cart-lib loaded or not. I think the logic is this: if cart-lib is load, we just use the stream from `ActiveCartFacade.getActive()`. If the cart-lib is not loaded, we check the `getCartStateFromBrowserStorage`, and user login or not, then decide which stream to use.

feature-libs/cart/main/root/config/default-cart-config.ts
```diff
@@ -0,0 +1,14 @@
+import { StorageSyncType } from '@spartacus/core';
+import { CartConfig } from './cart-config';
+
+export const defaultCartConfig: CartConfig = {
```
WeizhengSap, line 4: We need this config, so don't remove it.

plabadie:
Replaced by #14577

PR #14489, opened 2021-11-30
chore: Handle errors in checkout-steps to not cause infinite spinner

bgambocjaviniar:
This PR fixes the infinite spinner issue on the payment step.
It also address a case when normalizeHttpError is called with an already normalized error.
Finally, it aligns some subscription management in payment step components.

### DIFF
--- a/feature-libs/checkout/base/components/checkout-payment-method/checkout-payment-method.component.ts
+++ b/feature-libs/checkout/base/components/checkout-payment-method/checkout-payment-method.component.ts
@@ -42,8 +42,7 @@ export class CheckoutPaymentMethodComponent implements OnInit, OnDestroy {
   selectedMethod$: Observable<PaymentDetails>;
   isGuestCheckout = false;
   newPaymentFormManuallyOpened = false;
-  paymentSavingInProgress$: BehaviorSubject<boolean> =
-    new BehaviorSubject<boolean>(false);
+  paymentSavingInProgress$ = new BehaviorSubject<boolean>(false);
   subscriptions: Subscription = new Subscription();
 
   backBtnText = this.checkoutStepService.getBackBntText(this.activatedRoute);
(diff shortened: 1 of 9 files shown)


projects/core/src/util/normalize-http-error.ts
```diff
@@ -13,6 +13,10 @@ import { HttpErrorModel } from '../model/misc.model';
 export function normalizeHttpError(
   error: HttpErrorResponse | any
 ): HttpErrorModel | undefined {
+  if (error instanceof HttpErrorModel) {
```
bgambocjaviniar, line 16: reason for this is we need throw an error in the backoff, but backoff error is already normalized.

P.S - We can't just `return NEVER` instead of `return throwError(stuff)` in the back-off mechanism as it will not get caught in the subscription error of said function that failed.

feature-libs/checkout/base/components/checkout-payment-method/checkout-payment-method.component.ts
```diff
@@ -202,6 +203,7 @@ export class CheckoutPaymentMethodComponent implements OnInit, OnDestroy {
     this.subscriptions.add(
       this.checkoutPaymentFacade.createPaymentDetails(details).subscribe({
         complete: () => this.paymentSavingInProgress$.next(false),
+        error: () => this.paymentSavingInProgress$.next(false),
```
znikola, line 206: Let's add a test for this case

feature-libs/checkout/base/components/checkout-payment-method/checkout-payment-method.component.ts
```diff
@@ -177,6 +177,7 @@ export class CheckoutPaymentMethodComponent implements OnInit, OnDestroy {
     this.subscriptions.add(
       this.checkoutPaymentFacade.setPaymentDetails(paymentDetails).subscribe({
         complete: () => this.paymentSavingInProgress$.next(false),
+        error: () => this.paymentSavingInProgress$.next(false),
```
znikola, line 180: Let's add a test for this case

znikola (changes requested):

znikola (approved):

projects/core/src/util/normalize-http-error.ts
znikola: Discussed offline, good one.

PR #14494, opened 2021-12-01
chore: block continue button when no address is selected in new checkout

bgambocjaviniar:


bgambocjaviniar:
will be done in https://github.com/SAP/spartacus/pull/14495

PR #7247, opened 2020-04-14
refactor: Payment methods component

marlass:
Refactored components a bit to fix related bugs and to make it consistent with other checkout components behavior.

Closes #6839 closes #6820

PR #2189, opened 2019-04-29
Extensible checkout FR1

marlass:

PR #14495, opened 2021-12-01
fix: Improve payment and shipping components

bgambocjaviniar:
PR includes:

- tackling some `TODO:#checkout` from todo.md
- align and refactor checkout payment / shipping component
- fixes shipping allowing not selected address

### DIFF
--- a/docs/migration/5_0-checkout.md
+++ b/docs/migration/5_0-checkout.md
@@ -2,7 +2,7 @@
 
 Checkout refactoring + code splitting was done as part of the initiative of removing `NGRX` dependencies from Spartacus by replacing it with [commands and queries](https://sap.github.io/spartacus-docs/commands-and-queries/#page-title).
 
-As we want to reduce the bundle size in Spartacus, the checkout refactoring also consisted of decoupling the checkout logic into three different entry points: base, b2b, scheduled-replenishment. 
+As we want to reduce the bundle size in Spartacus, the checkout refactoring also consisted of decoupling the checkout logic into three different entry points. 
 
 - @spartacus/checkout/base - offers a basic checkout functionality, such as b2c.
 - @spartacus/checkout/b2b - offers a b2b checkout flow.
--- a/docs/migration/5_0-checkout.md
+++ b/docs/migration/5_0-checkout.md
@@ -1,8 +1,8 @@
 ## Checkout
 
-Checkout refactoring + code splitting was done as part of the initiative of removing `NGRX` dependencies from Spartacus by replacing it with [command and queries](https://sap.github.io/spartacus-docs/commands-and-queries/#page-title).
+Checkout refactoring + code splitting was done as part of the initiative of removing `NGRX` dependencies from Spartacus by replacing it with [commands and queries](https://sap.github.io/spartacus-docs/commands-and-queries/#page-title).
 
-As we want to reduce the bundle size in Spartacus, the checkout refactoring also consisted of decoupling the checkout logic into three different libraries: base, b2b, scheduled-replenishment. 
+As we want to reduce the bundle size in Spartacus, the checkout refactoring also consisted of decoupling the checkout logic into three different entry points: base, b2b, scheduled-replenishment. 
 
 - @spartacus/checkout/base - offers a basic checkout functionality, such as b2c.
 - @spartacus/checkout/b2b - offers a b2b checkout flow.
@@ -11,9 +11,11 @@ As we want to reduce the bundle size in Spartacus, the checkout refactoring also
 Note: Although the checkout libraries might still have some reference to some ngrx actions within the library, it is not possible to fully separate ngrx from checkout until the other libraries from Spartacus get refactored, such as Cart or User libraries.
 
 Benefits of converting to `command and queries`:
-  - easier to extend as everything is in classes, and it is easy to `extends` and override a method, as opposed to ngrx where you can not really extend a reducer or an effect.
-  - instead of using three actions in ngrx, you can subscribe to the command and get the result in the same method call
-  - similarly to command, you can subscribe to a query, and get the current state in the same call (loading, error, or the data itself).
+  - easier to extend as everything is in classes, and it is simple as overriding a method, as opposed to ngrx where you can not really extend a reducer or an effect.
+  -  commands are built in a more reactive way, and return the execution result in the same method call.
+  - similarly to the commands, listening to loading, error, and data state changes are as simple as calling one method and getting all the results in one call.
+
+For more about commands and queries, please see [this](https://sap.github.io/spartacus-docs/commands-and-queries/#page-title)
 
 ### Setup library
 
--- a/docs/migration/5_0-checkout.md
+++ b/docs/migration/5_0-checkout.md
@@ -1,9 +1,19 @@
 ## Checkout
 
-TODO:#checkout - write some introduction here, explaining the motive behind the refactor:
-  - subscribe to command and get the result in the same method call, instead of using three actions in ngrx
-  - the same is for the queries - subscribe to a query, and get the state feedback in the same call (loading, error or the data itself).
-  - easier to extend - everything is in classes, and it's easy to extends / override a method, as opposed to ngrx, where you can't really extend a reducer or an effect
+Checkout refactoring + code splitting was done as part of the initiative of removing `NGRX` dependencies from Spartacus by replacing it with [command and queries](https://sap.github.io/spartacus-docs/commands-and-queries/#page-title).
+
+As we want to reduce the bundle size in Spartacus, the checkout refactoring also consisted of decoupling the checkout logic into three different libraries: base, b2b, scheduled-replenishment. 
+
+- @spartacus/checkout/base - offers a basic checkout functionality, such as b2c.
+- @spartacus/checkout/b2b - offers a b2b checkout flow.
+- @spartacus/checkout/scheduled-replenishment - offers a scheduled replenishment checkout flow.
+
+Note: Although the checkout libraries might still have some reference to some ngrx actions within the library, it is not possible to fully separate ngrx from checkout until the other libraries from Spartacus get refactored, such as Cart or User libraries.
+
+Benefits of converting to `command and queries`:
+  - easier to extend as everything is in classes, and it is easy to `extends` and override a method, as opposed to ngrx where you can not really extend a reducer or an effect.
+  - instead of using three actions in ngrx, you can subscribe to the command and get the result in the same method call
+  - similarly to command, you can subscribe to a query, and get the current state in the same call (loading, error, or the data itself).
 
 ### Setup library
 
--- a/docs/migration/5_0-checkout.md
+++ b/docs/migration/5_0-checkout.md
@@ -2,7 +2,7 @@
 
 Checkout refactoring + code splitting was done as part of the initiative of removing `NGRX` dependencies from Spartacus by replacing it with [commands and queries](https://sap.github.io/spartacus-docs/commands-and-queries/#page-title).
 
-As we want to reduce the bundle size in Spartacus, the checkout refactoring also consisted of decoupling the checkout logic into three different entry points. 
+As we want to reduce the bundle size in Spartacus, the checkout refactoring also consisted of decoupling the checkout logic into three different entry points: 
 
 - @spartacus/checkout/base - offers a basic checkout functionality, such as b2c.
 - @spartacus/checkout/b2b - offers a b2b checkout flow.
--- a/feature-libs/checkout/b2b/components/checkout-review-submit/checkout-review-submit.component.ts
+++ b/feature-libs/checkout/b2b/components/checkout-review-submit/checkout-review-submit.component.ts
@@ -22,7 +22,7 @@ import {
 } from '@spartacus/core';
 import { Card } from '@spartacus/storefront';
 import { combineLatest, Observable } from 'rxjs';
-import { filter, map, switchMap } from 'rxjs/operators';
+import { filter, map } from 'rxjs/operators';
 
 @Component({
   selector: 'cx-review-submit',
@@ -71,22 +71,10 @@ export class B2BCheckoutReviewSubmitComponent extends CheckoutReviewSubmitCompon
     return this.checkoutPaymentTypeFacade.isAccountPayment();
   }
 
-  /**
-   *  TODO:#checkout - since the `getCostCenterState` returns the actual cost center,
-   * there's no need to use userCostCenterService.getActiveCostCenters()?
-   */
   get costCenter$(): Observable<CostCenter | undefined> {
-    return this.userCostCenterService.getActiveCostCenters().pipe(
-      filter((costCenters) => !!costCenters),
-      switchMap((costCenters) =>
-        this.checkoutCostCenterFacade
-          .getCostCenterState()
-          .pipe(
-            map((state) =>
-              costCenters.find((cc) => cc.code === state?.data?.code)
-            )
-          )
-      )
+    return this.checkoutCostCenterFacade.getCostCenterState().pipe(
+      filter((state) => !state.loading && !state.error),
+      map((state) => state.data)
     );
   }
 
--- a/feature-libs/checkout/b2b/components/checkout-shipping-address/checkout-shipping-address.component.html
+++ b/feature-libs/checkout/b2b/components/checkout-shipping-address/checkout-shipping-address.component.html
@@ -51,7 +51,7 @@
         <div class="col-md-12 col-lg-6">
           <button
             class="cx-btn btn btn-block btn-primary"
-            [disabled]="!selectedAddress?.id"
+            [disabled]="!(selectedAddress$ | async)?.id"
             (click)="next()"
           >
             {{ 'common.continue' | cxTranslate }}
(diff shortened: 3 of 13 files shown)


feature-libs/checkout/b2b/occ/config/default-occ-checkout-b2b-config.ts
```diff
@@ -15,9 +15,7 @@ export const defaultOccCheckoutB2BConfig: OccConfig = {
         setDeliveryAddress:
           'orgUsers/${userId}/carts/${cartId}/addresses/delivery',
         paymentTypes: 'paymenttypes',
-        // TODO:#checkout - shouldn't the following endpoints use `orgUsers`?
```
bgambocjaviniar, line 18: Shouldn't be orgUsers, because some b2b endpoints uses users. As Bill explained a while back to me when I asked, he says it's complicated and politic related things in naming for users / orgUsers 🤷‍♂️ 

https://api.cg79x9wuu9-eccommerc1-p4-public.model-t.myhybris.cloud/occ/v2/swagger-ui.html#/B2B%20Carts/replaceOrgCartCostCenter

bgambocjaviniar: I believe he also mentioned it will be unified to 'users' one day if I recall correctly

feature-libs/checkout/b2b/occ/config/default-occ-checkout-b2b-config.ts
```diff
@@ -15,9 +15,7 @@ export const defaultOccCheckoutB2BConfig: OccConfig = {
...
         setCartCostCenter: 'users/${userId}/carts/${cartId}/costcenter',
-        // TODO:#checkout - rename to b2bPlaceOrder? Similar to `scheduleReplenishmentOrder` from `feature-libs/checkout/scheduled-replenishment/occ/config/default-occ-scheduled-replenishment-config.ts`
```
bgambocjaviniar, line 20: Reason why I didn't change it to b2bPlaceOrder is because we just changed the 'channel' (user vs orgUsers). Going by this logic, I would change `setDeliveryAddress` to `setB2bDeliveryAddress` and `getCheckoutDetails` to `getB2bCheckoutDetails`

bgambocjaviniar: `I'm willing to change the keys named above to contain b2b, but let me know otherwise what you think` if it's just for semantics. However, as mentioned in the comment about `scheduledReplenishmentOrder` key, I would say it wasn't named `placeOrder` because we're not placing an order, but scheduling a replenishment order :P. Basically, when we place an order, the order is placed, but when we scheduled one, it is scheduled to place an order at the given date/time.

feature-libs/checkout/b2b/components/checkout-review-submit/checkout-review-submit.component.ts
```diff
@@ -71,22 +71,10 @@ export class B2BCheckoutReviewSubmitComponent extends CheckoutReviewSubmitCompon
     return this.checkoutPaymentTypeFacade.isAccountPayment();
   }
 
-  /**
-   *  TODO:#checkout - since the `getCostCenterState` returns the actual cost center,
```
bgambocjaviniar, line 75: Without the episode, it makes sense to use the super query as it contains the current set cost center for the user's cart

feature-libs/checkout/base/components/checkout-payment-method/checkout-payment-method.component.ts
```diff
@@ -90,16 +87,9 @@ export class CheckoutPaymentMethodComponent implements OnInit, OnDestroy {
       .pipe(
         filter((state) => !state.loading),
         map((state) => state.data),
-        tap((paymentInfo: any) => {
+        tap((paymentInfo) => {
           if (paymentInfo && !!Object.keys(paymentInfo).length) {
-            if (paymentInfo['hasError']) {
```
bgambocjaviniar, line 95: dead code that came from year 1 of Spartacus, and wasn't ever removed. There  is no longer any case where we mutate the response to include hasError attribute to a response

https://github.com/SAP/spartacus/pull/7247/files#diff-6a20f7cb50b2fa544425396960af44eae0a82887f4d4b057d114f36e81aa7f03R89

https://github.com/SAP/spartacus/pull/2189/files#diff-c45eb1dbd0ef5f712ba7188480c4a3d9d79f1b9917857727c4cbb1e9f603f3f5R77

The new way of how these errors got handled is through  interceptors ---- > **This is already handled elsewhere in the `bad-request.handler.ts`** . 
https://github.com/SAP/spartacus/blob/develop/projects/core/src/global-message/http-interceptors/handlers/bad-request/bad-request.handler.ts#L73

Screen Shot 2021-12-01 at 5 24 28 PM
Screen Shot 2021-12-01 at 5 22 59 PM

feature-libs/checkout/b2b/components/checkout-shipping-address/checkout-shipping-address.component.html
```diff
@@ -71,7 +71,6 @@ <h2 class="cx-checkout-title d-none d-lg-block d-xl-block">
         <cx-address-form
           [showTitleCode]="true"
           [setAsDefaultField]="!isGuestCheckout"
-          [addressData]="selectedAddress"
```
bgambocjaviniar, line 74: redundant as we do not edit in checkout steps, therefore not populating the input of the address form. However, in the my-account/address-book, we can edit, therefore addressData is used to populate the input with existing address data

docs/migration/5_0-checkout.md
```diff
@@ -1,9 +1,19 @@
 ## Checkout
 
-TODO:#checkout - write some introduction here, explaining the motive behind the refactor:
-  - subscribe to command and get the result in the same method call, instead of using three actions in ngrx
-  - the same is for the queries - subscribe to a query, and get the state feedback in the same call (loading, error or the data itself).
-  - easier to extend - everything is in classes, and it's easy to extends / override a method, as opposed to ngrx, where you can't really extend a reducer or an effect
+Checkout refactoring + code splitting was done as part of the initiative of removing `NGRX` dependencies from Spartacus by replacing it with [command and queries](https://sap.github.io/spartacus-docs/commands-and-queries/#page-title).
+
+As we want to reduce the bundle size in Spartacus, the checkout refactoring also consisted of decoupling the checkout logic into three different libraries: base, b2b, scheduled-replenishment.
```
znikola:
```suggestion
As we want to reduce the bundle size in Spartacus, the checkout refactoring also consisted of decoupling the checkout logic into three different entry points: 
```

docs/migration/5_0-checkout.md
```diff
@@ -1,9 +1,19 @@
...
+As we want to reduce the bundle size in Spartacus, the checkout refactoring also consisted of decoupling the checkout logic into three different libraries: base, b2b, scheduled-replenishment. 
+
+- @spartacus/checkout/base - offers a basic checkout functionality, such as b2c.
+- @spartacus/checkout/b2b - offers a b2b checkout flow.
+- @spartacus/checkout/scheduled-replenishment - offers a scheduled replenishment checkout flow.
+
+Note: Although the checkout libraries might still have some reference to some ngrx actions within the library, it is not possible to fully separate ngrx from checkout until the other libraries from Spartacus get refactored, such as Cart or User libraries.
+
+Benefits of converting to `command and queries`:
+  - easier to extend as everything is in classes, and it is easy to `extends` and override a method, as opposed to ngrx where you can not really extend a reducer or an effect.
```
znikola:
```suggestion
  - easier to extend as everything is in classes and it simple as overriding a method, as opposed to ngrx where you can not really extend a reducer or an effect.
```

docs/migration/5_0-checkout.md
```diff
@@ -1,9 +1,19 @@
...
+  - instead of using three actions in ngrx, you can subscribe to the command and get the result in the same method call
```
znikola:
```suggestion
  -  commands are build in a more reactive way, and return the execution result in the same method call.
```

docs/migration/5_0-checkout.md
```diff
@@ -1,9 +1,19 @@
...
+  - similarly to command, you can subscribe to a query, and get the current state in the same call (loading, error, or the data itself).
```
znikola:
```suggestion
  - similarly to the commands, listening to loading, error and data state changes is as simple as calling one method and getting all the results in one call.
```

docs/migration/5_0-checkout.md
```diff
@@ -1,9 +1,19 @@
...
```
znikola, line 19:
```suggestion
For more about commands and queries, please see [this](https://sap.github.io/spartacus-docs/commands-and-queries/#page-title)
```

feature-libs/checkout/b2b/components/checkout-review-submit/checkout-review-submit.component.ts
znikola: Nice!

feature-libs/checkout/base/components/checkout-shipping-address/checkout-shipping-address.component.ts
```diff
@@ -54,13 +53,8 @@ export class CheckoutShippingAddressComponent implements OnInit {
...
       map((state) => state.data),
       tap((address) => {
-        if (
-          address &&
-          (this.selectedAddress === undefined ||
-            this.selectedAddress.id !== address.id)
-        ) {
-          this.selectedAddress = address;
-          if (this.forceLoader) {
+        if (address && !!Object.keys(address).length) {
```
znikola: Why do we need `Object.keys(address).length` check?
I know we are doing it in other components. I'm just asking do you know if we get `{}` from back-end in some cases?

feature-libs/checkout/base/components/checkout-shipping-address/checkout-shipping-address.component.ts
```diff
@@ -54,13 +53,8 @@ export class CheckoutShippingAddressComponent implements OnInit {
...
+          if (this.shouldRedirect) {
```
znikola: I think you can just append `this.shouldRedirect` check to the above `if`

todo.md
```diff
@@ -1,13 +1,11 @@
 ## TODO:
 
-1. search for "// TODO:#checkout"
-2. catch refresh bug on b2b (account type on refresh displays the payment method step when it's not supposed to)
```
znikola, line 4: This is the guard problem, isn't it?

feature-libs/checkout/base/components/checkout-payment-method/checkout-payment-method.component.ts
```diff
@@ -90,16 +87,9 @@ export class CheckoutPaymentMethodComponent implements OnInit, OnDestroy {
...
-              Object.keys(paymentInfo).forEach((key) => {
-                if (key.startsWith('InvalidField')) {
-                  this.sendPaymentMethodFailGlobalMessage(paymentInfo[key]);
-                }
-              });
-              // TODO:#checkout this.checkoutService.clearCheckoutStep(3);
```
znikola, line 101: I guess we no longer need to think about this one?

feature-libs/checkout/base/components/checkout-payment-method/checkout-payment-method.component.ts
```diff
@@ -41,21 +39,20 @@ export class CheckoutPaymentMethodComponent implements OnInit, OnDestroy {
   existingPaymentMethods$: Observable<PaymentDetails[]>;
   isLoading$: Observable<boolean>;
   cards$: Observable<{ content: Card; paymentMethod: PaymentDetails }[]>;
-  selectedMethod$: Observable<PaymentDetails>;
+  selectedMethod$: Observable<PaymentDetails | undefined>;
```
znikola, line 42: Nice 👍🏻

feature-libs/checkout/base/components/checkout-payment-method/checkout-payment-method.component.ts
znikola: Cool!
I think we were wondering about this in one of our episodes, and you've solved the mystery now ;)

feature-libs/checkout/b2b/occ/config/default-occ-checkout-b2b-config.ts
znikola: Ok, that settles it!

feature-libs/checkout/b2b/occ/config/default-occ-checkout-b2b-config.ts
znikola: > Reason why I didn't change it to b2bPlaceOrder is because we just changed the 'channel' (user vs orgUsers). Going by this logic, I would change setDeliveryAddress to setB2bDeliveryAddress and getCheckoutDetails to getB2bCheckoutDetails

Yep, I don't mind doing this. I left a todo, and let's keep it in mind. If we decide to do it, let's do it at some point _after_ the cart merge, as we are making some changes to occ configs.

znikola: > Basically, when we place an order, the order is placed, but when we scheduled one, it is scheduled to place an order at the given date/time.

Yep, I get it.

znikola (changes requested):

feature-libs/checkout/base/components/checkout-payment-method/checkout-payment-method.component.ts
bgambocjaviniar: yup same case as being an old dead code

todo.md
bgambocjaviniar: yup my mistake and nice catch

docs/migration/5_0-checkout.md
bgambocjaviniar: done

docs/migration/5_0-checkout.md
bgambocjaviniar: done

docs/migration/5_0-checkout.md
bgambocjaviniar: done

docs/migration/5_0-checkout.md
bgambocjaviniar: done

docs/migration/5_0-checkout.md
bgambocjaviniar: done

feature-libs/checkout/base/components/checkout-shipping-address/checkout-shipping-address.component.ts
bgambocjaviniar: you're right about other components. Just double check in stackblitz for this tiny thing and I wouldn't include it in this case

bgambocjaviniar: done

todo.md
bgambocjaviniar: brought back

feature-libs/checkout/base/components/checkout-shipping-address/checkout-shipping-address.component.ts
bgambocjaviniar: fixed

docs/migration/5_0-checkout.md
```diff
@@ -1,9 +1,21 @@
 ## Checkout
 
-TODO:#checkout - write some introduction here, explaining the motive behind the refactor:
-  - subscribe to command and get the result in the same method call, instead of using three actions in ngrx
-  - the same is for the queries - subscribe to a query, and get the state feedback in the same call (loading, error or the data itself).
-  - easier to extend - everything is in classes, and it's easy to extends / override a method, as opposed to ngrx, where you can't really extend a reducer or an effect
+Checkout refactoring + code splitting was done as part of the initiative of removing `NGRX` dependencies from Spartacus by replacing it with [commands and queries](https://sap.github.io/spartacus-docs/commands-and-queries/#page-title).
+
+As we want to reduce the bundle size in Spartacus, the checkout refactoring also consisted of decoupling the checkout logic into three different entry points: base, b2b, scheduled-replenishment.
```
znikola: Lets just put colon here, as the unordered list on line 7 lists what are the entry points.

```suggestion
As we want to reduce the bundle size in Spartacus, the checkout refactoring also consisted of decoupling the checkout logic into three different entry points:
```

feature-libs/checkout/base/components/checkout-shipping-address/checkout-shipping-address.component.ts
```diff
@@ -53,10 +53,9 @@ export class CheckoutShippingAddressComponent implements OnInit {
       filter((state) => !state.loading),
       map((state) => state.data),
       tap((address) => {
-        if (address && !!Object.keys(address).length) {
-          if (this.shouldRedirect) {
-            this.next();
-          }
+        if (address && this.shouldRedirect) {
+          console.log('who');
```
znikola: Please remove this

znikola (changes requested):

znikola (approved):

docs/migration/5_0-checkout.md
bgambocjaviniar: done 100%

PR #14505, opened 2021-12-02
fix: Handle lazy load trigger cms mapping from multiple feature module

bgambocjaviniar:
Use case:

If there are many feature modules that uses the 'same' cms component to lazy load the feature, but one of the features has no cms configuration, then the mapping will use the latest configuration from the chunk. In this case, b2b checkout was loaded last before the base checkout. B2b checkout has no cms mapping that initializes `CheckoutPaymentDetails` and `CheckoutDeliveryMode`, therefore it's configuration is an empty object.

Example:
base checkout entrypoint gets loaded when `CheckoutPaymentDetails` and `CheckoutDeliveryMode` exist,  and has the cms mapping in here to a component

Whereas, b2b checkout entrypoints contains them as well, but has no cms mapping to a component. 

Therefore, we need to use the latest configuration found from a feature library

**Solution:**

If there are many same cms component, we should always use the latest existing chunk's configuration from any lazy loaded feature module.

Fixes: 
https://github.com/SAP/spartacus/blob/epic/checkout-refactor-c-and-q/todo.md
for 5. and 6. 

```
5. BUG - When doing b2b checkout:
select the credit card payment method
go all the way to the last step (review order)
notice the delivery method is there
go back to the previous step (payment), and refresh browser
go to the last step - the delivery mode is not there.

6. BUG - the payment method infinite spinner:
repeat the steps above
go to the first step (payment method)
notice the infinite spinner
```

### DIFF
--- a/projects/storefrontlib/cms-structure/services/cms-features.service.ts
+++ b/projects/storefrontlib/cms-structure/services/cms-features.service.ts
@@ -33,6 +33,11 @@ export class CmsFeaturesService {
   // maps componentType to feature
   private componentFeatureMap: Map<string, string> = new Map();
 
+  private componentExistingConfigurationMap: Map<
+    string,
+    CmsComponentMapping | undefined
+  > = new Map();
+
   /*
    * Contains either FeatureInstance or FeatureInstance resolver for not yet
    * resolved feature modules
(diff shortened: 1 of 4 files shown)


todo.md
```diff
@@ -113,6 +94,15 @@
       1. the _base_ checkout depends on _cart_
       2. the _b2b_ / _repl_ depend on _base_
       3. when landing on a _b2b_ step, will the order of chunks being loaded be the following: _cart_ first, then _base_, and lastly the _b2b_ chunk?
+8. There's a TS error on the last checkout step (review order) when running b2b. Seems related to import-export. (fix existing issue from develop)
```
bgambocjaviniar, line 97: moved this to near the end as this error is the same behavior from develop

znikola: I think we can remove it altogether, if the issue is present on develop.

projects/storefrontlib/cms-structure/services/cms-features.service.ts
```diff
@@ -169,9 +174,21 @@ export class CmsFeaturesService {
 
     // extract cms components configuration from feature config
     for (const componentType of featureConfig.cmsComponents ?? []) {
+      /**
+       * update same cms component mapping's configuration found
+       * across all feature instance if the configuration exist
+       * in order to use the latest configuration
+       **/
+      if (resolvedConfiguration.cmsComponents?.[componentType]) {
```
znikola, line 182: Can we come up with a test which can test this functionality?
Ideally, we should cover it with a unit test.
An e2e is welcomed as well.

bgambocjaviniar: unit test (will work on it)
e2e (no)

znikola:
Closing, as we decided to use one chunk for all the checkout variants

PR #14515, opened 2021-12-03
fix: add screen reader attribute to page title (#13556)

### LINKED ISSUE #13556: When the user changes the page this doesn't trigger Screen Reader narration

BenjaminEmiliani:
Closes #13556


saad-mo (dismissed):
The changes work to fix the narration issue. However, there is another issue with component caching. not exactly sure how to fix it but here is the description of the issue.

When navigating from a category search page to a landing page and then to another category search page, the bread crumb component has the text from the previous category search page and after a split second the text updates. It's a minor inconvenience for the sighted user, but for SR users, both text values get narrated which is confusing. 

Steps to reproduce: 
1. From the electronics home page click on `Film Cameras` in the top menu and observe "2 results for Film cameras" text and narration.
2. Click on `Homepage` in the top menu and observe "homepage" is narrated among other things 
3. Click on `Camcorders`  in the top menu and observe the initial text of "2 results for Film cameras" in the bread crumb heading and it gets updated to  "6 results for Hand-held Camcorders" after a split second. This also gets picked up by the Screen Reader and announces both the initial value and updated value.  

Maybe there is a way to clear component value when you navigate away from a pace so that the page load, the initial value is empty and the new value is populated.

saad-mo:
adding `debounceTime` to delay the rending of the title definitely improved the SR narration. I would let @developpeurweb make the final call here.

dydome (dismissed):
The problem still persist, may you take a look @BenjaminEmiliani

BenjaminEmiliani:
> The problem still persist, may you take a look @BenjaminEmiliani

@dydome Are you sure you tested on latest push? Currently on this branch the issue is fixed but using debounceTime

dydome:
https://user-images.githubusercontent.com/18736434/155111427-c9a205b8-463a-4f27-ad90-40f77c6830fb.mov

Yes, the problem still persists, please take a look.

dydome (approved):

PR #14521, opened 2021-12-04
chore: Carry over checkout-related features

znikola:
This PR carries over the features added to the old checkout from the moment we started working on it.

### DIFF
--- a/checkout-git-check.sh
+++ b/checkout-git-check.sh
@@ -9,16 +9,13 @@
 # =========================================================
 
 
-# feature-libs/checkout/components/components/payment-method/payment-form/payment-form.component.html
 # feature-libs/checkout/components/components/index.ts
 # feature-libs/checkout/components/components/cost-center/cost-center.module.ts
 # feature-libs/checkout/components/components/payment-type/payment-type.module.ts
 # feature-libs/checkout/components/components/shipping-address/shipping-address.module.ts
 # feature-libs/checkout/components/components/checkout-login/checkout-login.component.html
-# feature-libs/checkout/components/components/payment-method/payment-form/payment-form.component.html
 # feature-libs/checkout/components/order-confirmation/components/guest-register-form/guest-register-form.component.html
 # feature-libs/checkout/components/components/checkout-login/checkout-login.component.html
-# feature-libs/checkout/components/components/payment-method/payment-form/payment-form.component.html
 # feature-libs/checkout/components/order-confirmation/components/guest-register-form/guest-register-form.component.html
 # feature-libs/checkout/components/services/checkout-details.service.spec.ts
 # feature-libs/checkout/components/services/checkout-details.service.ts
--- a/checkout-git-check.sh
+++ b/checkout-git-check.sh
@@ -8,7 +8,6 @@
 # 4. switch back to the working branch, and carry over the changes from the listed files
 # =========================================================
 
-# feature-libs/checkout/components/components/place-order/place-order.component.ts
 # feature-libs/checkout/core/store/effects/checkout.effect.spec.ts
 # feature-libs/checkout/core/store/effects/checkout.effect.ts
 # feature-libs/checkout/core/facade/checkout-delivery.service.spec.ts
--- a/checkout-git-check.sh
+++ b/checkout-git-check.sh
@@ -13,8 +13,6 @@
 # feature-libs/checkout/root/pages/index.ts
 # feature-libs/checkout/root/pages/order-confirmation-order-entries-context.spec.ts
 # feature-libs/checkout/root/pages/order-confirmation-order-entries-context.ts
-# feature-libs/checkout/styles/components/_promotions.scss
-# feature-libs/checkout/styles/components/steps/review/_review-submit.scss
 
 
 # =========================================================
--- a/checkout-git-check.sh
+++ b/checkout-git-check.sh
@@ -9,7 +9,6 @@
 # =========================================================
 
 
-# feature-libs/checkout/components/components/index.ts
 # feature-libs/checkout/components/components/cost-center/cost-center.module.ts
 # feature-libs/checkout/components/components/payment-type/payment-type.module.ts
 # feature-libs/checkout/components/components/shipping-address/shipping-address.module.ts
--- a/checkout-git-check.sh
+++ b/checkout-git-check.sh
@@ -11,8 +11,6 @@
 
 # feature-libs/checkout/components/components/checkout-login/checkout-login.component.html
 # feature-libs/checkout/components/order-confirmation/components/guest-register-form/guest-register-form.component.html
-# feature-libs/checkout/components/components/checkout-login/checkout-login.component.html
-# feature-libs/checkout/components/order-confirmation/components/guest-register-form/guest-register-form.component.html
 # feature-libs/checkout/components/services/checkout-details.service.spec.ts
 # feature-libs/checkout/components/services/checkout-details.service.ts
 # feature-libs/checkout/components/components/checkout-progress/checkout-progress-mobile-bottom/checkout-progress-mobile-bottom.component.ts
--- a/checkout-git-check.sh
+++ b/checkout-git-check.sh
@@ -9,8 +9,6 @@
 # =========================================================
 
 
-# feature-libs/checkout/components/components/checkout-login/checkout-login.component.html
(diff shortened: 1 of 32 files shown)


checkout-git-check.sh
```diff
@@ -0,0 +1,57 @@
+#!/usr/bin/env bash
```
bgambocjaviniar, line 1: 😍

will do my own double check near the end as well.
But from the output of the script, looks like this PR tackles all the recent ones

feature-libs/checkout/base/components/services/express-checkout.service.ts
```diff
@@ -149,16 +157,14 @@ export class ExpressCheckoutService {
...
   }
 
   public trySetDefaultCheckoutDetails(): Observable<boolean> {
-    return combineLatest([this.deliveryModeSet$, this.paymentMethodSet$]).pipe(
-      map(
-        ([deliveryModeSet, paymentMethodSet]) =>
-          deliveryModeSet && paymentMethodSet
-      )
+    return this.paymentMethodSet$.pipe(
+      map((paymentMethodSet) => !!paymentMethodSet)
```
bgambocjaviniar, line 167: 👍  I see they're using Boolean constructor and you decide to use double `!!`

feature-libs/checkout/scheduled-replenishment/components/checkout-order-confirmation-thank-you-message/checkout-order-confirmation-thank-you-message.component.html
```diff
@@ -1,13 +1,13 @@
 <ng-container *ngIf="order$ | async as order">
   <div class="cx-page-header">
-    <h1 class="cx-page-title">
+    <span class="cx-page-title">
```
bgambocjaviniar, line 3: 👍  together with base

feature-libs/checkout/core/store/effects/checkout.effect.ts
```diff
@@ -274,13 +274,8 @@ export class CheckoutEffects {
...
           map(() => {
-            let digits = payload.paymentDetails.cardNumber;
-            digits = digits.substring(digits.length - 4, digits.length);
-            this.globalMessageService?.add(
-              {
-                key: 'paymentMethods.paymentMethodSelectedSucess',
-                params: { digits: digits },
-              },
+            this.globalMessageService.add(
+              { key: 'paymentMethods.paymentMethodSelectedSuccess' },
```
bgambocjaviniar, line 278: wondering why we're not doing the same as the one in develop? 

I'm in favor for not showing the last 4 digits as well, but maybe Benjamin asked Manu about if it should be displayed

bgambocjaviniar: same goes with the listeners for base

bgambocjaviniar (changes requested):

feature-libs/checkout/core/store/effects/checkout.effect.ts
bgambocjaviniar: Also, this thing had created a bug in a way that when we set payment details, it will throw the error message when checkout was cleared. Exist in develop and now in out checkout branch.

Basically, when we're in b2b, and when we navigate to account, it 'clears' checkout. So when we go back to credit card, and go through the checkout steps to payment details, it will fire the global error message. 

P.S. I would say create a TODO in our list to fix later unless you want to fix it in this PR.

feature-libs/checkout/base/root/checkout-root.module.ts
```diff
@@ -40,7 +47,23 @@ export function defaultCheckoutComponentsConfig() {
...
+    RouterModule.forChild([
+      {
+        // @ts-ignore
+        path: null,
+        canActivate: [CmsPageGuard],
+        component: PageLayoutComponent,
+        data: {
+          cxRoute: 'orderConfirmation',
+          cxContext: {
+            [ORDER_ENTRIES_CONTEXT]: OrderConfirmationOrderEntriesContext,
```
bgambocjaviniar, line 61: Just for my knowledge, I'm assuming this cxContext is used to differentiate a order type in order confirmation page, but can't seem to find how it's being used as it's not even being called 🤔 

If this works as how I think it should work, are we going to be dropping the replenishment order thank you component?

checkout-git-check.sh
znikola: Yep, we'll check regularly. Daily? Bi-daily? As frequently as we can.

After we check, we should:
 - update the date in script to match the checked date
 - push to epic directly
 - let each other know we did it

Better to check for the previous day in case something sneaks in after the check.

feature-libs/checkout/base/components/services/express-checkout.service.ts
znikola: Yeah, we're moving from it, it's not justified tbh.

feature-libs/checkout/scheduled-replenishment/components/checkout-order-confirmation-thank-you-message/checkout-order-confirmation-thank-you-message.component.html
znikola: Yes sir!

feature-libs/checkout/core/store/effects/checkout.effect.ts
znikola: > wondering why we're not doing the same as the one in develop?

I'm not comfortable manipulating the credit card info more than we need to on the FE. 
Another reason is: do we really need to specify which payment detail is selected?
In screen readers, when selecting a payment detail, as a user you'll know which one you want to select, and you just need a confirmation if it happened. To me, it seems kind of redundant to specify which payment detail was selected.

> but maybe Benjamin asked Manu about if it should be displayed

I don't see it the [ticket](https://github.com/SAP/spartacus/issues/13729). And the screenshot in the ticket clearly doesn't have any credit card digits.

> Exist in develop and now in out checkout branch

Can you provide the steps for repro? I was trying to repro it, but couldn't.

feature-libs/checkout/base/root/checkout-root.module.ts
znikola: > cxContext is used to differentiate a order type

yes it's used for import / export feature to "know" if it should place the "export to CSV" or "import from CSV" buttons on the page.

> but can't seem to find how it's being used

It is used, I answered it on Slack.

> if this works as how I think it should work, are we going to be dropping the replenishment order thank you component?

yep, that was my idea as well. added a todo for it.

bgambocjaviniar: Will post on the DM. Miguel agrees in keeping the last 4 digits.

feature-libs/checkout/core/store/effects/checkout.effect.ts
bgambocjaviniar: Will post on the DM. Miguel agrees in keeping the last 4 digits.

bgambocjaviniar: > Can you provide the steps for repro? I was trying to repro it, but couldn't.

1 - use credit card, and place order. and restart the checkout flow with card. You'll see the global message gets fired in the payment step because it calls to set default payment.

Another case
- make sure you have address / payment set

1 - choose account and click next to go to shipping address setp
2 - you decide to not checkout as account anymore, so you go back to the payment type step, and choose card
3 - when you get to the payment step, it will throw the global message automatically

(same case for both develop and this branch)

feature-libs/checkout/base/root/checkout-root.module.ts
bgambocjaviniar: to discuss

feature-libs/checkout/core/store/effects/checkout.effect.ts
bgambocjaviniar: ^ very weird as well when using express checkout. So when we get from cart page -> review-submit step, it will throw the error since under the hood, we dispatched the action to set payment details.

znikola: Fixed express checkout.
Let's discuss the other stuff offline.

PR #14556, opened 2021-12-08
fix: Save auth redirectUrl on NavigationStart event

Zeyber:


Zeyber:
Related to #14694. Closing as a better solution has been proposed.

PR #14561, opened 2021-12-08
Add e2e test for bugfix #14480.

nnamlaht:
Please review the additional e2e test for bugfix #14480.

### DIFF
--- a/projects/storefrontapp-e2e-cypress/cypress/integration/vendor/cds/profile-tag/default-events.e2e-spec.ts
+++ b/projects/storefrontapp-e2e-cypress/cypress/integration/vendor/cds/profile-tag/default-events.e2e-spec.ts
@@ -348,7 +348,7 @@ describe('Profile-tag events', () => {
 
 // For some reason these two events interfere with eachother. Only 1 needs to click the allow all banner
 // and the next will then have consent granted
-describe('Consent Changed', () => {
+describe('Consent Changed for anonymous user', () => {
   beforeEach(() => {
     cdsHelper.setUpMocks(strategyRequestAlias);
     navigation.visitHomePage({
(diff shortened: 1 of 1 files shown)


projects/storefrontapp-e2e-cypress/cypress/integration/vendor/cds/profile-tag/default-events.e2e-spec.ts
```diff
@@ -420,6 +420,31 @@ describe('Consent Changed', () => {
...
+      options: {
+        onBeforeLoad: profileTagHelper.interceptProfileTagJs,
+      },
+    });
+    profileTagHelper.waitForCMSComponents();
+    anonymousConsents.clickAllowAllFromBanner();
+    loginHelper.loginAsDefaultUser();
+  });
+  it('should not send a consentgranted = false event on a page refresh', () => {
+    cy.reload();
```
ShankariGiri-I, line 483: When viewing the test in a headed fashion (on chrome) ie `cypress run --config-file cypress.ci.json --spec "cypress/integration/vendor/cds/profile-tag/default-events.e2e-spec.ts" --headed`, I found that somehow the login doesn't actually go through and the check we perform immediately corresponds to an anonymous user.

This could be avoided by waiting for the loginNotification to be sent first and then reloading the page. For inspiration on how to do that, please check: https://github.com/SAP/spartacus/blob/4d17239588c73897e5f2966fd76f05c06fa02655/projects/storefrontapp-e2e-cypress/cypress/integration/vendor/cds/profile-tag/login-notification.e2e-spec.ts#L14. 

Additionally, once I reverted the fix for https://github.com/SAP/spartacus/pull/14480, one would expect this test to fail but it doesn't & hence looking at the headed version helped to identify that we expect things before a login actually happens.

giancorderoortiz:
Closing due to many months of inactivity.
If still relevant, please update PR with latest develop branch, and re-open with meaningful comments so that it can be evaluated by development team.

PR #14565, opened 2021-12-08
fix: Show global message only when the payment method is clicked on

### LINKED ISSUE #13729: During checkout: missing successful global message when user selects different Payment Method

znikola:
fixes #13729

### DIFF
--- a/feature-libs/checkout/components/components/payment-method/payment-method.component.ts
+++ b/feature-libs/checkout/components/components/payment-method/payment-method.component.ts
@@ -162,7 +162,7 @@ export class PaymentMethodComponent implements OnInit, OnDestroy {
       {
         key: 'paymentMethods.paymentMethodSelectedSuccess',
       },
-      GlobalMessageType.MSG_TYPE_CONFIRMATION
+      GlobalMessageType.MSG_TYPE_INFO
     );
     this.checkoutPaymentService.setPaymentDetails(paymentDetails);
   }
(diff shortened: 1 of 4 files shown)


feature-libs/checkout/components/components/payment-method/payment-method.component.ts
```diff
@@ -158,6 +158,12 @@ export class PaymentMethodComponent implements OnInit, OnDestroy {
   }
 
   selectPaymentMethod(paymentDetails: PaymentDetails): void {
+    this.globalMessageService.add(
+      {
+        key: 'paymentMethods.paymentMethodSelectedSuccess',
```
bgambocjaviniar: let's remove success from the key since we're using it to display as 'info' to have it vocalized this way instead.

bgambocjaviniar (changes requested):
As discussed with Miguel,

we fixed the payment method global message bug, where it gets thrown first time when we didn't even select a payment method during the checkout flow with an already existing set payment method.


This PR fixes that problem of bug, however, for future references, we still need to fix things for error thrown and retries to display a proper message (for every other instances in our codebase)

bgambocjaviniar (approved):

PR #14577, opened 2021-12-09
GH-13737: Lazy Loading Optimization - mini cart

### LINKED ISSUE #13737: Lazy Loading Optimization - mini cart.

plabadie:


feature-libs/cart/main/root/components/mini-cart/mini-cart.module.ts
```diff
@@ -17,6 +18,7 @@ import { MiniCartComponent } from './mini-cart.component';
       cmsComponents: {
         MiniCartComponent: {
           component: MiniCartComponent,
+          providers: [{ provide: MiniCartComponentService }],
```
WeizhengSap: `MiniCartComponentService` is `providedIn: 'root'`, so we don't need provide it in component, right?

feature-libs/cart/main/root/components/mini-cart/mini-cart.component.ts
```diff
@@ -0,0 +1,22 @@
...
+
+@Component({
+  selector: 'cx-mini-cart',
+  templateUrl: './mini-cart.component.html',
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class MiniCartComponent {
+  iconTypes = ICON_TYPE;
+
+  quantity$: Observable<number>;
```
WeizhengSap: `quantity$ = this.miniCartComponentService.getQuantity();`, just set it in the component, no need do it in `constructor` (same for `total$`).

feature-libs/cart/main/root/config/cart-config.ts
```diff
@@ -15,6 +15,9 @@ export abstract class CartConfig {
     validation?: {
       enabled?: boolean;
     };
+    storageType?:
```
WeizhengSap: We have ngrx state, so it seems there is no scenario that we need sync data in `sessionStorage`, right? I guess that's why we don't have `storageType` in any config.

projects/core/src/lazy-loading/facade-factory/facade-factory.service.ts
```diff
@@ -119,6 +121,17 @@ export class FacadeFactoryService {
       result[property] = this.get(resolver$, property as string);
     });
 
+    result[PROXY_FACADE_INSTANCE_PROP] = true;
```
WeizhengSap, line 124: This will add a property to each proxy facade created. And the important thing is: proxy facade created, does not mean the `ActiveCartService` is ready. This line `const resolver$ = this.getResolver(feature, facade, async);` triggers the lib loading (`this.featureModules.resolveFeature(feature)`). When lib loaded, we have something like this to replace the proxy facade.
```
{
    provide: ActiveCartFacade,
    useExisting: ActiveCartService,
 },
```
Theoretically, proxy facade created has no relationship with `ActiveCartService` is loaded.

projects/core/src/lazy-loading/facade-factory/facade-factory.service.ts
```diff
@@ -119,6 +121,17 @@ export class FacadeFactoryService {
...
+
     return result;
```
WeizhengSap, line 126: line 115:  `const result: any = new (class extends (facade as any) {})();` that means proxy is `extends` facade, while our facade service implements the facade (e.g. `export class MultiCartService implements MultiCartFacade`). So, we can use `instanceof` to check whether it is proxy or real facade service.
```
this.injector
      .get(ActiveCartFacade)
      .subscribe((value) =>
        console.log('facade: ', value, value instanceof ActiveCartFacade)
      );
```

feature-libs/cart/main/root/components/mini-cart/mini-cart.module.ts
plabadie: fixed

feature-libs/cart/main/root/components/mini-cart/mini-cart.component.ts
plabadie: fixed

feature-libs/cart/main/root/components/mini-cart/mini-cart-component.service.ts
```diff
@@ -0,0 +1,123 @@
...
+      map(
+        (activeCartFacade) =>
+          activeCartFacade !== undefined &&
+          !this.facadeFactoryService.isProxyFacadeInstance(activeCartFacade)
+      ),
+      distinctUntilChanged()
+    );
+  }
+
+  getCartStateFromBrowserStorage(): Observable<{ active: string } | undefined> {
```
WeizhengSap: can we merge functions `getCartStateFromBrowserStorage()` and `browserHasCartInStorage()` into one function `hasActiveCartInStorage()`?

feature-libs/cart/main/root/components/mini-cart/mini-cart-component.service.ts
```diff
@@ -0,0 +1,123 @@
...
+            map((cart) => cart.totalPrice?.formattedValue ?? '')
+          );
+        } else {
+          return of('');
+        }
+      })
+    );
+  }
+
+  watchUntilCartFacadeRequired() {
```
WeizhengSap: can we change its name to `activeCartRequired`? (login user must have active cart; if `cart-lib` loaded or there is `activeCartId`, active cart must exist). There is only one exception that active cart can be absence. That is NOT `hasCartInStorage || isUserLoggedIn || isCartFacadeLoaded`)

feature-libs/cart/main/root/components/mini-cart/mini-cart-component.service.ts
```diff
@@ -0,0 +1,123 @@
...
+} from 'rxjs/operators';
+import { CartConfig } from '../../config/cart-config';
+import { ActiveCartFacade } from '../../facade/active-cart.facade';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class MiniCartComponentService {
+  constructor(
+    protected activeCartFacade: ActiveCartFacade,
```
WeizhengSap, line 27: please clean the constructor parameters, some of them are not used.

plabadie: Thanks.  Good catch.

feature-libs/cart/main/root/components/mini-cart/mini-cart-component.service.ts
```diff
@@ -0,0 +1,123 @@
...
+            filter((cart) => !!cart.totalPrice),
+            map((cart) => cart.totalPrice?.formattedValue ?? '')
+          );
+        } else {
+          return of('');
+        }
+      })
+    );
+  }
+
```
WeizhengSap, line 76: Please add some doc to explain the logic, something like `login user must have active cart...`

feature-libs/cart/main/root/components/mini-cart/mini-cart-component.service.ts
```diff
@@ -0,0 +1,123 @@
...
+  browserHasCartInStorage(): Observable<boolean> {
+    return this.getCartStateFromBrowserStorage().pipe(
+      map((state) => Boolean(state?.active))
+    );
+  }
+
+  isCartFacadeLoaded(): Observable<boolean> {
```
WeizhengSap: can we also change the name?
```
return this.injector.get(ActiveCartFacade).pipe(
      map((facade) => !this.facadeFactoryService.isProxyFacadeInstance(facade)),
      distinctUntilChanged()
    );
```

feature-libs/cart/main/root/config/cart-config.ts
plabadie: I'll remove it.  We'll let the multi cart persistent service impose the storage type since only the local storage makes sense for storing the active cart.

feature-libs/cart/main/root/components/mini-cart/mini-cart-component.service.ts
plabadie: Fixed

feature-libs/cart/main/root/components/mini-cart/mini-cart-component.service.ts
plabadie: Good idea

feature-libs/cart/main/root/components/mini-cart/mini-cart-component.service.ts
```diff
@@ -0,0 +1,120 @@
...
+    );
+  }
+
+  /**
+   * This function supports lazy loading of the cart functoinality's code. We only call
+   * the activeCartFacade if we know there is actually a cart.
+   * Without a cart, we can return a default value and
+   * avoid loading the cart library code.
+   */
+  getTotalPrice(): Observable<string> {
```
WeizhengSap, line 62: except these 2 get functions, others can be `protected`, right?

feature-libs/cart/main/root/components/mini-cart/mini-cart-component.service.ts
plabadie: done

feature-libs/cart/main/root/components/mini-cart/mini-cart-component.service.ts
plabadie: done

feature-libs/cart/main/root/components/mini-cart/mini-cart-component.service.ts
plabadie: I guess they can.  Their purpose is really to help the 2 get functions accomplish their task.

feature-libs/cart/main/root/components/mini-cart/mini-cart-component.service.spec.ts
```diff
@@ -0,0 +1,352 @@
...
+        {
+          provide: StatePersistenceService,
+          useClass: MockStatePersistenceService,
+        },
+        {
+          provide: SiteContextParamsService,
+          useClass: MockSiteContextParamsService,
+        },
+        {
+          provide: CartConfig,
```
WeizhengSap: this can be removed

feature-libs/cart/main/root/components/mini-cart/mini-cart-component.service.ts
plabadie: Renamed it to `isActiveCartFacadeImplProvided`

feature-libs/cart/main/root/components/mini-cart/mini-cart-component.service.spec.ts
plabadie: Yes.  Good catch.

plabadie: done

feature-libs/cart/main/root/components/mini-cart/mini-cart-component.service.ts
plabadie: I made the protected

feature-libs/cart/main/root/components/mini-cart/mini-cart-component.service.ts
plabadie: Each function has it's different purpose and it also makes unit testing simipler/more straightforward.  I'd like to keep them as-is.

WeizhengSap (approved):

PR #14580, opened 2021-12-09
Move cart context interfaces to root

### LINKED ISSUE #14576: avoid importing cart components in other libs
importing cart's main components in other libs make the lib splited into several chunks.

WeizhengSap:
close #14576


feature-libs/cart/main/core/services/product-import-info.service.ts
```diff
@@ -1,43 +1,27 @@
...
+import { CartActions } from '../store/actions';
 
-/**
- * An abstract class for contexts based on cart (active cart, saved cart) which share part of logic - mainly handle cart actions and map results.
- */
-export abstract class CartOrderEntriesContext {
+@Injectable({
+  providedIn: 'root',
+})
+export class ProductImportInfoService {
```
plabadie, line 15: Does't removing CartOrderEntriesContext prevent the `AddEntries` method from beiing reused?

WeizhengSap: The shared `addEntries` is very small.

plabadie (approved):

PR #14587, opened 2021-12-10
Building maintenance branch only

hackergil:
Excluding release branches from Travis since they all fail and we don't backport e2e fixes

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,8 +8,8 @@ git:
 branches:
   only:
     - develop
+    - release/4.2.x
     - "/^epic\\/.*$/"
-    - "/^release\\/.*$/"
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -8,8 +8,8 @@ git:
 branches:
   only:
     - develop
+    - release/4.2.x
     - "/^epic\\/.*$/"
-    - "/^release\\/.*$/"
```
plabadie, line 12: Why would we only want to run the CI on 4.2.x and not any other release branches?

hackergil: it's on the PR description :)

hackergil:
After discussing with @plabadie let's not do this

PR #14600, opened 2021-12-13
chore: use adapter/connector model for REST APIs

sam-garland:


integration-libs/epd-visualization/root/connectors/scene/storage.connector.ts
```diff
@@ -0,0 +1,37 @@
+import { Injectable } from '@angular/core';
+import { Observable } from 'rxjs';
+import { NodesResponse } from './nodes-response';
+import { SceneAdapter } from './scene.adapter';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class StorageConnector {
```
Platonn: I cannot find this connector being used anywhere. But it should be used in services (instead of adapter). And only the connectors should touch adapters.

In other words, the dependencies between layers should look like here:
Component -> **Service -> Connector -> Adapter** -> http...

For more, see examples in other libs or see https://sap.github.io/spartacus-docs/connecting-to-other-systems/

I understand in most simple cases the connector layer is very simple.

Issue #14588, opened 2021-12-10
Cart is not anonymous error, for returning users with invalid OAuth token


neilhubertprice:
In a customised Spartacus UI, where the mini-cart is not shown prior to login: during login the merge cart functionality can incorrectly treat a logged-in cart for the user as if it is an anonymous cart.

This happens because the localStorage still contains the user's cartID from the last logged-in usage, which is then picked up during login but the code assumes that any cartID during login is for an anonymous cart.  The logic triggers a CreateCart API call with oldCartId set to the cartID from localStorage (this param expects an anonymous cart), and the toMergeCartGuid to the GUID from the same cart (most recent logged-in cart).  On the SAP Commerce side the createCart method in the CartsController attempts to validate that the oldCartID is for an anonymous cart & throws an error.

Returning users that still have a valid OAuth token do not go through the login process & so the issue cannot occur.  Users that are new to the website do not have a cartId in local storage, and so also cannot experience this issue.

Spartacus version 3.2

Steps to reproduce the behaviour:
1. Spartacus implementation with early login + customised so that no mini-cart displays until logged in
2. Login normally on first visit
3. Close browser
4. Wait until token is invalid, or delete token from localStorage
5. Return to website, which should then redirect to login
6. Login normally as per first visit
7. Spartacus issues a CreateCart request as described above, which returns a 400 error and the text 'Cart is not anonymous' 
8. 'Cart is not anonymous' error shown in red to the user after login

I believe the issue is within the logic of ActiveCartService.  That class creates a subscription to watch for User ID changes, such as login/logout.  The final segment of that subscription is:

_.subscribe(([previousUserId, userId, cartId]) => {
          // Only change of user and not a logout (current user id !== anonymous) should trigger loading mechanism
          if (this.isJustLoggedIn(userId, previousUserId)) {
            this.loadOrMerge(cartId, userId, previousUserId);
          }
        })_

In the loadOrMerge method that this calls, the execution ends up in this segment:

     _// We have particular cart locally, but we logged in, so we need to combine this with current cart or make it ours.
      this.multiCartService.mergeToCurrentCart({
        userId,
        cartId,
        extraData: {
          active: true,
        },
      });_

This piece of code implicitly assumes that the cartId given is an anonymous cart, but that is not true in the scenario described above.

Note: This issue is most likely not reproducible in the standard Spartacus UI as the mini-cart is always shown & so is initialising the cart handling before login is attempted.

**Expected behavior**
[...]

PR #14607, opened 2021-12-13
 Don't do the cart merge if session cart is the same as old cart

WeizhengSap:
#14588

### DIFF
--- a/feature-libs/cart/main/core/store/effects/cart.effect.spec.ts
+++ b/feature-libs/cart/main/core/store/effects/cart.effect.spec.ts
@@ -293,6 +293,21 @@ describe('Cart effect', () => {
 
       expect(cartEffects.mergeCart$).toBeObservable(expected);
     });
+
+    it('should do nothing if merged old cart is the same as session cart', () => {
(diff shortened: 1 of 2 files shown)


feature-libs/cart/main/core/store/effects/cart.effect.spec.ts
```diff
@@ -293,6 +293,21 @@ describe('Cart effect', () => {
...
+        cartId: 'xxx',
+        tempCartId: 'temp-uuid',
+      });
+
+      actions$ = hot('-a', { a: action });
+
+      cartEffects.mergeCart$.subscribe((emitted) => fail(emitted));
+      // just to get rid of the SPEC_HAS_NO_EXPECTATIONS message.
+      // The actual test is done in the subscribe part
+      expect(true).toBeTruthy();
```
plabadie: I think you can replace these 4 lines with:
```
expect(cartEffects.mergeCart$).toBeObservable(cold('--'));
```

plabadie (approved):

PR #14608, opened 2021-12-13
Don't do the cart merge if session cart is the same as old cart

### LINKED ISSUE #14588: Cart is not anonymous error, for returning users with invalid OAuth token

WeizhengSap:
close #14588

### DIFF
--- a/projects/core/src/cart/store/effects/cart.effect.spec.ts
+++ b/projects/core/src/cart/store/effects/cart.effect.spec.ts
@@ -1,7 +1,9 @@
+import { HttpErrorResponse } from '@angular/common/http';
 import { HttpClientTestingModule } from '@angular/common/http/testing';
 import { TestBed } from '@angular/core/testing';
 import { provideMockActions } from '@ngrx/effects/testing';
 import { StoreModule } from '@ngrx/store';
+import { normalizeHttpError } from '@spartacus/core';
 import { cold, hot } from 'jasmine-marbles';
 import { Observable, of, throwError } from 'rxjs';
 import { CLIENT_AUTH_FEATURE } from '../../../auth/client-auth/store/client-auth-state';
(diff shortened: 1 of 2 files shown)


plabadie:
I believe we should only backport fixes to the last major version.  If so, we should close this PR and keep the one for 3.4

plabadie (changes requested):
I believe we should only backport fixes to the last major version. If so, we should close this PR and keep the one for 3.4

PR #14624, opened 2021-12-15
feat: Add smoothing in payment type step in case of non-resolved requests

bgambocjaviniar:


feature-libs/checkout/b2b/components/checkout-payment-type/checkout-payment-type.component.spec.ts
```diff
@@ -109,52 +97,49 @@ describe('CheckoutPaymentTypeComponent', () => {
...
   });
 
-  it('should get all supported payment types', () => {
-    component.paymentTypes$
-      .subscribe((types) => {
-        expect(types).toBe(mockPaymentTypes);
-      })
-      .unsubscribe();
+  it('should get the loading state when selecting a payment type', (done) => {
+    component.selectedPaymentTypeIsLoading$
```
znikola: maybe we can add a test where we would check how the component html behaves when the state is loading? i.e. check for the spinner

znikola (changes requested):

feature-libs/checkout/b2b/components/checkout-payment-type/checkout-payment-type.component.spec.ts
```diff
@@ -190,4 +167,52 @@ describe('CheckoutPaymentTypeComponent', () => {
...
+
+      fixture.detectChanges();
+
+      expect(el.query(By.css('div.cx-spinner'))).toBeTruthy();
+    });
+
+    it('should NOT display spinner when user selects a new payment and response did not complete', () => {
+      (
+        component.changeSelectedPaymentTypeInProgress$ as BehaviorSubject<boolean>
+      ).next(false);
```
znikola, line 185: If the `changeSelectedPaymentTypeInProgress$` is false, then the user didn't select a payment type, did they?
If so, the title of the test is a bit confusing. Maybe we can just say: `should NOT display spinner when the payment type is NOT loading`?

feature-libs/checkout/b2b/components/checkout-payment-type/checkout-payment-type.component.spec.ts
```diff
@@ -190,4 +167,52 @@ describe('CheckoutPaymentTypeComponent', () => {
...
+
+      fixture.detectChanges();
+
+      expect(
+        el.query(By.css('.cx-checkout-btns .btn-primary')).nativeElement
+          .disabled
+      ).toBe(true);
+    });
+
+    it('should enable continue button when type selected is in progress', () => {
```
znikola:
```suggestion
    it('should enable continue button when the payment is selected', () => {
```

znikola (changes requested):

PR #14626, opened 2021-12-15
chore: Move models out of core in checkout

bgambocjaviniar:


feature-libs/cart/main/root/models/cart.model.ts
```diff
@@ -1,8 +1,8 @@
+import { PaymentDetails } from '@spartacus/checkout/base/root';
```
znikola: I'm wondering if this is a good idea - _checkout_ already depends on _cart_, but now we're making _cart_ depend on _checkout_.
Maybe this is a good sign that `PaymentDetails` actually belongs to _user_ lib?

feature-libs/cart/main/root/models/cart.model.ts
```diff
@@ -1,8 +1,8 @@
...
 import {
   Address,
```
znikola, line 2: A similar question is for the _order_ lib.

feature-libs/cart/main/root/models/cart.model.ts
```diff
@@ -1,8 +1,8 @@
...
   CostCenter,
   Currency,
```
znikola, line 4: Not sure how this happened, but once we run `yarn generate:deps` and `yarn config:update` we get the checkout ts configs inside of the checkout tsconfig itself 🤔

znikola (changes requested):

feature-libs/cart/main/root/models/cart.model.ts
bgambocjaviniar: moving to cart

PR #14642, opened 2021-12-16
Trigger loading order lib when place order

### LINKED ISSUE #14640: add the dependencies between checkout and order

WeizhengSap:
close #14640

### DIFF
--- a/feature-libs/checkout/components/components/place-order/place-order.component.ts
+++ b/feature-libs/checkout/components/components/place-order/place-order.component.ts
@@ -13,7 +13,8 @@ import {
   recurrencePeriod,
   ScheduleReplenishmentForm,
 } from '@spartacus/checkout/root';
-import { RoutingService } from '@spartacus/core';
+import { FeatureModulesService, RoutingService } from '@spartacus/core';
+import { ORDER_FEATURE } from '@spartacus/order/root';
 import { LaunchDialogService, LAUNCH_CALLER } from '@spartacus/storefront';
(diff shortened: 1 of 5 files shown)


feature-libs/checkout/components/components/place-order/place-order.component.ts
```diff
@@ -46,10 +47,15 @@ export class PlaceOrderComponent implements OnInit, OnDestroy {
...
     protected checkoutReplenishmentFormService: CheckoutReplenishmentFormService,
     protected launchDialogService: LaunchDialogService,
-    protected vcr: ViewContainerRef
+    protected vcr: ViewContainerRef,
+    protected featureModules: FeatureModulesService
   ) {}
 
   submitForm(): void {
+    if (this.featureModules.isConfigured(ORDER_FEATURE)) {
+      this.featureModules.resolveFeature(ORDER_FEATURE).subscribe();
```
Platonn, line 56: To avoid a race condition, we should wait unitl this observable emits, and only then call `this.checkoutService.placeOrder()`

Platonn: It would be also good to add a comment explaining why we this workaround, as it's not obvious: 
```
// TODO(#14649) `OccCheckoutAdapter.placeOrder()` needs `ORDER_NORMALIZER` to be provided, but it's provided only in the lazy-loaded ORDER_FEATURE
```

PR #14643, opened 2021-12-16
GH-14416: Optimize cart lib lazy loading - PDP

### LINKED ISSUE #14416: Lazy Loading Optimization - PDP
Evaluate if we could delay the loading of the cart lib on the PDP.

Likely, the wish list and the add to cart button trigger the load of the cart lib on the PDP.

Evaluate how we could carve out the add to cart button component so it does not eagerly load the cart library.

Some things to consider:
- We could leverage the new facades to load the bulk of the cart logic only when the add to cart button its clicked.
- We could implement loose coupling between the add to cart button and the add to cart modal. Such loose coupling its being implemented using events in the "cart toast" feature that can replace the add to cart modal.  Both options could use the same event.

Evaluate if we could change the wish list component to delay the load of the cart lib

plabadie:


projects/core/src/lazy-loading/facade-factory/facade-factory.service.ts
```diff
@@ -31,7 +39,8 @@ const PROXY_FACADE_INSTANCE_PROP = 'proxyFacadeInstance';
 export class FacadeFactoryService {
   constructor(
     protected featureModules: FeatureModulesService,
-    protected injector: Injector
+    protected injector: Injector,
```
WeizhengSap: It seems that  `injector` created in `protected injector: Injector,` is not used.

projects/core/src/lazy-loading/facade-factory/facade-factory.service.ts
```diff
@@ -131,7 +140,26 @@ export class FacadeFactoryService {
    * Facade proxy instances contain an object key to label them as such.
    * @param facade The facade object to evaluate
    */
-  isProxyFacadeInstance(facade: any) {
+  protected isProxyFacadeInstance(facade: any) {
```
WeizhengSap: maybe `private`?

feature-libs/cart/wish-list/root/components/index.ts
```diff
@@ -0,0 +1 @@
+export * from './add-to-wishlist/index';
```
WeizhengSap, line 1: this one is not exported in root's `index.ts`

feature-libs/cart/wish-list/root/components/add-to-wishlist/add-to-wish-list.component.ts
```diff
@@ -59,4 +61,42 @@ export class AddToWishListComponent {
       product.stock && product.stock.stockLevelStatus !== 'outOfStock'
     );
   }
+
+  protected getWishListLoading(): Observable<boolean> {
```
WeizhengSap, line 65: no need add these functions, just change the component's variable `wishListEntries$` and `loading$` to `get` function. something like this:
```
get wishListEntries$(): Observable<OrderEntry[]> {
    return this.wishListFacade.getWishList().pipe(
      filter((wishlist) => Boolean(wishlist)),
      map((wishList) => wishList.entries ?? [])
    );
  }
```
```
get loading$(): Observable<boolean> {
    return this.wishListFacade.getWishListLoading();
  }
```

feature-libs/cart/wish-list/root/components/add-to-wishlist/add-to-wish-list.component.ts
```diff
@@ -17,15 +24,10 @@ export class AddToWishListComponent {
...
-    .getWishList()
-    .pipe(
-      filter((wishlist) => Boolean(wishlist)),
-      map((wishList) => wishList.entries ?? [])
-    );
+  wishListEntries$: Observable<OrderEntry[]> = this.getWishListEntries();
 
   userLoggedIn$: Observable<boolean> = this.authService.isUserLoggedIn();
-  loading$: Observable<boolean> = this.wishListFacade.getWishListLoading();
+  loading$: Observable<boolean> = this.getWishListLoading();
```
WeizhengSap: how about this:
```
wishListEntries$: Observable<OrderEntry[]>;
loading$: Observable<boolean>;

  userLoggedIn$: Observable<boolean> = this.authService.isUserLoggedIn().pipe(
    tap((isLogin) => {
      if (isLogin) {
        this.wishListEntries$ = this.wishListFacade.getWishList().pipe(
          filter((wishlist) => Boolean(wishlist)),
          map((wishList) => wishList.entries ?? [])
        );
        this.loading$ = this.wishListFacade.getWishListLoading();
      }
    })
  );
```
In html tmeplate, when `wishListEntries$` or `loading$` is accessed, the user must be logged in.

WeizhengSap: or this:
```
  private _wishListEntries$: Observable<OrderEntry[]>;
  private _loading$: Observable<boolean>;

  get wishListEntries$(): Observable<OrderEntry[]> {
    if (this._wishListEntries$ === undefined) {
      this._wishListEntries$ = this.wishListFacade.getWishList().pipe(
        filter((wishlist) => Boolean(wishlist)),
        map((wishList) => wishList.entries ?? [])
      );
    }
    return this._wishListEntries$;
  }

  get loading$(): Observable<boolean> {
    if (this._loading$ === undefined) {
      this._loading$ = this.wishListFacade.getWishListLoading();
    }
    return this._loading$;
  }
```

feature-libs/cart/karma.conf.js
```diff
@@ -26,7 +26,7 @@ module.exports = function (config) {
         global: {
           statements: 90,
           lines: 90,
-          branches: 75,
+          branches: 74,
```
WeizhengSap: can we add more unit tests instead of lower the threshold?

feature-libs/cart/main/root/components/add-to-cart/add-to-cart.module.ts
```diff
@@ -0,0 +1,52 @@
...
+  KeyboardFocusModule,
+  ModalModule,
+  PromotionsModule,
+  SpinnerModule,
+} from '@spartacus/storefront';
+import { AddToCartComponent } from './add-to-cart.component';
+
+@NgModule({
+  imports: [
+    CommonModule,
```
WeizhengSap: except `CommonModule`, `ReactiveFormsModule`, `I18nModule` and `ItemCounterModule`, all other modules are not used, right?

feature-libs/cart/main/root/events/cart.events.ts
```diff
@@ -76,3 +76,12 @@ export class CartUpdateEntryFailEvent extends CartEvent {
   quantity: number;
   entry: OrderEntry;
 }
+
+export class CartUiEventAddToCart extends CxEvent {
```
WeizhengSap, line 80: We can use `CartAddEntryEvent`, right?

feature-libs/cart/main/root/components/add-to-cart/add-to-cart.component.ts
```diff
@@ -149,38 +132,22 @@ export class AddToCartComponent implements OnInit, OnDestroy {
...
-    this.activeCartService
-      .getEntries()
-      .pipe(take(1))
-      .subscribe((entries) => {
-        this.numberOfEntriesBeforeAdd = entries.length;
-        this.openModal();
-        this.activeCartService.addEntry(this.productCode, quantity);
+    this.activeCartService.addEntry(this.productCode, quantity);
+
+    // Because the cart library can be lazy loaded, we wait for
```
WeizhengSap, line 137: if we use `CartAddEntryEvent`, we don't need this block of codes.

feature-libs/cart/main/root/cart-root.module.ts
```diff
@@ -48,7 +48,7 @@ export function defaultCartComponentsConfig() {
       },
     ]),
   ],
-  exports: [MiniCartModule],
+  exports: [MiniCartModule, AddToCartModule],
```
WeizhengSap: `imports` or `exports`?

feature-libs/cart/main/components/events/add-to-cart-dialog-event.listener.ts
```diff
@@ -0,0 +1,61 @@
...
+      this.eventService.get(CartUiEventAddToCart).subscribe((event) => {
+        this.openModal(event);
+      })
+    );
+  }
+  protected openModal(event: CartUiEventAddToCart): void {
+    let modalRef: ModalRef;
+
+    let modalInstance: any;
+    modalRef = this.modalService.open(AddedToCartDialogComponent, {
```
WeizhengSap: `const modalRef = ...`, and remove `let modalRef: ModalRef;`

feature-libs/cart/main/components/events/add-to-cart-dialog-event.listener.ts
```diff
@@ -0,0 +1,61 @@
...
+      centered: true,
+      size: 'lg',
+    });
+    modalInstance = modalRef.componentInstance;
```
WeizhengSap: `const modalInstance = ...`, and remove `let modalInstance: any;`

feature-libs/cart/main/components/cart/add-to-cart/add-to-cart.module.ts
```diff
@@ -37,19 +30,8 @@ import { AddedToCartDialogComponent } from './added-to-cart-dialog/added-to-cart
     KeyboardFocusModule,
     ModalModule,
   ],
```
WeizhengSap: can we remove the unused imports?

feature-libs/cart/main/components/events/add-to-cart-dialog-event.listener.ts
```diff
@@ -0,0 +1,61 @@
...
+    // Display last entry for new product code. This always corresponds to
+    // our new item, independently of whether merging occured or not
+    modalInstance.entry$ = this.activeCartFacade.getLastEntry(
+      event.productCode
+    );
+    modalInstance.cart$ = this.activeCartFacade.getActive();
+    modalInstance.loaded$ = this.activeCartFacade.isStable();
+    modalInstance.quantity = event.quantity;
+    modalInstance.addedEntryWasMerged$ = this.activeCartFacade
```
WeizhengSap: There is no `addedEntryWasMerged$`, but instead `addedEntryWasMerged`.

feature-libs/cart/main/components/cart/add-to-cart/added-to-cart-dialog/added-to-cart-dialog.component.ts
```diff
@@ -83,11 +83,12 @@ export class AddedToCartDialogComponent implements OnInit {
...
   ngOnInit() {
-    this.addedEntryWasMerged$ = this.loaded$.pipe(
-      filter((loaded) => loaded),
-      switchMapTo(this.cartService.getEntries()),
-      map((entries) => entries.length === this.numberOfEntriesBeforeAdd)
-    );
+    this.addedEntryWasMerged =
+      getLastValueSync(
+        this.entry$.pipe(
+          map((cartEntry) => (cartEntry?.quantity ?? 0) > this.quantity)
```
WeizhengSap: if `cartEntry.quantity` is 1, and adding 1 same product, the value will be `false`. But, it should be `true`.
Screen Shot 2022-01-06 at 3 17 53 PM

Before, we checked `numberOfEntriesBeforeAdd`. If there are 8 entries, after adding new products, there are still 8 entries, that means we don't add new products, but instead updating the existing quantity.

feature-libs/cart/main/components/events/cart-components-event.module.ts
```diff
@@ -0,0 +1,7 @@
+import { NgModule } from '@angular/core';
+import { AddToCartDialogEventListener } from './add-to-cart-dialog-event.listener';
+
+@NgModule({})
+export class CartComponentsEventModule {
+  constructor(_addToCartDialogEventListener: AddToCartDialogEventListener) {}
```
WeizhengSap: can we move `add-to-cart-dialog-event.listener.ts` to `add-to-cart` folder, and remove this module? We can move this `constructor` to `AddToCartModule`.

feature-libs/cart/main/core/facade/active-cart.service.ts
```diff
@@ -287,20 +287,18 @@ export class ActiveCartService implements ActiveCartFacade, OnDestroy {
...
-        ? this.cartEntity$.pipe(
-            filter(() => !Boolean(getLastValueSync(this.isGuestCart())))
-          )
-        : this.cartEntity$
-    ).pipe(filter((cartState) => !cartState.loading));
+    const cartSelector$ = forGuestMerge
+      ? this.cartEntity$.pipe(
+          filter(() => !Boolean(getLastValueSync(this.isGuestCart())))
+        )
+      : this.cartEntity$;
```
WeizhengSap: I think we need this filter `filter((cartState) => !cartState.loading)`. When cart is loading, new cart should not be created. Without the filter, the state is this (create happens right after load):
Screen Shot 2022-01-06 at 4 37 51 PM

With the filter, the state is this (create happens when load fails):
Screen Shot 2022-01-06 at 4 38 58 PM

feature-libs/cart/main/root/components/mini-cart/mini-cart-component.service.ts
```diff
@@ -84,11 +82,11 @@ export class MiniCartComponentService {
    * Once the observable returned by activeCartRequired emits true, it completes.
    * activeCartRequired helps to make the mini cart compatible with some level of lazy loading.
    */
-  protected activeCartRequired() {
+  protected activeCartRequired(): Observable<boolean> {
```
WeizhengSap, line 85: Each time this function is called, there is a new stream created (which make `isFacadeImplProvided` called multiple times). Can we just create one stream for it and maybe use `shareReplay`? something like this:
`protected activeCartRequired$: Observable<boolean> = combineLatest...`?

feature-libs/cart/main/root/components/add-to-cart/add-to-cart.component.ts
```diff
@@ -149,38 +132,22 @@ export class AddToCartComponent implements OnInit, OnDestroy {
...
       return;
     }
-    this.activeCartService
-      .getEntries()
-      .pipe(take(1))
-      .subscribe((entries) => {
-        this.numberOfEntriesBeforeAdd = entries.length;
-        this.openModal();
-        this.activeCartService.addEntry(this.productCode, quantity);
+    this.activeCartService.addEntry(this.productCode, quantity);
```
WeizhengSap: For this component, can we also make cart-lib loaded for login user? Since, we have `mini-cart` and `wish-list` (both make cart-lib loaded for login user), so in PDP or PLP, cart-lib loaded for login user. But, without `mini-cart` or `wish-list`, there will be an issue. That is, when there is a `cartId` stored in `localStorage`, and the cart is removed from backend, the "add-to-cart" will not work well. When the `activeCartService` construct, it will load the cart which `id` is stored in browser or `current`.  If this loading fails, the `cartId` stored in storage will be removed (this is what we expected). But, if "add-to-cart" makes the loading happen (loading `current`, but the cartId is something stored in browser), the state will be wrong.

feature-libs/cart/main/core/facade/active-cart.service.ts
WeizhengSap: I made some changes in this PR: https://github.com/SAP/spartacus/pull/14747

feature-libs/cart/wish-list/root/components/index.ts
plabadie: Good catch.  I added the export.

projects/core/src/lazy-loading/facade-factory/facade-factory.service.ts
plabadie: Ok.

projects/core/src/lazy-loading/facade-factory/facade-factory.service.ts
plabadie: Indeed.  I removed it to clean up the code.

feature-libs/cart/main/root/cart-root.module.ts
plabadie: Import. This should have been `imports` for `MiniCartModule` to begin with. (my bad)

feature-libs/cart/main/root/components/add-to-cart/add-to-cart.module.ts
plabadie: Indeed, a cleanup was required here.  AddToCartComponent was previously declared in the same module file as other components and I had carried over all the imports.

feature-libs/cart/main/root/components/add-to-cart/add-to-cart.component.ts
plabadie: The use case is different than mini-cart or wish list, I'm not sure the same logic applies.  Let's discuss this one.

feature-libs/cart/main/components/cart/add-to-cart/added-to-cart-dialog/added-to-cart-dialog.component.ts
plabadie: I can repro.  This needs fixing.

feature-libs/cart/main/core/facade/active-cart.service.ts
plabadie: Can we take the fix from #14747 instead of this fix attempt?

WeizhengSap: yes, we can take the fix from #14747

feature-libs/cart/main/components/cart/add-to-cart/add-to-cart.module.ts
plabadie: I removed some imports that didn't seem used and I refactored a bit to reflect thee fact that AddedToCart-dialogComponent is now alone in his module.

feature-libs/cart/main/components/events/cart-components-event.module.ts
plabadie: Done. Suggestion applied.

feature-libs/cart/main/components/events/add-to-cart-dialog-event.listener.ts
plabadie: fixed

feature-libs/cart/main/components/events/add-to-cart-dialog-event.listener.ts
plabadie: fixed

feature-libs/cart/main/components/events/add-to-cart-dialog-event.listener.ts
plabadie: Thanks.  This is dead code that I should have deleted.

feature-libs/cart/main/components/cart/add-to-cart/added-to-cart-dialog/added-to-cart-dialog.component.ts
plabadie: This is fixed.

feature-libs/cart/karma.conf.js
plabadie: Got the branch coverage above the 75% mark again.

feature-libs/cart/wish-list/root/components/add-to-wishlist/add-to-wish-list.component.ts
plabadie: I understand the suggestion, but I'd like too keep the code as it is.  It might be subjective, but I find the logic more readable, modular and testable using functions.  You can look at the associated unit tests to see the kind of coverage this design allowed to achieve.

feature-libs/cart/main/root/components/mini-cart/mini-cart-component.service.ts
plabadie: Good suggestion. I lean however to still leave it as-is for now, but lets keep it in mind for a future improvement.  There is no resource intensive processing being done by the stream.  Moreover, the page header is not recreated each time a user navigates to a new page in the storefront.  Since the mini cart is typically in the the header, this lowers the amount of calls made.

feature-libs/cart/main/components/cart/index.ts
```diff
@@ -1,6 +1,5 @@
-export * from './add-to-cart/add-to-cart.component';
-export * from './add-to-cart/add-to-cart.module';
-export * from './add-to-cart/added-to-cart-dialog/added-to-cart-dialog.component';
+export * from './added-to-cart-dialog/added-to-cart-dialog.component';
+export * from './added-to-cart-dialog/added-to-cart-dialog.module';
```
WeizhengSap: just this `export * from './added-to-cart-dialog/index';`.

feature-libs/cart/main/components/cart/added-to-cart-dialog/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from './added-to-cart-dialog.component';
+export * from './added-to-cart-dialog.module';
```
WeizhengSap, line 3: do we also want to export event listener?

feature-libs/cart/main/components/cart/added-to-cart-dialog/added-to-cart-dialog-event.listener.ts
```diff
@@ -0,0 +1,52 @@
...
+    this.onAddToCart();
+  }
+
+  protected onAddToCart() {
+    this.subscription.add(
+      this.eventService.get(CartUiEventAddToCart).subscribe((event) => {
+        this.openModal(event);
+      })
+    );
+  }
```
WeizhengSap, line 27: please add an empty line here :))

feature-libs/cart/main/root/components/add-to-cart/add-to-cart.component.ts
```diff
@@ -149,36 +132,44 @@ export class AddToCartComponent implements OnInit, OnDestroy {
     if (!this.productCode || quantity <= 0) {
       return;
     }
+
+    let numberOfEntriesBeforeAdd = 0;
```
WeizhengSap: we don't need this variable, right? just pass `entries.length` to `createCartUiEventAddToCart`. Or ` const numberOfEntriesBeforeAdd = entries.length;` to make it more readable.

feature-libs/cart/main/components/cart/added-to-cart-dialog/added-to-cart-dialog.module.ts
```diff
@@ -0,0 +1,37 @@
...
+    RouterModule,
+    SpinnerModule,
+    PromotionsModule,
+    UrlModule,
+    IconModule,
+    I18nModule,
+    ItemCounterModule,
+    ModalModule,
+  ],
+  providers: [],
```
WeizhengSap: remove this empty providers?

feature-libs/cart/wish-list/root/components/add-to-wishlist/add-to-wish-list.component.ts
WeizhengSap: The thing is, when `getWishListEntries()` is called, we know that the user is login. So, `return of([]);` will never be executed (same as `return of(false);`). We  subscribe `this.authService.isUserLoggedIn()` in 2 places. When I read the codes, my first feeling is function `wishListDataRequired` is redundant.

feature-libs/cart/main/components/cart/added-to-cart-dialog/added-to-cart-dialog.module.ts
plabadie: Fixed

feature-libs/cart/main/components/cart/added-to-cart-dialog/index.ts
plabadie: I have no strong opinion on that topic. Since we export the rest of the UI code, I didn't see a reason not to export the listener.  
The checkout lib exports it's listener ( although it is in `core`, not `components`  )

Do you think it should be private instead?

WeizhengSap: I think we can export the listener. `onAddToCart` and `openModal` both are `protected`.

plabadie: Yes.

feature-libs/cart/main/components/cart/index.ts
plabadie: Yes, the index is better.

feature-libs/cart/main/components/cart/added-to-cart-dialog/index.ts
plabadie: Exported

feature-libs/cart/main/components/cart/added-to-cart-dialog/added-to-cart-dialog-event.listener.ts
plabadie: Yup, that looks better.

feature-libs/cart/main/root/components/add-to-cart/add-to-cart.component.ts
plabadie: There was a class property before and I made it a local var instead, but perhaps we don't event need that variable.  I'll change it

feature-libs/cart/wish-list/root/components/add-to-wishlist/add-to-wish-list.component.ts
plabadie: I updated it to follow your suggestion.

feature-libs/cart/wish-list/root/components/add-to-wishlist/add-to-wish-list.component.ts
```diff
@@ -59,4 +61,11 @@ export class AddToWishListComponent {
       product.stock && product.stock.stockLevelStatus !== 'outOfStock'
     );
   }
+
+  protected getWishListEntriesFromFacade(): Observable<OrderEntry[]> {
```
WeizhengSap: can we change the function name?

feature-libs/cart/main/components/cart/added-to-cart-dialog/added-to-cart-dialog.component.ts
```diff
@@ -82,24 +81,39 @@ export class AddedToCartDialogComponent implements OnInit {
...
+      switchMapTo(this.activeCartFacade.getEntries()),
+      map((entries) => entries.length === numberOfEntriesBeforeAdd)
+    );
+  }
   /**
    * Adds quantity and entryNumber form controls to the FormGroup.
    * Returns quantity form control.
    */
-  protected getQuantityFormControl(entry: OrderEntry): FormControl {
+  protected getQuantityFormControl(entry: OrderEntry | undefined): FormControl {
```
WeizhengSap: can be this `protected getQuantityFormControl(entry?: OrderEntry): FormControl`

feature-libs/cart/wish-list/root/components/add-to-wishlist/add-to-wish-list.component.ts
plabadie: Yes.  Changed to `getWishListEntries`

feature-libs/cart/main/components/cart/added-to-cart-dialog/added-to-cart-dialog.component.ts
plabadie: I'll look for alternatives

WeizhengSap: `(entry?: OrderEntry)` is the same as `(entry: OrderEntry | undefined)`, right? What alternative you want? I am also fine with this `protected getQuantityFormControl(entry: OrderEntry | undefined)`. :)

WeizhengSap (approved):

feature-libs/cart/main/components/cart/added-to-cart-dialog/added-to-cart-dialog.component.ts
plabadie: I'm sorry, my bad.  I misread your first comment.  And I meant that I was going to try to find alternatives to what I did.

I applied your suggestion : )

WeizhengSap (approved):

PR #14465, opened 2021-11-29
Add global messages for adding new payment and shipment

RadhepS:
closes: https://github.com/SAP/spartacus/issues/13699

PR #14401, opened 2021-11-23
feat: disable shipping method radio input when there is ongoing delivery mode request

wangzixi-diablo:
fixes: #14386

PR #14689, opened 2021-12-22
chore: Carry over changes from develop to checkout lib

bgambocjaviniar:
changes from december 7th and onwards 

changes to checkout

- https://github.com/SAP/spartacus/pull/14465 (carried over which includes add new payment and shipping global message) --> updated in new checkout to use listeners upon address or shipping successfully created
- https://github.com/SAP/spartacus/pull/14401 (did not carry over changes as those changes were 'fixes' for JALO)


checkout-git-check.sh
```diff
@@ -22,7 +22,7 @@ git pull
 # =========================================================
 # since_date: Update the date accordingly. How far back to check for changes
 # =========================================================
-since_date=2021-12-06
+since_date=2021-12-20
```
znikola, line 25: Nice that you used "yesterday's" date ;)

projects/storefrontapp-e2e-cypress/cypress/helpers/b2b/b2b-checkout.ts
```diff
@@ -93,8 +93,8 @@ export function selectAccountPayment() {
     'getShippingPage'
   );
   cy.get('button.btn-primary').click({ force: true });
-
   cy.wait(`@${shippingPage}`).its('response.statusCode').should('eq', 200);
+  cy.wait('@getCart').its('response.statusCode').should('eq', 200);
```
znikola, line 97: Was this removed accidentally, and you re-added it?

git-log-develop-from-dec7.txt
```diff
@@ -0,0 +1,414 @@
+BRIAN - DONE commit ccdb51175eec7c775425a288e5b62b355480b614
```
znikola: I think we can safely remove this file

git-log-develop-from-dec7.txt
```diff
@@ -0,0 +1,414 @@
...
+Author: Jerry Wang <58975336+wangzixi-diablo@users.noreply.github.com>
```
znikola: For Jerry's PR: 
What do you think, should we carry over the spinner?
The reason why I'm asking is what if Jalo fails after the retries? It's not likely to happen, but still technically possible if back-end is overloaded, or something.

znikola (changes requested):

git-log-develop-from-dec7.txt
bgambocjaviniar: left it for you if you wanted to double check haha

projects/storefrontapp-e2e-cypress/cypress/helpers/b2b/b2b-checkout.ts
bgambocjaviniar: yup

git-log-develop-from-dec7.txt
bgambocjaviniar: 1 - 🙈  develop delivery mode step 'styling' is broken due to his change
2 - was not a spinner. He disabled the inputs and continue button. Current fix that I just did utilizes the spinner upon selecting a delivery mode and disables the continue button as well. (Need to talk and normalize what way we choose because most of the time, we just use spinners instead of disabling the 'inputs')

znikola: Is it broken in the new checkout? If not, no need to "fix" develop, as the new checkout will be the default one (and possibly the only one).
Ideally, we want to use the progress button + disable the form while loading. But this is the task for the future, where we will refactor components.
Let's just use the spinner. Optionally, we can disable the form if it's easy enough to do so, or if it makes sense.

znikola (approved):

git-log-develop-from-dec7.txt
bgambocjaviniar: conclusion that we came up with:

- will discuss with UI/UX starting January on new checkout behavior in terms of loading 'spinners' or 'disabling + spinners'

PR #14690, opened 2021-12-22
feat: Convert guards to facades

znikola:
This PR converts the base and b2b guards into facade-service "pattern".

This is just a POC with the following cons:
- guards leak from `/components` entry point to `/root`
- it brings an additional complexity into guard implementations
- it deviates from the standard way of implementing guards

The pros are:
- it doesn't require `unifiedInjector` in order to trigger LL guards when they are needed
- it allows customers to extend our LL guards


znikola:
Closing as we don't want to go with this approach.

PR #14703, opened 2021-12-23
fix: Fix incorrect redirection when using code flow for auth

### LINKED ISSUE #14694: Code Auth Flow causes incorrect redirection when opening new tabs
**Describe the bug**
When configuring `OAuthLibConfig` to use `responseType: 'code'` which enables code auth flow (as opposed to the default implicit flow), logging in, and attempting to open a new tab via a link or changes the url in the browser's address bar, the user will be redirected to the previous url instead of the intended one.

Example:
- User logs in
- User is on cart page
- User opens new tab from category link
- Tab opens up also on cart page

4.2

**To Reproduce**
1. Setup 3rd party authentication server
2. Set `responseType: 'code'` in `AuthConfig`
3. Login
4. Navigate to cart page
5. Opens new tab from category link (or any link on the site)
6. Will be navigated back to cart page on new tab

**Expected behavior**
New tab should open on intended page.

 - OS: Windows 10
 - Browser Chrome

**Additional context**
This behaviour occurs because the `redirectUrl` stored in local storage (used to know where to send the user after logging in) is not updated due to the `NavigationEvent` that would update this being cancelled by a second `NavigationEvent`. This is done in `AuthRedirectService`.

This secondary `NavigationEvent` is thrown by `checkOAuthParamsInUrl()` in `AuthService` which is fired upon starting or returning to the Spartacus site. The `tryLogin()` method when using implicit flow will return `true` or `false` depending on if it succeeds and the proceeding condition will be met and cause the redirect. Because `tryLogin()` can fail and return false, the redirect works correctly for implicit flow.

However, for code flow, `tryLogin()` will always return `true` regardless of what happens within the method. Because of this, the redirect is always called even if we are not logging in anymore which causes the bug.

A proposed solution is to subscribe to `OAuthLibService` events and watch for the `'token_received'` event. If catching this event after calling `tryLogin()` using code flow, then the user must be logging in and we can call the redirect. Otherwise, don't call the redirect.

Zeyber:
Closes: #14694


znikola (approved):

znikola (approved):

hackergil (approved):

projects/core/src/auth/user-auth/facade/auth.service.ts
```diff
@@ -43,13 +44,26 @@ export class AuthService {
    */
   async checkOAuthParamsInUrl(): Promise<void> {
     try {
+      const tokenReceived = getLastValueSync(
+        this.oAuthLibWrapperService.events$.pipe(
+          filter((e) => e.type === 'token_received')
+        )
+      );
```
Platonn, line 51: Can you help me understand why we have used here `getLastValueSync`?
https://github.com/SAP/spartacus/blob/5df4e4c92ae3c24eb4df27b79b51b46f36fb732b/projects/core/src/util/rxjs/get-last-value-sync.ts#L3-L17
As far as I understand, the event `token_reveived` has a chance to occur only later, during the invocation of `tryLogin()`. But, because we attempt to pick the value from the observable synchronously before the `tryLogin() `is called, IMHO the const `tokenReceived` will be always falsy. Would you agree?

--
Instead, did you consider subscribing to the events$ observable (+ `pipe(filter(...), take(1))`) first, and checking if it emitted anything only later, after we have called `tryLogin()`?
PS. if we go that route, it would be good to clean up the subscription (unsubscribe) also when the event does not happen (i.e. when `take(1)` doesn't close the stream automatically).

PS2. Nitpick: It would be nice to also add some typing for `const tokenReceived`.

PR #14727, opened 2021-12-31
chore: Enhance the checkout e2e experience + fix replenishment bug in new checkout

bgambocjaviniar:
This PR fixes a bug where after scheduling an order we would see the regular order confirmation page (instead of the scheduled order confirmation page).

### DIFF
--- a/feature-libs/checkout/scheduled-replenishment/components/services/checkout-replenishment-form.service.ts
+++ b/feature-libs/checkout/scheduled-replenishment/components/services/checkout-replenishment-form.service.ts
@@ -9,7 +9,6 @@ import {
   DeliveryAddressSetEvent,
   DeliveryModeClearedEvent,
   DeliveryModeSetEvent,
-  OrderPlacedEvent,
   PaymentDetailsCreatedEvent,
   PaymentDetailsSetEvent,
 } from '@spartacus/checkout/base/root';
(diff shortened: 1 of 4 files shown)


feature-libs/checkout/scheduled-replenishment/components/services/checkout-replenishment-form.service.ts
```diff
@@ -64,8 +62,6 @@ export class CheckoutReplenishmentFormService implements OnDestroy {
         this.eventService.get(RestoreSavedCartSuccessEvent),
         this.eventService.get(PaymentDetailsCreatedEvent),
         this.eventService.get(PaymentDetailsSetEvent),
-        this.eventService.get(OrderPlacedEvent),
-        this.eventService.get(ReplenishmentOrderScheduledEvent),
```
znikola, line 68: I guess this was a mistake to include the repl event here, and then do `next` with `ORDER_TYPE.PLACE_ORDER`.
But why are we removing `OrderPlacedEvent`?

znikola (changes requested):

feature-libs/checkout/scheduled-replenishment/components/services/checkout-replenishment-form.service.ts
bgambocjaviniar: Nope

as mentioned, OrderPlacedEvent by default is using order_type.place_order, therefore no need to 'reset'

znikola (approved):

PR #14423, opened 2021-11-24
feat: epd-visualization integration library (#13933)

sam-garland:

PR #14753, opened 2022-01-11
feat: epd-visualization integration library

sam-garland:
Changed branch name as requested: `epic/epd-visualization-maintenance` -> `integration-libs/epd-visualization-maintenance`

Previous PR before branch change: https://github.com/SAP/spartacus/pull/14420

PR #14754, opened 2022-01-11
feat: epd-visualization integration library

sam-garland:
Changed branch name as requested: epic/epd-visualization-> integration-libs/epd-visualization

Previous PR before branch change: #14423


Platonn:
Hi @sam-garland 
Would you mind resolving the conflicts in this PR do develop, please?
I'm going do the last review comparing the changes: `this-PR --> develop` vs. `merged-PR --> release/4.3.x` branch.
Thanks in advance

sam-garland:
> Hi @sam-garland Would you mind resolving the conflicts in this PR do develop, please? I'm going do the last review comparing the changes: `this-PR --> develop` vs. `merged-PR --> release/4.3.x` branch. Thanks in advance

@Platonn I have changed the behaviour so that `cx-epd-visualization-compact-add-to-cart` buttons are not rendered when the `addToCart` feature is not configured (since these buttons will not do anything when clicked when the `addToCart` feature is not configured). 

I also notice that more manual steps are needed after the schematic runs.
After adding `provideConfig(defaultB2bOccConfig), provideConfig(defaultB2bCheckoutConfig)` to the `providers` section of `spartacus-configuration.module.ts` (and adding imports), I needed to add the `@spartacus/setup`, `@spartacus/checkout` and `@spartacus/order` packages to the `dependencies` section in the `package.json` and run `yarn install` before building.

Platonn:
@sam-garland 

> I have changed the behaviour so that cx-epd-visualization-compact-add-to-cart buttons are not rendered when the addToCart feature is not configured (since these buttons will not do anything when clicked when the addToCart feature is not configured).

Thank you for paing attention to the recent changes in develop with regards to the cart lib. That said, I'm not 100% sure the need to check if add-to-cart feature was installed in _this specific case._ Let me first consult with @plabadie and @WeizhengSap (authors of recent changes in the cart lib) and I'll get back to you.

Regarding the manual steps for installing b2b... understood, this is not comfortable situation for you. 

I would like to wait with deciding  whether we should add officialy an option `--b2b` to `@spartacus/schematics` installation schematics. It's becasue we plan in the future to recognize it automatically from the basesites configuration in the CMS.

Did you consider any other workarounds? For example:
a) writing your own script (kind of "internal schematics?") OR
b) change sample data: put your sample products into the electronics' product catalogue instead of the powertools' catalogue? then for testing purposes you could use the electronics' site OR
c) ...or maybe you have any other ideas for workaround?

sam-garland:
@Platonn 

> Thank you for paing attention to the recent changes in develop with regards to the cart lib. That said, I'm not 100% sure the need to check if add-to-cart feature was installed in _this specific case._ Let me first consult with @plabadie and @WeizhengSap (authors of recent changes in the cart lib) and I'll get back to you.

Sure.

> Regarding the manual steps for installing b2b... understood, this is not comfortable situation for you.
> 
> I would like to wait with deciding whether we should add officialy an option `--b2b` to `@spartacus/schematics` installation schematics. It's becasue we plan in the future to recognize it automatically from the basesites configuration in the CMS.
> 
> Did you consider any other workarounds? For example: a) writing your own script (kind of "internal schematics?") OR b) change sample data: put your sample products into the electronics' product catalogue instead of the powertools' catalogue? then for testing purposes you could use the electronics' site OR c) ...or maybe you have any other ideas for workaround?

Given that the long term plan is to detect the channel from the base site configuration, perhaps the `@spartacus/epd-visualization` schematic could:
- Read the `spartacus-configuration.module.ts` file to get the `baseSite` configuration
- Determine whether a B2B site is being used based on the name (defaulting to B2C)
- Add B2B configuration if required

Then the logic above could be replaced in the future if there is a more reliable way of determining whether the configured base site is a B2B site.

Platonn:
Hi @sam-garland 

### Suggestion to remove some unnecessary code
Since you just extend the TS class `AddToCartComponent` and do nothing more around the add-to-cart feature, I believe we can get rid of some unnecessary code:
- in your component:
  - unnecessary check whether feature was configured
  - unnecessary custom constructor (with FeatureModulesService dependency):
  https://github.com/SAP/spartacus/blob/bc49b9dcaf38db53914a67f0088825ccebbb1d30/integration-libs/epd-visualization/components/visual-picking/visual-picking-tab/product-list/compact-add-to-cart/compact-add-to-cart.component.ts#L28-L51
  - ... and the check for `requiredFeaturesEnabled` in the template
  https://github.com/SAP/spartacus/blob/bc49b9dcaf38db53914a67f0088825ccebbb1d30/integration-libs/epd-visualization/components/visual-picking/visual-picking-tab/product-list/compact-add-to-cart/compact-add-to-cart.component.html#L2
... and in your modules you don't need to import the original `AddToCartModule` anytime:
  - https://github.com/SAP/spartacus/blob/34c9a5a4d78f76de931e89ef7c305d3b93c2fcc5/integration-libs/epd-visualization/components/visual-picking/visual-picking-tab/product-list/compact-add-to-cart/compact-add-to-cart.module.ts#L29
  - https://github.com/SAP/spartacus/blob/34c9a5a4d78f76de931e89ef7c305d3b93c2fcc5/integration-libs/epd-visualization/components/visual-picking/visual-picking-tab/product-list/visual-picking-product-list.module.ts#L25
  - https://github.com/SAP/spartacus/blob/34c9a5a4d78f76de931e89ef7c305d3b93c2fcc5/integration-libs/epd-visualization/components/visual-picking/visual-picking-tab/product-list/visual-picking-product-list.module.ts#L29

### Explanation:
1. No need to import `AddToCartModule`
   If you referenced the original component via the selector `<cx-add-to-cart>` in the html of some of your components, you would need to import the original `AddToCartModule`. But it's not the case for you. You just extend a TS class. 

2. No need to check if a feature was configured
   You don't relay in any way on the fact that the `ADD_TO_CART_FEATURE` is configured for lazy loading (in other words, you don't need to check if `featureModulesService.isConfigured(ADD_TO_CART_FEATURE)`).

    More info, how lazy loading works in your case:
    You just import a TS class `import { xxx } from '@spartacus/different-lib-that-might-be-lazy-loaded-separately`). Then the `xxx` class (which in your case is `AddToCartComponent`) will be extracted by Webpack during the compilation into a separate shared JS chunk. Then in runtime the Webpack will trigger loading of the shared chunk, when either the chunk of your EPD's feature module is lazy-loaded or JS chunks of other features importing the `AddToCartComponent` class are lazy-loaded (including the default module of ADD_TO_CART_FEATURE). 

### Note on a possible, completely different solution (which I don't want to go now):
Instead of creating a brand new component `CompactAddToCartComponent` with custom HTML and inherited TS, you could consider reusing the original component and only change it's look-and-feel via pure CSS + pass the different text into it (the icon instead of the default "add to cart" text) via a new `@Input` or preferrably the `<ng-content>` slot.
This would require the following steps:
- add a support of passing a custom content into the original add-to-cart component (via a new `@input` or `<ng-content>` slot)
- reference `<cx-add-to-cart>` instead of `<cx-epd-visualization-compact-add-to-cart>` inside the html template of the parent component `<cx-epd-visualization-compact-add-to-cart>`
- pass into the `<cx-add-to-cart>` the custom content (icon)
- then you could also drop your custom component `cx-epd-visualization-compact-add-to-cart`
- import `AddToCartModule` in `VisualPickingProductListModule` to make the component visible `<cx-add-to-cart>` in the template of the parent component `<cx-epd-visualization-compact-add-to-cart>`

That said, I don't like to go this route in 5.0 due to 2 reasons:
1. in `release/4.3.x` you already merged a version that uses a custom Compact add to cart component. I don't want to introduce inconsistencies, at least in this PR.
2. importing the original `AddToCartModule` inside your  module would unnecessarily provide the cms mapping config for `ProductAddToCartComponent` cms component inside the child injector of your lazy-loaded module. See https://github.com/SAP/spartacus/blob/34c9a5a4d78f76de931e89ef7c305d3b93c2fcc5/feature-libs/cart/base/components/add-to-cart/add-to-cart.module.ts#L11-L20 . Although it wouldn't cause any errors, I believe, this would be semantically incorrect, as only one lazy-loaded chunk should contain a cms mapping for a certain component type. 

    Btw. Now I realized there is an important problem with too many responsibilities of `AddToCartModule`: exposing a component's selector for static usage + providing a cms mapping for it. As a possible improvement in the future we could move out the cms mapping's providers to a separate method so importing the module won't provide the mapping by default. See the example below:
   ```ts
    @NgModule({
      imports: [CommonModule, ReactiveFormsModule, I18nModule, ItemCounterModule],
      declarations: [AddToCartComponent],
      exports: [AddToCartComponent],
    })
    export class AddToCartModule {
      public static forRoot() {
        return {
          ngModule: AddToCartModule,
          providers: [
              cmsComponents: {
                ProductAddToCartComponent: {
                  component: AddToCartComponent,
                  data: {
                    inventoryDisplay: false,
                  },
                },
              },
            }),
          ],
        };
      }
    }
  ```

sam-garland:
@Platonn - 

I have removed the redundant module references as requested since these are not needed at all.

The issue I was trying to solve with the feature check is not related to importing the `AddToCartComponent` code. The issue is that if the end user runs the schematics in `@spartacus/schematics` and enables the EPD Visualization Integration feature, but does not enable the Cart feature then the `ActiveCartService` implementation is not loaded in the storefront application (since `facadeProviders` is not being included in the `providers` array of any loaded `NgModule`). This leads to visible `CompactAddToCartComponent` buttons that do not do anything when clicked being rendered.

I think it is entirely reasonable to use the visual spare part picking without having the Cart feature enabled (there are no doubt other kinds of actions that might be useful aside from Add to Cart). So I don't think it would be good to make the Cart feature a required dependency of the EPD Visualization Integration feature.

When the Cart feature is not enabled, the customer would most likely need to replace the `VisualPickingProductListComponent` to change the templates. However I was trying to achieve a nicer "out of the box" behaviour when the customer creates a storefront using schematics. I was hoping to have consistent/reasonable behaviour regardless of which other features the user enables.

Using a slot for the button in the list item template would add some additional flexibility, but I would think the same problem would exist when the Cart feature is not enabled. If the slot contains a `CompactAddToCartComponent` button (as it would do by default) the `CompactAddToCartComponent` button would still be rendered but not do anything when clicked.

The change I have made to hide the buttons when the Cart feature is not enabled is not an ideal solution on its own since the list layout still has space reserved for a column that does not contain anything:
image

Some more special case code would be needed to avoid the ugliness of an extra column. 

Just removing all the special case code would certainly be the simplest option.

Let me know what approach you think is best.

Platonn:
@sam-garland 

Thanks for your explanation. Now I understand your motivation for adding the check whether the storefront is capable of having a cart and adding items to it.

TBH I'm not sure the OOTB Spartacus is prepared at this moment for a partial usage of the storefront (without cart functionality). I would need to consult again with @plabadie and @WeizhengSap. Despite the customer can uncheck a  checkbox "Cart" in the CLI while running `ng add @spartacus/schematics`, I don't know whether the rest of OOTB Spartacus will work correctly then. If it's possible, then **we need to find a proper way of checking the existence of the cart functionality.**


I guess the name of the service method is a bit confusing `FeatureModulesService.isConfigured`. Maybe we should  rename to `.isLazyLoadingConfigured` or something similar? 🤷 I'll discuss with the Spartacus core team.


Checking the `featureModulesService.isConfigured(ADD_TO_CART_FEATURE)` means nothing more that the "Add to cart" button component is technically extracted to a tiny dedicated lazy loaded module (apart from the rest of the core cart functionality module) . But the absence of this configuration for the key `ADD_TO_CART_FEATURE` doesn't mean that the cart functionality is not used. The JS for add-to-cart functionality can be theoretically  bundled in a different lazy-loaded module (according to any customizations the customers may have). 

I confirm, this extra configuration for lazy-loading ADD_TO_CART_FEATURE will be created by default by schematics only in the new fresh installed Spartacus apps since 5.0. But for customers migrating their apps manually from Spartacus 4.x to 5.0 this configuration might be missing (unless we document in the migration docs it needs to be added manually or we might create migration-schematics for adding it - that's possible).

Nevertheless, the absence of this config for key ADD_TO_CART_FEATURE doesn't mean semantically that the storefront has disabled the add-to-cart functionality. Therefore I suggest no to rely on this check in your code.

Even if we would check if the `.isConfigured(CART_BASE_FEATURE)` (string key for the heart of the cart mechanism) , we would only check if the base cart module is configured for lazy loading. But not every customer uses lazy loading. I imagine there might be some customers (e.g. migrating from old  Spartacus 1.x or 2.x, where lazy loading was not yet supported nor configurable) that don't configure the lazy loading themselves so quickly after bumping the `@spartacus/*` packages version to 4.x.

I guess you only wanted to make the life of customers easier when they extend your feature and don't use cart functionality, by taking advantage of an existing check `.isConfigured()`, which seemed to be a low hanging fruit. But this condition in the code IMHO doesn't check what you expected.
**Btw. As far as I understand, the support of absent cart functionality wasn't implemented in your feature released in 4.3.0, right?**



I would need to think more and discuss with the Spartacus core team if there exists a proper check to determine the absence of the cart functionality, that you could use in your logic. But I don't want to delay the merge of this PR because of it.

**What do you think about removing the check in this PR? And we could possibly get back to the topic, when we have some idea how to check the existence of the cart functionality correctly?**

sam-garland:
I have reverted the changes to try to detect whether the Cart feature is enabled given that the check was not doing what I had assumed and cannot be relied upon.

@Platonn - Please let me know if/when there is a better solution for detecting whether the ActiveCartService can be used.

projects/storefrontapp-e2e-cypress/cypress/integration/vendor/cds/profile-tag/default-events.spec.ts
```diff
@@ -0,0 +1,489 @@
...
+  createProductQuery,
+  QUERY_ALIAS,
+} from '../../../../helpers/product-search';
+import {
+  cdsHelper,
+  strategyRequestAlias,
+} from '../../../../helpers/vendor/cds/cds';
+import { profileTagHelper } from '../../../../helpers/vendor/cds/profile-tag';
+
+describe('Profile-tag events', () => {
```
Platonn: Are those changes really related to this PR?

projects/storefrontapp-e2e-cypress/cypress/integration/vendor/epd-visualization/accessibility/visual-picking-tab-tabbing-order.e2e-spec.ts
```diff
@@ -0,0 +1,14 @@
+import {
+  configureDefaultProduct,
+  verifyTabbingOrder,
+} from '../../../../helpers/vendor/epd-visualization/visual-picking-tab';
+
+context("Tabbing order - tests don't require user to be logged in", () => {
```
Platonn, line 6: Are those changes really related to this PR?

projects/storefrontapp-e2e-cypress/cypress/helpers/vendor/epd-visualization/visual-picking-tab.ts
```diff
@@ -0,0 +1,150 @@
+export function configureDefaultProduct() {
```
Platonn, line 1: Are those changes really related to this PR?

projects/storefrontapp-e2e-cypress/cypress/support/a11y-tab.commands.ts
```diff
@@ -33,7 +39,7 @@ declare global {
   }
 }
 
-Cypress.Commands.add('pressTab', () => {
+Cypress.Commands.add('pressTab', (shiftModifier?: boolean) => {
```
Platonn, line 42: Are those changes really related to this PR?

projects/storefrontapp-e2e-cypress/cypress/support/utils/a11y-tab.ts
```diff
@@ -18,3 +18,12 @@ export function getNextFocusableElement(
     ? elements[activeElementIndex + 1]
     : elements[0];
 }
+
+export function getPreviousFocusableElement(
```
Platonn, line 22: Are those changes really related to this PR?

projects/storefrontapp-e2e-cypress/cypress/integration/vendor/epd-visualization/accessibility/visual-picking-tab-tabbing-order.e2e-spec.ts
Platonn: My bad. Looks good.

projects/storefrontapp-e2e-cypress/cypress/integration/vendor/cds/profile-tag/default-events.spec.ts
Platonn: I can see in the PR to 4.3.x we have only changed the path of this file
https://github.com/SAP/spartacus/pull/14753/files#diff-f0e7debfd7f5766491e0bcf6e6e36a7ebaf209e2cc19a22c8002657ab9d2b4c1
image

projects/storefrontapp-e2e-cypress/cypress/helpers/vendor/epd-visualization/visual-picking-tab.ts
Platonn: My bad. Looks good

projects/storefrontapp-e2e-cypress/cypress/support/utils/a11y-tab.ts
Platonn: My bad. Looks good

projects/storefrontapp-e2e-cypress/cypress/support/a11y-tab.commands.ts
Platonn: My bad. Looks good.

Sorry for spam

Platonn:
Hi @sam-garland 
I did the last check to compare changes in PR to 4.x vs. changes in this PR to develop.

I'm sorry I raised a few questions "are those changes really related to this PR" and only later I realized those changes are ok. :( 

There is one change related to CDS' e2e which I believe shouldn't be part of this PR.

Besides that, looks good! Thanks for your updates.

When changes in  CDS' e2e file are sorted out, I'm happy to approve the PR.

projects/storefrontapp-e2e-cypress/cypress/integration/vendor/cds/profile-tag/default-events.spec.ts
sam-garland: @Platonn

What happened here was that the name of the file was changed in the `develop` branch and in my set of changes and the merge process added the file that had been renamed in my set of changes:

image

I have removed the file added as a result of my renaming so that there is no change introduced relative to the `develop` branch.

Platonn (approved):

PR #14758, opened 2022-01-11
Issues with CI build on develop branch

### LINKED ISSUE #14757: issues with develop build on CI
Having the following error when running e2es in develop

> Opening Cypress...

  You passed the --group flag, but this group name has already been used for this run.
  The existing run is: https://dashboard.cypress.io/projects/k3nmep/runs
  The --group flag you passed was: B2B
  The --ciBuildId flag you passed was: 244608914
   If you are trying to parallelize this run, then also pass the --parallel flag, else pass a different group name.
   It also looks like you also passed in an explicit --ci-build-id flag.
  This is only necessary if you are NOT running in one of our supported CI providers.
  This flag must be unique for each new run, but must also be identical for each machine you are trying to --group or run in --parallel.
  https://on.cypress.io/run-group-name-not-unique

hackergil:
Fixes #14757


giancorderoortiz (approved):
Removes execution of flaky tests. 
Runs both core and supplemental tests.

PR #14763, opened 2022-01-13
fix: Delivery mode infinite loop

bgambocjaviniar:
This PR fixes the following:

1. user starts the checkout, and adds / selects an existing address
2. user navigates to delivery mode step, and selects a delivery mode
3. user wants to change an address used in the 1st step, and they navigate back to the address checkout step
4. they add an address, and press continue
5. there's an infinite spinner shown now. 

The reason for the above bug is because the checkout details (aka super query) was reset when a new address was added, and therefore the delivery mode step was waiting for it to be available. The correct fix for this is to make sure the address checkout step completes all of its operations before allowing the user to move to the next step.

### DIFF
--- a/feature-libs/checkout/base/components/checkout-shipping-address/checkout-shipping-address.component.ts
+++ b/feature-libs/checkout/base/components/checkout-shipping-address/checkout-shipping-address.component.ts
@@ -1,7 +1,10 @@
 import { ChangeDetectionStrategy, Component, OnInit } from '@angular/core';
 import { ActivatedRoute } from '@angular/router';
 import { ActiveCartFacade } from '@spartacus/cart/main/root';
-import { CheckoutDeliveryAddressFacade } from '@spartacus/checkout/base/root';
+import {
+  CheckoutDeliveryAddressFacade,
+  CheckoutDeliveryModesFacade,
+} from '@spartacus/checkout/base/root';
 import {
   Address,
   getLastValueSync,
@@ -10,7 +13,7 @@ import {
 } from '@spartacus/core';
 import { Card } from '@spartacus/storefront';
 import { BehaviorSubject, combineLatest, Observable } from 'rxjs';
-import { filter, map, tap } from 'rxjs/operators';
+import { filter, map, switchMap, take, tap } from 'rxjs/operators';
 import { CheckoutStepService } from '../services/checkout-step.service';
 
 export interface CardWithAddress {
@@ -41,7 +44,8 @@ export class CheckoutShippingAddressComponent implements OnInit {
     protected activatedRoute: ActivatedRoute,
     protected translationService: TranslationService,
     protected activeCartFacade: ActiveCartFacade,
-    protected checkoutStepService: CheckoutStepService
+    protected checkoutStepService: CheckoutStepService,
+    protected checkoutDeliveryModesFacade: CheckoutDeliveryModesFacade
   ) {}
 
   get isGuestCheckout(): boolean {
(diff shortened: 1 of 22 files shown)


feature-libs/checkout/base/core/facade/checkout-delivery-address.service.ts
```diff
@@ -66,7 +66,8 @@ export class CheckoutDeliveryAddressService
                     },
                     DeliveryAddressCreatedEvent
                   )
-                )
+                ),
+                switchMap((address) => this.setDeliveryAddress(address))
```
bgambocjaviniar: 🤔

feature-libs/checkout/base/components/checkout-shipping-address/checkout-shipping-address.component.ts
```diff
@@ -168,24 +172,33 @@ export class CheckoutShippingAddressComponent implements OnInit {
   }
 
   addAddress(address: Address | undefined): void {
-    if (address) {
```
bgambocjaviniar, line 149: 👍  was thinking of flattening later after we fix the issue, but coolio that you did it

feature-libs/checkout/base/components/checkout-shipping-address/checkout-shipping-address.component.ts
```diff
@@ -142,17 +174,42 @@ export class CheckoutShippingAddressComponent implements OnInit {
   }
 
   selectAddress(address: Address): void {
-    this.checkoutDeliveryAddressFacade.setDeliveryAddress(address);
+    this.busy$.next(true);
+    this.checkoutDeliveryAddressFacade
+      .setDeliveryAddress(address)
+      .pipe(switchMap(() => this.prepareDeliveryModes$))
```
bgambocjaviniar: Not sure about this use case as we're basically doubling the work on what is already done in the command already and its side effects in the listeners, such as deleting the delivery mode and resetting query and stuff. But in the switchMap stream, we're doing it again.

In conclusion, I don't think we need to touch this stream other than creating the address*

projects/storefrontapp-e2e-cypress/cypress/helpers/b2b/b2b-checkout.ts
```diff
@@ -29,6 +29,20 @@ import {
 } from '../checkout-flow';
 import { generateMail, randomString } from '../user';
 
+export const GET_CHECKOUT_SUPER_QUERY_ENDPOINT_ALIAS =
+  'GET_CHECKOUT_SUPER_QUERY';
+
+export function interceptB2BCheckoutSuperQueryEndpoint() {
```
znikola: Let's call it "checkout details" instead of the "super query"

projects/storefrontapp-e2e-cypress/cypress/helpers/b2b/b2b-checkout.ts
```diff
@@ -29,6 +29,20 @@ import {
 } from '../checkout-flow';
 import { generateMail, randomString } from '../user';
 
+export const GET_CHECKOUT_SUPER_QUERY_ENDPOINT_ALIAS =
+  'GET_CHECKOUT_SUPER_QUERY';
```
znikola: Let's call it "checkout details" instead of the "super query"

projects/storefrontapp-e2e-cypress/cypress/helpers/b2b/b2b-checkout.ts
```diff
@@ -148,15 +162,23 @@ export function selectAccountShippingAddress() {
 }
 
 export function selectAccountDeliveryMode() {
+  const getCheckoutSuperQueryAlias = interceptB2BCheckoutSuperQueryEndpoint();
```
znikola: Let's call it "checkout details" instead of the "super query"

projects/storefrontapp-e2e-cypress/cypress/helpers/checkout-flow.ts
```diff
@@ -20,6 +20,20 @@ import {
...
+export function interceptB2CCheckoutSuperQueryEndpoint() {
+  cy.intercept(
+    'GET',
+    `${Cypress.env('OCC_PREFIX')}/${Cypress.env(
+      'BASE_SITE'
+    )}/users/current/carts/**/*?fields=deliveryAddress(FULL),deliveryMode(FULL),paymentInfo(FULL)*`
+  ).as(GET_CHECKOUT_SUPER_QUERY_ENDPOINT_ALIAS);
+
+  return GET_CHECKOUT_SUPER_QUERY_ENDPOINT_ALIAS;
+}
```
znikola: Let's call it "checkout details" instead of the "super query"

projects/storefrontapp-e2e-cypress/cypress/helpers/checkout-forms.ts
```diff
@@ -32,47 +32,40 @@ export function fillShippingAddress(
...
-        cy.get('[formcontrolname="postalCode"]')
-          .clear()
-          .type(shippingAddress.address.postal);
-        cy.get('[formcontrolname="phone"]').clear().type(shippingAddress.phone);
-        if (submitForm) {
-          cy.get('button.btn-primary').should('be.enabled').click();
-        }
-      });
+
+  cy.get('cx-address-form').within(() => {
```
znikola, line 36: 👍🏻

znikola (changes requested):

feature-libs/checkout/base/core/facade/checkout-delivery-address.service.ts
znikola: We discussed it.

feature-libs/checkout/base/components/checkout-shipping-address/checkout-shipping-address.component.ts
znikola: Sounds good.

PR #14764, opened 2022-01-13
chore: Update todo + fix some misc items

### LINKED ISSUE #8880: Changed guards in checkout components
With B2B Checkout we no longer have all guards for delivery mode, payment method, review that we previously had.
Check if that was an incident or it is needed for B2B.

Also why `CheckoutStep` was altered?

bgambocjaviniar:
This PR addresses some of the misc tasks:
- improve e2es
- fix build warnings for ng-packagr.json
- cleans up / updates the `todo.md`, and removes the previously done tasks
- renames events to have the `Checkout` prefix
- removes `TODO(#8880):`, and closes #8880


feature-libs/checkout/base/root/events/checkout.events.ts
```diff
@@ -105,31 +105,31 @@ export abstract class CheckoutDeliveryModeEvent extends CheckoutEvent {}
...
 /**
  * Emit this event to force delivery modes reset
  */
-export class ResetDeliveryModesEvent extends CheckoutDeliveryModeEvent {
+export class ResetCheckoutDeliveryModesEvent extends CheckoutDeliveryModeEvent {
   /**
    * Event's type
    */
-  static readonly type = 'ResetDeliveryModesEvent';
+  static readonly type = 'ResetCheckoutDeliveryModesEvent';
```
znikola: Does it makes sense if we swap the `Reset` and `Checkout`?
All events start with "checkout" prefix, and I'm wondering if it makes sense to name this one `CheckoutResetDeliveryModesEvent`?

todo.md
```diff
@@ -61,16 +46,15 @@
...
       };
-10. align the event names - prefix them with Checkout?
-11. Rename b2b and repl endpoint config keys - https://github.com/SAP/spartacus/pull/14495/files#r760445274
-12. When we were renaming components / folders to have the checkout prefix, we intentionally left out the components' prefix untouched.
-   4.  Rename the checkout components' selectors to have the checkout prefix?
-13. query debounce - `feature/query-debounce`
-14. converters and any - https://github.com/SAP/spartacus/pull/14165#discussion_r751912800
-15. Look into `TODO(#8880):`
-16. check changes to the old checkout
+6. Rename b2b and repl endpoint config keys - https://github.com/SAP/spartacus/pull/14495/files#r760445274 (I think it's best to not rename them. What do you think?)
```
znikola: ok, let's keep them as is

todo.md
```diff
@@ -61,16 +46,15 @@
...
+7. When we were renaming components / folders to have the checkout prefix, we intentionally left out the components' prefix untouched.
+   4.  Rename the checkout components' selectors to have the checkout prefix? (revisit as I think it's fine)
```
znikola: You think it's fine to rename it, or to keep it as is?

todo.md
```diff
@@ -61,16 +46,15 @@
...
+8. query debounce - `feature/query-debounce`
+9. converters and any - https://github.com/SAP/spartacus/pull/14165#discussion_r751912800
+10. Look into `TODO(#8880):`
```
znikola: I think it's pretty much addressed by you, so we can remove this todo?

znikola (changes requested):

todo.md
```diff
@@ -1,34 +1,19 @@
...
    - (related to above) sample data changes: https://github.tools.sap/cx-commerce/spartacussampledata/pull/211 (source: https://sap-cx.slack.com/archives/C02L8BUATM5/p1638283007005900)
-7.  check the event listeners for the following scenario:
+3.  check the event listeners for the following scenario:
     1.  a user started the checkout, entered their delivery address, and set the delivery mode, and the data is sent on the back-end for the active cart
     2.  the user changes their mind, and navigates away from the checkout page to homepage, and refreshes the browser.
     3.  after it, they decide to change their address in the profile menu. 
     4.  if they now start the checkout (and LL the feature), the current back-end data is _not_ valid for the active cart - we must reset the set delivery mode, and load the supported delivery modes again for the new address.
     5.  if the listener was in the root module, it can listen to the userupdateaddress event, ll the checkout, and issue a reset query event
```
znikola: I just checked this, and it seems to be working as expected, so we can remove it

feature-libs/checkout/base/root/events/checkout.events.ts
znikola: The same goes for: `ReloadCheckoutQueryEvent` and `ResetCheckoutQueryEvent`.

feature-libs/checkout/base/root/events/checkout.events.ts
```diff
@@ -105,31 +105,31 @@ export abstract class CheckoutDeliveryModeEvent extends CheckoutEvent {}
 /**
  * Emit this event to force delivery modes reload
  */
-export class ReloadDeliveryModesEvent extends CheckoutDeliveryModeEvent {
+export class CheckoutReloadDeliveryModesEvent extends CheckoutDeliveryModeEvent {
```
znikola, line 108: Can we add a `TODO:#checkout - not used?` for `CheckoutReloadDeliveryModesEvent` and `ReloadCheckoutQueryEvent`? These don't seem to be dispatched anywhere

znikola (changes requested):

todo.md
bgambocjaviniar: basically, 3. right? for `check the event listeneres for the following scenario`


* I removed it in the todo now though

todo.md
bgambocjaviniar: I was fine to keep it as it is, but after thoroughly thinking about it again, I think I'll change it now to include the prefix in the selector.

One reason why I said it was fine was because it's just for the dom, and I don't know if having checkout-*anything always to be displayed as an element tag is worth it.

But another reason against this logic would be that when we generate angular components using angular cli, it usually puts the entire folder name with a prefixed of wtv we have in angular.json. Therefore, I am going to include the prefix to have the 'entire' folder name for convention reasons.

bgambocjaviniar: Actually, before I do this change, I need to talk about whether we're still keeping the old checkout (one of my questions in our DM)

Because when I mentioned overhead, we would need to update the reference in old checkout to cx-checkout-* as well.

feature-libs/checkout/base/root/events/checkout.events.ts
bgambocjaviniar: I didn't add the todo because I was thinking that we set this up in the service in a way that a customer can always fire those events to 'reload' them anyhow if they ever need it

todo.md
znikola: let's talk about it, and do it in a separate PR if decide to go for it .

feature-libs/checkout/base/root/events/checkout.events.ts
znikola: I guess that was the idea. Not sure if it's realistic to actually be used. Maybe we can document it, so the events will have some exposure and a potential to be used. I added this to todo.md.

todo.md
znikola: Yep, thanks!

PR #14647, opened 2021-12-17
feature/CXINT-22 CDC-COMM:SPA UI - Headless Commerce: Login & Registration

SherwinVarghese:
Requirement Details:

https://wiki.wdf.sap.corp/wiki/display/GIGYA/SAP+Commerce+Cloud+-+Built-in+Extension

Login & Registration:

The current built-in extension allows the registration and login flows to be performed without the use of the screensets. Instead, the WebSDK functions - such as gigya.accounts.login or gigya.accounts.register - can be used, as per below recording.

PR #14768, opened 2022-01-13
CXCDS-2462 Fix for displaying Merchandising Carousel in b2b Spartacus

marek-sap:


hackergil (changes requested):
Hi Marek. Can you please add a unit test for your change? Thanks.

marek-sap:
Here is the original ticket: 
https://jira.tools.sap/browse/CXCDS-2462

We needed to implement this fix, because displaying carousel is related to consent changed event (and consent currently are not shown in B2B storefront)

integration-libs/cds/src/merchandising/facade/cds-merchandising-user-context.service.ts
```diff
@@ -65,7 +72,17 @@ export class CdsMerchandisingUserContextService {
   }
 
   private getConsentReferenceContext(): Observable<MerchandisingUserContext> {
+    let operators: MonoTypeOperatorFunction<ConsentChangedPushEvent>[] = [];
+    if (!!environment.b2b) {
```
Platonn, line 76: We shouldn't use the `environment` in the libs. For more, see comment https://github.com/SAP/spartacus/pull/14647/files#r771495613

marek-sap: Ok, then how to distinguish in libs if spartacus is working with b2b enabled or not?

Platonn: Did you try checking the `channel` property of the `BaseSite` model returned from the method `BaseSiteService.get()`?

Platonn: Ideally, I'd love if we could avoid depending on the info from the `BaseSiteService` and it would be great if you wouldn't need the conditional `startWith` ;) Just to double check: Did you explore alternative approaches to achieve your business goal, that don't depend on reading the `BaseSite` data (in other workds: that don't check b2c vs b2b)?

Additional note on performance: 
The method `.get()` will return immediately in case when the data have been loaded before (it is the case for example when the [automatic basesite detection by URL was enabled](https://sap.github.io/spartacus-docs/automatic-context-configuration/)).
But will return with a delay (because of the roundtrip to the backend for data), when the method was not accessed before.

Therefore, if you discover performance slowdown when depending on the `.get()` method, you might consider triggering the load of data earlier, if only you find it justified.

Platonn: @znikola helped me realize that currently by default the call for `BaseSite` model won't return the `channel` property. To make it work, we need to add `channel` it in the `fields?` in the configuration of the base site endpoint.
https://github.com/SAP/spartacus/blob/807f6ee39fb7f9455b17c167be9683adf5483694/projects/core/src/occ/adapters/site-context/default-occ-site-context-config.ts#L13

--

But again, I'd love to avoid depending in CDS library on the info about b2b/b2c. Let's discuss first if it's definitely needed

cc @znikola @bgambocjaviniar

Issue #14710, opened 2021-12-24
Follow-up for Delivery Mode component: use spinner for future checkout refactoring


wangzixi-diablo:
This is a follow-up for https://github.com/SAP/spartacus/issues/14386.

In https://github.com/SAP/spartacus/issues/14386, the delivery mode radio input control is disabled temporarily when there's an on-going HTTP PUT request sent and executed in commerce cloud backend server. The radio input control is re-enabled once the HTTP PUT request is finished or responded with error.

Nik and Brian are doing checkout refactoring verification and give the following feedback based on the fix above:

1. remove feature flag usage as the change should be considered as a bug fix on important patches, in this case feature flag is not needed.
2. use a spinner to hide the input control, in order to conform with all the spinners we have in checkout. This is necessary for the discussion on the user interface and experience in the new checkout library with UX team in January 2022.

PR #14798, opened 2022-01-18
feat: use spinner to hide delivery mode radio input when there's ongoing HTTP put request

wangzixi-diablo:
fixes: #14710

- the delivery mode radio input control is hidden temporarily by a spinner when there's an on-going HTTP PUT request sent and executed in commerce cloud backend server. 

- We introduce a new constructor dependency, CheckoutService, in CheckoutDeliveryService. We mark it as optional to avoid potential compilation error when you upgrade to new minor version and have previously extended it by calling super() constructor with less parameters. 

- in CheckoutDeliveryService we evaluate availability of CheckoutService dependency. If it's available, use its method isLoading to return checkout loading flag, otherwise fallback to the old NgRx store approach.

- we recommend you to pass this new constructor dependency in super in case you have extended CheckoutDeliveryService.

### DIFF
--- a/projects/core/src/checkout/facade/checkout-delivery.service.ts
+++ b/projects/core/src/checkout/facade/checkout-delivery.service.ts
@@ -8,6 +8,7 @@ import {
   take,
   tap,
   withLatestFrom,
+  map,
 } from 'rxjs/operators';
 import { UserIdService } from '../../auth/user-auth/facade/user-id.service';
 import { ActiveCartService } from '../../cart/facade/active-cart.service';
@@ -26,6 +27,7 @@ import {
   StateWithCheckout,
 } from '../store/checkout-state';
 import { CheckoutSelectors } from '../store/selectors/index';
+import { CheckoutService } from './checkout.service';
 
 @Injectable({
   providedIn: 'root',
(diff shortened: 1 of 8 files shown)


projects/core/src/checkout/facade/checkout-delivery.service.ts
```diff
@@ -34,7 +36,8 @@ export class CheckoutDeliveryService {
   constructor(
     protected checkoutStore: Store<StateWithCheckout | StateWithProcess<void>>,
     protected activeCartService: ActiveCartService,
-    protected userIdService: UserIdService
+    protected userIdService: UserIdService,
+    protected checkoutService: CheckoutService
```
znikola: We should do this in a non-breaking way.
I suggest to mark the injection with `@Optional`, and as TS optional (`checkoutService?: CheckoutService`).
If the service is provided and present, we can use it.

projects/core/src/checkout/facade/checkout-delivery.service.ts
```diff
@@ -34,7 +36,8 @@ export class CheckoutDeliveryService {
...
   ) {}
```
znikola, line 42: It would be good to notify the customer about this change in case they are extending this service, so that they can pass it to `super()` call.
The best way to do this would be to update the 3.4 docs, but also write this explanation in the body of this PR, and just copy it to the merge commit when merging. This is useful for clients who will go to our GH and check recent commits.

znikola (changes requested):

znikola (approved):

PR #14809, opened 2022-01-19
chore: Centralize store usage

znikola:
This PR centralizes the usage of the legacy ngrx store.

### DIFF
--- a/feature-libs/checkout/base/core/facade/checkout-delivery-address.service.spec.ts
+++ b/feature-libs/checkout/base/core/facade/checkout-delivery-address.service.spec.ts
@@ -1,5 +1,5 @@
 import { inject, TestBed } from '@angular/core/testing';
-import { MockStore, provideMockStore } from '@ngrx/store/testing';
+import { provideMockStore } from '@ngrx/store/testing';
 import { ActiveCartFacade } from '@spartacus/cart/main/root';
 import {
   CheckoutDeliveryAddressClearedEvent,
@@ -11,10 +11,10 @@ import {
 import {
   Address,
   EventService,
+  LoadUserAddressesEvent,
   OCC_USER_ID_ANONYMOUS,
   OCC_USER_ID_CURRENT,
   QueryState,
-  UserActions,
   UserIdService,
 } from '@spartacus/core';
 import { of } from 'rxjs';
@@ -61,7 +61,6 @@ class MockCheckoutQueryFacade implements Partial<CheckoutQueryFacade> {
 describe(`CheckoutDeliveryAddressService`, () => {
   let service: CheckoutDeliveryAddressService;
   let connector: CheckoutDeliveryAddressConnector;
-  let store: MockStore;
   let checkoutQuery: CheckoutQueryFacade;
   let eventService: EventService;
   let userIdService: UserIdService;
@@ -84,7 +83,6 @@ describe(`CheckoutDeliveryAddressService`, () => {
 
     service = TestBed.inject(CheckoutDeliveryAddressService);
     connector = TestBed.inject(CheckoutDeliveryAddressConnector);
-    store = TestBed.inject(MockStore);
     checkoutQuery = TestBed.inject(CheckoutQueryFacade);
     eventService = TestBed.inject(EventService);
     userIdService = TestBed.inject(UserIdService);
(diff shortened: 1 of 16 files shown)


feature-libs/checkout/base/core/facade/checkout-delivery-address.service.spec.ts
```diff
@@ -147,26 +143,26 @@ describe(`CheckoutDeliveryAddressService`, () => {
...
-        new UserActions.LoadUserAddresses(mockUserId)
+      expect(eventService.dispatch).toHaveBeenCalledWith(
+        { userId: mockUserId },
+        LoadUserAddressesEvent
       );
     });
 
-    // TODO:#deprecation-checkout Replace with event testing once we remove ngrx store.
-    it(`should NOT dispatch UserActions.LoadUserAddresses when the use is anonymous`, () => {
+    it(`should NOT dispatch LoadUserAddressesEvent when the use is anonymous`, () => {
```
znikola:
```suggestion
    it(`should NOT dispatch LoadUserAddressesEvent when the user is anonymous`, () => {
```

feature-libs/checkout/base/core/facade/checkout-delivery-modes.service.ts
```diff
@@ -99,16 +101,14 @@ export class CheckoutDeliveryModesService
...
-                   */
-                  this.store.dispatch(
-                    new CartActions.LoadCart({
+                  this.eventService.dispatch(
+                    {
                       userId,
                       cartId,
-                    })
+                      // TODO:#checkout - check this
+                      cartCode: cartId,
```
bgambocjaviniar, line 112: yup this is fine, but cartId would either be the guid for anonymous or the cartId similarly to cartCode

bgambocjaviniar: Can be seen in `feature-libs/cart/saved-cart/core/events/saved-cart-event.builder.ts`, where cartCode is the actual code and cartId is anon/actualCode

feature-libs/checkout/scheduled-replenishment/core/facade/checkout-scheduled-replenishment.service.ts
```diff
@@ -40,12 +38,15 @@ export class CheckoutScheduledReplenishmentService
...
                 this.checkoutFacade.setOrder(replenishmentOrder);
-                /**
-                 * TODO:#deprecation-checkout We have to keep this here, since the cart feature is still ngrx-based.
-                 * Remove once it is switched from ngrx to c&q.
-                 * We should dispatch an event, which will remove the cart
-                 */
-                this.store.dispatch(new CartActions.RemoveCart({ cartId }));
+                this.eventService.dispatch(
+                  {
+                    userId,
```
bgambocjaviniar, line 43: To me, it doesn't hurt to add extra payload to the event, but do we want to show the userId. Also, should we use cartCode? 

I think it will confuse the customers haha if we include cartCode when cartId is both guid and the actual cartCode too depending on the user

feature-libs/checkout/base/root/events/checkout-legacy-store-event.listener.ts
```diff
@@ -0,0 +1,104 @@
...
+          new CartActions.LoadCart({
+            userId,
+            cartId,
+          })
+        );
+      })
+    );
+
+    this.subscriptions.add(
+      this.eventService.get(RemoveCartEvent).subscribe(({ cartId }) => {
```
bgambocjaviniar, line 90: for things like LoadCartEvent and RemoveCartEvent, when we actually dispatch them in our app, we added additional payload. I understand we're destructing it here, but I guess we just want to give the users options when they subscribe to the event to use the other data right?

projects/core/src/user/events/user.events.ts
```diff
@@ -1,23 +1,30 @@
 import { CxEvent } from '../../event/cx-event';
 import { Address } from '../../model/address.model';
 
-export abstract class UserAddressEvent extends CxEvent {}
+export abstract class UserAddressEvent extends CxEvent {
+  userId: string;
```
bgambocjaviniar, line 5: 👍  awesome instead of having to repeat ourselves

bgambocjaviniar (changes requested):
more of questions than changes below, but it's ready to be merged

also just to make aware in cart-lib
there will exist a createcartevent https://github.com/SAP/spartacus/pull/14839/files

feature-libs/checkout/base/root/events/checkout-legacy-store-event.listener.spec.ts
```diff
@@ -0,0 +1,102 @@
...
+      ],
+    });
+
+    TestBed.inject(CheckoutLegacyStoreEventListener);
+    store = TestBed.inject(MockStore);
+  });
+
+  describe(`onUserAddressAction`, () => {
+    it(`should dispatch UserActions.LoadUserAddresses`, () => {
+      spyOn(store, 'dispatch').and.stub();
```
bgambocjaviniar: think we can move this under line 42 since all use test cases needs to stub the dispatch

feature-libs/checkout/base/core/facade/checkout-delivery-modes.service.ts
znikola: In this case it will _always_ be `cart.code`, because we did precondition check for the logged-in cart.

feature-libs/checkout/scheduled-replenishment/core/facade/checkout-scheduled-replenishment.service.ts
znikola: Not sure what else can we include?
As for the confusion - I wrote some jsdocs in the CartEvent, let me know if it's not clear.

feature-libs/checkout/base/root/events/checkout-legacy-store-event.listener.ts
znikola: Yes.
Also, CDS is leveraging the additional payload which we don't require here

feature-libs/checkout/base/root/events/checkout-legacy-store-event.listener.spec.ts
znikola: yep, nice :)

Issue #14810, opened 2022-01-19
avoid breaking changes in CheckoutDeliveryService constructor change for 4.2.x


wangzixi-diablo:
- When back-porting https://github.com/SAP/spartacus/pull/14714 from develop branch to 4.2.x using https://github.com/SAP/spartacus/pull/14777, a breaking change in CheckoutDeliveryService constructor is introduced.  This breaking change should be avoided by following this [guideline](https://sap.github.io/spartacus-docs/breaking-changes/#adding-new-constructor-dependencies-in-minor-versions).

- Meanwhile, the unneeded `FeaturesConfigModule` in delivery-mode.component.spec.ts should also be removed.

PR #14813, opened 2022-01-19
fix: avoid breaking changes in CheckoutDeliveryService constructor change for 4.2.x

wangzixi-diablo:
fixes: #14810

### DIFF
--- a/feature-libs/checkout/core/facade/checkout-delivery.service.ts
+++ b/feature-libs/checkout/core/facade/checkout-delivery.service.ts
@@ -42,7 +42,7 @@ export class CheckoutDeliveryService implements CheckoutDeliveryFacade {
     @Optional() protected checkoutService?: CheckoutService
   ) {}
 
-  private isCheckoutDetailsLoading$ = this.checkoutService
+  protected isCheckoutDetailsLoading$ = this.checkoutService
     ? this.checkoutService.isLoading()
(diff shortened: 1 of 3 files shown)


feature-libs/checkout/core/facade/checkout-delivery.service.ts
```diff
@@ -39,9 +39,13 @@ export class CheckoutDeliveryService implements CheckoutDeliveryFacade {
     protected processStateStore: Store<StateWithProcess<void>>,
     protected activeCartService: ActiveCartService,
     protected userIdService: UserIdService,
-    protected checkoutService: CheckoutService
+    @Optional() protected checkoutService?: CheckoutService
   ) {}
 
+  private isCheckoutDetailsLoading$ = this.checkoutService
```
znikola: Nice. Let's mark it as `protected`, per our new convention.

znikola (changes requested):

feature-libs/checkout/core/facade/checkout-delivery.service.ts
znikola: I add a type, e.g. `protected isCheckoutDetailsLoading$: Observable<something>...`

znikola (dismissed):

znikola (approved):

PR #14817, opened 2022-01-20
feat: Quick Order Screen Reader improvements (#13654)

### LINKED ISSUE #13654: Quick Order: Accessibility Pattern revision

dydome:
close #13654


developpeurweb (changes requested):
The alert and the input field hint work well for all SR.
Now, for some reason, the search query highlight stays as part of the `alt` attribute, it would be ideal to clean it up, if not possible, then in this case, since the image is NOT-a-link, it's OK to set it up as `role="presentation"` so it gets ignored by the SR. We have the product name being a link and also focusable therefore `role="presentation"` it's fine, if no other solution can be found.
Screen Shot 2022-01-24 at 1 32 27 PM

dydome:
@developpeurweb I have changed alt content to the product name and also added role="presentation" as it is good 
 way I think.

feature-libs/cart/quick-order/components/quick-order/table/item/quick-order-item.component.html
```diff
@@ -9,8 +9,9 @@
         >
           <cx-media
             [container]="entry?.product?.images?.PRIMARY"
-            [alt]="entry.product.summary ?? entry.product.name"
+            [alt]="entry?.product?.name"
             format="cartIcon"
+            role="presentation"
```
developpeurweb: Seems the `alt` attribute is clean and works well now, so please remove `role="presentation"` as both cannot coexist. Their purposes contradict each other.

developpeurweb (changes requested):

feature-libs/cart/quick-order/components/quick-order/table/item/quick-order-item.component.html
dydome: Removed

developpeurweb (approved):

Issue #13910, opened 2021-09-28
[BLOCKED #14989] Migrate A11Y Tabbing Tests (replaces tabbing-order.flaky.e2e-spec.ts)


Zeyber:
This is the master ticket to migrate tests from the current `tabbing-order.e2e-spec.ts` spec to the new method defined in #13686. This test is targeted at migrating the existing tests already and should clean up Tabbing Order Config files.

Tests:
- [ ] Header - Desktop (not logged in) (#14819)
- [ ] Header - Mobile (not logged in) (#14819)
- [ ] Header Sub Categories - Desktop (#14819)
- [ ] Header Sub Categories - Mobile (#14819)
- [ ] Home page (#14819)
- [ ] Footer (#14819)
- [ ] Login page (#14820)
- [ ] Register page (#14822)
- [ ] Reset password (#14823)
- [ ] Product List (#14151)
- [ ] Product Page (#14824)
- [ ] Product Page Tabs (#14824)
- [ ] Add to cart
- [ ] Cart (#14825)
- [ ] Store finder (#14272)
- [ ] Store finder search results (#14272)
- [ ] Store details page (#14272)
- [ ] Countries list (#14272)
- [ ] Stores list (#14272)
- [ ] Stock Notification
- [ ] ASM
- [ ] Header - Desktop (logged in) (#14820)
- [ ] Header - Mobile (logged in) (#14820)
- [ ] My Account - Desktop (#14820)
- [ ] My Account - Mobile (#14820)
- [ ] Checkout (#14405)
- [ ] Shipping address (#14405)
- [ ] Delivery mode (#14405)
- [ ] Payment details (#14405)
- [ ] Review order (#14405)
- [ ] Order History
- [ ] Notification preference
- [ ] Change password (#14823)
- [ ] Personal details
- [ ] Update email
- [ ] Close account
- [ ] Consent Management
- [ ] Address Book (Form)
- [ ] Payment Details
- [ ] Order Details
- [ ] Wishlist
- [ ] Save for later
- [ ] My Interest
- [ ] Coupons
- [ ] My-coupons
- [ ] Stock Notification
- [ ] Consignment Tracking
- [ ] Saved Cart - #14945

Zeyber:
Blocked by #14989 due to disruptive changes to happen with `domSnapshot()`(eg. changing snapshot formats).

PR #14823, opened 2022-01-20
[BLOCKED #14989] test: Reset and forgot password a11y tabbing test migration

Zeyber:
Part of #13910
Reopening from #14141


Zeyber:
Blocked by #14989 due to disruptive changes to happen with `domSnapshot()`(eg. changing snapshot formats).

PR #14828, opened 2022-01-21
fix: Update checkout leaks and normalize checkout steps

bgambocjaviniar:


feature-libs/checkout/b2b/components/checkout-delivery-address/checkout-delivery-address.component.ts
```diff
@@ -34,7 +34,7 @@ export interface CardWithAddress {
 }
 
 @Component({
-  selector: 'cx-shipping-address',
+  selector: 'cx-delivery-address',
```
znikola, line 37: If we decide to stick with this change, let's document it (see slack)

projects/assets/src/translations/en/payment.ts
```diff
@@ -14,7 +14,7 @@ export const payment = {
     saveAsDefault: 'Save as default',
     setAsDefault: 'Set as default payment method',
     billingAddress: 'Billing address',
-    sameAsShippingAddress: 'Same as shipping address',
+    sameAsDeliveryAddress: 'Same as shipping address',
```
znikola:
```suggestion
    sameAsDeliveryAddress: 'Same as delivery address',
```

znikola: Also document

feature-libs/checkout/base/components/checkout-delivery-mode/checkout-delivery-mode.component.html
```diff
@@ -41,26 +41,22 @@ <h2 class="cx-checkout-title d-none d-lg-block d-xl-block">
...
-            deliveryModeInvalid || (changeDeliveryModeInProgress$ | async)
-          "
-          (click)="next()"
-        >
-          {{ 'common.continue' | cxTranslate }}
-        </button>
-      </div>
+  <div class="row cx-checkout-btns">
+    <div class="col-md-12 col-lg-6">
+      <button class="btn btn-block btn-action" (click)="back()">
```
znikola: Should we also disable the _back_ button if busy$?

feature-libs/checkout/base/components/checkout-delivery-mode/checkout-delivery-mode.component.ts
```diff
@@ -28,17 +28,14 @@ export class CheckoutDeliveryModeComponent implements OnInit, OnDestroy {
 
   selectedDeliveryModeCode$: Observable<string | undefined>;
   supportedDeliveryModes$: Observable<DeliveryMode[]>;
-  continueButtonPressed = false;
```
znikola, line 31: 🙏🏻

feature-libs/checkout/base/components/checkout-payment-method/checkout-payment-method.component.html
```diff
@@ -1,8 +1,9 @@
-<ng-container *ngIf="cards$ | async as cards">
```
znikola, line 1: Nice!

feature-libs/checkout/base/components/checkout-payment-method/checkout-payment-method.component.html
```diff
@@ -1,8 +1,9 @@
...
-  <h2 class="cx-checkout-title d-none d-lg-block d-xl-block">
-    {{ 'paymentForm.payment' | cxTranslate }}
-  </h2>
-  <ng-container *ngIf="!shouldRedirect && !(isLoading$ | async); else loading">
+<h2 class="cx-checkout-title d-none d-lg-block d-xl-block">
+  {{ 'paymentForm.payment' | cxTranslate }}
+</h2>
+
+<ng-container *ngIf="cards$ | async as cards; else loading">
+  <ng-container *ngIf="!shouldRedirect && !(isUpdated$ | async); else loading">
```
znikola: Does it make sense to merge these two ng-containers?

feature-libs/checkout/base/components/checkout-payment-method/checkout-payment-method.component.ts
```diff
@@ -36,20 +36,23 @@ import { CheckoutStepService } from '../services/checkout-step.service';
...
   isGuestCheckout = false;
   newPaymentFormManuallyOpened = false;
-  paymentSavingInProgress$ = new BehaviorSubject<boolean>(false);
-  shouldRedirect: boolean;
+  shouldRedirect: boolean = false;
 
   backBtnText = this.checkoutStepService.getBackBntText(this.activatedRoute);
 
-  protected deliveryAddress: Address | undefined;
+  isUpdated$: Observable<boolean> = combineLatest([
```
znikola: I think we chose the wrong name in the delivery-address cmp. It seems better to name it `isUpdating$`. What do you think?

feature-libs/checkout/base/components/checkout-payment-method/checkout-payment-method.component.html
```diff
@@ -28,7 +29,10 @@ <h2 class="cx-checkout-title d-none d-lg-block d-xl-block">
           class="cx-payment-card col-md-12 col-lg-6"
           *ngFor="let card of cards; let i = index"
         >
-          <div class="cx-payment-card-inner">
+          <div
+            class="cx-payment-card-inner"
+            (click)="selectPaymentMethod(card.paymentMethod)"
```
znikola: Nice. Previously we had to click on the actual text, right? And now it's possible to click on the whole card?

znikola (changes requested):

feature-libs/checkout/base/components/checkout-delivery-mode/checkout-delivery-mode.component.html
bgambocjaviniar: I think it's fine like they're not moving to the next step, but going back. What do you think?

I say it's fine because from above and if they want to get through the next step, the data of said step will be completed / done / finalized being going to the next step

feature-libs/checkout/b2b/components/checkout-delivery-address/checkout-delivery-address.component.ts
bgambocjaviniar: on hold ( slack )

feature-libs/checkout/base/components/checkout-payment-method/checkout-payment-method.component.html
bgambocjaviniar: yup

projects/assets/src/translations/en/payment.ts
bgambocjaviniar: if we're going through with this https://github.com/SAP/spartacus/pull/14828/files#r790410745 (notably slack)

projects/assets/src/translations/en/payment.ts
```diff
@@ -14,7 +14,7 @@ export const payment = {
...
     billingAddressSameAsShipping:
```
znikola, line 18: Adding a comment here because GH doesn't allow me to add it to the right place:
We should also rename the CMS component name in the sample data `CheckoutShippingAddress` to `CheckoutDeliveryAddress`. This means we will have to update the mapping in both base and b2b.
Maybe you can add a TODO:#checkout?

feature-libs/checkout/base/components/checkout-delivery-mode/checkout-delivery-mode.component.html
znikola: Not sure. What if  the user clicks back while there's an operation running, and in the previous component we get the result of the running operation (which no longer relevant / valid)?
But this is no longer valid, as we are hiding buttons when the spinner is shown.

projects/assets/src/translations/en/payment.ts
bgambocjaviniar: I was thinking from the accelerator SS that I showed you. Most things are delivery-address, but labels are Same as shipping address. Should we do it like that?

feature-libs/checkout/b2b/components/checkout-delivery-address/checkout-delivery-address.component.ts
bgambocjaviniar: keeping except changing the path references /checkout/shipping-address until we update sample data

feature-libs/checkout/b2b/components/checkout-delivery-address/checkout-delivery-address.component.html
```diff
@@ -1,15 +1,25 @@
...
+
+<ng-container *ngIf="state$ | async as state">
+  <ng-container
+    *ngIf="!state.shouldRedirect && !state.isUpdating; else loading"
+  >
     <ng-container
-      *ngIf="cards?.length && !addressFormOpened; else newAddressForm"
+      *ngIf="
+        isAccountPayment || (state.cards?.length && !addressFormOpened);
+        then hasExistingAddresses;
```
bgambocjaviniar: 👍  only way that I can think of with the global state above to have the button 'outside' the spinner

bgambocjaviniar: idea was to flatten all and just use state. place in other places, but I needed the back and continue button out when selecting the cards for both delivery address and payment method

projects/assets/src/translations/en/payment.ts
bgambocjaviniar: as mentioned earlier today since we won't do it like accelerator where path shows delivery-address, but label shows shipping-address, I changed all references for those labels and value to have Delivery instead of Shipping.

In the e2e, I did not update the methods and variables because I find it redundant fixing that 'now' haha, but I did change those references/variables in our checkout lib (except old checkout since we're dropping that)

feature-libs/checkout/b2b/components/checkout-delivery-address/checkout-delivery-address.component.html
bgambocjaviniar: no need anymore since we're putting the buttons in the spinner too now

feature-libs/checkout/base/components/checkout-delivery-mode/checkout-delivery-mode.component.html
```diff
@@ -1,17 +1,17 @@
...
+      !(isSetDeliveryModeBusy$ | async);
+    else loading
+  "
+>
+  <div [formGroup]="mode">
+    <div class="row">
+      <div class="col-md-12 col-lg-9">
         <div
           class="form-check"
           *ngFor="let mode of supportedDeliveryModes$ | async"
```
bgambocjaviniar, line 17: reminder

feature-libs/checkout/base/components/checkout-payment-method/checkout-payment-method.component.html
```diff
@@ -1,75 +1,82 @@
...
-  <ng-template #loading>
-    <div class="cx-spinner"><cx-spinner></cx-spinner></div>
+  <ng-template #newPaymentForm>
+    <cx-payment-form
+      (setPaymentDetails)="setPaymentDetails($event)"
+      (closeForm)="hideNewPaymentForm()"
+      (goBack)="back()"
+      [paymentMethodsCount]="cards?.length || 0"
+      [setAsDefaultField]="!isGuestCheckout"
+      [loading]="busy$ | async"
```
bgambocjaviniar: reminder

feature-libs/checkout/base/components/checkout-delivery-mode/checkout-delivery-mode.component.html
znikola: For what? :)

feature-libs/checkout/base/components/checkout-payment-method/checkout-payment-form/checkout-payment-form.component.html
```diff
@@ -197,16 +197,17 @@
...
               else billingAddress
             "
           >
             <cx-card
-              [content]="getAddressCardContent(shippingAddress)"
+              [content]="getAddressCardContent(deliveryAddress)"
             ></cx-card>
           </ng-container>
 
+          <!-- TODO:#checkout do we really want this? We can always pass more inputs to the copied address form component to make it more modular -->
```
znikola, line 210: Copy pasted from somewhere?

bgambocjaviniar: nono it's to keep and do later on since it's the copy pasted address form , but with limited inputs

If you recall, 

I wanted to re-use the address-form, and pass an input to it to hide the respective things we don't want to see in payment method

projects/assets/src/translations/en/payment.ts
bgambocjaviniar: yup yup I left the todo's already in places that when we update sample data, we change the paths but that also means to change the flex cms component

Issue #10251, opened 2020-12-07
Make spartacus work with both b2c / b2b at the same time


znikola:
We should spike how much effort there is to make Spartacus work with both b2c and b2b environments at the same time.

The first starting point would be to check if the APIs are actually allowing us to achieve this.

If this pre-condition is satisfied, we should then look into our modules and check how we could re-structure them in order to be able to run Spartacus against both b2c and b2b environments, at the same time.

There are a lot other things to consider and check as part of this spike, which are not yet obvious until we actually try to do it.

Some other things to consider:
- leverage the `channel` from the base site. Use this to dynamically call the appropriate endpoints (e.g. use different adapters to call either /orgUser API or regular /user APIs.
- checkout - can we have two different checkouts for b2c and b2b?
- configuration - are there any overlaps between the current b2c / b2b configs? If so, would one override the other?

Platonn:
May be related  to https://github.com/SAP/spartacus/issues/9724

Other angles to consider:
- installation script (`run.sh`) - drop the variable `ADD_B2B_LIBS`, because ideally we shouldn't t decide in the compilation time whether storefront is b2b or b2c

Platonn:
Firstly we need to do a prerequisite for `channel` from https://github.com/SAP/spartacus/issues/11040

Automaticaly provide b2b configs when channel is b2b: Add a config initializer that populates the configuration with `defaultB2bOccConfig` and `defaultB2bCheckoutConfig`, when the `channel` is B2B. 

It needs a spike: There might occur some problems (but don't have to) in places where we consume the checkout config or occ config statically (via usual `CONFIG` token), but not via methods `ConfigInitializerService.getStable('backend.occ')` or `ConfigInitializerService.getStable('checkout')`.

PR #14837, opened 2022-01-21
fix: add B2B config in epd-visualization schematic

sam-garland:


Platonn:
Hi @sam-garland 
If I understood correctly, in the fresh app with installed Spartacus, when you visit the B2B basesite, the following OCC endpoint is used for adding to cart: `<base-site>/users/anonymous/carts/`, and you get the error response `It's not allowed to execute this call from the current channel`.
And you discovered that using /orgUsers/ instead of /users/ (`<base-site>/orgUsers/anonymous/carts/`) fixes the problem.

Actually, the same issue happens in the fresh app installed via schematics generated from the `develop` branch, when you visit the B2B storefront (powertools-spa). The problem is not in your integration library, so it would be good to avoid implementing the workaround for this problem specifically for your library, if possible. 

### The reason of the problem

> If you are using schematics to set up your Spartacus project for the first time, there are important considerations to be aware of. For example, Spartacus does not support B2C and B2B storefronts **running together in a single storefront application.**

(source: https://sap.github.io/spartacus-docs/schematics/#adding-spartacus-core-libraries-and-features-to-your-angular-project)

In details:
The fresh app by default uses the feature "automatic multi-site configuration". This allows for dynamic recognition of the basesite by the URL - so you can run multiple storefronts in a single Spartacus app. But as of today we don't support yet the dynamic recognition of the basesite's channel (b2b vs b2c) see issue #10251, therefore the OCC configuration of endpoints is not automatically updated on-the-fly to be B2C- or B2B-specific. So you need to configure b2b/b2c endpoints ahead of time, before compiling the app.

If EPD visualisations can work both with B2B and B2C basesite, IMHO we shouldn't via schematics configure statically the B2B OCC endpoints in the name of customer in their app. 

@sam-garland **Do you think you could test your feature in the fresh app using just a B2C basesite (e.g. based on electronics)?**

If you want to test using B2B (powertools based) basesite, you would need to add manually in the fresh app the b2b config, because Spartacus doesn't have at the moment any explicit global flag in the schematics (like `--b2b`) to add it conviniently in the automated way.
Btw. the reason there is no such flag is we have plans to recognize it automatically in the runtime (in #10251)

sam-garland:
@Platonn - OK, I will instead make a change to the documentation to describe this.

PR #14839, opened 2022-01-21
GH-14833: Replace isFacadeImplProvided with CreateCartEvent

### LINKED ISSUE #14833: Remove isFacadeImplProvided
Remove isFacadeImplProvided and implement alternative logic in mini-cart

plabadie:


feature-libs/cart/main/root/components/mini-cart/mini-cart-component.service.ts
```diff
@@ -103,6 +105,14 @@ export class MiniCartComponentService {
     );
   }
 
+  protected isCartCreated(): Observable<boolean> {
+    return this.eventService.get(CreateCartEvent).pipe(
+      map((_) => true),
+      take(1),
```
WeizhengSap: do we need this `take(1), startWith(false)` if we have `      takeWhile((hasCart) => !hasCart, true)`?

plabadie: The intent was that, as soon as we have one cart created, we have the info we want and we don't need to listen for more.

...but, I agree that the takeWhile fills that role a little further.  I'll remove the take(1)

WeizhengSap (approved):

PR #14859, opened 2022-01-24
Handle add to cart errors

### LINKED ISSUE #14762: handle add to cart error
If the "add to cart" has error, we need a better way to handle it. For example, for anonymous user, add one product in cart, then remove the cart from backoffice. Add this product again. Since the active cart is gone, so the "addEntry" is failed. We can see the error message of "cart not found", but the modal is empty with spinner running.
Screen Shot 2022-01-12 at 12 04 55 PM

WeizhengSap:
close #14762

### DIFF
--- a/feature-libs/cart/main/components/cart/added-to-cart-dialog/added-to-cart-dialog-event.listener.spec.ts
+++ b/feature-libs/cart/main/components/cart/added-to-cart-dialog/added-to-cart-dialog-event.listener.spec.ts
@@ -107,5 +107,14 @@ describe('AddToCartDialogEventListener', () => {
         mockFailEvent.error
       );
     });
+
+    it('Should do nothing if the active modal is not the cart dialog', () => {
+      spyOn(modalService, 'getActiveModal').and.returnValue(null);
+      spyOn(mockInstance, 'dismissModal').and.stub();
+      listener['closeModal'](mockFailEvent);
+      expect(mockInstance.dismissModal).not.toHaveBeenCalledWith(
+        mockFailEvent.error
+      );
+    });
   });
 });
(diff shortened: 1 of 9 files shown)


projects/assets/src/translations/en/common.ts
```diff
@@ -133,6 +133,8 @@ export const common = {
     },
     cartNotFound: 'Cart not found.',
     invalidCodeProvided: 'Invalid code provided.',
+    unknownIdentifier: 'Item not found',
```
plabadie: Should the message end with a dot? (most of them seem to have one)

plabadie: Item as in "cart item" or as any kind of item not found?

WeizhengSap: any kind of item.

feature-libs/cart/main/core/http-interceptors/handlers/bad-cart-request.handler.ts
```diff
@@ -44,6 +45,22 @@ export class BadCartRequestHandler extends HttpErrorHandler {
       });
   }
 
+  protected handleOtherCartErrors(
+    _request: HttpRequest<any>,
+    response: HttpErrorResponse
+  ): void {
+    this.getErrors(response)
+      .filter((e) => e.reason !== 'notFound' || e.subjectType !== 'cart')
```
plabadie, line 53: instead of `e.reason !== 'notFound' || e.subjectType !== 'cart')`, could we use ` !isCartNotFoundError(e)`

feature-libs/cart/main/components/cart/added-to-cart-dialog/added-to-cart-dialog-event.listener.spec.ts
```diff
@@ -75,4 +98,14 @@ describe('AddToCartDialogEventListener', () => {
       );
     });
   });
+
+  describe('closeModal', () => {
+    it('Should close the add to cart dialog', () => {
```
plabadie, line 103: perhaps we can also test what happens when the active modal ref is not equal to the listener modal ref.  It might help the unit test branch coverge %

feature-libs/cart/main/core/http-interceptors/handlers/bad-cart-request.handler.ts
WeizhengSap: `isCartNotFoundError(e)` also check `!isSelectiveCart(error.subject)`

plabadie (approved):

PR #14876, opened 2022-01-26
refactor: Create entry points for special components

### LINKED ISSUE #14814: Create entry points for special components
These components:
- mini-cart
- add-too-cart
- add-to-wish-list

now live in `cart/main/root` and `cart/wishlist/root`
We want to see if we can improve the design by  moving them to a separate entry point and evaluate if we can provide the option to lazy load them.

* Required: whatever the solution, it has to play well with the schematics installer, the one that installs spartacus in an angular app and the one that installs just the cart lib and wish list lib ( they might be the same under the hood though  )

WeizhengSap:
close #14814


projects/storefrontapp/src/app/spartacus/features/cart-feature.module.ts
```diff
@@ -15,6 +15,12 @@ import { I18nConfig, provideConfig } from '@spartacus/core';
           module: () =>
             import('@spartacus/cart/main').then((m) => m.CartModule),
         },
+        /*[MINI_CART]: {
+          module: () =>
+            import('@spartacus/cart/main/components/cart/mini-cart').then(
+              (m) => m.MiniCartModule
+            ),
+        },*/
```
plabadie: dead code

feature-libs/cart/main/root/cart-root.module.ts
```diff
@@ -24,6 +27,20 @@ export function defaultCartComponentsConfig() {
...
+            (m) => m.MiniCartModule
+          ),
+      },
+      [ADD_TO_CART]: {
+        cmsComponents: ['ProductAddToCartComponent'],
+        module: () =>
+          import('@spartacus/cart/main/components/cart/add-to-cart').then(
+            (m) => m.AddToCartModule
+          ),
+      },
```
plabadie: Does this work?

I was under the impression that dynamic imports only woorked from the shell app.  But perhaps this works since rot is eager loaded.  Lets discuss.

feature-libs/cart/main/root/feature-name.ts
```diff
@@ -1,2 +1,5 @@
 export const CART_FEATURE = 'cart';
 export const CART_CORE_FEATURE = 'cartCore';
+
+export const MINI_CART = 'miniCart';
+export const ADD_TO_CART = 'addToCart';
```
plabadie: Should we add a "component" suffix?  I ddon't know.  Lets discuss.

feature-libs/cart/main/core/facade/active-cart.service.ts
```diff
@@ -61,7 +59,6 @@ export class ActiveCartService implements ActiveCartFacade, OnDestroy {
   );
 
   constructor(
-    protected store: Store<StateWithMultiCart>,
```
plabadie: `store` wasn't used?

WeizhengSap: not used

feature-libs/cart/main/root/cart-root.module.ts
WeizhengSap: it works

WeizhengSap: I moved it to shell app.

plabadie: In a way, I was hoping this would work.  If this needs to move to the shell app, we have to update the installation schematics to support it.

projects/storefrontapp/src/app/spartacus/features/cart-base-feature.module.ts
```diff
@@ -18,6 +20,18 @@ import { I18nConfig, provideConfig } from '@spartacus/core';
...
+            import('@spartacus/cart/base/components/mini-cart').then(
+              (m) => m.MiniCartModule
+            ),
+        },
+        [ADD_TO_CART]: {
+          module: () =>
+            import('@spartacus/cart/base/components/add-to-cart').then(
+              (m) => m.AddToCartModule
+            ),
+        },
```
plabadie, line 34: We can't have this here without the associated schematics support.  Did havinig these in the CartBaseRootModule work?  Let's discuss.

feature-libs/cart/base/root/feature-name.ts
```diff
@@ -1,2 +1,5 @@
 export const CART_BASE_FEATURE = 'cartBase';
 export const CART_BASE_CORE_FEATURE = 'cartBaseCore';
+
+export const MINI_CART = 'miniCart';
+export const ADD_TO_CART = 'addToCart';
```
plabadie: Not strongly againist having those here, but perhaps these should live with the corresponding feature.  On the other hand, this would create alot of ooverhead for a one component entry point, so maybe having those here is a good compromise.

projects/storefrontapp/src/app/spartacus/features/cart-base-feature.module.ts
```diff
@@ -18,6 +20,18 @@ import { I18nConfig, provideConfig } from '@spartacus/core';
           module: () =>
             import('@spartacus/cart/base').then((m) => m.CartBaseModule),
         },
+        [MINI_CART]: {
```
plabadie: Let's add the _FEATURE suffix to the 2 new feature constants to be consistent.

projects/storefrontapp/src/app/spartacus/features/wish-list-feature.module.ts
```diff
@@ -18,6 +19,12 @@ import { I18nConfig, provideConfig } from '@spartacus/core';
           module: () =>
             import('@spartacus/cart/wish-list').then((m) => m.WishListModule),
         },
+        [ADD_TO_WISHLIST]: {
```
plabadie: Add _FEATURE suffix to constant

feature-libs/cart/schematics/add-cart/index.ts
```diff
@@ -105,17 +111,31 @@ function addCartBaseFeature(options: SpartacusCartOptions): Rule {
...
+      },
+      {
+        name: MINI_CART_MODULE,
+        importPath: MINI_CART_ENTRY_POINT,
+      },
+      {
+        name: ADD_TO_CART_MODULE,
+        importPath: ADD_TO_CART_ENTRY_POINT,
+      },
+    ],
```
plabadie, line 130: It's missing the corresponding config for the wishlist.

feature-libs/cart/main/root/feature-name.ts
plabadie: We add a _FEATURE suffix to the const name

projects/storefrontapp/src/app/spartacus/features/cart-base-feature.module.ts
plabadie: Associated installation schematics are now added.

PR #14880, opened 2022-01-27
fix: Add aria attributes to tab paragraph container [REQUIRES WINDOWS]

### LINKED ISSUE #11637: Collapsible elements need aria-controls and aria-expanded [work on PR #14880/Windows]
Product tabs in Product Details Page and Product facets in Product Listing Page need aria attributes such as aria-controls and aria-expanded.

Probably a service and directive will be helpful to track state of aria-expanded

Zeyber:
PDP Part of #11637


developpeurweb:
**Testing outcome:**

- No violations were found with the AMP test.
- **JAWS** - overrides custom pattern and suggests navigation with up/down arrows, they work, but then when using tab it goes to next tab and not the tab panel.
- **NVDA/VO** - reads expanded/collapsed state but then when trying to go to associated tab panel, the user needs to keep tabbing to reach it, without any indication it has been reached.

**Some thoughts:**

In this case, it would be `aria-selected` instead of `aria-expanded`, but still, the pattern doesn't make sense for SRs as it is now. Because it get's "expanded" or "selected" but then when tabbing, the next tab button it's in the order path, and there is no way for the user to know when it has landed on the tab panel corresponding to the button.

So the suggestion here is to use an **Accordion** to keep a consistent focus sequence, instead of a Tab group.

developpeurweb:
Unfortunately, the PDP remains unchanged regarding SR compatibility, because the source order doesn't help. 

We currently have this:

Screen Shot 2022-02-01 at 4 17 02 PM

**It would be great** if we could have it as per screenshots below. With divs having tabindex so they can receive focus `<div tabindex="0">` so users won't skip to the following button and omit the panel's content.

Screen Shot 2022-02-01 at 4 20 56 PM

Screen Shot 2022-02-01 at 4 22 47 PM

Zeyber:
Unassigning myself so a dev who can use JAWS and NVIDIA (ie. Windows machine) can take over this PR. Currently seems to work fine on Mac with Chrome Vox and Voice Over Screen Readers but there are issues only apparent the windows readers.

FollowTheFlo:
Note for reviewer: 
breakpointService has been removed as it was decided to apply accordion on Pdp to all screen sizes. 
Tabs and Accordion Mode can be switched using css var $useAccordionOnly.

developpeurweb (changes requested):
The accordion toggle works now cross-SR/browser. However, there are a couple of new minor issues:
1. When focusing on accordion buttons, all SRs read the panel's content as opposed to only the button label, except for the Specs button (based on PLP, maybe try using `aria-label` to remove the bloat). Likewise, since this is a temporary solution and not the final component, pls try removing `aria-owns` if this makes the bug go away, it's not needed in this case.


2. The icon + and - also gets vocalized. Adding `role="presentation"` to it seems to fix it.
Screen Shot 2022-03-02 at 12 32 53 PM

developpeurweb (dismissed):
This works well for JAWS*/NVDA/VO
*Works well on JAWS+Firefox. Bugs with JAWS+Chrome (it will be addressed on a separate bug ticket if persistent).

docs/migration/5_0.md
```diff
@@ -416,12 +417,6 @@ The following keys have been added to `configurator-common.ts`:
 - key `common.navigation.shopAll` has been removed.
 - added new keys: `common.navigation.categoryNavLabel`, `common.navigation.footerNavLabel`
 
-### TabParagraphContainerComponent
```
pawelfras, line 419: `could:` consider to add an information that `BreakpointService` has been removed from component.

docs/migration/css-changes-in-version-5.md
```diff
@@ -43,6 +43,7 @@ title: Changes to Styles in 5.0
 
 - wrapped everything in `%pdpTabs`inside a `> div {...}` to restore styling after template changes
 - wrapped everything in `%pdpFlat`inside a `> div {...}` to restore styling after template changes
+- `span.accordion-icon` has been added for screen reader improvments.
```
pawelfras, line 46: `query:` maybe it's worth to mention about new var `$useAccordionOnly`?

projects/storefrontlib/cms-components/content/tab-paragraph-container/tab-paragraph-container.component.html
```diff
@@ -5,20 +5,23 @@
...
+        <button
+          [class.active]="i === activeTabNum"
+          (click)="select(i, $event)"
+          [attr.aria-expanded]="i === activeTabNum"
+        >
           {{
             component.title | cxTranslate: { param: tabTitleParams[i] | async }
           }}
+
+          <span class="accordion-icon" aria-hidden="”true”"></span>
```
Matejk00: Wondering if this double quote is required: `aria-hidden="”true”"`
I think we should try with `aria-hidden="true"`

projects/storefrontstyles/scss/components/content/tab-paragraph-container/_tab-paragraph-container.scss
```diff
@@ -1,4 +1,51 @@
...
+    padding-bottom: 0;
+    padding-inline-start: 4%;
+    width: 100%;
+    text-align: start;
+    height: 63px;
+    > span.accordion-icon {
+      &:before {
+        content: '+';
+        margin: 0px 15px;
+        float: right;
```
morganm58: The float will cause problems with directionality. We can remove the float and position this.

projects/storefrontstyles/scss/components/content/tab-paragraph-container/_tab-paragraph-container.scss
```diff
@@ -1,4 +1,51 @@
...
+    border-top: 1px var(--cx-color-light) solid;
+    font-size: 18px;
+    padding-top: 15px;
+    padding-inline-end: 0;
+    padding-bottom: 0;
+    padding-inline-start: 4%;
+    width: 100%;
+    text-align: start;
+    height: 63px;
+    > span.accordion-icon {
```
morganm58, line 16: add `position: absolute;` and `inset-inline-end: 0;` to span.accordion

projects/storefrontstyles/scss/components/content/tab-paragraph-container/_tab-paragraph-container.scss
```diff
@@ -1,4 +1,51 @@
...
+  > button {
+    border-top: 1px var(--cx-color-light) solid;
+    font-size: 18px;
+    padding-top: 15px;
+    padding-inline-end: 0;
+    padding-bottom: 0;
+    padding-inline-start: 4%;
+    width: 100%;
+    text-align: start;
+    height: 63px;
```
morganm58, line 14: Add 'position: relative;' after line 14

projects/storefrontstyles/scss/components/content/tab-paragraph-container/_tab-paragraph-container.scss
```diff
@@ -1,4 +1,51 @@
 $useTabsOnPDP: false !default;
+$useAccordionOnly: false !default;
+
+@mixin accordion {
+  > button {
+    border-top: 1px var(--cx-color-light) solid;
+    font-size: 18px;
```
morganm58: font-size: 18px should be change to a variable. `font-size: var(--cx-font-size, 1.125rem);`

projects/storefrontstyles/scss/components/content/tab-paragraph-container/_tab-paragraph-container.scss
```diff
@@ -1,4 +1,51 @@
...
+    padding-inline-start: 4%;
+    width: 100%;
+    text-align: start;
+    height: 63px;
+    > span.accordion-icon {
+      &:before {
+        content: '+';
+        margin: 0px 15px;
+        float: right;
+        font-size: 33px;
```
morganm58: font-size: 33px should be change to a variable. `font-size: var(--cx-font-size, 2.063rem);`

morganm58 (dismissed):

projects/storefrontstyles/scss/components/content/tab-paragraph-container/_tab-paragraph-container.scss
FollowTheFlo: it slightly modified vertical align of +/- icons, which is fixed by modifying 'bottom' property. 
Tested with resizing/zooming screen, all ok.

projects/storefrontlib/cms-components/content/tab-paragraph-container/tab-paragraph-container.component.html
FollowTheFlo: thanks, looks like prettier did not catch it.

FollowTheFlo:
@developpeurweb the committed fix from today didn't modify the a11y behavior. tested on windows nvda/jaws.
I suggest to create a new ticket specific to Chrome/JAWS as you mentioned, so we can go ahead with tickets depending of this one.

developpeurweb (approved):
CSS approved.
Still works for JAWS/NVDA/VO.

pawelfras (approved):

Matejk00 (approved):

projects/storefrontlib/cms-components/content/tab-paragraph-container/tab-paragraph-container.component.ts
```diff
@@ -34,8 +32,7 @@ export class TabParagraphContainerComponent implements AfterViewInit, OnInit {
   constructor(
     public componentData: CmsComponentData<CMSTabParagraphContainer>,
     protected cmsService: CmsService,
-    protected winRef: WindowRef,
```
Zeyber, line 37: Missing migrations for this constructor change

Issue #13536, opened 2021-08-23
E2E Tests for b2b saved cart


KateChuen:
Looking at the feature of b2b saved cart feature, I think we could add these e2e tests to increase the test coverage.

Test scenarios proposed:

-> Updating quantity in a saved cart
     - Have a saved cart with a few items, for eg, 3 different items
     - Go to the saved cart
     - Update the quantity of one of the items
     - Remove one item from the saved cart 
     - Make the cart active
     - Proceed to checkout

-> Clone a saved cart twice
    - Have a cart
    - Save the cart, with a name, for eg nameA.
    - Make the saved cart active and clone it (so now the cloned cart is named Copy of nameA)
    - Go to cart, and save the cart again, with the same name, nameA
    - Make the saved cart active and clone it (so the cloned cart is also named Copy of nameA)
    - Verify that the ID is different and unique for the cloned carts with same name. 

-> Make saved cart active through saved cart details page
    - Have a couple of saved carts.
    - Select one saved cart to go to its saved cart details.
    - Make the saved cart active from there

KateChuen:
Moved to https://jira.tools.sap/browse/CXSPA-252

PR #14536, opened 2021-12-06
Feature/gh 13536 saved cart e2e improvements

dydome:
close #13536

PR #14882, opened 2022-01-27
add: Added proceed to checkout steps from saved cart (#13536)

dydome:


projects/storefrontapp-e2e-cypress/cypress/helpers/b2b/b2b-checkout.ts
```diff
@@ -138,10 +139,12 @@ export function selectAccountShippingAddress() {
   );
 
   // Accessibility
-  verifyTabbingOrder(
-    'cx-page-layout.MultiStepCheckoutSummaryPageTemplate',
-    config.shippingAddressAccount
-  );
+  if (getViewport() === 'desktop') {
```
bgambocjaviniar: please remove as I don't think we're supposed to do this. Did someone say mention we're going to move forward to only test in desktop?

what was the use case if ever someone said to start making it like this?

projects/storefrontapp-e2e-cypress/cypress/helpers/b2b/b2b-checkout.ts
```diff
@@ -303,6 +308,7 @@ export function placeOrder(orderUrl: string) {
     'getOrderConfirmationPage'
   );
 
+  cy.get('cx-place-order input').click();
```
bgambocjaviniar: this breaks the e2e 🤔 

it unchecks the checkbox, therefore you're not able to place an order

bgambocjaviniar (changes requested):
see comments

bgambocjaviniar (changes requested):
2 things that you need to do:

1 - either bring this back to the other PR or bring the other changes here https://github.com/SAP/spartacus/pull/14536 (I don't see a point on having 2 separate 'saved-cart' e2e PR)
- Be careful because from my observation here, there were 'not useful' improvements that broke the e2es https://github.com/SAP/spartacus/pull/14536#issuecomment-1031024329

2 - Related to the DM, where you mentioned minicart breaking the e2e sometimes. See screenshot. There's finally an example on which you were talking about mini cart being 0 vs 1. I would say it's either a bug that was introduced pre-cart in terms of a cart service or such as the e2e was deterministic. Maybe it has something to do with the user being 're-used' when it's not supposed to because according to the video, it shows the minicart having an item when it's not supposed to, so my guess would be the user is somehow being re-used OR a cart bug was introduced as again the e2e test that was created is deterministic. https://dashboard.cypress.io/projects/k3nmep/runs/27915/test-results?actions=%5B%5D&browsers=%5B%5D&groups=%5B%5D&isFlaky=%5B%5D&modificationDateRange=%7B%22startDate%22%3A%221970-01-01%22%2C%22endDate%22%3A%222038-01-19%22%7D&orderBy=EXECUTION_ORDER&oses=%5B%5D&specs=%5B%5D&statuses=%5B%7B%22value%22%3A%22FAILED%22%2C%22label%22%3A%22FAILED%22%7D%5D&testingTypesEnum=%5B%5D&utm_source=github.

All in all, something to do. for 2-. Try to reproduce the use case, if it fails to reproduce, try to make sure each users is distinct because I'm thinking somehow the user is being re-used now. But I would look at the reproducing first.

Screen Shot 2022-02-06 at 10 56 28 PM

PR #14911, opened 2022-01-31
fix: Add aria attributes to facets

### LINKED ISSUE #11637: Collapsible elements need aria-controls and aria-expanded [work on PR #14880/Windows]

Zeyber:
PLP Part of #11637


developpeurweb:
**Test outcome for PLP:**
- **JAWS** - Brakes pattern. Reads whole content of the panel, then button expanded state at the end, ENTER key gets overridden and will not work to enter into the panel,  it only works for expanding button. 
- **NVDA** -  Forces accessible pattern, so it's mostly fine.
- **VO** - Follows implemented custom pattern, which never lands on the button, but on the first item.

**Some thoughts/options**

- Removing the double-enter and double-escape pattern.
- Aux key events (up/down arrow to jump groups, TAB to go one-by-one)

**Also:**
- Let's try IDs without spaces and associating the buttons.
- Do we need IDs? if we remove aria-controls and use aria-label, will it sound the same?

Screen Shot 2022-01-31 at 5 51 48 PM
Screen Shot 2022-01-31 at 5 55 34 PM
Screen Shot 2022-01-31 at 5 58 55 PM

developpeurweb:
- We still have some issues with `aria-labelledby` linking to IDs.
- Is it possible to remove the double-enter and double-escape pattern? 
It would be a better cross-screen-reader experience if the user just follows the source order, item-by-item when they are available once a given panel is open. The facet "group skipping" has different behaviour across SRs.

Screen Shot 2022-02-01 at 3 43 51 PM
Screen Shot 2022-02-01 at 3 44 20 PM

developpeurweb (dismissed):
This works well on JAWS/NVDA/VO

docs/migration/5_0.md
```diff
@@ -359,6 +359,11 @@ The following keys have been added to `configurator-common.ts`:
 
 - `store-finder-list-item.component.html` markup structure changed. `h2.cx-store-name > button` changed to `a.cx-store-name` for screen reader improvements.
 
+### FacetListComponent
```
Matejk00: It is doubled section for `### FacetListComponent` in 5_0 document. We can add this info to existing one under "Template Changes".

Matejk00 (changes requested):

Matejk00 (approved):

developpeurweb (approved):
This works well with JAWS/NVDA/VO

PR #14919, opened 2022-01-31
fix: add checkout step to title (#13539)

### LINKED ISSUE #13539: Page title for checkout doesn't change along with checkout steps

BenjaminEmiliani:
Closes #13539


projects/storefrontlib/cms-components/navigation/page-header/page-title.component.ts
```diff
@@ -26,9 +28,14 @@ export class PageTitleComponent implements OnInit {
   }
 
   private setTitle(): void {
-    this.title$ = this.pageMetaService.getMeta().pipe(
-      filter(isNotNullable),
-      map((meta) => (meta.heading || meta.title) ?? '')
+    this.title$ = combineLatest([
```
bgambocjaviniar: as mentioned, do not couple 'any checkout' logic to this 'generic' page title

feature-libs/checkout/base/core/services/checkout-page-meta.resolver.ts
```diff
@@ -46,14 +45,8 @@ export class CheckoutPageMetaResolver
...
    */
-  resolveTitle(): Observable<string> {
-    return this.cart$.pipe(
-      switchMap((c) =>
-        this.translationService.translate('pageMetaResolver.checkout.title', {
-          count: c.totalItems,
-        })
-      )
-    );
+  resolveTitle(): Observable<string | undefined> {
```
bgambocjaviniar, line 51: from the first hint provided in DM, you can use resolveHeading() to change the breadcrumb title and resolveTitle will change the actual title displayed in the header and web browser

bgambocjaviniar (changes requested):
^

projects/storefrontlib/cms-components/navigation/page-header/page-title.component.ts
bgambocjaviniar: put it back to what it was and use resolveHeading() in the checkout-page resolver --> I was hinting in our messages that this functionality existed from the hints I provided 😉 
```suggestion
     this.title$ = this.pageMetaService.getMeta().pipe(
      filter(isNotNullable),
      map((meta) => (meta.heading || meta.title) ?? '')
```

feature-libs/checkout/base/core/services/checkout-page-meta.resolver.ts
```diff
@@ -41,19 +40,22 @@ export class CheckoutPageMetaResolver
...
    * @override
-   * Resolves the page title from the translation `pageMetaResolver.checkout.title`. The
-   * cart total item `count` is passed to the translation, so it can be used in the title.
+   * Resolves the page title for the Checkout Page to include checkout step.
+   * The page title used by the browser (history, tabs) and crawlers.
    *
    * The title from the page data is ignored for this page title.
    */
-  resolveTitle(): Observable<string> {
-    return this.cart$.pipe(
```
bgambocjaviniar, line 50: we can safely remove cart$ from this file as it's no longer being used nor it was even being used for a lonnngggg time apparently haha

feature-libs/checkout/base/core/services/checkout-page-meta.resolver.spec.ts
```diff
@@ -58,14 +58,29 @@ describe('CheckoutPageMetaResolver', () => {
...
         result = meta;
       })
       .unsubscribe();
 
-    expect(result).toEqual('pageMetaResolver.checkout.title count:5');
+    expect(result).toEqual('Checkout Delivery Mode');
+  });
+
+  it('should resolve checkout heading', () => {
+    let result!: string;
```
bgambocjaviniar: instead of saying it will always be a string, let's put the type as  `string | undefined`, similarly to line 59

bgambocjaviniar (changes requested):

feature-libs/checkout/base/core/services/checkout-page-meta.resolver.ts
```diff
@@ -27,8 +26,6 @@ export class CheckoutPageMetaResolver
   extends PageMetaResolver
   implements PageTitleResolver, PageDescriptionResolver, PageRobotsResolver
```
bgambocjaviniar: I missed this, but for convention on interfaces that helps to 'include' the abstract methods to be implemented, you should implement PageHeadingResolver

feature-libs/checkout/base/core/services/checkout-page-meta.resolver.spec.ts
```diff
@@ -58,14 +58,29 @@ describe('CheckoutPageMetaResolver', () => {
   it(`should resolve page title`, () => {
     let result: string | undefined;
 
+    basePageMetaResolver.resolveTitle = createSpy().and.returnValue(
```
bgambocjaviniar, line 62: let's include resolveTitle method under the mock class in line 23

bgambocjaviniar (approved):

feature-libs/checkout/base/core/services/checkout-page-meta.resolver.ts
```diff
@@ -41,19 +43,22 @@ export class CheckoutPageMetaResolver
...
+    return this.basePageMetaResolver.resolveTitle();
+  }
+
+  /**
+   * Resolves the page heading for the Checkout Page.
+   * The page heading is used in the UI (`<h1>`), where as the page
+   * title is used by the browser and crawlers.
+   */
+  resolveHeading(): Observable<string> {
+    return this.translationService.translate('pageMetaResolver.checkout.title');
```
znikola, line 61: Question: aren't we "loosing" the information about the total items in the current cart (which we had before)?

---------/---------
/reply:
@znikola in terms of the logic yes, but not in terms of the application. What I mean by that is the method previously translated  'pageMetaResolver.checkout.title' and would pass totalItems of the cart as a parameter. But this key in our base application does not handle any parameters:
Screen Shot 2022-02-07 at 9 07 50 AM

And if a customer is using a custom translation key that expects the parameter, then it will be reported to them in the 5.0 md that the method no longer translates this key with a parameter

PR #14927, opened 2022-02-01
feat: new global message type to enable screen reader vocalization on demand (#14371)

### LINKED ISSUE #14371: Create a Service for Dynamic Vocalized Feedback (SR improvement)

BenjaminEmiliani:
Closes #14371


projects/storefrontlib/shared/components/assistive-technology-message/assistive-technology-message.directive.ts
```diff
@@ -0,0 +1,48 @@
+import {
+  Directive,
+  ElementRef,
+  HostListener,
+  Input,
+  Optional,
+  TemplateRef,
+} from '@angular/core';
+import { GlobalMessageService } from '../../../../core/src/global-message/facade/global-message.service';
+import { GlobalMessageType } from '../../../../core/src/global-message/models/global-message.model';
```
Pucek9: wrong import paths

projects/storefrontlib/shared/components/assistive-technology-message/assistive-technology-message.directive.ts
```diff
@@ -0,0 +1,48 @@
...
+
+@Directive({
+  selector: '[cxAtMessage]',
+})
+export class AtMessageDirective {
+  /**
+   * Usage [cxAtMEssage]="'translatableKey' | cxTranslate"
+   */
+  @Input() cxAtMessage: string;
```
WeizhengSap, line 18: maybe just input "translatableKey"? This key can be passed to `this.globalMessageService.add`.

projects/storefrontlib/shared/components/assistive-technology-message/assistive-technology-message.directive.ts
```diff
@@ -0,0 +1,48 @@
...
+   * Emit assistive global meesage to improve screen reader vocalization.
+   * @param event
+   */
+  @HostListener('click', ['$event'])
+  handleClick(event: MouseEvent): void {
+    event?.preventDefault();
+
+    if (event?.target === this.host) {
+      this.globalMessageService.add(
+        this.cxAtMessage,
```
WeizhengSap, line 42: if `this.cxAtMessage` is empty (`undefined`), we can get the message just from `innerText`.

projects/storefrontlib/shared/components/assistive-technology-message/assistive-technology-message.directive.ts
```diff
@@ -0,0 +1,48 @@
...
+    return !!this.templateRef
+      ? this.templateRef.elementRef.nativeElement.parentElement
+      : this.elementRef.nativeElement;
+  }
+
+  /**
+   * Emit assistive global meesage to improve screen reader vocalization.
+   * @param event
+   */
+  @HostListener('click', ['$event'])
```
WeizhengSap, line 36: If this is just POC, that's fine; otherwise we also need to listen to other events.

projects/storefrontlib/shared/components/assistive-technology-message/assistive-technology-message.directive.ts
BenjaminEmiliani: @WeizhengSap I am not too sure if this is needed for the reason that the entire purpose of this extension of global message is to make screen reader technologies vocalize a descriptive text that is not on the DOM. So the usage strictly requires passing a descriptive message. 

I do see how it would serve as a fallback but the innerText of the element is not the intended message. Thoughts?

WeizhengSap: OK, if `cxAtMessage` is empty, do nothing. (I thought we need do check of `cxAtMessage` to prevent empty key/text passed to globalMessage.add)

projects/storefrontlib/shared/components/assistive-technology-message/assistive-technology-message.directive.ts
BenjaminEmiliani: @WeizhengSap  Yes this is an option as well, but since we have to considers translatable keys that have parameters (I.e. 
Screen Shot 2022-02-02 at 9 47 51 AM

that means that the input of the directive would have to look like this:
Screen Shot 2022-02-02 at 9 48 39 AM

Whereas I feel that translating first in template and passing translated string is simpler and uniform with the rest of the application:
Screen Shot 2022-02-02 at 9 51 28 AM

projects/storefrontlib/shared/components/assistive-technology-message/assistive-technology-message.directive.ts
WeizhengSap: Please consider to add other listeners; at least `tab` key need be listened, right?

BenjaminEmiliani: @WeizhengSap I am not sure what other listeners are needed. since 'click' handles mouse click and 'enter' key. 

When I think of `tab` key, it means user is shifting focus from one element to the next. And when we do that, accessible labels and texts are automatically read by screen readers

WeizhengSap (dismissed):

WeizhengSap (approved):

projects/storefrontlib/cms-components/misc/global-message/global-message.component.html
```diff
@@ -1,4 +1,10 @@
 <div *ngIf="messages$ | async as messages">
+  <div
+    class="cx-visually-hidden"
+    *ngFor="let confMsg of messages[messageType.MSG_TYPE_ASSISTIVE]"
```
Pucek9, line 4: `confMsg`?

BenjaminEmiliani: Will change

PR #14942, opened 2022-02-03
feat: Cart page cms components refactor for screen readers and tabbing (#13590)

### LINKED ISSUE #13590: Coupon field is not discoverable by screen reader users before getting to Checkout button
**Describe the bug**

- The coupon field is after the checkout button, therefore, it might go unnoticed by a blind user if the SR doesn't vocalize it, before hitting the "Proceed" button.
- This is information the user definitely needs to hear, check with designers and devs to see if we can place this better, basically anywhere before the "Proceed to Checkout" button.

- Library version: 4.0.0

**To Reproduce**
2. Add product to cart.
3. Go to the cart page and **pretend you cannot see what is on the screen.**
4. Checkout products using the keyboard.
5. Listen how you miss the coupon field when reaching the **Proceed to Checkout** button.

**Expected behavior**

- Coupon field is discoverable by screen reader users before getting to the **Proceed to Checkout** button

**Screenshots**

dydome:
close #13590

Please test it with server: https://cx-platform-current.eastus.cloudapp.azure.com:9002/


feature-libs/cart/base/components/cart-proceed-to-checkout/cart-proceed-to-checkout.component.html
```diff
@@ -0,0 +1,12 @@
+<cx-progress-button
+  (clickEvent)="disableButtonWhileNavigation()"
+  [class]="'btn btn-primary btn-block'"
+  [disabled]="cartValidationInProgress"
+  [loading]="cartValidationInProgress"
+  [routerLink]="{ cxRoute: 'checkout' } | cxUrl"
+>
+  {{
+    (!cartValidationInProgress ? 'cartDetails.proceedToCheckout' : '')
```
bgambocjaviniar: cart-lib fixed this, where they added text, but that's an issue since text wrapped due to the length of it. https://github.com/SAP/spartacus/issues/14952

Currently causing the same issue as it was pre-cart-lib merge. To discuss with Patrick and Wei on a fix.

bgambocjaviniar: Basically, Patrick and Wei added text because when passing an empty string, there's an error in devtools

feature-libs/cart/base/components/cart-totals/cart-totals.component.ts
```diff
@@ -1,59 +1,18 @@
...
-  entries$: Observable<OrderEntry[]>;
-  cartValidationInProgress = false;
 
-  protected subscription = new Subscription();
-
-  constructor(
-    protected activeCartService: ActiveCartFacade,
-    protected router?: Router
-  ) {}
+  constructor(protected activeCartService: ActiveCartFacade) {}
```
bgambocjaviniar, line 13: conventions from now for all 'facades', let's name the variable for them activeCartFacade

```suggestion
  constructor(protected activeCartFacade: ActiveCartFacade) {}
```

bgambocjaviniar (changes requested):
After above fixes are applied, coordinate with GC on when to update the servers, so we can get the PR in, **BUT make sure you tell him checkout is soon to be ready to merge and it needs to update the servers too due to sample data updates too, so I don't know whether we want this PR until checkout is merged as well.**

feature-libs/cart/base/components/cart-totals/cart-totals.component.ts
dydome: we are mostly using this approach -> activeCartService: ActiveCartFacade, but also agree that in the repo there are 2 ways.

feature-libs/cart/base/components/cart-proceed-to-checkout/cart-proceed-to-checkout.component.html
dydome: I saw it, there is already Pr with Fix so I have updated this part in my PR also.

developpeurweb (dismissed):
Works well with JAWS/NVDA/VO and no violations found for AMP. That's awesome!

Screen Shot 2022-02-18 at 3 16 10 PM

bgambocjaviniar (dismissed):

szczepaniak-michal (approved):

PR #14954, opened 2022-02-04
chore: add plugin to feature configration [POC]

### LINKED ISSUE #15231: [POC] add plugin into feature configuration
Wrapper module can be programmatically created.

WeizhengSap:
close #14954


integration-libs/digital-payments/src/digital-payments.module.ts
```diff
@@ -1,6 +1,7 @@
-import { DpCheckoutModule } from './checkout/dp-checkout.module';
 import { NgModule } from '@angular/core';
+import { CheckoutModule } from '@spartacus/checkout';
+import { DpCheckoutModule } from './checkout/dp-checkout.module';
 @NgModule({
-  imports: [DpCheckoutModule],
+  imports: [CheckoutModule, DpCheckoutModule],
```
plabadie: Just a question:  Is this an example of what would be a "built-in" wrapper module?

I thought we had to create an extra module to be the wrapper.

WeizhengSap: This is can be a wrapper module, because `DigitalPaymentsModule` itself is a wrapper module wrapping `DpCheckoutModule`.  If we create another wrapper, then there will be a wrapper inside another wrapper.

integration-libs/digital-payments/src/checkout/dp-checkout.module.ts
```diff
@@ -28,6 +29,7 @@ import { DpPaymentMethodModule } from './cms-components/dp-payment-method/dp-pay
       useExisting: OccDpRequestNormalizer,
       multi: true,
     },
+    DpCheckoutPaymentService,
```
znikola: A question: what is the reason for providing the service in this module? it seems it's already provided in `root` and used withing the DP lib?

znikola (changes requested):

projects/storefrontapp/src/app/spartacus/features/digital-payments-feature.module.ts
```diff
@@ -1,14 +1,21 @@
...
   providers: [
     provideConfig(<I18nConfig>{
+      featureModules: {
+        [CHECKOUT_FEATURE]: {
+          module: () =>
+            import('@spartacus/digital-payments').then(
+              (m) => m.DigitalPaymentsModule
+            ),
+        },
+      },
```
plabadie, line 20: I believe there must be corresponding changes in the installation schematics to support the changes done to this file.

integration-libs/digital-payments/src/checkout/dp-checkout.module.ts
WeizhengSap: Without it, there will be `NullInjectorError: No provider for DigitalPaymentsAdapter!`.

WeizhengSap: Same thing for connector, e.g. `CartConnector` also has `providedIn: 'root'`, but if not put it in module's provider list, there will be "No provier error".

projects/storefrontapp/src/app/spartacus/features/b2b-digital-payments.module.ts
```diff
@@ -0,0 +1,8 @@
+import { NgModule } from '@angular/core';
+import { CheckoutScheduledReplenishmentModule } from '@spartacus/checkout/scheduled-replenishment';
+import { DigitalPaymentsModule } from '@spartacus/digital-payments';
+
+@NgModule({
+  imports: [DigitalPaymentsModule, CheckoutScheduledReplenishmentModule],
```
bgambocjaviniar:
```suggestion
  imports: [CheckoutScheduledReplenishmentModule, DigitalPaymentsModule],
```

projects/storefrontapp/src/app/spartacus/features/digital-payments-feature.module.ts
```diff
@@ -17,4 +16,40 @@ import {
     }),
   ],
 })
-export class DigitalPaymentsFeatureModule {}
+export class DigitalPaymentsFeatureModule {
+  static forB2C(): ModuleWithProviders<DigitalPaymentsFeatureModule> {
```
bgambocjaviniar: that's a cool way of doing things instead of creating separate 'feature-libs'

I'd like Nik's opinion on this because to me it's about consistency vs 'new' consistency (if we're going to go this route in the future)

(example: to me old approach is creating different set of DigitalPaymentsModule wrapper module in the library)

p.s I like this approach :)

projects/storefrontapp/src/app/spartacus/features/digital-payments-feature.module.ts
```diff
@@ -17,4 +16,40 @@ import {
...
+                  (m) => m.DigitalPaymentsModule
+                ),
+            },
+          },
+        }),
+      ],
+    };
+  }
+
+  static forB2B(): ModuleWithProviders<DigitalPaymentsFeatureModule> {
```
bgambocjaviniar: for b2b should point to b2b checkout

and another one for replenishment

`forScheduledReplenishment()` pointing to the replenishment/digitalpayment module

projects/storefrontapp/src/app/spartacus/features/b2b-digital-payments.module.ts
bgambocjaviniar: folder should be called scheduled-replenishment-digital-payments.module.ts

and you should create another one for the b2b module (which is without replenishment)

projects/storefrontapp/src/app/spartacus/features/digital-payments-feature.module.ts
bgambocjaviniar: it's just a different way of how we will write schematics in the end (old vs new)

bgambocjaviniar: another drawback is more code for customer if they decide to not use our schematics ootb

because if we put each of these wrapper module within our shellapp, it's more code to do in the customer side

whereas if we place the 'wrapper module' within the dp-lib, they can always use that instead of us putting this b2b dp module in out shellapp

bgambocjaviniar: ^

Screen Shot 2022-02-18 at 12 09 55 PM

WeizhengSap: I can move the wrapper modules inside lib. But, each wrapper module will need one entry point.

integration-libs/digital-payments/src/checkout/cms-components/dp-payment-method/dp-payment-method.component.ts
```diff
@@ -21,21 +21,15 @@ import { DP_CARD_REGISTRATION_STATUS } from '../../../utils/dp-constants';
...
     const queryParams = this.activatedRoute.snapshot.queryParamMap.get(
       DP_CARD_REGISTRATION_STATUS
     );
-    if (queryParams) {
-      return true;
-    } else {
-      return false;
-    }
+
+    return Boolean(queryParams);
```
znikola: Let's just use `!!queryParams`

projects/storefrontapp/src/app/spartacus/features/digital-payments-feature.module.ts
znikola: I like this approach, but one drawback is that the unused static methods are not tree-shaken.
It's not adding much of bytes to the bundle, but we'll have weigh in the trade-offs

integration-libs/digital-payments/scheduled-replenishment/replenishment-digital-payments.module.ts
```diff
@@ -0,0 +1,8 @@
+import { NgModule } from '@angular/core';
+import { CheckoutScheduledReplenishmentModule } from '@spartacus/checkout/scheduled-replenishment';
+import { DigitalPaymentsModule } from '@spartacus/digital-payments';
+
+@NgModule({
+  imports: [DigitalPaymentsModule, CheckoutScheduledReplenishmentModule],
+})
+export class ReplenishmentDigitalPaymentsModule {}
```
znikola: This is the price of no composability

znikola (changes requested):

PR #14971, opened 2022-02-08
Feature/gh 13916 jest refactor

psmul:


feature-libs/asm/jest.ts
```diff
@@ -22,3 +22,15 @@ Object.defineProperty(document.body.style, 'transform', {
   },
 });
 */
+
+import 'zone.js/fesm2015/zone-testing-bundle.min.js';
```
Platonn: 1. Do we need to include zone also for unit tests of schematics? I guess no.
2. Btw. why not just `import 'zone.js'`?

(similar for other libs)

feature-libs/asm/package.json
```diff
@@ -24,12 +24,14 @@
     "@angular/common": "^13.2.0",
     "@angular/core": "^13.2.0",
     "@angular/forms": "^13.2.0",
+    "@angular/platform-browser-dynamic": "^13.2.0",
```
Platonn: Why do we need this peerDep platform-browser-dynamic specifically in ASM package?

(similar for other libs)

feature-libs/asm/package.json
```diff
@@ -24,12 +24,14 @@
...
     "@ngrx/effects": "^13.0.2",
     "@ngrx/store": "^13.0.2",
     "@spartacus/core": "4.1.0-next.0",
     "@spartacus/schematics": "4.1.0-next.0",
     "@spartacus/storefront": "4.1.0-next.0",
-    "rxjs": "^6.6.0"
+    "rxjs": "^6.6.0",
+    "zone.js": "~0.11.4"
```
Platonn: Why do we need this peerDep Zone.js specifically in ASM package?

(similar for other libs)

feature-libs/product-configurator/jest.schematics.config.js
```diff
@@ -11,10 +12,9 @@ module.exports = {
       tsconfig: '<rootDir>/tsconfig.schematics.json',
     },
   },
-
   collectCoverage: false,
   coverageReporters: ['json', 'lcov', 'text', 'clover'],
-  coverageDirectory: '<rootDir>/../../coverage/storefinder/schematics',
+  coverageDirectory: '<rootDir>/../../coverage/product-configurator/schematics',
```
Platonn, line 17: good catch 👀

package.json
```diff
@@ -210,10 +210,11 @@
...
     "stylelint-scss": "^3.18.0",
     "through2": "^3.0.1",
-    "ts-jest": "^26.4.4",
+    "ts-jest": "^27.1.3",
     "ts-morph": "^9.1.0",
     "ts-node": "~8.3.0",
     "typescript": "^4.4.4",
-    "webpack-cli": "^3.3.12"
+    "webpack-cli": "^3.3.12",
+    "jest-preset-angular": "^11.1.0"
```
Platonn: Nitpick: Let's add it via yarn command: `yarn add -D jest-preset-angular` so it's placed in alphabetical order into package.json. Otherwise the future similar `yarn add` command might move your dependency alphabetically up in the list and cause unnecessary changes to the future PRs.

projects/core/package.json
```diff
@@ -23,8 +23,8 @@
     "@ngrx/router-store": "^13.0.2",
     "@ngrx/store": "^13.0.2",
     "angular-oauth2-oidc": "^13.0.1",
-    "i18next": "^21.6.6",
-    "i18next-http-backend": "^1.3.1",
+    "i18next": "^21.6.10",
+    "i18next-http-backend": "^1.3.2",
```
Platonn: I don't mind having those changes here.
Just double checking: is it related to fixing Jest?

integration-libs/cdc/jest.ts
```diff
@@ -0,0 +1,36 @@
...
+});
+Object.defineProperty(document.body.style, 'transform', {
+  value: () => {
+    return {
+      enumerable: true,
+      configurable: true,
+    };
+  },
+});
+*/
```
Platonn: 1. I wonder if/when we will transition all unit tests to jest. If we have plans to do it, let's create a proposal ticket for it and let's put the required steps (code snippets) there in the ticket instead of putting long dead code comments. Or preferably let's put into the ticket just a link to a guideline how to migrate to Jest (I guess you copy pasted those snippets from some guideline). 
1. Just curious: Why we added this comment specifically only for a few integration libs (e.g. cdc, cds, dp)? But not for other libs? 
3. btw. Those commands are opinionated workarounds, overwriting global window object's properties. I would like to avoid them at all, if possible. In the future, if/when we transition to Jest, I hope they won't be needed. 🤞  But I understand now it might be required if only we transitioned all to Jest today.

feature-libs/asm/jest.ts
psmul: Simple `import 'zone.js'` breaks unit tests. I assume this specific version is prepared to be working with NG13 somehow, but I didn't dive into the internals.

feature-libs/asm/package.json
psmul: Actually this is something that our `yarn config:update` did, nothing was done manually here. I assume it's due to the fact of using imports from `@angular/platform-browser-dynamic` in `jest.ts` file responsible for creating testing space for unit tests.

feature-libs/asm/package.json
psmul: Same as above. Probably our config script catches library usages in code (somehow?) and applies missing libs in depedencies.

integration-libs/cdc/jest.ts
psmul: 1. I think such ticket already exists.
2. Actually I must admit, it was copy&paste between different locations. I could have thought of it better, sorry.
3. Agreed, all comments will be gone from all these types of files. 👍🏼

package.json
psmul: Oh good catch and idea, thank you!

feature-libs/asm/package.json
Platonn: IMHO those packages (zone and platform browser) should not be listed as peer dependencies of all our libraries. 
Since we're not using those packages in the code of our libraries, it's not justified to list them as our peer deps. Please note this package.json will be shipped to NPM. Therefore customers installing our lib will be adviced by `yarn` or `npm` commands to install missing peer deps (or complain about incompatible versions).

We use those packages only in the setup of unit tests.

Let's fix our `config:update` script, to ignore also the new files `jest.ts` as well:  https://github.com/SAP/spartacus/blob/8afd1fdb58d4ac1028daedab317463ed8e163d66/tools/config/manage-dependencies.ts#L460-L462

btw. what about renaming those files from `jest.ts` to `test-jest.ts`, so it's obvious that this file is related to bootstraping the tests?

feature-libs/asm/jest.ts
Platonn: I can see the docs of the jest-angular-preset [suggest importing from`/fesm2015/`](https://thymikee.github.io/jest-preset-angular/docs/guides/angular-13+/#using-es-modules). I just want to double check if its really needed.

1. Did you try importing both zone.js and zone.js/testing, just as we do in files `test.ts`, e.g.:
https://github.com/SAP/spartacus/blob/ebcb4c9e582d385ab817bb384a07abb6715ece48/projects/core/src/test.ts#L3-L5

2. Could you attach here the error message when using those imports, please?

integration-libs/cdc/jest.ts
Platonn: ad 1. What do you think about moving out those commented out code snippets to the ticket description?

feature-libs/asm/jest.ts
psmul: Actually importing both lines works well. I get green and finished results. I assume we want to go this way? (2 imports as mentioned above)

feature-libs/asm/package.json
psmul: I like the idea with re-naming, it's always good to stick to convention. (in this case with karma test.ts files, although it would suggest renaming test to karma-test 🧐 )

integration-libs/cdc/jest.ts
psmul: 1. Done 👍🏼  #15004

feature-libs/asm/jest.ts
Platonn: If you confirm it helps, I'm for it.

PR #14974, opened 2022-02-08
feat: improve screen reader support for storefinder

### LINKED ISSUE #13740: Find a Store: On the Store Details page, there is no vocalized feedback of context change when landing

Matejk00:
Closes: #13740


developpeurweb (dismissed):
Approving since page changes are detected and being announced in JAWS/NVDA/VO. There is header/footer vocalization bloat, which is also present on other pages, so the latter will be addressed in a different ticket.

szczepaniak-michal (approved):

PR #14991, opened 2022-02-10
fix: Missing focus management and narration on PDP when 'Show Reviews' is clicked

### LINKED ISSUE #14030: [Not fixed in #11637] Missing focus management and narration on PDP when 'Show Reviews' is clicked
Address after #11637 

**Describe the bug**
For a11y users, there is not focus shift or vocalized feedback when users select 'Show reviews' link on a product display page.

- 4.1

**To Reproduce**
1. Click on any product with reviews
2. Click on 'Show reviews'
4. Observe current focus and lack of context feedback

**Expected behavior**
Focus shifts to 'Reviews' tab. 
vocalized feedback that user is in the Reviews section

Pucek9:
closes #14030

### DIFF
--- a/projects/storefrontlib/cms-components/content/tab-paragraph-container/tab-paragraph-container.component.ts
+++ b/projects/storefrontlib/cms-components/content/tab-paragraph-container/tab-paragraph-container.component.ts
@@ -83,6 +83,14 @@ export class TabParagraphContainerComponent implements AfterViewInit, OnInit {
         } else {
           this.activeTabNum = tabNum;
         }
+        setTimeout(() => {
+          const firstFocusableElement = this.winRef.document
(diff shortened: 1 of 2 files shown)


FollowTheFlo (approved):
Tested successfully, focus goes on 'Write a review' button when Show reviews is pressed.

developpeurweb (changes requested):
As detailed on the ticket, the expectation is to move the focus to the "Reviews" panel, as the triggering button reads **"Show"**. Currently, it moves focus to the  _"Write a Review"_ button. Which doesn't "show" the reviews.
However, since we're moving from Tabs to Accordions, this needs to be retested when **11637** is merged.

Screen Shot 2022-03-07 at 5 22 50 PM

PR #15042, opened 2022-02-16
Added to Cart Toast Unit Tests

### LINKED ISSUE #13349: Added to Cart Toast Unit tests

bit172:
closes #13349


feature-libs/cart/base/components/added-to-cart-toast/added-to-cart-toast.component.spec.ts
```diff
@@ -0,0 +1,242 @@
...
+    component.ngOnDestroy();
+  });
+
+  it('should create', () => {
+    expect(component).toBeTruthy();
+  });
+
+  it('should call the active cart service', () => {
+    const spy = spyOn(activeCartService, 'getLastEntry').and.callThrough();
+    component.ngOnInit();
```
szczepaniak-michal: why do we need ngOnInit() on every test case here? Could we achieve the same goal with beforeEach?

feature-libs/cart/base/components/added-to-cart-toast/added-to-cart-toast.component.spec.ts
```diff
@@ -0,0 +1,242 @@
...
+      const res = component.isHeaderInView();
+      expect(res).not.toBeTruthy();
+    });
+  });
+
+  describe('Toast Positions', () => {
+    describe('Toast Opening', () => {
+      it('should be fixed when the header is fixed', () => {
+        spyOn(component, 'isHeaderFixed').and.returnValue(true);
+        let positionClasses = component.getToastStyles(
```
szczepaniak-michal: we could declare `positionClasses` only once in the describe block and only do the assignments in test cases

feature-libs/cart/base/components/added-to-cart-toast/added-to-cart-toast.component.ts
```diff
@@ -54,7 +54,7 @@ export class AddedToCartToastComponent implements OnInit, OnDestroy {
     this.headerElement = this.winRef.document.querySelector('header');
 
     if (!this.customClass) {
-      this.customClass = 'cx-added-cart-toast';
+      this.customClass = 'cx-added-to-cart-toast';
```
szczepaniak-michal, line 57: Some inconsistency in naming, in selectors I see `added-to-cart` and i.e. in constants `ADD_TO_CART`. Let's decide on one and apply it on all cases

bit172: I think I'll be using `ADDED_TO_CART` to refer to components that are used when added to the cart (like an effect from adding to cart). Add to cart should refer to actions that do the adding.

szczepaniak-michal (approved):

Issue #9445, opened 2020-10-21
E2E tests for auth changes


marlass:
We need to write e2e for the new functionality in auth module.

Platonn:
It's been quite a time since creating this ticket and its description is a bit vague. So let me deduce its details:
Deducing from the context (time of creation and visibility in the board https://github.com/SAP/spartacus/projects/78), it refers to the enhanced auth capabilities introduced in Spartacus 2.0, see https://sap.github.io/spartacus-docs/session-management, e.g. the support of other auth flows - Implicit Flow or Authorization Code Flow.

Currently our e2e cover only the Resource Owner Password Flow (which is a default). But this flow is not recommended due to security reasons and customers should the Authorization Code Flow (most recommended) or Implicit Flow. But our e2e don't cover those 2 flows yet. 

If possible, I'd suggest to start from testing the most recommended - Authorization Code Flow (or if not possible, then Implicit Flow). In this flow, the user is redirected to a different domain to fill a login form (to the auth server page), and after successfull login, the user is redirected back to Spartacus to the previously visited route (or to the login-protected route which the user attempted to visit).

For more, let's watch the knowledge transfer recordings from @marlass.

Note: to use implicit flow you might need first to enable it in backoffice:
![image](https://user-images.githubusercontent.com/4001059/158995444-c4043115-b7b0-4a9f-a58a-3695d9f3108f.png)
 


### Plan of attack:
 - [ ] [INFRA] We need first to check if our test server exposes the external login page. If not, we need to setup our test servers. 
   - Done: The URL to the auth page is `/authorizationserver/login.jsp`, e.g. https://40.76.109.9:9002/authorizationserver/login.jsp . The alternative alias is `/authorizationserver/oauth/authorize` - see our default auth config https://github.com/SAP/spartacus/blob/e5b5a373490c3462a218188bc69e5d1d84b92ec5/projects/core/src/auth/user-auth/config/default-auth-config.ts#L9
 - [ ] Read the docs of Spartacus [configuring-authorization-code-flow-or-implicit-flow](https://sap.github.io/spartacus-docs/session-management/#configuring-authorization-code-flow-or-implicit-flow) and learn how to configure Spartacus with OAuth flow (Authorization Code flow or Implicit flow) with external login page. 
[...]

PR #15179, opened 2022-02-18
fix: Fix redirection when return from auth server using code flow

### LINKED ISSUE #14694: Code Auth Flow causes incorrect redirection when opening new tabs

Zeyber:
Closes: #14694


projects/core/src/auth/user-auth/facade/auth.service.ts
```diff
@@ -43,13 +43,15 @@ export class AuthService {
...
-      const tokenReceived = getLastValueSync(
-        this.oAuthLibWrapperService.events$.pipe(
-          filter((e) => e.type === 'token_received')
-        )
-      );
+    let tokenReceived: OAuthEvent | undefined;
+    const sub: Subscription = this.oAuthLibWrapperService.events$
+      .pipe(
+        take(1),
+        filter((e: OAuthEvent) => e.type === 'token_received')
```
Platonn: As far as I understand: In case when other unrelated event is emitted first from events$ (due to any reason), we stop listening immediately and miss the event `token_received` that is emitted later. 

To avoid this issue, I suggest to filter() first, and only later take(1). 
```suggestion
        filter((e: OAuthEvent) => e.type === 'token_received')
        take(1),
```

projects/core/src/auth/user-auth/facade/auth.service.ts
```diff
@@ -43,13 +43,15 @@ export class AuthService {
    * Check params in url and if there is an code/token then try to login with those.
    */
   async checkOAuthParamsInUrl(): Promise<void> {
-    try {
-      const tokenReceived = getLastValueSync(
-        this.oAuthLibWrapperService.events$.pipe(
-          filter((e) => e.type === 'token_received')
-        )
-      );
+    let tokenReceived: OAuthEvent | undefined;
```
Platonn, line 46: What do you think about adding a code comment explaining why we need this variable?
```ts
/**
  * The event 'token_received' is emitted, when `oAuthLibWrapperService.tryLogin()`
  * can obtain the token from the URL params (which means we've just returned from 
  * an external login page to Spartacus).
  * 
  * Please note that this event is not emitted when `oAuthLibWrapperService.tryLogin()`
  * can obtain the token from the storage (e.g. on refresh of the Spartacus page).
  */
let tokenReceived: OAuthEvent | undefined;
```

projects/core/src/auth/user-auth/facade/auth.service.ts
```diff
@@ -43,13 +43,15 @@ export class AuthService {
...
+    const sub: Subscription = this.oAuthLibWrapperService.events$
+      .pipe(
+        take(1),
+        filter((e: OAuthEvent) => e.type === 'token_received')
+      )
+      .subscribe((e: OAuthEvent) => (tokenReceived = e));
 
+    try {
       const result = await this.oAuthLibWrapperService.tryLogin();
```
Platonn, line 66: What do you think about adding a code comment explaining the connection between `tryLogin()` and the event `token_received`?  IMHO the connection is not obvious. It is known only from the source code of the OAuthLib.
```ts

// The method `oAuthLibWrapperService.tryLogin()` obtains the token either from the URL params
// or from the storage. To distinguish those 2 cases, we observe the event `token_received`.
//
// The event 'token_received' is emitted, when the method `oAuthLibWrapperService.tryLogin()`
// can obtain the token from the URL params (which means we've just returned from 
// an external authorization page to Spartacus).
// 
// But the event 'token_received' is not emitted when the method `oAuthLibWrapperService.tryLogin()`
// can retrieve the token from the storage (e.g. on refresh of the Spartacus page).
const result = await this.oAuthLibWrapperService.tryLogin();
```

--
Side note:
Ideally, I'd love to encapsulate this implementation detail of the OAuthLib **inside** our `this.oAuthLibWrapperService.tryLogin();`, so the `AuthService` facade is more clean, readable and testable .
This would require changing the signature of `oAuthLibWrapperService.tryLogin()` from `Promise<boolean>` to `Promise<OAuthTryLoginResult>`. (note: the name of the interface is up to discussion)
```ts
interface OAuthTryLoginResult {
   result: boolean,

   /** 
     * It's true only when received the token from URL params.
     * Note: it's false in particular, when the token was retrieved from the storage.
     */
   tokenReceived: boolean,
} 
```

This would be a breaking change.
What do you think about merging the current fix just as is to the latest 4.x, but add the improvements in 5.0 (move the complexity of subscriptions to inside tryLogin()) + describe breaking changes in the migration docs?

Platonn (dismissed):

hackergil (changes requested):
Is there a reason why this PR is not (unit) tested?

Zeyber:
> Is there a reason why this PR is not (unit) tested?

Units tests already exist for the provided functions. What this PR fixes requires e2e testing will be provided here: https://github.com/SAP/spartacus/issues/9445

Platonn (approved):

hackergil (approved):

PR #15211, opened 2022-02-22
feat: Vocalize loading spinner status

### LINKED ISSUE #13698: Percentage of a page loaded not being vocalized
**Describe the bug**
- Currently, there's no hint as to what percentage of the page has been loaded, so a Screen Reader user wouldn't know when a page has loaded fully.
- This is particularly helpful for users with slow internet connections.
- This could be related to Lazy Loading.

- Library version: 4.0.0.

**Expected behavior**
- Screen Reader vocalizes the percentage of the page that has loaded.
- "Loading" messages get vocalized using `aria-live="polite" aria-atomic="true"`

**Screenshots**

Some examples of how this feature is expected to be vocalized:

Screen Shot 2020-10-07 at 6 17 29 PM

Screen Shot 2020-10-07 at 6 05 41 PM

Matejk00:
Closes: #13698


FollowTheFlo (dismissed):
From my tests. The fix does what is described, when  spinner appears, 'loading...' is vocalized
image
image

FollowTheFlo:
I tried 5 different solutions: add hidden div directly, wrap hidden div in dummy cpnt, globalMessageService, Directive that manipulate DOM, manipulate DOM in ngOnDestroy of Spinner cpnt.
I kept the most simple 'add hidden div directly' as it does not need any import and no DOM manipulation needed that could create race conditions, better performance.
Adding 
**<div role="status" [attr.aria-label]="'common.loaded' | cxTranslate"></div>** 
in each Cpnts html using  cx-spinner

developpeurweb (dismissed):
This doesn't work on JAWS, but maybe because the narration bloat takes over the "loaded" narration.
Also `role="status"` doesn't work on JAWS unless it's preloaded.
I'm approving it, since it works on NVDA/VO and this is a minor feature. After the bloat is gone, if loading silence persists, a new ticket will be issued.

szczepaniak-michal (approved):

PR #15224, opened 2022-02-23
feat: Schematics - order feature modules

### LINKED ISSUE #14959: Schematics - order of feature modules
```
In SpartacusFeaturesModule, ...featureModules, is the last in provider list. So, configuration of[CHECKOUT_FEATURE]: {...} in digital-payment will replace the one in checkout feature.
```

The above comment is from Wei's from this PR: https://github.com/SAP/spartacus/pull/14954
It got me thinking about our current installation schematics. As we didn't have a reason to think about the order of feature before, we are simply _not_ ensuring this.
Let's take a look at the following use cases:

### Installing feature using the main prompt

When we run `ng add @spartacus/schematics` we will be prompted to select which features to install.

The order of installation is not deterministic, and it depends on multiple factors:
- the order in which it appears in the prompt list
- the order in which they are selected - as we depend on the Angular passing us the list of selected features, which might be changed if the user selects a feature, then deselects, and then selects again.
- some internal dependency discovery that which we have (think of the case where customers select a feature which a has a dependency on another Spartacus features, which isn't selected).

Because DP has to be listed after Checkout, we need to ensure this by ordering the imports of features in spartacus-features module.
Currently, this is _not_ the case, and can be seen in the following examples:
- `ng add @spartacus/schematics`
- select _only_ DP. Checkout is not selected, but it will be installed as it's a dependency of DP.
- when the schematics finishes, in the Spartacus features module, the order will the following: `DigitalPaymentsFeatureModule,CheckoutFeatureModule,UserFeatureModule`.

We can see the order is wrong - the `DigitalPaymentsFeatureModule` should be listed _after_ `CheckoutFeatureModule`.

### Installing a feature using its own schematics

Let's say the app already exists with some features installed (but not the `Checkout`), and the user runs `ng add @spartacus/digital-payments`.
It's hard to imagine Spartacus without Checkout, but technically it's possible. The result of the installation will be exactly the same as the one above.

## How to solve the problem

We could try to determine the order of the feature module imports by running the following algorithm:

- when we have the list of selected features
  - detect their dependencies
  - store them in a certain data structure
- once we have the list of all spartacus features which should be installed
  - determine which features is the most dependent on. We could do this by e.g. counting how many times appears in the data structure from above
  - as the last step in the installation process, we could re-arrange the list of features in the spartacus features module according to the order 

Since we want to make sure the order is preserved, it would be good to add a comment above the import whose order is critical.
E.g.
```ts
imports: [
...
CheckoutFeatureModule,
// should be imported after CheckoutFeatureModule
DigitalPaymentsFeatureModule,
...
]
```

## Conclusion

To me, this is something that we want to address, and invest time, because it can save us a lot of time with potential mis-configuration issues.

znikola:
closes #14959


feature-libs/tracking/schematics/add-tracking/index.ts
```diff
@@ -64,15 +64,15 @@ function addGtm(options: SpartacusTrackingOptions): Rule {
...
       rootModule: {
-        importPath: SPARTACUS_TMS_CORE,
         name: TMS_BASE_MODULE,
+        importPath: SPARTACUS_TMS_CORE,
         content: `${TMS_BASE_MODULE}.forRoot()`,
       },
-      featureModule: {
-        importPath: SPARTACUS_TMS_GTM,
-        name: TMS_GTM_MODULE,
-      },
```
znikola, line 75: Just re-arranged, no changes

feature-libs/tracking/schematics/add-tracking/index.ts
```diff
@@ -99,15 +99,15 @@ function addAep(options: SpartacusTrackingOptions): Rule {
...
       rootModule: {
-        importPath: SPARTACUS_TMS_CORE,
         name: TMS_BASE_MODULE,
+        importPath: SPARTACUS_TMS_CORE,
         content: `${TMS_BASE_MODULE}.forRoot()`,
       },
-      featureModule: {
-        importPath: SPARTACUS_TMS_AEP,
-        name: TMS_AEP_MODULE,
-      },
```
znikola, line 110: Same as above - re-arranging

integration-libs/digital-payments/schematics/add-digital-payments/index.ts
```diff
@@ -50,8 +50,8 @@ function addDigitalPayments(options: SpartacusDigitalPaymentsOptions): Rule {
       folderName: DIGITAL_PAYMENTS_FOLDER_NAME,
       moduleName: DIGITAL_PAYMENTS_MODULE_NAME,
       featureModule: {
-        importPath: SPARTACUS_DIGITAL_PAYMENTS,
         name: DIGITAL_PAYMENTS_MODULE,
+        importPath: SPARTACUS_DIGITAL_PAYMENTS,
```
znikola: Re-arranged

projects/schematics/src/add-spartacus/spartacus-features.ts
```diff
@@ -41,30 +41,26 @@ function configureSpartacusModules(
         .getFilePath()
         .includes(`${SPARTACUS_FEATURES_MODULE}.module.ts`)
     ) {
-      [
-        `// Auth Core
```
znikola, line 45: We are removing these comments, as it's not easy to handle them in the customers app - when ordering feature modules, they are removed)

projects/schematics/src/migrations/mechanism/dependency-management/dependency-management.ts
```diff
@@ -83,8 +81,8 @@ function collectSpartacusLibraryDependencies(packageJson: any): {
 }
 
 function collectSpartacusPeerDeps(
-  collectedDeps: string[],
-  name: string
+  name: string,
+  collectedDeps: string[]
```
znikola, line 85: Re-arranged the params, causing the changes across the file

projects/schematics/src/shared/utils/package-utils.ts
```diff
@@ -116,7 +105,7 @@ export function readPackageJson(tree: Tree): any {
 
 export function cleanSemverVersion(versionString: string): string {
   if (isNaN(Number(versionString.charAt(0)))) {
-    return versionString.substr(1, versionString.length - 1);
+    return versionString.substring(1, versionString.length);
```
bgambocjaviniar, line 108: 👍 

substr going to be deprecated one day

projects/schematics/src/shared/feature-libs-constants.ts
```diff
@@ -0,0 +1,257 @@
...
+
+export const SPARTACUS_CORE = '@spartacus/core';
+export const SPARTACUS_STOREFRONTLIB = '@spartacus/storefront';
+export const SPARTACUS_ASSETS = '@spartacus/assets';
+export const SPARTACUS_STYLES = '@spartacus/styles';
+
+export const SPARTACUS_SETUP = '@spartacus/setup';
+export const SPARTACUS_SETUP_SSR = `${SPARTACUS_SETUP}/ssr`;
+
+export const CORE_SPARTACUS_SCOPES = [
```
bgambocjaviniar: I understand we moved these things to it's own constants as the constants file we were leveraging was getting big anyhow. I do not have a problem with this approach, but maybe we can change the filename to encapsulate the overall contents as I believe, some of the scopes aren't part of a 'feature', but more of the 'core' part of Spartacus. 

To be honest, I can't think of a filename that can encapsulates the contents in here. But if we move line [7-26](https://github.com/SAP/spartacus/pull/15224/files#diff-8a2edcc5bf618cefbd40578f754bdad00ee3dd9726e149aa0c9d79b8fa12b6afR7-R26) back to constants then the current filename makes sense. Same for line [109-115](https://github.com/SAP/spartacus/pull/15224/files#diff-8a2edcc5bf618cefbd40578f754bdad00ee3dd9726e149aa0c9d79b8fa12b6afR109-R115)

bgambocjaviniar: But please disregard this message if so as it is more semantics what I saying than a problem :P

projects/schematics/src/shared/updateable-constants.ts
```diff
@@ -0,0 +1,245 @@
...
+  [SPARTACUS_ORGANIZATION]: [
+    CLI_ORGANIZATION_ADMINISTRATION_FEATURE,
+    CLI_ORGANIZATION_ORDER_APPROVAL_FEATURE,
+  ],
+  [SPARTACUS_PRODUCT]: [
+    CLI_PRODUCT_BULK_PRICING_FEATURE,
+    CLI_PRODUCT_VARIANTS_FEATURE,
+    CLI_PRODUCT_IMAGE_ZOOM_FEATURE,
+  ],
+  [SPARTACUS_PRODUCT_CONFIGURATOR]: [
```
bgambocjaviniar, line 139: never noticed, but shouldn't product configurator be an integration-lib?

projects/schematics/src/shared/utils/graph-utils.ts
```diff
@@ -0,0 +1,76 @@
+export class Graph {
```
bgambocjaviniar, line 1: **not important at all**, but is it worth to add a print graph function for anyone who wants to see current graph 'outlook' when testing

projects/schematics/src/shared/utils/graph-utils_spec.ts
```diff
@@ -0,0 +1,112 @@
...
+      const result = kahnsAlgorithm(graph);
+      expect(result).toEqual([
+        SPARTACUS_PRODUCT,
+        SPARTACUS_QUALTRICS,
+        SPARTACUS_SMARTEDIT,
+        SPARTACUS_STOREFINDER,
+      ]);
+    });
+
+    it('should be able to detect a cyclic dependency', () => {
```
bgambocjaviniar, line 97: awesome

projects/storefrontapp/src/app/spartacus/spartacus-features.module.ts
```diff
@@ -182,21 +182,28 @@ if (environment.epdVisualization) {
 
     /************************* Feature libraries *************************/
     UserFeatureModule,
+
     CartBaseFeatureModule,
     WishListFeatureModule,
+    SavedCartFeatureModule,
+    QuickOrderFeatureModule,
+    ImportExportFeatureModule,
+
```
bgambocjaviniar, line 191: the extra break lines are for separation I'm assuming for some libs right

projects/storefrontlib/cms-components/myaccount/my-interests/my-interests.module.ts
```diff
@@ -1,18 +1,19 @@
...
-import { SpinnerModule } from '../../../shared/components/spinner/spinner.module';
 import {
   AuthGuard,
   CmsConfig,
   I18nModule,
   provideDefaultConfig,
   UrlModule,
 } from '@spartacus/core';
-import { CmsPageGuard, PageLayoutComponent } from '../../../cms-structure';
+import { CmsPageGuard } from '../../../cms-structure/guards/cms-page.guard';
```
bgambocjaviniar, line 11: 👍

bgambocjaviniar (changes requested):
tests in DM

projects/schematics/src/shared/utils/graph-utils.ts
bgambocjaviniar: actually maybe not only for unit test to verify. 

Maybe we can add a --verbose flag, which enables logs to be shown. So when we do ng add stuff in customer's app, we can verify the directed graph

projects/schematics/src/shared/updateable-constants.ts
znikola: It might belong there, but not sure if we had the concept of "integration libs" before the configurator features was introduced 🤔
We can always move it, as it's not a breaking change

projects/storefrontapp/src/app/spartacus/spartacus-features.module.ts
znikola: Yes, it's supposed to separate imports from each of the libs

projects/schematics/src/shared/feature-libs-constants.ts
znikola: Right. Maybe just "lib-constants"?

projects/schematics/src/shared/updateable-constants.ts
bgambocjaviniar: think cds was the first integration lib, so maybe configurator was misplaced

projects/storefrontapp/src/app/spartacus/spartacus-features.module.ts
bgambocjaviniar: yup I see

so my follow-up to this was, wouldn't we separate these too

```
    AsmFeatureModule,
    StorefinderFeatureModule,
    QualtricsFeatureModule,
    SmartEditFeatureModule,
```

projects/schematics/src/shared/utils/graph-utils.ts
znikola: Done

projects/schematics/src/shared/utils/config-utils.ts
```diff
@@ -36,11 +36,29 @@ export function getSpartacusProviders(
...
+}
+
 const EMPTY_SPACE_REG_EXP = /\s+/gm;
-export function normalizeConfiguration(config: string | Node): string {
+const COMMA_REG_EXP = /,+/gm;
+function normalizeConfiguration(config: string | Node): string {
   let newConfig = typeof config === 'string' ? config : config.getText();
 
   newConfig = newConfig.trim();
+  newConfig = newConfig.replace(COMMA_REG_EXP, '');
```
bgambocjaviniar, line 61: 👍

projects/schematics/src/shared/utils/new-module-utils.ts
```diff
@@ -189,20 +187,11 @@ function isDuplication(
...
-    return false;
+  if (propertyName !== 'providers') {
+    return isTypeTokenDuplicate(initializer, content);
   }
 
-  return isTypeTokenDuplicate(initializer, content);
+  return isSpartacusConfigDuplicate(content, initializer);
 }
 
 function isTypeTokenDuplicate(
```
bgambocjaviniar, line 197: not needed, but if you want to increase the coverage, then maybe test the duplication for imports, exports, and declarations

projects/schematics/src/shared/utils/workspace-utils_spec.ts
```diff
@@ -14,7 +14,7 @@ import {
 import { Schema as WorkspaceOptions } from '@schematics/angular/workspace/schema';
 import * as path from 'path';
 import { Schema as SpartacusOptions } from '../../add-spartacus/schema';
-import { SPARTACUS_CORE } from '../../shared/feature-libs-constants';
+import { SPARTACUS_CORE } from '../libs-constants';
```
bgambocjaviniar, line 17: coolio how'd you spot that (wasn't an issue, but cool haha. I just noticed this was the only difference from the rest, where it was just renamed, but makes sense for this since it's only 1 level up)

projects/schematics/src/shared/utils/lib-utils.ts
```diff
@@ -69,6 +69,10 @@ export interface LibraryOptions extends Partial<ExecutionOptions> {
   project: string;
   lazy?: boolean;
   features?: string[];
+  /**
+   * When enabled, prints the additional logs.
+   */
+  debug?: boolean;
```
bgambocjaviniar, line 75: I'm all in for debug instead of log from what I said without looking at your PR before :d

projects/schematics/src/shared/utils/lib-utils.ts
```diff
@@ -1047,3 +1026,95 @@ export function runExternalSpartacusLibrary(
...
+      if (collectedModules.warnings.length) {
+        context.logger.warn(
+          'The following modules were not recognized due to various reasons:'
+        );
+        for (const warning of collectedModules.warnings) {
+          context.logger.warn(warning);
+        }
+      }
+
+      const orderedModules: string[] = spartacusCoreModules
```
bgambocjaviniar, line 1110: coolio, I see how it's done now :d 

Took a closer look today under the hood 👍

projects/schematics/src/shared/utils/lib-utils.ts
```diff
@@ -1047,3 +1026,95 @@ export function runExternalSpartacusLibrary(
...
+): Rule {
+  return (tree: Tree, context: SchematicContext): void => {
+    context.logger.info('Ordering the installed Spartacus features...');
+
+    const basePath = process.cwd();
+    const { buildPaths } = getProjectTsConfigPaths(tree, options.project);
+    for (const tsconfigPath of buildPaths) {
+      const { appSourceFiles } = createProgram(tree, basePath, tsconfigPath);
+
+      let spartacusFeaturesModule: SourceFile | undefined;
```
bgambocjaviniar: Maybe we can re-purpose `spartacusFeatureModuleExists` function to `getSpartacusFeatureModule`, where it returns the source else return undefined. 

As I see, it is being used in line 300 for the `checkAppStructure` to verify  if it is true. It won't affect the condition since it returns an object instead of undefined. Therefore, we can just re-use it here instead I think

projects/schematics/src/shared/utils/lib-utils.ts
```diff
@@ -306,6 +280,8 @@ export function addLibraryFeature<T extends LibraryOptions>(
       config.dependencyManagement
         ? installRequiredSpartacusFeatures(config.dependencyManagement, options)
         : noop(),
+
+      orderInstalledFeatures(options),
```
bgambocjaviniar, line 284: Not sure if it makes sense to do it now, but instead of ordering at every spartacus lib installation, maybe we can just do it at the end.

Can't find where, but I'm assuming there's a tree of things, where we know how many features are going to be installed beforehand, so we maybe only run after the last added installation to slightly improve performance, and to not order and save the spartacus-features.module every time too

projects/schematics/src/shared/utils/lib-utils.ts
```diff
@@ -1047,3 +1026,95 @@ export function runExternalSpartacusLibrary(
...
+          sourceFile
+            .getFilePath()
+            .includes(`${SPARTACUS_FEATURES_MODULE}.module.ts`)
+        ) {
+          spartacusFeaturesModule = sourceFile;
+          break;
+        }
+      }
+
+      if (!spartacusFeaturesModule) {
```
bgambocjaviniar, line 1071: why do we continue if there's no spartacus features module found

I'm assuming it's to keep the schematics running right, and not break right?

bgambocjaviniar (dismissed):

projects/schematics/src/shared/utils/new-module-utils.ts
znikola: It's being tested implicitly when we compare the snapshots runs

projects/schematics/src/shared/utils/workspace-utils_spec.ts
znikola: I think VSCode did it when I renamed the file :)

projects/schematics/src/shared/utils/lib-utils.ts
znikola: Good catch! Done.

projects/schematics/src/shared/utils/lib-utils.ts
znikola: We continue because we are iterating over the `tsconfigPaths`, and the module might be in some other project.

projects/schematics/src/shared/utils/lib-utils.ts
znikola: I wanted to optimize it, but I wasn't able to find out how to determine when the schematics will finish the execution.
We have a complex dependency management, and we add schematics installation task for each of the dependencies that need to be installed. Those dependencies might require another  set of dependencies, etc., and it's very difficult to foresee the last step.

At the end of the day, we are at least safe.

bgambocjaviniar (approved):

PR #15226, opened 2022-02-23
feat: Add vocalization for Order Confirmation (#13587)

### LINKED ISSUE #13587: [Needs assistive message feature] Order Confirmation needs vocalized feedback
**Describe the bug**

- There is no verbal confirmation alert that a transaction has happened. Before actually landing on the confirmation page.
- Once on the Confirmation page, there's no feedback, (this is mainly because the Page title is not being read).

- Library version: 4.0.0.

**To Reproduce**
2. Add products to cart.
3. Checkout products.
4. Hear absence of feedback.

**Expected behavior**
- Right after hitting the **Place Order** button, Screen Reader vocalizes: _"Thank you for your order! An invoice has been sent by email. Redirecting to the Confirmation page."_
- When landing on the Confirmation page, Screen Reader vocalizes: _"Order Confirmation, Confirmation of Order 00000012"_
- Use: `aria-live="polite/assertive"` and `aria-atomic="true"`.

**Screenshots**
Screen Shot 2021-08-31 at 3 04 54 PM

pawelfras:
closes #13587


feature-libs/checkout/base/components/checkout-order-confirmation/checkout-order-confirmation-thank-you-message/checkout-order-confirmation-thank-you-message.component.html
```diff
@@ -1,5 +1,12 @@
 <ng-container *ngIf="order$ | async as order">
-  <div class="cx-page-header">
+  <div
+    class="cx-page-header"
+    [attr.aria-label]="
+      'checkoutOrderConfirmation.orderConfirmation' | cxTranslate
+    "
+    aria-atomic="true"
+    aria-live="polite"
+  >
```
BenjaminEmiliani: I'm almost certain this will not work on Windows using NVDA or JAWS, it only works on Voice Over. Issue here is ` aria-live="polite"` region is inside a container that is not pre-rendered (`ng-container *ngIf="order$`).

feature-libs/checkout/base/assets/translations/en/checkout.i18n.ts
```diff
@@ -44,5 +44,7 @@ export const checkout = {
     createAccount: 'Create an account?',
     createAccountForNext:
       'Create an account for <{{email}}> for a faster checkout on your next visit.',
+    redirectingToConfirmationPage: 'Redirecting to the Confirmation Page.',
+    orderConfirmation: 'Order Confirmation',
```
BenjaminEmiliani: I don't think we needs these new keys. The `redirectingToConfirmationPage` is not necessary in my opinion but maybe confirm with @developpeurweb. And the `orderConfirmation` key is very similar to `'checkoutOrderConfirmation.confirmationOfOrder'`.

feature-libs/checkout/base/components/checkout-place-order/checkout-place-order.component.html
```diff
@@ -24,9 +20,14 @@
...
   <button
-    type="submit"
+    (click)="submitForm()"
     class="btn btn-primary btn-block"
     [disabled]="termsAndConditionInvalid"
+    [cxAtMessage]="[
+      'checkoutOrderConfirmation.thankYou' | cxTranslate,
+      'checkoutOrderConfirmation.invoiceHasBeenSentByEmail' | cxTranslate,
+      'checkoutOrderConfirmation.redirectingToConfirmationPage' | cxTranslate
+    ]"
```
BenjaminEmiliani: So during testing I realized something I did not think about before. These messages will be narrated to the user even if there is a failure placing an order (server error, or connection error). So I think we need to rethink the approach. I think maybe instead of using the directive, we can emit an ASSISTIVE global message in the component `checkout-order-confirmation-thank-you-message` inside the Angular lifecycle hook: `ngAfterViewInit()`.

pawelfras: Maybe part of vocalisation could be done like that, but then user will stay without any information after click on "Place Order". It has to be discussed with @developpeurweb.

feature-libs/checkout/base/assets/translations/en/checkout.i18n.ts
pawelfras: Yeah, to be discussed with @developpeurweb.

dydome (dismissed):

developpeurweb (dismissed):
Works well with JAWS/NVDA/VO

dydome (dismissed):

dydome (approved):

PR #15230, opened 2022-02-23
chore: Removing unused incubator library

hackergil:
Incubator library won't be published anymore since we don't use it.


angular.json
```diff
@@ -349,43 +349,6 @@
         }
```
znikola, line 349: Can you add some useful information to the PR description?
When merging to develop, this should automatically be copied by GH, and used in the release notes

angular.json
```diff
@@ -349,43 +349,6 @@
...
       }
```
znikola, line 350: We should add one line to our deprecation docs and mention about the removal of incubator lib

znikola (changes requested):

angular.json
hackergil: Done

angular.json
hackergil: Done

znikola (approved):

PR #15237, opened 2022-02-24
decision: install extensions for existing lazy loaded feature

Platonn:


docs/decisions/install-extensions-for-existing-lazy-loaded-feature/install-extensions-for-existing-lazy-loaded-feature.md
```diff
@@ -0,0 +1,395 @@
+## 1. Title
+Install extensions for existing lazy loaded feature
+
+## 2. Status
+Proposed
```
znikola: Approved?

docs/decisions/install-extensions-for-existing-lazy-loaded-feature/install-extensions-for-existing-lazy-loaded-feature.md
```diff
@@ -0,0 +1,395 @@
...
+
+## 3. Involved areas
+extensibility, lazy-loading, dependency injection, schematics installation
+
+## 4. Context
+_Explain why the decision is being taken_
+
+New opt-in Spartacus features should be able to extend the existing lazy-loadable Spartacus features. We didn't have a need for it yet, but now we have a big need. It’s because we extract more-and-more code to lazy-loaded modules and we decouple more-and-more code into independent, opt-in, composable “plugin-like” libraries. And the usual techniques for extending services, don't work in the case of lazy-loaded modules.
```
znikola:
```suggestion
New opt-in Spartacus features should be able to extend the existing lazy-loadable Spartacus features. We didn't have a need for it yet, but now this changes due to extracting more-and-more code to lazy-loaded modules and decoupling into independent, opt-in, composable “plugin-like” libraries. Also, the usual techniques for extending services don't work in the case of lazy-loaded modules.
```

docs/decisions/install-extensions-for-existing-lazy-loaded-feature/install-extensions-for-existing-lazy-loaded-feature.md
```diff
@@ -0,0 +1,395 @@
+## 1. Title
+Install extensions for existing lazy loaded feature
```
znikola, line 4: In points 4 and beyong we left the `docs/decisions/.DECISION-TEMPLATE.md`'s description before starting the actual description. 
For some consistency's sake (and to be concise) I suggest to leave the descriptions for points 1, 2, and 3 as well

docs/decisions/install-extensions-for-existing-lazy-loaded-feature/install-extensions-for-existing-lazy-loaded-feature.md
```diff
@@ -0,0 +1,395 @@
...
+        // CheckoutModule and provides custom extensions
+    },
+  },
+}),
+```
+
+The following diagram shows that e.g. digital payments module imports statically the original checkout module. And in the app our dynamic import points to the digital payments module.
+![](./pre-baked-digital-payments-module-diagram.png)
+
+Unfortunately, this approach allows for installing automatically (via schematics) only one extension feature for one base feature. This is a blocker issue. Taking as an example the Checkout feature, customers might want to use many possible opt-in extensions originating from vaious Spartacus libraries, e.g. `DigitalPaymentsModule`, `B2bCheckoutModule` or `ScheduledReplenishmentCheckoutModule`. The following diagram pictures the dillema, to which library's the dynamic import should point to:
```
znikola: I would also mention how this is not a scalable solution, as it assumes only the pre-baked feature module is used (e.g. CheckoutModule), and doesn't take into an account its other variants (b2b / repl). If we took this approach, it would require us to provide pre-baked modules for every combination of features, which would be:
- hard to maintain
- add great complexity to the schematics installer
- increase the bundle size, since we would be shipping extra pre-baked module which might never be used

docs/decisions/install-extensions-for-existing-lazy-loaded-feature/install-extensions-for-existing-lazy-loaded-feature.md
```diff
@@ -0,0 +1,395 @@
...
+![](./pre-baked-wrapper-modules-diagram.png)
+
+Note: at the time of writing, this solution is used in the `develop` branch. See `DigitalPaymentsModule` as an example: https://github.com/SAP/spartacus/blob/780cd570ca56b2f55a94872b4c0f7ae30b5fdccd/integration-libs/digital-payments/src/digital-payments.module.ts#L5-L8
+
+#### Pros:
+- You can install an extension for an existing lazy loaded feature
+
+#### Cons:
+- You cannot install many extensions for one one base feature
```
znikola, line 97: - not scalable
- bigger bundle size
- hard to maintain

docs/decisions/install-extensions-for-existing-lazy-loaded-feature/install-extensions-for-existing-lazy-loaded-feature.md
```diff
@@ -0,0 +1,395 @@
...
+
+The working example can be found in the PoC PR: https://github.com/SAP/spartacus/pull/14871
+
+The following diagram shows possibility of composing many extensions for the checkout feature:
+![](./wrapper-module-in-the-app.png)
+
+
+Assuming, that the lazy-loadable feature has its own wrapper module in the app, the installation of the extension features means appending the extension module inside the wrapper module.
+
+The installer should basically locate the import of the base feature module in the customer's codebase (e.g. import of `CheckoutModule`), and append the extension module (e.g. `DigitalPayments`) after it. The following example shows the content of the wrapper module, including the base `CheckoutModule` (being a marker for the schematics installer) and the extensions modules `B2bCheckoutModule` and `DigitalPaymentsModule` imported after the marker:
```
znikola, line 148: We can also mention that all the "wrapped" features are being bundled in one chunk and loaded in one go.

docs/decisions/install-extensions-for-existing-lazy-loaded-feature/install-extensions-for-existing-lazy-loaded-feature.md
```diff
@@ -0,0 +1,395 @@
...
+- [x] does it increase the production main JS bundle size?
+  - NO
+- [x] does it cause producing more JS chunks than before?
+  - NO
+- [x] will it break lazy loading when we statically import the base module inside the wrapper module?
+  - NO
+- [x] should we adapt our _installation schematics_ to create wrapper modules OOTB? 
+  - YES
+- [x] should we create wrapper modules for all features or only those that are extended by other opt-in features?
+  - ONLY THE EXTENDED ONES, not to introduce unnecessary boilerplate in the app
```
znikola, line 201:
```suggestion
  - ONLY THE EXTENDABLE ONES, not to introduce unnecessary boilerplate in the app
```

docs/decisions/install-extensions-for-existing-lazy-loaded-feature/install-extensions-for-existing-lazy-loaded-feature.md
```diff
@@ -0,0 +1,395 @@
...
+        // point to library's path 👆
+    },
+  },
+}),
+```
+
+## 5. Alternatives considered
+_List the alternative options you considered with their pros and cons_
+
+### Option 0 (currently used): Pre-baked wrapper modules in the Spartacus extension library
```
znikola:
```suggestion
### Option 1 (currently used): Pre-baked wrapper modules in the Spartacus extension library
```
Let's start from 1?

docs/decisions/install-extensions-for-existing-lazy-loaded-feature/install-extensions-for-existing-lazy-loaded-feature.md
```diff
@@ -0,0 +1,395 @@
...
+- [x] how should the installer behave, when there is no marker module in the app (e.g. because customer dismantled it into smaller modules)
+  - WARN in the console and explain how to import the module manually
+- [x] how should the installer behave, when customer is not using lazy loading (i.e. when the marker module is imported only statically)
+  - NOTHING DIFFERENT, just append the extension module after the base module
+- [x] should we update our installer, so it first installs the base feature and only later the extensions?
+  - YES, we need to add a logic that analyzes the dependencies between feature and extension modules and then installs all features in the correct order
+  
+#### Plan of implementation
+
+After 5.0:
```
znikola:
```suggestion
After 5.0:
```
no longer the case, it is _actually_ (😉) for 5.0

docs/decisions/install-extensions-for-existing-lazy-loaded-feature/install-extensions-for-existing-lazy-loaded-feature.md
```diff
@@ -0,0 +1,395 @@
...
+1. implement schematics util that breaks down the given dynamic import into a wrapper module, e.g:
+
+    `ng g @spartacus/schematics:wrapper --module=CheckoutModule --module-path=@spartacus/checkout/base`
```
znikola:
```suggestion
    `ng g @spartacus/schematics:wrapper --markerModuleName=CheckoutModule --featureModuleName=CheckoutB2BModule`
```

That's how the command will look like. No need to provide the paths

docs/decisions/install-extensions-for-existing-lazy-loaded-feature/install-extensions-for-existing-lazy-loaded-feature.md
```diff
@@ -0,0 +1,395 @@
...
+2. implement schematics util that appends the given extension module after another "marker" module, e.g:
+    
+    `ng g @spartacus/schematics:append-module --module=DigitalPaymentsModule --module-path=@spartacus/digital-payments --after-module=CheckoutModule --after-module-path=@spartacus/checkout/base`
```
znikola: No need for this. All the wrapper module schematic needs is the 
- marker (`markerModuleName` option)
- and the module which should be appended (`featureModuleName` option)

docs/decisions/install-extensions-for-existing-lazy-loaded-feature/install-extensions-for-existing-lazy-loaded-feature.md
```diff
@@ -0,0 +1,395 @@
...
+3. change the implementation of the installation-schematics of our extension features (e.g. DigitalPayments, CheckoutB2b) use the two util schematics above. For example: to install the Digital Payments feature (`ng add @spartacus/digital-payments`): first make sure to break down the dynamic import of `CheckoutModule` into a local wrapper module and only then append the `DigitalPaymentModule` after the `CheckoutModule`.
+
+Issues with current develop branch:
+1. because of putting the pre-baked modules into customers app (so the base module is implicitly imported inside), there is no notion about the base module the customer's code. Therefore the base module cannot be the marker for generating the wrapper modules in the future.
+   - possible solutions:
+     1. workaround: the future schematics generating the wrapper module for the Checkout will accept an array of possible maker modules (e.g. `CheckoutModule, CheckoutB2BModule, DigitalPaymentsModule`)
+        - cons: we will have to change in 6.0 those pre-baked wrapper modules (it's tech dept to pay in the future)
```
znikola: No longer the case

docs/decisions/install-extensions-for-existing-lazy-loaded-feature/install-extensions-for-existing-lazy-loaded-feature.md
```diff
@@ -0,0 +1,395 @@
+## 1. Title
```
znikola, line 1: What do you think to link this ticket https://github.com/SAP/spartacus/issues/15380 when we start to mention schematics in this doc?

docs/decisions/install-extensions-for-existing-lazy-loaded-feature/install-extensions-for-existing-lazy-loaded-feature.md
```diff
@@ -0,0 +1,395 @@
...
+_What becomes easier or more difficult to do because of this change?_
+
+TODO
+
+### Order of installing libraries separately becomes important
+When installing libraries separately via bash installation script, the order of installing libraries becomes important, e.g.:
+```bash
+ng add --skip-confirmation @spartacus/checkout --interactive false
+ng add --skip-confirmation @spartacus/digital-payments --interactive false
+# not the extension should go after the default feature 👆
```
znikola:
```suggestion
# note the extension should go after the default feature 👆
```
?

docs/decisions/install-extensions-for-existing-lazy-loaded-feature/install-extensions-for-existing-lazy-loaded-feature.md
```diff
@@ -0,0 +1,395 @@
...
+```
+
+Instead of installing libraries separately, it's recommended to run `ng add @spartacus/schematics --features=...` (which automatically takes care about the order of installing libraries):
```
znikola, line 317: When invoking the libraries' schematics, we will also analyze the cross-feature dependencies, create the dependency graph, and install all the missing libraries.
There might be some corner cases where this won't work, like in cases where customers heavily customized their apps, and in those we will abort the installer with appropriate messages.

znikola (changes requested):

docs/decisions/install-extensions-for-existing-lazy-loaded-feature/install-extensions-for-existing-lazy-loaded-feature.md
Platonn: I'm a developer, I start counting from 0 😜.
... kidding, I just added this option on top very recently and was lazy to update all the numbers

You're right. I'll update all the numbers starting from 1 😉

docs/decisions/install-extensions-for-existing-lazy-loaded-feature/install-extensions-for-existing-lazy-loaded-feature.md
Platonn: What if the customer creates a custom module with the same name in their codebase?

docs/decisions/install-extensions-for-existing-lazy-loaded-feature/install-extensions-for-existing-lazy-loaded-feature.md
Platonn: IMHO we cannot predict which features won't be extended in the future by other extensions. So theoretically every feature is "extendable".
My proposal is to leave as-is the original dynamic imports pointing to Spartacus lib path. And break them down into local wrapper modules only on-demand, when some extension is being installed.

docs/decisions/install-extensions-for-existing-lazy-loaded-feature/install-extensions-for-existing-lazy-loaded-feature.md
Platonn: I've removed now all the example cli commands. They were added just as an example. But they were not meant to be run via CLI, but rather programatically. So it was misleading anyway

docs/decisions/install-extensions-for-existing-lazy-loaded-feature/install-extensions-for-existing-lazy-loaded-feature.md
Platonn: I'm happy to add more details to this doc, if it contributes to the decision, context, or pros/cons. At the same time I'd like to balance the level of details to be eaisly digestable for the reader.

If some info impacts the decision, I'd like to copy/rephrase it directly in this doc, instead of linking to external resources (like ticket). I'd like this document to be standalone, comprehensive and immutable over time (in terms of git commit history). In the future, if we tweak/change the decision, this should be reflected in the git commit history of `develop`.

Would you mind proposing what and where should I add in this doc?

znikola: Ok, that makes sense. 
I was thinking we could link it like: for the current docs, "please visit this page", or "see this ticket/PR".

docs/decisions/install-extensions-for-existing-lazy-loaded-feature/install-extensions-for-existing-lazy-loaded-feature.md
znikola: We check if the module is imported from an appropriate library - e.g. we check if CheckoutB2BModule is imported from `@spartacus/checkout/b2b`

Platonn:
decided. moved to wiki

PR #15240, opened 2022-02-24
chore: fix a bug in page-layout-handler.ts

### LINKED ISSUE #15236: Page layout handler is not called on page rendering
VariantConfiguratorPageLayoutHandler's handle method is no longer called on navigating to a new page. It is only called once after a browser refresh.

I cannot exactly tell what broke it, but it worked in develop branch at least until February 1st (with commit 3813105)

WeizhengSap:
close #15236


projects/storefrontlib/cms-structure/page/page-layout/page-layout.service.ts
```diff
@@ -45,7 +46,7 @@ export class PageLayoutService {
       }),
       switchMap(({ slots, pageTemplate, breakpoint }) => {
         let result = of(slots);
-        for (const handler of getLastValueSync(this.handlers$) || []) {
+        for (const handler of this.handlers || []) {
```
plabadie, line 49: Could there be a race condition here?  (like theoretically could the value of `this.handlers` change while looping over it? ).  Could a `withLatestFrom` or another rxjs operator mitigate this?

Not too sure about this myself, lets discuss

WeizhengSap: The race condition should not happen. CmsPageGuard trigger the module loading. Only when module is loaded (which provides `PAGE_LAYOUT_HANDLER`), then `PageLayoutService` can getSlot.

plabadie (approved):

PR #15250, opened 2022-02-24
fix: Do not allow user to select an already selected address or payment in checkout

### LINKED ISSUE #15249: Checkout: do not allow card selection of already selected address/payment
currently, we can select already selected address/payment cards, which fires requests to set it again in the backend, but we do not need to set it again when it is already selected.

Moreover, every time we select a new address, it deletes the delivery mode set value from the backend. Therefore, clicking the same address should not be allowed as it puts the delivery method back to 'TBD' in the order overview section.

bgambocjaviniar:
closes GH-15249

### DIFF
--- a/feature-libs/checkout/base/components/checkout-payment-method/checkout-payment-method.component.spec.ts
+++ b/feature-libs/checkout/base/components/checkout-payment-method/checkout-payment-method.component.spec.ts
@@ -41,6 +41,35 @@ const mockPaymentDetails: PaymentDetails = {
   cvn: '123',
 };
 
+const mockPayments: PaymentDetails[] = [
+  {
+    id: 'non default method',
+    accountHolderName: 'Name',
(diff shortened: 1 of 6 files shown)


feature-libs/checkout/base/components/checkout-payment-method/checkout-payment-method.component.spec.ts
```diff
@@ -531,5 +533,34 @@ describe('CheckoutPaymentMethodComponent', () => {
         <any>mockActivatedRoute
       );
     });
+
+    it('should be able to payment method', () => {
```
znikola:
```suggestion
    it('should be able to select a payment method', () => {
```
?

znikola (changes requested):

znikola (approved):

PR #15251, opened 2022-02-25
fix: Remove assets from EPD peer dependencies and fix 4.3.x pipeline

Zeyber:


integration-libs/epd-visualization/.release-it.json
```diff
@@ -34,7 +34,6 @@
             "peerDependencies.@spartacus/core",
             "peerDependencies.@spartacus/styles",
             "peerDependencies.@spartacus/storefront",
-            "peerDependencies.@spartacus/assets",
```
Platonn, line 37: Nice. 
I believe we need to repeat the same change in develop, to avoid similar problem when we'll release 5.0 out from develop branch: https://github.com/SAP/spartacus/blob/develop/integration-libs/epd-visualization/.release-it.json#L37

Platonn (dismissed):

integration-libs/epd-visualization/.release-it.json
Zeyber: PR for change on develop: https://github.com/SAP/spartacus/pull/15260

Platonn (approved):

PR #15266, opened 2022-02-28
fix: add region input to card element (#15246)

### LINKED ISSUE #15246: Aria Roles without names in Review Order
**Describe the bug**

- AMP detects aria roles without accessible names.
- This is probably coming from cx-card(s) of the previous steps where regions have names, **proceed carefully.** 

Screen Shot 2022-02-24 at 4 00 03 PM

**To Reproduce**
1. Go to the last step on checkout
2. Run AMP and see the error.

**Expected behaviour**

- No AMP errors.
- **Best solution:** Remove roles at the last step **without** affecting previous steps.
- **Alternative solution:** If roles can't be removed at the last step, then **add `aria-label` and remove `tabindex="-1"`** so the user can jump to them with the rotor. Suggested names: _Review Payment, Review Shipping Address, Review Shipping Method_.

BenjaminEmiliani:
Closes #15246


projects/storefrontlib/shared/components/card/card.component.ts
```diff
@@ -44,6 +44,9 @@ export class CardComponent implements OnInit {
   @Input()
   border = false;
 
+  @Input()
```
Zeyber: We usually don't want to add more properties as inputs if we don't need to. We already have an input of `content: Card`. Could we add `role`to the Card interface and pass it in with the `content`input?

Zeyber (changes requested):

projects/storefrontlib/shared/components/card/card.component.ts
BenjaminEmiliani: Good catch, I will try that approach

projects/storefrontlib/shared/components/card/card.component.ts
```diff
@@ -21,6 +21,7 @@ export interface Card {
   actions?: Array<CardAction | CardLinkAction>;
   deleteMsg?: string;
   label?: string;
+  isRegion?: boolean;
```
Zeyber: Why do we need to limit `role` to `region` only? Couldn't we have this as a `string` or even better `enum` with role types?

Zeyber (changes requested):

projects/storefrontlib/shared/components/card/card.component.ts
BenjaminEmiliani: I will explore this

Zeyber (dismissed):

Zeyber (approved):

Issue #15267, opened 2022-03-01
Anonymous users are not redirected to login page after trying to access a checkout step route since 3.0.0


bgambocjaviniar:
**Old checkout bug behavior since Spartacus version 3.0.0** 

Checkout auth guard is not redirecting anonymous users that has no active cart id. They get stuck in a blank page if they were to try to visit the checkout route directly

bgambocjaviniar:
QA:

as an anonymous user

- clear all storage (local / session / cookies)
- refresh the page

try to visit a checkout route when active cart is empty. You can view the active cart being empty through the synced local storage with our redux store under the `spartacus⚿electronics-spa⚿cart`. It should show an empty string ""  for "active". 

Example: 
paste the route in your url with electronics-spa/en/USD/checkout/payment-details
or electronics-spa/en/USD/checkout/delivery-address
etc

Expected result: You should be redirected to the login page

---------
OLD behaviour since 3.0: you would have been stuck in a blank page

giancorderoortiz:
@KateChuen please add this issue to list of critical issues to check for 5.0. Security concern.

KateChuen:
Tested using develop branch.
Now, user is redirected to Login page. 
With an empty string for "active" cart

Tested with checkout/delivery-address, checkout/delivery-mode, checkout/payment-details, checkout/review-order
![Screen Shot 2022-03-24 at 12 42 55 PM](https://user-images.githubusercontent.com/45409515/160141351-a9cebf5e-2ca6-4ba0-86e0-26c7e5c8af4c.png)

PR #15268, opened 2022-03-01
fix: Anonymous users are not redirected to login page after trying to access a checkout step route (epic)

bgambocjaviniar:
fixes issue that existed since 3.0.0

closes GH-15267


bgambocjaviniar:
closing as we're merging to develop

PR #15269, opened 2022-03-01
fix: Anonymous users are not redirected to login page after trying to access a checkout step route

### LINKED ISSUE #15267: Anonymous users are not redirected to login page after trying to access a checkout step route since 3.0.0

bgambocjaviniar:
fixes issue that existed since 3.0.0

`Checkout auth guard is not redirecting anonymous users that has no active cart id as one of the stream in the combineLatest 'isStable'` always returns false 

closes https://github.com/SAP/spartacus/issues/15267


projects/core/src/state/utils/processes-loader/processes-loader.selectors.ts
```diff
@@ -1,7 +1,9 @@
 import { ProcessesLoaderState } from './processes-loader-state';
 
 export function isStableSelector<T>(state: ProcessesLoaderState<T>): boolean {
-  return state.processesCount === 0 && !state.loading;
+  return state.processesCount === undefined && !state.loading
+    ? true
+    : state.processesCount === 0 && !state.loading;
```
znikola: this is kind of tough to read ;/
can we break it down somehow?

znikola (dismissed):

feature-libs/cart/base/core/facade/multi-cart.service.spec.ts
```diff
@@ -232,15 +242,19 @@ describe('MultiCartService', () => {
...
           cartId: 'xxx',
+          extraData: {
+            active: true,
+          },
         })
       );
+
       service
-        .isStable('cartId')
+        .isStable('xxx')
```
bgambocjaviniar, line 252: just a note for the reviewer,

this test was broken since the creation of it because when it got to the selectors, it always uses the `initialLoaderState` instead of the actual process cart state as the cartId was not the same as the one that was dispatched due to links below

basically, 'cartId' was being passed, but the store had 'xxx' as the actual cart data

- https://github.com/SAP/spartacus/blob/develop/projects/core/src/state/utils/entity-processes-loader/entity-processes-loader.selectors.ts#L30
- https://github.com/SAP/spartacus/blob/develop/projects/core/src/state/utils/entity-loader/entity-loader.selectors.ts#L15

projects/storefrontapp-e2e-cypress/cypress/helpers/checkout-flow.ts
```diff
@@ -173,9 +174,7 @@ export function goToProductDetailsPage() {
 
 export function addProductToCart() {
   cy.get('cx-item-counter').findByText('+').click();
-  cy.get('cx-add-to-cart')
-    .findByText(/Add To Cart/i)
-    .click();
+  addToCart();
```
bgambocjaviniar, line 177: reason why I added this function that waits for the cart to be 'refreshed' is because cypress did not wait and when we proceeded to checkout, it returned us to homepage due to the `CartNotEmptyGuard` as the cart was empty when it clicked proceed to checkout

bgambocjaviniar: this applies to all instances in the application where we do 

```
  cy.get('cx-add-to-cart')
    .findByText(/Add To Cart/i)
    .click();
```

feature-libs/cart/base/core/store/selectors/multi-cart.selector.spec.ts
```diff
@@ -194,21 +195,58 @@ describe('Multi Cart selectors', () => {
   });
 
   describe('getCartIsStableSelectorFactory', () => {
-    it('should return cart stability flag', () => {
-      let result;
+    it('should return true when cart is stable when there is no active cart', (done) => {
```
bgambocjaviniar, line 198: when I say there is no active cart, it means active is currently set to empty string AND if you look at the cart process state, it does not have the processCount set as active cart is empty

WeizhengSap (approved):

PR #15270, opened 2022-03-01
feat: Add alt text for image link with no image (#15235)

### LINKED ISSUE #15235: Unavailable image link needs alt text
**Describe the bug**
AMP detects missing accessible name for unavailable image, since this is not a real image, the link itself needs text. I suggest adding `aria-label` to the link.

Screen Shot 2022-02-23 at 8 19 31 PM

**Expected behaviour**
We have two options, please implement whichever is possible.

- **Preferred** solution: SR says vocalizes: _"no image available, product name"_ 

Screen Shot 2022-02-23 at 8 27 10 PM

- **Generic** solution: SR vocalizes: _"no image available"_ for all unavailable images.

Screen Shot 2022-02-23 at 8 24 22 PM

dydome:
close #15235


projects/storefrontlib/cms-components/product/product-list/product-list-item/product-list-item.component.html
```diff
@@ -1,6 +1,12 @@
 <div class="row">
   <div class="col-12 col-md-4">
     <a
+      [attr.aria-label]="
+        !product.images?.hasOwnProperty('PRIMARY')
+          ? ('productDetails.noProductImage'
+            | cxTranslate: { product: product.name })
+          : ''
```
pawelfras: `should:` please set `undefined` if this attribute is not need, as with empty string `aria-label` will still exist for an element

pawelfras (approved):

PR #15275, opened 2022-03-01
GH-15005 Use "totalUnitCount" over "totalItems" for Saved Cart Dialog

### LINKED ISSUE #15005: Saved Cart qty is totalItems instead of totalUnitCount in SavedCartFormDialogComponent
**Describe the bug**
When a saved cart is made active, the SavedCartFormDialogComponent shows a small summary of the saved cart. The qty that is presented is not equal to the total amount of products in the cart, but is equal to the amount of unique products. This is not consistent with the qty that is presented in the saved carts list.

- Library version: 4.3

**To Reproduce**
1. Create a cart with 2 unique products
2. Make sure that the quantities of both products are not the same
3. Go to cart
4. Save cart
5. Go to saved carts page
6. Check quantity in list (this equals totalUnitCount)
7. Click "make cart active"
8. Check quantity in modal (this equals totalItems, should be totalUnitCount)

**Expected behavior**
In the SavedCartFormDialogComponent, the value of qty is the totalItems. This refers to the total number of unique items. This should be the totalUnitCount as it represents the total amount of product in the cart. This makes it also in line with the saved carts list, where the totalUnitCount is used as well.

**Desktop:**
 - OS: macOS Big Sur 
 - Browser: Firefox
 - Version: 97.0

justinlee01:
Fixes: https://github.com/SAP/spartacus/issues/15005

SavedCartFormDialogComponent should be using totalUnitCount rather than totalItems for displaying quantity in the cart

### DIFF
--- a/feature-libs/cart/saved-cart/components/saved-cart-form-dialog/saved-cart-form-dialog.component.html
+++ b/feature-libs/cart/saved-cart/components/saved-cart-form-dialog/saved-cart-form-dialog.component.html
@@ -89,7 +89,7 @@
                 {{ 'savedCartDialog.quantity' | cxTranslate }}
               </div>
               <div class="cx-saved-cart-value">
-                {{ cart?.totalItems }}
+                {{ cart?.totalUnitCount }}
               </div>
             </div>
 


justinlee01:
Will rename branch

PR #15280, opened 2022-03-02
feature/GH-14685: fix contrast issues

### LINKED ISSUE #14685: AMP: Contrast issues (Severity 4)
Severity 4

There are several issues wrt contrast.

AMP-Note (example): "This SPAN contains text with a background color of #f4f4f4 rgb(244, 244, 244) and foreground color of #747881 rgb(116, 120, 129) that is less than 18 point in size; or bold text less than 14 point in size that has a luminosity contrast ratio of 4.023, which is below 4.5:1"

Tester remark:
It has to be checked whether those contrast issues are not part of theme-styling.
Only configuration-specific styling might be adapted.

Example:
Contrast example

schuppell:


schuppell:
Please wait before merging!

ChristophHi (approved):

PR #15291, opened 2022-03-03
feat: add vocalisation for star rating

### LINKED ISSUE #15039: screen reader: Missing vocal feedback required on review stars
**Steps:**
Go to a pdp
Write a review
Focus on the review stars

**Result:**
Missing the vocal feedback "required".

reviewStars

Matejk00:
Closes: #15039


projects/assets/src/translations/en/product.ts
```diff
@@ -61,6 +61,8 @@ export const product = {
     reviewTitle: 'Review Title',
     writeYourComments: 'Write your comments',
     rating: 'Rating',
+    ariaLabelRating: 'Product rating, required',
+    ariaLabelRate: 'Add rate: {{rating}}',
```
BenjaminEmiliani: Hey @Matejk00 I think the names of these two keys can be changed to something more descriptive of their value rather than their purpose. What I mean is, instead of:
`ariaLabelRating: 'Product rating, required',`
It can be:
`ratingRequired: 'Product rating, required',`

Just something to think about

BenjaminEmiliani: One other thing to consider is maybe add 'stars' to `ariaLabelRate`. i.ie:
`ariaLabelRate: 'Add rate: {{rating}} stars'`

BenjaminEmiliani (commented):
Works well with VO, only left some comments related to syntax and wording

BenjaminEmiliani (dismissed):
Works good on VO. I doubt behaviour will be different on NVDA/JAWs but it could be worth confirming

developpeurweb:
Works well on JAWS/NVDA/VO.

dydome (approved):

PR #15292, opened 2022-03-03
feat: Vocalisation for PLP after rearrangement by sorting/facets/view (#15217)

### LINKED ISSUE #15217: Missing vocalized confirmation when PLP rearranges results on page
**Describe the bug**
When facets, sort by or display grid are triggered, the results on the page will rearrange without any audible confirmation that the page view has changed.

**To Reproduce**
1. Go to PLP using a screen reader and keyboard
2. Trigger any of the following: facets, sort by, display grid button.
3. Hear the silence.

**Expected behaviour**
For any of the aforementioned options being triggered a message will be vocalized: _"Page view updated with your selected options"_.

Screen Shot 2022-02-22 at 10 09 23 AM

pawelfras:
closes #15217


developpeurweb (changes requested):
This works for NVDA and VO. 
But not for JAWS, neither on Chrome nor Firefox. Needs to work on JAWS on at least one of the two browsers.

developpeurweb (dismissed):
The update message is heard after the JAWS bloat. NVDA/VO work as expected.

Matejk00 (dismissed):

Matejk00 (approved):

PR #15297, opened 2022-03-03
fix: i18n when using JSON in SSR - set `reloadInterval: false`

### LINKED ISSUE #14372: JSON translations don't work in SSR Spartacus 4.x

Platonn:
Previously, the periodical reloading of JSON translations in SSR was enabled by default by `i18next-http-backend`. The never-ending async task `setInterval()` invoked inside `i18next-http-backend` caused the SSR to hang.

To prevent it, we're setting explicitly `reloadInterval: false` in the options of `i18next`.

fixes #14372

### DIFF
--- a/projects/core/src/i18n/i18next/i18next-init.ts
+++ b/projects/core/src/i18n/i18next/i18next-init.ts
@@ -43,7 +43,7 @@ export function i18nextInit(
               loadPath,
               request: i18nextGetHttpClient(httpClient),
 
-              // Disable the periodical reloading, because the `setInterval()` used inside `i18next-http-backend` would prevent the SSR to finish
+              // Disable the periodical reloading. Otherwise the SSR would not finish due to the pending task `setInterval()`
               // See source code: https://github.com/i18next/i18next-http-backend/blob/00b7e8f67abf8372af17529b51190a7e8b17e3d8/i18nextHttpBackend.js#L94-L95
               reloadInterval: false,
             };
--- a/projects/core/src/i18n/i18next/i18next-init.ts
+++ b/projects/core/src/i18n/i18next/i18next-init.ts
@@ -43,8 +43,8 @@ export function i18nextInit(
               loadPath,
               request: i18nextGetHttpClient(httpClient),
 
-              // Disable the periodical reloading. Otherwise the SSR would not finish due to the pending task `setInterval()`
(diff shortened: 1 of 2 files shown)


projects/core/src/i18n/i18next/i18next-init.ts
```diff
@@ -16,7 +16,7 @@ export function i18nextInit(
   configInit: ConfigInitializerService,
   languageService: LanguageService,
   httpClient: HttpClient,
-  serverRequestOrigin: string,
+  serverRequestOrigin: string | null,
```
Platonn, line 19: just fixed typing by the way. SERER_REQUEST_ORIGIN is `@Optional`

znikola (approved):

PR #15303, opened 2022-03-03
fix: fix export button align issue

### LINKED ISSUE #14535: Export button not aligned properly on order-confirmation page
When an order is placed, we can see the "export to CSV" button underneath the cart items on the order-confirmation page.
It is styled nicely for desktop resolution, but it appears to be misaligned on lower resolutions (see screen-shot):
Screenshot 2021-12-06 at 11 19 48

kimhw0630:
fixed: https://github.com/SAP/spartacus/issues/14535 
fix for export button align issue in the order confirmation screen


giancorderoortiz:
Does it work well in mobile too ?

PR #15304, opened 2022-03-03
fix: Adjust aria attributes to vocalize All Stores result count

### LINKED ISSUE #15243: Store Finder: View all stores doesn't vocalize and display number of results
**Steps:**
Go to electronics -> Find a store 
Click on "View all stores"

**Result:**
Before it used to vocalize the number of result found. (was tested in https://github.com/SAP/spartacus/issues/13738)
But it looks like it's broken now, and it says '111' which is meaningless to the user. 

storeFinder1

FollowTheFlo:
Closes #15243

### DIFF
--- a/feature-libs/storefinder/components/store-finder-stores-count/store-finder-stores-count.component.html
+++ b/feature-libs/storefinder/components/store-finder-stores-count/store-finder-stores-count.component.html
@@ -4,12 +4,7 @@
     else loading
   "
 >
-  <div
-    class="cx-count container"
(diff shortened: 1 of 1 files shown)


Matejk00 (approved):

Matejk00:
Looks good from code perspective. Would be nice to re-test this solution cross all Screen Readers before merge.

developpeurweb (approved):
Works well with JAWS/NVDA/VO

PR #15305, opened 2022-03-03
GH-15301: Generate Cart Lib and Checkout C&Q constructor change schematics

plabadie:
Replaces PR #15247

PR #15307, opened 2022-03-04
GH-15306: Generate Cart Lib and Checkout C&Q migration schematics & docs

### LINKED ISSUE #15306: Generate Cart Lib and Checkout C&Q migration schematics & docs
The constructor schematics are part of issue GH-15301

plabadie:


projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/cart-lib-checkout.migration.ts
```diff
@@ -0,0 +1,1719 @@
+import { MethodPropertyDeprecation } from 'projects/schematics/src/shared/utils/file-utils';
+
+export const CART_LIB_AND_CHECKOUT_MIGRATION: MethodPropertyDeprecation[] = [
+  {
+    class: 'AddToSavedCartComponent',
+    importPath: '@spartacus/cart/saved-cart/components',
+    deprecatedNode: 'activeCartService',
+    comment:
+      '// TODO:Spartacus - Property "activeCartService" was removed from Class "AddToSavedCartComponent". ',
```
bgambocjaviniar: was renamed, and not removed

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/cart-lib-checkout.migration.ts
```diff
@@ -0,0 +1,1719 @@
...
+  },
+  {
+    class: 'SavedCartAdapter',
+    importPath: '@spartacus/cart/saved-cart/core',
+    deprecatedNode: 'saveCart',
+    comment:
+      '// TODO:Spartacus - Method "saveCart" was removed from Class "SavedCartAdapter". ',
```
bgambocjaviniar: moved to `CartAdapter` and renamed to Method `save`

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/cart-lib-checkout.migration.ts
```diff
@@ -0,0 +1,1719 @@
...
+  },
+  {
+    class: 'SavedCartConnector',
+    importPath: '@spartacus/cart/saved-cart/core',
+    deprecatedNode: 'saveCart',
+    comment:
+      '// TODO:Spartacus - Method "saveCart" was removed from Class "SavedCartConnector". ',
```
bgambocjaviniar: moved to CartConnector and renamed to Method save

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/cart-lib-checkout.migration.ts
```diff
@@ -0,0 +1,1719 @@
...
+    deprecatedNode: 'getSaveCartEndpoint',
+    comment:
+      '// TODO:Spartacus - Method "getSaveCartEndpoint" was removed from Class "OccSavedCartAdapter". ',
+  },
+  {
+    class: 'OccSavedCartAdapter',
+    importPath: '@spartacus/cart/saved-cart/occ',
+    deprecatedNode: 'saveCart',
+    comment:
+      '// TODO:Spartacus - Method "saveCart" was removed from Class "OccSavedCartAdapter". ',
```
bgambocjaviniar: moved to CartAdapter and renamed to Method save

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/cart-lib-checkout.migration.ts
```diff
@@ -0,0 +1,1719 @@
...
+    deprecatedNode: 'actionsSubject',
+    comment:
+      '// TODO:Spartacus - Property "actionsSubject" was removed from Class "SavedCartOrderEntriesContext". ',
+  },
+  {
+    class: 'CartNotEmptyGuard',
+    importPath: '@spartacus/checkout/components',
+    deprecatedNode: 'activeCartService',
+    comment:
+      '// TODO:Spartacus - Property "activeCartService" was removed from Class "CartNotEmptyGuard". ',
```
bgambocjaviniar: was renamed to activeCartFacade

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/cart-lib-checkout.migration.ts
```diff
@@ -0,0 +1,1719 @@
...
+  },
+  {
+    class: 'CheckoutAuthGuard',
+    importPath: '@spartacus/checkout/components',
+    deprecatedNode: 'activeCartService',
+    comment:
+      '// TODO:Spartacus - Property "activeCartService" was removed from Class "CheckoutAuthGuard". ',
```
bgambocjaviniar: was renamed to activeCartFacade

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/cart-lib-checkout.migration.ts
```diff
@@ -0,0 +1,1719 @@
...
+    deprecatedNode: 'userService',
+    comment:
+      '// TODO:Spartacus - Property "userService" was removed from Class "CheckoutAuthGuard". ',
+  },
+  {
+    class: 'CheckoutGuard',
+    importPath: '@spartacus/checkout/components',
+    deprecatedNode: 'activeCartService',
+    comment:
+      '// TODO:Spartacus - Property "activeCartService" was removed from Class "CheckoutGuard". ',
```
bgambocjaviniar: was renamed to activeCartFacade

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/cart-lib-checkout.migration.ts
```diff
@@ -0,0 +1,1719 @@
...
+  },
+  {
+    class: 'CheckoutLoginComponent',
+    importPath: '@spartacus/checkout/components',
+    deprecatedNode: 'activeCartService',
+    comment:
+      '// TODO:Spartacus - Property "activeCartService" was removed from Class "CheckoutLoginComponent". ',
```
bgambocjaviniar: was renamed to activeCartFacade

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/cart-lib-checkout.migration.ts
```diff
@@ -0,0 +1,1719 @@
...
+  },
+  {
+    class: 'CheckoutOrderSummaryComponent',
+    importPath: '@spartacus/checkout/components',
+    deprecatedNode: 'activeCartService',
+    comment:
+      '// TODO:Spartacus - Property "activeCartService" was removed from Class "CheckoutOrderSummaryComponent". ',
```
bgambocjaviniar: was renamed to activeCartFacade

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/cart-lib-checkout.migration.ts
```diff
@@ -0,0 +1,1719 @@
...
+  },
+  {
+    class: 'CheckoutProgressMobileTopComponent',
+    importPath: '@spartacus/checkout/components',
+    deprecatedNode: 'activeCartService',
+    comment:
+      '// TODO:Spartacus - Property "activeCartService" was removed from Class "CheckoutProgressMobileTopComponent". ',
```
bgambocjaviniar: was renamed to activeCartFacade

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/cart-lib-checkout.migration.ts
```diff
@@ -0,0 +1,1719 @@
...
+  },
+  {
+    class: 'CheckoutProgressMobileTopComponent',
+    importPath: '@spartacus/checkout/components',
+    deprecatedNode: 'ngOnInit',
+    comment:
+      '// TODO:Spartacus - Method "ngOnInit" was removed from Class "CheckoutProgressMobileTopComponent". ',
+  },
+  {
+    class: 'CheckoutStepsSetGuard',
```
bgambocjaviniar: confusing guard and related to the constructor one. Possibility of removing and to go with the source of truthy md?

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/cart-lib-checkout.migration.ts
```diff
@@ -0,0 +1,1719 @@
...
+    deprecatedNode: 'checkoutDetailsService',
+    comment:
+      '// TODO:Spartacus - Property "checkoutDetailsService" was removed from Class "ExpressCheckoutService". ',
+  },
+  {
+    class: 'ExpressCheckoutService',
+    importPath: '@spartacus/checkout/components',
+    deprecatedNode: 'checkoutPaymentService',
+    comment:
+      '// TODO:Spartacus - Property "checkoutPaymentService" was removed from Class "ExpressCheckoutService". ',
```
bgambocjaviniar: renamed to checkoutPaymentFacade

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/cart-lib-checkout.migration.ts
```diff
@@ -0,0 +1,1719 @@
...
+    deprecatedNode: 'clearCheckoutService',
+    comment:
+      '// TODO:Spartacus - Property "clearCheckoutService" was removed from Class "ExpressCheckoutService". ',
+  },
+  {
+    class: 'ExpressCheckoutService',
+    importPath: '@spartacus/checkout/components',
+    deprecatedNode: 'setShippingAddress',
+    comment:
+      '// TODO:Spartacus - Method "setShippingAddress" was removed from Class "ExpressCheckoutService". ',
```
bgambocjaviniar: renamed to setDeliveryAddress

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/cart-lib-checkout.migration.ts
```diff
@@ -0,0 +1,1719 @@
...
+    deprecatedNode: 'checkoutDeliveryService',
+    comment:
+      '// TODO:Spartacus - Property "checkoutDeliveryService" was removed from Class "ExpressCheckoutService". ',
+  },
+  {
+    class: 'ExpressCheckoutService',
+    importPath: '@spartacus/checkout/components',
+    deprecatedNode: 'checkoutDetailsService',
+    comment:
+      '// TODO:Spartacus - Property "checkoutDetailsService" was removed from Class "ExpressCheckoutService". ',
```
bgambocjaviniar: not sure what to say about this, but this was split into checkoutDeliveryAdddressFacade, checkoutPaymentFacade, checkoutDeliveryModesFacade as they are all pulling from the checkoutqueryservice, which is the checkoutdetailsservice

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/cart-lib-checkout.migration.ts
```diff
@@ -0,0 +1,1719 @@
...
+    deprecatedNode: 'setShippingAddress',
+    comment:
+      '// TODO:Spartacus - Method "setShippingAddress" was removed from Class "ExpressCheckoutService". ',
+  },
+  {
+    class: 'NotCheckoutAuthGuard',
+    importPath: '@spartacus/checkout/components',
+    deprecatedNode: 'activeCartService',
+    comment:
+      '// TODO:Spartacus - Property "activeCartService" was removed from Class "NotCheckoutAuthGuard". ',
```
bgambocjaviniar: renamed to activeCartFacade

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/cart-lib-checkout.migration.ts
```diff
@@ -0,0 +1,1719 @@
...
+    deprecatedNode: 'placeOrder',
+    comment:
+      '// TODO:Spartacus - Method "placeOrder" was removed from Class "CheckoutConnector". ',
+  },
+  {
+    class: 'CheckoutCostCenterService',
+    importPath: '@spartacus/checkout/core',
+    deprecatedNode: 'activeCartService',
+    comment:
+      '// TODO:Spartacus - Property "activeCartService" was removed from Class "CheckoutCostCenterService". ',
```
bgambocjaviniar: renamed to activeCartFacade

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/cart-lib-checkout.migration.ts
```diff
@@ -0,0 +1,1719 @@
...
+    deprecatedNode: 'checkoutStore',
+    comment:
+      '// TODO:Spartacus - Property "checkoutStore" was removed from Class "CheckoutCostCenterService". ',
+  },
+  {
+    class: 'CheckoutCostCenterService',
+    importPath: '@spartacus/checkout/core',
+    deprecatedNode: 'getCostCenter',
+    comment:
+      '// TODO:Spartacus - Method "getCostCenter" was removed from Class "CheckoutCostCenterService". ',
```
bgambocjaviniar: was renamed to getCostCenterState -ish. Last time it returned  cost center code, but now it returns the query state and cost center data, which has the cost center code.

If above is true that it's getCostCenterState instead, then I argue setCostCenter is not the same as it uses commands under the hood instead of ngrx. 

Although, I would add that setCostCenter returns `Observable<QueryState<CostCenter | undefined>>` instead of a void type

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/cart-lib-checkout.migration.ts
```diff
@@ -0,0 +1,1719 @@
...
+  },
+  {
+    class: 'CheckoutPageMetaResolver',
+    importPath: '@spartacus/checkout/core',
+    deprecatedNode: 'activeCartService',
+    comment:
+      '// TODO:Spartacus - Property "activeCartService" was removed from Class "CheckoutPageMetaResolver". ',
```
bgambocjaviniar: was renamed to activeCartFacade

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/cart-lib-checkout.migration.ts
```diff
@@ -0,0 +1,1719 @@
...
+    deprecatedNode: 'cart$',
+    comment:
+      '// TODO:Spartacus - Property "cart$" was removed from Class "CheckoutPageMetaResolver". ',
+  },
+  {
+    class: 'CheckoutPageMetaResolver',
+    importPath: '@spartacus/checkout/core',
+    deprecatedNode: 'translation',
+    comment:
+      '// TODO:Spartacus - Property "translation" was removed from Class "CheckoutPageMetaResolver". ',
```
bgambocjaviniar: was renamed to translationService

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/cart-lib-checkout.migration.ts
```diff
@@ -0,0 +1,1719 @@
...
+  },
+  {
+    class: 'CheckoutPaymentAdapter',
+    importPath: '@spartacus/checkout/core',
+    deprecatedNode: 'create',
+    comment:
+      '// TODO:Spartacus - Method "create" was removed from Class "CheckoutPaymentAdapter". ',
```
bgambocjaviniar: was renamed to createPaymentDetails

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/cart-lib-checkout.migration.ts
```diff
@@ -0,0 +1,1719 @@
...
+  },
+  {
+    class: 'CheckoutPaymentAdapter',
+    importPath: '@spartacus/checkout/core',
+    deprecatedNode: 'loadCardTypes',
+    comment:
+      '// TODO:Spartacus - Method "loadCardTypes" was removed from Class "CheckoutPaymentAdapter". ',
```
bgambocjaviniar: was renamed to getCardTypes

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/cart-lib-checkout.migration.ts
```diff
@@ -0,0 +1,1719 @@
...
+  },
+  {
+    class: 'CheckoutPaymentAdapter',
+    importPath: '@spartacus/checkout/core',
+    deprecatedNode: 'set',
+    comment:
+      '// TODO:Spartacus - Method "set" was removed from Class "CheckoutPaymentAdapter". ',
```
bgambocjaviniar: was renamed to setPaymentDetails

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/cart-lib-checkout.migration.ts
```diff
@@ -0,0 +1,1719 @@
...
+  },
+  {
+    class: 'CheckoutPaymentConnector',
+    importPath: '@spartacus/checkout/core',
+    deprecatedNode: 'create',
+    comment:
+      '// TODO:Spartacus - Method "create" was removed from Class "CheckoutPaymentConnector". ',
```
bgambocjaviniar: relevant connector and occ adapter for CheckoutPayment needs to be changed to rename of those values above

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/cart-lib-checkout.migration.ts
```diff
@@ -0,0 +1,1719 @@
...
+    deprecatedNode: 'set',
+    comment:
+      '// TODO:Spartacus - Method "set" was removed from Class "CheckoutPaymentConnector". ',
+  },
+  {
+    class: 'CheckoutPaymentService',
+    importPath: '@spartacus/checkout/core',
+    deprecatedNode: 'actionAllowed',
+    comment:
+      '// TODO:Spartacus - Method "actionAllowed" was removed from Class "CheckoutPaymentService". ',
```
bgambocjaviniar: renamed to checkoutPreconditions, which returns `Observable<[string, string]>` instead of boolean

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/cart-lib-checkout.migration.ts
```diff
@@ -0,0 +1,1719 @@
...
+  },
+  {
+    class: 'CheckoutPaymentService',
+    importPath: '@spartacus/checkout/core',
+    deprecatedNode: 'activeCartService',
+    comment:
+      '// TODO:Spartacus - Property "activeCartService" was removed from Class "CheckoutPaymentService". ',
```
bgambocjaviniar: was renamed to activeCartFacade

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/cart-lib-checkout.migration.ts
```diff
@@ -0,0 +1,1719 @@
...
+  },
+  {
+    class: 'CheckoutPaymentService',
+    importPath: '@spartacus/checkout/core',
+    deprecatedNode: 'checkoutService',
+    comment:
+      '// TODO:Spartacus - Property "checkoutService" was removed from Class "CheckoutPaymentService". ',
```
bgambocjaviniar: should be removed as it is the same as this (not included) https://github.com/SAP/spartacus/pull/15305/files#r821023502

this is a 'false' positive constructor param 'property' that was added around 4.3.x and pre-cart-lib commit

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/cart-lib-checkout.migration.ts
```diff
@@ -0,0 +1,1719 @@
...
+    deprecatedNode: 'checkoutStore',
+    comment:
+      '// TODO:Spartacus - Property "checkoutStore" was removed from Class "CheckoutPaymentService". ',
+  },
+  {
+    class: 'CheckoutPaymentService',
+    importPath: '@spartacus/checkout/core',
+    deprecatedNode: 'getPaymentDetails',
+    comment:
+      '// TODO:Spartacus - Method "getPaymentDetails" was removed from Class "CheckoutPaymentService". ',
```
bgambocjaviniar: possibly renamed to getPaymentDetailsState, which returns `Observable<QueryState<PaymentDetails | undefined>>` instead of `Observable<PaymentDetails>` 

Not sure if we want to say it was the 'same'

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/cart-lib-checkout.migration.ts
```diff
@@ -0,0 +1,1719 @@
...
+    deprecatedNode: 'loadCheckoutDetails',
+    comment:
+      '// TODO:Spartacus - PropertySignature "loadCheckoutDetails" was removed from Interface "CheckoutOccEndpoints". ',
+  },
+  {
+    class: 'CheckoutOccEndpoints',
+    importPath: '@spartacus/checkout/occ',
+    deprecatedNode: 'paymentTypes',
+    comment:
+      '// TODO:Spartacus - PropertySignature "paymentTypes" was removed from Interface "CheckoutOccEndpoints". ',
```
bgambocjaviniar: not sure what to say about this since it moved to the b2b  occ checkout b2b endpoints, which is 'new'

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/cart-lib-checkout.migration.ts
```diff
@@ -0,0 +1,1719 @@
...
+  },
+  {
+    class: 'CheckoutOccEndpoints',
+    importPath: '@spartacus/checkout/occ',
+    deprecatedNode: 'placeOrder',
+    comment:
+      '// TODO:Spartacus - PropertySignature "placeOrder" was removed from Interface "CheckoutOccEndpoints". ',
```
bgambocjaviniar: was moved to OrderOccEndpoints

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/cart-lib-checkout.migration.ts
bgambocjaviniar: similarly to other properties related to b2b / scheduled-replenishment from this interface

bgambocjaviniar (commented):
part 1* of this file

RESOLVED comments below because it is part of phase 2

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/cart-lib-checkout.migration.ts
```diff
@@ -0,0 +1,1719 @@
+import { MethodPropertyDeprecation } from 'projects/schematics/src/shared/utils/file-utils';
```
bgambocjaviniar: pipeline is complaining because of absolute path

Didn't take a look at the script, but why is it using absolute path, whereas the constructor deprecation PR one uses relative path

projects/schematics/src/migrations/5_0/removed-public-api-deprecations/cart-lib-checkout.migration.ts
```diff
@@ -0,0 +1,1522 @@
+import { DeprecatedNode } from 'projects/schematics/src/shared/utils/file-utils';
```
bgambocjaviniar: same as above

projects/schematics/src/migrations/5_0/removed-public-api-deprecations/removed-public-api-deprecations.ts
```diff
@@ -1,14 +1,15 @@
...
+  CONFIGURATOR_EVENT_LISTENER,
   HANDLE_VOUCHER_OPERATION_ERROR,
+  SPARTACUS_CORE,
+  SPARTACUS_PRODUCT_CONFIGURATOR_RULEBASED,
 } from '../../../shared/constants';
 import { DeprecatedNode } from '../../../shared/utils/file-utils';
 import { removedPublicApiDeprecation } from '../../mechanism/removed-public-api-deprecations/removed-public-api-deprecation';
+import { CART_LIB_CHECKOUT_REMOVED_PUBLIC_API_DATA } from './cart-lib-checkout.migration';
 
 export const REMOVED_PUBLIC_API_DATA: DeprecatedNode[] = [
```
bgambocjaviniar: Coming from the constructor deprecation PR, shouldn't we normalize the way it is generating the files, where we generate the migration and it will include all the existing + newly generated ones (all in one), and naming the variable `REMOVED_PUBLIC_API_DATA` --> depending on the respective migrations

Another example for `METHODS_AND_PROPERTIES_DEPRECATIONS_DATA` ---> for the methods-and-properties migration

etc

bgambocjaviniar (commented):
phase 2:

looking into the deleted members due to rename

specifically places in:
- methods-and-properties
- removed-public-api

another specific is for checkout to look into the renamed as some 'splits' the logic into separate checkout libs, and constructor

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/cart-lib-checkout.migration.ts
plabadie: This is my bad.  The script generates the array of migrations, but noot the rest.

projects/schematics/src/migrations/5_0/removed-public-api-deprecations/removed-public-api-deprecations.ts
plabadie: Agreed.  I'll rename it like we did for the constructor PR

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/cart-lib-checkout.migration.ts
plabadie: Fixed

projects/schematics/src/migrations/5_0/removed-public-api-deprecations/cart-lib-checkout.migration.ts
plabadie: fixed

projects/schematics/src/migrations/5_0/constructor-deprecations/constructor-deprecations.ts
```diff
@@ -1,13 +1,15 @@
...
+import { TAB_PARAGRAPH_CONTAINER_COMPONENT_MIGRATION } from './data/tab-paragraph-container.component.migration';
 
 export function migrate(): Rule {
   return (tree: Tree, context: SchematicContext) => {
-    return migrateConstructorDeprecation(
-      tree,
-      context,
-      CONSTRUCTOR_DEPRECATIONS_DATA
-    );
+    return migrateConstructorDeprecation(tree, context, [
```
bgambocjaviniar, line 9: I'm assuming this was changed manually for the sake of fixing the conflict from the breaking changes that were introduced recently, but I'm assuming we're going to re-generate this later instead of manually adding it to the array

bgambocjaviniar: Or we can just add them in the breaking-change.json manually before regenerating (assuming some changes are between 4.0.0 >= x <= 4.2.x

bgambocjaviniar: same for other  deprecations, if applicable

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/methods-and-properties-deprecations.ts
```diff
@@ -1,35 +1,17 @@
...
 export function migrate(): Rule {
   return (tree: Tree, context: SchematicContext) => {
-    return migrateMethodPropertiesDeprecation(
-      tree,
-      context,
-      METHODS_AND_PROPERTIES_DEPRECATIONS_DATA
-    );
+    return migrateMethodPropertiesDeprecation(tree, context, [
+      ...METHODS_AND_PROPERTIES_DEPRECATIONS_DATA,
+      ...CONFIGURATOR_ATTRIBUTE_MULTI_SELECTION_BUNDLE_COMPONENT_MIGRATION,
```
bgambocjaviniar: ConfiguratorAttributeMultiSelectionBundleComponent was caught by the breaking change.

Can we add a comment in the breaking-change.json like others to handle this case, so that we'll have everything inside `METHODS_AND_PROPERTIES_DEPRECATIONS_DATA`

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/methods-and-properties-deprecations.ts
```diff
@@ -1,35 +1,17 @@
...
+      ...CONFIGURATOR_ATTRIBUTE_SINGLE_SELECTION_BUNDLE_COMPONENT_MIGRATION,
```
bgambocjaviniar: same as above https://github.com/SAP/spartacus/pull/15307#discussion_r830739994

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/methods-and-properties-deprecations.ts
```diff
@@ -1,35 +1,17 @@
...
+      ...CONFIGURATOR_ATTRIBUTE_HEADER_COMPONENT_MIGRATION,
```
bgambocjaviniar: same as above as well https://github.com/SAP/spartacus/pull/15307#discussion_r830739994

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/methods-and-properties.migration.ts
```diff
@@ -0,0 +1,1790 @@
+import { MethodPropertyDeprecation } from '../../../../shared/utils/file-utils';
+
+export const METHODS_AND_PROPERTIES_DEPRECATIONS_DATA: MethodPropertyDeprecation[] =
+  [
+    {
+      class: 'QuickOrderFormComponent',
+      importPath: '@spartacus/cart/quick-order/components',
+      deprecatedNode: 'globalMessageService',
+      comment:
+        "// TODO:Spartacus - Property 'globalMessageService' was removed from Class 'QuickOrderFormComponent'. ",
```
bgambocjaviniar, line 38: question which would relate to other things here

in the sequence of our migrations, our constructor deprecations removed the dependency, but should we mention additionally that it was removed as a comment in methods-and-properties, where we mentioned the param is gone.

nothing important, but just a question that I have in mind.

bgambocjaviniar: My question came from thinking about constructors, where we add new dependency, therefore 'new' params, but we don't mention that

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/methods-and-properties.migration.ts
```diff
@@ -0,0 +1,1790 @@
...
+      class: 'AddToSavedCartComponent',
+      importPath: '@spartacus/cart/saved-cart/components',
+      deprecatedNode: 'activeCartService',
+      comment:
+        "// TODO:Spartacus - Property 'activeCartService' was removed from Class 'AddToSavedCartComponent'. ",
+    },
+    {
+      class: 'SavedCartAdapter',
+      importPath: '@spartacus/cart/saved-cart/core',
+      deprecatedNode: 'saveCart',
```
bgambocjaviniar, line 113: NOT related, but I always forget to ask

don't have to answer, but I'll leave it here for me to remember

Why did this move to the cart adapter as save cart is part of the Saved Cart controller in the backend, so I would guess to look at the saved cart domain in the frontend

projects/schematics/src/migrations/5_0/constructor-deprecations/constructor-deprecations.ts
plabadie: Let's discuss, not sure I understand.

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/methods-and-properties.migration.ts
plabadie: The constructor doc and schematics already advertise the new construtor params.

As for removed params iin construtors, they will show up as deleted properties in here as well.

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/methods-and-properties-deprecations.ts
plabadie: For now cases like these are present in the breaking change doc, but not in the migration schematics.

I was under the impression that methods and props schematics were meant to address deleted methods and props.  Let's discuss how we addrssess the cases of changed methods and props and if they need to be in the schematics after all.

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/methods-and-properties-deprecations.ts
plabadie: same as above :)

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/methods-and-properties-deprecations.ts
plabadie: same as above :)

projects/schematics/src/migrations/5_0/constructor-deprecations/constructor-deprecations.ts
bgambocjaviniar: got it, but in short 
```suggestion
    return migrateConstructorDeprecation
    (
    tree, 
    context, 
    CONSTRUCTOR_DEPRECATIONS_DATA
    );
```

for every deprecation, we want it to all be included in their respective auto generated file. In this case, inside `CONSTRUCTOR_DEPRECATIONS_DATA`. 

I was asking if this was temporarily to fix the 'conflicts' from people merging deprecations still as this PR is only for 'rest' and not constructor deprecation

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/methods-and-properties-deprecations.ts
bgambocjaviniar: to discuss, as I'm a bit confused here

bgambocjaviniar: to discuss~

I understand your concern as I saw the under the hood generation of this automated migration files even though you said to just look into the generated file instead But, I was interested so I took a bit of a peek :P 

All that being said, I believe it should cover additional added props to methods that is not optional, therefore it is required and a breaking change to 'old' users.

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/methods-and-properties-deprecations.ts
bgambocjaviniar: https://github.com/SAP/spartacus/pull/15307#discussion_r830739003

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/methods-and-properties.migration.ts
bgambocjaviniar: Yup I understand that, where the deleted properties are added here as well.

Since that is the case, shouldn't we mention the new properties that were added as well (which are the new dependencies for the class)

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/methods-and-properties.migration.ts
```diff
@@ -0,0 +1,1790 @@
...
+      comment:
+        "// TODO:Spartacus - Method 'saveCart' was removed from Class 'OccSavedCartAdapter'. ",
+    },
+    {
+      class: 'NewSavedCartOrderEntriesContext',
+      importPath: '@spartacus/cart/saved-cart/root',
+      deprecatedNode: 'actionsSubject',
+      comment:
+        "// TODO:Spartacus - Property 'actionsSubject' was removed from Class 'NewSavedCartOrderEntriesContext'. ",
```
bgambocjaviniar, line 171: Out of many cases, this is an example of the case I was talking about, where a property getting deleted due to the dependency to `ActionsSubject` being removed (which is good) https://github.com/SAP/spartacus/pull/15307#discussion_r830746461

However, the case I am mentioning is about also including in the migration saying `importInfoService` was added since a new dependency was introduced `ProductImportInfoService`

bgambocjaviniar: So it would be `// TODO:Spartacus - Property 'importInfoService' was added from Class 'NewSavedCartOrderEntriesContext'.`

Although, thinking about it now, it does not make sense mentioning these properties as they're not important in a constructor. The variable property can be named anything. 

All in all, should we mention constructor properties in the migrations?

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/methods-and-properties.migration.ts
```diff
@@ -0,0 +1,1790 @@
...
+      deprecatedNode: 'actionsSubject',
+      comment:
+        "// TODO:Spartacus - Property 'actionsSubject' was removed from Class 'SavedCartOrderEntriesContext'. ",
+    },
+    {
+      class: 'CartNotEmptyGuard',
+      importPath: '@spartacus/checkout/components',
+      deprecatedNode: 'activeCartService',
+      comment:
+        "// TODO:Spartacus - Property 'activeCartService' was removed from Class 'CartNotEmptyGuard'. ",
```
bgambocjaviniar, line 206: related to this question about whether we should even include constructor properties deletion/renamed/added to methods-and-properties. https://github.com/SAP/spartacus/pull/15307/files#r830749914

This case is a good example too as this was just renamed to activeCartFacade, but again, I would say it's not important to know what the property named changed to

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/methods-and-properties.migration.ts
```diff
@@ -0,0 +1,1790 @@
...
+    },
+    {
+      class: 'CheckoutAuthGuard',
```
bgambocjaviniar, line 209: case of the split checkout issue, but will be handled manually

bgambocjaviniar: won't be mentioning the other checkout things as we said it's manual things for Nik and I to do, for example, files like ShippingAddressComponent couldn't be compared to CheckoutDeliveryAddressComponent

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/methods-and-properties.migration.ts
```diff
@@ -0,0 +1,1790 @@
...
+    },
+    {
+      class: 'CheckoutProgressMobileTopComponent',
+      importPath: '@spartacus/checkout/components',
+      deprecatedNode: 'ngOnInit',
+      comment:
+        "// TODO:Spartacus - Method 'ngOnInit' was removed from Class 'CheckoutProgressMobileTopComponent'. ",
+    },
+    {
+      class: 'CheckoutStepsSetGuard',
```
bgambocjaviniar, line 293: Another case of split b2c and b2b logic for checkout

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/methods-and-properties.migration.ts
```diff
@@ -0,0 +1,1790 @@
...
+      deprecatedNode: 'checkoutDeliveryService',
+      comment:
+        "// TODO:Spartacus - Property 'checkoutDeliveryService' was removed from Class 'ExpressCheckoutService'. ",
+    },
+    {
+      class: 'ExpressCheckoutService',
+      importPath: '@spartacus/checkout/components',
+      deprecatedNode: 'checkoutDetailsService',
+      comment:
+        "// TODO:Spartacus - Property 'checkoutDetailsService' was removed from Class 'ExpressCheckoutService'. ",
```
bgambocjaviniar, line 346: won't mention some checkout things anymore as we agreed checkout will have to be checked manually and added to breaking-change.json, but last flag here to mention that it would be great to know what has been added as well

but again, maybe we're dropping constructor property changes for methods-and-properties (to discuss)

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/methods-and-properties.migration.ts
```diff
@@ -0,0 +1,1790 @@
...
+    },
+    {
+      class: 'CheckoutPaymentService',
+      importPath: '@spartacus/checkout/core',
+      deprecatedNode: 'resetSetPaymentDetailsProcess',
+      comment:
+        "// TODO:Spartacus - Method 'resetSetPaymentDetailsProcess' was removed from Class 'CheckoutPaymentService'. ",
+    },
+    {
+      class: 'CheckoutState',
```
bgambocjaviniar: another case of mentioning about new properties being added instead of removing all of these old ones from the CheckoutState interface

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/methods-and-properties.migration.ts
```diff
@@ -0,0 +1,1790 @@
...
+      importPath: '@spartacus/checkout/core',
+      deprecatedNode: 'steps',
+      comment:
+        "// TODO:Spartacus - PropertySignature 'steps' was removed from Interface 'CheckoutState'. ",
+    },
+    {
+      class: 'CheckoutOccEndpoints',
```
bgambocjaviniar: case of another split as we have b2b endpoints

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/methods-and-properties.migration.ts
```diff
@@ -0,0 +1,1790 @@
...
+      deprecatedNode: 'isEmail',
+      comment:
+        "// TODO:Spartacus - Method 'isEmail' was removed from Class 'ActiveCartService'. ",
+    },
+    {
+      class: 'ActiveCartService',
+      importPath: '@spartacus/core',
+      deprecatedNode: 'isEmpty',
+      comment:
+        "// TODO:Spartacus - Method 'isEmpty' was removed from Class 'ActiveCartService'. ",
```
bgambocjaviniar, line 801: will stop flagging things like this, but just mentioning that it was moved to utils.ts, so in phase 2, we'll add a comment to use isEmpty from utils.ts ( more of a reminder)

bgambocjaviniar (commented):
Part 1: will continue on, but stopped at SelectiveCartService in methods-and-properties (checkpoint for me to remember - line 1049 'class: UpdateUserAddressEvent')

methods-and-properties looking good to me other than checking manually again for checkout (Nik and I), and to check if things are really deleted, moved, or just renamed (manual step in phase 2)

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/methods-and-properties.migration.ts
```diff
@@ -0,0 +1,1790 @@
...
+    },
+    {
+      class: 'ReplenishmentOrderHistoryComponent',
+      importPath: '@spartacus/order/components',
+      deprecatedNode: 'userReplenishmentOrderService',
+      comment:
+        "// TODO:Spartacus - Property 'userReplenishmentOrderService' was removed from Class 'ReplenishmentOrderHistoryComponent'. ",
+    },
+    {
+      class: 'OrderAdapter',
```
bgambocjaviniar, line 1217: will most likely be checked with checkout as this is like a 'CheckoutFacade' for placing an order

projects/schematics/src/migrations/5_0/rename-symbol/rename-symbol.ts
```diff
@@ -1,10 +1,12 @@
 import { Rule, Tree } from '@angular-devkit/schematics';
 import { RenamedSymbol } from '../../../shared/utils/file-utils';
 import { migrateRenamedSymbols } from '../../mechanism/rename-symbol/rename-symbol';
+import { CART_LIB_CHECKOUT_RENAMED_SYMBOLS_DATA } from './cart-lib-checkout.migration';
 import { CHECKOUT_RENAMED_SYMBOLS_DATA } from './checkout-rename-symbol';
 
 export const RENAMED_SYMBOLS_DATA: RenamedSymbol[] = [
   ...CHECKOUT_RENAMED_SYMBOLS_DATA,
+  ...CART_LIB_CHECKOUT_RENAMED_SYMBOLS_DATA,
```
plabadie: Update constant naming to aling with others.

projects/schematics/src/migrations/5_0/constructor-deprecations/constructor-deprecations.ts
plabadie: The extra two were added after we ran the compare script.
They might go away in the future is we run the script again for the time period between now and the release date.

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/methods-and-properties.migration.ts
plabadie: The new properties (that might replace the deleted ones) are meant to be mentioned in the migration comment of the deleted property.
If it's just a new standalone property, it't not in scope of the migration and breaking change handling to list new things.

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/methods-and-properties.migration.ts
plabadie: It's handled by the constructor schematics.

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/methods-and-properties.migration.ts
plabadie: Let's figure out if the schematics support interfaces.
If not, we'll generate sschematics only for classes.

projects/schematics/src/migrations/5_0/rename-symbol/cart-lib-checkout.migration.ts
```diff
@@ -0,0 +1,1354 @@
...
+    newImportPath: '@spartacus/cart/wish-list/components',
+  },
+  {
+    previousNode: 'WishListItemComponent',
+    previousImportPath: '@spartacus/storefront',
+    newImportPath: '@spartacus/cart/wish-list/components',
+  },
+  {
+    previousNode: 'WishListModule',
+    previousImportPath: '@spartacus/storefront',
```
WeizhengSap: This `WishListModule` is different with the one in `@spartacus/cart/wish-list`. It is rename to `WishListComponentsModule` in `@spartacus/cart/wish-list/components`

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/methods-and-properties.migration.ts
```diff
@@ -0,0 +1,1790 @@
+import { MethodPropertyDeprecation } from '../../../../shared/utils/file-utils';
+
+export const METHODS_AND_PROPERTIES_DEPRECATIONS_DATA: MethodPropertyDeprecation[] =
+  [
+    {
+      class: 'QuickOrderFormComponent',
+      importPath: '@spartacus/cart/quick-order/components',
+      deprecatedNode: 'globalMessageService',
```
WeizhengSap, line 36: `globalMessageService` is constructor parameter. We already handle it in "constructor" deprecation. Do we also process it in "methods and properties" deprecation?

projects/schematics/src/migrations/5_0/rename-symbol/cart-lib-checkout.migration.ts
```diff
@@ -0,0 +1,1354 @@
...
+    newImportPath: '@spartacus/order/components',
+  },
+  {
+    previousNode: 'OrderHistoryModule',
+    previousImportPath: '@spartacus/storefront',
+    newImportPath: '@spartacus/order/components',
+  },
+  {
+    previousNode: 'OrderModule',
```
WeizhengSap: This module is rename to `OrderComponentsModule`, and in '@spartacus/order/components'

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/methods-and-properties.migration.ts
WeizhengSap: function `saveCart` is not only used in SavedCart domain, it also used in other places. So, that's why we move this function to `base`.

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/methods-and-properties.migration.ts
plabadie: Yes, I think so.  This migration schematic covers the use case of a class that extends QuickOrderFormComponent and uses the globalMessageService property in its code.

projects/schematics/src/migrations/5_0/rename-symbol/cart-lib-checkout.migration.ts
plabadie: Right, thanks!

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/methods-and-properties.migration.ts
plabadie: They don't support interfaces.  We'll restrict the schematics to only classes as this is what the schematics expect.

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/methods-and-properties-deprecations.ts
plabadie: The script now generates migration schematics for updated methods&props (their signature or type changed) in addition to the deleted methods and props.

projects/schematics/src/migrations/5_0/rename-symbol/rename-symbol.ts
plabadie: Naming fixed.

projects/schematics/src/migrations/5_0/rename-symbol/cart-lib-checkout.migration.ts
plabadie: created followup ticket #15448 to review these kind of cases.

projects/schematics/src/migrations/5_0/rename-symbol/cart-lib-checkout.migration.ts
plabadie: created followup ticket https://github.com/SAP/spartacus/issues/15448 to review these kind of cases.

tools/breaking-changes/generate-methods-props.ts
```diff
@@ -53,13 +54,37 @@ function getSchematicsData(breakingChange: any): any {
   schematicsData.class = breakingChange.apiElementName;
   schematicsData.importPath = breakingChange.entryPoint;
   schematicsData.deprecatedNode = breakingChange.deletedMember.name;
-  schematicsData.comment = `${breakingChange.deletedComment} ${breakingChange.migrationComment}`;
+  schematicsData.comment = getShematicsComment(breakingChange);
```
bgambocjaviniar:
```suggestion
schematicsData.comment = getSchematicsComment(breakingChange);
```

very minor --> getSch*

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/methods-and-properties.migration.ts
```diff
@@ -2,20 +2,66 @@ import { MethodPropertyDeprecation } from '../../../../shared/utils/file-utils';
...
+      importPath: '@spartacus/cart/quick-order/components',
+      deprecatedNode: 'activeCartService',
+      comment:
+        "The type of property 'activeCartService' changed to: 'activeCartService: ActiveCartFacade' ",
+    },
+    {
+      class: 'QuickOrderFormComponent',
+      importPath: '@spartacus/cart/quick-order/components',
+      deprecatedNode: 'cd',
+      comment: "The type of property 'cd' changed to: 'cd: ChangeDetectorRef' ",
```
bgambocjaviniar: not sure about these being thrown as comments. They didn't change, whereas changes like the above `QuickOrderComponent` is viable

bgambocjaviniar: Actually, my question would be more towards 

- should we flag the ones that became required instead of optional. I think yes, but the message is confusing

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/methods-and-properties.migration.ts
```diff
@@ -30,6 +76,13 @@ export const METHODS_AND_PROPERTIES_DEPRECATIONS_DATA: MethodPropertyDeprecation
       comment:
         "// TODO:Spartacus - Method 'search' was removed from Class 'QuickOrderService'. Use 'searchProducts' instead.",
     },
+    {
+      class: 'QuickOrderConfig',
```
bgambocjaviniar, line 83: cool we never had abstract class flagged before

MINOR: Maybe we should improve the generation of comments, where we replace the `\n` with a new line generation, or simper if the string contains '\n`, we replace it with a space

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/methods-and-properties.migration.ts
```diff
@@ -65,13 +118,34 @@ export const METHODS_AND_PROPERTIES_DEPRECATIONS_DATA: MethodPropertyDeprecation
...
       deprecatedNode: 'registerDeleteSavedCartEvents',
       comment:
         "// TODO:Spartacus - Method 'registerDeleteSavedCartEvents' was removed from Class 'SavedCartEventBuilder'. It was moved to 'CartEventBuilder', and was renamed to 'registerDeleteCart'.",
     },
+    {
+      class: 'SavedCartService',
+      importPath: '@spartacus/cart/saved-cart/core',
+      deprecatedNode: 'getSavedCart',
+      comment:
+        "The 'getSavedCart' method's signature changed to: 'getSavedCart( cartId: string): Observable<StateUtils.ProcessesLoaderState<Cart | undefined>>'",
```
bgambocjaviniar: Awesome for supporting this as I think we forgot about thinking of return types of methods as well
💯

bgambocjaviniar: I think something is wrong as it's being duplicated to the facade too. Line [178](https://github.com/SAP/spartacus/pull/15307/commits/593a52de4156a9d37749c651c2c97376cfbc23cb#diff-81358b4cd74a506d7e4c9f3b72e9b62f769b823eb7ea689ac7c42f363526bfddR178)

bgambocjaviniar: same for other service and proxy facade counterparts

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/methods-and-properties.migration.ts
```diff
@@ -590,6 +762,27 @@ export const METHODS_AND_PROPERTIES_DEPRECATIONS_DATA: MethodPropertyDeprecation
       comment:
         "// TODO:Spartacus - Property 'cartSelector$' was removed from Class 'ActiveCartService'. ",
     },
+    {
+      class: 'ActiveCartService',
+      importPath: '@spartacus/core',
+      deprecatedNode: 'getEntry',
+      comment:
+        "The 'getEntry' method's signature changed to: 'getEntry( productCode: string): Observable<OrderEntry | undefined>'",
```
bgambocjaviniar: Not sure why this isn't complaining like other service / proxy facade duplications

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/methods-and-properties.migration.ts
bgambocjaviniar: However, should we include configs? (didn't flag all configs)

Most of the time like this and other configs like GlobalMessageConfig includes only new properties, which are optional, therefore not a breaking change

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/methods-and-properties.migration.ts
```diff
@@ -688,13 +937,104 @@ export const METHODS_AND_PROPERTIES_DEPRECATIONS_DATA: MethodPropertyDeprecation
...
+      deprecatedNode: 'store',
+      comment:
+        "The type of property 'store' changed to: 'store: Store<StateWithProcess<void>>' ",
+    },
+    {
+      class: 'Command',
+      importPath: '@spartacus/core',
+      deprecatedNode: 'execute',
+      comment:
+        "The 'execute' method's signature changed to: 'execute( parameters: PARAMS): Observable<RESULT>'",
```
bgambocjaviniar: These are generics, and I don't think we should mention this (same for other commands in the file)

But how would we figure this out for automation

bgambocjaviniar: With above, maybe it's best to keep it then?

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/methods-and-properties.migration.ts
```diff
@@ -1423,6 +1980,41 @@ export const METHODS_AND_PROPERTIES_DEPRECATIONS_DATA: MethodPropertyDeprecation
       comment:
         "// TODO:Spartacus - Property 'userOrderService' was removed from Class 'OrderCancellationService'. ",
     },
+    {
+      class: 'OrderDetailItemsComponent',
+      importPath: '@spartacus/storefront',
+      deprecatedNode: 'cancel$',
+      comment:
+        "The type of property 'cancel$' changed to: 'cancel$: Observable<Consignment[] | undefined>' ",
```
bgambocjaviniar: nothing changed for these properties in this class `OrderDetailItemsComponent`

they're the same

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/methods-and-properties.migration.ts
```diff
@@ -1465,13 +2092,41 @@ export const METHODS_AND_PROPERTIES_DEPRECATIONS_DATA: MethodPropertyDeprecation
       comment:
         "// TODO:Spartacus - Property 'userReplenishmentOrderService' was removed from Class 'ReplenishmentOrderDetailsService'. ",
     },
+    {
+      class: 'ReplenishmentOrderHistoryComponent',
+      importPath: '@spartacus/storefront',
+      deprecatedNode: 'replenishmentOrders$',
+      comment:
+        "The type of property 'replenishmentOrders$' changed to: 'replenishmentOrders$: Observable<ReplenishmentOrderList | undefined>' ",
```
bgambocjaviniar: Another example of property type did not change

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/methods-and-properties.migration.ts
```diff
@@ -688,13 +937,104 @@ export const METHODS_AND_PROPERTIES_DEPRECATIONS_DATA: MethodPropertyDeprecation
...
+      deprecatedNode: 'load',
+      comment:
+        "The 'load' method's signature changed to: 'load( userId: string cartId: string): Observable<Cart | undefined>'",
+    },
+    {
+      class: 'OccCartVoucherAdapter',
+      importPath: '@spartacus/core',
+      deprecatedNode: 'getCartVoucherEndpoint',
+      comment:
+        "The 'getCartVoucherEndpoint' method's signature changed to: 'getCartVoucherEndpoint( userId: string cartId: string): string'",
```
bgambocjaviniar: not a problem, but to mention that this is a case of a method's param not having a type before, therefore was implicitly typed as any for the `cartId`, but now it is explicitly typed as a string.  --> but again we can keep it and not really a breaking change

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/methods-and-properties.migration.ts
```diff
@@ -1311,6 +1721,27 @@ export const METHODS_AND_PROPERTIES_DEPRECATIONS_DATA: MethodPropertyDeprecation
       comment:
         "// TODO:Spartacus - Property 'numberOfEntriesBeforeAdd' was removed from Class 'AddedToCartDialogComponent'. ",
     },
+    {
+      class: 'AddressBookComponent',
+      importPath: '@spartacus/storefront',
+      deprecatedNode: 'getCardContent',
+      comment:
+        "The 'getCardContent' method's signature changed to: 'getCardContent( address: Address): Observable<{ role: string; textBold: string; text: string[]; actions: { name: string; event: string; }[]; header: string; deleteMsg: string; label: string; }>'",
```
bgambocjaviniar: another case of implicit typing. Method from 4.3 and 5.0 does not have an explicit return type, but the returned value has additional properties.

I would say this is good that it caught it, but the previous case was different, where it wasn't really a problem.

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/methods-and-properties.migration.ts
```diff
@@ -1339,13 +1770,118 @@ export const METHODS_AND_PROPERTIES_DEPRECATIONS_DATA: MethodPropertyDeprecation
...
+      deprecatedNode: 'selectiveCartService',
+      comment:
+        "The type of property 'selectiveCartService' changed to: 'selectiveCartService: SelectiveCartFacade' ",
+    },
+    {
+      class: 'CartItemComponent',
+      importPath: '@spartacus/storefront',
+      deprecatedNode: 'isProductOutOfStock',
+      comment:
+        "The 'isProductOutOfStock' method's signature changed to: 'isProductOutOfStock( product: any): any'",
```
bgambocjaviniar: this didn't change at all

bgambocjaviniar (commented):
1 - noticed code cleanup was part of finalizing this initiative, so I took a quick peek on files for unused things

`tools/breaking-changes/compare.ts`
- memberTypes
- commented console logs
- getGenericCompareBreakingChange

`tools/breaking-changes/extract.ts`
- commented console.logs

`tools/breaking-changes/generate-const.ts`
- import execSync

`tools/breaking-changes/generate-deleted.ts`
- import execSync

`tools/breaking-changes/generate-doc.ts`
- getFullNewName
- import execSync

`tools/breaking-changes/generate-methods-props.ts`
- import execSync

`tools/breaking-changes/generate-moved.ts`
- import execSync

`tools/breaking-changes/generate-stats.ts`
- import execSync

`tools/breaking-changes/parse.ts`
- getCanonicalReference

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/methods-and-properties.migration.ts
bgambocjaviniar: related to comment https://github.com/SAP/spartacus/pull/15307#discussion_r835383778

as these services had duplicated counterpart to there Facade, whereas this one did not have a duplicated one (no duplicated thing found for ActiveCartFacade)

tools/breaking-changes/generate-methods-props.ts
plabadie: fixed

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/methods-and-properties.migration.ts
plabadie: I updated the comment to show the type before and after to make it clear to the reader what change happened to the property (a bit like what we generate in the md dock).

In this case it produces 
```
"The type of property 'cd: ChangeDetectorRef | undefined' changed to: 'cd: ChangeDetectorRef' ",
```

Adding this additional info does add more meaning to the comment.

As for making sure the change is really a breaking change, we can discuss which logic we could add to filter out changes that we are sure are not breaking.  Currently, any change to a property's type is considered potentially breaking.

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/methods-and-properties.migration.ts
plabadie: Good catch for `\n`.  I didn't think we would have some in property types.  I'll "sanitize" the property types..

plabadie: As for this kind of type change not being breaking, I don't see how we can catch this with a script.  Let's discuss.

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/methods-and-properties.migration.ts
plabadie: I don't understand what you mean by duplicated.  Yes, the service implements the facade, but in the end they are two different things and I think it's ok to document them both.  Let's discuss.

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/methods-and-properties.migration.ts
plabadie: I believe it's because ActiveCartFacade did not exist yet in 4.x

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/methods-and-properties.migration.ts
plabadie: If you look at the import path, the `OrderDetailItemsComponent` from  `@spartacus/storefront` is the source of the comparison.

With the new comment being generated, there is also more info provided about the alleged difference.:
```
    {
      class: 'OrderDetailItemsComponent',
      importPath: '@spartacus/storefront',
      deprecatedNode: 'cancel$',
      comment:
        "The type of property 'cancel$: Observable<Consignment[]>' changed to: 'cancel$: Observable<Consignment[] | undefined>' ",
    },

```

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/methods-and-properties.migration.ts
plabadie: Same explanation as for `OrderDetailItemsComponent`

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/methods-and-properties.migration.ts
plabadie: Correct.  Let's discuss if it's ok to leave those cases in.  We could document the fact that some changes in the list are "potential" breaking changes depending on how the code is used and extended.

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/methods-and-properties.migration.ts
plabadie: This one is tricky.  Let's discuss if it's still somewhat useful to list those cases.

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/methods-and-properties.migration.ts
plabadie: I agree.  According to the api extractor, the return type changed from boolean to any.
Again, this is a case of implicit function return types that causes a debatable result.

### Method isProductOutOfStock changed.


Previous version: 

```

isProductOutOfStock(
  product: any
): boolean

```


Current version: 

```

isProductOutOfStock(
  product: any
): any

```

plabadie: I validated the data and this result comes out of the api extractor output as is.

This class moved from storefrontlib to the cart library.  Could there be differences in the compilation options that result in a different interpretation from thee api extractor?

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/methods-and-properties.migration.ts
plabadie: We could review manually the results and make sure we review each Config class. I'll document the suggested process in the readme.

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/methods-and-properties.migration.ts
plabadie: These cases may or may not result in breaking changes.  In this specific occurrence, it might result in customer custom code having compilation errors.  (we decided to keep this.)

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/methods-and-properties.migration.ts
plabadie: We say it's good to have these cases documented. Not 100% sure it's a breaking change, but good to document regardless.

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/methods-and-properties.migration.ts
plabadie: Edge case,  ideally, we remove these cases from the breaking change list, but gif they make it into the migration doc it's ok.

plabadie: Will push a small PR to add a type on that method.  #15478

tools/breaking-changes/generate-methods-props.ts
```diff
@@ -105,13 +107,21 @@ function isMethodOrProperty(memberKind: string): boolean {
...
-function normalizeNewLines(doc: string) {
+function normalizeNewLinesForFunction(doc: string) {
   doc = doc.slice(1).slice(0, -1); // Remove first and last chars.  They are \n.
   doc = doc.replace(/\(\n/, '('); // Remove \n next to the opening parenthesis.
   doc = doc.replace(/\n\)/, ' )'); // Remove \n next to the closing parenthesis.
   return doc.replace(/\n/g, ','); // The rest of the \n are between params.  Replace with `,`
 }
+
+function normalizePropertyDoc(doc: string): string {
+  return normalizeNewLinesForProperty(doc).replace(/\s+/g, ' ').trim();
```
bgambocjaviniar: What do you think about

doc = doc.replace(/\n/g, '');
doc = doc.replace(/\s+/g, ' ').trim();

return doc;

tools/breaking-changes/readme.md
```diff
@@ -23,5 +27,53 @@ Run `yarn compare`.  This compares both ./src/*/public-api.json files to create
...
+Typically the documentation is placed in the folder `docs/migration`
+
+
+# Post generation tasks:
+
+## Manual review of all the deleted API element.
+
+Some cases requires manual review to complete.  The preferable way to deal with those is to update the data in `breaking-changes.json`.  This way, the migration assets can be re-generated to reflect the changes.
+
+The API elements that were detected as deleted by the script need human intervention.  Theere are different cases:
```
bgambocjaviniar: There instead of Theere

bgambocjaviniar: LGTM otherwise

tools/breaking-changes/compare.ts
```diff
@@ -5,7 +5,6 @@ import * as common from './common';
 // --------------------------------------------------
 // Main Logic
 // --------------------------------------------------
-const memberTypes = new Set();
```
bgambocjaviniar, line 8: getGenericCompareBreakingChange function is not being used as well

bgambocjaviniar (changes requested):

tools/breaking-changes/generate-methods-props.ts
plabadie: I agree.  This is more readable.

bgambocjaviniar (approved):

PR #15312, opened 2022-03-04
fix: fix export button align issue

### LINKED ISSUE #14535: Export button not aligned properly on order-confirmation page

kimhw0630:
fix to align export button with other components in order-confirmation page
closes #14535


feature-libs/cart/import-export/components/export-entries/export-order-entries.component.ts
```diff
@@ -15,6 +15,8 @@ import { ExportOrderEntriesToCsvService } from './export-order-entries-to-csv.se
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
 export class ExportOrderEntriesComponent {
+  @HostBinding('class') styles = 'container';
```
bit172, line 18: Is there a reason why you used hostbinding here instead of applying the styles directly?

kimhw0630: @bit172 
I don't think you can add a style class in the template for the host element ...
Do you mean like below in the component property?
image

Thanks

bit172: I meant that it's not really necessary to add a class `.container` if you can apply the style directly to `cx-export-order-entries`, unless there's a reason that I'm not aware of

kimhw0630: @bit172 
I used the existing "container"  class (I think this class coming from bootstrap) because its parent's children containers use this class  (e.g.cx-order-confirmation-items) so it should have the same behavior in different browser sizes.

bit172 (approved):

Issue #14955, opened 2022-02-05
Qty in added to cart modal does not match qty in cart if qty is greater than 9999


KateChuen:
**Environment Details**
Browser: Chrome
Device: mac
Frontend: develop branch on my local
Backend: 40.76.109.9

**Steps to Reproduce**
Login on electronics storefront
Go to any powershot A480 product, for eg product ID 1934795
Click on "Add to Cart" 
In the added-to-cart modal, change the qty to 100000

**Actual Result**
You'll see the qty in the qty box is still 100000, but by the total number of items in cart, only 9999 was added. 
If you go to cart, you'll see the qty next to the item is 9999. 

However, you can then go ahead and either, increase the qty by clicking on the + sign, or manually typing in 100000 again. 
<img width="849" alt="Screen Shot 2022-02-04 at 9 17 58 PM" src="https://user-images.githubusercontent.com/45409515/152625471-a99f3d3c-e53b-4784-8780-76028e7127fb.png">

<img width="919" alt="Screen Shot 2022-02-04 at 9 16 17 PM" src="https://user-images.githubusercontent.com/45409515/152625422-2342007f-8b37-460d-b557-ca2f22b3c4b4.png">

<img width="827" alt="Screen Shot 2022-02-04 at 9 21 08 PM" src="https://user-images.githubusercontent.com/45409515/152625447-8fa82717-dab1-4252-8d44-f1036a2fdb82.png">

**Expected Result**
As least the number in the qty box of the added-to-cart modal should be the same as the number in the cart. It should update the qty in the added to cart modal to 9999.

Xymmer:
investigate if related to spartacus/cartlib; if so, fix for 5.0.
if not (so it's occ api), create jira and add jira issue link to this ticket, and close this one and let Bill know.

there might be a ticket for 9999 issue already though
older issue: #13632 but maybe not same

KateChuen:
Tested on develop. 
- Now the qty in the added-to-cart modal corresponds to the qty in cart. 
- We can't add more than 9999 in both the added-to-cart modal and on the cart page. 
<img width="797" alt="Screen Shot 2022-02-28 at 11 19 13 AM" src="https://user-images.githubusercontent.com/45409515/156021400-d54a8892-77e3-4f75-ab46-eff9a382ef09.png">


[...]

PR #15314, opened 2022-03-04
chore: fix a bug in cart-item-list.component.ts

### LINKED ISSUE #15286: Input of item counter in SavedCartDetailsItems is disabled
**Describe the bug**
The middle section input of the item-counter in Saved Cart Details is disabled, but the quantity can be incremented/decremented with [-] [+] buttons.

**To Reproduce**
1. Create a shopping cart and save cart
2. Navigate to 'Saved Carts' under 'My Account'
3. Click on the saved cart
4. See error on item counter

**Expected behavior**
Uniform behaviour of item-counter  across features

**Screenshots**
Screen Shot 2022-03-02 at 4 07 40 PM

WeizhengSap:
close #15286


feature-libs/cart/base/components/cart-shared/cart-item-list/cart-item-list.component.ts
```diff
@@ -168,12 +168,18 @@ export class CartItemListComponent implements OnInit, OnDestroy {
...
-      const group = new FormGroup({
-        entryNumber: new FormControl(item.entryNumber),
-        quantity: new FormControl(item.quantity, { updateOn: 'blur' }),
-      });
-
-      this.form.setControl(controlName, group);
+      const control = this.form.get(controlName);
+      if (control) {
+        if (control.get('quantity')?.value !== item.quantity) {
+          control.patchValue({ quantity: item.quantity }, { emitEvent: false });
```
bgambocjaviniar, line 174: In what case do we ever land in here? I tried smoke testing this, but I can't find an instance on where this fires, except the unit tests due to direct manipulation of items.

However, I understand if we have an existing control, then let's just give the value to what the item currently have as quantity. But to me it's the same if we remove the if else and just create a new control, but again, this never happens on the app, but only in the unit tests

------------
Basically, I favor keeping what it was before, but changing 

this.form.setControl(controlName, group);

to 

this.form.addControl(controlName, group);
**as it was before**

feature-libs/cart/base/components/cart-shared/cart-item-list/cart-item-list.component.spec.ts
```diff
@@ -297,6 +297,17 @@ describe('CartItemListComponent', () => {
...
 
+    it('should update controls when quantity change', () => {
+      fixture.detectChanges();
+      const mockItem0Qty = mockItem0.quantity;
+      mockItem0.quantity = 20;
+      component.items = [mockItem0, mockItem1];
+      expect(
+        component.form.controls[mockItem0.entryNumber].get('quantity')?.value
+      ).toEqual(20);
+      mockItem0.quantity = mockItem0Qty;
```
bgambocjaviniar, line 308: 👍  I understand we want to set back the quantity to the same it was before to not cause any leak for other future assertions for people, but first I need clarify[ this question before approving](https://github.com/SAP/spartacus/pull/15314/files#r824142667) (we might even remove this if you agree)

feature-libs/cart/base/components/cart-shared/cart-item-list/cart-item-list.component.spec.ts
```diff
@@ -297,6 +297,17 @@ describe('CartItemListComponent', () => {
       ).toBeDefined();
     });
 
+    it('should update controls when quantity change', () => {
```
bgambocjaviniar, line 300: I believe this should be 'existing', but we might remove depending on the answer in https://github.com/SAP/spartacus/pull/15314/files#r824142667
```suggestion
    it('should update existing controls when quantity change', () => {
```

bgambocjaviniar (changes requested):
running all e2es* just to verify if saved cart got fixed for being able to focus and other e2es working as intended

Will revert previous commit once pipeline is done, but also change may be required depending on the answer related to https://github.com/SAP/spartacus/pull/15314/files#r824142667

EDIT: 

all e2es looking good 👍

feature-libs/cart/base/components/cart-shared/cart-item-list/cart-item-list.component.ts
WeizhengSap: we use `setControl` because of this ticket https://github.com/SAP/spartacus/issues/14955.  We have to set back the qty for the case that the counter greater than the limit.

bgambocjaviniar: quick call during the day as I need to verify something

bgambocjaviniar (approved):

PR #15332, opened 2022-03-09
feat: Add Video Component

### LINKED ISSUE #15213: Implement smart-edit video component

RadhepS:
Closes: https://github.com/SAP/spartacus/issues/15213


projects/storefrontlib/cms-components/content/video/video.component.ts
```diff
@@ -0,0 +1,150 @@
...
+
+  protected setMedia(
+    video?: CmsBannerComponentMedia,
+    media?: CmsBannerComponentMedia | CmsResponsiveBannerComponentMedia
+  ) {
+    if (video) {
+      this.source = this.mediaService.getMedia(video)?.src;
+    }
+
+    if (media) {
```
WeizhengSap: need to check whether `data.containerBackground === ContainerBackgroundOptions.UPLOAD_RESPONSIVE_IMAGE` as well

something like this:
```
if (
        data.containerBackground ===
          ContainerBackgroundOptions.UPLOAD_RESPONSIVE_IMAGE &&
        data.videoMedia
      ) {
        this.videoImage = this.mediaService.getMedia(
          data.videoMedia as MediaContainer
        );
      }
```

projects/storefrontlib/cms-components/content/video/video.component.ts
```diff
@@ -0,0 +1,150 @@
...
+          })
+      );
+    } else {
+      this.routerLink = this.getRouterLink(data);
+    }
+    this.cd.markForCheck();
+  }
+
+  protected getRouterLink(data: CmsVideoComponent): string | any[] | undefined {
+    if (data.url) {
```
WeizhengSap, line 79: `data.url` has high priority than `data.contentPage`. So, check `url` first, if it does not exist, then check `contentPage`, then `product` and `category`.

projects/storefrontlib/cms-components/content/video/video.component.ts
```diff
@@ -0,0 +1,150 @@
...
+            const pageLabel = page.label || '';
+            this.routerLink = this.urlService.transform({
+              cxRoute: pageLabel.substring(1),
+            });
+          })
+      );
+    } else {
+      this.routerLink = this.getRouterLink(data);
+    }
+    this.cd.markForCheck();
```
WeizhengSap: this one has to be inside `subscribe`, right?

projects/storefrontlib/cms-components/content/video/video.component.ts
```diff
@@ -0,0 +1,150 @@
...
+    protected component: CmsComponentData<CmsVideoComponent>,
+    protected mediaService: MediaService,
+    protected urlService: SemanticPathService,
+    protected cmsService: CmsService,
+    protected cd: ChangeDetectorRef,
+    protected renderer: Renderer2
+  ) {}
+
+  ngAfterViewChecked(): void {
+    this.renderer.setStyle(
```
WeizhengSap: why we have this? Can we just add `[height]` in `<video..>`?

projects/storefrontlib/cms-components/content/video/video.component.html
```diff
@@ -0,0 +1,25 @@
+<div
+  *ngIf="data$ | async as data; else loading"
+  class="video-container"
+  tabindex="-1"
```
WeizhengSap, line 4: what it is used for?

projects/storefrontlib/cms-components/content/video/video.component.html
```diff
@@ -0,0 +1,25 @@
...
+>
+  <a *ngIf="data.overlayTitle" [routerLink]="routerLink">{{
+    data.overlayTitle
+  }}</a>
```
WeizhengSap, line 8: need to confirm with UI devs, whether the link is added only to `overlayTitle` or to the whole video

projects/storefrontlib/cms-components/content/video/video.component.ts
RadhepS: Done

projects/storefrontlib/cms-components/content/video/video.component.ts
RadhepS: Done

projects/storefrontlib/cms-components/content/video/video.component.ts
RadhepS: Ya, changed

projects/storefrontlib/cms-components/content/video/video.component.ts
RadhepS: It's because [height] only supports the size in pixels. So we wouldn't be able to set 100% height using that attribute for example.

Maybe there's a better way, let me know if you think there is

projects/storefrontlib/cms-components/content/video/video.component.html
RadhepS: I'm not which one you're specifically referring to but:

Div - Used as container to help with positioning elements especially overlayTitle
loading - Used to display spinner if data is not ready (I saw this in other CMS components)
tabindex="-1" - Used to not make the div tabable so it is more seemless. (You would need to tab twice to get to the video element before, now it's just once)

projects/storefrontlib/cms-components/content/video/video.component.html
RadhepS: It was added to the overlayTitle because on Safari, you need to begin the video by clicking on the center of the screen which triggered the routing.

We will modify the player in the next iteration when we add closed captions

projects/storefrontstyles/scss/components/content/video/_video.scss
```diff
@@ -0,0 +1,27 @@
...
+      font-size: 3vw;
+    }
+    @include type('3');
+    position: absolute;
+    top: 2.5rem;
+    left: 2.5rem;
+    z-index: 1;
+    padding: 0.5rem;
+    margin-inline-end: var(--cx-margin-right, 2.5rem);
+    color: black;
```
morganm58: this color should be a variable. color: var(--cx-color-text);

projects/storefrontstyles/scss/components/content/video/_video.scss
```diff
@@ -0,0 +1,27 @@
...
+    background-color: rgba(150, 150, 150, 0.8);
+    border-radius: 0.2rem;
+    font-weight: bold;
```
morganm58: the font weight should also be a variable `font-weight: var(--cx-font-weight-bold);`

projects/storefrontlib/cms-components/content/video/video.component.ts
```diff
@@ -0,0 +1,146 @@
...
+  autoPlay: boolean;
+  loop: boolean;
+  mute: string | undefined;
+  height: string;
+  protected subscriptions = new Subscription();
+
+  data$: Observable<CmsVideoComponent> = this.component.data$.pipe(
+    tap((data) => {
+      this.styleClasses = data.styleClasses;
+      this.setMedia(data.video, data.videoMedia, data.containerBackground);
```
WeizhengSap: can we just pass `data` to these functions?

projects/storefrontlib/cms-components/content/video/video.component.ts
WeizhengSap: something like this:
```
 [style.height.%]="height === '100%' ? '100' : null"
 [style.height.px]="height !== '100%' ? height : null"
```
if we use this way, remove `px` from `${videoContainerHeight}px`. Or `height` can be either `FULL` or some value, then we check `height === 'FULL' ? 100: null`.

projects/storefrontstyles/scss/components/content/video/_video.scss
RadhepS: Done

projects/storefrontlib/cms-components/content/video/video.component.ts
RadhepS: I was debating on this too, changed

projects/storefrontstyles/scss/components/content/video/_video.scss
RadhepS: Done

projects/storefrontlib/cms-components/content/video/video.component.ts
RadhepS: Modified approach, let me know what you think.

I call setHeight() directly from the template. If there is no pixel height defined then default will be 100%

RadhepS: Also I didn't know about style.height.px/%, that's very cool and good to know! Thanks for that

WeizhengSap: that's fine. But the function name should be `getHeight` instead of `setHeight`, or maybe just use `get height` (use getter and setter).

projects/storefrontlib/cms-components/content/video/video.component.ts
```diff
@@ -0,0 +1,125 @@
...
+      });
+    } else if (data.category) {
+      this.routerLink = this.urlService.transform({
+        cxRoute: 'category',
+        params: { code: data.category },
+      });
+    }
+  }
+
+  protected setHeight(data: CmsVideoComponent) {
```
WeizhengSap: rename to `getHeight`, and add return type `string | undefined`. `else {return undefined}` can be removed.

RadhepS: Done

projects/storefrontlib/cms-components/content/video/video.component.html
```diff
@@ -0,0 +1,25 @@
...
+  <video
+    *ngIf="source"
+    [style.height.px]="getHeight(data)"
+    controls
+    [poster]="thumbnail?.src"
+    [loop]="loop"
+    [autoplay]="autoPlay"
+    [muted]="mute"
+    [attr.aria-label]="'player.label' | cxTranslate"
+  >
```
WeizhengSap, line 18: I tested, the poster is gone, and height is not set. where you set the default height `100%`?

RadhepS: We already talked about why the poster is gone.

 I updated the CSS to make the video player 100% by default.
Tested it to make sure, [style.height.px] overrides it when set

projects/storefrontlib/cms-components/content/video/video.component.ts
```diff
@@ -0,0 +1,124 @@
...
+        cxRoute: 'category',
+        params: { code: data.category },
+      });
+    }
+  }
+
+  getHeight(data: CmsVideoComponent): number | undefined {
+    if (
+      data.containerSize === ContainerSizeOptions.DEFINE_CONTAINER_HEIGHT &&
+      data.videoContainerHeight
```
WeizhengSap: no need check `data.videoContainerHeight`, just return it, because the return type can be `undefined`

projects/storefrontlib/cms-components/content/video/video.component.ts
```diff
@@ -0,0 +1,114 @@
...
+    this.autoPlay = data.autoPlay === 'true';
+    this.loop = data.loop === 'true';
+    this.mute = data.mute === 'true' ? 'muted' : undefined;
+  }
+
+  protected setRouting(data: CmsVideoComponent) {
+    if (data.url) {
+      this.routerLink = data.url;
+    } else if (data.contentPage) {
+      this.subscriptions.add(
```
WeizhengSap: if there is `take(1)`, `this.subscriptions` not required, right?

projects/storefrontlib/cms-components/content/video/video.component.ts
```diff
@@ -0,0 +1,114 @@
...
+  source: string | undefined;
+  thumbnail: Media | undefined;
+  routerLink: string | any[] | undefined;
+  autoPlay: boolean;
+  loop: boolean;
+  mute: string | undefined;
+  protected subscriptions = new Subscription();
+
+  data$: Observable<CmsVideoComponent> = this.component.data$.pipe(
+    tap((data) => {
```
WeizhengSap, line 40: maybe also add "distinctUntilChanged", so the "set" functions are not called multi times

projects/storefrontlib/cms-components/content/video/video.component.ts
```diff
@@ -0,0 +1,114 @@
...
+      this.subscriptions.add(
+        this.cmsService
+          .getPage({
+            id: data.contentPage,
+            type: PageType.CONTENT_PAGE,
+          })
+          .pipe(take(1))
+          .subscribe((page) => {
+            const pageLabel = page.label || '';
+            this.routerLink = this.urlService.transform({
```
WeizhengSap: "this.routerLink = pageLabel".

WeizhengSap (approved):

PR #15337, opened 2022-03-09
fix: Expose missing interceptors in public API (#12979)

### LINKED ISSUE #12979: Expose missing interceptors in public API
I confirmed with @WeizhengSap, there is no reason to hide interceptors in private api. Let's make them public for extensibility:
- [x] OccPersonalizationTimeInterceptor
- [x] OccPersonalizationIdInterceptor
- [x] CmsTicketInterceptor

- [x] Let's also move the private constant `const PERSONALIZATION_TIME_KEY = 'personalization-time';` to the body of the `OccPersonalizationTimeInterceptor` as a `protected readonly` property .
- [x] Similar for `const PERSONALIZATION_ID_KEY = 'personalization-id';` in `OccPersonalizationIdInterceptor`

kpawelczak:
Closes: #12979

### DIFF
--- a/feature-libs/smartedit/root/public_api.ts
+++ b/feature-libs/smartedit/root/public_api.ts
@@ -2,3 +2,4 @@ export * from './config/smart-edit-config';
 export * from './feature-name';
 export * from './services/smart-edit-launcher.service';
 export * from './smart-edit-root.module';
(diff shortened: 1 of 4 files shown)


bit172 (approved):
LGTM

PR #15338, opened 2022-03-09
fix: improve a11y on NVDA/JAWS for added to cart (#13581)

### LINKED ISSUE #13581: [needs assistive directive] JAWS/NVDA Add-to-cart needs meaningful vocalized feedback for screen reader users
**Describe the bug**
- If the user is on the results page and hits the add to cart button the dialog gets vocalized only as "updating cart" but there is no vocalization of:
- An item being successfully added or
- The item was already in the cart and the quantity was updated. 

- Library version: 4.0.0.

**To Reproduce**

2. Tab to any product category. E.g. "Film Cameras"
3. Tab to the first product.
4. Add product using the keyboard (Enter or Spacebar).
5. Listen to screen reader feedback.

**Expected behavior**
- Include success/warning message in the same context as the _"Updating cart"_ message.
- Screen reader vocalizes:  _"1 item has been added to your cart"_ or if the product is already in the cart...
- Screen reader vocalizes: _"The item was already in the cart. The quantity was updated."_ 
- Use: `aria-live="polite"` and `aria-atomic="true"`.

**Screenshots**

Screen Shot 2021-08-31 at 11 34 51 AM
Screen Shot 2021-08-31 at 11 32 58 AM

BenjaminEmiliani:
Closes #13581


Matejk00 (dismissed):

developpeurweb (approved):
- This works well for VO.
- NVDA alerts of a dialog opening and starts Read Mode after, so there is detection of a change of context and the narration of the dialog contents. 
- JAWS only alerts of a dialog opening, so it hints change of context, however, it doesn't start Read Mode automatically on dialog, so the user will have to manually explore the dialog to find out what content is about. We tried many approaches but it seems this is the way JAWS handles dialogs.
- Approving for now, and we'll wait for feedback from auditors on this one.

PR #15369, opened 2022-03-14
chore: Remove deprecated code and add EventService as required dependency to ComponentWrapperDirective and InnerComponentsHostDirective (#14960)

### LINKED ISSUE #14960: Make `EventService` a required dependency of `ComponentWrapperDirective`
Now `EventService` is an optional dependency ( with `?`). Moreover it has a decorator `@Optional()` (which is btw. redundant since the EventService is provided in the root injector).

We can make EventService a required dependency.

Moreover, we should also add `EventService` as a new required dependency of `InnerComponentsHostDirective` and pass it to the constructor of `new ComponentWrapperDirective(` in line 54.

Then we should add the migration constructor schematics for both classes `ComponentWrapperDirective` and `InnerComponentsHostDirective`

dydome:
close #14960


projects/schematics/src/migrations/5_0/constructor-deprecations/data/component-wrapper.directive.migration.ts
```diff
@@ -0,0 +1,61 @@
...
+    {
+      className: COMPONENT_HANDLER_SERVICE,
+      importPath: SPARTACUS_STOREFRONTLIB,
+    },
+    {
+      className: CMS_INJECTOR_SERVICE,
+      importPath: SPARTACUS_STOREFRONTLIB,
+    },
+    {
+      className: EVENT_SERVICE,
```
Zeyber: I think this would add the `EventService` a second time to the constructor.

Zeyber (changes requested):

Zeyber (dismissed):

Zeyber (dismissed):

szczepaniak-michal (dismissed):

szczepaniak-michal (approved):

PR #15373, opened 2022-03-15
Fix failing cypress tests

### LINKED ISSUE #15372: Fix failing cypress tests

Larisa-Staroverova:


projects/storefrontapp-e2e-cypress/cypress/helpers/product-configurator-cpq.ts
```diff
@@ -348,3 +349,12 @@ export function waitForProductCardsLoad(expectedLength: number) {
 export function checkSuccessMessageNotDisplayed(): void {
   globalMessage.getSuccessAlert().should('not.exist');
 }
+
+/**
+ * Searches for a product by a product name.
+ *
+ * @param {string} productName - Product name
+ */
+export function searchForProduct(productName: string): void {
```
ChristophHi, line 358: Why do we a CPQ specific method here?

ChristophHi (approved):

projects/storefrontapp-e2e-cypress/cypress/helpers/product-configurator-cpq.ts
Larisa-Staroverova: `product-configurator.ts` has its own `searchForProduct` method that does not work for CPQ. The reason is that one waits for a `GET` request that does not take place. That's why I decided to "override"  `searchForProduct` method for CPQ that does not contain waiting for a `GET` request.
For sure, one could import product search service directly and call `searchForProduct` method

PR #15375, opened 2022-03-15
Site freezes when logging out of ASM (#14914)

### LINKED ISSUE #14914: [repro] Site freezes when logging out of ASM
**Describe the bug**
At the moment of logging in as asm with any account and later, closing said session, if we try a search in the searchbox or go to another page we will see that we will not be able to do it.
If we look at the 'Network' tab of the DevTools we will see that no requests are executed, and it seems that in Redux, the navigations are canceled and the "Load Page Data" action never fails or completes successfully.

**Version of Spartacus**
- Library version: 3.4 but I was also able to reproduce the bug in the [Spartacus Demo](https://spartacus-demo.eastus.cloudapp.azure.com/electronics-spa/en/USD/)

**To Reproduce**
1. Login in ASM mode
2. Log out of ASM
3. Try performing a search in the searchbox or go to another page (e.g, the PDP of a product)

**Expected behavior**
That when you log out of ASM you can navigate correctly through the site

**Desktop:**
 - OS: Linux Ubuntu 20.04 LTS and Windows 10
 - Browser: Chrome, Firefox and Edge

kpawelczak:
Closes: #14914 

Changed logout guard, now it will prevent not logged-in users from entering the 'logout' route.


bit172 (approved):
Logging out from ASM doesn't freeze anymore. LGTM

integration-libs/cdc/root/guards/cdc-logout.guard.spec.ts
```diff
@@ -18,6 +18,10 @@ class MockAuthService implements Partial<AuthService> {
   coreLogout() {
     return Promise.resolve();
   }
+
+  isUserLoggedIn(): Observable<boolean> {
+    return of(true);
+  }
```
plabadie, line 24: This function does not seem used in the test.  Can it be removed? If some intended test coverage is missing instead, could we add the coverage?

Issue #15232, opened 2022-02-23
Spare part list filter value not cleared on navigation


sam-garland:
**Describe the bug**

When the user enters a value in the input rendered by `VisualPickingProductFilterComponent` the value is retained after navigation and is still present when the product details page for a different product is shown. The filter value should be cleared when navigation occurs.

**Tell us the version of Spartacus**
- 4.3

**To Reproduce**
1. Use a Spartacus site that has the `EPD Visualization` integration feature enabled
2. Navigate to a product that has `SPAREPART` product references
3. Type a value in the `VisualPickingProductFilterComponent` input
4. Navigate to a different product that has SPAREPART product references (such as by clicking a navigation node and clicking the link for a product in the search results)

**Observed behavior**

After navigating to a new product, the `VisualPickingProductFilterComponent` has the same filter value set that was set previously for a different product. The list is not filtered according to the value in the filter input.

**Expected behavior**

The `VisualPickingProductFilterComponent` is a live filter, so the set of items shown in the list and the opacity of the objects in the viewport should always correspond to the filter value shown in the input. 
The `filter` property of the `VisualPickingProductFilterService` should not persist after navigation and should be set to an empty string when a navigation occurs.

**Screenshots**
Filter entered for product CX704:
![image](https://user-images.githubusercontent.com/52560012/155400245-19ef9fc2-abb8-405f-8af4-03194e782a9c.png)

Same value in filter after navigating to CX704_2D (but filter not even applied):
![image](https://user-images.githubusercontent.com/52560012/155400478-ad564dc4-0cfb-4655-aba8-de687ed42230.png)

**Desktop (please complete the following information):**
 - OS: not relevent (but it was OSX)
 - Browser: not relevant (but it was Chrome)
 - Version: not relevant (but it was 98)

PR #15384, opened 2022-03-16
fix: spare part list filter input value not reset on navigation

sam-garland:
[Spare part list filter value not cleared on navigation #15232](https://github.com/SAP/spartacus/issues/15232)

- Sets the `filter` property of the VisualPickingProductFilterService to an empty string on navigation.
- Ensures that Observables used in the corresponding tests complete (to avoid interaction between tests caused by an Observable that outlives the test it relates to).


sam-garland:
The travis job is running in https://github.com/SAP/spartacus/pull/15383
No need for this PR.

Issue #14997, opened 2022-02-10
AMP validation errors on Results Page


developpeurweb:
**Describe the bug**
For some reason, this bug reappeared.

- A label needs to be associated with the combobox, also the combobox is missing some aria attributes.

**Tell us the version of Spartacus**
- Library version: develop (5.0)

**Screershots**

![Screen Shot 2022-02-09 at 7 41 57 PM](https://user-images.githubusercontent.com/15113219/153506096-c6cfeb0f-8d67-4b86-9155-a79743b98cc9.png)
![Screen Shot 2022-02-09 at 7 42 24 PM](https://user-images.githubusercontent.com/15113219/153506114-1bb0761e-0ff5-43c1-9cbc-b53ab370172d.png)

BenjaminEmiliani:
Discussion with: @developpeurweb @Dalvir-Personal and @hackergil concluded as follows. 

Ticket is on hold temporarily. Solutions have been explored and tested. Those that solve the error require a significant change to the code of the application in many locations. Reason for this:

Error reported by Access Assistant is on an inner element of `ng-select` (Angular component) not exposed in the API. So we would need to bind specific aria-attributes to each instance of `ng-select` or wrap `ng-select` in a custom cx element  and replace all instances of `ng-select` in the application. 

Both these options require a fair amount of time and do not yield added value to the user experience of accessibility users, thus, the implementation of such solution would be for the sake of dismissing an automated reported error that is exemptible by manual testing.

developpeurweb:
For the record. We have been in this component before, GH-13433, with different errors.

KateChuen:
Tested on develop branch of spartacus using yarn
More extensive testing (on other screenreaders and devices) will be done with RC.

<img width="787" alt="Screen Shot 2022-04-06 at 4 11 28 PM" src="https://user-images.githubusercontent.com/45409515/162224807-09198315-015c-45af-8532-41f905e3214c.png">

<img width="1374" alt="Screen Shot 2022-04-06 at 4 12 40 PM" src="https://user-images.githubusercontent.com/45409515/162224841-e5ba65ee-3f13-405c-acef-b38e8e02354d.png">

PR #15408, opened 2022-03-17
fix: add aria-controls to sorting components (#15401)

### LINKED ISSUE #15401: Add ariaControls input to SortingComponent instances
Recent change (#14997) added a directive to handle aria-label and aria-controls for ng-select to comply with SAP AMP. `<cx-sorting>` element uses ng-select and we need to pass a new input ariaControls to the sorting component selector.

BenjaminEmiliani:
Closes #15401
In previous PR, #14997, we added the ng-select a11y to the cx-sorting component. This component is reusable and requires a custom input for ariaControls.


FollowTheFlo (dismissed):
Tested all of them with amp and SR except order approvals and order returns, 
no AMP violations concerning cx-sorting.
No change of behaviour on SR which is expected.
Code is correct to me with safety nets for null values, covers all scenarios i could think.

FollowTheFlo (approved):

PR #15410, opened 2022-03-17
chore: cx-media gives error when used with sticktTemplates enabled

### LINKED ISSUE #14037: cx-media gives error when [container] prop is used with strictTemplate: true
**Describe the bug**
I'm using cx-media component to show some product images. In the source code of Spartacus there's examples with cx-media which is accepting images in the [container] property from the default data model of the Product.
But in the newest version of Spartacus with Angular 12 the strictTemplate property is enabled by default (which is good), but when I pass something like this to the cx-media component:
` <cx-media [container]="product.images?.PRIMARY" format="product"></cx-media>`

I get a complain from the compiler:
`Type 'ImageGroup | ImageGroup[]' is not assignable to type 'MediaContainer'.`

This is a bummer, because usually the cx-media component is used very often with product images and the strictTemplate option is usually really helpful so I don't want to disable it.

- Library version: 4.1.1

**To Reproduce**
1. Create new project with the latest Spartacus libraries
2. Use cx-media component in an example component somewhere in the application
3. Get current product and assign product.images.PRIMARY to the [container] property of the cx-media component (make sure that the strictTemplate flag is set to true).

**Expected behavior**
The container property of cx-media component should accept types of ImageGroup | ImageGroup[] or there should be other property which accepts ImageGroup | ImageGroup[].

**Screenshots**
image

 - OS: Windows 10
 - Browser: Chrome
 - Version:  latest version

WeizhengSap:
close #14037


projects/storefrontlib/shared/components/media/media.component.ts
```diff
@@ -91,7 +97,7 @@ export class MediaComponent implements OnChanges {
    */
   protected create(): void {
     this.media = this.mediaService.getMedia(
-      this.container,
+      this.container instanceof Array ? this.container[0] : this.container,
```
bgambocjaviniar, line 100: After the DM, it works as expected in strict typing and allowing it to pass for the use case of the ticket.

However, I forgot to ask about if it was indeed an image group array, then why does it have to be the first imagegroup?

projects/storefrontlib/shared/components/media/media.service.ts
```diff
@@ -38,7 +38,7 @@ export class MediaService {
    * for specific formats.
    */
   getMedia(
-    mediaContainer: MediaContainer | Image,
+    mediaContainer?: MediaContainer | Image,
```
bgambocjaviniar, line 41: doesn't hurt to make it optional, but do we need to? I think we can keep it as a required param for getMedia

I don't think there's an instance, where we do not pass a media container. Moreover, we're checking if it's undefined in the end in [line 46](https://github.com/SAP/spartacus/pull/15410/files#diff-b2b4a1ad47f7561a671934b3afed849d7676a290627693fa3c558fe0aeab1e9bR46)

bgambocjaviniar, line 41: doesn't hurt to make it optional, but do we need to? I understand we can now explicitly pass undefined to the `input container` but I think we can keep it as required for the getMedia function in order for customers to always atleast pass a media. What do you think?

Reason why I'm saying this is ebcause in [line 46](https://github.com/SAP/spartacus/pull/15410/files#diff-b2b4a1ad47f7561a671934b3afed849d7676a290627693fa3c558fe0aeab1e9bR46) we're getting out of the function if it's undefined.

WeizhengSap: This is called from `MediaComponent`, which `contaner` (which is an input) can be `undefined`. Also, in our `interface` of cms models, all properties are optional (`?`). So, if make this required, we have to force the property passed is not optional.

projects/storefrontlib/shared/components/media/media.component.ts
WeizhengSap: The `container` should never been an array. We make `container` as `ImageGroup | ImageGroup[]`, because product's `PRIMARY` is `ImageGroup` and `GALLERY` is `ImageGroup[]`. We don't separate `PRIMARY` or `GALLERY`. But, there is no use case, we pass `GALLERY` list to it. (We can pass one item in `GALLAERY` to it). So, if the list is passed (should never happen), then we only process the first one (it is the default one, which should be the same to PRIMARY)

bgambocjaviniar (approved):

PR #15423, opened 2022-03-21
chore: Fix arrow navigation in quick order form with screen reader enable (#15308)

### LINKED ISSUE #15308: Arrow navigation not working in Quick Order search with Chrome JAWS
**Describe the bug**
If we want to search for a product in the main search once we type 3 letters we can see the list with suggestions, after hitting the arrow down on the keyboard we can enter the list and navigate to the first element on the list. The same situation is in the quick order feature on the list page. This is working fine without and for most screen readers, except for **Chrome+JAWS** where arrow navigation works to reach the item but `Enter` or `Spacebar` won't work to trigger selection. However, it works for Firefox+JAWS.

- This bug was found as a result of this effort GH-15233
- A **solution** to a similar issue was found in GH-13360

- Library version: 4.3

**To Reproduce**
1. Go to the Quick Order using Chrome for Windows
2. Enable JAWS
3. Tab to search box
4. Write 3 random characters
5. Wait until the suggestion list appears
6. Click the arrow down on your keyboard
7. Try to add a product by using `Enter` or `Spacebar`.

**Expected behavior**
Should enter the list and select any element on the list.

dydome:
close #15308


Matejk00 (approved):

FollowTheFlo (approved):
code change is fine
tested on JAWS and NVDA (ff and chrome) successfully

developpeurweb (approved):
Works well with JAWS/NVDA/VO.

PR #15427, opened 2022-03-21
test: E2E Tests for OAuth code and implicit authentication flows

### LINKED ISSUE #9445: E2E tests for auth changes

Zeyber:
Closes: #9445


Zeyber:
Redirects from the authentication server do not seem to be working in Cypress. We can see the redirect response `302` in the network traffic `https://40.76.109.9:9002/authorizationserver/oauth/authorize?response_type=token&client_id=client4kyma&state=YzFHSFNuSTFGbFpaMWZJMGJvSEpsT0V-b3F2OHdvVGZZQnU3OW1lMVNHVU5n&redirect_uri=https%3A%2F%2Flocalhost%3A4200&scope=openId`. But the redirection is not done and the test stops on the external page.

image


A workaround could be to redirect based on the returned `redirect_uri`. However, we can't seem to see this request from the test suite (possibly related to https://github.com/cypress-io/cypress/issues/3190) so we can't get the uri.

PR #15428, opened 2022-03-21
replaced defaultGlobalMessageConfigFactory (#13571)

### LINKED ISSUE #13571: Verify is defaultGlobalMessageConfigFactory still needed
We were forced to using the factory for configuration with TS dynamic keys, but probably in new Angular, it should work for standard configuration too.
- [x] verify is it possible to use `provideDefaultConfig` for this case
if it works:
- [x] replace factory by standard config for global messages
- [x] update docs for global messages

@Optional()
- [ ] create ticket for other occurences

kpawelczak:
Closes: #13571


bit172 (changes requested):
~~It works but since it is a breaking change, it should be documented.~~
It should be fine since it does not involve public apis

bit172 (dismissed):

bit172 (approved):

Issue #14604, opened 2021-12-13
Fix refreshing of customized price


marek-sap:
Refreshing of customized prices is not working in Spartacus.
I have created customized prices for products based on:
https://help.sap.com/viewer/4c33bf189ab9409e84e589295c36d96e/2105/en-US/8c5af47f8669101488e59df61b2f6ab4.html
Then i assigned those customizations to my testing user.

In Spartacus storefront after login to my testing user, price didn't changed.
After refreshing the page, price changed to customized value.
Similar situation is for logout event => price is not changing after logout. I need to refresh the page to get default price.

It seems that login and logout events are not refreshing prices.

Steps to reproduce:
1. Create some customized prices based on: https://help.sap.com/viewer/4c33bf189ab9409e84e589295c36d96e/2105/en-US/8c5af47f8669101488e59df61b2f6ab4.html
and assign them to customer (testing user).
2. Go to Spartacus storefront
3. Navigate to product for which you customized price. (You should see default price)
4. Login to you testing user (for which customized price should be applied)
5. After successful login product price still have default value (It should be changed after login)
6. Reload the page, now you should see customized price.

Xymmer:
Anette is asking that this be backported to 4.x as well a fixed for 5.0
let's evaluate the work needed and where the problem lies so we can fit it in.
could be: too fast, spartacus using cache, or occ api returning cached value.

WeizhengSap:
To QA:
1. enable b2b, choose one product from powertools (e.g. `3755219` which is also a product in carousel in homepage)
2. go to backoffice, create a new price row (set product to `3755219`, set customer to some b2b customer, and set Price to `10`)
![Screen Shot 2022-03-23 at 3 00 10 PM](https://user-images.githubusercontent.com/44440575/159765681-b884004f-272e-4f19-a773-215b38f069d4.png)
)
3. in Spartacus, check the price of product `3755219`. For login user, it is `10$`, for anonymous user, it is `48$`. (Check it in homepage, and pdp both)

Xymmer:
to confirm, we are putting this in 5.0 and backporting to 4.x when approved

KateChuen:
Tested on develop branch, using yarn, pointing to dev7 backend.

- Verified that specific user (gi.sun) sees the customized price (18.18) both on pdp and homepage.
[...]

PR #15441, opened 2022-03-23
chore: Fix refreshing of customized price

WeizhengSap:
#14604


projects/core/src/state/utils/entity/entity.action.ts
```diff
@@ -5,11 +5,11 @@ export const ENTITY_REMOVE_ALL_ACTION = '[ENTITY] REMOVE ALL';
 
 export interface EntityMeta {
   entityType: string;
-  entityId: string | string[];
+  entityId?: string | string[];
```
plabadie, line 8: Why does this field ( and further down the road function param) has to be now optional?

Is it because of this call https://github.com/SAP/spartacus/pull/15441/files#r843298544

projects/core/src/occ/adapters/product/default-occ-product-config.ts
```diff
@@ -12,6 +12,7 @@ export const defaultOccProductConfig: OccConfig = {
           details:
             'products/${productCode}?fields=averageRating,stock(DEFAULT),description,availableForPickup,code,url,price(DEFAULT),numberOfReviews,manufacturer,categories(FULL),priceRange,multidimensional,tags,images(FULL)',
           attributes: 'products/${productCode}?fields=classifications',
+          price: 'products/${productCode}?fields=price(formattedValue)',
```
plabadie, line 15: Does the fact that the price is ini 2 scopes affect the way we want it to refresh?

projects/core/src/product/store/actions/product.action.ts
```diff
@@ -38,5 +40,16 @@ export class LoadProductSuccess extends EntityScopedLoaderActions.EntityScopedSu
   }
 }
 
+export class ClearProductPrice extends EntityScopedLoaderActions.EntityScopedResetAction {
+  readonly type = CLEAR_PRODUCT_PRICE;
+  constructor() {
+    super(PRODUCT_DETAIL_ENTITY, undefined, ProductScope.PRICE);
```
plabadie, line 46: What's the undefined field here?   Is int because we want to clear the price of all the products and not the price of a specific product?

plabadie: What is the expected behaviour of `EntityScopedLoaderActions.EntityScopedResetAction`?  Does it simply erase the data from the store, or does it refresh thee scope data with new up to date data by doing new backend calls?

projects/core/src/occ/adapters/product/default-occ-product-config.ts
plabadie, line 15: Having taken a look at the solution, one question comes to mind.  the price was already in 2 of thee existing scopes.  Could we have fixed the issue without creating a new scope that contains just the price?  What is the added value of creating a new scope in this case?

projects/core/src/product/store/effects/product.effect.ts
```diff
@@ -71,6 +72,12 @@ export class ProductEffects {
     );
   }
 
+  @Effect()
+  clearProductPrice$: Observable<ProductActions.ClearProductPrice> = this.actions$.pipe(
+    ofType(AuthActions.LOGOUT, AuthActions.LOGIN),
+    map(() => new ProductActions.ClearProductPrice())
+  );
+
```
plabadie, line 80: Instead of a new effect/action/scope, could we simply have use a reloading trigger in the existing scopes that carry the price?  https://sap.github.io/spartacus-docs/loading-scopes/#reloading-triggers
Assuming there are login/logout events we could use for that.  So something like:

```
{
  backend: {
    loadingScopes: {
      product: {
        details: {
          reloadOn: [LoginEvent, LogoutEvent]
        },
        list: {
          reloadOn: [LoginEvent, LogoutEvent]
        }
      },
    },
  },
}
```

projects/core/src/state/utils/entity/entity.action.ts
WeizhengSap: No. It should be optional. Because in `entity.reducer.ts` there is check `action.meta.entityId !== undefined` (line 26)

projects/core/src/product/store/actions/product.action.ts
WeizhengSap: yes, this `undefined` will clear all scoped data (in our case, prices of all products). It is designed as this.
`EntityScopedLoaderActions.EntityScopedResetAction` erase data from store, no refresh.

projects/core/src/occ/adapters/product/default-occ-product-config.ts
WeizhengSap: `default` and `details` both are too heavy. We only want to update `price`. So, that's why this light scope is created

projects/core/src/product/store/effects/product.effect.ts
WeizhengSap: This is good! But we will reloadOn `price`.

projects/core/src/product/store/actions/product.action.ts
plabadie: Do you know whih mechanism reefreshes the data in this case?

projects/core/src/product/store/effects/product.effect.ts
plabadie: We tried `reloadOn` and it doesn't work on the login page.  It seems this feature needs active subscriptions on the data for it to refresh.  This is not a limitation we can live with for this use case.

plabadie (dismissed):

plabadie (approved):

PR #15444, opened 2022-03-24
feat: add stock level label in merchandising carousel

marek-sap:


integration-libs/cds/src/merchandising/cms-components/merchandising-carousel/merchandising-carousel.component.html
```diff
@@ -37,6 +37,12 @@ <h4>{{ item.name }}</h4>
       <div class="price">
         {{ item.price?.formattedValue }}
       </div>
+      <div class="price" *ngIf="item.stock?.stockLevel>0">
+        {{ item.stock?.stockLevelStatus }} : {{ item.stock?.stockLevel }}
+      </div>
+      <div class="price" *ngIf="item.stock?.stockLevel===undefined || item.stock?.stockLevel<=0">
```
bgambocjaviniar: 👍 

We can keep like this, but what do you think of wrapping it in an `else` clause instead.

For example,

```
      <div class="price" *ngIf="item.stock?.stockLevel > 0; else noStocks">
        {{ item.stock?.stockLevelStatus }} : {{ item.stock?.stockLevel }}
      </div>
      <ng-template #noStocks>
        <div class="price">
          {{ item.stock?.stockLevelStatus }}
        </div>
      </ng-template>
```

bgambocjaviniar: you should run `yarn prettifer:fix` to fix the formatter of this template as it will throw an error in the CI

bgambocjaviniar (changes requested):
If there's no need to change the template to what was suggested, feel free to ping me, and i'll approve after prettifying your files

AND

to fix the your unit tests
- you can run `ng test cds` to verify your unit tests

bgambocjaviniar (approved):

PR #15475, opened 2022-03-29
fix: include and fix broken unit tests for b2b .get() entity when it's not in store

MarcinGanczorz:


MarcinGanczorz:
i'll re-crate it from branch instead of fork

PR #15491, opened 2022-03-30
fix: Support redirects to protected pages from external authenticators

### LINKED ISSUE #13622: Support redirects from external auth pages to pages with AuthGuard
When we implemented auth in Spartacus we didn't assumed that there will be a case when we redirect from auth provider page to the page that is protected with AuthGuard.

When this case happens and we use Authorization Code Flow that require a bit more time to finish (it does a call with code to exchange it for a token) the guards might be invoked too quickly (before we get the token).

**How we should behave in this situation?**

We should have some flag that says if the login is in progress. If so we should wait in AuthGuard for it to finish first before me make a decision to redirect user to login page.

Zeyber:
Closes: #13622


projects/core/src/auth/user-auth/services/oauth-lib-wrapper.service.ts
```diff
@@ -16,6 +16,11 @@ import { AuthConfigService } from './auth-config.service';
 export class OAuthLibWrapperService {
   events$: Observable<OAuthEvent> = this.oAuthService.events;
 
+  /**
+   * Returns true if the `tryLogin()` call has not completed.
+   */
+  loginInProgress$: Observable<boolean> = new BehaviorSubject<boolean>(false);
```
Pucek9: You needn't define an explicit type of property if it's easily inferred by TS. Also calling a method with false causes the type (boolean) to be easily pressed.
image

In this case, you tricked TS that it's an `Observable` by unnecessary type narrowing, making you unable to explicitly refer to the `next` method (which occurs in `BehaviorSubject`, but not in `Observable`) if you have to use an assertion somewhere, there is most likely an oversight somewhere.

Pucek9 (approved):

projects/core/src/auth/user-auth/services/oauth-lib-wrapper.service.ts
```diff
@@ -16,6 +16,11 @@ import { AuthConfigService } from './auth-config.service';
 export class OAuthLibWrapperService {
   events$: Observable<OAuthEvent> = this.oAuthService.events;
 
+  /**
+   * Returns true if the `tryLogin()` call has not completed.
+   */
+  loginInProgress$ = new BehaviorSubject(false);
```
Pucek9:
```suggestion
  protected loginInProgress$ = new BehaviorSubject(false);

  isLoginInProgress(): Observable<boolean> { return this.loginInProgress$.asObservable(); }
```

Pucek9 (changes requested):
If we dont want to return explicite BehaviorSubject, let's return .asObservable() instead

Pucek9 (approved):
I was not able to confirm improvement by QA because it's hard to reproduce the issue on develop.

projects/core/src/auth/user-auth/facade/auth.service.ts
```diff
@@ -113,8 +113,13 @@ export class AuthService {
    * Returns `true` if the user is logged in; and `false` if the user is anonymous.
    */
   isUserLoggedIn(): Observable<boolean> {
-    return this.authStorageService.getToken().pipe(
-      map((userToken) => Boolean(userToken?.access_token)),
+    return combineLatest([
+      this.authStorageService.getToken(),
+      this.oAuthLibWrapperService.isLoginInProgress(),
```
Platonn, line 118: Did you consider any other alternative solutions?

Now `isUserLoggedIn()` indirectly depends now on calling `tryLogin()` specifically.

Nitpicks on details:
- Are there any other cases (besides invoking `tryLogin()`) when the “login process is in progress” and we should allow for observing it via `loginInProgress$`? Currently it’s bound only to the custom implementation of `tryLogin()`. And what about other processes when “login is in progress”? I don’t know TBH, maybe calling another method of `OAuthService`? e.g. `tryLoginWithBlahBlah` etc. or calling the OCC endpoint for login?
- should `isUserLoggedIn()` return false when the “login is in progress”? Or should it hold emitting any values until the login is in progress (e.g. use rxjs `filter("...login is in progress...")`)? And if we hold emiting values, what side effects might it cause? 

The above nitpicks are not so important, they just refer to the current implementation. I’m more interested in whether you considered any alternative approaches in general.
I have a gut feeling the with current approach we’re fixing the problem near the symptoms, but don’t in the roots. But I might be wrong 😉 
Anyway, I’m happy to understand the roots of the problem.

Platonn (changes requested):
I'd be happy to understand the roots of the problem so to see if there are any possible alternative solutions.

As we agreed offline, I'm setting status "Request changes" so it's not merged accidentally.

PR #15497, opened 2022-03-30
feat: Added a note when editing a root unit (#8519)

### LINKED ISSUE #8519: Display note when editing root unit, about not changing parent
When editing root unit, the parent unit field is not displayed.
That's fine, but add a note:

"This unit is the top-level unit for your organization. It cannot be assigned to another unit."

kimhw0630:
add a note when editing a root unit.
image

Closes #8519


feature-libs/organization/administration/components/unit/form/unit-form.component.html
```diff
@@ -48,26 +48,32 @@
         [control]="form.get('approvalProcess.code')"
       ></cx-form-errors>
     </label>
-
-    <label *ngIf="form.get('parentOrgUnit')" formGroupName="parentOrgUnit">
-      <span class="label-content required">{{
-        'orgUnit.form.parentOrgUnit' | cxTranslate
+    <ng-container *ngIf="form.get('parentOrgUnit'); else parentOrgUnitNotes">
+      <label *ngIf="form.get('parentOrgUnit')" formGroupName="parentOrgUnit">
```
Pucek9: It's unnecessary to check conditions inside the block if it was already checked upstream.
```suggestion
      <label formGroupName="parentOrgUnit">
```

Pucek9:
What do you think about setting e.g. `margin-top: 1rem;` for this new span?
image

feature-libs/organization/administration/components/unit/form/unit-form.component.html
kimhw0630: oh.. i forgot to remove inside condtion... thanks

Pucek9 (approved):

PR #15514, opened 2022-03-31
chore: Improving ci pipeline total build time and execution when pull requests contains non-relevant files for CI

bgambocjaviniar:


.github/workflows/ci.yml
```diff
@@ -38,6 +51,16 @@ jobs:
         uses: actions/setup-node@v2
         with:
           node-version: ${{ env.NODE_VERSION }}
+      - name: Cache node_modules
+        id: cache-node-modules
+        uses: actions/cache@v2
+        with:
+          path: | 
+            node_modules
```
RadhepS, line 59: should we try to install root package in case there is a difference between the node modules in develop vs the PR?
(For example the PR introduces an additional library or library versions differ)

.github/workflows/ci.yml
```diff
@@ -38,6 +51,16 @@ jobs:
...
+            projects/storefrontapp-e2e-cypress/node_modules
```
RadhepS, line 60: same as other comment, should we try to install the cypress package as well in case? (library versions might differ)

.github/workflows/ci.yml
```diff
@@ -80,61 +123,74 @@ jobs:
...
+          restore-keys: nodemodules-${{ github.event.pull_request.base.sha }}
       - name: Run e2es
         env: 
           SPA_ENV: ci,b2c
         run: |
           ci-scripts/e2e-cypress.sh
   b2c_ssr_e2e_tests:
-    needs: no_retries
+    needs: [no_retries, validate_e2e_execution]
     name: E2E tests for SSR B2C
```
RadhepS, line 152: should we add the environment variable here? SPA_ENV: ci,b2c

.github/workflows/ci.yml
```diff
@@ -28,6 +27,20 @@ jobs:
...
+      SHOULD_RUN_E2E: ${{ steps.save-e2e-output-result.outputs.SHOULD_RUN_E2E }}
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Determine whether to run e2es
+        id: save-e2e-output-result
+        run: |         
+          source ci-scripts/validate-e2e-execution.sh
+          echo "::set-output name=SHOULD_RUN_E2E::$(echo $RUN_E2E)"
```
RadhepS, line 43: i dont see this used in an if statement to determine whether we want to run tests or not.
i see it on your testing branch. im guessing this is because this PR is most likely still in-progress?

bgambocjaviniar: nay it is here because that's how to set outputs in github actions. Without this, you won't be able to share data between jobs as it will be blank

.github/workflows/ci.yml
bgambocjaviniar: yea I guess we can, but did you determine if it was a mis-config / design from travis?

.github/workflows/ci.yml
bgambocjaviniar: for this particular job, no because unit tests job won't test e2e.

Moreover, the e2e job(s) install packages in both places from the script, so no need as well

ci-scripts/validate-e2e-execution.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
+set -e
+set -o pipefail
+
+FILES=$(git diff --name-only origin/develop)
```
RadhepS: like we talked about, this can be improved so it gets the diff of the branch this PR is trying to merge into (can be done later)

.github/workflows/ci.yml
```diff
@@ -7,7 +7,6 @@ on:
       - integration-libs/**
```
RadhepS, line 7: and like we talked about, the issue where the pipeline didn't start.
you can let me know when that's fixed

ci-scripts/validate-e2e-execution.sh
bgambocjaviniar: done

.github/workflows/ci.yml
bgambocjaviniar: done

RadhepS (approved):

PR #15523, opened 2022-04-01
Feat: rule based configurator -navigation from conflicting characteristic to conflicting group

### LINKED ISSUE #15498: Enable "View Details" Link (identify conflict and trigger navigation) and restore "Conflict Detected - View Details" text
To navigate to the conflict we need name of the conflict attribute as well as the id of the conflict group. The name is trivial, because conflict group attribute and cstic group attribute share the same name. getting the conflict group id is more complex. given the cstic group attribute key, we search for a conflict group containing an attribute with the same key. we use the id of the first conflict group found for navigation.

Changsuwan:


feature-libs/product-configurator/rulebased/components/attribute/header/configurator-attribute-header.component.ts
```diff
@@ -153,17 +153,39 @@ export class ConfiguratorAttributeHeaderComponent
...
+  findConflictGroupId(
+    configuration: Configurator.Configuration,
+    currentAttribute: Configurator.Attribute
+  ): string {
+    let groupId = '';
+    configuration.flatGroups
+      .filter(
+        (group) => group.groupType === Configurator.GroupType.CONFLICT_GROUP
+      )
+      .forEach((group) => {
```
ChristophHi: Shouldn't we stop after we found the first conflict group with the desired attribute? This will always loop through all conflict groups

feature-libs/product-configurator/rulebased/components/attribute/header/configurator-attribute-header.component.ts
```diff
@@ -153,17 +153,39 @@ export class ConfiguratorAttributeHeaderComponent
...
+          this.focusAttribute(this.attribute.name);
+        }
+      });
+  }
+
+  findConflictGroupId(
+    configuration: Configurator.Configuration,
+    currentAttribute: Configurator.Attribute
+  ): string {
+    let groupId = '';
```
ChristophHi: Why initialize with ''? In case we don't find anything, groupId can stay undefined

feature-libs/product-configurator/rulebased/components/attribute/header/configurator-attribute-header.component.ts
```diff
@@ -153,17 +153,39 @@ export class ConfiguratorAttributeHeaderComponent
...
-              configuration,
-              groupId
-            );
-            this.focusAttribute(this.attribute.name);
-          }
+          groupId = this.attribute.groupId;
+        } else {
+          groupId = this.findConflictGroupId(configuration, this.attribute);
         }
+        if (groupId) {
```
ChristophHi, line 163: What happens if groupId is not defined i.e. group was not found?

ChristophHi (dismissed):

feature-libs/product-configurator/rulebased/components/attribute/header/configurator-attribute-header.component.ts
Changsuwan: @Uli-Tiger can we use **find** instead?

feature-libs/product-configurator/rulebased/components/attribute/header/configurator-attribute-header.component.ts
Uli-Tiger: @ChristophHi: We will stay on the current page. This is the same behavior, as for the already implemented navigation from conflict to group (when the attribute is not found). Any other suggestion?

ChristophHi: Then I suggest to write a warning in case dev mode is active. This may help us and projects to identify issues. See e.g. ConfiguratorCartEntryBundleInfoService. (The situation might happen if the needed OCC enhancements are not present)

feature-libs/product-configurator/rulebased/components/attribute/header/configurator-attribute-header.component.ts
```diff
@@ -153,19 +153,37 @@ export class ConfiguratorAttributeHeaderComponent
...
+  findConflictGroupId(
+    configuration: Configurator.Configuration,
+    currentAttribute: Configurator.Attribute
+  ): string | undefined {
+    return configuration.flatGroups
+      .filter(
+        (group) => group.groupType === Configurator.GroupType.CONFLICT_GROUP
+      )
+      .find((group) => {
+        return group.attributes?.find(
```
ChristophHi, line 198: The branch where attributes is undefined is not covered, could you maybe at that in the test?

Uli-Tiger (approved):

Uli-Tiger:
closes #15565

PR #15524, opened 2022-04-01
fix: Flaky Replenishment Order E2E Tests (#14802)

### LINKED ISSUE #14802: B2B Replenishment E2E High Failure Rate in Pipeline. Improvement Request.
B2B Replenishment Tests have a high failure rate in pipeline.

It's decreasing overall performance of the test suite.
It causes unnecessary repetition in order to get a green build.
Request timeout was increased to 45 secs, but no improvement was seen.

**Request:**

- Please analyze failures in pipeline (not localhost, nor dev server)
- Address issues where e2e fail. Consider backend maybe under high load.
- Modify e2e test so that it's flexible at certain junctions where backend maybe under high load.
- Test modifications in a dev server (headless mode) at least 25 times at different intervals throughout the day.

---

As a starting point, please look at B2B replenishment errors in https://dashboard.cypress.io/projects/k3nmep/runs/27216/overview

Screen Shot 2022-01-18 at 11 20 27 AM

Screen Shot 2022-01-18 at 11 14 10 AM

pawelfras:
closes #14802


projects/storefrontapp-e2e-cypress/cypress/helpers/b2b/b2b-checkout.ts
```diff
@@ -74,10 +61,15 @@ export function addB2bProductToCartAndCheckout() {
 }
 
 export function enterPONumber() {
+  const getPaymentTypes = interceptPaymentTypesEndpoint();
+
+  cy.wait(`@${getPaymentTypes}`).its('response.statusCode').should('eq', 200);
```
bgambocjaviniar: This will always be flaky as there's going to have a race condition at times. Recent commits 'checks' shows e2e being more flaky now

This `wait` should be when we click the button to start the checkout process, for example, clicking the 'proceed to checkout button', however, do note, this would be fine if both functions, `enterPoNumber` and `addB2bProductToCartAndCheckout`, were in the same `it` (but it's not)

The `wait` should be within this [function](https://github.com/SAP/spartacus/blob/develop/projects/storefrontapp-e2e-cypress/cypress/integration/b2b/regression/checkout/b2b-account-checkout-flow.core-e2e-spec.ts#L29) `addB2bProductToCartAndCheckout`

projects/storefrontapp-e2e-cypress/cypress/helpers/b2b/b2b-checkout.ts
```diff
@@ -128,38 +119,41 @@ export function selectCreditCardPayment() {
 }
 
 export function selectAccountShippingAddress() {
+  const getCartWithDeliveryAddress = interceptCartWithDeliveryAddressEndpoint();
+  const updateAddress = interceptUpdateAddressEndpoint();
+  const getCheckoutDetails = interceptCheckoutB2BDetailsEndpoint();
+  const pudDeliveryMode = interceptPutDeliveryModeEndpoint();
```
bgambocjaviniar:
```suggestion
  const putDeliveryMode = interceptPutDeliveryModeEndpoint();
```

projects/storefrontapp-e2e-cypress/cypress/helpers/b2b/b2b-checkout.ts
```diff
@@ -436,3 +430,63 @@ export function reviewB2bOrderConfirmation(
...
+    method: 'GET',
+    path: `${Cypress.env('OCC_PREFIX')}/${Cypress.env(
+      'BASE_SITE'
+    )}/paymenttypes*`,
+  }).as(alias);
+
+  return alias;
+}
+
+export function interceptCartWithDeliveryAddressEndpoint(): string {
```
bgambocjaviniar: we don't need this. In checkout, all 'set' checkout details are found in the checkout details request. 

This function is intercepting the same thing as `interceptCheckoutB2BDetailsEndpoint`, therefore we do not it.

projects/storefrontapp-e2e-cypress/cypress/helpers/b2b/b2b-checkout.ts
```diff
@@ -128,38 +119,41 @@ export function selectCreditCardPayment() {
...
+
+  //wait for updating address
+  cy.wait(`@${updateAddress}`).its('response.statusCode').should('eq', 200);
+
+  //wait for response with updated address
+  cy.wait(`@${getCartWithDeliveryAddress}`)
+    .its('response.statusCode')
+    .should('eq', 200);
 
   cy.get('cx-card').within(() => {
```
bgambocjaviniar, line 143: I would say we don't need to add line 133 to 144, but we can keep.

But let me know of your opinion. Reason why we don't need is we're not waiting for anything data specific and the test is already deterministic. But the real reason is cy.get('cx-card') will wait for 30s anyhow for the card to appear. Although, I agree having extra checks for the response to actually come back is useful, but for this case, I really don't think we need it.

P.S. As mentioned in some other comment, but replace `getCartWithDeliveryAddress` with `getCheckoutDetails` if you decide to keep line 133 to 144.

I would also argue that I'm pretty sure this part of checkout (shipping address) never had an issue/flaky unless said otherwise!

bgambocjaviniar (changes requested):

projects/storefrontapp-e2e-cypress/cypress/helpers/b2b/b2b-checkout.ts
bgambocjaviniar: P.S

for flaky fixing tests, you should run them in headless like 10-20 times to verify if it actual got fixed, and like maybe ~5 times in CI.

Locally, you can include this in your root `package.json`. Make sure your cypress.json have the right b2b configs and that your actual application is on b2b.

```
testcy": "for i in {1..20}; do yarn e2e:run --spec cypress/integration/b2b/regression/checkout/b2b-account-checkout-flow.core-e2e-spec.ts; done",
```

`cypress.json`

  >   ....
    "BASE_SITE": "powertools-spa",
    ...
    "OCC_PREFIX_USER_ENDPOINT": "orgUsers",
   ....

pawelfras: This is very good point! I was wondering about this problem over the weekend before I saw your comment and came to the same conclusions - we have to run this `wait` beforehand, because there is a chance that by checking it in the current step the data is already downloaded, so we intercepting the query that has already been made before.

projects/storefrontapp-e2e-cypress/cypress/helpers/b2b/b2b-checkout.ts
pawelfras: The reason why I've added those checks was that when I ran test cases with throttling (Slow 3G) sometimes card wasn't appear and I was getting an error. It doesn't cost much but can help to avoid some flakiness in the future. I'll replace `getCartWithDeliveryAddress` with `getCheckoutDetails`.
Screenshot 2022-04-04 at 11 11 47

projects/storefrontapp-e2e-cypress/cypress/helpers/b2b/b2b-checkout.ts
pawelfras: Yeah, that's true. Sorry, I had to miss it.

projects/storefrontapp-e2e-cypress/cypress/helpers/b2b/b2b-checkout.ts
bgambocjaviniar: As discussed, I concluded that's not a use case, where we are throttling the browser on the testrunner, therefore not a flaky test as we are running headlessly.

Although will keep to make the e2e more bulletproof.

bgambocjaviniar (approved):

PR #15527, opened 2022-04-01
chore: Update dependencies on certain github actions and installation script

### LINKED ISSUE #15525: Update dependencies on certain github actions and installation script
some of these are failing due to using node 12 images

but the installation script is failing for another reason. To verify and fix.

Expectation: Making the CI pipeline on develop truthy with only flaky e2es instead of other errors

bgambocjaviniar:
closes GH-15525


.github/workflows/cache-builded-libs.yml
```diff
@@ -1,6 +1,7 @@
 on:
   push:
     branches:
+      - feature/GH-15525
```
bgambocjaviniar: will remove, but to test

projects/storefrontapp/server.ts
```diff
@@ -63,7 +63,7 @@ export function app() {
 }
 
 function run() {
-  const port = process.env.PORT || 4000;
+  const port = process.env['PORT'] || 4000;
```
bgambocjaviniar, line 66: to verify as we're exporting this in the script and to update reference if ever in the schematics for add ssr

when we build:ssr

Screen Shot 2022-04-01 at 10 50 13 AM

.github/workflows/install.yml
```diff
@@ -4,6 +4,7 @@ on:
       - 'release/4.**'
       - 'epic/**'
       - 'develop'
+      - feature/GH-15525
```
bgambocjaviniar: will remove but to test

feature-libs/order/styles/components/order-history/_index.scss
```diff
@@ -8,7 +9,6 @@
 @import './return-request-totals';
 @import './cancel-order';
 @import './consignment-tracking';
-@import 'cart';
```
bgambocjaviniar: if this was to work, it was not a great solution as we would have been providing the entire cart styles from the order lib. But good thing our installation script caught the issue on the customer facing app.

Potential solutions to bring up to team:
- webpack to resolve for local development in order to actually use @spartacus/cart, which will work in customer's app. But again, it causes issues bringing the entire cart styles until we properly split styles, which is in our backlog. Although, I'm sceptical that this was to work if cart library was not included (did not test this)
- move the specific cart item styles in the @spartacus/styles, so that it can be available and re-used throughout features
- copy paste the 'specific' styles in order lib and re-use it there.

Pucek9: additional ideas:
- using classes instead placeholder selector, components from order lib already use cart lib, so we are ensured that it should be loaded already. Styles specified for components, probably will be in another class, then the element will have 2 classes: `cx-cart-item-list cx-specified-list`) [#15569](https://github.com/SAP/spartacus/issues/15569)
- keep placeholder selectors which would be nice to override in specified patch: `@spartacus/cart/shared`

bgambocjaviniar: I'll drop fixing it in this PR as I saw it will be worked on your PR https://github.com/SAP/spartacus/pull/15564

projects/storefrontapp/server.ts
bgambocjaviniar: Since angular13, new generated shellapp through ng cli provides by default the `noPropertyAccessFromIndexSignature` property, which won't allow access to index signatures using `dots`, but only through `indexed`

RadhepS (approved):
LGTM

PR #15536, opened 2022-04-04
feat: Generate wrapper modules

znikola:
Title.


feature-libs/cart/schematics/add-cart/__snapshots__/index_spec.ts.snap
```diff
@@ -44,10 +44,20 @@ import { CmsConfig, I18nConfig, provideConfig } from \\"@spartacus/core\\";
       [CART_BASE_FEATURE]: {
         module: () =>
           import('@spartacus/cart/base').then((m) => m.CartBaseModule),
-      }, [MINI_CART_FEATURE]: {
+      },
+    }
+  }),
+  provideConfig(<CmsConfig>{
```
bgambocjaviniar, line 50: I understand this is just a snap, but providing a config for the same 'object' instead of encapsulating them into one was part of the initiative on tackling all whitespaces and trailing commas? Basically normalizing the configs right?

feature-libs/checkout/b2b/root/checkout-b2b-root.module.ts
```diff
@@ -35,7 +34,7 @@ export function defaultCheckoutComponentsConfig() {
 }
```
bgambocjaviniar, line 34: just a question as I forgot why we had to spread the mapping from base and b2b to b2b and repl respectively.

My question is geared towards whether we can removed them due to the fact we're having this wrapper and eagerly loading them on the 'feature-module' now instead

projects/schematics/src/shared/utils/feature-utils.ts
```diff
@@ -1,67 +1,768 @@
...
+      tree,
+      options.project
+    );
+
+    // fresh installation
+    // TODO:#schematics - use options.internal.?
+    // TODO:#schematics - check if this is properly set when installing though a lib (e.g. run ng add @spa/dp)
+    if (!spartacusFeatureModuleExists) {
+      if (options.debug) {
+        context.logger.info(`✅  Analysis of installed features complete.`);
```
bgambocjaviniar: nitpicking, but since it's a fresh installation, then there's no installed features to analyze, therefore wouldn't it be best to bring line 628 - 641 to line 623 so that we won't have `Analyzing installed features..` fire if ever.

This would mean we can update this message to something more verbose. Something along the lines of 'Fresh Spartacus installation. Begin scaffolding?'

feature-libs/product/schematics/add-product/index.ts
```diff
@@ -28,17 +25,9 @@ export function addSpartacusProduct(options: SpartacusProductOptions): Rule {
     );
 
     return chain([
+      analyzeInstalledFeatures(options, features),
```
bgambocjaviniar: reminder to discuss

projects/schematics/src/shared/utils/feature-utils.ts
bgambocjaviniar: or skip the log and just return the noop when we move it above the other log

feature-libs/product/schematics/add-product/index.ts
bgambocjaviniar: was this 'pre-check' done to help alleviate the issue existing users to properly follow our 'convention' of things for feature modules

feature-libs/smartedit/schematics/add-smartedit/index.ts
```diff
@@ -29,51 +23,11 @@ export function addSmartEditFeatures(options: SpartacusSmartEditOptions): Rule {
     const features = analyzeCrossFeatureDependencies(
       options.features as string[]
     );
-    const overrides = buildSmartEditConfig(options);
 
     return chain([
-      addFeatures(options, features, overrides),
+      analyzeInstalledFeatures(options, features),
+      addFeatures(options, features),
```
bgambocjaviniar, line 29: 👍  for normalization of adding features and making it possible with xxxes-schematics-config.ts

bgambocjaviniar (commented):
first 65 files :x (today was hectic)

very minor changes, and a few questions

projects/schematics/src/shared/lib-configs/integration-libs/cdc-schematics-config.ts
```diff
@@ -35,24 +36,29 @@ export const CDC_SCHEMATICS_CONFIG: FeatureConfig = {
...
   },
-  customConfig: {
-    import: [
-      {
-        moduleSpecifier: SPARTACUS_CDC_ROOT,
-        namedImports: [CDC_CONFIG],
-      },
-    ],
-    content: `<${CDC_CONFIG}>{
+  customConfig: () => ({
```
bgambocjaviniar: should we do this in-line function, or should we return it from a function similarly to other custom config instances in our application

projects/schematics/src/shared/lib-configs/smartedit-schematics-config.ts
```diff
@@ -36,3 +44,26 @@ export const SMARTEDIT_SCHEMATICS_CONFIG: FeatureConfig = {
...
+        },
+      ],
+      content: `<${SMART_EDIT_CONFIG}>{
+        smartEdit: {
+          storefrontPreviewRoute: '${
+            options.storefrontPreviewRoute ??
+            'STOREFRONT_PREVIEW_ROUTE_PLACEHOLDER'
+          }',
+          allowOrigin: '${options.allowOrigin ?? 'ALLOWED_ORIGIN_PLACEHOLDER'}',
+        },\n}`,
```
bgambocjaviniar: nitpicking, but do we need the `\n` for the new line. I believe the other 'customConfigs' in our application doesn't need the extra `,\n`

projects/schematics/src/shared/schematics-config-mappings_spec.ts
```diff
@@ -0,0 +1,175 @@
...
+      ]);
+      expect(result.get(CHECKOUT_B2B_FEATURE_NAME)).toEqual([
+        CHECKOUT_B2B_ROOT_MODULE,
+      ]);
+      expect(result.get(CHECKOUT_SCHEDULED_REPLENISHMENT_FEATURE_NAME)).toEqual(
+        [CHECKOUT_SCHEDULED_REPLENISHMENT_ROOT_MODULE]
+      );
+      expect(result.get(ORDER_FEATURE_NAME)).toEqual([ORDER_ROOT_MODULE]);
+      expect(result.get(CDC_FEATURE_NAME)).toEqual([CDC_ROOT_MODULE]);
+      expect(result.get(CDS_FEATURE_NAME)).toEqual([]);
```
bgambocjaviniar, line 100: 👍

projects/schematics/src/shared/schematics-config-mappings.ts
```diff
@@ -0,0 +1,301 @@
...
+   */
+  featureRootModuleMapping,
+  /**
+   * Mapping of schematics configurations to the Spartacus library.
+   *
+   * E.g.:
+   *
+   * {
+   * ...,
+   * '@spartacus/checkout': [baseConfig, b2bConfig, scheduledReplenishmentConfig],
```
bgambocjaviniar: think this doc is incorrect now compared to what it really did before

The mapping is feature-name to the entire config object instead of the mainscope
Checkout-B2b: entire-config-object from the CHECKOUT_B2B_SCHEMATICS_CONFIG that contains, lib, foldername,modulename, featuremodule, etc

bgambocjaviniar (commented):
part 2 88/130

need a proper sit down during working time for the rest

feature-libs/cart/schematics/add-cart/__snapshots__/index_spec.ts.snap
znikola: Hmm, no, I think it's just a consequence of some internal refactoring because now we have to provide more granular configurations.
I actually prefer it to be like this, as it allows for easy config shifting.

feature-libs/checkout/b2b/root/checkout-b2b-root.module.ts
znikola: I don't quite understand to which mapping spread are you referring? The one with CMS components (`CHECKOUT_B2B_CMS_COMPONENTS`)?

We have to do it because our deepMerge of configs doesn't support array merging.

projects/schematics/src/shared/utils/feature-utils.ts
znikola: Good point, addressed.

feature-libs/product/schematics/add-product/index.ts
znikola: This code is now updated, please take another look.
The pre-check is done for two reasons:
1. Check if some required features are not configured. For more, see `If the feature is not configured, we can not proceed` in https://github.com/SAP/spartacus/issues/15380
2. It checks if we could order the modules in Spartacus feature module and in wrapper modules.

projects/schematics/src/shared/lib-configs/integration-libs/cdc-schematics-config.ts
znikola: I think it's simple enough to be inlined. Do you think we should not inline it?

projects/schematics/src/shared/lib-configs/smartedit-schematics-config.ts
znikola: Hmm, good catch. The generated code will just look nice if there's a new line after the custom config, but let's remove it for consistency's sake.

projects/schematics/src/shared/schematics-config-mappings.ts
znikola: Yep, a great catch, it's a leftover from the previous version.

projects/storefrontapp/src/app/spartacus/features/user-profile-wrapper.module.ts
```diff
@@ -13,4 +13,4 @@ if (environment.cdc) {
   declarations: [],
   imports: [UserProfileModule, ...extensions],
 })
-export class UserWrapperModule {}
+export class UserProfileWrapperModule {}
```
bgambocjaviniar, line 16: I believe you need to update the snaps + spec files for path as it's changed to user-profile-wrapper.module.ts

projects/schematics/src/shared/lib-configs/integration-libs/cdc-schematics-config.ts
bgambocjaviniar: reminder that it was meant about consistency from the rest of the customConfig across the application, which are within a function

An example of another simple inlined function is the smartedit/tracking custom config. (CDC is currently adding the option to allow configuring javascriptUrl)

projects/schematics/src/shared/libs-constants.ts
```diff
@@ -167,52 +167,51 @@ export const SPARTACUS_CONFIGURATION_MODULE = 'spartacus-configuration';
 /***** File structure end *****/
 
 /***** CLI start *****/
```
bgambocjaviniar: nitpicking, but since we swapped from CLI_*, I think it should be feature name now

projects/schematics/src/shared/schematics-config-mappings.ts
```diff
@@ -0,0 +1,302 @@
...
+import {
+  USER_ACCOUNT_SCHEMATICS_CONFIG,
+  USER_PROFILE_SCHEMATICS_CONFIG,
+} from './lib-configs/user-schematics-config';
+import { Module, SchematicConfig } from './utils/lib-utils';
+
+/**
+ * A list of all schematics feature configurations.
+ * _Must_ be updated when adding a new schematics
+ * library or a feature.
```
bgambocjaviniar, line 51: might be best to add it in here `### Configuring Schematics` for `creating-lib.md` in our spartacus repo

projects/schematics/src/shared/utils/feature-utils.ts
```diff
@@ -1,67 +1,903 @@
...
+
+export interface FeatureModuleImports {
+  importPath: string;
+  moduleNode: Expression | Identifier;
+}
+/**
+ * Represents the result of the `collectInstalledModules` function,
+ * and contains all features imported in the `spartacus-features.module.ts` file.
+ */
+export interface SpartacusFeatureModule {
```
bgambocjaviniar: Do we still need this interface

projects/schematics/src/shared/utils/feature-utils.ts
```diff
@@ -1,67 +1,903 @@
...
-          `No feature config found for ${feature}. Please check if you added the schematics config to the projects/schematics/src/shared/updateable-constants.ts`
+          `[internal] No feature config found for ${feature}. ` +
+            `Please check if  the schematics config is added to projects/schematics/src/shared/schematics-config-mappings.ts`
         );
       }
 
-      const config =
-        configurationOverrides?.[feature]?.schematics ??
-        schematicsConfiguration;
+      // TODO:#schematics - fix the interactivity for the CDS / ASM, etc.
```
bgambocjaviniar, line 180: 👍

projects/schematics/src/shared/utils/feature-utils.ts
```diff
@@ -1,67 +1,903 @@
...
+    if (analysis.feature) {
+      features.push({ element, feature: analysis.feature });
+      continue;
+    }
+  }
+
+  return {
+    core,
+    features,
+    empty,
```
bgambocjaviniar, line 353: I see, that's how you fixed the ordering issue like digital payments

projects/schematics/src/shared/utils/feature-utils.ts
```diff
@@ -1,67 +1,903 @@
...
+          false
+        )?.replaceWithText(`[${ordered.join(',\n')}]`);
+
+        saveAndFormat(spartacusFeaturesModule);
+        break;
+      }
+    }
+
+    if (options.debug) {
+      context.logger.info(`✅ Ordering Spartacus feature modules...`);
```
bgambocjaviniar: ordering complete!

projects/schematics/src/shared/utils/feature-utils.ts
```diff
@@ -1,67 +1,903 @@
...
+
+  return undefined;
+}
+
+/**
+ * Searches through all the source files,
+ * and looks for the either the static or
+ * dynamic imports of the feature module.
+ */
+function findFeatureModule(
```
bgambocjaviniar, line 750: not here, but related to the wrapper-modle/index.ts

maybe rename the other one to a more specific one

projects/schematics/src/shared/utils/graph-utils_spec.ts
```diff
@@ -173,38 +143,40 @@ describe('Graph utils', () => {
...
-        CLI_CART_IMPORT_EXPORT_FEATURE,
-        CLI_ASM_FEATURE,
-      ]);
+      expect(crossFeatureInstallationOrder).toMatchInlineSnapshot(`
+        Array [
+          "Account",
+          "Profile",
+          "Cart",
+          "Saved-Cart",
+          "WishList",
```
bgambocjaviniar, line 152: Cool how the algo re-ordered it based on properly putting dependencies and made it look better :P

projects/schematics/src/shared/utils/import-utils.ts
```diff
@@ -45,3 +73,148 @@ export function getImportDeclaration(
...
+}
+
+/**
+ * Returns the import path, e.g. @spartacus/cart/base
+ */
+export function getDynamicImportImportPath(
+  arrowFunction: ArrowFunction
+): string | undefined {
+  return getDynamicImportCallExpression(arrowFunction)
+    ?.getFirstDescendantByKind(tsMorph.SyntaxKind.StringLiteral)
```
bgambocjaviniar, line 117: I have to say, ts-morph is quite awesome 👍 
I need to dabble into it

projects/schematics/src/shared/utils/lib-utils.ts
```diff
@@ -180,9 +216,15 @@ export function shouldAddFeature(
 
 export function addLibraryFeature<T extends LibraryOptions>(
   options: T,
-  config: FeatureConfig
+  config: SchematicConfig
 ): Rule {
   return (tree: Tree, context: SchematicContext) => {
+    if (options.debug) {
+      context.logger.info(
```
bgambocjaviniar: let's make use of your debugLog()

projects/schematics/src/shared/utils/lib-utils.ts
```diff
@@ -452,9 +417,33 @@ function addFeatureModule<T extends LibraryOptions>(
   };
 }
 
+/**
+ * Checks if the given feature module is already being
```
bgambocjaviniar: Maybe something like "Checks if the given feature module is already imported in a wrapper module"

projects/schematics/src/shared/utils/workspace-utils.ts
```diff
@@ -177,7 +179,11 @@ export function validateSpartacusInstallation(packageJson: any): void {
 }
 
 export function scaffoldStructure(options: SpartacusOptions): Rule {
-  return (_tree: Tree) => {
+  return (_tree: Tree, context: SchematicContext) => {
+    if (options.debug) {
+      context.logger.info(`⌛️ Scaffolding Spartacus file structure...`);
```
bgambocjaviniar: maybe throughout the application, you should make use of debugLog as mentioned previously.

You used it in line 207 here, so maybe change all instances to use that of 

```
debugLog(xxx/formatFeatureStart/formatFeatureComplete, options.debug);

```

projects/schematics/src/wrapper-module/index.ts
```diff
@@ -0,0 +1,563 @@
...
+      updateWrapperModule({
+        project: options.project,
+        moduleName: options.markerModuleName,
+        markerFeatureModulePath,
+        debug: options.debug,
+      }),
+      updateWrapperModule({
+        project: options.project,
+        moduleName: options.featureModuleName,
+        markerFeatureModulePath,
```
bgambocjaviniar, line 547: Kind of confused when I'm trying to understand

why do we have to updateWrapperModule twice for the marker module name and the feature module that will be added to the wrapper

projects/storefrontapp/src/app/spartacus/features/user-profile-wrapper.module.ts
znikola: We didn't change the generated file's name. 
@Platonn also pointed out that we should change it, but for this to we would have to split the Profile and Account features into separate feature modules.

projects/schematics/src/shared/utils/feature-utils.ts
znikola: No, good catch!

projects/schematics/src/shared/schematics-config-mappings.ts
znikola: Nice, added.

projects/schematics/src/shared/utils/feature-utils.ts
znikola: Good catch, updated.

projects/schematics/src/shared/utils/import-utils.ts
znikola: For sure, it's the bomb :)

projects/schematics/src/shared/utils/lib-utils.ts
znikola: Nice, thanks :)

projects/schematics/src/shared/utils/lib-utils.ts
znikola: Done. I renamed the `debugLog` to `debugLogRule`

projects/schematics/src/shared/utils/workspace-utils.ts
znikola: I updated all the instances that I could.
Note that the new `debugLogRule` is just a rule that logs the given message, and we can't use it in places where we don't have a chain of rules being returned, or have an early return statement.

projects/schematics/src/shared/utils/feature-utils.ts
znikola: Nice, done.

projects/schematics/src/wrapper-module/index.ts
znikola: Because we need to statically import both the marker (`markerModuleName`) and target feature module (`featureModuleName`) to the wrapper module.

bgambocjaviniar (approved):

PR #15542, opened 2022-04-04
fix: remove facets on following search result of 0 (#15531)

### LINKED ISSUE #15531: Facets are not updated when the next search result is 0
**Environment Details**
Browser: Chrome
Device: mac
Frontend: develop branch on my local
Backend: 40.76.109.9

**Steps to Reproduce**
Go on electronics storefront
Search for eg: cam -25%
There will be 134 results returned.
Now search for eg: cam -25%~`!@#$^&*()_+-={}|[]\:";'<>?,./"
There will be 0 results returned.

**Actual Result**
However, the facets are not updated accordingly. It still shows the same facets as the previous search. 
Screen Shot 2022-04-01 at 3 44 28 PM

**Note**:
However, if the first search was: cam -25%~`!@#$^&*()_+-={}|[]\:";'<>?,./"
Then the it's displayed properly
Screen Shot 2022-04-01 at 3 52 37 PM

BenjaminEmiliani:
Closes #15531


projects/storefrontlib/cms-components/product/product-list/product-facet-navigation/services/product-facet.service.ts
```diff
@@ -24,6 +24,9 @@ export class ProductFacetService {
     protected productListComponentService: ProductListComponentService
   ) {}
 
+  model$: Observable<ProductSearchPage> =
+    this.productListComponentService.model$;
+
```
WeizhengSap: This issue is caused by `filterForPage`.
line 90 should be `return page.freeTextSearch === state.params.query.split(':')[0];`

WeizhengSap (approved):

PR #15545, opened 2022-04-05
feat: Custom protocol handling for generic links

### LINKED ISSUE #15526: Custom Protocol Handler for Generic Links
While we can fix bugs/add handlers for protocols as they appear in generic links from the CMS, a custom protocol handler could be used to solve the issue instead of needing to upgrade to a patched version.

This handler would react to a given protocol within a link via regex and transform the link accordingly given custom logic.

See https://github.com/SAP/spartacus/pull/15429/files for an example of an OOTB protocol fix that needs to be backported.

There are many protocols that could occur, therefore we should add this custom handler to avoid back porting for every fixed/added OOTB handler: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#attr-href

Zeyber:
Closes: #15526


projects/storefrontlib/shared/components/generic-link/generic-link.service.ts
```diff
@@ -0,0 +1,22 @@
...
+  MAILTO_PROTOCOL_REGEX,
+  TEL_PROTOCOL_REGEX,
+} from './generic-link.models';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class GenericLinkService {
+  /**
+   * Returns true when the @Input `url` is a string starting with `http://`, `https://`, `mailto:`, `tel:`.
```
Platonn:
```suggestion
   * Returns true when the given `url` is a string starting with `http://`, `https://`, `mailto:`, `tel:`.
```

projects/storefrontlib/shared/components/generic-link/generic-link.service.ts
```diff
@@ -0,0 +1,22 @@
...
+import {
+  HTTP_PROTOCOL_REGEX,
+  MAILTO_PROTOCOL_REGEX,
+  TEL_PROTOCOL_REGEX,
+} from './generic-link.models';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class GenericLinkService {
```
Platonn: This service has a very specialized responsibility - being an extension point for the `GenericLinkComponent`. IMHO the service name **`GenericLinkService`** in `@spartacus/storefront` shounds a bit to general.
What do you think about naming it to `GenericLinkComponentService` so it corresponds directly to the `GenericLinkComponent`?

projects/storefrontlib/shared/components/generic-link/generic-link.models.ts
```diff
@@ -0,0 +1,14 @@
...
+
+/**
+ * Pattern matching string starting with `mailto:`.
+ */
+export const MAILTO_PROTOCOL_REGEX: RegExp = /^mailto:/i;
+
+/**
+ * Pattern matching string starting with `tel:`.
+ */
+export const TEL_PROTOCOL_REGEX: RegExp = /^tel:/i;
```
Platonn: Nitpick: now those items are loose members of public API of `@spartacus/storefront`  and customers may use  them freely in various places in their customisations, e.g `import { TEL_PROTOCOL_REGEX } from '@spartacus/storefront'`. That's not too bad. But at the same time, I'm not convinced if it's necessary to expose them as loose items.

What do you think about moving those consts to protected fields of the service, so they are scoped only to the service's logic?

projects/storefrontlib/shared/components/generic-link/generic-link.service.ts
```diff
@@ -0,0 +1,22 @@
...
+  /**
+   * Returns true when the @Input `url` is a string starting with `http://`, `https://`, `mailto:`, `tel:`.
+   */
+  isExternalUrl(url: string | any[]): boolean {
+    return (
+      (typeof url === 'string' && HTTP_PROTOCOL_REGEX.test(url)) ||
+      MAILTO_PROTOCOL_REGEX.test(<string>url) ||
+      TEL_PROTOCOL_REGEX.test(<string>url)
```
Platonn: if you wrap all regex checks in one parenthesis, you won't need to cast `<string>url`
```suggestion
      typeof url === 'string' && 
      (
        HTTP_PROTOCOL_REGEX.test(url) ||
        MAILTO_PROTOCOL_REGEX.test(url) ||
        TEL_PROTOCOL_REGEX.test(url)
      )
```
+/- fix indentation (run prettier)

Platonn (dismissed):

Platonn (approved):

PR #15548, opened 2022-04-05
fix: remove cart with invalid products (#14228)

### LINKED ISSUE #14228: Import to saved cart creates a saved cart even when there's no valid products added
**Describe the bug**
Using the import to cart feature in saved cart, we're able to create a saved cart by trying to import an invalid product to the saved cart list.

**To Reproduce**
1. Login and head to saved carts under the my-account dropdown menu
2. use a csv file with an invalid product for the catalog. Ex: trying to import products that is b2c in b2b, vice-versa
3. See created 'empty saved cart'

**Expected behavior**
Should not create an empty saved cart as when you visit the details page of that empty cart, it will try and delete it.

**Screenshots**
IMG_0155

I539321:
Hi!

We've added another test case for deleting the saved cart if it only tries to import invalid products. Perhaps the two outer `describe` methods can be merged (different classes were required for `ProductImportInfoService`, hence we separated). Let us know if testing should be done differently.

There's a UI aspect to this ticket, which is not included in the current ticket's scope, would you like us to create another ticket for that @Xymmer?

Closes #14228.


feature-libs/cart/saved-cart/components/page-context/saved-carts-page/new-saved-cart-order-entries.context.ts
```diff
@@ -39,6 +41,23 @@ export class NewSavedCartOrderEntriesContext implements AddOrderEntriesContext {
...
     return this.add(products, savedCartInfo).pipe(
+      tap((cartId: string) => {
+        this.importInfoService
+          .getResults(cartId)
+          .pipe(
+            take(products.length),
+            every(
+              (productInfo: ProductImportInfo) =>
+                productInfo.statusCode ===
+                ProductImportStatus.UNKNOWN_IDENTIFIER
```
RadhepS: I think we should also look for UNKNOWN_ERROR

feature-libs/cart/saved-cart/components/page-context/saved-carts-page/new-saved-cart-order-entries.context.ts
```diff
@@ -39,6 +41,23 @@ export class NewSavedCartOrderEntriesContext implements AddOrderEntriesContext {
...
+            )
+          )
+          .subscribe((isInvalid: boolean) => {
+            if (isInvalid) {
+              this.savedCartService.deleteSavedCart(cartId);
```
RadhepS, line 58: When we try to import a csv with an invalid product, we get two errors:
image

Can we prevent the cart not found error? Ideally preventing the creation of the cart rather than deleting.

RadhepS (commented):
When we try to import a csv with an invalid product, we get two errors:
image

Can we stop the creation of the cart instead of deleting it?

feature-libs/cart/saved-cart/components/page-context/saved-carts-page/new-saved-cart-order-entries.context.ts
I539321: Hi, thanks for the review!

AFAIK, in order to get the information on whether the import was successful or not, you need to create a cart and pass the ID of that during the OCC call, so we're not sure if we can prevent the creation of that beforehand. This was also our initial idea, as to not have this redundant step, but no luck. If you're aware of a way (perhaps another OCC endpoint that should be used), please let us know.

RadhepS: Yes we currently do not have an OCC endpoint for this. This solution should be fine for now. Well done :)

RadhepS (changes requested):
Thoughts on the other comment?

feature-libs/cart/saved-cart/components/page-context/saved-carts-page/new-saved-cart-order-entries.context.ts
I539321: Alright, just pushed it, let me know if any change should be made to the tests :)

I539321:
> Thoughts on the other comment?

Should be addressed now, let me know if anything else needs to be discussed :)

feature-libs/cart/saved-cart/components/page-context/saved-carts-page/new-saved-cart-order-entries.context.spec.ts
```diff
@@ -44,7 +50,15 @@ class MockProductImportInfoService {
   getResults = createSpy().and.returnValue(of(mockProductImportInfo));
 }
 
-describe('NewSavedCartOrderEntriesContext', () => {
+const mockFailingProductImportInfo = {
+  productCode: 'testProductCode',
+  statusCode: ProductImportStatus.UNKNOWN_IDENTIFIER,
```
RadhepS, line 112: Can we add a product case where the status code is UNKNOWN_ERROR?

So something like:
mockFailingProductImportWithUnknownError

RadhepS (changes requested):

feature-libs/cart/saved-cart/components/page-context/saved-carts-page/new-saved-cart-order-entries.context.spec.ts
I539321: Hey, pushed a commit regarding this, let me know if you'd like something changed!

RadhepS (approved):

PR #15552, opened 2022-04-05
refactor: Add pagination for loading multiple CMS components

### LINKED ISSUE #13080: Requesting a large amount of components at once from the CmsService is failing
- Description:
In my project I have a very large navigation tree. Once the CategoryNavigationComponent/NavigationService try to retrieve the navigation-node-components, all the requests are sent to the CmsService at once.
The call to the backend then looks like this:
`/cms/components?fields=DEFAULT&currentPage=0&pageSize=713&componentIds=<huge_list_of_components> `
Which results in a server error:
Status Code: 431 Request Header Fields Too Large

- Library version: 3.4

- Screenshot:
image

 - OS: Windows
 - Browser: chrome

WeizhengSap:
close #13080


projects/core/src/cms/connectors/component/cms-component.connector.ts
```diff
@@ -47,14 +47,27 @@ export class CmsComponentConnector {
...
         if (missingIds.length > 0) {
-          return this.cmsComponentAdapter
-            .findComponentsByIds(missingIds, pageContext)
-            .pipe(
-              map((loadedComponents) => [
-                ...configuredComponents.filter(Boolean),
-                ...loadedComponents,
-              ])
+          const pageSize = this.config.componentsLoading?.pageSize || 50;
+          const p: Observable<any>[] = [];
```
bgambocjaviniar:
```suggestion
          const p: Observable<CmsComponent[]>[] = [];
```

projects/core/src/cms/connectors/component/cms-component.connector.ts
```diff
@@ -47,14 +47,27 @@ export class CmsComponentConnector {
...
+          const num = Math.ceil(missingIds.length / pageSize);
```
bgambocjaviniar: Maybe let's call it totalPages

what do you think as I think it's more verbose for anyone to understand

projects/core/src/cms/connectors/component/cms-component.connector.ts
bgambocjaviniar: What about the name of cmsComponents instead of p?

projects/core/src/cms/connectors/component/cms-component.connector.ts
```diff
@@ -47,14 +47,27 @@ export class CmsComponentConnector {
...
+                'DEFAULT',
+                curr,
+                pageSize
+              )
             );
+            curr++;
+          }
+          return zip(...p).pipe(
+            map((comps) =>
+              [...configuredComponents.filter(Boolean)].concat(...comps)
```
bgambocjaviniar: let's keep the variable named loadedComponents for readability

projects/core/src/cms/connectors/component/cms-component.connector.ts
```diff
@@ -47,14 +47,27 @@ export class CmsComponentConnector {
...
-            .findComponentsByIds(missingIds, pageContext)
-            .pipe(
-              map((loadedComponents) => [
-                ...configuredComponents.filter(Boolean),
-                ...loadedComponents,
-              ])
+          const pageSize = this.config.componentsLoading?.pageSize || 50;
+          const p: Observable<any>[] = [];
+          const num = Math.ceil(missingIds.length / pageSize);
+          let curr = 0;
```
bgambocjaviniar: currentPage for readability

projects/core/src/cms/connectors/component/cms-component.connector.ts
bgambocjaviniar: Just wondering (not important), but the logic is implemented the same in the backend? (I didn't take a look, but I tried different sizes, and it looks like it does the ceiling always)

bgambocjaviniar (changes requested):
Although the implementation is good, where it makes multiple `same` requests to get the different response based on the 'page' in the backend, I do not think this solves the issue of having the query string being fairly large (100+ cms components in the query string). 

I think the under the hood implementation is almost there. Instead of working with page size and current/next page, we would need `to split in pieces the missingIds` instead of splitting how to get the response data.

With that being said, we can keep all the additional optional params for extensibility, but we should not have a default page size as in the occ-cms-component.adapter.ts, we will use the length of the ids (the length of each splitted query string requests)

bgambocjaviniar (approved):

PR #15563, opened 2022-04-06
fix: improve SR vocalization of applied filters (#15562)

### LINKED ISSUE #15562: Improve SR a11y of Applied Filters on PLP
**Current Behavior**
Keyboard navigation through applied filters does not provide meaningful narration for screen reader users. 

**To Reproduce**
1. Go to any PLP with facets
2. Apply one or more filters
3. Tab through applied filters with a screen reader on
4. SR only vocalizes value of filter

**Expected behavior**
More context in SR vocalization

BenjaminEmiliani:
Closes #15562


developpeurweb (dismissed):
Works well for JAWS/NVDA/VO

developpeurweb (dismissed):

Pucek9:
IMO converting links to have button behavior is over-engineering.

If I were to vote for solutions, they would be in that order:

1. Find out if it could be a real button or if it really makes sense for the UX (change to the button)
3. Since it was a link, leave it as a link and treat it as a link (no changes in this area, the user must use enter, without possibility using space)
5. If we want to have links that we treat as buttons, maybe let's prepare a directive?
6. The solution I just pushed (fewer changes)
7. Your early solution (too many changes)

developpeurweb (approved):
Still works well for JAWS/NVDA/VO

Pucek9 (approved):

PR #15564, opened 2022-04-06
hotfix: import cart styles in order lib

Pucek9:
hotfix

### DIFF
--- a/feature-libs/order/styles/components/order-history/_index.scss
+++ b/feature-libs/order/styles/components/order-history/_index.scss
@@ -8,7 +8,7 @@
 @import './return-request-totals';
 @import './cancel-order';
 @import './consignment-tracking';
-@import 'cart';
+@import '@spartacus/cart';
 
 $order-history-allowlist: cx-amend-order-actions, cx-amend-order-items,
   cx-cancel-order, cx-consignment-tracking, cx-tracking-events, cx-order-history,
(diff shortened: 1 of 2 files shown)


Platonn (approved):

feature-libs/order/styles/components/order-history/_index.scss
```diff
@@ -8,7 +8,7 @@
 @import './return-request-totals';
 @import './cancel-order';
 @import './consignment-tracking';
-@import 'cart';
+@import '@spartacus/cart';
```
bgambocjaviniar, line 11: 🔴

bgambocjaviniar (changes requested):
As discussed, we don't have to merge this PR as it's not needed ASAP.
Michal will find other solutions and PoC in this PR instead of providing the entire cart styles again within the order lib.

P.S do note adding the mapping in the webpack config and styles are not enough as the installation script fails at times still as seen in my PR previously. We would need to increase the bundle size budget to 3mb or 2.6mb as it sometimes go over the 2.5mb threshold by 15 kb, or so.

Screen Shot 2022-04-07 at 8 16 06 AM

feature-libs/order/styles/components/order-history/_index.scss
bgambocjaviniar: should place this at the top in case we're going to override something with common classes and selectors (if any)

however, this is not applicable anymore as we agreed that you're going to PoC and provide a granular solution instead of the entire lib

PR #15571, opened 2022-04-07
chore: Use classes instead of placeholder selectors for list items 

### LINKED ISSUE #15569: Broken schematics build (styles inheritance) after cart table refactor

Pucek9:
Closes #15569

### DIFF
--- a/feature-libs/cart/base/styles/components/_cart-item-list.scss
+++ b/feature-libs/cart/base/styles/components/_cart-item-list.scss
@@ -1,4 +1,4 @@
-%cx-cart-item-list {
+.cx-cart-item-list {
   @extend .cx-agnostic-table;
 
   td.cx-actions {
--- a/feature-libs/cart/base/styles/components/_index.scss
+++ b/feature-libs/cart/base/styles/components/_index.scss
@@ -17,7 +17,7 @@
 @import './clear-cart';
 
 $cart-components-allowlist: cx-mini-cart, cx-add-to-cart, cx-item-counter,
-  cx-added-to-cart-dialog, cx-cart-details, cx-cart-item, cx-cart-item-list,
+  cx-added-to-cart-dialog, cx-cart-details, cx-cart-item,
   cx-cart-totals, cx-cart-coupon, cx-applied-coupons, cx-save-for-later,
   cx-order-summary, cx-cart-validation-warnings, cx-cart-item-validation-warning,
   cx-cart-proceed-to-checkout, cx-clear-cart-dialog, cx-clear-cart !default;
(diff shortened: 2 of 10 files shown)


feature-libs/cart/base/styles/components/_index.scss
```diff
@@ -17,7 +17,7 @@
 @import './clear-cart';
 
 $cart-components-allowlist: cx-mini-cart, cx-add-to-cart, cx-item-counter,
-  cx-added-to-cart-dialog, cx-cart-details, cx-cart-item, cx-cart-item-list,
+  cx-added-to-cart-dialog, cx-cart-details, cx-cart-item,
```
bgambocjaviniar, line 20: we lose the option of making the style for a component 'toggleable', which I think is fine if we agree that this is the approach we want to take with our current limitation on our styles architecture 

P.S. When I mean toggleable, I mean won't let the users have the option to skip the styles for 'said' component

feature-libs/cart/base/styles/components/_cart-item-list.scss
```diff
@@ -1,4 +1,4 @@
-%cx-cart-item-list {
+.cx-cart-item-list {
```
bgambocjaviniar, line 1: I understand we have to go with this approach if we're going to have it re-used in each respective components

If we decide to go this approach, maybe leave a comment explaining why we're doing this until we have a better solution.

feature-libs/order/components/amend-order/amend-order-items/amend-order-items.component.ts
```diff
@@ -9,6 +9,7 @@ import { OrderAmendService } from '../amend-order.service';
   selector: 'cx-amend-order-items',
   templateUrl: './amend-order-items.component.html',
   changeDetection: ChangeDetectionStrategy.OnPush,
+  host: { class: 'cx-cart-item-list cx-amend-order-items' },
```
bgambocjaviniar, line 12: we no longer need `cx-amend-order-items` nor it's stylesheet right?

feature-libs/order/components/return-request-detail/return-request-items/return-request-items.component.ts
```diff
@@ -7,6 +7,7 @@ import { ReturnRequestService } from '../return-request.service';
   selector: 'cx-return-request-items',
   templateUrl: './return-request-items.component.html',
   changeDetection: ChangeDetectionStrategy.OnPush,
+  host: { class: 'cx-cart-item-list cx-return-request-items' },
```
bgambocjaviniar, line 10: we no longer need `cx-return-request-items` nor it's stylesheet right?

bgambocjaviniar (commented):
Looks like a clean solution, but it goes against what we're doing.

Let's take today and Friday to see if we can figure something else, but I'm likely to lean towards this solution if we won't place this logic in our global styles lib.
One thing with this approach is what if the customer does not use our cart library and just uses the order library, then the order 'cart items' won't be styled.

 Although, I believe this shouldn't be a use case and just a misconfiguration right?

PR #15582, opened 2022-04-08
chore: Add breaking change script/tool

### LINKED ISSUE #15581: Merge breaking change tool to develop
Create a new branch with just the tool and not all the migration files.

plabadie:


tools/breaking-changes/api-extractor.json
```diff
@@ -0,0 +1,341 @@
+/**
+ * Config file for API Extractor.  For more info, please visit: https://api-extractor.com
+ */
+{
+  "$schema": "https://developer.microsoft.com/json-schemas/api-extractor/v7/api-extractor.schema.json",
```
WeizhengSap, line 5: I think we can remove those comments in this file. First, we cannot add such comments in json; and that schema link really give all info about the file.

tools/breaking-changes/generate-const.ts
```diff
@@ -0,0 +1,116 @@
+import deepEqual from 'deep-equal';
```
WeizhengSap: Can we change the file name? From the name, I think it will generate schematics const variables.

tools/breaking-changes/clone.sh
```diff
@@ -0,0 +1,10 @@
+mkdir -p src
+git clone git@github.com:SAP/spartacus.git ./src/old
```
WeizhengSap, line 2: I have to run it: `git clone https://github.com/SAP/spartacus.git ./src/old`

tools/breaking-changes/readme.md
```diff
@@ -0,0 +1,85 @@
...
+
+- checkout the approproate commit in ./src/old and ./src/new  
+These will be the 2 versions compared for breaking changes.
+
+- Build Spartacus libs
+In both ./src/old and ./src/new, run `yarn install` and `yarn build:libs`.
+
+
+- Extract the publiic API.
+Run `yarn extract-all` in this folder.  This will extract the public api in ./src/*/temp folder into many files. (one per entry point)
```
WeizhengSap: in which folder?

tools/breaking-changes/clone.sh
plabadie: What do you mean?

tools/breaking-changes/api-extractor.json
plabadie: IT's meant do document the different configs we could apply to this file.   I thought it could be useful down the road.

tools/breaking-changes/clone.sh
WeizhengSap: I run this `clone.sh`, and got `Permission denied`.

tools/breaking-changes/api-extractor.json
WeizhengSap: Comments are not permitted in json; and we can get all info from that schema link, right?

tools/breaking-changes/generate-const.ts
plabadie: Right.  I wanted the name to be short, but I agree const might be confusing.

tools/breaking-changes/readme.md
plabadie: I added more details.

tools/breaking-changes/generate-const.ts
plabadie: Fixed

tools/breaking-changes/api-extractor.json
plabadie: The Microsoft api extractor tool generates the config file like this (with all the comments), so I presumed that having comments in there was fine.  I didn't see a compelling reason to erase everything.  If you feel strongly about, that's ok and we can discuss.

tools/breaking-changes/clone.sh
plabadie: I can't see why.  Could git have messed up the exec flag?
if you run `ls -l` , can you see the exec flag?
If you run `chmod 744 clone.sh` does it fix it?

plabadie: in the end this file is half documentation, half convenience.  There really are just 2 commands in there.

WeizhengSap (approved):

PR #15599, opened 2022-04-13
Reordered the scss files load

### LINKED ISSUE #15534: Default value of $font-url in styles.scss could not be overwritten by customers
**Describe the bug**

For certain reasons ( refer to this stackoverflow [discussion](https://stackoverflow.com/questions/71705078/sap-spartacus-sparta-theme-requesting-open-sans-by-default-not-dsgvo-complia) for detail ) some customers would like to overwrite the default value of variable `$font-url` in styles.scss. When customers set new value to variable `$font-url`, the new value does appear in compiled `style.css`. However, the default value `https://fonts.googleapis.com/css?family=Open+Sans:300,400,600,700` still remains in compiled `style.css`. As a result, both fonts will be loaded in the runtime. Customer would like the default font is NOT loaded any more.

- Library version: 4.1

**To Reproduce**
1. Create a new Storefront using schematics via command line:

`ng add @spartacus/schematics@4.1`

2. open styles.scss file in src folder, overwrite the `$font-url` value with below code:

```scss
/* You can add global styles to this file, and also import other style files */

$font-url:'https://fonts.googleapis.com/css?family=Tangerine';
$styleVersion: 4.1;
@import '~@spartacus/styles/index';
```

3. the default font url still exists in compiled `style.css`, as a result it is still loaded in the runtime:

image

image

**Expected behavior**

Only the customized font url is loaded in the runtime. The default font url, `https://fonts.googleapis.com/css?family=Open+Sans:300,400,600,700`, should **NOT** be loaded.

Dalvir-Personal:
fixes GH-15534


morganm58 (approved):

bgambocjaviniar (changes requested):
as discussed, please do not merge this re-order as it is not a viable solution

cart and order lib should be re-worked to not re-import the base styles

morganm58 (changes requested):
Brian proposed a solution that solves the problem without having potential future issues. cart and order libs are duplicating the styles

feature-libs/cart/_index.scss
```diff
@@ -20,3 +19,9 @@ $useLatestStyles: true;
 @import './quick-order/index';
 @import './wish-list/index';
 @import './import-export/index';
+
+@import 'cxbase/blocks/modal';
```
bgambocjaviniar, line 23: let's include the package path and place it under the spartacus/styles/scss/core to bundle them

`@import '~@spartacus/styles/scss/cxbase/blocks/modal';`

feature-libs/cart/_index.scss
```diff
@@ -20,3 +19,9 @@ $useLatestStyles: true;
...
+
+@import 'misc/hamburger';
```
bgambocjaviniar, line 25: Took a glance, but I don't think the cart styles is using anything from the hamburger stylesheet right?

feature-libs/cart/_index.scss
```diff
@@ -20,3 +19,9 @@ $useLatestStyles: true;
...
+@import 'misc/dialog';
```
bgambocjaviniar, line 26: Don't think it's being used anywhere in the cart styles

feature-libs/cart/_index.scss
```diff
@@ -20,3 +19,9 @@ $useLatestStyles: true;
...
+@import 'misc/table';
```
bgambocjaviniar, line 27: Same as previous. Let's include the package name

`@import '~@spartacus/styles/scss/misc/table';`

bgambocjaviniar: What do you think about just putting this import in `feature-libs/cart/base/styles/components/_cart-item-list.scss` 

as that's the only place that is using extending .cx-agnostic-table

OR 

do you think it should be at the top level?

feature-libs/order/styles/_index.scss
```diff
@@ -1,2 +1,2 @@
-@import '~@spartacus/styles';
+@import '~@spartacus/styles/scss/core';
```
bgambocjaviniar, line 1: I haven't checked yet, but is core the only necessary things for the order stylesheet.

It is using something from modal, tables, etc? Same reasoning as in the cart stylesheet

integration-libs/epd-visualization/_index.scss
```diff
@@ -1,10 +1,9 @@
-@import '~@spartacus/styles/scss/core';
-@import '~@spartacus/styles/scss/app';
```
bgambocjaviniar, line 2: Why did app get removed. Was there no instances of some re-used classes from our styles?

projects/storefrontstyles/scss/theme/sparta/_variables.scss
```diff
@@ -9,11 +9,13 @@
...
 
-//fonts (see _fonts.scss to import)
-$font-family-sans-serif: 'Open Sans', -apple-system, BlinkMacSystemFont,
-  'Segoe UI', 'Roboto', 'Helvetica Neue', Arial, sans-serif, 'Apple Color Emoji',
+$font-family: 'Open Sans', -apple-system, BlinkMacSystemFont, 'Segoe UI',
+  'Roboto', 'Helvetica Neue', Arial, sans-serif, 'Apple Color Emoji',
   'Segoe UI Emoji', 'Segoe UI Symbol';
 
+//fonts (see _fonts.scss to import)
+$font-family-sans-serif: $font-family !default;
```
bgambocjaviniar, line 17: is this necessary? If not, let's keep it as what it was before as I don't see a need to set it as a default from the issue at hand

projects/storefrontstyles/scss/theme/sparta/_variables.scss
```diff
@@ -54,7 +56,7 @@ $theme-colors: (
 );
 
 //type — https://github.com/twbs/bootstrap/blob/v4-dev/scss/_variables.scss#L225
-$font-size-base: 1rem; //16px
+$font-size-base: 1rem !default; //16px
```
bgambocjaviniar, line 59: same for above, do we really want it as default?

projects/storefrontstyles/scss/theme/sparta/_variables.scss
```diff
@@ -54,7 +56,7 @@ $theme-colors: (
...
 // $h1-font-size: $font-size-base * 2.375; //38px
 $h1-font-size: $font-size-base * 1.85; //36px
 $h2-font-size: $font-size-base * 1.75; //28px
```
bgambocjaviniar, line 62: @morganm58 

all that being said, please create a new branch with the proper conventions!

bgambocjaviniar (changes requested):
@morganm58 

the changes you applied looks good, but I would like the following to follow the convention we have in place

1 - make a new PR as all branches that is not a release or an epic are called feature. So name the branch `feature/GH-15534`
2 - @morganm58  after pushing a new branch for the one in develop (do not create those other branches that will point to maintenance branches as it just causes overhead to change down the road IF there will have additional changes). Leave the imports where they are and do not re-order them
3 - name the pull request as `style: Fixes re-import of entire spartacus/styles` <-- something along those lines so that when we merge, it will be there automatically without manually changing it

apply changes requested in other changes below (IF applicable)

integration-libs/epd-visualization/_index.scss
bgambocjaviniar: Same applies for cart and order. Have to make sure we didn't miss anything

bgambocjaviniar:
closing in favor for https://github.com/SAP/spartacus/pull/15619

PR #15600, opened 2022-04-13
Reordered css files load

### LINKED ISSUE #15534: Default value of $font-url in styles.scss could not be overwritten by customers

Dalvir-Personal:


morganm58 (dismissed):

bgambocjaviniar (changes requested):
as discussed, please do not merge this re-order as it is not a viable solution

cart and order lib should be re-worked to not re-import the base styles

bgambocjaviniar:
closing in favor for https://github.com/SAP/spartacus/pull/15619

PR #15607, opened 2022-04-14
fix: ASM lib doesn't allow for override styles

### LINKED ISSUE #15434: ASM lib doesn't allow for override styles
**Describe the bug**
Most libraries have separated `styles` part. Styling in ASM is placed together with components, and styles are not exported.

- Library version: 4

**Expected behavior**
We would like to have
`feature-libs/asm/_index.scss`
`feature-libs/asm/styles` with all files
and everything which is necessary to allow for overriding

**Additional context**
Please remember about usage placeholder selectors (%)

Pucek9:
Closes #15434

### DIFF
--- a/feature-libs/asm/package.json
+++ b/feature-libs/asm/package.json
@@ -11,11 +11,6 @@
   "homepage": "https://github.com/SAP/spartacus",
   "repository": "https://github.com/SAP/spartacus/tree/develop/feature-libs/asm",
   "license": "Apache-2.0",
-  "exports": {
-    ".": {
-      "sass": "./_index.scss"
-    }
-  },
   "scripts": {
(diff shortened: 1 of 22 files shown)


feature-libs/asm/package.json
```diff
@@ -11,6 +11,11 @@
   "homepage": "https://github.com/SAP/spartacus",
   "repository": "https://github.com/SAP/spartacus/tree/develop/feature-libs/asm",
   "license": "Apache-2.0",
+  "exports": {
+    ".": {
+      "sass": "./_index.scss"
+    }
+  },
```
Pucek9, line 18: with possible backport for 4.3.x, this fragment should not be added (breaks build)

znikola (approved):

PR #15613, opened 2022-04-20
Focus value for conflict solver navigation

### LINKED ISSUE #15604: Fix focus for conflict solver navigation

Larisa-Staroverova:


feature-libs/product-configurator/rulebased/components/attribute/header/configurator-attribute-header.component.ts
```diff
@@ -165,7 +165,18 @@ export class ConfiguratorAttributeHeaderComponent
...
+          );
+          if (selectedValue) {
+            const valueUiKey = this.createAttributeValueUiKey(
+              this.attribute.name,
+              selectedValue.valueCode
+            );
+            this.focusValue(this.attribute.name, valueUiKey);
+          } else {
+            this.focusValue(this.attribute.name);
+          }
```
Uli-Tiger: Cant't this code be part of the focusValue method? - especially as the navigateToGroup() method is quite lengthy and this is not about navigation but about focus handling.
So we would just call _this.focusAttribute(this.attribute)_ and then do the checking around selected value inside.

feature-libs/product-configurator/rulebased/components/service/configurator-storefront-utils.service.ts
```diff
@@ -138,12 +139,40 @@ export class ConfiguratorStorefrontUtilsService {
...
+
   /**
-   * Find an attribute by its name in the form and focus it.
+   * Find a value by its UI key in the form and focus it.
    *
-   * @param {string} name - Attribute name
+   * @param {string} attributeName - Attribute name
+   * @param {string} valueUiKey - Value UI key
    */
-  focusAttribute(name: string): void {
```
Uli-Tiger, line 146: isn't removing a method is an incompatible change - are you sure about this?

feature-libs/product-configurator/rulebased/components/service/configurator-storefront-utils.service.ts
```diff
@@ -138,12 +139,40 @@ export class ConfiguratorStorefrontUtilsService {
...
+  focusValue(attributeName: string, valueUiKey?: string): void {
```
Uli-Tiger: if we just keep the method name and add the optional parameter 'valueUiKey' the change would be compatible?

feature-libs/product-configurator/rulebased/components/attribute/header/configurator-attribute-header.component.spec.ts
```diff
@@ -781,7 +791,78 @@ describe('ConfigAttributeHeaderComponent', () => {
...
+      component.attribute.values = [value1, value2, value3];
+
+      spyOn(configurationGroupsService, 'navigateToGroup');
+      fixture.detectChanges();
+
+      component.navigateToGroup();
+      expect(configurationGroupsService.navigateToGroup).toHaveBeenCalledTimes(
+        1
+      );
+    });
```
Uli-Tiger: this test does not test the changed code. It is just a repetition of the test
_should navigate from conflict group to regular group that contains attribute which is involved in conflict_, which is written above.
i would expect that the test asserts that _focusValue_ was called with the expected arguments

feature-libs/product-configurator/rulebased/components/attribute/header/configurator-attribute-header.component.spec.ts
```diff
@@ -781,7 +791,78 @@ describe('ConfigAttributeHeaderComponent', () => {
...
```
Uli-Tiger: same as above. test does not test written code. expect focusValue was called is missing

feature-libs/product-configurator/rulebased/components/attribute/header/configurator-attribute-header.component.spec.ts
```diff
@@ -790,6 +871,11 @@ describe('ConfigAttributeHeaderComponent', () => {
       testScheduler.run(() => {
         component.groupType = Configurator.GroupType.CONFLICT_GROUP;
         component.attribute.groupId = ConfigurationTestData.GROUP_ID_2;
+        const value1 = createValue('1', 'val1', false);
+        const value2 = createValue('2', 'val2', false);
+        const value3 = createValue('3', 'val3', true);
+
+        component.attribute.values = [value1, value2, value3];
```
Uli-Tiger: i believe this setup can be removed, as it is not required for this test. you are calling focusValue directly, which just delegates to the focusService without touching the attribute.

feature-libs/product-configurator/rulebased/components/attribute/header/configurator-attribute-header.component.spec.ts
```diff
@@ -781,7 +791,78 @@ describe('ConfigAttributeHeaderComponent', () => {
...
+
+        component['focusValue'](component.attribute.name);
+      });
+
+      expect(
+        configuratorStorefrontUtilsService.focusValue
+      ).toHaveBeenCalledTimes(1);
+    });
+
+    it('should call focusValue with attribute name', () => {
```
Uli-Tiger: test has the same name as the test in 834 - since we are just testing that component.focusValue is delegating to the focusService one test of those can be deleted

Uli-Tiger (changes requested):
i believe we need to consider compatibility and  the component test need a little bit of rework.

feature-libs/product-configurator/rulebased/components/service/configurator-storefront-utils.service.ts
Larisa-Staroverova: This method occurred in the period as the conflict solver navigation had been implemented that's why no incompatible change is expected

feature-libs/product-configurator/rulebased/components/service/configurator-storefront-utils.service.ts
Larisa-Staroverova: According to my point of view the name of the method should be descriptive

feature-libs/product-configurator/rulebased/components/attribute/header/configurator-attribute-header.component.ts
Larisa-Staroverova: Good point! I will refactor `this.ocusValue(this.attribute)` method

feature-libs/product-configurator/rulebased/components/attribute/header/configurator-attribute-header.component.spec.ts
Larisa-Staroverova: Please, see my rework

feature-libs/product-configurator/rulebased/components/service/configurator-storefront-utils.service.spec.ts
```diff
@@ -278,27 +301,17 @@ describe('ConfigUtilsService', () => {
...
+        classUnderTest.focusValue(attribute);
         expect(keyboardFocusService.findFocusable).toHaveBeenCalledTimes(1);
       });
 
       it('should not delegate to keyboard focus service', () => {
         spyOn(windowRef, 'isBrowser').and.returnValue(false);
         spyOn(keyboardFocusService, 'findFocusable').and.callThrough();
-        classUnderTest.focusValue('ATTR_2');
+        classUnderTest.focusValue(attribute);
         expect(keyboardFocusService.findFocusable).toHaveBeenCalledTimes(0);
```
Uli-Tiger, line 388: I think just to check that _keyboardFocusService.findFocusable_ was called is not sufficient. it's one of the first things this method does, but its not about what the method actually is doing. 
the main purpose of _focusValue_ is to choose the correct element to focus on based on the given attribute and then to actually calling focus on this element.
so i would expect that we set up list of elements and check that focus is invoked on the expected element and not on the other elements

feature-libs/product-configurator/rulebased/components/service/configurator-storefront-utils.service.ts
```diff
@@ -181,20 +187,24 @@ export class ConfiguratorStorefrontUtilsService {
...
-                valueUiKey
-              )
-            : this.getFocusableElementByAttributeId(
-                focusableElements,
-                attributeName
-              );
-
+          );
+        }
+        if (foundFocusableElement === undefined) {
```
Uli-Tiger: why checking explicitly for undefined? - what about null values, i doubt we can focus on them, so maybe you want to change this line to `if (!foundFocusableElemen){`

feature-libs/product-configurator/rulebased/components/attribute/header/configurator-attribute-header.component.spec.ts
Uli-Tiger: still there is no need for values here. We just pass on the attribute, whether it has values or not. So i believe we can removed this along with the createValue method. Still you test will pass.

Uli-Tiger (changes requested):
Productive code looks cleaner now.
Would still like to see more detailed expectations in the test checking the expected behavior of the focusValue() method.

feature-libs/product-configurator/rulebased/components/service/configurator-storefront-utils.service.spec.ts
Larisa-Staroverova: Reworked!

ChristophHi (dismissed):

feature-libs/product-configurator/rulebased/components/service/configurator-storefront-utils.service.spec.ts
```diff
@@ -249,66 +270,84 @@ describe('ConfigUtilsService', () => {
...
-        spyOn(keyboardFocusService, 'findFocusable').and.callThrough();
+        attribute.name = 'NO_ATTR_2';
+        const focusedElement = spyFocus();
+
         classUnderTest.focusValue(attribute);
-        expect(keyboardFocusService.findFocusable).toHaveBeenCalledTimes(0);
+        expect(keyboardFocusService.findFocusable).toHaveBeenCalledTimes(1);
+        if (focusedElement) {
+          expect(focusedElement.focus).toHaveBeenCalled();
+        }
```
Uli-Tiger: why is this expectation within if-statement, i would have expected `expect(focusedElement.focus).not.toHaveBeenCalled()`

feature-libs/product-configurator/rulebased/components/service/configurator-storefront-utils.service.spec.ts
```diff
@@ -249,66 +270,84 @@ describe('ConfigUtilsService', () => {
...
+      });
 
+      it('should not set focus because attribute does not contain any values', () => {
+        spyOn(windowRef, 'isBrowser').and.returnValue(true);
+        const focusedElement = spyFocus();
         classUnderTest.focusValue(attribute);
         expect(keyboardFocusService.findFocusable).toHaveBeenCalledTimes(1);
+        if (focusedElement) {
+          expect(focusedElement.focus).toHaveBeenCalled();
+        }
```
Uli-Tiger: based on the test name i would have expect a `expect(focusedElement.focus).not.toHaveBeenCalled()`

feature-libs/product-configurator/rulebased/components/service/configurator-storefront-utils.service.spec.ts
```diff
@@ -249,66 +270,84 @@ describe('ConfigUtilsService', () => {
...
+        const value2 = createValue('value_2', true);
+        const value3 = createValue('value_3', false);
         attribute.values = [value1, value2, value3];
+        const focusedElement = spyFocus();
 
         classUnderTest.focusValue(attribute);
         expect(keyboardFocusService.findFocusable).toHaveBeenCalledTimes(1);
+        if (focusedElement) {
+          expect(focusedElement.focus).toHaveBeenCalled();
+        }
```
Uli-Tiger: why an if  statement and not just 
`expect(focusedElement.focus).toHaveBeenCalled()`

Uli-Tiger (changes requested):
I really like that spec 'should set focus because attribute contains selected value' now expects that focus method is called. that is an example for an excellent test. 
On little thing, instead of having expectation within an if statement, i like to ask you to remove and if and either expect that something is called or expect that something is not called.

Uli-Tiger (dismissed):

Uli-Tiger (approved):

PR #15616, opened 2022-04-20
chore: Cache node_modules to be used for the pipeline

bgambocjaviniar:


.github/workflows/cache-node-modules.yml
```diff
@@ -0,0 +1,35 @@
...
+      - 'epic/**'
+      - 'release/**'
+name: Cache node modules
+
+env:
+  NODE_VERSION: '16'
+
+jobs:
+  cacheNodeModules:   
+    name: Cache root node_modules
```
RadhepS: nitpick... think it can be 'cache node modules'? since we cache cypress ones that isnt on the root level?

bgambocjaviniar: It would be 'root' of cypress :P

RadhepS (approved):

PR #15628, opened 2022-04-21
feat: added to cart toast

bit172:
`default-cart-config.ts` requires some changes to make the feature active.

```typescript
    // ...
    addToCartFeedback: {
      feedback: ADDED_TO_CART_FEEDBACK.TOAST,
     // ...
    },

```


developpeurweb:
- Please make sure this works with JAWS/NVDA
- Also, keyboard-only users may not be able to reach the toast on time, please make sure the duration can be extended/configurable.

feature-libs/cart/base/assets/translations/en/cart.i18n.ts
```diff
@@ -76,4 +76,12 @@ export const cart = {
       'The product configuration requires additional entries in the back end. As a result, you cannot proceed. Please contact support.',
     inProgress: 'Processing',
   },
+  cartToast: {
+    entriesAdded: '{{ quantity }} items added',
```
plabadie: There seems to be a mechanism available to manage plurals.  Look for i18n keys that have the `_other`  suffix.  (I think is used to be `_plural` before.)

feature-libs/cart/base/components/add-to-cart/add-to-cart.component.ts
```diff
@@ -46,8 +46,11 @@ export class AddToCartComponent implements OnInit, OnDestroy {
   hasStock: boolean = false;
   inventoryThreshold: boolean = false;
 
-  showInventory$: Observable<boolean | undefined> | undefined =
-    this.component?.data$.pipe(map((data) => data.inventoryDisplay));
+  showInventory$:
+    | Observable<boolean | undefined>
+    | undefined = this.component?.data$.pipe(
+    map((data) => data.inventoryDisplay)
+  );
```
plabadie, line 53: What is the change in the code above about?

feature-libs/cart/base/components/add-to-cart/add-to-cart.component.ts
```diff
@@ -46,8 +46,11 @@ export class AddToCartComponent implements OnInit, OnDestroy {
...
```
plabadie, line 54: Overall, did only the format change in this file?

feature-libs/cart/base/components/added-to-cart-dialog/added-to-cart-dialog-event.listener.spec.ts
```diff
@@ -48,6 +50,12 @@ class MockModalService {
   }
 }
 
+const mockCartConfig: CartConfig = {
+  cart: {
+    addToCartFeedback: { feedback: ADDED_TO_CART_FEEDBACK.MODAL },
+  },
+};
+
```
plabadie, line 58: There could be coverage about what happens when the feedback config is not modal.

feature-libs/cart/base/components/added-to-cart-toast/added-to-cart-toast-component.service.ts
```diff
@@ -0,0 +1,54 @@
+import { Injectable } from '@angular/core';
+import { Observable, of } from 'rxjs';
+import { CartToastItem, OrderEntry } from '../../root';
```
plabadie: Fix this import path.  Should be `from '@spartacus/cart/base/root';`

feature-libs/cart/base/components/added-to-cart-toast/added-to-cart-toast-event.listener.ts
```diff
@@ -0,0 +1,67 @@
+import { ComponentRef, Injectable, OnDestroy } from '@angular/core';
+import { EventService } from '@spartacus/core';
+import { LaunchDialogService, LAUNCH_CALLER } from '@spartacus/storefront';
+import { Subscription } from 'rxjs';
+import { filter, take } from 'rxjs/operators';
+import {
+  ADDED_TO_CART_FEEDBACK,
+  CartConfig,
+  CartUiEventAddToCart,
+} from '../../root';
```
plabadie: Fix this import path. Should be from '@spartacus/cart/base/root';

feature-libs/cart/base/components/added-to-cart-toast/added-to-cart-toast.component.ts
```diff
@@ -0,0 +1,192 @@
...
+import { WindowRef } from '@spartacus/core';
+import { Observable, OperatorFunction, Subscription } from 'rxjs';
+import { filter, take } from 'rxjs/operators';
+import {
+  ActiveCartFacade,
+  CartToastItem,
+  CartUiEventAddToCart,
+  CART_TOAST_STATE,
+  OrderEntry,
+} from '../../root';
```
plabadie: Fix this import path. Should be from '@spartacus/cart/base/root';

projects/storefrontlib/layout/launch-dialog/config/launch-config.ts
```diff
@@ -90,4 +90,5 @@ export const enum LAUNCH_CALLER {
   ANONYMOUS_CONSENT = 'ANONYMOUS_CONSENT',
   REPLENISHMENT_ORDER = 'REPLENISHMENT_ORDER',
   PLACE_ORDER_SPINNER = 'PLACE_ORDER_SPINNER',
+  ADDED_TO_CART_TOAST = 'ADDED_TO_CART_TOAST',
```
plabadie: We should use type augmentation and declare this value in the cart lib.
Search for `const enum LAUNCH_CALLER` for examples.

projects/storefrontapp/src/app/spartacus/features/added-to-cart-toast-feature.module.ts
```diff
@@ -0,0 +1,32 @@
...
+      },
+      i18n: {
+        resources: addedToCartToastTranslations,
+        chunks: addedToCartToastTranslationChunksConfig,
+        fallbackLang: 'en',
+      },
+    }),
+  ],
+})
+export class AddedToCartToastFeatureModule {}
```
plabadie: I don't think the cart toast should be a standalone feature.  Rather, it should be part of the cart base like the dialog.  Let's discuss.

feature-libs/cart/base/styles/components/_added-to-cart-toast.scss
```diff
@@ -0,0 +1,101 @@
+:root {
+  --cx-added-to-cart-toast-width: 375px;
+  --cx-added-to-cart-toast-header-offset: 192px;
+  --cx-added-to-cart-toast-background: var(--cx-color-inverse);
+  --cx-added-to-cart-toast-translateY: -100%;
+}
+
+cx-added-to-cart-toast {
```
plabadie, line 8: Many component style definition start with a `%`, would that apply here?

feature-libs/cart/base/core/services/cart-config.service.spec.ts
```diff
@@ -20,7 +20,12 @@ describe('CartConfigService', () => {
 
   describe('isSelectiveCartService', () => {
     it('should return true if selectiveCart enabled is set to true', () => {
-      mockCartConfig.cart = { selectiveCart: { enabled: true } };
+      mockCartConfig.cart = {
+        selectiveCart: { enabled: true },
+        addToCartFeedback: {
+          feedback: ADDED_TO_CART_FEEDBACK.MODAL,
+        },
```
plabadie, line 27: I don't think the add to cart feedback should be mocked in this spec.  It doesn't seem related to the test.

feature-libs/cart/base/components/added-to-cart-toast/added-to-cart-toast.module.ts
```diff
@@ -0,0 +1,18 @@
...
+import { MediaModule } from '@spartacus/storefront';
+import { AddedToCartToastEventListener } from './added-to-cart-toast-event.listener';
+import { AddedToCartToastComponent } from './added-to-cart-toast.component';
+import { defaultAddedToCartToastLayoutConfig } from './default-added-to-cart-toast-layout.config';
+
+@NgModule({
+  imports: [CommonModule, MediaModule, RouterModule, UrlModule, I18nModule],
+  declarations: [AddedToCartToastComponent],
+  exports: [AddedToCartToastComponent],
+  providers: [provideConfig(defaultAddedToCartToastLayoutConfig)],
```
plabadie: To provide default config we should use `provideDefaultConfig`

feature-libs/cart/base/components/added-to-cart-toast/added-to-cart-toast-event.listener.ts
```diff
@@ -0,0 +1,67 @@
...
+        this.component.timeout = timeout ? timeout : 3000;
+        this.component.init();
+      });
+  }
+
+  protected addToast(event: CartUiEventAddToCart): void {
+    if (!this.component) {
+      this.renderToastUi();
+    }
+    this.component?.addToast(event);
```
plabadie, line 60: I think the event shoult not leave the event listener class.  It would be a looser coupling and better encapsulate the purpose of the even listener.  Data should be taken out of the event and passed as arguments to the component.

feature-libs/cart/base/components/added-to-cart-toast/added-to-cart-toast-event.listener.ts
```diff
@@ -0,0 +1,67 @@
...
+      .pipe(filter(Boolean), take(1))
+      .subscribe((component: ComponentRef<AddedToCartToastComponent>) => {
+        this.component = component.instance;
+        this.component.timeout = timeout ? timeout : 3000;
+        this.component.init();
+      });
+  }
+
+  protected addToast(event: CartUiEventAddToCart): void {
+    if (!this.component) {
```
plabadie, line 57: Why do we attempt reusing the component.  The add to cart dialog seems to dismiss everything when the dialog is closed and does not have to manage the component instance this way.  Lets discuss

feature-libs/cart/base/components/added-to-cart-toast/added-to-cart-toast.component.ts
```diff
@@ -0,0 +1,192 @@
...
+    this.setOffsetHeight();
+
+    this.activeCartService
+      .getLastEntry(event.productCode)
+      .pipe(
+        filter(Boolean) as OperatorFunction<OrderEntry | undefined, OrderEntry>,
+        take(1)
+      )
+      .subscribe((entries) => {
+        let toastContainerClass = `${this.toastContainerBaseClass} `;
```
plabadie, line 79: could `toastContainerClass` be a const?  If so, perhaps it can be not used at all if the string is used on once place.

feature-libs/cart/base/components/added-to-cart-toast/added-to-cart-toast.component.ts
```diff
@@ -0,0 +1,192 @@
...
+
+        // trigger enter animations with class change
+        toastItem.baseClass = this.getToastStyles();
+
+        if (this.scrollEventUnlistener) {
+          this.scrollEventUnlistener();
+        }
+
+        // wait for enter animation end
+        setTimeout(() => {
```
plabadie, line 96: I'd like to understand this part.  Lets discuss.

feature-libs/cart/base/components/added-to-cart-toast/added-to-cart-toast.component.ts
```diff
@@ -0,0 +1,192 @@
...
+  }
+
+  closeToast(toastItem: CartToastItem) {
+    if (toastItem.timeoutRef) {
+      clearTimeout(toastItem.timeoutRef);
+    }
+    this._closeToast(toastItem);
+  }
+
+  _closeToast = (toastItem: CartToastItem) => {
```
plabadie, line 147: why the underscore?

feature-libs/cart/base/root/models/cart.model.ts
```diff
@@ -191,3 +192,28 @@ export enum CartValidationStatusCode {
   PRICING_ERROR = 'pricingError',
   UNRESOLVABLE_ISSUES = 'unresolvableIssues',
 }
+
+export interface AddToCartToastConfig {
+  timeout?: number;
+}
+
+export enum ADDED_TO_CART_FEEDBACK {
```
plabadie, line 196: Naming suggestion: `ADDED_TO_CART_FEEDBACK_TYPE`

feature-libs/cart/base/root/models/cart.model.ts
```diff
@@ -191,3 +192,28 @@ export enum CartValidationStatusCode {
...
+  productName: string;
+  quantity: number;
+  unitPrice: string;
+  baseClass: string;
+  images?: Images;
+  timeoutRef?: ReturnType<typeof setTimeout>;
+  scrollUnlistener?: () => void;
+}
+
+export enum CART_TOAST_STATE {
```
plabadie, line 211: CART_TOAST_STATE doesn't seem like a model that has a use outside of the AddToCartToastComponent.

feature-libs/cart/base/root/models/cart.model.ts
```diff
@@ -191,3 +192,28 @@ export enum CartValidationStatusCode {
...
+export interface AddToCartToastConfig {
+  timeout?: number;
+}
+
+export enum ADDED_TO_CART_FEEDBACK {
+  MODAL,
+  TOAST,
+  NONE,
+}
+export interface CartToastItem {
```
plabadie, line 201: If CartToastItem is only used with the AddToCartToastComponent and its service, perhaps we put `CartToastItem` in a model file next to these classes.  Lets discuss.

feature-libs/cart/base/root/models/cart.model.ts
```diff
@@ -191,3 +192,28 @@ export enum CartValidationStatusCode {
   PRICING_ERROR = 'pricingError',
   UNRESOLVABLE_ISSUES = 'unresolvableIssues',
 }
+
+export interface AddToCartToastConfig {
+  timeout?: number;
+}
```
plabadie: I think the convention is to declare the config types in the feature-config.ts file, like cart-config.ts.
If this is the case, declare `AddToCartToastConfig` in cart-config.ts

feature-libs/cart/base/root/models/cart.model.ts
plabadie: [2022/06/07]  - I think `ADDED_TO_CART_FEEDBACK` should be declared in cart-config.ts as well.

feature-libs/cart/base/components/add-to-cart/add-to-cart.component.ts
bit172: Looks like the formatting changed. I must have saved this file and formatted it.

feature-libs/cart/base/components/add-to-cart/add-to-cart.component.ts
bit172: Right, only the formatting changed here

projects/storefrontapp/src/app/spartacus/features/added-to-cart-toast-feature.module.ts
bit172: Looks like an old file I forgot to remove

feature-libs/cart/base/core/services/cart-config.service.spec.ts
bit172: I added this since the test was failing, due to a lack of default behavior when no feedback config is provided.

feature-libs/cart/base/root/models/cart.model.ts
```diff
@@ -191,3 +192,24 @@ export enum CartValidationStatusCode {
...
 }
+
+export enum ADDED_TO_CART_FEEDBACK {
+  MODAL,
+  TOAST,
+  NONE,
+}
+export interface CartToastItem {
+  productName: string;
+  quantity: number;
```
plabadie, line 203: [2022/06/07]  - Note for discussion: do we keep the NONE option

feature-libs/cart/base/components/added-to-cart-toast/added-to-cart-toast.component.ts
```diff
@@ -0,0 +1,192 @@
...
+    protected winRef: WindowRef,
+    protected renderer: Renderer2,
+    protected cd: ChangeDetectorRef,
+    protected addedToCartToastService: AddedToCartToastComponentService,
+    protected activeCartService: ActiveCartFacade,
+    protected el: ElementRef
+  ) {}
+
+  init(): void {
+    if (!this.customClass) {
```
plabadie: [2022/06/07]  - Concept of custom class doesn't seem to be used in the end.  If so, we consider a cleanup and remove it.

feature-libs/cart/base/components/added-to-cart-toast/added-to-cart-toast.component.ts
```diff
@@ -0,0 +1,192 @@
...
+  CartToastItem,
+  CartUiEventAddToCart,
+  CART_TOAST_STATE,
+  OrderEntry,
+} from '@spartacus/cart/base/root';
+import { AddedToCartToastComponentService } from './added-to-cart-toast-component.service';
+
+@Component({
+  selector: 'cx-added-to-cart-toast',
+  templateUrl: './added-to-cart-toast.component.html',
```
plabadie, line 23: [2022/06/07]  - Unless there is a reason not to, the component should update on push.

feature-libs/cart/base/components/added-to-cart-toast/added-to-cart-toast.component.ts
plabadie: [2022/06/07]  - Answer: because of a name clash.  There is probably a way to handle this in a less confusing way.

plabadie (changes requested):

feature-libs/cart/base/styles/components/_added-to-cart-toast.scss
plabadie: Actually, now I know that this class should start with a % to make it an extensible sass abstract class.  Please implemente it as such.

feature-libs/cart/base/components/added-to-cart-toast/added-to-cart-toast.component.ts
```diff
@@ -0,0 +1,202 @@
...
+    protected winRef: WindowRef,
+    protected renderer: Renderer2,
+    protected cd: ChangeDetectorRef,
+    protected addedToCartToastService: AddedToCartToastComponentService,
+    protected activeCartService: ActiveCartFacade,
+    protected el: ElementRef
+  ) {}
+
+  init(): void {
+    this.baseClass = 'cx-added-to-cart-toast';
```
plabadie, line 55: This baseClass thing doesn't look right.  Components are styled via a scss file for the component and are targetd via their selector anyways.  I don't understand the purpose of this implementation and it seems like the base styling should be done following the Spartacus convention.  ( unless there is a good reason not to do so in this case. )

feature-libs/cart/base/components/added-to-cart-toast/added-to-cart-toast.component.ts
```diff
@@ -0,0 +1,202 @@
...
+
+        // trigger enter animations with class change
+        toastItem.baseClass = this.getToastStyles();
+
+        if (this.scrollEventUnlistener) {
+          this.scrollEventUnlistener();
+        }
+
+        // wait for enter animation end
+        setTimeout(() => {
```
plabadie, line 96: There are 2 timeouts in the component and a third on in the component service.  This seems like a lot of timeouts for a toast that simply needs to disappear after a given delay.  Are they all required?  Can the implementation be simplified?

feature-libs/cart/base/components/added-to-cart-toast/added-to-cart-toast.component.ts
```diff
@@ -0,0 +1,202 @@
...
+
+  headerElement: HTMLElement | null;
+
+  toastContainerClass: string;
+
+  timeout: number;
+
+  toastContainerBaseClass = 'toast-container';
+
+  lastToastItem: CartToastItem;
```
plabadie, line 36: The goal of handling one toast (because we don't need more anyways) was to simplify the implementation to reflect the simple UI requirement where at no time we need more than one toast to show.

While on the UI we display only one toast at a time, the implementation add more logic on top of the existing code, which is the opposite of the simplification we were seeking.

For example, we should consider all the arrays and the loops in the toast handling.  We could start with the component service and then the places where it is called.

feature-libs/cart/base/components/added-to-cart-toast/added-to-cart-toast-event.listener.spec.ts
```diff
@@ -0,0 +1,91 @@
...
+          useValue: mockCartConfig,
+        },
+      ],
+    });
+
+    listener = TestBed.inject(AddedToCartToastEventListener);
+    launchDialogService = TestBed.inject(LaunchDialogService);
+  });
+
+  describe('onAddToCart', () => {
```
plabadie, line 71: We should also cover that the toast is not displayed if the config is not set to display the toast

feature-libs/cart/base/components/added-to-cart-toast/added-to-cart-toast-event.listener.ts
```diff
@@ -0,0 +1,67 @@
...
+
+    this.subscription.add(
+      this.eventService.get(CartUiEventAddToCart).subscribe((event) => {
+        this.addToast(event);
+      })
+    );
+  }
+
+  protected renderToastUi(): void {
+    const timeout = this.cartConfig.cart?.addToCartFeedback.toast?.timeout;
```
plabadie, line 42: Timeout config should likely be encapsulated in the component class.  Ther doesn't seem to be an added value to pass it down from here, it's always the same value.

feature-libs/cart/base/components/added-to-cart-toast/added-to-cart-toast.component.html
```diff
@@ -0,0 +1,29 @@
+<ng-container *ngIf="cartToasts$ | async as cartToasts">
```
plabadie, line 1: How is accessibility handled?

The recommendation from the a11y guys was to use the "assistive message directive" somehow.

plabadie (changes requested):
See PR comments for requested changes or things to discuss.

PR #15637, opened 2022-04-22
feature/cxspa-299: [POC] add context in InnerComponentsHostDirective

### LINKED ISSUE #14332: Context directive
Currently we have different mechanisms of providing context for specific features. Some of them have limitations, they are not ergonomic, and it's a hassle to add new: 
- CartItemContext + CartItemContextSource
- ProductListItemContext + ProductListItemContextSource
- OrderEntriesContext
- generic OutletContext (which has limitations, that's why above 3 were created)

What we need to standardize our approach of passing context to components: child / inner / outlet. 

The benefits:
- having one standardized way of providing and consuming contexts in the extendable parts of the UI
- having an ergonomic API to define the context in the html template (via a directive)
- it's a missing piece for current UI extenisbility mechanisms: Outlets and InnerComponents directives 
  - we need those extensibility mechanisms being fully-fledged to solve our problems when splitting code into feature libraries
  - they might help us to expose better fine gained IU extensibility of the Cart component and Product Listing component (which was always a pain for customers, becasue they had to replace the whole CMS components)
  - if we need to support other CMS backends and operate on more generic layout data structures (e.g. allowing for deep nesting of slots/components), above extensibility mechanisms might serve well

Requirements [DRAFT]:
- We need a way of providing a context for the child and descendant components, independently from the standard Angular inputs and outputs, so the context is passed deep in the components tree.
- Going forward with development, adding new parts of the context shouldn't be hassle.
- We need a distinction of the context source and the readonly context (i.e. consumers shouldn't be able amend directly the context provided by the ancestor component)
- We need the ability of descendant components communicating back with the ancestor component
- ergonomic API to define and consume the context
- we should be able to extend the ancestor's context in the child component, so it can pass the enhanced context to its descendants

WeizhengSap:
close #14332


projects/storefrontlib/cms-components/product/product-list/product-list-item/product-list-item.component.html
```diff
@@ -50,7 +50,10 @@ <h2 [innerHtml]="product.nameHtml"></h2>
       </div>
       <div class="col-12 col-md-4">
         <ng-template [cxOutlet]="ProductListOutlets.ITEM_ACTIONS">
-          <ng-container cxInnerComponentsHost></ng-container>
+          <ng-container
+            cxInnerComponentsHost
+            [cxInnerComponentsContext]="{ product: product }"
```
plabadie: use shortcut? `{ product}`

projects/storefrontlib/cms-structure/page/model/cms-component-context.ts
```diff
@@ -0,0 +1,8 @@
+import { Observable } from 'rxjs';
+
+export abstract class CmsComponentContextData<T = any> {
+  /**
+   * Provides cms component context as an observable
+   */
+  context$: Observable<T>;
```
plabadie: Use `<T = unknown>`

feature-libs/cart/base/components/add-to-cart/add-to-cart.component.ts
```diff
@@ -63,7 +63,7 @@ export class AddToCartComponent implements OnInit, OnDestroy {
     protected activeCartService: ActiveCartFacade,
     protected component: CmsComponentData<CmsAddToCartComponent>,
     protected eventService: EventService,
-    @Optional() protected productListItemContext?: ProductListItemContext
+    @Optional() protected context?: CmsComponentContextData
```
plabadie: Add type safety

WeizhengSap: `CmsComponentContextData<Product>`

projects/storefrontlib/cms-components/product/product-list/product-list-item/product-list-item.component.html
WeizhengSap: or just `product`, not wrappered.

Platonn: In general, for being future-proof and extendable, I'm all for wrapping contexts and any payloads inside the wrapper objects. Why: we never know if in the future we (or customer) want to pass more data entities into the payload

PR #15650, opened 2022-04-26
ref types and async handling in auth-state-persistence.service

jolle22:
Refactor the usage of Subject and Observables in the auth-state-persistence.service


Pucek9:
We dont use "$" for methods which returns observable, only properties use this naming convention

jolle22:
> We dont use "$" for methods which returns observable, only properties use this naming convention

Thank you for the feedback. I Removed it, along with the related changes :)

PR #15658, opened 2022-04-26
feature/cxspa-265: enable strict mode for storefrontlib/layout

WeizhengSap:
enable strict mode for `storefrontlib/layout`


projects/storefrontlib/layout/breakpoint/breakpoint.service.ts
```diff
@@ -167,7 +167,7 @@ export class BreakpointService {
   }
 
   protected getMinSize(breakpoint: BREAKPOINT): number {
-    return (this.config[breakpoint] as BreakPoint)?.min;
+    return (this.config[breakpoint] as BreakPoint)?.min ?? 0;
```
bgambocjaviniar: I don't think this should be returning 0. I think it should return null similarly to getMaxSize. Reason being in line 134 of this file, we sometimes do not want to return anything and if it is null, then it will be 0. However, if we set this 'default' value, then the behavior will change. It will no longer be 0, but 0 + 1.

projects/storefrontlib/layout/a11y/keyboard-focus/persist/persist-focus.service.ts
```diff
@@ -18,15 +18,15 @@ export class PersistFocusService extends BaseFocusService {
...
+  get(group?: string | null): string | undefined {
     return this.focus.get(group || GLOBAL_GROUP);
   }
 
   /**
    * Persist the keyboard focus state for the given key. The focus is stored globally
    * or for the given group.
    */
-  set(key: string, group?: string) {
+  set(key?: string, group?: string | null) {
```
bgambocjaviniar: instead of making the `key` optional, let's still make it mandatory.

But make the type `string | undefined`. I don't think it makes sense just to allow customers to do .set(), which won't do anything in the end it won't go through the if condition

projects/storefrontlib/layout/a11y/keyboard-focus/autofocus/auto-focus.service.ts
```diff
@@ -14,9 +14,9 @@ export class AutoFocusService extends EscapeFocusService {
    * Returns the first focusable child element of the host element.
    */
   findFirstFocusable(
-    host: HTMLElement,
+    host: HTMLElement | null | undefined,
```
bgambocjaviniar, line 17: I removed undefined from the type, and it still works. Why is it here?

projects/storefrontlib/layout/a11y/keyboard-focus/autofocus/auto-focus.service.ts
```diff
@@ -29,7 +29,10 @@ export class AutoFocusService extends EscapeFocusService {
   /**
    * Indicates whether any of the focusable child elements is focused.
    */
-  hasPersistedFocus(host: HTMLElement, config: PersistFocusConfig): boolean {
+  hasPersistedFocus(
+    host: HTMLElement | null | undefined,
```
bgambocjaviniar, line 33: I removed undefined from the type, and it still works. Why is it here?

projects/storefrontlib/layout/a11y/keyboard-focus/autofocus/auto-focus.service.ts
```diff
@@ -40,15 +43,18 @@ export class AutoFocusService extends EscapeFocusService {
    * @param group the optional group for the persistent state, to separate different focus
    *   groups and remain the persistence
    */
-  protected getPersisted(host: HTMLElement, group?: string): HTMLElement {
+  protected getPersisted(
+    host?: HTMLElement | null,
```
bgambocjaviniar, line 47: I don't think we should make host optional. Similar reason as [this](https://github.com/SAP/spartacus/pull/15658/files#r870749898)

projects/storefrontlib/layout/a11y/keyboard-focus/lock/lock-focus.directive.ts
```diff
@@ -129,10 +129,11 @@ export class LockFocusDirective
        * we persist the group key to the children, so that they can taken this
        * into account when they persist their focus state.
        */
-      if (!!this.group) {
+      if (this.group) {
```
bgambocjaviniar: We can keep `!!this.group`. It just helps convert it to a boolean in the end whether it is an empty string, null, or undefined, it won't ever go inside the condition

projects/storefrontlib/layout/a11y/keyboard-focus/persist/persist-focus.service.ts
```diff
@@ -43,7 +43,10 @@ export class PersistFocusService extends BaseFocusService {
    * Returns the group for the host element based on the configured group or
    * by the `data-cx-focus-group` attribute stored on the host.
    */
-  getPersistenceGroup(host: HTMLElement, config?: PersistFocusConfig): string {
-    return config?.group ? config.group : host.getAttribute(FOCUS_GROUP_ATTR);
+  getPersistenceGroup(
+    host: HTMLElement | undefined | null,
```
bgambocjaviniar, line 47: I removed undefined from the type, and it still works. Why is it here?

projects/storefrontlib/layout/a11y/keyboard-focus/services/select-focus.util.ts
```diff
@@ -26,19 +26,19 @@ export class SelectFocusUtility {
   protected focusableSelectorSuffix =
     ':not([disabled]):not([hidden]):not([aria-hidden])';
 
-  query(host: HTMLElement, selector: string): HTMLElement[] {
+  query(host: HTMLElement | null | undefined, selector: string): HTMLElement[] {
```
bgambocjaviniar, line 29: I removed undefined from the type, and it still works. Why is it here?

projects/storefrontlib/layout/a11y/keyboard-focus/services/select-focus.util.ts
```diff
@@ -26,19 +26,19 @@ export class SelectFocusUtility {
...
     }
     return Array.from(
-      host.querySelectorAll(selector) as NodeListOf<HTMLElement>
+      host?.querySelectorAll(selector) as NodeListOf<HTMLElement>
     );
   }
 
   findFirstFocusable(
-    host: HTMLElement,
+    host: HTMLElement | null | undefined,
```
bgambocjaviniar, line 39: I removed undefined from the type, and it still works. Why is it here?

projects/storefrontlib/layout/a11y/keyboard-focus/services/select-focus.util.ts
```diff
@@ -57,7 +57,7 @@ export class SelectFocusUtility {
    * @param invisible indicates whether hidden focusable elements should be returned
    */
   findFocusable(
-    host: HTMLElement,
+    host: HTMLElement | null | undefined,
```
bgambocjaviniar, line 60: I removed undefined from the type, and it still works. Why is it here?

projects/storefrontlib/layout/a11y/keyboard-focus/trap/trap-focus.service.ts
```diff
@@ -57,11 +57,11 @@ export class TrapFocusService extends TabFocusService {
     }
   }
 
-  protected getTrapStart(trap: TrapFocusType): boolean {
+  protected getTrapStart(trap?: TrapFocusType): boolean {
```
bgambocjaviniar: I don't think we should make trap optional. Similar reason as [this](https://github.com/SAP/spartacus/pull/15658/files#r870749898)

projects/storefrontlib/layout/a11y/keyboard-focus/trap/trap-focus.service.ts
```diff
@@ -57,11 +57,11 @@ export class TrapFocusService extends TabFocusService {
...
     return trap === true || trap === TrapFocus.start;
   }
 
-  protected getTrapEnd(trap: TrapFocusType): boolean {
+  protected getTrapEnd(trap?: TrapFocusType): boolean {
     return trap === true || trap === TrapFocus.end;
   }
```
bgambocjaviniar, line 66: I don't think we should make trap optional. Similar reason as [this](https://github.com/SAP/spartacus/pull/15658/files#r870749898)

projects/storefrontlib/layout/a11y/keyboard-focus/trap/trap-focus.service.ts
```diff
@@ -57,11 +57,11 @@ export class TrapFocusService extends TabFocusService {
     }
   }
 
-  protected getTrapStart(trap: TrapFocusType): boolean {
+  protected getTrapStart(trap?: TrapFocusType): boolean {
     return trap === true || trap === TrapFocus.start;
   }
 
-  protected getTrapEnd(trap: TrapFocusType): boolean {
+  protected getTrapEnd(trap?: TrapFocusType): boolean {
```
bgambocjaviniar: I don't think we should make trap optional. Similar reason as [this](https://github.com/SAP/spartacus/pull/15658/files#r870749898)

projects/storefrontlib/layout/a11y/skip-link/service/skip-link.service.ts
```diff
@@ -51,34 +51,33 @@ export class SkipLinkService {
     const target =
       skipLink.target instanceof HTMLElement
         ? skipLink.target
-        : (skipLink.target as Element).parentElement;
+        : (skipLink.target as Element | undefined)?.parentElement;
```
bgambocjaviniar, line 54: instead of as Element | undefined, let's make it as unknown as Element. 

Reason being, we do not know what the target is actually, but we know it is some sort of element that may have a .parentElement.

---
We have a few instances that follows this practice in the codebase for setting types like these. If you search for as unknown as, you'll find some instances

projects/storefrontlib/layout/direction/direction.service.ts
```diff
@@ -81,7 +81,7 @@ export class DirectionService implements OnDestroy {
    * Sets the direction attribute for the given element. If the direction is undefined, the `dir`
    * attribute is removed.
    */
-  setDirection(el: HTMLElement, direction: DirectionMode): void {
+  setDirection(el: HTMLElement, direction?: DirectionMode): void {
```
bgambocjaviniar: I don't think we should make direction optional. Similar reason as [this](https://github.com/SAP/spartacus/pull/15658/files#r870749898)

projects/storefrontlib/layout/loading/defer-loader.service.ts
```diff
@@ -48,7 +47,7 @@ export class DeferLoaderService {
   }
 
   private shouldLoadInstantly(
-    elementLoadingStrategy: DeferLoadingStrategy
+    elementLoadingStrategy?: DeferLoadingStrategy
```
bgambocjaviniar: I don't think we should make elementLoadingStrategy optional. Similar reason as [this](https://github.com/SAP/spartacus/pull/15658/files#r870749898)

projects/storefrontlib/layout/theme/theme.service.ts
```diff
@@ -27,10 +27,10 @@ export class ThemeService {
     // Theme value is a string. It is put in the generic multi-value
     // property of the SiteContextConfig. So the array's first item
     // is the theme value.
-    this.setTheme(this.config.context[THEME_CONTEXT_ID]?.[0]);
+    this.setTheme(this.config.context?.[THEME_CONTEXT_ID]?.[0]);
   }
 
-  setTheme(theme: string): void {
+  setTheme(theme?: string): void {
```
bgambocjaviniar: I don't think we should make theme optional. Similar reason as [this](https://github.com/SAP/spartacus/pull/15658/files#r870749898)

projects/storefrontlib/router/on-navigate.service.ts
```diff
@@ -54,11 +54,10 @@ export class OnNavigateService {
           const previousRoute = event[0];
           const currentRoute = event[1];
 
-          if (currentRoute.position) {
+          const position = currentRoute.position;
```
bgambocjaviniar, line 57: not part of layout, but it works 👍

bgambocjaviniar (changes requested):
I have made in another branch the 'layout' folder of storefrontlib stricter as well as I needed to understand a few reasons why some things needed to be null / undefined.

Let me know if there's anything wrong with the direction on going optional param vs required param.

projects/storefrontlib/layout/a11y/keyboard-focus/autofocus/auto-focus.service.ts
WeizhengSap: It is also called by `scrollToTarget` in `skip-link.service.ts`. `target` can be undefined

projects/storefrontlib/layout/a11y/keyboard-focus/autofocus/auto-focus.service.ts
WeizhengSap: that's becuse `host` in `findFirstFocusable` is `HTMLElement | null | undefined`.

projects/storefrontlib/layout/a11y/keyboard-focus/autofocus/auto-focus.service.ts
WeizhengSap: `host` can be `HTMLElement | null | undefined`. When the parameter is `| undefined`, I will make it optional.

projects/storefrontlib/layout/a11y/keyboard-focus/persist/persist-focus.service.ts
WeizhengSap: as reason for `host`

projects/storefrontlib/layout/a11y/keyboard-focus/services/select-focus.util.ts
WeizhengSap: same host

projects/storefrontlib/layout/a11y/keyboard-focus/services/select-focus.util.ts
WeizhengSap: same host

projects/storefrontlib/layout/a11y/skip-link/service/skip-link.service.ts
WeizhengSap: skipLink.target is optional. It can be `undefined`. `as unknown as` is similar to `as any`. It is unsafe.

projects/storefrontlib/layout/breakpoint/breakpoint.service.ts
WeizhengSap: It should be `-1` because of this `this.getMinSize(prev) + 1 || 0`. Or return `null`, but need check it when `this.getMinSize(next) + 1`. To make it consistent with `getMaxSize`, I will make it `null`.

projects/storefrontlib/layout/a11y/skip-link/service/skip-link.service.ts
bgambocjaviniar: I think we had a discussion about this a while back, but it is 'safer' to use unknown than any. They're not necessarily the same as when you define something as 'unknown', it enforces a type check. https://dmitripavlutin.com/typescript-unknown-vs-any/#2-the-mental-model-of-unknown-vs-any

Basically, before using something that is 'unknown', it would need to type-check before using it.

projects/storefrontlib/layout/breakpoint/breakpoint.service.ts
bgambocjaviniar: unresolving, because ` getSize(breakpoint: BREAKPOINT): number`  is now complaining for strict type on compilation. 

Reason being we made getMinSize potentially return null as well, so getSize need to be typed as `number | null`, which is fine because if you look at the test case in the spec file, it could have been falsy (which was returning null) https://github.com/SAP/spartacus/blob/develop/projects/storefrontlib/layout/breakpoint/breakpoint.service.spec.ts#L190. But need to handle another place for the breakpoint when we type getSize as `number | null` as well.

projects/storefrontlib/layout/a11y/skip-link/service/skip-link.service.ts
bgambocjaviniar: After verifying again, I looked at the docs https://developer.mozilla.org/en-US/docs/Web/API/Node/parentElement, and it is just the Node Element that has parentElement, so it is fine to keep this way

Although, I also agree that `as unknown as` is just making it `as any` in the end.

projects/storefrontlib/layout/a11y/keyboard-focus/autofocus/auto-focus.service.ts
bgambocjaviniar: makes sense! it ties to this here https://github.com/SAP/spartacus/pull/15658#discussion_r870762291 since we're saying to put | undefined instead

projects/storefrontlib/layout/a11y/keyboard-focus/autofocus/auto-focus.service.ts
bgambocjaviniar: makes sense! it ties to this here https://github.com/SAP/spartacus/pull/15658#discussion_r870762291 since we're saying to put | undefined instead

projects/storefrontlib/layout/a11y/keyboard-focus/autofocus/auto-focus.service.ts
bgambocjaviniar: yup actually, this case here is a good one to make it optional instead of adding undefined type to the host param. Referencing cases like the one we spoke about for `set(key: string | undefined, group?: string | null) {`

projects/storefrontlib/layout/a11y/keyboard-focus/persist/persist-focus.service.ts
bgambocjaviniar: makes sense! it ties to this here https://github.com/SAP/spartacus/pull/15658#discussion_r870762291 since we're saying to put | undefined instead

projects/storefrontlib/layout/a11y/keyboard-focus/services/select-focus.util.ts
bgambocjaviniar: makes sense! it ties to this here https://github.com/SAP/spartacus/pull/15658#discussion_r870762291 since we're saying to put | undefined instead

projects/storefrontlib/layout/a11y/keyboard-focus/services/select-focus.util.ts
bgambocjaviniar: makes sense! it ties to this here https://github.com/SAP/spartacus/pull/15658#discussion_r870762291 since we're saying to put | undefined instead

projects/storefrontlib/layout/a11y/keyboard-focus/services/select-focus.util.ts
bgambocjaviniar: makes sense! it ties to this here https://github.com/SAP/spartacus/pull/15658#discussion_r870762291 since we're saying to put | undefined instead

bgambocjaviniar (approved):

PR #15671, opened 2022-04-27
feat: Allow CartSnapshotEvents to be sent on cart change

### LINKED ISSUE #15670: Add support to send cart-snapshot events to CDS whenever there is a change in the cart
**Is your feature request related to a problem? Please describe.**
Team Penguins of ISS is currently working on a new type of product recommendation: complementary products. This should display products, which ML considers good additions to the selection of products that the customer currently has in their cart. For this to work, ISS must get a reliable snapshot of how the customer's cart looks like.

**Describe the solution you'd like**
To achieve the above requirement, cds-integration library is adapted to send a new `CartSnapshotEvent` to event-layer containing the whole cart whenever there is an addition/modification/removal of items in the cart. This will then be picked up by profile-tag which converts this to an array of productSkus and sends that as a slimEvent to the edge service of ISS for further processing thus enabling the ML systems to have a good overview of the cart contents. The whole cart is sent to avoid further changes in spa/cds-integration-lib and enable PT to be the curator of the content needed even if the format changes later.

**Describe alternatives you've considered**
An alternative was to do this in profile-tag directly by talking to the occ-endpoints that are responsible for getting the anonymous and current user's carts. However, this was devoted as its just re-inventing the active-cart-facade thats readily available in spartacus and exposing the occ base-url to profile-tag thus introducing unwanted coupling.

**Additional context**
https://jira.tools.sap/browse/CXCDS-3688 describes the feature request in detail.

ShankariGiri-I:
Closes https://github.com/SAP/spartacus/issues/15670


nnamlaht (dismissed):

bgambocjaviniar (approved):

integration-libs/cds/src/profiletag/services/profile-tag-push-events.service.ts
```diff
@@ -350,6 +355,25 @@ export class ProfileTagPushEventsService {
     );
   }
 
+  protected cartChangedEvent(): Observable<ProfileTagPushEvent> {
```
bgambocjaviniar, line 368: After you message again in the thread again, I took a peek at the code instead of approving an already approved PR, but wouldn't the snippet achieve the same result? 

Instead of treating the pipe like a single function, let's treat it as a 'chain of async tasks'. Basically, streams in a declarative manner. Let me know if it works for you~ (also if you try to find in the codebase, we don't usually go for pipe(() => ...)
```
  protected cartChangedEvent(): Observable<ProfileTagPushEvent> {
    return merge(
      this.eventService.get(CartAddEntrySuccessEvent),
      this.eventService.get(CartUpdateEntrySuccessEvent),
      this.eventService.get(CartRemoveEntrySuccessEvent)
    ).pipe(
      withLatestFrom(
        this.activeCartFacade.getActive().pipe(
          distinctUntilChanged(),
          filter((cart) => cart.code !== undefined)
        )
      ),
      map(([_, cart]) => new CartSnapshotPushEvent({ cart }))
    );
  }
```

P.S 
1. you can always filter for cart.code to be undefined, but wouldn't we say that if cart returns an empty object, then it means we currently have no active cart, which in turns mean, there is no code as having a cart would mean that there is an attached code to it always I believe

2. don't think you need distinctUntilChanged when the service already does it 

screenshot from `feature-libs/cart/base/core/facade/active-cart.service.ts`

Screen Shot 2022-04-29 at 11 24 03 AM

bgambocjaviniar (changes requested):
Took a look instead of approving from someone already approving due to code owners. Please see comment

integration-libs/cds/src/profiletag/services/profile-tag-push-events.service.ts
ShankariGiri-I: Hi Brian! Thanks for having a look :) 

However, I am afraid that the solution that you are proposing won't work because of the following reasons:

1. When the user adds an item to the cart for the first time, the current code sends 2 CartSnapshot events one with productSkus: [] and the other with the new item in the cart. The filtering for the cart.code was done to avoid sending events when the cart is not initialized yet and that was the case. However, with the change u proposed, only one event is sent with empty skus thus not capturing the cart state on AddToCart event.
2. When the user removes the last item from the cart, as of now a CartSnapshot event is sent with empty skus. However, with the change that u proposed, its sending it with the sku id 🤔 

I do agree with point 2 :) Let me know what u think of this :)

bgambocjaviniar: double checked and worked on it~

integration-libs/cds/src/profiletag/services/profile-tag-push-events.service.ts
```diff
@@ -350,6 +355,22 @@ export class ProfileTagPushEventsService {
...
+    return merge(
+      this.eventService.get(CartAddEntrySuccessEvent),
+      this.eventService.get(CartUpdateEntrySuccessEvent),
+      this.eventService.get(CartRemoveEntrySuccessEvent)
+    ).pipe(
+      switchMapTo(this.activeCartFacade.takeActive()),
```
bgambocjaviniar, line 374: Let me know if this is working fine for you now~ 

If so, we'll wait for https://github.com/SAP/spartacus/pull/15716 to be merged

Finally, once this is merged, make sure to backport it to `release/4.3.x` to be released as a patch. However, do not forget to handle the constructor overload. An example of handling constructor deprecation can be found [here](https://github.com/SAP/spartacus/blob/release/4.3.x/projects/storefrontlib/cms-components/cart/add-to-cart/add-to-cart.component.ts#L58-L84)

bgambocjaviniar (changes requested):
To be reviewed https://github.com/SAP/spartacus/pull/15671/files#r868235621

Moreover, you can review among yourselves the backport (review/approve)

integration-libs/cds/src/profiletag/services/profile-tag-push-events.service.ts
ShankariGiri-I: Yes, it works for us & thanks a ton for adapting cart service! I fixed the unit and e2e tests because only a single event is sent on the first AddToCart event unlike earlier when 2 were sent. This is also fine by us 👍

bgambocjaviniar (dismissed):

integration-libs/cds/src/profiletag/services/profile-tag-push-events.service.ts
bgambocjaviniar: don't forget to run yarn prettier:fix in root to prettify your files if you don't have prettier enabled in your IDE

bgambocjaviniar (approved):

PR #15698, opened 2022-05-02
Feature/cxspa 332

FollowTheFlo:
PDF Component implementation, this PR includes:
Business logic and styling
Not to take in account in this PR: logo and Accessibility as still need to be defined.

### DIFF
--- a/.env-cmdrc
+++ b/.env-cmdrc
@@ -1,6 +1,6 @@
 {
   "dev": {
-    "CX_BASE_URL": "https://40.76.109.9:9002"
+    "CX_BASE_URL": "https://52.188.69.146:59002"
   },
   "local": {
     "CX_BASE_URL": "https://localhost:9002"
--- a/.env-cmdrc
+++ b/.env-cmdrc
@@ -1,6 +1,6 @@
 {
   "dev": {
-    "CX_BASE_URL": "https://52.188.69.146:59002"
+    "CX_BASE_URL": "https://40.76.109.9:9002"
   },
   "local": {
     "CX_BASE_URL": "https://localhost:9002"
--- a/projects/assets/src/translations/en/pdf.ts
+++ b/projects/assets/src/translations/en/pdf.ts
@@ -1,5 +1,5 @@
 export const pdf = {
   pdf: {
-    document: 'Document',
+    defaultName: 'Document',
   },
 };
--- a/projects/assets/src/translations/en/pdf.ts
+++ b/projects/assets/src/translations/en/pdf.ts
@@ -1,0 +1,5 @@
+export const pdf = {
+  pdf: {
+    document: 'Document',
+  },
+};
--- a/projects/assets/src/translations/en/pdf.ts
+++ b/projects/assets/src/translations/en/pdf.ts
@@ -1,5 +1,5 @@
 export const pdf = {
   pdf: {
-    defaultName: 'Document',
+    defaultTitle: 'Document',
   },
 };
--- a/projects/core/src/model/cms.model.ts
+++ b/projects/core/src/model/cms.model.ts
@@ -189,6 +189,12 @@ export interface CmsAddToCartComponent extends CmsComponent {
   inventoryDisplay?: boolean;
 }
 
+export interface CmsPDFDocumentComponent extends CmsComponent {
+  pdfFile?: CmsBannerComponentMedia;
+  name?: string;
+  height: number;
+}
+
 export interface CmsVideoComponent extends CmsComponent {
   overlayTitle?: string;
   autoPlay?: string;
--- a/projects/core/src/model/cms.model.ts
+++ b/projects/core/src/model/cms.model.ts
@@ -192,7 +192,7 @@ export interface CmsAddToCartComponent extends CmsComponent {
 export interface CmsPDFDocumentComponent extends CmsComponent {
   pdfFile?: CmsBannerComponentMedia;
   name?: string;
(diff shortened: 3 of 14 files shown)


.env-cmdrc
```diff
@@ -1,6 +1,6 @@
 {
   "dev": {
-    "CX_BASE_URL": "https://40.76.109.9:9002"
+    "CX_BASE_URL": "https://52.188.69.146:59002"
```
RadhepS: should be changed back when merging

projects/core/src/model/cms.model.ts
```diff
@@ -189,6 +189,12 @@ export interface CmsAddToCartComponent extends CmsComponent {
   inventoryDisplay?: boolean;
 }
 
+export interface CmsPDFDocumentComponent extends CmsComponent {
+  pdfFile?: CmsBannerComponentMedia;
+  name?: string;
+  height: number;
```
RadhepS: height should be optional since it can be undefined from smartedit

projects/storefrontlib/cms-components/content/pdf/pdf.component.html
```diff
@@ -0,0 +1,14 @@
+<ng-container *ngIf="data$ | async as data">
+  <div class="pdf-container">
+    <span
+      tabindex="0"
+      aria-label="Download PDF file"
+      (click)="!data?.pdfFile?.url || download(data.pdfFile)"
```
RadhepS: not able to start your branch but,

do we need `!data`? Also, do we need the download function or can we always just use the URL?

projects/storefrontlib/cms-components/content/pdf/pdf.component.html
```diff
@@ -0,0 +1,14 @@
+<ng-container *ngIf="data$ | async as data">
+  <div class="pdf-container">
+    <span
+      tabindex="0"
+      aria-label="Download PDF file"
```
RadhepS: aria label should be translated unless this will be done in a difference PR

RadhepS (changes requested):

.env-cmdrc
FollowTheFlo: yes , i did it on purpose, the plan is to set it on epic/pdf, will be change back before merge to develop.

projects/storefrontlib/cms-components/content/pdf/pdf.component.html
```diff
@@ -0,0 +1,10 @@
+<ng-container *ngIf="data$ | async as data">
+  <div class="pdf-container">
+    <a [ngClass]="url ? 'active' : 'disabled'" [href]="url" target="_blank">
+      {{ data?.name || data?.pdfFile?.altText || 'document.pdf' }}
```
RadhepS: We can add translation to the fallback document.pdf text

RadhepS (changes requested):

projects/storefrontstyles/scss/components/content/pdf/_pdf.scss
```diff
@@ -0,0 +1,24 @@
+%cx-pdf {
```
RadhepS, line 1: Can you have Mikel review the CSS please?

projects/storefrontlib/cms-components/content/pdf/pdf.component.html
RadhepS: Can you verify with UX/UI whether we should append .pdf to the name of the file? Easier for people to know its a pdf file but, I noticed other websites don't include the .pdf extension in the name

For example:
image
vs
image

projects/storefrontstyles/scss/components/content/pdf/_pdf.scss
```diff
@@ -0,0 +1,24 @@
...
+  .pdf-container {
+    padding: 15px;
+    font-size: 0.875 rem;
+    a {
+      color: var(--cx-color-text);
+    }
+    cx-icon {
+      margin-left: 10px;
```
morganm58: This should be `margin-inline-start: 0.625rem;` This will make it work with directionality. And we should be using rem instead of px for padding, margin and font-sizes

projects/storefrontstyles/scss/components/content/pdf/_pdf.scss
```diff
@@ -0,0 +1,24 @@
+%cx-pdf {
+  .pdf-container {
+    padding: 15px;
+    font-size: 0.875 rem;
```
morganm58: Font-sizes should be variables. `font-size: var(--cx-font-size, 0.875rem);`

projects/storefrontstyles/scss/components/content/pdf/_pdf.scss
```diff
@@ -0,0 +1,24 @@
+%cx-pdf {
+  .pdf-container {
+    padding: 15px;
```
morganm58: Can you change the px to rem. this would be 0.938rem

projects/assets/src/translations/en/pdf.ts
```diff
@@ -0,0 +1,5 @@
+export const pdf = {
+  pdf: {
+    document: 'Document',
```
RadhepS: i think this can be `defaultName: 'Document'`

RadhepS (changes requested):

morganm58 (approved):

RadhepS (approved):

PR #15701, opened 2022-05-02
Feature/cxspa 267: enable strict mode for storefrontlib/shared

WeizhengSap:
enable strict mode for `storefrontlib/shared`


projects/storefrontlib/shared/utils/validators/custom-form-validators.ts
```diff
@@ -189,7 +189,7 @@ export class CustomFormValidators {
     startDateKey: string,
     endDateKey: string,
     getDate: (value: string) => Date
-  ): (FormGroup) => any {
+  ): (_: FormGroup) => any {
```
bgambocjaviniar: this file is still complaining about 'strictness' for

```
  projects/storefrontlib/shared/utils/validators/custom-form-validators.ts:191:14
    191     getDate: (value: string) => Date
                     ~~~~~~~~~~~~~~~~~~~~~~~
    The expected type comes from the return type of this signature.
```

projects/storefrontlib/shared/components/list-navigation/sorting/sorting.component.html
```diff
@@ -10,6 +10,6 @@
   }"
 >
   <ng-option *ngFor="let sort of sortOptions" [value]="sort.code">{{
```
bgambocjaviniar, line 12: this file is still complaining about 'strictness' for
line 8 fails in strict mode as well due to the `nullish coalescing` requiring the LHS to be null or undefined. 

We can do the following:

1 - ariaLabel: ariaLabel || ('productList.sortResults' | cxTranslate)
OR
2- in the component, we make the input string | undefined, since the LHS expects it to have a null / undefined

projects/storefrontlib/cms-components/myaccount/payment-methods/payment-methods.component.ts
```diff
@@ -78,7 +78,7 @@ export class PaymentMethodsComponent implements OnInit {
           actions.push({ name: textDelete, event: 'edit' });
           const card: Card = {
             role: 'region',
-            header: defaultPayment ? textDefaultPaymentMethod : null,
+            header: defaultPayment ? textDefaultPaymentMethod : undefined,
```
bgambocjaviniar, line 81: 👍  I know this was to fix the header issue from card.component, but just mentioning if it was not done in /storefrontlib/cms-component that the template is complaining and other instances in this file (did not check the other PR)

projects/storefrontlib/shared/components/anonymous-consents-dialog/anonymous-consent-dialog.component.ts
```diff
@@ -110,14 +115,17 @@ export class AnonymousConsentDialogComponent implements OnInit, OnDestroy {
             templates.forEach((template) => {
               const consent = this.getCorrespondingConsent(template, consents);
               if (
-                (consent && consent.consentState == null) ||
-                this.anonymousConsentsService.isConsentWithdrawn(consent)
+                consent &&
+                (consent.consentState == null ||
```
bgambocjaviniar, line 119: 👍 

makes sense

projects/storefrontlib/shared/components/anonymous-consents-dialog/anonymous-consent-dialog.component.ts
```diff
@@ -128,10 +136,11 @@ export class AnonymousConsentDialogComponent implements OnInit, OnDestroy {
   }
 
   private isRequiredConsent(template: ConsentTemplate): boolean {
-    return (
-      Boolean(this.config.anonymousConsents) &&
-      Boolean(this.config.anonymousConsents.requiredConsents) &&
-      this.config.anonymousConsents.requiredConsents.includes(template.id)
+    return Boolean(
```
bgambocjaviniar, line 139:
```suggestion
    return Boolean(
      template.id &&
      this.config.anonymousConsents?.requiredConsents &&
      this.config.anonymousConsents.requiredConsents.includes(template.id)
    );
```

projects/storefrontlib/shared/components/card/card.component.html
```diff
@@ -80,7 +80,7 @@
     <!-- Actions -->
     <div *ngIf="content.actions && !editMode" class="cx-card-actions">
       <div *ngFor="let action of content.actions">
-        <div [ngSwitch]="action.event">
+        <div [ngSwitch]="$any(action).event">
```
bgambocjaviniar: I don't know what to say about this since we're avoiding the problem altogether from the strict template checking

I was trying some things out, but maybe we can create some user defined guards to be used in the template https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards

same for the changes below in this file

projects/storefrontlib/shared/components/form/file-upload/file-upload.component.ts
```diff
@@ -52,7 +52,9 @@ export class FileUploadComponent implements ControlValueAccessor {
   }
 
   get selectedFiles(): File[] | undefined {
-    return Array.from(this.fileInput.nativeElement.files);
+    if (this.fileInput.nativeElement.files) {
+      return Array.from(this.fileInput.nativeElement.files);
+    }
```
bgambocjaviniar, line 57: Let's return undefined after the `if` in a case `this.fileInput.nativeElement.files` does not exist

projects/storefrontlib/shared/components/generic-link/generic-link.component.ts
```diff
@@ -65,22 +65,22 @@ export class GenericLinkComponent implements OnChanges {
   /**
    * The part with the path of the local url.
    */
-  get routerUrl(): any[] {
+  get routerUrl(): any[] | undefined {
```
bgambocjaviniar:
```suggestion
  get routerUrl(): string[] | undefined {
```

projects/storefrontlib/shared/components/generic-link/generic-link.component.ts
```diff
@@ -103,7 +103,7 @@ export class GenericLinkComponent implements OnChanges {
    */
   protected splitUrl(url: string = ''): RouteParts {
     const { queryParams, fragment } = this.router.parseUrl(url);
-    const [, path] = url.match(this.URL_SPLIT);
+    const [, path] = url.match(this.URL_SPLIT) ?? [];
```
bgambocjaviniar: I wonder if this is the correct approach

Due to the following reasons:
1 - it won't ever be undefined or null due to the fact we have a default value of ''. In any case we split, it will return something.
2 - if for 'x' reason, it is undefined or null, so it will get to the [], then const [, path] = [] does not make sense.

Maybe the fix would be ?? [, url], but would need to handle the undefined 'state'

projects/storefrontlib/shared/components/list-navigation/pagination/pagination.builder.ts
```diff
@@ -135,9 +137,10 @@ export class PaginationBuilder {
     };
 
     const addLastGap = () => {
-      const nextPageNumber = pages[pages.length - 1].number + 1;
+      const number = pages[pages.length - 1].number;
```
bgambocjaviniar: let's call it pageNumber

projects/storefrontlib/shared/components/list-navigation/pagination/pagination.builder.ts
```diff
@@ -310,17 +313,20 @@ export class PaginationBuilder {
...
+      const minStart = Math.max(0, current - delta);
+      // ensures that we start with at least 1 and do not pass the last range
+      const maxStart = Math.max(0, pageCount - count - 1);
 
-    // ensure that we get at least a full range at the end
-    return Math.min(maxStart, minStart);
+      // ensure that we get at least a full range at the end
+      return Math.min(maxStart, minStart);
+    }
+    return 0;
```
bgambocjaviniar: to verify (smoke test) 

should it be 0 or should we return null and handle it everywhere

projects/storefrontlib/shared/components/list-navigation/pagination/pagination.component.ts
```diff
@@ -68,7 +68,7 @@ export class PaginationComponent {
    * @param type PaginationItemType
    * @returns string
    */
-  getAriaLabel(label: string, type: PaginationItemType): string {
+  getAriaLabel(label?: string, type?: PaginationItemType): string {
```
bgambocjaviniar, line 71: I think this is an okay case to use optional instead of undefined (from the discussion we had before in the other PR)

projects/storefrontlib/shared/components/media/media.service.ts
```diff
@@ -154,11 +159,12 @@ export class MediaService {
     }
 
     const srcset = formats.reduce((set, format) => {
-      if (!!media[format.code]) {
+      const image = (media as MediaContainer)[format.code];
+      if (image) {
```
bgambocjaviniar: nitpicking, but let's keep it as !!image

projects/storefrontlib/shared/components/modal/modal.service.ts
```diff
@@ -39,21 +39,21 @@ export class ModalService {
...
 
-  getActiveModal(): ModalRef {
+  getActiveModal(): ModalRef | null {
     const modal = this.modals[this.modals.length - 1];
     return modal ? modal : null;
   }
 
   dismissActiveModal(reason?: any): void {
-    const modal: ModalRef = this.getActiveModal();
+    const modal: ModalRef | null = this.getActiveModal();
```
bgambocjaviniar: I think we can just remove : ModalRef | null since it's implicit from the method

bgambocjaviniar: same for the bottom one

projects/storefrontlib/shared/components/popover/popover.component.html
```diff
@@ -13,7 +13,7 @@
     </button>
   </div>
   <ng-container *ngIf="isTemplate">
-    <ng-container *ngTemplateOutlet="content"></ng-container>
+    <ng-container *ngTemplateOutlet="$any(content)"></ng-container>
```
bgambocjaviniar: might need to use the user defined guard as mentioned earlier

projects/storefrontlib/shared/components/popover/popover.component.ts
```diff
@@ -160,18 +160,18 @@ export class PopoverComponent implements OnInit, OnDestroy, AfterViewChecked {
     this.eventSubject.next(PopoverEvent.ESCAPE_KEYDOWN);
   }
 
-  protected isClickedOnPopover(event) {
+  protected isClickedOnPopover(event: any) {
```
bgambocjaviniar: instead of any, maybe put it as `MouseEvent` since it comes from line 149 on outsideClick which shows that it's a MouseEvent

projects/storefrontlib/shared/components/table/table-renderer.service.ts
```diff
@@ -128,7 +128,7 @@ export class TableRendererService {
    */
   getDataOutletContext(
     type: string,
-    options: TableOptions,
+    options: TableOptions | undefined,
```
bgambocjaviniar, line 131: 👍 

since we're accepting undefined, let's type these from table.component.ts

```
  private get layout(): TableLayout | undefined {
    return this.structure?.options?.layout;
  }

  /**
   * Helper method to return the deeply nested type.
   */
  private get type(): string {
    return this.structure?.type;
  }

  private get options(): TableOptions | undefined {
    return this.structure?.options;
  }
```

projects/storefrontlib/shared/components/table/table.component.html
```diff
@@ -19,7 +19,7 @@
             [cxOutlet]="getDataOutletRef(cell)"
             [cxOutletContext]="getDataOutletContext(cell, item)"
           >
-            {{ item[cell] }}
+            {{ $any(item)[cell] }}
```
bgambocjaviniar, line 22: in this case, I believe we don't need a user defined guard since the data is generic

bgambocjaviniar (changes requested):

projects/storefrontlib/shared/utils/forms/title-utils.ts
```diff
@@ -10,9 +10,12 @@ export const titleScores = {
 };
 
 export function sortTitles(title1: Title, title2: Title) {
```
bgambocjaviniar: : number

projects/storefrontlib/shared/components/popover/popover.component.ts
bgambocjaviniar: same for the bottom one

projects/storefrontlib/shared/components/split-view/split-view.service.ts
```diff
@@ -17,7 +17,7 @@ export class SplitViewService {
 
   protected splitViewCount = 1;
 
-  protected _views$: BehaviorSubject<any[]> = new BehaviorSubject([]);
+  protected _views$: BehaviorSubject<any[]> = new BehaviorSubject<any[]>([]);
```
bgambocjaviniar:
```suggestion
    protected _views$: BehaviorSubject<SplitViewState[]> = new BehaviorSubject<
    SplitViewState[]
  >([]);
```

projects/storefrontlib/shared/components/table/table.service.ts
```diff
@@ -79,18 +79,19 @@ export class TableService {
     type: string,
     breakpoint: BREAKPOINT,
     defaultStructure?: ResponsiveTableConfiguration
-  ): TableStructureConfiguration {
+  ): TableStructureConfiguration | undefined {
     if (!this.config.table?.[type]) {
-      return null;
+      return undefined;
```
bgambocjaviniar: Maybe null plays a big role for the table service

I believe we can keep it to `return null` and typed `as | null` assuming that it explicitly returned as null and not `return;`

```
    if (!this.config.table?.[type]) {
      return;
```

projects/storefrontlib/shared/utils/validators/custom-form-validators.ts
bgambocjaviniar: also instead of => any

we can do => ValidationErrors | null

projects/storefrontlib/cms-components/myaccount/payment-methods/payment-methods.component.ts
WeizhengSap: it is done in other PR.

projects/storefrontlib/shared/components/list-navigation/pagination/pagination.builder.ts
WeizhengSap: I will make it return null.

projects/storefrontlib/shared/components/generic-link/generic-link.component.ts
WeizhengSap: I do this:
```
     const [, path] = url.match(this.URL_SPLIT) ?? [, ''];

    // wrap path in an array, to have the Angular-like path format
    return { path: [path ?? ''], queryParams, fragment };
```

projects/storefrontlib/shared/components/list-navigation/sorting/sorting.component.html
WeizhengSap: I didn't see the complaining.

projects/storefrontlib/shared/utils/validators/custom-form-validators.ts
WeizhengSap: I didn't see the complaining.

bgambocjaviniar: as show in DM
Screen Shot 2022-05-17 at 11 17 58 AM

projects/storefrontlib/shared/components/list-navigation/sorting/sorting.component.html
bgambocjaviniar: as shown in DM
Screen Shot 2022-05-17 at 11 17 18 AM

projects/storefrontlib/shared/components/card/card.component.html
```diff
@@ -120,7 +120,7 @@
           </button>
           <a
             *ngSwitchDefault
-            href="{{ action.link }}"
+            href="{{ $any(action).link }}"
```
bgambocjaviniar: don't forget this one as well 

is it possible to make isCardAction working for both? or we might have to define another function?

bgambocjaviniar (changes requested):
should be good after fixing the comment above

But also, you have a unit test that is failing

projects/storefrontlib/shared/components/card/card.component.html
WeizhengSap: I will remove `isCardAction(action)`, because it narrow `action` to `CardAction`. the `*ngSwitchDefault` is used for `CardLinkAction`. So, I need think about another solution.

bgambocjaviniar (approved):

PR #15702, opened 2022-05-04
Feature/cxspa 206

kimhw0630:


feature-libs/asm/styles/components/_asm-main-ui.component.scss
```diff
@@ -103,6 +103,20 @@
       display: flex;
       justify-content: flex-end;
       align-items: center;
+      .cx-asm-customer-list {
+        cursor: pointer;
+        cx-icon {
+          padding: 0 7px 0 0;
```
morganm58: for directionality this one should be
```
padding-top: 0;
padding-inline-end: 7px;
padding-bottom: 0;
padding-inline-start: 0;
```

feature-libs/asm/styles/components/_customer-list.component.scss
```diff
@@ -0,0 +1,185 @@
...
+        margin: 0;
+      }
+      .cx-pagination-buttons {
+        display: flex;
+        align-items: center;
+        cx-icon {
+          color: $sapFioriLinkColor;
+        }
+        cx-icon.previous {
+          padding: 0 10px 0 0;
```
morganm58: This one should be changed for directionality as well.
```
padding-top: 0;
padding-inline-end: 10px;
padding-bottom: 0;
padding-inline-start: 0;
```

feature-libs/asm/styles/components/_customer-list.component.scss
```diff
@@ -0,0 +1,185 @@
...
+        }
+        cx-icon.next {
+          padding: 0 0 0 10px;
```
morganm58: This one should be changed for directionality as well.
```
padding-top: 0;
padding-inline-end: 0;
padding-bottom: 0;
padding-inline-start: 10px;
```

feature-libs/asm/styles/components/_customer-list.component.scss
```diff
@@ -0,0 +1,185 @@
...
+        }
+        button {
+          text-decoration: none;
+          color: $sapFioriLinkColor;
+        }
+        .cx-btn-previous {
+          span {
+            padding: 0 0 0 5px;
```
morganm58: This one should be changed for directionality as well.
```
padding-top: 0;
padding-inline-end: 0;
padding-bottom: 0;
padding-inline-start: 5px;
```

feature-libs/asm/styles/components/_customer-list.component.scss
```diff
@@ -0,0 +1,185 @@
...
+          }
+        }
+        .cx-btn-next {
+          span {
+            padding: 0 5px 0 0;
```
morganm58: This one should be changed for directionality as well.
```
padding-top: 0;
padding-inline-end: 5px;
padding-bottom: 0;
padding-inline-start: 0;
```

feature-libs/asm/styles/components/_customer-list.component.scss
```diff
@@ -0,0 +1,185 @@
...
+          }
+        }
+        div:first-child {
+          font-size: 0.875rem;
+          color: $sapFioriTextColor;
+          padding: 0 10px 0 10px;
```
morganm58: this one can be changed to 
padding: 0 10px;

feature-libs/asm/styles/components/_customer-list.component.scss
```diff
@@ -0,0 +1,185 @@
...
+        }
+      }
+      .table-header-cell {
+        line-height: 40px;
+        margin: 0 15px 0 0;
```
morganm58: This one should be changed for directionality as well.
```
margin-top: 0;
margin-inline-end: 15px;
margin-bottom: 0;
margin-inline-start: 0;
```

feature-libs/asm/styles/components/_customer-list.component.scss
```diff
@@ -0,0 +1,185 @@
...
+    .cx-avatar {
+      width: 44px;
+      height: 44px;
+      line-height: 44px;
+      border-radius: 50%;
+      color: #fff;
+      text-align: center;
+      background: #9259cd;
+      text-transform: uppercase;
+      margin: 2px 10px 2px 0;
```
morganm58: This one should be changed for directionality as well.
```
margin-top: 2px;
margin-inline-end: 10px;
margin-bottom: 2px;
margin-inline-start: 0;
```

feature-libs/asm/styles/components/_customer-list.component.scss
```diff
@@ -0,0 +1,185 @@
...
+  }
+
+  .cx-empty-state {
+    display: flex;
+    justify-content: center;
+    margin: 50px 0 0 0;
+  }
+  ng-select {
+    .ng-select-container {
+      border: 0 !important;
```
morganm58: We try to stay away from using !important in our scss. can this be made more specific to avoid the use of the !important?

feature-libs/asm/styles/components/_customer-list.component.scss
```diff
@@ -0,0 +1,185 @@
...
+      .ng-arrow-wrapper {
+        .ng-arrow {
+          border-color: $sapFioriLinkColor transparent transparent !important;
```
morganm58: Same comment as above. Can this be made more specific to avoid the use of the !important?

feature-libs/asm/styles/components/_customer-list.component.scss
```diff
@@ -0,0 +1,185 @@
...
+        }
+      }
+    }
+  }
+  .ng-select-focused {
+    outline: none !important;
```
morganm58: Same comment as above. Can this be made more specific to avoid the use of the !important.

morganm58:
The pagination doesn't match what is in the visual. The dropdown styling is also different
Screen Shot 2022-07-01 at 11 05 16 AM
Screen Shot 2022-07-01 at 11 05 23 AM
Screen Shot 2022-07-01 at 11 05 33 AM
Screen Shot 2022-07-01 at 11 12 10 AM

kimhw0630:
> The pagination doesn't match what is in the visual. The dropdown styling is also different Screen Shot 2022-07-01 at 11 05 16 AM Screen Shot 2022-07-01 at 11 05 23 AM Screen Shot 2022-07-01 at 11 05 33 AM Screen Shot 2022-07-01 at 11 12 10 AM

@morganm58 
For the drop-down, we are using common controls (ng-select and cx-sort).
For the pagination currently we don't get all required fields from the backend so we can't match the design right now.  We have already informed Veronica.

morganm58 (dismissed):

feature-libs/asm/root/facade/asm-facade.service.ts
```diff
@@ -0,0 +1,24 @@
+import { Injectable } from '@angular/core';
+import { facadeFactory, QueryState } from '@spartacus/core';
```
WeizhengSap: can we rename it to `customer-list.facade.ts`? Or merge these 2 functions into `asm.facade.ts` which is created in https://github.com/SAP/spartacus/pull/15776/files.

feature-libs/asm/root/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from './facade/index';
+export * from './model/index';
```
WeizhengSap, line 2: We can put all exported stuff in index.ts
and `export * from './index';` in public_api.ts.

feature-libs/asm/root/model/customer-lists-page.model.ts
```diff
@@ -0,0 +1,14 @@
+import { UserGroup } from './user-group.model';
```
WeizhengSap: can we put these model interfaces into one file something like this: `customer-list.model.ts`

feature-libs/asm/occ/adapters/occ-asm.adapter.ts
```diff
@@ -40,9 +69,16 @@ export class OccAsmAdapter implements AsmAdapter {
...
+    let params: HttpParams = new HttpParams().set(
+      'baseSite',
+      this.activeBaseSite
+    );
+
+    if (typeof options['sort'] !== 'undefined') {
+      params = params.set('sort', '' + options.sort);
+    } else {
+      params = params.set('sort', 'byNameAsc');
+    }
```
WeizhengSap, line 90: can be this `params = params.set('sort', options.sort ?? 'byNameAsc');`.

feature-libs/asm/occ/adapters/occ-asm.adapter.ts
```diff
@@ -52,6 +88,14 @@ export class OccAsmAdapter implements AsmAdapter {
       params = params.set('pageSize', '' + options.pageSize);
     }
 
+    if (typeof options['currentPage'] !== 'undefined') {
+      params = params.set('currentPage', '' + options.currentPage);
```
WeizhengSap: why `'' + options.currentPage`? convert number to string? can it be this?: 
```
if (options.pageSize !== undefined) {
      params = params.set('pageSize', options.pageSize);
    }
```

feature-libs/asm/occ/adapters/occ-asm.adapter.ts
```diff
@@ -29,7 +33,32 @@ export class OccAsmAdapter implements AsmAdapter {
   ) {
     this.baseSiteService
       .getActive()
-      .subscribe((value) => (this.activeBaseSite = value));
+      .subscribe((value) => (this.activeBaseSite = value)); // TODO: Concurrency issue?
+  }
+
+  customerLists(): Observable<CustomerListsPage> {
+    const headers = InterceptorUtil.createHeader(
```
WeizhengSap: this header seems used in all functions in this adapter. Maybe make it as `protected` class variable.

feature-libs/asm/core/facade/asm.service.ts
```diff
@@ -14,7 +17,20 @@ import { AsmSelectors } from '../store/index';
   providedIn: 'root',
 })
 export class AsmService {
-  constructor(protected store: Store<StateWithAsm>) {}
+  constructor(
```
WeizhengSap: facade codes need to be organized with the code in this PR https://github.com/SAP/spartacus/pull/15776/files. Either create 2 facades `asm facade` and `customers facade`, or just one facade`asm facade`.

feature-libs/asm/core/facade/asm-query.service.ts
```diff
@@ -0,0 +1,53 @@
...
+  CommandStrategy,
+  Query,
+  QueryService,
+  QueryState,
+} from '@spartacus/core';
+import { Observable } from 'rxjs';
+import { AsmConnector } from '../connectors/asm.connector';
+
+@Injectable()
+export class AsmQueryService implements AsmFacadeService {
```
WeizhengSap: `AsmQueryService` is not a good name. We can just create queries in facade service.

feature-libs/asm/core/facade/asm-query.service.ts
```diff
@@ -0,0 +1,53 @@
...
+  ) {}
+
+  getCustomerLists(): Observable<QueryState<CustomerListsPage>> {
+    return this.customerListQuery$.getState();
+  }
+
+  getCustomers(
+    options: CustomerSearchOptions = {}
+  ): Observable<CustomerSearchPage> {
+    return this.customerSearchCommand$.execute(options);
```
WeizhengSap: I understand we use `command` because we cannot pass parameter to `query`. Can you take a look of this code, and consider whether we can use a similar way here.
```
protected searchInput$ = new Subject<{ term: string; pagination: number, ... }>();
performSearch(term: string, pagination: number, ...): void {
  this.searchInput$.next({ term, pagination, ... });
}
protected searchQuery: Query<unknown> = this.query.create(
  () =>
    this.searchInput$.pipe(
      switchMap(({ term, pagination, ... }) =>
        this.connector.search(term, pagination, ...)
      )
    )
);
```

feature-libs/asm/components/customer-list/customer-list.model.ts
```diff
@@ -0,0 +1,7 @@
+import { CustomerListColumnActionType } from '@spartacus/asm/root';
+import { User } from '@spartacus/core';
+
+export interface CustomerListActionEvent {
```
WeizhengSap: Can we rename to it `CustomerListAction`? We have `CxEvent`, this interface is not Spartacus' event.

feature-libs/asm/components/customer-list/customer-list.component.ts
```diff
@@ -0,0 +1,243 @@
...
+    protected translation: TranslationService
+  ) {
+    this.breakpoint$ = this.getBreakpoint();
+  }
+
+  ngOnInit(): void {
+    this.customerListsPage$ = this.asmService
+      .getCustomerLists()
+      .pipe(
+        filter((queryState) => queryState.loading === false),
```
WeizhengSap: can we move this `filter` and `map` into service? So, calling `this.asmService.getCustomerLists()` just get observable of customerListsPage.

feature-libs/asm/components/customer-list/customer-list.component.ts
```diff
@@ -0,0 +1,243 @@
...
+        map((queryState) => {
+          return queryState.data;
+        })
+      )
+      .pipe(
+        map((customerListsPage) => customerListsPage),
```
WeizhengSap: why we have this `map` and `filter`?

feature-libs/asm/components/customer-list/customer-list.component.ts
```diff
@@ -0,0 +1,243 @@
...
+    protected asmService: AsmService,
+    protected breakpointService: BreakpointService,
+    protected asmConfig: AsmConfig,
+    protected translation: TranslationService
+  ) {
+    this.breakpoint$ = this.getBreakpoint();
+  }
+
+  ngOnInit(): void {
+    this.customerListsPage$ = this.asmService
```
WeizhengSap: we can move this logic out of `ngOnInit`. So, just `customerListsPage$: Observable<CustomerListsPage | undefined> = this.asmService.getCustomerLists()....`

feature-libs/asm/components/customer-list/customer-list.component.ts
```diff
@@ -0,0 +1,243 @@
...
+  fetchCustomers(): void {
+    const pageSize =
+      this.asmConfig.asm?.customerList?.pageSize ?? this.PAGE_SIZE;
+    if (this.selectedUserGroupId) {
+      const options: CustomerSearchOptions = {
+        customerListId: this.selectedUserGroupId,
+        pageSize: pageSize,
+        currentPage: this.currentPage,
+      };
+      if (this.sortCode) {
```
WeizhengSap, line 139: if `this.sortCode` is empty string `''`, it is also false. So, each time, we select a customer group, the sort is always `byNameAsc`

feature-libs/asm/components/customer-list/customer-list.component.ts
```diff
@@ -0,0 +1,243 @@
...
+            }
+          })
+        );
+    }
+  }
+
+  onChangeCustomerGroup(): void {
+    this.currentPage = 0;
+    this.sorts = null;
+    this.sortCode = '';
```
WeizhengSap, line 131: we set `this.sortCode` to empty string, but in line 103, it will be false.

feature-libs/asm/components/customer-list/customer-list.component.ts
```diff
@@ -0,0 +1,243 @@
...
+    if (this.selectedUserGroupId) {
+      const options: CustomerSearchOptions = {
+        customerListId: this.selectedUserGroupId,
+        pageSize: pageSize,
+        currentPage: this.currentPage,
+      };
+      if (this.sortCode) {
+        options.sort = this.sortCode;
+      }
+      this.customerSearchPage$ = of(options)
```
WeizhengSap: why we use `of` and `switchMap` here? can it be just this?:
```
    const options: CustomerSearchOptions = {
        customerListId: this.selectedUserGroupId,
        pageSize: this.customerListConfig?.pageSize ?? this.PAGE_SIZE,
        currentPage: this.currentPage,
        sort: this.sortCode,
      };

      this.loaded = false;
      this.customerSearchPage$ = this.asmService.searchCustomers(options).pipe(
        tap((result) => {
          this.loaded = true;
          ....
        })
      );
```

feature-libs/asm/components/customer-list/customer-list.component.ts
```diff
@@ -0,0 +1,243 @@
...
+        )
+        .pipe(
+          tap((result) => {
+            this.loaded = true;
+            if (result.sorts) {
+              this.sorts = result.sorts;
+              this.sortCode = result.pagination?.sort;
+            }
+            if (result.entries.length < pageSize) {
+              this.maxPage = result.pagination?.currentPage || 0;
```
WeizhengSap, line 119: `maxPage` here is confusing. Maybe replaced by a boolean variable `isLastPage`. (`this.isLastPage = result.entries.length < pageSize`)

feature-libs/asm/components/customer-list/customer-list.component.ts
kimhw0630: The sorting depends on the customer group for example if the customer group id is 'A' then we only can sort by 'date' and if the customer group id is 'B' then we only can sort by 'name'.  If wrong sorting is passed then the backend ignores and send back us the default sorting.  so we removed the default sort (byNameAsc) if the customer group is passed in occ-asm.adapter.ts

feature-libs/asm/core/facade/asm-query.service.ts
ercultimate: Can you explain more why the suggestion is more desirable?  

As I understand, the suggestion is what the Command object does.  The critical behavioral difference is that, with the suggestion, all consumers of `getCustomers()` would receive the same value.  The connector result acts like a "singleton" provider, where there can only be 1 value that is the latest page request.  This would prevents multiple consumers from having separate page states.

feature-libs/asm/components/customer-list/customer-list.component.ts
ercultimate: I would prefer not to implement this change. This can make the component more involved to unit test.  

Assume the component initialization is done during class construction, per the suggestion.  In the unit test, the component class is typically already constructed by the time a specific unit test or describe block becomes active.  If you need to change the return values of spies for that test/describe, you need to re-construct the component so that the new values are used in the component initialization.

feature-libs/asm/occ/adapters/occ-asm.adapter.ts
ercultimate: This suggestion is not sufficient.  There are 3 cases here:
1. Sort is provided
   - use provided value 
1. Sort not provided, normal customer search
   - use default value
1. Sort not provided, customer list search
   - do not send default, let api select default value
  
The valid sort values for a customer list search are defined per the search strategy used for the list.  The sort value for a normal customer search is highly likely to not be valid.  Unless it is explicitly specified, the sort value is not sent so that the API can respond with its default sort method.  The chosen sort value is readable in the response should you need to make UI adjustments.

As a side note, I don't believe we should be providing a hard-coded default for sort for any case.

feature-libs/asm/occ/adapters/occ-asm.adapter.ts
```diff
@@ -35,27 +37,69 @@ export class OccAsmAdapter implements AsmAdapter {
...
+      {},
+      {
+        baseSite: false,
+        prefix: false,
+      }
+    );
+
+    return this.http
+      .get<CustomerListsPage>(url, { headers, params })
+      .pipe(this.converterService.pipeable(CUSTOMER_LISTS_NORMALIZER));
```
WeizhengSap: better handle error as well: `catchError((error) => throwError(normalizeHttpError(error))),`

feature-libs/asm/core/store/actions/customer.action.ts
```diff
@@ -44,9 +56,41 @@ export class CustomerSearchReset extends StateUtils.LoaderResetAction {
   }
 }
 
+export class CustomerListCustomersSearch extends StateUtils.LoaderLoadAction {
```
WeizhengSap, line 59: This action.ts looks good me, just one small thing: Why `CustomerListCustomersSearch`? (there are 2 customer in the name). how about `CustomerListSearch`?

feature-libs/asm/core/facade/facade-providers.ts
```diff
@@ -0,0 +1,17 @@
...
+
+import { Provider } from '@angular/core';
+import { AsmCustomerListFacade } from '@spartacus/asm/root';
+import { AsmCustomerListService } from './asm-customer-list.service';
+
+export const facadeProviders: Provider[] = [
+  AsmCustomerListService,
+  {
+    provide: AsmCustomerListFacade,
+    useExisting: AsmCustomerListService,
```
WeizhengSap, line 16: Do we want to make `AsmService` as facade?

feature-libs/asm/core/facade/asm.service.ts
```diff
@@ -20,7 +22,24 @@ import { AsmSelectors } from '../store/index';
...
+  /**
+   * @deprecated since 5.1
+   */
+  constructor(store: Store<StateWithAsm>);
+  constructor(
+    protected store: Store<StateWithAsm>,
+    @Optional() protected asmFacade?: AsmCustomerListFacade
+  ) {}
+
+  getCustomerLists(): Observable<QueryState<CustomerListsPage>> | undefined {
```
WeizhengSap: since we have `AsmCustomerListFacade`, so no need add this function in `AsmService`, right? Customer can get this function from `AsmCustomerListFacade`.

angular.json
```diff
@@ -1,5 +1,8 @@
 {
   "$schema": "./node_modules/@angular/cli/lib/config/schema.json",
+  "cli": {
+    "analytics": false
+  },
```
WeizhengSap: can we revert it back?

feature-libs/asm/core/facade/facade-providers.ts
ercultimate: `AsmService` cannot be converted to a facade without breaking changes.

feature-libs/asm/occ/adapters/occ-asm.adapter.ts
```diff
@@ -35,27 +39,70 @@ export class OccAsmAdapter implements AsmAdapter {
...
+      {},
+      {
+        baseSite: false,
+        prefix: false,
+      }
+    );
+
+    return this.http.get<CustomerListsPage>(url, { headers, params }).pipe(
+      this.converterService.pipeable(CUSTOMER_LISTS_NORMALIZER),
+      catchError((error) => throwError(normalizeHttpError(error)))
```
WeizhengSap: should catchError before convert, right?

feature-libs/asm/core/facade/asm.service.ts
```diff
@@ -53,6 +53,38 @@ export class AsmService {
     );
   }
 
+  /**
+   * Search for customers in a customer list
+   */
+  customerListCustomersSearch(options: CustomerSearchOptions): void {
```
WeizhengSap, line 59: can we also move them to `AsmCustomerListService`?

feature-libs/asm/root/facade/asm-customer-list.facade.ts
```diff
@@ -0,0 +1,25 @@
...
+  useFactory: () =>
+    facadeFactory({
+      facade: AsmCustomerListFacade,
+      feature: ASM_FEATURE,
+      methods: ['getCustomerLists', 'getCustomerListsState'],
+    }),
+})
+export abstract class AsmCustomerListFacade {
+  abstract getCustomerLists(): Observable<CustomerListsPage | undefined>;
+  abstract getCustomerListsState(): Observable<QueryState<CustomerListsPage>>;
```
WeizhengSap, line 24: this one is not used, right? I am fine to expose it.

feature-libs/asm/core/facade/asm-customer-list.service.ts
```diff
@@ -0,0 +1,33 @@
...
+import { AsmCustomerListFacade, CustomerListsPage } from '@spartacus/asm/root';
+import { Query, QueryService, QueryState } from '@spartacus/core';
+import { Observable } from 'rxjs';
+import { AsmConnector } from '../connectors/asm.connector';
+
+@Injectable()
+export class AsmCustomerListService implements AsmCustomerListFacade {
+  protected customerListQuery$: Query<CustomerListsPage> =
+    this.queryService.create(() => this.asmConnector.customerLists(), {
+      reloadOn: undefined,
```
WeizhengSap, line 17: do we want to reload the data when language change?

feature-libs/asm/root/facade/asm-customer-list.facade.ts
ercultimate: Correct, it is not used.  I had seen other services that use queries exposing both.

feature-libs/asm/occ/adapters/occ-asm.adapter.ts
ercultimate: I don't believe the order really matters for these operations, but I see that conventionally the error normalization is done first. 

The `pipeable` method uses `map` and does not interact with the observable's error out, so an error would pass by the converter with either ordering.

feature-libs/asm/core/facade/asm.service.ts
ercultimate: Not at the moment.  The models used are located in the `core` module while the facade is located in the `root` module.  Moving the methods would require a breaking change to move the models to `root`.

projects/storefrontlib/cms-components/misc/icon/icon.model.ts
```diff
@@ -45,8 +45,14 @@ export enum ICON_TYPE {
   OFF = 'OFF',
   LINK_OUT = 'LINK_OUT',
   EXPAND_ARROWS = 'EXPAND_ARROWS',
+  CLIPBOARD_LIST = 'CLIPBOARD_LIST',
```
WeizhengSap, line 48: If adding them is breaking change, maybe augment this enum in asm.

projects/storefrontlib/cms-components/misc/icon/fontawesome-icon.config.ts
```diff
@@ -45,9 +45,15 @@ export const fontawesomeIconConfig: IconConfig = {
       SORT_DOWN: 'fas fa-sort-down',
       SORT: 'fas fa-sort',
       EXPAND_ARROWS: 'fas fa-expand',
+      CLIPBOARD_LIST: 'fas fa-clipboard-list',
```
WeizhengSap, line 48: If adding them is breaking change, maybe augment this config in asm.

feature-libs/asm/core/connectors/asm.connector.spec.ts
```diff
@@ -64,13 +85,27 @@ describe('AsmConnector', () => {
...
+  });
+
+  it('should return customerLists results', (done) => {
+    spyOn(asmAdapter, 'customerLists').and.returnValue(
+      of(mockCustomerListPage)
+    );
+    asmConnector.customerLists().subscribe((results) => {
+      expect(results).toEqual(mockCustomerListPage);
+      done();
+    });
```
WeizhengSap, line 120: missing test for bindCart

projects/storefrontlib/cms-components/misc/icon/icon.model.ts
ercultimate: I don't believe this is a breaking change.  We are only adding new values to the enum, so no previously used values would be changed.  Additionally, enums cannot be extended, so there is little chance of conflicts.

projects/storefrontlib/cms-components/misc/icon/fontawesome-icon.config.ts
ercultimate: I don't believe this is a breaking change.  We are only adding new values to the configuration, so no previously used values would be changed.  The configurations are deep merged into one another, so any conflicting values provided by the merchant would be overridden.

feature-libs/asm/core/connectors/asm.connector.spec.ts
ercultimate: Tests added.

WeizhengSap (dismissed):

feature-libs/asm/root/config/asm-config.ts
```diff
@@ -3,7 +3,9 @@
  *
  * SPDX-License-Identifier: Apache-2.0
  */
-import { OccConfig } from '@spartacus/core';
+import { OccConfig, User } from '@spartacus/core';
+import { ICON_TYPE } from '@spartacus/storefront';
+import { CustomerListColumnActionType } from '../model/customer-list.model';
 
 export abstract class AsmConfig extends OccConfig {
```
WeizhengSap: no need to `extends OccConfig`, right? This config has nothing related to occ.

feature-libs/asm/components/customer-list/customer-list.component.html
```diff
@@ -0,0 +1,354 @@
...
+          cellContent;
+          context: { column: column, customerEntry: customerEntry }
+        "
+      ></ng-container>
+    </button>
+  </div>
+</ng-template>
+<ng-template #cellContent let-customerEntry="customerEntry" let-column="column">
+  <span *ngIf="!column.icon">{{
+    column.renderer ? column.renderer(customerEntry) : ''
```
justinlee01: I want to say Angular supports the Elvis operator in template HTML now
```suggestion
    column.renderer?.(customerEntry) || ''
```

feature-libs/asm/root/config/asm-config.ts
```diff
@@ -13,6 +15,19 @@ export abstract class AsmConfig extends OccConfig {
     customerSearch?: {
       maxResults?: number;
     };
+    customerList?: {
```
WeizhengSap, line 18: Why some of the asm config is moved from core to root?

justinlee01: I can explain this.
For this PR (https://github.com/SAP/spartacus/pull/16095) I added an interceptor `UserIdHttpHeaderInterceptor` that lived in `@spartacus/asm/root`.
As it needed to be turned on/off - depending on the customer's backend - a configuration was added to `AsmConfig`.
This caused a circular dependency - `@spartacus/asm/root` depended on `@spartacus/asm/core` which depends on `root`.
Patrick and I concluded the configuration should have been in `root` originally. There is precedence for this - for example, `CartConfig` is in root.
Let me know if you have any further questions.

feature-libs/asm/components/asm-main-ui/asm-main-ui.component.ts
```diff
@@ -160,32 +174,32 @@ export class AsmMainUiComponent implements OnInit {
...
-          this.startCustomerEmulationSession(selectedUser);
-          if (actionType === CustomerListColumnActionType.ORDER_HISTORY) {
-            this.routingService.go({ cxRoute: 'orders' });
+    this.launchDialogService?.openDialogAndSubscribe(
+      LAUNCH_CALLER.ASM_CUSTOMER_LIST,
+      this.element
+    );
+
+    this.subscription.add(
+      this.launchDialogService?.dialogClose
```
WeizhengSap: each time function `showCustomList` is called, a new subscription will be created. So, the code block `    this.subscription.add(...)` can be moved into `ngOnInit`, right?

kimhw0630: thanks, you are right we can move into ngOnInit.

feature-libs/asm/root/config/asm-config.ts
justinlee01: I suppose so, if it is old. `CmsConfig` and `AnonymousConsentsConfig` extend `OccConfig`, unless they're not meant to.

WeizhengSap: CmsConfig needs extends OccConfig, but I don't think AnonymousConsentsConfig should extend OccConfig.

kimhw0630: extends OccConfig removed.. thanks

WeizhengSap (approved):

feature-libs/asm/components/customer-list/customer-list.component.ts
```diff
@@ -0,0 +1,264 @@
...
+      (customerEntry.lastName?.charAt(0) ?? '')
+    );
+  }
+
+  startColumnAction(
+    customerEntry: User,
+    action: CustomerListColumnActionType
+  ): void {
+    this.selectedCustomer = customerEntry;
+    let closeValue: CustomerListAction = {
```
justinlee01, line 178:
```suggestion
    const closeValue: CustomerListAction = {
```

feature-libs/asm/components/customer-list/customer-list.component.ts
```diff
@@ -0,0 +1,264 @@
...
+
+  goToNextPage(): void {
+    if (this.currentPage >= this.maxPage) {
+      this.currentPage = this.maxPage;
+    } else {
+      if (this.loaded) {
+        this.currentPage++;
+        this.fetchCustomers();
+      }
+    }
```
justinlee01, line 198:
```suggestion
    } else if (this.loaded) {
      this.currentPage++;
      this.fetchCustomers();
    }
```

feature-libs/asm/components/customer-list/customer-list.component.ts
```diff
@@ -0,0 +1,264 @@
...
+        ]) => {
+          return {
+            byNameAsc: textByNameAsc,
+            byNameDesc: textByNameDesc,
+            byOrderDateAsc: textByOrderDateAsc,
+            byOrderDateDesc: textByOrderDateDesc,
+            byDateAsc: textByDateAsc,
+            byDateDesc: textByDateDesc,
+          };
+        }
```
justinlee01, line 249: I'm surprised prettier did not truncate ths

```suggestion
        ]) => ({
          byNameAsc: textByNameAsc,
          byNameDesc: textByNameDesc,
          byOrderDateAsc: textByOrderDateAsc,
          byOrderDateDesc: textByOrderDateDesc,
          byDateAsc: textByDateAsc,
          byDateDesc: textByDateDesc,
        })
```

feature-libs/asm/components/customer-list/customer-list.component.ts
kimhw0630: we will update in ASM 360

feature-libs/asm/components/customer-list/customer-list.component.ts
kimhw0630: we will update in ASM 360

feature-libs/asm/components/customer-list/customer-list.component.ts
kimhw0630: we will update in ASM 360

justinlee01 (approved):

PR #15709, opened 2022-05-05
chore: fix the bug that mini-cart total is not updated after checkout

WeizhengSap:


feature-libs/cart/base/components/mini-cart/mini-cart-component.service.ts
```diff
@@ -61,10 +60,9 @@ export class MiniCartComponentService {
     return this.activeCartRequired().pipe(
       switchMap((activeCartRequired) => {
         if (activeCartRequired) {
-          return this.activeCartFacade.getActive().pipe(
-            filter((cart) => !!cart.totalPrice),
-            map((cart) => cart.totalPrice?.formattedValue ?? '')
-          );
+          return this.activeCartFacade
+            .getActive()
+            .pipe(map((cart) => cart.totalPrice?.formattedValue ?? '$0.00'));
```
znikola: What if the default currency is not USD?

znikola (changes requested):

znikola (approved):

PR #15714, opened 2022-05-09
feat: Add BOPIS Skeleton

cawmichae:
Initial commit. Adds skeleton feature library for BOPIS, as well as a
reference to it in the storefront project.

CXSPA-6


feature-libs/cart/base/components/add-to-cart/add-to-cart.module.ts
```diff
@@ -1,12 +1,19 @@
...
 import { AddToCartComponent } from './add-to-cart.component';
 
 @NgModule({
-  imports: [CommonModule, ReactiveFormsModule, I18nModule, ItemCounterModule],
+  imports: [
+    CommonModule,
+    ReactiveFormsModule,
+    I18nModule,
+    ItemCounterModule,
+    PickupInStoreModule,
```
Pucek9: I would avoid such a dependency for the sake of
- circular dependency is possible - I think that `cart/pickup-in-store` will use `cart/base` (models?).
- `cart/base` should work as well without `cart/pickup-in-store module` - in this case, it's hardcoded, not all customers will need and want that

I would consider using cxOutlets or some other way to inject this module/component here - my favorite is to make it independent at all and use cms component, but since it has been agreed with the architect (was it?) - please tell me what has been agreed.

Pucek9:
To complete this feature, please add:
- root module
- feature module and register it in storefront app like [here](https://github.com/SAP/spartacus/compare/epic-pickup-in-store-poc#diff-14b359a06c80571d3436cf0a7874dc3f7661b5f6aa60aff6d5d0093085ca00bdR210)  (`PickupInStoreFeatureModule`)
- core / occ modules (I suspect it will use API)

at the end, to verify build the application in production mode

integration-libs/cds/tsconfig.lib.json
```diff
@@ -63,6 +67,12 @@
       "@spartacus/cart/wish-list": ["dist/cart/wish-list"],
       "@spartacus/cart/wish-list/root": ["dist/cart/wish-list/root"],
       "@spartacus/core": ["dist/core"],
+      "@spartacus/order/assets": ["dist/order/assets"],
```
Pucek9, line 76: Please avoid unnecessary change order of paths eg. in tsconfig files, it's not related to pickup-in-store and it's hard to review then.

If you are using some automated tool/configuration and recommend it, please suggest it in a separate ticket / pull request so that all developers in the project have the same rules/configuration.

mechris3: This was indeed auto-generated by Yarn. We don't really have an opinion on it, but when one of the tools did it, assumed it must be necessary.

feature-libs/cart/base/components/add-to-cart/add-to-cart.module.ts
mechris3: We haven't spoken to an arhictect. Most of these changes were based on what we saw in the POC. We are not so sure how cxOutlets work. Perhaps we could have a call later?

integration-libs/cds/tsconfig.lib.json
SVendittelli: Removed all the changes not related to pickup in store

feature-libs/cart/base/components/add-to-cart/add-to-cart.component.ts
```diff
@@ -57,6 +58,8 @@ export class AddToCartComponent implements OnInit, OnDestroy {
     quantity: new FormControl(1, { updateOn: 'blur' }),
   });
 
+  readonly CartOutlets = CartOutlets;
```
Pucek9, line 61:
```suggestion
```

feature-libs/cart/base/components/add-to-cart/add-to-cart.component.ts
```diff
@@ -10,6 +10,7 @@ import {
 import { FormControl, FormGroup } from '@angular/forms';
 import {
   ActiveCartFacade,
+  CartOutlets,
```
Pucek9, line 13:
```suggestion
```

SVendittelli:
Replaced by #15765

PR #15717, opened 2022-05-09
test: fix flaky cart validation e2e by inputting correct cart length

RadhepS:
close: https://jira.tools.sap/browse/CXSPA-399

The test passed before because for a brief moment after delete cart item, the item will remain in the cart with a quantity of 0.


projects/storefrontapp-e2e-cypress/cypress/helpers/cart-validation.ts
```diff
@@ -44,7 +44,7 @@ export function addMultipleProductsToCart(products) {
 
 export function removeItemAndCheckCartEntriesNumber(
   product,
-  expectedCartLength = 0
```
RadhepS, line 47: removed as I don't think this should be here. The default value is not used by anything referencing it and I think the caller should have to input their expected length.

WeizhengSap (approved):

PR #15724, opened 2022-05-10
feature: Improve PasswordVisibilityToggleDirective uasge

Pucek9:


Pucek9:
After discussing with @znikola we decided to reject this improvement. Code will stay less DRY, but more obvious at first glance (especially for less experienced devs)

PR #15744, opened 2022-05-13
Bugfix/cxspa-307

adolfobee:
This should fix the box shadowing when inside the product view after you click the arrows
Before:
image
After:
image


feature-libs/product/image-zoom/styles/_product-image-zoom-view.scss
```diff
@@ -20,6 +20,11 @@ cx-product-image-zoom-view {
         align-self: center;
         font-size: 2rem;
       }
+
+      .btn:focus,.btn:active {
+        outline: none !important;
```
developpeurweb: - We avoid using `!important` statements as per the [UI Styling and Accessibility Onboarding](https://wiki.wdf.sap.corp/wiki/display/spar/UI+Styling+and+Accessibility+Onboarding) guide _Styling> Avoid> 4_.
- In this case, adding `outline:none` even without the `!important` statement will also remove the blue Accessibility focus ring for keyboard navigation, which is mandatory.
- Everything seems to work fine by just removing line 25.

developpeurweb (changes requested):

developpeurweb (approved):

PR #15756, opened 2022-05-17
CXINT-78 Adding parameters jsSDKUrl, baseSite and sessionExpiration in CDC schematics

SherwinVarghese:
- Provided configurable parameters for CDC schematics to allow build pipelines to generate Spartacus modules
- Allows a new Angular project to add CDC schematics like -
 `ng add @spartacus/cdc@latest --interactive false  --javascriptUrl="https://eu1.gigya.com/API-KEY" --baseSite="electroncis-spa" --sessionExpiration=7200`

### DIFF
--- a/integration-libs/cdc/schematics/add-cdc/index.ts
+++ b/integration-libs/cdc/schematics/add-cdc/index.ts
@@ -29,7 +29,7 @@ import {
 } from '../constants';
 
 export function addCdcFeature(options: SpartacusCdcOptions): Rule {
-  return (tree: Tree, _context: SchematicContext): Rule => {
+  return (tree: Tree, context: SchematicContext): Rule => {
     const packageJson = readPackageJson(tree);
     validateSpartacusInstallation(packageJson);
 
@@ -37,20 +37,20 @@ export function addCdcFeature(options: SpartacusCdcOptions): Rule {
       addPackageJsonDependenciesForLibrary(peerDependencies, options),
 
       shouldAddFeature(CLI_CDC_FEATURE, options.features)
-        ? addCdc(options, _context)
+        ? addCdc(options, context)
         : noop(),
     ]);
   };
 }
 
 function addCdc(options: SpartacusCdcOptions, context: SchematicContext): Rule {
-  if (!options.jsSDKUrl) {
+  if (!options.javascriptUrl) {
     context.logger.warn(
-      `CDC JS SDK URL is not provided. Please run the schematic again, or make sure you update the jsSDKUrl.`
+      `CDC JS SDK URL is not provided. Please run the schematic again, or make sure you update the javascriptUrl.`
     );
   }
 
-  let jsSDKUrl = options.jsSDKUrl? options.jsSDKUrl: "<url-to-cdc-script>";
+  const JS_SDK_URL_PLACEHOLDER = '<url-to-cdc-script>';
 
   return addLibraryFeature(options, {
     folderName: CDC_FOLDER_NAME,
@@ -79,7 +79,9 @@ function addCdc(options: SpartacusCdcOptions, context: SchematicContext): Rule {
           cdc: [
             {
               baseSite: '${options.baseSite}',
-              javascriptUrl: '${jsSDKUrl}',
+              javascriptUrl: '${
+                options.javascriptUrl ?? JS_SDK_URL_PLACEHOLDER
+              }',
               sessionExpiration: ${options.sessionExpiration}
             },
           ],
--- a/integration-libs/cdc/schematics/add-cdc/index.ts
+++ b/integration-libs/cdc/schematics/add-cdc/index.ts
@@ -12,13 +12,13 @@ import {
   CDC_ROOT_MODULE,
   CLI_CDC_FEATURE,
   CLI_USER_PROFILE_FEATURE,
-  LibraryOptions as SpartacusCdcOptions,
   readPackageJson,
   shouldAddFeature,
   SPARTACUS_CDC,
   SPARTACUS_USER,
   validateSpartacusInstallation,
 } from '@spartacus/schematics';
+import { Schema as SpartacusCdcOptions } from './schema';
 import { peerDependencies } from '../../package.json';
 import {
   CDC_CONFIG,
@@ -37,13 +37,21 @@ export function addCdcFeature(options: SpartacusCdcOptions): Rule {
       addPackageJsonDependenciesForLibrary(peerDependencies, options),
 
       shouldAddFeature(CLI_CDC_FEATURE, options.features)
-        ? addCdc(options)
+        ? addCdc(options, _context)
         : noop(),
     ]);
   };
 }
 
-function addCdc(options: SpartacusCdcOptions): Rule {
+function addCdc(options: SpartacusCdcOptions, context: SchematicContext): Rule {
+  if (!options.jsSDKUrl) {
+    context.logger.warn(
+      `CDC JS SDK URL is not provided. Please run the schematic again, or make sure you update the jsSDKUrl.`
+    );
+  }
+
+  let jsSDKUrl = options.jsSDKUrl? options.jsSDKUrl: "<url-to-cdc-script>";
+
   return addLibraryFeature(options, {
     folderName: CDC_FOLDER_NAME,
     moduleName: CDC_MODULE_NAME,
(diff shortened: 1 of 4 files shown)


integration-libs/cdc/schematics/add-cdc/schema.json
```diff
@@ -25,6 +25,23 @@
...
+    "baseSite": {
+      "type": "string",
+      "description": "Name of the Spartcus base site",
+      "x-prompt": "[CDC] What is the base site? (optional)",
+      "default": "electronics-spa"
+    },
+    "sessionExpiration": {
+      "type": "number",
+      "description": "Session Expiration time in seconds",
+      "x-prompt": "[CDC] What is the CDC Session expiration time (in seconds)? (optional)",
```
bgambocjaviniar: To make things clearer, I believe these are all required than optional by looking at the 'old' config.

If that's the case, let us omit (optional) and place the default property for these three CLI params as it was in statically set in the config before.

Reason why I'm saying I'm thinking they are NOT optional is because if they are indeed optional, we should be omitting the creation of these properties in the config. For example as this was used as an example when we spoke in terms of what CDS/ISS did, you would need to create a different config for these different scenarios. https://github.com/SAP/spartacus/blob/develop/integration-libs/cds/src/schematics/add-cds/index.ts#L45-L87

integration-libs/cdc/schematics/add-cdc/schema.ts
```diff
@@ -0,0 +1,7 @@
+import { LibraryOptions } from '@spartacus/schematics';
+
+export interface Schema extends LibraryOptions {
+  baseSite?: string;
```
bgambocjaviniar: if my assumption is correct https://github.com/SAP/spartacus/pull/15756/files#r875015142, then remove the optional param and make it required

integration-libs/cdc/schematics/add-cdc/index.ts
```diff
@@ -37,13 +37,21 @@ export function addCdcFeature(options: SpartacusCdcOptions): Rule {
       addPackageJsonDependenciesForLibrary(peerDependencies, options),
 
       shouldAddFeature(CLI_CDC_FEATURE, options.features)
-        ? addCdc(options)
+        ? addCdc(options, _context)
```
bgambocjaviniar: `_` as a variable or before a variable is used to mention that it won't be used. Since we are now using context, let us remove `_`.

Therefore, line 40 should have `addCdc(options, context)` and line 32 should have `return (tree: Tree, context: SchematicContext): Rule => {`

integration-libs/cdc/schematics/add-cdc/index.ts
```diff
@@ -37,13 +37,21 @@ export function addCdcFeature(options: SpartacusCdcOptions): Rule {
...
 
-function addCdc(options: SpartacusCdcOptions): Rule {
+function addCdc(options: SpartacusCdcOptions, context: SchematicContext): Rule {
+  if (!options.jsSDKUrl) {
+    context.logger.warn(
+      `CDC JS SDK URL is not provided. Please run the schematic again, or make sure you update the jsSDKUrl.`
+    );
+  }
+
+  let jsSDKUrl = options.jsSDKUrl? options.jsSDKUrl: "<url-to-cdc-script>";
```
bgambocjaviniar: Since we're not re-assigning the variable, `const` should be used.

However, as mentioned [here](https://github.com/SAP/spartacus/pull/15756/files#r875015142), we most likely do not have to check as it will be a required param (if ever, but with a default value) 
Basically, if they decide to just press enter and skip the interactive params setting, they'll get the default value.

integration-libs/cdc/schematics/add-cdc/index.ts
```diff
@@ -70,9 +78,9 @@ function addCdc(options: SpartacusCdcOptions): Rule {
       content: `<${CDC_CONFIG}>{
           cdc: [
             {
-              baseSite: 'electronics-spa',
-              javascriptUrl: '<url-to-cdc-script>',
-              sessionExpiration: 3600
+              baseSite: '${options.baseSite}',
+              javascriptUrl: '${jsSDKUrl}',
```
bgambocjaviniar: In any case, if ever options.jsSDK is not available with defaults and such, therefore null / undefined, we can handle this easily by doing the snippet below

```suggestion
              javascriptUrl: '${options.jsSDKUrl ?? 'JS_SDK_URL_PLACEHOLDER'}',
```

Basically, make use of the nullish coalescing operator to catch undefined and null values.
This applies for baseSite and sessionExpiration.

Again, based on what I said here https://github.com/SAP/spartacus/pull/15756/files#r875015142 as it seems these config properties are required and NOT optional.

Just to repeat, if they were optional, for instance, sessionExpiration is optional, then you would have to configure the object as 

```
{
 baseSite: 'xxx',
 javaScriptUrl: 'xxx'
}
```

^ omitted sessionExpiration from the config as it is optional

bgambocjaviniar (changes requested):
Please do the following before taking a look at the reviews:

1 - please clone source and make a pull request from there instead of forking
2 - if you are using vscode, make sure you have prettier enabled as you can see in the checks from the screenshot, it is failing the `Validation Check` due to the fact that the code format does not adhere our 'convention'.
Screen Shot 2022-05-17 at 12 34 40 PM
P.S If you do not have prettier properly configured, you can just run `yarn prettier:fix` at the root level of the folder in order to automatically prettify files.

3- look at the code review changes.

integration-libs/cdc/schematics/add-cdc/schema.json
SherwinVarghese: Hi @bgambocjaviniar,

thanks for the review.

However, the `jsSDKUrl` is not optional and it is required to be given by the customer.
Unless the `jsSDKUrl` is not provided, the entire CDC functionality does not work.

Also, we do not have a default URL that can be used if the user does not provide any, since it would point to our dev. CDC instance. 

This was the reason for keeping `jsSDKUrl`mandatory.

Even if we have multiple configurations, we might not be able to expose our dev. CDC instances.
The other 2 configurations like `baseSite` and `sessionExpiration` have fallback default values which can be used, so they were mentioned as optional.


Please let me know your thoughts.

integration-libs/cdc/schematics/add-cdc/schema.ts
SherwinVarghese: sure @bgambocjaviniar , will do that, thanks!

integration-libs/cdc/schematics/add-cdc/index.ts
SherwinVarghese: agreed

integration-libs/cdc/schematics/add-cdc/index.ts
SherwinVarghese: that is correct, all these parameters are required, but the only difference is that `baseSite` and `sessionExpiration` can take in the predefined value, but for `jsSDKUrl`, the URL must be provided by the user.

SherwinVarghese:
Hi @bgambocjaviniar,


- Point 1: I am not able to push directly to the Spartacus repository.

`git push origin bugfix/CXINT-78`

```
 git push origin bugfix/CXINT-78
ERROR: Permission to SAP/spartacus.git denied to SherwinVarghese.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

Current git remotes `git remote -v`

```
origin	git@github.com:SAP/spartacus.git (fetch)
origin	git@github.com:SAP/spartacus.git (push)
sherwin	git@github.com:SherwinVarghese/spartacus.git (fetch)
sherwin	git@github.com:SherwinVarghese/spartacus.git (push)
```

If you have access rights, please provide write access to the repo.

- Point 2 and 3: I would be adopting the comments and doing `yarn prettier`

thanks

integration-libs/cdc/schematics/add-cdc/schema.json
bgambocjaviniar: I believe I already explained my thoughts. 

To simplify things
- all your options are required due to the fact that you have default values in case the user presses 'enter'. Meaning we still want to have those required configs to be placed in the file.
- Yes jsSDKUrl is not optional and I did not say anything about it being an issue, but nothing stops the user from pressing enter. Again, we would need a default value placed it in as mentioned in other instances below.

It is just a matter of following how 'CDS' constructed it as you pointed out somewhere in a slack thread.

bgambocjaviniar: I see that you took a look at the rest of the comments afterwards. https://github.com/SAP/spartacus/pull/15756#discussion_r876555817

integration-libs/cdc/schematics/add-cdc/index.ts
bgambocjaviniar: it can be predefined from the user, but nothings stops the user from pressing enter. Therefore, as mentioned previously, we would want to place a default value in case that happens.

All you need to do in the end afterwards is to document and let your customers know what they need to do. Example this CDS docs that explains the flow of things https://sap.github.io/spartacus-docs/cds-integration/#enabling-intelligent-selling-services-for-sap-commerce-cloud-in-spartacus

SherwinVarghese:
- Closing the duplicate PR.

PR #15772, opened 2022-05-20
CXSPA-445. E2E Test Run Dev CI7. Perf

giancorderoortiz:
DO NOT MERGE!

Test with CX 2105.11 (latest patch)


giancorderoortiz:
No longer needed

PR #15777, opened 2022-05-20
CXINT-78 Adding parameters jsSDKUrl, baseSite and sessionExpiration In CDC schematics

SherwinVarghese:
- Provided configurable parameters for CDC schematics to allow build pipelines to generate Spartacus modules
- Allows a new Angular project to add CDC schematics like -
```ng add @spartacus/cdc@latest --interactive false --javascriptUrl="https://eu1.gigya.com/API-KEY" --baseSite="electroncis-spa" --sessionExpiration=7200```

### DIFF
--- a/integration-libs/cdc/schematics/add-cdc/index.ts
+++ b/integration-libs/cdc/schematics/add-cdc/index.ts
@@ -50,8 +50,6 @@ function addCdc(options: SpartacusCdcOptions, context: SchematicContext): Rule {
     );
   }
 
-  const JS_SDK_URL_PLACEHOLDER = '<url-to-cdc-script>';
-
   return addLibraryFeature(options, {
     folderName: CDC_FOLDER_NAME,
     moduleName: CDC_MODULE_NAME,
@@ -80,7 +78,7 @@ function addCdc(options: SpartacusCdcOptions, context: SchematicContext): Rule {
             {
               baseSite: '${options.baseSite}',
               javascriptUrl: '${
-                options.javascriptUrl ?? JS_SDK_URL_PLACEHOLDER
+                options.javascriptUrl || 'JS_SDK_URL_PLACEHOLDER'
               }',
               sessionExpiration: ${options.sessionExpiration}
             },
--- a/integration-libs/cdc/schematics/add-cdc/index.ts
+++ b/integration-libs/cdc/schematics/add-cdc/index.ts
@@ -12,13 +12,13 @@ import {
   CDC_ROOT_MODULE,
   CLI_CDC_FEATURE,
   CLI_USER_PROFILE_FEATURE,
-  LibraryOptions as SpartacusCdcOptions,
   readPackageJson,
   shouldAddFeature,
   SPARTACUS_CDC,
   SPARTACUS_USER,
   validateSpartacusInstallation,
 } from '@spartacus/schematics';
+import { Schema as SpartacusCdcOptions } from './schema';
 import { peerDependencies } from '../../package.json';
 import {
   CDC_CONFIG,
@@ -29,7 +29,7 @@ import {
 } from '../constants';
 
 export function addCdcFeature(options: SpartacusCdcOptions): Rule {
-  return (tree: Tree, _context: SchematicContext): Rule => {
+  return (tree: Tree, context: SchematicContext): Rule => {
     const packageJson = readPackageJson(tree);
     validateSpartacusInstallation(packageJson);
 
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: integration-libs/cdc/schematics/add-cdc/index.ts [hunk 1]
```diff
@@ -70,9 +78,11 @@ function addCdc(options: SpartacusCdcOptions): Rule {
       content: `<${CDC_CONFIG}>{
           cdc: [
             {
-              baseSite: 'electronics-spa',
-              javascriptUrl: '<url-to-cdc-script>',
-              sessionExpiration: 3600
+              baseSite: '${options.baseSite}',
+              javascriptUrl: '${
+                options.javascriptUrl ?? JS_SDK_URL_PLACEHOLDER
```
bgambocjaviniar: I went ahead and verified the returned value of `javascriptUrl` when the user just presses `enter` (skipping the option to put a url themselves). The value is an empty string `''` instead of `undefined`, therefore nullish coalescing won't work in this case as you mentioned it was an `undefined` value that was returned before.

However, when it is automatically generated without the use of 'interactability', then it is undefined. All in all, if you tried manually to do `ng add @spartacus/cdc`, then you'll get an empty string instead of the default value.

Please change `??` to `||`.

full snippet for javascriptUrl below

```
javascriptUrl: `${options.javascriptUrl || 'JS_SDK_URL_PLACEHOLDER'}`,
```  

P.S. Instead of storing it in line 53, please just make it a string as a placeholder as shown in the snippet above.
Reason being, we have an ongoing  schematics 'refactor' work that makes default url like these to be put as 'SOMETHING_URL_PLACEHOLDER'. **HOWEVER**, if you are eager to leave it as '`<url-to-cdc-script>`', then put string instead as a one liner instead of storing it.

integration-libs/cdc/schematics/add-cdc/index_spec.ts
```diff
@@ -122,6 +122,44 @@ describe('Spartacus CDC schematics: ng-add', () => {
...
+            'ng-add',
+            { ...cdcFeatureOptions, javascriptUrl: '<dc>.gigya.com/<api-key>' },
+            appTree
+          )
+          .toPromise();
+      });
+
+      it('should set the given javascriptUrl', () => {
+        const module = appTree.readContent(featureModulePath);
+        expect(module).toContain<string>(`<dc>.gigya.com/<api-key>`);
```
bgambocjaviniar: instead of `expect(module).toContain<string>('<dc>.gigya.com/<api-key>');`

put `expect(module).toMatchSnapshot();`

P.S. go to `integration-libs/cdc` and run `yarn test:schematics` 
this will basically create an entry under the snapshot for this test case when you change it to match snapshot. This snap will help with future changes as it will detect the output of the snap if it is different to what it was before.

bgambocjaviniar: ^ snap is part of jest features

integration-libs/cdc/schematics/add-cdc/index_spec.ts
```diff
@@ -122,6 +122,44 @@ describe('Spartacus CDC schematics: ng-add', () => {
...
+            firstMessage = log.message;
+          }
+        });
+
+        appTree = await schematicRunner
+          .runSchematicAsync('ng-add', { ...cdcFeatureOptions }, appTree)
+          .toPromise();
+      });
+
+      it('show the warning', () => {
```
bgambocjaviniar: to semantically follow the naming of assertion tests, let us put it as `it('should show the warning', () => {`

bgambocjaviniar (changes requested):

integration-libs/cdc/schematics/add-cdc/index.ts
```diff
@@ -29,21 +29,29 @@ import {
...
 
-function addCdc(options: SpartacusCdcOptions): Rule {
+function addCdc(options: SpartacusCdcOptions, context: SchematicContext): Rule {
+  if (!options.javascriptUrl) {
+    context.logger.warn(
+      `CDC JS SDK URL is not provided. Please run the schematic again, or make sure you update the javascriptUrl.`
+    );
+  }
+
+  const JS_SDK_URL_PLACEHOLDER = '<url-to-cdc-script>';
```
bgambocjaviniar: please remove from this https://github.com/SAP/spartacus/pull/15777/files#r880884597

integration-libs/cdc/schematics/add-cdc/index.ts
SherwinVarghese: thanks for the update Brian, I will keep it as `JS_SDK_URL_PLACEHOLDER` so that it is consistent across the libraries.

Will write some tests around the same too.

integration-libs/cdc/schematics/add-cdc/index_spec.ts
SherwinVarghese: agreed

integration-libs/cdc/schematics/add-cdc/index_spec.ts
SherwinVarghese: yes, will update the test case Brian.

bgambocjaviniar (changes requested):
I will be updating the snap and test as what was pushed broke the pipeline without verification

bgambocjaviniar (approved):

PR #15781, opened 2022-05-22
Fix routing of ParagraphComponent HTML Links containing queryParams #15780

### LINKED ISSUE #15780: ParagraphComponent Link does not work with Query Parameter or Fragment
**Describe the bug**
With the new Spartacus Version 4.3.3, the ParagraphComponent will use the Angular Routing for handling HTML links. This works fine for "normal" links, but does not work for URLs with Query Params or Fragments in it (like `/test?someQuery=true`). Angular router expects to have the query params in a separate object in the navigate function.

- Spartacus 4.3.3 (latest currently)

**To Reproduce**
1. Create a CMS Page with a ParagraphComponent on it. The content should be: `<a href="/test?someQuery=true">Some Testlink</a>`.
2. Navigate to the page created in step 1
3. Click on the "Some Testlink" link.
4. Spartacus will use the Angular Router to navigate to `test%3FsomeQuery%3Dtrue` (notice the escaped questionmark and equal sign). The params are not correcty passed to the browser.

**Expected behavior**
The browser should navigate to `/test?someQuery=true`

**Desktop (my specs, but this bug exists on all browsers):**
 - OS: Ubuntu
 - Browser Chrome
 - Version 101.0.4951.64

**Additional context**
Bug only exists since Spartacus Version 4.3.3

dev-jan:
Use the Angular Routing navigateByUrl() method to correctly route all kinds
of URLs, including the ones with query params and fragments. Otherwise
URLs containing query params or fragments are broken.

Closes https://github.com/SAP/spartacus/issues/15780


dev-jan:
@Platonn any chance you can review this fix?

PR #15785, opened 2022-05-23
feature/CXINT-22 CDC-COMM:SPA UI - Headless Commerce: Login & Registration 

SherwinVarghese:
Requirement Details:

https://wiki.wdf.sap.corp/wiki/display/GIGYA/SAP+Commerce+Cloud+-+Built-in+Extension

Login & Registration:

The current built-in extension allows the registration and login flows to be performed without the use of the screensets. Instead, the WebSDK functions - such as gigya.accounts.login or gigya.accounts.register - can be used, as per below recording.


feature-libs/user/account/components/login-form/login-form.module.ts
```diff
@@ -37,7 +38,7 @@ import { LoginFormComponent } from './login-form.component';
             {
               provide: LoginFormComponentService,
               useClass: LoginFormComponentService,
-              deps: [AuthService, GlobalMessageService, WindowRef],
+              deps: [AuthService, GlobalMessageService, WindowRef, CdcJsService],
```
Platonn: let's overwrite in CMS mapping in CDC module, not in original user module

integration-libs/cdc/core/auth/services/user-authentication/cdc-login.service.ts
```diff
@@ -0,0 +1,42 @@
...
+import { CdcJsService } from "integration-libs/cdc/root";
+import { AuthService } from "projects/core/src/auth";
+import { GlobalMessageService } from "projects/core/src/global-message";
+import { WindowRef } from "projects/core/src/window";
+import { BehaviorSubject } from "rxjs";
+
+@Injectable({
+  providedIn: 'root',
+})
+export class CdcLoginComponentService extends LoginFormComponentService {
```
Platonn: cdc-login-component.service.ts ?

integration-libs/cdc/core/auth/services/user-authentication/cdc-login.service.ts
```diff
@@ -0,0 +1,42 @@
...
+  login() {
+    this.busy$.next(true);
+    this.cdcJsService.didLoad().subscribe((cdcLoaded) => {
+      if (!cdcLoaded) {
+        console.log(
+          'CDC Gigya SDK not loaded, logging in using normal service'
+        );
+        super.login();
+      } else {
+        console.log('Logging in using CDC Gigya SDK');
```
Platonn: let's drop console.logs unless they are really justified

integration-libs/cdc/core/profile/facade/facade-providers.ts
```diff
@@ -0,0 +1,31 @@
...
+    useExisting: UserEmailService,
+  },
+  {
+    provide: UserPasswordFacade,
+    useExisting: UserPasswordService,
+  },
+  {
+    provide: UserProfileFacade,
+    useExisting: UserProfileService,
+  },
```
Platonn: Do we need to provide originals? (and UserProfileConnector original too )? Let's double check.

integration-libs/cdc/root/service/cdc-js.service.ts
```diff
@@ -38,7 +39,8 @@ export class CdcJsService implements OnDestroy {
     protected auth: AuthService,
     protected zone: NgZone,
     protected userProfileFacade: UserProfileFacade,
-    @Inject(PLATFORM_ID) protected platform: any
+    @Inject(PLATFORM_ID) protected platform: any,
+    protected globalMessageService: GlobalMessageService
```
Platonn, line 50: **If it goes to a minor (4.4 or 5.1), then we need to mitigate breaking changes**
To mitigate breaking change for customer, new contructor params should be optional (`?` in TS)

And handle the logic when new dependency is null/undefined, e.g.:
```ts
this.globalMessageService?.add(...)
```

**If it goes to major (5.0)** 
let's just document the change and provide migrations
Kris TODO: link here to our docs for changing constructor

integration-libs/cdc/core/profile/cdc-user-profile.module.ts
```diff
@@ -0,0 +1,13 @@
+import { NgModule } from '@angular/core';
+import { UserProfileComponentsModule } from '@spartacus/user/profile/components';
+import { UserProfileOccModule } from '@spartacus/user/profile/occ';
+import { CDCUserProfileCoreModule } from './cdc-user-profile-core.module';
+
+@NgModule({
+  imports: [
+    CDCUserProfileCoreModule,
+    UserProfileOccModule,
+    UserProfileComponentsModule,
```
Platonn: Let's double check if we need to import original modules here?

Platonn: Let's follow the example of DigitalPaymentsModule, which imports the original feature module + our overwrites/extensions

Platonn: Note: In the short future we will change the architecture -> the wrapper modules will live in the customer's codebase. See PR https://github.com/SAP/spartacus/pull/15459

projects/storefrontapp/src/app/spartacus/features/cdc-feature.module.ts
```diff
@@ -27,6 +30,29 @@ import { CmsConfig, provideConfig } from '@spartacus/core';
...
+          providers: [
+            {
+              provide: LoginFormComponentService,
+              useClass: CdcLoginComponentService,
+              deps: [AuthService, GlobalMessageService, WindowRef, CdcJsService],
+            },
+          ],
+        },
+      },
+    }),
```
Platonn: Let's change it to:
```ts
provideConfig(<CmsConfig>{
      featureModules: {
        [USER_ACCOUNT_FEATURE]: {
          module: () =>
            import('@spartacus/cdc/???').then((m) => m.CDCUserAccountModule),
        },
      },
    }),
```

btw. we might consider renaming of the modules and paths, e.g.
```
import('@spartacus/cdc/login').then((m) => m.CDCLoginModule),
```

integration-libs/cdc/root/service/cdc-js.service.ts
Platonn: Here's the link to the docs: 
https://sap.github.io/spartacus-docs/breaking-changes/#adding-new-constructor-dependencies-in-minor-versions

Platonn: Please discuss with Bill first the targeted release version, so you know which coding strategy to use (backward compatibiity vs. breaking change followed with migration docs 5_0.md and constructor update schematics)

projects/storefrontapp/src/app/spartacus/features/cdc-feature.module.ts
Platonn: As we discussed in the call, the possible alternatives for new entry points are:
(we like options 2 and 3 more, with a tendency to 2, although it's not very strong opinion)

1. in the root folder (sibling to /core, /components)
```
@spartacus/cdc/core
@spartacus/cdc/components
NEW:
@spartacus/cdc/login
@spartacus/cdc/register
@spartacus/cdc/update-profile (in the future)
```

2. nest inside /components/
```
@spartacus/cdc/core
@spartacus/cdc/components
NEW:
@spartacus/cdc/components/login
@spartacus/cdc/components/register
@spartacus/cdc/components/update-profile (in the future)
```


3. nest inside new folder /extensions/
```
@spartacus/cdc/root
@spartacus/cdc/components
NEW:
@spartacus/cdc/extensions/login
@spartacus/cdc/extensions/register
@spartacus/cdc/extensions/update-profile (in the future)
```

I have hard time choosing between 2. and 3. But I'm a little more fond of 2. (nest in /components/).

### In the future, when we implement CDC update-profile:
In the future, when we implement the CDC update-profile component, it will be no longer optimal for customers to load in production the JS logic for both screenset component and the individual extended components. Then I'll suggest to allow customers for choosing (via installation schematics) which implementation do they want: screen set (dynamic iframe driven) vs. Spartacus extended components. And then avoid loading JS of unused implementation variant

Btw. Then it will make sense (although it's only a cosmetic change), to move out the GigyaRaasComponent (iframe creator) from:
```
@spartacus/cdc/components -> @spartacus/cdc/components/screen-set
```

integration-libs/cdc/core/profile/facade/cdc-user-register.service.ts
```diff
@@ -0,0 +1,106 @@
...
+} from '@spartacus/user/profile/root';
+import {
+  UserProfileConnector, UserProfileService
+} from 'feature-libs/user/profile/core';
+import { CdcJsService } from 'integration-libs/cdc/root';
+import { Observable } from 'rxjs';
+import { tap } from 'rxjs/operators';
+
+@Injectable()
+export class CDCUserRegisterService implements UserRegisterFacade {
```
Platonn: As we discussed in the call, instead of overwriting UserRegisterFacade, let's introduce an intermediate layer of the `register-component.service`, similar to `login-form-component.service`. Thanks to that we don't need to implement the guest register method which doesn't make sense in the context of CDC. See the diagram below:
image

feature-libs/user/account/components/login-form/login-form.module.ts
SherwinVarghese: agreed.

integration-libs/cdc/core/auth/services/user-authentication/cdc-login.service.ts
SherwinVarghese: noted, thanks

integration-libs/cdc/core/auth/services/user-authentication/cdc-login.service.ts
SherwinVarghese: agreed

integration-libs/cdc/root/service/cdc-js.service.ts
SherwinVarghese: It is decided to go for 5.0 version, would be proceeding with documentation update.
thanks

integration-libs/cdc/core/profile/facade/cdc-user-register.service.ts
SherwinVarghese: agreed, will be creating a new `register-component.service.ts` file

integration-libs/cdc/core/profile/facade/facade-providers.ts
SherwinVarghese: Hi @Platonn,

While removing the `UserProfileService` from the file, I get the error in runtime -

The register page does not load.

```
core.mjs:6485 ERROR Error: NullInjectorError: No provider for UserProfileService!
    at NullInjector.get (core.mjs:11120:1) [angular]
    at R3Injector.get (core.mjs:11287:1) [angular]
    at R3Injector.get (core.mjs:11287:1) [angular]
    at R3Injector.get (core.mjs:11287:1) [angular]
    at R3Injector.get (core.mjs:11287:1) [angular]
    at injectInjectorOnly (core.mjs:4765:1) [angular]
    at Module.ɵɵinject (core.mjs:4769:1) [angular]
    at Object.CDCUserRegisterService_Factory [as factory] (cdc-user-register.service.ts:22:36) [angular]
    at R3Injector.hydrate (core.mjs:11457:1) [angular]
    at R3Injector.get (core.mjs:11276:1) [angular]
    at injectInjectorOnly (core.mjs:4765:1) [angular]
    at ɵɵinject (core.mjs:4769:1) [angular]
    at Object.factory (core.mjs:11544:1) [angular]
    at R3Injector.hydrate (core.mjs:11457:1) [angular]
defaultErrorLogger @ core.mjs:6485
handleError @ core.mjs:6532
next @ core.mjs:26046
__tryOrUnsub @ Subscriber.js:183
next @ Subscriber.js:122
_next @ Subscriber.js:72
next @ Subscriber.js:49
next @ Subject.js:39
emit @ core.mjs:22423
(anonymous) @ core.mjs:25514
invoke @ zone.js:372
run @ zone.js:134
runOutsideAngular @ core.mjs:25387
onHandleError @ core.mjs:25514
handleError @ zone.js:376
runTask @ zone.js:181
invokeTask @ zone.js:487
ZoneTask.invoke @ zone.js:476
data.args.<computed> @ zone.js:2541
setTimeout (async)
scheduleTask @ zone.js:2543
scheduleTask @ zone.js:393
onScheduleTask @ zone.js:283
scheduleTask @ zone.js:386
scheduleTask @ zone.js:221
scheduleMacroTask @ zone.js:244
scheduleMacroTaskWithCurrentZone @ zone.js:679
(anonymous) @ zone.js:2585
proto.<computed> @ zone.js:975
hostReportError @ hostReportError.js:2
error @ Observer.js:11
__tryOrUnsub @ Subscriber.js:183
error @ Subscriber.js:135
_error @ Subscriber.js:75
error @ Subscriber.js:55
_error @ Subscriber.js:75
error @ Subscriber.js:55
_error @ tap.js:56
error @ Subscriber.js:55
_trySubscribe @ Observable.js:50
subscribe @ Observable.js:28
call @ tap.js:16
subscribe @ Observable.js:23
call @ finalize.js:11
subscribe @ Observable.js:23
launchComponent @ component-wrapper.directive.ts:110
(anonymous) @ component-wrapper.directive.ts:73
__tryOrUnsub @ Subscriber.js:183
next @ Subscriber.js:122
_next @ Subscriber.js:72
next @ Subscriber.js:49
(anonymous) @ subscribeToArray.js:3
_trySubscribe @ Observable.js:42
subscribe @ Observable.js:28
(anonymous) @ defer.js:15
_trySubscribe @ Observable.js:42
subscribe @ Observable.js:28
ngOnInit @ component-wrapper.directive.ts:67
callHook @ core.mjs:2542
callHooks @ core.mjs:2511
executeInitAndCheckHooks @ core.mjs:2462
refreshView @ core.mjs:9499
refreshEmbeddedViews @ core.mjs:10609
refreshView @ core.mjs:9508
refreshEmbeddedViews @ core.mjs:10609
refreshView @ core.mjs:9508
refreshEmbeddedViews @ core.mjs:10609
refreshView @ core.mjs:9508
refreshComponent @ core.mjs:10655
refreshChildComponents @ core.mjs:9280
refreshView @ core.mjs:9534
refreshEmbeddedViews @ core.mjs:10609
refreshView @ core.mjs:9508
refreshEmbeddedViews @ core.mjs:10609
refreshView @ core.mjs:9508
refreshComponent @ core.mjs:10655
refreshChildComponents @ core.mjs:9280
refreshView @ core.mjs:9534
refreshEmbeddedViews @ core.mjs:10609
refreshView @ core.mjs:9508
refreshEmbeddedViews @ core.mjs:10609
refreshView @ core.mjs:9508
refreshComponent @ core.mjs:10655
refreshChildComponents @ core.mjs:9280
refreshView @ core.mjs:9534
renderComponentOrTemplate @ core.mjs:9598
tickRootContext @ core.mjs:10829
detectChangesInRootView @ core.mjs:10854
detectChanges @ core.mjs:21439
tick @ core.mjs:26449
(anonymous) @ core.mjs:26304
invoke @ zone.js:372
onInvoke @ core.mjs:25488
invoke @ zone.js:371
run @ zone.js:134
run @ core.mjs:25342
next @ core.mjs:26303
__tryOrUnsub @ Subscriber.js:183
next @ Subscriber.js:122
_next @ Subscriber.js:72
next @ Subscriber.js:49
next @ Subject.js:39
emit @ core.mjs:22423
checkStable @ core.mjs:25410
onHasTask @ core.mjs:25505
hasTask @ zone.js:426
_updateTaskCount @ zone.js:447
_updateTaskCount @ zone.js:274
runTask @ zone.js:195
drainMicroTaskQueue @ zone.js:582
Promise.then (async)
scheduleMicroTask @ zone.js:565
scheduleTask @ zone.js:396
onScheduleTask @ zone.js:283
scheduleTask @ zone.js:386
scheduleTask @ zone.js:221
scheduleMicroTask @ zone.js:241
scheduleResolveOrReject @ zone.js:1266
resolvePromise @ zone.js:1204
(anonymous) @ zone.js:1120
webpackJsonpCallback @ jsonp chunk loading:77
(anonymous) @ default-feature-libs_user_profile_components_public_api_ts-feature-libs_user_profile_occ_publ-4d6199.js:2
Show 79 more frames
```

integration-libs/cdc/core/profile/cdc-user-profile.module.ts
SherwinVarghese: Hi @Platonn,

While removing the UserProfileService from the file, I get the error in runtime -

The register page does not load.

```
ERROR TypeError: Cannot read properties of undefined (reading 'pipe')
    at ComponentWrapperDirective.launchComponent (component-wrapper.directive.ts:97:8)
    at SafeSubscriber._next (component-wrapper.directive.ts:73:16)
    at SafeSubscriber.__tryOrUnsub (Subscriber.js:183:1)
    at SafeSubscriber.next (Subscriber.js:122:1)
    at Subscriber._next (Subscriber.js:72:1)
    at Subscriber.next (Subscriber.js:49:1)
    at Observable._subscribe (subscribeToArray.js:3:1)
    at Observable._trySubscribe (Observable.js:42:1)
    at Observable.subscribe (Observable.js:28:1)
    at Observable._subscribe (defer.js:15:1)
    at Observable._trySubscribe (Observable.js:42:1)
    at Observable.subscribe (Observable.js:28:1)
    at ComponentWrapperDirective.ngOnInit (component-wrapper.directive.ts:67:8)
    at callHook (core.mjs:2542:1)
    
    ```

File: integration-libs/cdc/components/login/cdc-user-login.module.ts [hunk 9]
```diff
@@ -0,0 +1,33 @@
...
+    FormErrorsModule,
+    SpinnerModule,
+  ],
+  providers: [
+    provideConfig(<CmsConfig>{
+      cmsComponents: {
+        ReturningCustomerLoginComponent: {
+          component: LoginFormComponent,
+          guards: [NotAuthGuard],
+          providers: [ ...facadeProviders],
```
Platonn: Could you list those providers directly here, please? Those are just services providers for the component.

The term and concept of `facadeProviders` is reserved for providing the concrete Services for the  [Facades (proxy for lazy loading the service's implementation)](https://sap.github.io/spartacus-docs/proxy-facades/#defining-proxy-facades)

feature-libs/user/profile/components/register/register-component.service.ts
```diff
@@ -0,0 +1,52 @@
...
+  register(user: UserSignUp): Observable<User> {
+    return this.registerCommand.execute({ user });
+  }
+
+  /**
+   * Returns titles that can be used for the user profiles.
+   */
+  getTitles(): Observable<Title[]> {
+    return this.userProfile.getTitles();
+  }
```
Platonn: Instead of copy-pasting the implementation (and code comments like `// this is a compatibility mechanism only...` 😅 ) of the `UserRegisterService`, let's just delegate the calls to this service (to be precise: to it's facade `UserRegisterFacade`)

See the bottom right of the following chart https://github.com/SAP/spartacus/pull/15785#discussion_r884857549



```suggestion
export class RegisterComponentService {
  constructor(
    protected userRegisterFacade: UserRegisterFacade
  ) {}

  /**
   * Register a new user.
   *
   * @param submitFormData as UserRegisterFormData
   */
  register(user: UserSignUp): Observable<User> {
    return this.userRegisterFacade.register(user);
  }

  /**
   * Returns titles that can be used for the user profiles.
   */
  getTitles(): Observable<Title[]> {
    return this.userRegisterFacade.getTitles();
  }
```

integration-libs/cdc/components/login/facade/facade-providers.ts
```diff
@@ -0,0 +1,15 @@
...
+import { CdcJsService } from 'integration-libs/cdc/root';
+import { AuthService } from 'projects/core/src/auth';
+import { GlobalMessageService } from 'projects/core/src/global-message';
+import { WindowRef } from 'projects/core/src/window';
+import { CdcLoginComponentService } from '.';
+
+export const facadeProviders: Provider[] = [
+  {
+    provide: LoginFormComponentService,
+    useClass: CdcLoginComponentService,
```
Platonn: For consistency, could you align the naming with the original `LoginFormComponentService`, please: 
`CdcLoginComponentService` -> `CdcLoginFormComponentService`

integration-libs/cdc/components/register/facade/cdc-register-component.service.ts
```diff
@@ -0,0 +1,85 @@
...
+            {
+              key: 'errorHandlers.scriptFailedToLoad',
+            },
+            GlobalMessageType.MSG_TYPE_ERROR
+          );
+          return new Observable<User>();
+        }
+    }));
+      //return cdcLoaded?  this.registerWihoutScreenSetCommand.execute({ user }): this.registerCommand.execute({ user });
+    //}));
```
Platonn: Let's clean up the commented-out code here and in other places in this PR

integration-libs/cdc/components/register/facade/facade-providers.ts
```diff
@@ -0,0 +1,31 @@
...
+  },
+  {
+    provide: UserProfileFacade,
+    useExisting: UserProfileService,
+  },
+  {
+    provide: UserRegisterFacade,
+    useExisting: CDCRegisterComponentService,
+  },
+];
```
Platonn: We shouldn't re-declare all those facade's providers. They are already defined in the original User Profile feature. 

Morevoer my intention was that `CDCRegisterComponentService` should overwrite the `RegiserComponentService` but not the global facade (`UserRegisterService`/`UserRegisterFacade`).

The `CDCRegisterComponentService` should overwrite `RegiserComponentService` on the CMS-component mapping level - similar as `CdcLoginFormComponentService` overwrites `LoginFormComponentService` on the CMS-component mapping level.
See diagram https://github.com/SAP/spartacus/pull/15785#discussion_r884857549

integration-libs/cdc/root/service/cdc-js.service.ts
Platonn: Therefore I think we should provide the migration docs (and possibly migration schematics) in this PR.

integration-libs/cdc/components/login/cdc-user-account.module.ts
```diff
@@ -0,0 +1,13 @@
+import { NgModule } from '@angular/core';
+import { UserAccountCoreModule } from '@spartacus/user/account/core';
+import { UserAccountOccModule } from '@spartacus/user/account/occ';
+import { CDCUserAccountComponentsModule } from './cdc-user-account-component.module';
+
+@NgModule({
+  imports: [
+    UserAccountCoreModule,
+    UserAccountOccModule,
+    CDCUserAccountComponentsModule,
```
Platonn: I believe this could be replaced to just `imports: [UserAccountModule, CDCUserLoginModule]`.
But let's leave it for now. It will be refactored anyway, when we introduce installation schematics that create wrapper modules in the customer's app.

integration-libs/cdc/root/service/cdc-js.service.ts
Platonn: Therefore I think we should provide the migration docs (and possibly migration schematics) in this PR.

projects/storefrontapp/src/app/spartacus/features/cdc-feature.module.ts
```diff
@@ -12,12 +14,23 @@ import { CmsConfig, provideConfig } from '@spartacus/core';
           javascriptUrl: '',
           sessionExpiration: 3600,
         },
+        // {
+        //   baseSite: 'electronics-spa',
+        //   javascriptUrl:
+        //     'https://cdns.us1.gigya.com/JS/gigya.js?apikey=4_Ds_nEtQ8W58KZt0ckVVwvQ',
+        //   sessionExpiration: 3600,
+        // },
```
znikola: Is it safe to remove these commented-out lines?

feature-libs/user/profile/components/register/register-component.service.ts
SherwinVarghese: agreed, thanks!

integration-libs/cdc/components/login/cdc-user-login.module.ts
SherwinVarghese: Noted, thanks

integration-libs/cdc/components/login/facade/facade-providers.ts
SherwinVarghese: agreed, would be following the naming of files as in the core library.
- `LoginFormComponentService` -> `CDCLoginFormComponentService`
- `RegisterComponentService`-> `CDCRegisterComponentService`

integration-libs/cdc/components/register/facade/facade-providers.ts
SherwinVarghese: hi @Platonn,

yes it makes sense, I was trying to follow the structuring of the core library modules for login / registration. Will add the `CDCRegisterComponentService` to overwrite `RegiserComponentService` CMS-component mapping.

thanks

integration-libs/cdc/components/login/cdc-user-account.module.ts
SherwinVarghese: will check if this is necessary with CMS-component mapping.

feature-libs/user/profile/components/register/register-component.service.ts
```diff
@@ -0,0 +1,26 @@
...
+import { Observable } from 'rxjs';
+
+@Injectable()
+export class RegisterComponentService {
+  constructor(protected userRegisterService: UserRegisterService) {}
+
+  /**
+   * Register a new user.
+   *
+   * @param submitFormData as UserRegisterFormData
```
znikola: Please remove this line.

feature-libs/user/profile/components/register/register-component.service.ts
```diff
@@ -0,0 +1,26 @@
+import { Injectable } from '@angular/core';
+import { Title, User } from '@spartacus/core';
+import { UserRegisterService } from '@spartacus/user/profile/core';
+import { UserSignUp } from '@spartacus/user/profile/root';
+import { Observable } from 'rxjs';
+
+@Injectable()
+export class RegisterComponentService {
+  constructor(protected userRegisterService: UserRegisterService) {}
```
znikola: Please inject the `UserRegisterFacade` instead of the implementation service

integration-libs/cdc/components/register/cdc-user-profile.module.ts
```diff
@@ -0,0 +1,21 @@
...
+    UserProfileModule,
+  ],
+  providers: [
+
+    CDCRegisterComponentService,
+    {
+      provide: UserRegisterFacade,
+      useExisting: CDCRegisterComponentService,
+    },
+  ],
```
znikola: Let's remove this part, as the `CDCRegisterComponentService` should be provided using the CMS component mapping's `providers` array.

projects/schematics/src/shared/constants.ts
```diff
@@ -1073,3 +1073,5 @@ export const CSS_V5_DOCS_URL =
 export const BANNER_COMPONENT = 'BannerComponent';
 export const GENERIC_LINK_COMPONENT = 'GenericLinkComponent';
 export const GENERIC_LINK_COMPONENT_SERVICE = 'GenericLinkComponentService';
+
+export const CDC_JS_SERVICE = 'CdcJsService';
```
znikola, line 1141: A note for myself to double check if need this here.

feature-libs/user/profile/components/register/register-component.service.ts
znikola: Also, please rename the `userRegisterService` variable to `userRegisterFacade`

integration-libs/cdc/components/register/cdc-user-profile.module.ts
znikola: You should do it like it's done in `integration-libs/cdc/components/login/facade/cdc-login-form.module.ts`.

znikola: First, let's adapt the existing `RegisterComponent`:

- Please inject the `RegisterComponentService` into it.
- Next, you need to call the service's methods

```ts
ngOnInit() {
    this.titles$ = this.registerComponentService.getTitles().pipe(
      map((titles: Title[]) => {
        return titles.sort(sortTitles);
      })
    );
...
}
...

  registerUser(): void {
    this.isLoading$.next(true);
    this.registerComponentService
      .register(this.collectDataFromRegisterForm(this.registerForm.value))
      .subscribe({
        next: () => this.onRegisterUserSuccess(),
        complete: () => this.isLoading$.next(false),
      });
  }
```

- Please remove the `protected userRegister: UserRegisterFacade` injection from the constructor, as it's no longer needed.

znikola: Next, you need to:
- create the cdc register component, and extend the existing RegisterComponet from user-profile
- create the cdc register component module
- in the module, override the default CMS mapping, and specify the newly created cdc register component
- add the `providers` array to the CMS mapping. Please match the deps according to the `RegisterComponentService`'s constructor.
```ts
      providers: [
            {
              provide: RegisterComponentService,
              useClass: RegisterComponentService,
              deps: [...],
            },
          ],
```

znikola: As a final step, please document these changes in the breaking changes doc file, and please write a constructor migration for this component.

integration-libs/cdc/components/register/facade/cdc-register-component.service.ts
```diff
@@ -0,0 +1,69 @@
...
+  CommandService,
+  GlobalMessageService,
+  GlobalMessageType,
+  UserActions
+} from '@spartacus/core';
+import { User } from '@spartacus/user/account/root';
+import { RegisterComponentService } from '@spartacus/user/profile/components';
+import { UserRegisterService } from '@spartacus/user/profile/core';
+import { UserSignUp } from '@spartacus/user/profile/root';
+import { CdcJsService } from 'integration-libs/cdc/root';
```
znikola: Please fix all the imports that are using an absolute path, such as this one for example.
The imports should either be related (if they're from the same entry point), or they should import from `@spartacus/...`

integration-libs/cdc/components/register/facade/cdc-register-component.service.ts
```diff
@@ -0,0 +1,69 @@
...
+    protected cdcJSService: CdcJsService,
+    protected globalMessageService: GlobalMessageService
+  ) {
+    super(userRegisterService);
+  }
+
+  /**
+   * Register a new user.
+   *
+   * @param submitFormData as UserRegisterFormData
```
znikola: Please remove this jsdoc comment, as it's not relevant

feature-libs/user/account/components/login-form/login-form.module.ts
```diff
@@ -10,7 +10,7 @@ import {
   NotAuthGuard,
   provideDefaultConfig,
   UrlModule,
-  WindowRef,
+  WindowRef
```
znikola: hmm, seems that you didn't run prettier.
Please run `yarn prettier:fix` locally. I suggest doing it in a separate commit, as the number of files can get quite big

feature-libs/user/profile/assets/translations/en/user-profile.ts
```diff
@@ -57,6 +57,7 @@ export const userProfile = {
     bothPasswordMustMatch: 'Both password must match',
     titleRequired: 'Title is required.',
     postRegisterMessage: 'Please log in with provided credentials.',
+    loginSuccessful: 'Login Success',
```
znikola: I can't seem to find where is this used?

integration-libs/cdc/components/login/facade/cdc-login-form-component.service.ts
```diff
@@ -0,0 +1,46 @@
+import { Injectable } from "@angular/core";
+import { CdcJsService } from "@spartacus/cdc/root";
+import { LoginFormComponentService } from "@spartacus/user/account/components";
+import { AuthService } from "projects/core/src/auth";
+import { GlobalMessageService, GlobalMessageType } from "projects/core/src/global-message";
+import { WindowRef } from "projects/core/src/window";
```
znikola: Please don't use absolute path imports, but rather import from `@spartacus/...` or use relative imports if you're importing something from the same entry point

integration-libs/cdc/components/login/facade/cdc-login-form-component.service.ts
```diff
@@ -0,0 +1,46 @@
...
+  ) {
+    super(auth, globalMessageService, winRef);
+  }
+
+  protected busy$ = new BehaviorSubject(false);
+
+  login() {
+    this.busy$.next(true);
+    this.cdcJsService.didLoad().subscribe((cdcLoaded) => {
+      if (cdcLoaded) {
```
znikola: I see you are setting the `busy$` to `false` in the `else` part of this `if` condition.
Do you need to do the same in case the `cdcLoaded` is `true`?

integration-libs/cdc/components/login/facade/cdc-login-form-component.service.ts
```diff
@@ -0,0 +1,46 @@
...
+  constructor(
+    protected auth: AuthService,
+    protected globalMessageService: GlobalMessageService,
+    protected winRef: WindowRef,
+    protected cdcJsService: CdcJsService
+  ) {
+    super(auth, globalMessageService, winRef);
+  }
+
+  protected busy$ = new BehaviorSubject(false);
```
znikola: No need to declare this again, as you are overriding the `LoginFormComponentService`

integration-libs/cdc/components/login/facade/cdc-login-form-component.service.ts
```diff
@@ -0,0 +1,46 @@
+import { Injectable } from "@angular/core";
+import { CdcJsService } from "@spartacus/cdc/root";
+import { LoginFormComponentService } from "@spartacus/user/account/components";
+import { AuthService } from "projects/core/src/auth";
+import { GlobalMessageService, GlobalMessageType } from "projects/core/src/global-message";
+import { WindowRef } from "projects/core/src/window";
+import { BehaviorSubject } from "rxjs";
+
+@Injectable({
+  providedIn: 'root',
```
znikola: Please remove the `providedIn: 'root'` part

integration-libs/cdc/components/login/facade/cdc-login-form-component.service.ts
```diff
@@ -0,0 +1,46 @@
...
+    protected cdcJsService: CdcJsService
+  ) {
+    super(auth, globalMessageService, winRef);
+  }
+
+  protected busy$ = new BehaviorSubject(false);
+
+  login() {
+    this.busy$.next(true);
+    this.cdcJsService.didLoad().subscribe((cdcLoaded) => {
```
znikola: Let's cleanup the subscription on the `ngOnDestroy` lifecycle hook.

integration-libs/cdc/components/login/facade/cdc-login-form.module.ts
```diff
@@ -0,0 +1,53 @@
...
+    FormErrorsModule,
+    SpinnerModule,
+    UserAccountModule,
+  ],
+  providers: [
+    provideDefaultConfig(<CmsConfig>{
+      cmsComponents: {
+        ReturningCustomerLoginComponent: {
+          component: LoginFormComponent,
+          guards: [NotAuthGuard],
```
znikola: There's no need to define the guard again.
I believe this line can be removed

integration-libs/cdc/components/login/facade/cdc-login-form.module.ts
```diff
@@ -0,0 +1,53 @@
...
+  UrlModule,
+  WindowRef
+} from '@spartacus/core';
+import { FormErrorsModule, SpinnerModule } from '@spartacus/storefront';
+import { UserAccountModule } from '@spartacus/user';
+import {
+  LoginFormComponent,
+  LoginFormComponentService
+} from '@spartacus/user/account/components';
+import { CdcJsService } from 'integration-libs/cdc/root';
```
znikola: Please fix the import

integration-libs/cdc/components/login/facade/cdc-login-form.module.ts
```diff
@@ -0,0 +1,53 @@
...
+  imports: [
+    CommonModule,
+    FormsModule,
+    ReactiveFormsModule,
+    RouterModule,
+    UrlModule,
+    I18nModule,
+    FormErrorsModule,
+    SpinnerModule,
+    UserAccountModule,
```
znikola: A note for myself: double check when doing wrapper schematics

integration-libs/cdc/components/register/cdc-user-profile.module.ts
```diff
@@ -0,0 +1,21 @@
+import { NgModule } from '@angular/core';
+import { UserProfileModule } from '@spartacus/user';
+import {
+  UserRegisterFacade
+} from '@spartacus/user/profile/root';
+import { CDCRegisterComponentService } from './facade/cdc-register-component.service';
+
+@NgModule({
+  imports: [
+    UserProfileModule,
```
znikola: Note for myself: double check when doing wrapper module

znikola (changes requested):

projects/storefrontapp/src/app/spartacus/features/cdc-feature.module.ts
SherwinVarghese: Hi Nik,

this was for dev system and should be removed..
thanks

feature-libs/user/profile/components/register/register-component.service.ts
SherwinVarghese: agreed

feature-libs/user/profile/components/register/register-component.service.ts
SherwinVarghese: sure Nik

integration-libs/cdc/components/login/facade/cdc-login-form.module.ts
SherwinVarghese: agreed, removed the `guards` field

integration-libs/cdc/components/login/facade/cdc-login-form.module.ts
SherwinVarghese: sure.

feature-libs/user/account/components/login-form/login-form.module.ts
SherwinVarghese: sure Nik, will do so

feature-libs/user/profile/assets/translations/en/user-profile.ts
SherwinVarghese: it will be removed

integration-libs/cdc/components/login/facade/cdc-login-form-component.service.ts
SherwinVarghese: agreed, most of the times, VS Code tries to resolve imports locally.

integration-libs/cdc/components/login/facade/cdc-login-form-component.service.ts
SherwinVarghese: noted, thanks

integration-libs/cdc/components/login/facade/cdc-login-form-component.service.ts
SherwinVarghese: that is correct!

integration-libs/cdc/components/login/facade/cdc-login-form-component.service.ts
SherwinVarghese: noted.
thanks

integration-libs/cdc/components/login/facade/cdc-login-form-component.service.ts
SherwinVarghese: thanks for the suggestion Nik. After the login an action is triggered from the CDC JS service.
Nevertheless, we can set `busy$` to` false` here as well

integration-libs/cdc/components/register/facade/cdc-register-component.service.ts
SherwinVarghese: sure Nik, autocomplete problems 😅

integration-libs/cdc/components/register/facade/cdc-register-component.service.ts
SherwinVarghese: updated the jsDoc

integration-libs/cdc/root/service/cdc-js.service.ts
SherwinVarghese: added migration docs

feature-libs/user/profile/components/register/register.module.ts
```diff
@@ -35,6 +35,12 @@ import { RegisterComponent } from './register.component';
         RegisterCustomerComponent: {
           component: RegisterComponent,
           guards: [NotAuthGuard],
+          providers: [
+            {
+              provide: UserRegisterFacade,
+              useExisting: RegisterComponentService,
+            },
+          ],
```
znikola: Please check how it was done in feature-libs/user/account/components/login-form/login-form.module.ts, and analogical to that use `useClass` and provide `deps` to provide dependencies that match `RegisterService`'s constrcutor

integration-libs/cdc/components/login/facade/cdc-login-form-component.service.ts
znikola: Yeah, it often happens to me as well

integration-libs/cdc/components/login/facade/cdc-login-form-component.service.ts
```diff
@@ -19,28 +18,37 @@ export class CdcLoginFormComponentService extends LoginFormComponentService {
...
+            GlobalMessageType.MSG_TYPE_ERROR
+          );
+          this.busy$.next(false);
+        }
+      })
+    );
+  }
+
+  ngOnDestroy(): void {
+    if (this.subscription) {
```
znikola: Nice that you are unsubscribing.
This is a nitpick, but you don't really need to check if the `this.subscription` exists, as it's always instantiated as the service's property. I'll leave it up to you if want to keep it or not.

feature-libs/user/profile/components/register/register-component.service.ts
SherwinVarghese: updated the jsDoc to `@param user as UserSignUp`

integration-libs/cdc/components/login/facade/cdc-login-form.module.ts
```diff
@@ -0,0 +1,57 @@
...
```
znikola: We can remove the import of `UserAccountModule` as it's being used in the user-account-wrapper module.
Please double-check if your feature is working as expected after doing this.

znikola (changes requested):

feature-libs/user/profile/components/register/register.module.ts
SherwinVarghese: sure Nik, will do

integration-libs/cdc/components/login/facade/cdc-login-form-component.service.ts
SherwinVarghese: agreed Nik, it is one line less of code 😁

feature-libs/user/profile/components/register/register.component.ts
```diff
@@ -68,7 +65,7 @@ export class RegisterComponent implements OnInit, OnDestroy {
   );
 
   constructor(
-    protected userRegister: UserRegisterFacade,
+    protected registerComponentService: RegisterComponentService,
```
Platonn, line 68: Let's move the new constructor parameter as the last one.
It's because the [schematics for constructor migration](https://github.com/SAP/spartacus/pull/15785/files#diff-376763235e53d104df837029fdaf3789a7a339ed73003a4a5076c1cd944d78abR23-R27) will add the new paramerer as the last one in customer's codebase. 

That's how the `addParams` migration works - it adds a param at the last place.
image

feature-libs/user/profile/components/register/register.component.ts
```diff
@@ -68,7 +65,7 @@ export class RegisterComponent implements OnInit, OnDestroy {
...
     protected globalMessageService: GlobalMessageService,
     protected fb: FormBuilder,
     protected router: RoutingService,
```
Platonn:
```suggestion
    protected globalMessageService: GlobalMessageService,
    protected fb: FormBuilder,
    protected router: RoutingService,
    protected registerComponentService: RegisterComponentService,
```

Let's move the new constructor parameter to the last position in constructor parameres
It's because the [schematics for constructor migration](https://github.com/SAP/spartacus/pull/15785/files#diff-376763235e53d104df837029fdaf3789a7a339ed73003a4a5076c1cd944d78abR23-R27) will add the new paramerer as the last one in customer's codebase. 

That's how the `addParams` migration works - it adds a param at the last place.
image

integration-libs/cdc/components/login/facade/index.ts
```diff
@@ -0,0 +1 @@
+export * from './cdc-login-form-component.service';
```
Platonn, line 1:
```suggestion
export * from './cdc-login-form-component.service';
export * from './cdc-login-form.module';
```

let's expose in public API also the module

Platonn: Let's export also the module here:

```suggestion
export * from './cdc-login-form-component.service';
export * from './cdc-login-form.module';
```

integration-libs/cdc/components/login/index.ts
```diff
@@ -0,0 +1,3 @@
+export * from './facade/cdc-login-form.module';
+export * from './facade/index';
```
Platonn: The files CdcLoginFormModule and CdcLoginFormComponent are not facades and I don't think they should be in a folder named `/facade` . I guess it's a leftover name after a few refactors.

Let's rename the folder here from `/facade` for `/login-form`

integration-libs/cdc/components/register/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from './facade/index';
```
Platonn: Instead of having folder `/facade` here, let's create 2 folders:
/register
/forgot-password

Then let's move there relevant modules and components to those folders. And let's export them in their local `index.ts` files. 

Then in this file, let's just re-export those index.ts files:
```suggestion
export * from './forgot-password/index';
export * from './register/index';
```

integration-libs/cdc/core/auth/services/user-authentication/cdc-user-authentication-token.service.ts
```diff
@@ -32,11 +32,11 @@ export class CdcUserAuthenticationTokenService {
...
-      .set('UID', UID)
-      .set('UIDSignature', UIDSignature)
-      .set('timeStamp', signatureTimestamp)
-      .set('idToken', idToken)
-      .set('baseSite', baseSite);
+      .set('UID', encodeURIComponent(UID))
+      .set('UIDSignature', encodeURIComponent(UIDSignature))
+      .set('signatureTimestamp', encodeURIComponent(signatureTimestamp))
+      .set('id_token', encodeURIComponent(idToken))
+      .set('baseSite', encodeURIComponent(baseSite));
```
Platonn, line 45: Is this change really related to this PR? Or it's unrelated bugfix? 

If it's a bugfix, should we backport it separtely also to previous versions of Spartacus (the latest 3.x and 4.x)?

I can see we're renaming here 2 parameres: `timeStamp` and `idToken`. What is the reason of renaming them? Do you think it can be perceived a breaking change?

integration-libs/cdc/components/login/facade/cdc-login-form-component.service.spec.ts
```diff
@@ -0,0 +1,125 @@
...
+
+class MockLoginFormComponentService
+  implements Partial<LoginFormComponentService>
+{
+  login = createSpy();
+}
+
+describe('CdcLoginComponentService', () => {
+  let cdcLoginService: CdcLoginFormComponentService;
+  // let loginFormService: LoginFormComponentService;
```
Platonn: Let's remove the commented out code 
```suggestion
```

integration-libs/cdc/components/login/facade/cdc-login-form.module.ts
SherwinVarghese: sure Nik.

feature-libs/user/profile/components/register/register.component.ts
SherwinVarghese: sure Kris, will update the ordering of `RegisterComponentService`

integration-libs/cdc/core/auth/services/user-authentication/cdc-user-authentication-token.service.ts
SherwinVarghese: Hi Kris,

These changes were introduced as a part of the customer issue fix.
The backend API was changed to accept the parameter - `timeStamp` as `signatureTimestamp` and `idToken` as `id_token` while refactoring the backend code.

integration-libs/cdc/components/login/facade/cdc-login-form-component.service.spec.ts
SherwinVarghese: agreed, removed the commented code.

integration-libs/cdc/components/login/index.ts
SherwinVarghese: okay Kris

integration-libs/cdc/components/register/index.ts
SherwinVarghese: okay Kris, will do so

integration-libs/cdc/core/auth/services/user-authentication/cdc-user-authentication-token.service.ts
Platonn: This change seems to be a bugfix unrelated to the subject of this PR.
Therefore, for clarity of commits (and the possible ease of backporting of the bugfix), what do you think about moving this change to a separate bugfix PR?

Btw. Do I understand correctly, that this change is a breaking change (as it changes the contract of properties sent to the backend)? If yes, then formally we should release it only in the major version (5.0). So then there will be no backporting to 3.x and 4.x, I guess.

Only customers having the latest backend version (containing the backend fix with the changed contract) will benefit from this frontend's change, but for others (using older backends), this frontend's change will completely break the CDC integration, right? If yes, I think we should document it somewhere in 5.0 migration docs - what is needed in backend to make CDC work.

Moreover, independently, for customers that will still use 3.x or 4.x for some time, I believe we will need some documentation how they can workaround the bug (by writing some code both on the frontend and backend manually). What do you think about it? (But anyway, I believe the documentation of the workaround for users of 3.x and 4.x is beyond the scope of this PR)

integration-libs/cdc/root/service/cdc-js.service.ts
```diff
@@ -162,6 +167,146 @@ export class CdcJsService implements OnDestroy {
...
+      });
+    }
+  }
+
+  /**
+   * Trigger CDC User registration using CDC APIs.
+   *
+   * @param response
+   */
+  onInitRegistrationHandler(user: UserSignUp, response: any) {
```
Platonn: nitpick: I think we can make this method `protected`, as it's not meant to be called from outside. Same for other helper/handler methods in this PR.

projects/storefrontapp/src/app/spartacus/features/cdc-feature.module.ts
Platonn: Pasting here the conclusion from the offline chat, for future reference:

Lately, we realized that names of entry points `@spartacus/cdc/user-profile` and `@spartacus/cdc/user-account` could better suit our needs than entry points named `@spartacus/cdc/components/register` and `@spartacus/cdc/components/login` . That’s becasue now we need to put somewhere the Forgot Password feature - which neither fits well in `/register` nor in `/login`, but rather in `/user-profile`. Moreover in the future we’re going to add the Update Profile feature, which also fits well to `/user-profile`.

I’m aware some time ago we changed our mind from naming entry points and modules `user-profile` and `user-account` to `register` and `login` . And I’m sorry for back-and-forth suggestions in this topic.
The addition of Forgot Password and future Update Profile helped me realize that it makes sense to have `cdc/user-profile` and `cdc/user-account` entry points (and modules).

Platonn:
✅ **Lazy loaded chunks:** In the fresh app, the comparison of JS chunks in dev mode looks as expected (a little more bytes in `main` js, more in the lazy loaded chunks wrappers for `user-account` and `user-profile`, and separate additinoal chunk for `cdc`).
See: OOTB Spartacus vs OOTB Spartacus with CDC extension 
image

✅ **QA result:** I can use login, register and forgot-password features in our repo app, and in the fresh app, successfully

Platonn:
✅ QA result:
I can use login, register and forget-password features in our repo app, and in the fresh app, successfully

Platonn (dismissed):

integration-libs/cdc/core/auth/services/user-authentication/cdc-user-authentication-token.service.ts
SherwinVarghese: Hi @Platonn,

Thanks for bringing this dicussion. This fix was introduced as a part of the customer ticket. The customer is on Spartacus 4.3. so these changes are backported to 4.x version currently.

We can think of backporting the fixes for 3.x if needed.

thanks.

integration-libs/cdc/root/service/cdc-js.service.ts
SherwinVarghese: sure Kris, will do

integration-libs/cdc/root/service/cdc-js.service.ts
```diff
@@ -170,81 +170,112 @@ export class CdcJsService implements OnDestroy {
...
+          finalizeRegistration: true,
+          callback: (response: any) => {
+            this.zone.run(() => {
+              if (response && response.status === 'FAIL') {
+                this.handleRegisterError(response);
+                isRegistered.next(false);
+              } else if (response && response.status === 'OK') {
+                isRegistered.next(true);
+              }
+              isRegistered.complete();
```
Platonn: [MAJOR]
The response from the CDC server is rich, but here we return only true/false. In the future we might want to return more information about the response, Eg. the error message.
What do you think about returning at least an object, that contains (at least) the `status` property? Thanks to that, in the future we can add more properties to the response, whenever we want, without a breaking change. TBH I wouldn't mind if we passed also the error message and other info that you think might be a good result of this function.

Similar remark applies also for: Login and Forgot Password.

integration-libs/cdc/root/service/cdc-js.service.ts
```diff
@@ -170,81 +170,112 @@ export class CdcJsService implements OnDestroy {
...
+          'gigya'
+        ]?.accounts?.login({
+          loginID: email,
+          password: password,
+          callback: (response: any) => {
+            this.zone.run(() => {
+              if (response && response.status === 'FAIL') {
+                this.handleLoginError(response);
+                isLoggedIn.next(false);
+              } else if (response && response.status === 'OK') {
```
Platonn: [QUESTION]
Can the response `status` be different than 'FAIL' or 'OK'? If yes, then our method will just complete the stream without emitting any value or error.

Same question applies also for: Register and Forgot Password

integration-libs/cdc/root/service/cdc-js.service.ts
```diff
@@ -170,81 +170,112 @@ export class CdcJsService implements OnDestroy {
...
-        'gigya'
-      ]?.accounts?.initRegistration({
-        callback: (response: any) => {
-          this.zone.run(() => this.onInitRegistrationHandler(user, response));
-        },
-      });
-    }
+  registerUserWithoutScreenSet(user: UserSignUp): Observable<Boolean> {
+    return new Observable<Boolean>((initRegistration) => {
+      if (user.uid && user.password) {
```
Platonn: [MINOR]
For this function to be bullet-proof, it would be good to emit an error, when this condition is negative. Otherwise the subscriber of this method will never get any value and his subscription will be pending forever (being a memory leak)

integration-libs/cdc/root/service/cdc-js.service.spec.ts
```diff
@@ -318,47 +318,51 @@ describe('CdcJsService', () => {
...
+      service
+        .registerUserWithoutScreenSet({
+          uid: 'uid',
+          password: 'password',
+        })
+        .subscribe(() => {
+          expect(
+            winRef.nativeWindow['gigya'].accounts.initRegistration
+          ).toHaveBeenCalled();
+        });
```
Platonn: [MINOR]
It would be good call [`done()` function of jasmine](https://jasmine.github.io/tutorials/async#callbacks), after the `expect()`, inside the subscription's callback.
If we didn't put `done()` here, when the method doesn't emit anything, 
the test would pass. Which would false positive result. 

Similar remark applies also to other tests in this PR, which have `expect()` inside the subscription's callback.

integration-libs/cdc/user-profile/forgot-password/cdc-forgot-password-component.service.ts
```diff
@@ -59,8 +61,8 @@ export class CDCForgotPasswordComponentService
             },
             GlobalMessageType.MSG_TYPE_ERROR
           );
-          this.busy$.next(false);
         }
+        this.busy$.next(false);
```
Platonn: [MINOR]
Now we set `busy$` to false immediately when the function is called, without waiting for `this.cdcJsService.resetPasswordWithoutScreenSet` to finish.

[QUESTION]
Maybe(?) we could reuse the methods of the super class `ForgotPasswordComponentService`: `onSuccess()` and `onError()` https://github.com/SAP/spartacus/blob/69d81dd41713d2547a991f0a5c87a4030295610b/feature-libs/user/profile/components/forgot-password/forgot-password-component.service.ts#L65-L77? I'm not insisting on it, just asking. For example:
- in the subscription to `this.cdcJsService.resetPasswordWithoutScreenSet`, we could potentially call `onSuccess()` or `onError()`, depending on the result of the operation
- if `cdcLoaded` is `false`, we could also call `onError()`

integration-libs/cdc/user-profile/register/cdc-register-component.service.ts
```diff
@@ -22,10 +21,16 @@ export class CDCRegisterComponentService extends RegisterComponentService {
         new Observable<User>((userRegistered) => {
           // Registering user through CDC Gigya SDK
           if (user.firstName && user.lastName && user.uid && user.password) {
-            this.cdcJSService.registerUserWithoutScreenSet(user);
+            this.cdcJSService
+              .registerUserWithoutScreenSet(user)
+              .subscribe((isRegistered) => {
+                if (isRegistered) {
+                  userRegistered.complete();
```
Platonn: [MINOR]
It would be good to emit something (`userRegistered.next(something)`). Otherwise if someone subscribes to the .register() method, their subscription callback will won't be executed, e.g.:
```
// '🎉' won't be console logged, because the observable never emits something; it only completes
CdcUserRegisterComponentService
   .register().subscribe(()=>console.log('🎉'))
```

For more on subscription callbacks, see example: https://stackblitz.com/edit/rxjs-observable-only-complete?file=index.ts

integration-libs/cdc/user-profile/register/cdc-register-component.service.ts
```diff
@@ -0,0 +1,73 @@
...
+          // CDC Gigya SDK not loaded, show error
+          this.globalMessageService.add(
+            {
+              key: 'errorHandlers.scriptFailedToLoad',
+            },
+            GlobalMessageType.MSG_TYPE_ERROR
+          );
+          return new Observable<User>((userNotRegistered) => {
+            userNotRegistered.complete();
+          });
```
Platonn: In case when we couldnt register an user, we should throw an error.
So instead of just closing the stream, lets throw an error.
```suggestion
           throw new Error(....)
```

Platonn: handled in https://github.com/SAP/spartacus/pull/16160

integration-libs/cdc/user-profile/register/cdc-register-component.service.ts
Platonn: simplified and fixed in https://github.com/SAP/spartacus/pull/16160

integration-libs/cdc/user-profile/forgot-password/cdc-forgot-password-component.service.ts
Platonn: - regarding [MINOR] -  issue fixed in https://github.com/SAP/spartacus/pull/16160
- regarding [QUESTION] - we're not using super.onSuccess/onError, because we don't want to trigger the global message here (as it's already triggered from the CDC JS Service)

integration-libs/cdc/root/service/cdc-js.service.spec.ts
Platonn: Will be handled in https://github.com/SAP/spartacus/pull/16169

integration-libs/cdc/root/service/cdc-js.service.ts
Platonn: fixed in https://github.com/SAP/spartacus/pull/16160

integration-libs/cdc/root/service/cdc-js.service.ts
Platonn: fixed in https://github.com/SAP/spartacus/pull/16160

integration-libs/cdc/root/service/cdc-js.service.spec.ts
```diff
@@ -343,7 +343,7 @@ describe('CdcJsService', () => {
   });
 
   describe('onInitRegistrationHandler', () => {
-    it('should register the user', (done) => {
+    it('should register the user', () => {
```
znikola: I suggest keeping the `done`, but calling it from the `complete` channel.
Basically, something like this:

```ts
service['onInitRegistrationHandler'](
        {
          uid: 'uid',
          password: 'password',
          firstName: 'fname',
          lastName: 'lname',
        },
        { regToken: 'TOKEN' }
      ).subscribe({
        complete: () => {
          expect(
            winRef.nativeWindow['gigya'].accounts.register
          ).toHaveBeenCalledWith({
            email: 'uid',
            password: 'password',
            profile: {
              firstName: 'fname',
              lastName: 'lname',
            },
            regToken: 'TOKEN',
            finalizeRegistration: true,
            callback: jasmine.any(Function),
          });
          done();
        },
      });
```

znikola (changes requested):

znikola (dismissed):

znikola (dismissed):

znikola (approved):

PR #15786, opened 2022-05-24
feat: better extensibility for configurator attribute components

ChristophHi:


feature-libs/product-configurator/rulebased/components/composition/configurator-attribute-composition.directive.ts
```diff
@@ -0,0 +1,55 @@
...
+} from '@angular/core';
+import { ConfiguratorAttributeCompositionConfig } from './configurator-attribute-composition.config';
+import { ConfiguratorAttributeCompositionContext } from './configurator-attribute-composition.model';
+
+@Directive({
+  selector: '[cxConfiguratorAttributeComponent]',
+})
+export class ConfiguratorAttributeCompositionDirective implements OnInit {
+  @Input('cxConfiguratorAttributeComponent')
+  context: ConfiguratorAttributeCompositionContext;
```
Platonn: Just for this directive to be bullet-proof, let's discuss what should happen when the value of the context `@Input()` changes over time?
Shouldn't we reset the `vcr` and do the `renderComponent` again?

feature-libs/product-configurator/rulebased/components/form/configurator-form.component.ts
```diff
@@ -100,4 +100,8 @@ export class ConfiguratorFormComponent implements OnInit {
   createGroupId(groupId?: string): string | undefined {
     return this.configUtils.createGroupId(groupId);
   }
+
+  getComponentKey(attribute: Configurator.Attribute): string {
+    return 'AttributeType_' + attribute.uiType;
```
Platonn: Nitpick: let's move string `'AttributeType_'` to a separate `protected readonly` property

feature-libs/product-configurator/rulebased/components/attribute/types/drop-down/configurator-attribute-drop-down.module.ts
```diff
@@ -19,6 +20,15 @@ import { ConfiguratorPriceModule } from '../../../price/configurator-price.modul
     ReactiveFormsModule,
     ConfiguratorPriceModule,
   ],
+  providers: [
+    provideDefaultConfig(<ConfiguratorAttributeCompositionConfig>{
+      productConfigurator: {
+        attributeComponentAssignment: {
+          AttributeType_dropdown: ConfiguratorAttributeDropDownComponent,
```
Platonn, line 39: Nitpick: for type safety, what do you think about taking string `AttributeType_dropdown` (and other keys for similar configs) from some enum?

feature-libs/product-configurator/rulebased/components/attribute/footer/configurator-attribute-footer.module.ts
```diff
@@ -13,6 +14,15 @@ import { ConfiguratorAttributeFooterComponent } from './configurator-attribute-f
     I18nModule,
     IconModule,
   ],
+  providers: [
+    provideDefaultConfig(<ConfiguratorAttributeCompositionConfig>{
+      productConfigurator: {
+        attributeComponentAssignment: {
```
Platonn: Nitpick: for ergonomics of using this config, did you consider shorter config key instead of `attributeComponentAssignment`? maybe simply `attributeComponent`? or just `attribute`?

ChristophHi: Yes, I see. But I would rather go for 'assignment' (because we are assigning components to key (representing attribute types) here, would you agree?

feature-libs/product-configurator/rulebased/components/composition/configurator-attribute-composition.directive.ts
ChristophHi: @Platonn TBH I don't understand when this would happen, because on every new submission of a runtime configuration, e.g. on changing an attribute value, new instances of the directive are created.
(Is that somewhat surprising to you?)

feature-libs/product-configurator/rulebased/components/attribute/types/drop-down/configurator-attribute-drop-down.module.ts
ChristophHi: I am not sure, because I want projects to be able to define their custom attribute UI types in the backend, and, in order to support them in SPA, enable them to just add a configuration + the respective component.
So we must not restrict the keys for our assignment.

Larisa-Staroverova (approved):

PR #15807, opened 2022-06-01
refactor: Enable strict mode for core lib 

WeizhengSap:


projects/core/src/auth/client-auth/services/client-error-handling.service.ts
```diff
@@ -26,6 +27,7 @@ export class ClientErrorHandlingService {
     next: HttpHandler
   ): Observable<any> {
     return this.clientTokenService.refreshClientToken().pipe(
+      filter(isNotUndefined),
```
plabadie: Would it be best to make clientTokenService.refreshClientToken not emit undefined?

Maybe it's all the same in the end.  It's just that I noticed that you added `| undefined` to the return value of `refreshClientToken` and in practice, `undefined` is not a value the caller considers.

projects/core/src/auth/client-auth/services/client-token.service.ts
```diff
@@ -21,7 +21,7 @@ export class ClientTokenService {
    * Returns a client token. The client token from the store is returned if there is one.
    * Otherwise a new token is fetched from the backend and saved in the store.
    */
-  getClientToken(): Observable<ClientToken> {
+  getClientToken(): Observable<ClientToken | undefined> {
```
plabadie, line 25: If callers need to filter out undefined, maybe we don't return undefined.  What do you think?

projects/core/src/auth/client-auth/services/client-token.service.ts
```diff
@@ -43,7 +43,7 @@ export class ClientTokenService {
    * Fetches a clientToken from the backend and saves it in the store where getClientToken can use it.
    * The new clientToken is returned.
    */
-  refreshClientToken(): Observable<ClientToken> {
+  refreshClientToken(): Observable<ClientToken | undefined> {
```
plabadie: If callers tend to filter out undefined, maybe we don't return undefined.  What do you think?

projects/core/src/cms/services/cms-structure-config.service.ts
```diff
@@ -137,15 +139,19 @@ export abstract class CmsStructureConfigService {
     }
 
     for (const position of Object.keys(slots)) {
-      if (!Object.keys(pageStructure.page.slots).includes(position)) {
+      if (
+        pageStructure.page &&
+        pageStructure.page.slots &&
+        !Object.keys(pageStructure.page.slots).includes(position)
+      ) {
```
plabadie, line 145: Question: can  
```
pageStructure.page &&
pageStructure.page.slots &&
```

be replaced by 
```
pageStructure.page?.slots &&
```

I think they are equivalent. Are they?

projects/core/src/cms/store/effects/navigation-entry-item.effect.ts
```diff
@@ -52,10 +52,10 @@ export class NavigationEntryItemEffects {
                 )
             )
           );
-        } else if (data.ids.pageIds.length > 0) {
+          //} else if (data.ids.pageIds.length > 0) {
           // TODO: future work
           // dispatch action to load cms page one by one
-        } else if (data.ids.mediaIds.length > 0) {
+          //} else if (data.ids.mediaIds.length > 0) {
```
plabadie, line 58: Perhaps the dead code should be removed.

projects/core/src/occ/adapters/cms/legacy-occ-cms-component.adapter.ts
```diff
@@ -46,6 +46,7 @@ export class LegacyOccCmsComponentAdapter extends OccCmsComponentAdapter {
       )
       .pipe(
         pluck('component'),
+        map((compts) => compts ?? []),
```
plabadie: `component` instead of `compts` would improve readability since it's what has just been plucked.  
It's a suggestion, it's your call in the end.

projects/core/src/occ/adapters/cms/occ-cms-component.adapter.ts
```diff
@@ -57,6 +57,7 @@ export class OccCmsComponentAdapter implements CmsComponentAdapter {
       )
       .pipe(
         pluck('component'),
+        map((compts) => compts ?? []),
```
plabadie: `component` instead of `compts` would improve readability since it's what has just been plucked.  
It's a suggestion, but it's your call.

projects/core/src/occ/adapters/product/converters/occ-product-search-page-normalizer.ts
```diff
@@ -86,7 +87,9 @@ export class OccProductSearchPageNormalizer
       target.facets = target.facets.map((facetSource: Facet) => {
         const { topValues, ...facetTarget } = facetSource;
         facetTarget.topValueCount =
-          topValues?.length > 0 ? topValues.length : this.DEFAULT_TOP_VALUES;
+          topValues && topValues.length > 0
```
plabadie, line 90: Is there a difference for strict mode between

```
topValues && topValues.length > 0
```

and 

```
topValues?.length > 0
```

projects/core/src/occ/adapters/site-context/site-context.interceptor.ts
```diff
@@ -50,7 +49,11 @@ export class SiteContextInterceptor implements HttpInterceptor {
     request: HttpRequest<any>,
     next: HttpHandler
   ): Observable<HttpEvent<any>> {
-    if (request.url.includes(this.occEndpoints.getBaseUrl())) {
+    if (
+      this.activeCurr &&
+      this.activeLang &&
+      request.url.includes(this.occEndpoints.getBaseUrl())
+    ) {
```
plabadie: for activeCurr or activeLang, there should perhaps bean error message to the console if they are flasy since we omit to set them on the request.  At least there would be some feedback to devs that might get an issue with this.

projects/core/src/occ/adapters/site-context/site-context.interceptor.ts
```diff
@@ -50,7 +49,11 @@ export class SiteContextInterceptor implements HttpInterceptor {
     request: HttpRequest<any>,
     next: HttpHandler
   ): Observable<HttpEvent<any>> {
-    if (request.url.includes(this.occEndpoints.getBaseUrl())) {
+    if (
+      this.activeCurr &&
+      this.activeLang &&
+      request.url.includes(this.occEndpoints.getBaseUrl())
```
plabadie: There is a slight change in behaviour.   If `activeCurr` is flasy, `activeLang` is not going to be used and vice versia.

projects/core/src/occ/services/occ-endpoints.service.ts
```diff
@@ -110,7 +110,7 @@ export class OccEndpointsService {
       const { urlParams, queryParams } = attributes;
 
       if (urlParams) {
-        url = StringTemplate.resolve(url, attributes.urlParams, true);
+        url = StringTemplate.resolve(url, urlParams, true);
```
plabadie, line 113: I guess you don't need to destructure urlParams and can use `attributes.urlParams` everywhere instead.

projects/core/src/product/event/product-event.builder.ts
```diff
@@ -49,7 +51,10 @@ export class ProductEventBuilder {
             name: toggled.facetName,
             valueCode: toggled.facetValueCode,
             valueName: toggled.facetValueName,
-            selected: curr.breadcrumbs.length > prev.breadcrumbs.length,
+            selected:
+              curr.breadcrumbs &&
```
plabadie, line 55: if `curr.breadcrumbs` is undefined, will `selected` be assigned `false` or `undefined` ?

`selected` is type boolean.

projects/core/src/product/event/product.events.ts
```diff
@@ -11,21 +11,21 @@ export class FacetChangedEvent extends CxEvent {
   /**
    * Facet code
    */
-  code: string;
+  code?: string;
```
plabadie, line 14: Would it have been possible to change the code that uses this interface instead of making all fields optional?

projects/core/src/auth/client-auth/services/client-token.service.ts
WeizhengSap: the client token is read from ngrx store. The initial state value of the token is `undefined`. So, it is possible that the function returns `undefined`. This function is also used in `client-token.interceptor.ts`, which does not filter out the `undefined` value.

projects/core/src/auth/client-auth/services/client-token.service.ts
WeizhengSap: is it possible that the `refresh token` failed, then the token is `undefined`? I can move the filter to this function.

projects/core/src/cms/store/effects/navigation-entry-item.effect.ts
WeizhengSap: navigation entry can also point to medias and pages. Now, we only handle links. So, this dead code is a reminder for future work.

projects/core/src/occ/adapters/product/converters/occ-product-search-page-normalizer.ts
WeizhengSap: `topValues?.length > 0` has error, because `topValues` can be undefined, which does not has `length`

projects/core/src/occ/adapters/site-context/site-context.interceptor.ts
WeizhengSap: I change it to this: 
```
setParams: {
          lang: this.activeLang ?? '',
          curr: this.activeCurr ?? '',
        },
 ```

projects/core/src/occ/adapters/site-context/site-context.interceptor.ts
WeizhengSap: the request will be `curr=&lang=`. The backend will omit these 2 params.

projects/core/src/occ/services/occ-endpoints.service.ts
WeizhengSap: yes, but destructure it make the codes more clean.

projects/core/src/product/event/product-event.builder.ts
WeizhengSap: `false`. Actually, if `breadcrumbs` is `undefined`. `toggled` is `undefined`. So, the event will not be created.

projects/core/src/product/event/product.events.ts
WeizhengSap: this code is actually the `facetCode` in `Breadcrumb`. Our model interfaces have all fields as optional, even they should not be.

projects/core/src/auth/client-auth/services/client-token.service.ts
plabadie: Fair points.

projects/core/src/product/event/product-event.builder.ts
plabadie: For the kind of code that initialize `selected`, though, there might be something do improve there.  I have a hard time wrapping my head around the result of the expression for some reason, but I tried it in this snipped and it seems that the result of the whole thing can be undefined unless what is in each section is cast into a boolean, see: https://stackblitz.com/edit/typescript-zsm5ev?file=index.ts

I don't remember from  the top of my head if we have other cases like this in the PR.  Perhaps we double check,

projects/core/src/product/event/product.events.ts
plabadie: So you're saying it's fine we make such fields optional?

projects/core/src/auth/client-auth/services/client-token.service.ts
plabadie: This was for discussion more than a recommendation.  I say we do what has the least impact on the previous behaviour.  The token refresh logic is not trivial to test.

If it is what you had done to begin with, then I'm sorry I brought this up and perhaps we should leave it like you had done it in the first place.

WeizhengSap: I already changed it

projects/core/src/product/event/product.events.ts
WeizhengSap: `facetCode` is optional in `Breadcrumb`. This `code` is `facetCode`. So, it should be the same type as `facadeCode`. If we don't want it optional, we can set it to `''`.

projects/core/src/product/event/product-event.builder.ts
WeizhengSap: I am wrong. The value should be `undefined`. The `selected` is `boolean | undefined`. So, it should be fine to assign to `undefined`. That's why it can be compiled.

plabadie (approved):

PR #15816, opened 2022-06-06
fix: Move cart core code out of the 'main' chunk

bgambocjaviniar:
closes https://jira.tools.sap/browse/CXSPA-542


feature-libs/cart/base/root/facade/multi-cart.facade.ts
```diff
@@ -209,6 +210,16 @@ export abstract class MultiCartFacade {
    */
   abstract assignEmail(cartId: string, userId: string, email: string): void;
 
+  /**
+   * Remove cart
+   *
+   * Used for cleaning cart in local state, when we get information that it no longer exists in the backend.
```
znikola:
```suggestion
   Removes the cart from the state.
```

feature-libs/cart/base/root/facade/multi-cart.facade.ts
```diff
@@ -209,6 +210,16 @@ export abstract class MultiCartFacade {
...
+   * For removing particular cart in both places use DeleteCart actions.
```
znikola: I'm confused by this comment. Which both places? Back-end and state?

znikola: If so, maybe we can rephrase it somehow. ```To remove a cart on the back-end, please use `DeleteCart` actions.```
Moreover, why actions? We don't have a facade in front of those ngrx actions?

feature-libs/checkout/base/root/events/checkout-legacy-store-event.listener.spec.ts
```diff
@@ -16,14 +14,13 @@ import createSpy = jasmine.createSpy;
...
 const mockUserId = 'testUserId';
-const mockCartId = 'testCartId';
 
 class MockEventService implements Partial<EventService> {
   get = createSpy().and.returnValue(mockEventStream$.asObservable());
   dispatch = createSpy();
 }
 
-describe(`CheckoutPaymentEventListener`, () => {
+describe(`CheckoutLegacyStoreEventListener`, () => {
```
znikola, line 23: 🙏🏻

feature-libs/cart/base/core/facade/multi-cart.service.ts
```diff
@@ -334,6 +334,18 @@ export class MultiCartService implements MultiCartFacade {
...
   }
 
+  /**
+   * Remove Cart
+   *
+   * Used for cleaning cart in local state, when we get information that it no longer exists in the backend.
+   * For removing particular cart in both places use DeleteCart actions.
+   *
+   * @param cartId
+   */
```
znikola: There's already a jsdoc in the abstract class, which will be shown in an IDE.
Therefore, we don't need it here.

znikola (changes requested):

feature-libs/cart/base/root/facade/multi-cart.facade.ts
znikola: Also, is there a reason for having both remove and delete actions? Why we don't remove the cart with one action from both state and back-end? @WeizhengSap

feature-libs/cart/base/core/facade/multi-cart.service.ts
bgambocjaviniar: I can remove it, but then it will look weird as other exposed functions in this service has them as well

do you want me to remove all the rest too instead?

bgambocjaviniar: removed as discussed~

feature-libs/cart/base/root/facade/multi-cart.facade.ts
znikola: Wei replied on Slack: https://sap-cx.slack.com/archives/C02L8BUATM5/p1654801507446859

znikola (approved):

PR #15817, opened 2022-06-06
feat: Merchant can customize Product Carousels with inner components ex. Add To Cart 

justinlee01:
Closes [CXSPA-191](https://jira.tools.sap/browse/CXSPA-191)

Reusing item template code in product-carousel.component.html into its own component (ProductCarouselItemComponent)
Creating a stylesheet for the above component
The above component accepts inner components. to be used for Add to Cart or Configure components


projects/storefrontapp-e2e-cypress/cypress/integration/regression/product-carousel/product-carousel.non-core-e2e-spec.ts
```diff
@@ -0,0 +1,39 @@
+import * as cart from '../../../helpers/cart';
```
plabadie: What I meant by non core is to rename it back to what it was before, which I presume was  `product-carousel.e2e-spec.ts`

cmanciero (approved):

plabadie (approved):

PR #15826, opened 2022-06-09
Generate an error when invalid cost center is selected

ArkadiySukharenko:
https://jira.tools.sap/browse/CXSPA-411

Generates an error message when invalid cost center is selected during cart checkout in asm session.


justinlee01 (commented):
An interceptor seems like an odd way to solve this, though I see NgRx befuddles the solution a little. Could we use an Effect instead to listen for the `setCostCenter` failure? Is Patrick aware of this solution?

feature-libs/cart/base/core/http-interceptors/handlers/bad-cost-center-request.handler.ts
```diff
@@ -0,0 +1,55 @@
...
+  protected getErrors(response: HttpErrorResponse): ErrorModel[] {
+    return (response.error?.errors || []).filter(
+      (error: any) => error.type !== 'JaloObjectNoLongerValidError'
+    );
+  }
+
+  private isCostCenterRequest(errorResponse: HttpErrorResponse): boolean {
+    if (errorResponse?.url) {
+      const url = new URL(errorResponse.url);
+      return url.pathname.endsWith('costcenter') && new URLSearchParams(url.search).has('costCenterId');
```
plabadie: The url is configurable.  I wonder if this is fine or if we should find another way to narrow down to which response the error is applicable.

feature-libs/cart/base/core/http-interceptors/handlers/bad-cost-center-request.handler.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+
+
+class MockGlobalMessageService {
+  add() {}
+  remove() {}
+}
+
+describe('BadCostCenterRequestHandler', () => {
+  let service: BadCostCenterRequestHandler;
+  let globalMessageService: GlobalMessageService;
```
plabadie: naming the `service` var to something more self explanatory would make the code more readable.  The new name could be as simple as `handler`.

feature-libs/cart/base/core/http-interceptors/handlers/bad-cost-center-request.handler.ts
```diff
@@ -0,0 +1,55 @@
...
+      super.hasMatch(errorResponse) &&
+      this.getErrors(errorResponse).some(isEntityValidationError) &&
+      this.isCostCenterRequest(errorResponse)
+    );
+  }
+
+  handleError(_request: HttpRequest<any>, response: HttpErrorResponse): void {
+    this.getErrors(response)
+      .filter((e) => isEntityValidationError(e))
+      .forEach(() => {
```
plabadie: Do we really want to display multiple occurences of the same message for the same backend call?

feature-libs/cart/base/core/utils/utils.ts
```diff
@@ -59,6 +59,10 @@ export function isCartError(error: ErrorModel): boolean {
   );
 }
 
+export function isEntityValidationError(error: ErrorModel): boolean {
```
plabadie: `EntityValidationError` doesn't sound cart specific.

Wherever we host this util function it will also need corresponding unit tests.

feature-libs/cart/base/core/http-interceptors/handlers/bad-cost-center-request.handler.ts
```diff
@@ -0,0 +1,55 @@
...
+  GlobalMessageType,
+  HttpErrorHandler,
+  HttpResponseStatus,
+  Priority,
+} from '@spartacus/core';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class BadCostCenterRequestHandler extends HttpErrorHandler {
```
plabadie: It seems that this error handler should be moved to the b2b checkout feature.  The endpoint to set the cost center on the cart, which produces the error we want to handle is part of the b2b checkout feature and is called within there.

projects/assets/src/translations/en/common.ts
```diff
@@ -142,6 +142,7 @@ export const common = {
     voucherExceeded:
       'This coupon has exceeded the number of times it can be used',
     unknownIdentifier: 'Item not found.',
+    invalidCostCenter: 'Invalid cost center.',
```
plabadie: Should probably move to the same feature as the handler.

feature-libs/cart/base/core/http-interceptors/handlers/bad-cost-center-request.handler.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+  it('should match cost center error', () => {
+    spyOn(HttpErrorHandler.prototype, 'hasMatch').and.returnValue(true);
+    expect(service.hasMatch(MockCostCenterErrorResponse)).toBe(true);
+  });
+
+  it('should not have a match when super.hasMatch() is false', () => {
+    spyOn(HttpErrorHandler.prototype, 'hasMatch').and.returnValue(false);
+    expect(service.hasMatch(MockCostCenterErrorResponse)).toBe(false);
+  });
+
```
plabadie: We could also add covergate for when the request is not from the set cost center endpoint.

feature-libs/cart/base/core/http-interceptors/handlers/bad-cost-center-request.handler.ts
ArkadiySukharenko: There's no other way I can think of to identify the response as a cost center request. The only response we get is
`{
   "errors" : [ {
      "message" : "The application has encountered an error",
      "type" : "EntityValidationError"
   } ]
}`

justinlee01: This may be quite difficult to do, but you could use the coded value in `defaultOccCostCentersConfig` in the file `default-occ-cost-centers-config.ts`.

Zeyber (changes requested):
Is it possible validate this scenario in an e2e test too?

projects/storefrontapp-e2e-cypress/cypress/integration/b2b/regression/checkout/asm-cost-center.e2e-spec.ts
```diff
@@ -0,0 +1,72 @@
...
+    asm.agentLogin();
+  });
+
+  it('should start customer emulation', () => {
+    cy.get('cx-csagent-login-form').should('not.exist');
+    cy.get('cx-customer-selection').should('exist');
+    cy.get('cx-customer-selection form').within(() => {
+      cy.get('[formcontrolname="searchTerm"]')
+        .should('not.be.disabled')
+        .type('william.hunter@pronto-hw.com');
```
Zeyber: This email recurs a lot. Lets set a constant for this so its less places to edit if the email should change.

projects/storefrontapp-e2e-cypress/cypress/integration/b2b/regression/checkout/asm-cost-center.e2e-spec.ts
```diff
@@ -0,0 +1,72 @@
...
+
+  it('should not show error on valid cost center', () => {
+    alerts.getErrorAlert().then(() => {
+      alerts.getErrorAlert().within(() => {
+        cy.get('button').click();
+      });
+    });
+    cy.intercept('PUT', '*costcenter?costCenterId=*').as('costCenterReq');
+    cy.get('cx-cost-center').within(() => {
+      cy.get('select').select('Pronto_Services');
```
Zeyber: Opinionated. But I prefer to set constants for options like these like `const costCenterToSelect = 'Pronto_Services'` as I feel its easier to read and debug later.

projects/storefrontapp-e2e-cypress/cypress/integration/b2b/regression/checkout/asm-cost-center.e2e-spec.ts
```diff
@@ -0,0 +1,72 @@
...
+  it('should show error on invalid cost center', () => {
+    b2bCheckout.addB2bProductToCartAndCheckout();
+    cy.get('cx-payment-type').within(() => {
+      cy.findByText('Account').click({ force: true });
+    });
+    cy.get('button.btn-primary').should('be.enabled').click({ force: true });
+    cy.intercept('PUT', '*costcenter?costCenterId=*').as('costCenterReq');
+
+    cy.get('cx-cost-center').within(() => {
+      cy.get('select').select('Rustic_Global');
```
Zeyber: Opinionated. But I prefer to set constants for options like these like const costCenterToSelect = 'Rustic_Global' as I feel its easier to read and debug later.

feature-libs/checkout/b2b/checkout-b2b.module.ts
```diff
@@ -5,15 +5,24 @@
...
 
 @NgModule({
   imports: [
     CheckoutB2BComponentsModule,
     CheckoutB2BCoreModule,
     CheckoutB2BOccModule,
   ],
+  providers: [
+    {
+      provide: HttpErrorHandler,
```
Zeyber: I think that we should move this intercepter into the `CheckoutB2BCoreModule` as the http-interceptors live in that directory.

feature-libs/checkout/b2b/core/http-interceptors/bad-request/bad-cost-center-request.handler.ts
```diff
@@ -0,0 +1,59 @@
...
+    }
+  }
+
+  protected getErrors(response: HttpErrorResponse): ErrorModel[] {
+    return (response.error?.errors || []).filter(
+      (error: any) => error.type !== 'JaloObjectNoLongerValidError'
+    );
+  }
+
+  private isCostCenterRequest(errorResponse: HttpErrorResponse): boolean {
```
Zeyber: We try not to make methods `private` as it makes it makes functions inaccessible from an extensibility perspective. We use `protected` instead so that they are still able to be accessed by lib consumers but not intentionally so.

feature-libs/checkout/b2b/core/http-interceptors/bad-request/bad-cost-center-request.handler.ts
```diff
@@ -0,0 +1,59 @@
...
+      this.globalMessageService.add(
+        { key: 'httpHandlers.invalidCostCenter' },
+        GlobalMessageType.MSG_TYPE_ERROR
+      );
+    }
+  }
+
+  protected getErrors(response: HttpErrorResponse): ErrorModel[] {
+    return (response.error?.errors || []).filter(
+      (error: any) => error.type !== 'JaloObjectNoLongerValidError'
```
Zeyber: Can we make an `enum` for error types specific to this handler?

projects/assets/src/translations/en/common.ts
Zeyber: To clarify this, I think plabadie is talking about `feature-libs/checkout/b2b/assets/translations/en/checkout-b2b.i18n.ts`

Zeyber (changes requested):

feature-libs/cart/base/core/http-interceptors/handlers/bad-cost-center-request.handler.ts
ArkadiySukharenko: Based on discussion with Patrick we decided to keep it as is.

feature-libs/checkout/b2b/core/http-interceptors/bad-request/bad-cost-center-request.handler.ts
```diff
@@ -0,0 +1,73 @@
...
+      this.isCostCenterRequest(errorResponse)
+    );
+  }
+
+  handleError(_request: HttpRequest<any>, response: HttpErrorResponse): void {
+    if (this.getErrors(response).some((e) => this.isEntityValidationError(e))) {
+      this.translationService
+        .translate('checkoutB2B.invalidCostCenter')
+        .subscribe((result: string) => {
+          this.globalMessageService.add(
```
Zeyber: You can pass keys to the `globalMessage.add` method like: 
`this.globalMessage.add({ key: 'checkoutB2B.invalidCostCenter' }, GlobalMessageType.MSG_TYPE_CONFIRMATION);`. You don't need to make a subscription to the translation service.

Zeyber (changes requested):

Zeyber (approved):

justinlee01 (approved):

PR #15837, opened 2022-06-10
fix: AMP violations on My Interests page (CXSPA-499)

kimhw0630:
Ensure data table headers cells are not blank


developpeurweb (changes requested):
Tables are standard across Spartacus starting version 5.0, there is no need for empty columns anymore (which are detected by Screen Readers). This arrangement is very similar to that of the Wish List, maybe some code can be recycled from there.

**Some observations:**

1. Please standardize columns to one column header per column. Currently, Screen Reader detects 5 columns instead of 4. The table is also missing the caption as we have it in the Wish List and Cart Page code. E.g., `<caption class="cx-visually-hidden"> Wish list contents. </caption>`
5 cols

2. The "Remove" button column is missing the "ACTIONS" header.
missing col header

3. The "Remove" button is standard, please copy over the styles from Wish List or Cart Page.
std action link wish list
std action link cart

4. The "Remove" button on Wish List and Cart Page both have a Screen-Reader-only notification "Selected item has been removed", please implement it accordingly.
SR item removed



5. When using "Sort by" filters there is also a Screen-Reader-only notification, to tell the user the page view has changed. Please refer to the PLP to see how it works.
page updated msg


6. I get 2 AMP warnings that are easy to fix. One heading jump, might as well just remove it since it's the same as the H1. The colour contrast is a theme bug, please send it to @morganm58 as a separate ticket, he knows what to do.
h3 bug

contrast bug

Please ping me directly, on Slack, if any doubts.

developpeurweb (approved):
LGTM. Works well with JAWS, NVDA and VO.

PR #15838, opened 2022-06-10
feat: Fetching products in carousel with stock information

justinlee01:
Closes [CXSPA-569](https://jira.tools.sap/browse/CXSPA-569)


projects/storefrontlib/cms-components/product/carousel/product-carousel/product-carousel.component.spec.ts
```diff
@@ -157,9 +158,23 @@ describe('ProductCarouselComponent', () => {
...
+        ProductScope.PRICE,
+        ProductScope.DETAILS,
+      ];
+
+      let items: Observable<Product>[] | undefined;
+      component.items$.subscribe((i) => {
+        expect(productService.get).toHaveBeenCalledTimes(2);
+        expect(productService.get).toHaveBeenCalledWith('1', scopes);
+        expect(productService.get).toHaveBeenCalledWith('2', scopes);
+        items = i;
```
plabadie: Do you need the done() function for assertions in a subscription?

justinlee01: Ah, yes, I think there should be a done(). Thanks for reminding me.

justinlee01: Actually, line 177 essentially checks that lines 171-175 are called, but I like the neater approach of putting all assertions in the callback.

projects/storefrontlib/cms-components/product/carousel/product-carousel/product-carousel.component.ts
```diff
@@ -15,7 +15,11 @@ import { CmsComponentData } from '../../../../cms-structure/page/model/cms-compo
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
 export class ProductCarouselComponent {
-  protected readonly PRODUCT_SCOPE = ProductScope.LIST;
+  protected readonly PRODUCT_SCOPE = [
+    ProductScope.LIST,
+    ProductScope.PRICE,
+    ProductScope.DETAILS,
```
plabadie: Is there a way to not load the whole product details in order to fix the product stock issue?  
Reasoning: The purpose of the scopes were to delay the moment we need to load the whole product details, usually when the user hits the product details page.

justinlee01: I'm not sure - these are the only scopes available for fetching products, at least written in SPA code:

- LIST = 'list'
- DETAILS = 'details'
- ATTRIBUTES = 'attributes'
- VARIANTS = 'variants'
- PRICE = 'price'

DETAILS seems to be the only adequate one here.

bullfroggy-sap (dismissed):

projects/storefrontlib/cms-components/product/carousel/product-carousel/product-carousel.component.ts
plabadie: Which fields are required for the out of stock check to work properly?
Perhaps there is a way to get those fields without all the product details. I don't know exactly what or how, this is a suggestion.

Perhaps the add to cart button on the search results page and/or category page can be examined for how they do it.  I don't know if they react to out of stock status and it they do, perhaps we can see if they work with less data than all the product details.

justinlee01: The search page does look for stock information - the add to cart button needs the stock info to know whether to hide itself.
Interestingly enough, the MerchandisingCarouselComponent also uses DETAILS.

I can create a new scope mapping for this case specifically, something called STOCK.

Screen Shot 2022-06-23 at 3 46 22 PM

plabadie: At a high level, a STOCK scope sounds like a good idea and optimal since it's probably a common use case that the product carousel would be used on the home page.
I don't know how involving this modification can be though.  I never added a scope.

In the end, I'm not 100% against using details, but if there is a way to be optimal with a reasonable/small amount of effort, it might be worth it since this component is likely used on the home page.

justinlee01: Turned out to be pretty simple after all - PR updated.

plabadie (approved):

justinlee01:
For Iceman reference:

To reiterate from standup, all of the research here was done on `OccProductAdapter`. This code doesn't do much if the CMS adapter does not use the product scopes.
For each product scope, the adapter calls `OccEndpointsService::buildUrl` to get each specific product endpoint with the product scope.
The adapter then uses `OccRequestsOptimizerService::scopedDataLoad` to mash all the URLs together.

And so that is how the scopes are combined to get the precise data - I'm not sure how contradictions are resolved (i.e scopes both refer to stocks, so which one wins? My guess would be the last one).

bullfroggy-sap (approved):

kimhw0630 (approved):

PR #15869, opened 2022-06-17
style: fixed star rating component styling for seo performance

morganm58:
fixes CXSPA-440
https://jira.tools.sap/browse/CXSPA-440


projects/storefrontstyles/scss/components/product/_star-rating.scss
```diff
@@ -9,6 +13,10 @@
     font-size: 20px;
     margin: 0 2.5px;
 
+    @include media-breakpoint-down(sm) {
+      font-size: 48px;
```
developpeurweb, line 17: @morganm58 Although it passes validation, this seems so big it takes the focal point away from the product. Please validate with designers and @Xymmer 

Screen Shot 2022-06-21 at 3 49 05 PM

Here's an alternative using padding only

Screen Shot 2022-06-21 at 3 47 31 PM

developpeurweb (changes requested):

PR #15789, opened 2022-05-25
chore: Re-wire the events in checkout for better readability and tracing

bgambocjaviniar:
closes https://jira.tools.sap/browse/CXSPA-519

PR #15870, opened 2022-06-20
fix: Spinner in guest checkout after event refactoring

bgambocjaviniar:
related https://jira.tools.sap/browse/CXSPA-603

bug started from https://github.com/SAP/spartacus/pull/15789

### DIFF
--- a/feature-libs/cart/base/root/events/multi-cart-event.listener.ts
+++ b/feature-libs/cart/base/root/events/multi-cart-event.listener.ts
@@ -23,7 +23,9 @@ export class MultiCartEventListener implements OnDestroy {
   protected onCartBaseAction(): void {
     this.subscriptions.add(
       this.eventService.get(LoadCartEvent).subscribe(({ userId, cartId }) => {
-        this.multiCartFacade.loadCart({ userId, cartId });
+        if (userId && cartId) {
+          this.multiCartFacade.loadCart({ userId, cartId });
+        }
       })
     );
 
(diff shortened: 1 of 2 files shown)


feature-libs/cart/base/root/events/multi-cart-event.listener.ts
```diff
@@ -23,7 +23,9 @@ export class MultiCartEventListener implements OnDestroy {
   protected onCartBaseAction(): void {
     this.subscriptions.add(
       this.eventService.get(LoadCartEvent).subscribe(({ userId, cartId }) => {
-        this.multiCartFacade.loadCart({ userId, cartId });
+        if (userId && cartId) {
```
bgambocjaviniar, line 26: reason why this needs to be done was from the previous checkout event refactoring https://github.com/SAP/spartacus/pull/15789

We now dispatch an extra layer of event to help with readability + tracing, but now it causes issues, where some events requires additional payloads.

------
This is really not the issue to fix in guest checkout. This fixes the error we get after login in, where it says "lack of permissions". The fix for guest checkout can be found in checkout-event.module.ts as it was forgotten to bootstrap the 'new' event listener

bgambocjaviniar: for the reviewer:

OPTION 1: conditional userId and cartId to dispatch load cart

OR

OPTION 2: should I change the logic for the LoginEvent to accept the userId and cartId as the payload? This will remove this conditional statement. I believe we can go this route as it make sense that the user might want to know know the current userId and cartId of the logged in user. What do you think?

WeizhengSap (approved):

PR #15875, opened 2022-06-20
fix: Site freezes after agent logs out of ASM

bullfroggy-sap:
Revert ["Site freezes when logging out of ASM"](https://github.com/SAP/spartacus/commit/10bf6c09ce691e1967d751eeb909f16dcef32cb3)
Explicitly set logout process back to false after the operation has completed
Create e2e and unit tests to verify new logout behavior

### DIFF
--- a/integration-libs/cdc/root/guards/cdc-logout.guard.spec.ts
+++ b/integration-libs/cdc/root/guards/cdc-logout.guard.spec.ts
@@ -18,10 +18,6 @@ class MockAuthService implements Partial<AuthService> {
   coreLogout() {
     return Promise.resolve();
   }
-
-  isUserLoggedIn(): Observable<boolean> {
-    return of(true);
-  }
 }
 
 const gigya = {
--- a/projects/storefrontapp-e2e-cypress/cypress/helpers/vendor/cds/merchandising-carousel.ts
+++ b/projects/storefrontapp-e2e-cypress/cypress/helpers/vendor/cds/merchandising-carousel.ts
@@ -344,12 +344,13 @@ export function navigateToHomepage(): void {
 
 export function navigateToCategory(
   categoryName: string,
-  categoryCode: string
+  categoryCode: string,
+  force: boolean = true
 ): void {
   const categoryPage = waitForCategoryPage(categoryCode, 'getCategory');
   cy.get('cx-category-navigation cx-generic-link a')
     .contains(categoryName)
-    .click({ force: true });
+    .click({ force });
   cy.wait(`@${categoryPage}`).its('response.statusCode').should('eq', 200);
 }
 
--- a/projects/storefrontapp-e2e-cypress/cypress/integration/regression/asm/asm.e2e-spec.ts
+++ b/projects/storefrontapp-e2e-cypress/cypress/integration/regression/asm/asm.e2e-spec.ts
@@ -8,6 +8,8 @@ import { getSampleUser } from '../../../sample-data/checkout-flow';
 import { clearAllStorage } from '../../../support/utils/clear-all-storage';
 import * as consent from '../../../helpers/consent-management';
 import { fillShippingAddress } from '../../../helpers/checkout-forms';
+import { waitForCategoryPage, waitForPage } from '../../../helpers/checkout-flow';
+import { navigateToCategory, navigateToHomepage } from '../../../helpers/vendor/cds/merchandising-carousel';
 
 let customer: any;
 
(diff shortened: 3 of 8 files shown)


projects/storefrontapp-e2e-cypress/cypress/helpers/vendor/cds/merchandising-carousel.ts
```diff
@@ -344,12 +344,13 @@ export function navigateToHomepage(): void {
...
   categoryName: string,
-  categoryCode: string
+  categoryCode: string,
+  force: boolean = true
 ): void {
   const categoryPage = waitForCategoryPage(categoryCode, 'getCategory');
   cy.get('cx-category-navigation cx-generic-link a')
     .contains(categoryName)
-    .click({ force: true });
+    .click({ force });
```
plabadie, line 353: Was this test flaky?

integration-libs/cdc/root/guards/cdc-logout.guard.spec.ts
```diff
@@ -18,10 +18,6 @@ class MockAuthService implements Partial<AuthService> {
   coreLogout() {
     return Promise.resolve();
   }
-
-  isUserLoggedIn(): Observable<boolean> {
-    return of(true);
-  }
```
plabadie, line 24: Why remove this?

bullfroggy-sap: It was added in the previous PR which was reverted.

projects/storefrontapp-e2e-cypress/cypress/helpers/vendor/cds/merchandising-carousel.ts
bullfroggy-sap: I'm not sure if it was flaky, but it seemed like bad practice to always force the click. That leads to clicks occurring while the ui is not technically [actionable](https://docs.cypress.io/guides/core-concepts/interacting-with-elements#Actionability). In my opinion, if a test is written correctly, force should never be necessary. So I made it optional to force the click (though force is still true by default to avoid breaking changes).

integration-libs/cdc/root/guards/cdc-logout.guard.spec.ts
plabadie: Ah ok, thanks.

projects/storefrontapp-e2e-cypress/cypress/helpers/vendor/cds/merchandising-carousel.ts
plabadie: We have many clicks with `force: true` in the test code base.  I agree that it is best not to use `force: true`, but in practice, not using it made many tests too flaky to be usable.  Expect to see many `force: true` in the tests code.
In this case, If the default is true then this should be fine, however I think the asm test should use force true for reliability.

Another reason why force: true is used a lot is that it's not within the scope of most tests to make assertions  about the navigation menu.  More over, I think that force: true is useful in the mobile viewports where the navigation menu is buried under the hamburger menu.

plabadie:
We use 2 branch name prefixes:  `epic` (`epic/CXSPA-9999`) for epics and `feature`  (`feature/CXSPA-9999`) for all the rest, even bug fixes like this PR.

bullfroggy-sap:
> We use 2 branch name prefixes: `epic` (`epic/CXSPA-9999`) for epics and `feature` (`feature/CXSPA-9999`) for all the rest, even bug fixes like this PR.

Oh, my bad. Will close this and make a new one with the correct prefix.

projects/storefrontapp-e2e-cypress/cypress/integration/regression/asm/asm.e2e-spec.ts
```diff
@@ -81,6 +89,13 @@ context('Assisted Service Module', () => {
       cy.get('button[title="Close ASM"]').click();
       cy.get('cx-asm-main-ui').should('exist');
       cy.get('cx-asm-main-ui').should('not.be.visible');
+
+      // CXSPA-301/GH-14914
+      // Must ensure that site is still functional after service agent logout
+      navigateToHomepage();
+      cy.get('cx-storefront.stop-navigating').should('exist');
+      navigateToCategory('Brands', 'brands', false);
```
plabadie, line 97: See other comment about using force: true

bullfroggy-sap: I made sure this is in a stable place before clicking so that force true is not needed.

PR #15883, opened 2022-06-22
chore: fix "Public API Changes" bot 

Platonn:
**Improvements**:
- make the "Public API changes" bot back working in CI (get rid of all the errors in the script)
  - side effect: reduce the time of executing "Public API changes" bot from 2h back to ~30min (see [test PR](https://github.com/SAP/spartacus/actions/runs/2580268027))
- reduced the warnings noise in logs:
  - disabled API-extractor warnings not relevant in our project, e.g. [warning for missing release tags like alpha/beta/internal/public in our TSDocs](https://api-extractor.com/pages/messages/ae-missing-release-tag/),
  - disabled API-extractor warnings about missing exports, because sometimes caused false-positive warnings e.g. when an interface is exported, but from a different library
  - disabled all warnings from TSDocs parser (checking the syntax of TSDocs is not in scope of our breaking changes detection bot; generated ~30k warnings per script execution before)
- reduced the warnings noise in the report comment:
  - API Extractor warnings are not printed to the report comment, but only in the console of the script - to avoid fake reports because of a different file path, for example:
     ```diff
     -// /github/workspace/branch-clone/dist/<...file path> - <warning message>
     +// /github/workspace/dist/<...file path> - <warning message>
    ```

**Changes**:
- analyze d.ts instead of TS files in dependent libraries (e.g. when finding `import {X} from '@spartacus/core'`, analyze `dist/core/.../.d.ts` instead of `/projects/core/.../.ts`)
- disable api-extractor warning of type `ae-missing-release-tag` and `ae-forgotten-export`
- disable all TSDoc warnings (of type `tsdocMessageReporting`)
- disable reporting warnings in the markdown report, but leave them in the script console output
- add explicit return type for `APP_INITIALIZER`s that returned `()=>Promise<Config>` - to avoid api-extractor error `ERROR: Failed to fetch entity for import() type node: import("@spartacus/core").Config`
  - by the way, I added also missing return types for other `APP_INITIALIZER`s, which was not essential for this PR
- by the way, removed some redundant JSDoc notices `@member` from `occ-models.ts` - which caused warnings in api-extractor and were redundant

**Implementation notes**:
Previously for API Extractor tool we used the tsconfig with path mappings to source code (e.g. for resolving an import from `@spartacus/core` it looked up the source code TS files in `projects/core/*`).

According to https://api-extractor.com/pages/messages/ae-wrong-input-file-type/ , it seems that we were using the api-extractor tool not in a way that the tool's authors assumed. We analyzed TS files (only in dependency packages), but api-extractor expects to analyze d.ts

Moreover, since API Extractor v7.24.0 (released 2022-05-14), analyzing TS files throws an error (see [changelog](https://github.com/microsoft/rushstack/blob/main/apps/api-extractor/CHANGELOG.md#7240) and [troubleshooting page](https://api-extractor.com/pages/messages/ae-wrong-input-file-type/)), which caused our CI script for api-extractor to fail since 2022-05-14.

In this PR, we start using the "production" version of `tsconfig` (`tsconfig.app.prod.json` instead of `tsconfig.json`), which uses `"paths"` mapping to the `/dist` folder, which contains d.ts files.


tsconfig.api-extractor.json
```diff
@@ -1,6 +1,6 @@
 {
   //TODO: Remove file and reference to it when everything is adjusted to strict mode
-  "extends": "./tsconfig.json",
+  "extends": "./projects/storefrontapp/tsconfig.app.prod.json",
```
Platonn, line 3: Thanks to using production `"paths"` mappings in tsconfig, the dependencies types are resolved based on their compiled d.ts files instead of the source code .ts files

bgambocjaviniar (commented):
^ the merges above from develop was just to test the core pipeline :P 
looking good so far

I'll be adding comments if ever for this PR soon when I test shortly

feature-libs/asm/root/asm-loader.module.ts
```diff
@@ -26,7 +26,7 @@ export class AsmLoaderModule {}
  * to a real component; the router and state aren't available in an optimized
  * way during the APP_INITIALIZER.
  */
-export function asmFactory(asmEnablerService: AsmEnablerService) {
+export function asmFactory(asmEnablerService: AsmEnablerService): () => void {
   const isReady = () => {
```
bgambocjaviniar, line 30: not related to this file, but I believe you missed 2 last typings, which are:
 
- `feature-libs/asm/core/asm-core.module.ts` for `asmStatePersistenceFactory`
- `projects/storefrontlib/layout/a11y/skip-link/skip-link.module.ts` for `skipLinkFactory`

projects/core/src/config/config-initializer/config-initializer.module.ts
```diff
@@ -1,28 +1,29 @@
+import { LOCATION_INITIALIZED } from '@angular/common';
 import {
   APP_INITIALIZER,
   ModuleWithProviders,
   NgModule,
   Optional,
 } from '@angular/core';
-import { ConfigInitializerService } from './config-initializer.service';
+import { Config } from '../config-tokens';
```
bgambocjaviniar, line 8: 👍 since it came from @spartacus/core for relative paths instead of others like in feature-libs that will take from @spartacus/core

.github/api-extractor-action/api-extractor.json
```diff
@@ -307,6 +307,11 @@
...
         "logLevel": "warning"
         // "addToApiReportFile": false
+      },
+
+      // Spartaus: currently we don't use the release tags @alpha/@beta/@internal/@public in our JSDocs:
+      "ae-missing-release-tag": {
+        "logLevel": "none"
       }
 
       // "ae-extra-release-tag": {
```
bgambocjaviniar, line 323: In the report, there exist `ae-forgotten-export`, but I don't see a config option to disable it https://api-extractor.com/pages/messages/ae-forgotten-export/. 

Moreover, I don't see a reason for why it would complain/warn when I do see those models/interfaces/enums are being exported. 

Examples can be found from the report [here - fake develop](https://github.com/SAP/spartacus/pull/15908#issuecomment-1168097346)

bgambocjaviniar (changes requested):

.github/api-extractor-action/api-extractor.json
bgambocjaviniar: Not sure what to do with those warnings to be honest.

bgambocjaviniar (changes requested):
Should be ready to merge after fixing the missing typings AND answering whether the `ae-forgotten-export` is an issue` as the [report looks good to me](https://github.com/SAP/spartacus/pull/15908#issuecomment-1168097346) and [these logs showing the cache is being used from base](https://github.com/SAP/spartacus/runs/7083916038?check_suite_focus=true#step:4:91)

feature-libs/asm/root/asm-loader.module.ts
Platonn: Thanks for noticing. Now added types in https://github.com/SAP/spartacus/pull/15883/commits/825b245c3e0aa485cc672f8fef20cf04bdcd5f16

.github/api-extractor-action/api-extractor.json
Platonn: Those warnings are not helpful also because the `OrderEntry` for example IS exported, but from a different library. So the warning is false-positive in this case. Perhaps it's due to our misconfiguration or misusage of the API Extractor tool on our side... we would need to investigate more.

Getting notified about potentially forgotten exports is not a primary purpose of our Public API Changes bot. So I don't want invest time into investigating.

Although getting hints like “You forgot to export something” would be nice nice…. if doens’t work correctly ootb in our case, I'm for disabling this warning.

bgambocjaviniar (approved):

PR #15885, opened 2022-06-22
feat: add purchase order column to order history list

kimhw0630:
if  purchaseOrderNumber property existence from the backend result and then show PO Number column in Order history


feature-libs/order/components/order-history/order-history.component.spec.ts
```diff
@@ -120,6 +144,20 @@ class MockReplenishmentOrderHistoryFacade
   }
 }
 
+@Directive({
```
justinlee01: I know several existing SPA files do this, but I wonder if this should be declared in the `describe`. It seems like a waste of memory otherwise.

justinlee01 (dismissed):

feature-libs/order/components/order-history/order-history.component.html
```diff
@@ -53,92 +53,242 @@ <h2 *ngIf="!type.replenishmentOrder">
...
-              </tr>
-            </tbody>
-          </table>
+
+          <ng-container *cxFeatureLevel="'5.1'">
+            <table
+              role="table"
+              id="order-history-table"
+              class="table cx-order-history-table"
+              [ngClass]="{ 'cx-order-history-table-po': hasPONumber }"
```
Zeyber: Could we move the featureLevel condition into the class on `hasPONumber` so that we can add deprecation notes to the newly exposed flag? We shouldn't need to duplicate the table in the template if the flag itself depends on the feature level.

feature-libs/order/components/order-history/order-history.component.spec.ts
Zeyber: It could be easier to use a spy to mock the `isLevel` condition on the `FeatureConfigService`.

Zeyber (changes requested):

feature-libs/order/components/order-history/order-history.component.html
kimhw0630: done thanks

feature-libs/order/components/order-history/order-history.component.html
```diff
@@ -86,7 +108,44 @@ <h2 *ngIf="!type.replenishmentOrder">
...
                   >
                 </td>
-                <td class="cx-order-history-placed">
+                <ng-container *ngIf="hasPONumber">
+                  <td role="cell" class="cx-order-history-po">
+                    <div class="cx-order-history-label">
+                      {{ 'orderHistory.PONumber' | cxTranslate }}
+                    </div>
+                    <a
+                      *ngIf="order.purchaseOrderNumber"
```
Zeyber, line 114: Do we need this condition if `hasPONumber` is already true?

feature-libs/order/components/order-history/order-history.component.html
```diff
@@ -53,12 +53,33 @@ <h2 *ngIf="!type.replenishmentOrder">
...
-          <table id="order-history-table" class="table cx-order-history-table">
+
+          <table
+            role="table"
+            id="order-history-table"
+            class="table cx-order-history-table"
+            [ngClass]="{ 'cx-order-history-table-po': hasPONumber }"
+          >
+            <ng-container *cxFeatureLevel="'5.1'">
+              <caption class="cx-visually-hidden">
```
Zeyber: Can we put `*cxFeatureLevel` on the `caption` element instead?

feature-libs/order/components/order-history/order-history.component.ts
```diff
@@ -42,6 +50,11 @@ export class OrderHistoryComponent implements OnDestroy {
         if (orders?.pagination?.sort) {
           this.sortType = orders.pagination.sort;
         }
+        if (orders?.orders?.length) {
+          this.hasPONumber =
+            orders.orders[0].purchaseOrderNumber !== undefined &&
```
Zeyber: We could simplify this to `orders.orders[0].purchaseOrderNumber && this.featureConfigService?.isLevel('5.1')`.

feature-libs/order/components/order-history/order-history.component.ts
```diff
@@ -29,11 +35,13 @@ export class OrderHistoryComponent implements OnDestroy {
...
     protected orderHistoryFacade: OrderHistoryFacade,
     protected translation: TranslationService,
-    protected replenishmentOrderHistoryFacade: ReplenishmentOrderHistoryFacade
+    protected replenishmentOrderHistoryFacade: ReplenishmentOrderHistoryFacade,
+    @Optional() protected featureConfigService?: FeatureConfigService
   ) {}
 
   private PAGE_SIZE = 5;
   sortType: string;
+  hasPONumber = false;
```
Zeyber: Could we use an observable for this so we can use `async` pipe in the template?

feature-libs/order/components/order-history/order-history.component.spec.ts
```diff
@@ -223,6 +262,37 @@ describe('OrderHistoryComponent', () => {
     expect(elements.length).toEqual(2);
     expect(component.sortType).toEqual('byDate');
   });
+
+  it('should display PO Number & Cost Center', () => {
+    spyOn(featureConfigService, 'isLevel')
```
Zeyber: Do we need this when our mock service already returns `true` for `isLevel()`?

Zeyber (changes requested):
Could you also please write some `TODO:` cleanup notes for removing `FeatureConfigService` for when we move to the next major?

feature-libs/order/components/order-history/order-history.component.html
kimhw0630: Yes, we need it due to an accessibility problem with an empty anchor tag value.
PONumber is for checking property existence and order.purchaseOrderNumber for checking null.

feature-libs/order/components/order-history/order-history.component.spec.ts
kimhw0630: we want to return true only if 5.1
spyOn(featureConfigService, 'isLevel')
      .withArgs('5.1')
      .and.returnValue(true)
      .withArgs('!5.1')
      .and.returnValue(false);

feature-libs/order/components/order-history/order-history.component.ts
kimhw0630: We don't want to construct multiple static observables and re-assign the observable variable, the same applies to the existing 'sortType'

feature-libs/order/components/order-history/order-history.component.ts
kimhw0630: We want to check 'purchaseOrderNumber' property exists not the value.  if purchaseOrderNumber exists and the value is empty then the backend returns an empty string.

feature-libs/order/components/order-history/order-history.component.html
kimhw0630: done thanks~

feature-libs/order/components/order-history/order-history.component.spec.ts
Zeyber: Could you please point out where these args are inputted? I am not aware of us using version flags for unit tests in the CI. It also appears that this test should fail in the instance of `< 5.1` if this were the case.

feature-libs/order/components/order-history/order-history.component.ts
Zeyber: Okay. In that case, could we simplify to `orders.orders[0].purchaseOrderNumber !== undefined && this.featureConfigService?.isLevel('5.1')`?

Zeyber: Better yet, we can remove the `if` and have: 
`this.hasPONumber = orders?.orders?.[0].purchaseOrderNumber !== undefined && this.featureConfigService?.isLevel('5.1')`

feature-libs/order/components/order-history/order-history.component.ts
Zeyber: Okay. Lets leave this undefined like `hasPONumber: boolean` as the true/false value will become meaningful when it is instantiated later.

feature-libs/order/components/order-history/order-history.component.ts
Zeyber: It could also be nice to add a `TODO:`about cleaning up the feature level in this equation when moving to the next major.

feature-libs/order/components/order-history/order-history.component.spec.ts
kimhw0630: thanks, you are right we don't need 'spyOn'.  tested both 5.1 and 4.2.

feature-libs/order/components/order-history/order-history.component.ts
kimhw0630: done thanks~

feature-libs/order/components/order-history/order-history.component.ts
kimhw0630: done thanks~

feature-libs/order/components/order-history/order-history.component.spec.ts
```diff
@@ -223,6 +260,31 @@ describe('OrderHistoryComponent', () => {
     expect(elements.length).toEqual(2);
     expect(component.sortType).toEqual('byDate');
   });
+
+  it('should display PO Number & Cost Center', () => {
```
Zeyber, line 264: We should also test that columns don't appear when there is no `PONumber`. These expects could be added to an existing test.

Zeyber (changes requested):

feature-libs/order/components/order-history/order-history.component.spec.ts
kimhw0630: line 265 ~ 271 is checking for it without PONumber and
line 273 ~ 285 is fo with PONumber
Thanks~

Zeyber (approved):

PR #15887, opened 2022-06-23
test: Enhance checkout replenishment e2e

bgambocjaviniar:
two things:

- ~~remove date bug, where 24h time for midnight translates to 24:xx instead of 00:xx. Basically, backend for the replenishment endpoint does not accept 24 for midnight, but 00  / reference to fix conversion https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Locale/hourCycle~~ (WILL be done in https://github.com/SAP/spartacus/pull/15901)
- decouple the long checkout replenishment into multiple specs in order to remove persisted 'context' application state

closes https://jira.tools.sap/browse/CXSPA-530

-----------

EDIT 1 for flaky found from related test
- 7 out of 15 tests, a new flaky showed related to the test from this [commit](https://github.com/SAP/spartacus/commit/166c1594dd2d1cf3eb78d7c8447a3dee028026ae) `166c159`  and shown in this [dashboard](https://dashboard.cypress.io/projects/k3nmep/runs/31629/test-results?actions=%5B%5D&browsers=%5B%5D&groups=%5B%5D&isFlaky=%5B%5D&modificationDateRange=%7B%22startDate%22%3A%221970-01-01%22%2C%22endDate%22%3A%222038-01-19%22%7D&orderBy=EXECUTION_ORDER&oses=%5B%5D&specs=%5B%5D&statuses=%5B%7B%22value%22%3A%22FAILED%22%2C%22label%22%3A%22FAILED%22%7D%5D&testingTypesEnum=%5B%5D&utm_source=github)
- - from the video, I can see it is related to a retry due to JALO happening in this test (checkout has a retry mechanism, where if an error is related to JALO, then it will retry)
- - - solution: make fixtures in order to bypass the JALo error that can happen (basically stubbing the response)


RadhepS (dismissed):

bgambocjaviniar (commented):
https://github.com/SAP/spartacus/pull/15887/commits/166c1594dd2d1cf3eb78d7c8447a3dee028026ae

will restart the /15 test after figuring out how to tackle the e2e. This is a new flake that I never seen replenishment failed before. From the dashboard of this commit https://dashboard.cypress.io/projects/k3nmep/runs/31629/test-results/3407843b-1fc5-43dc-8c45-dcd597778b09, it shows that it failed in shipping address of `account` payment because the request failed due to the get checkout details endpoint (JALO error 100%), and then tries to wait for the update address, which won't fire since it happens after getting the details, but the updateAddress is blocking the retry call for the next ones.

Will try things out soon

projects/storefrontapp-e2e-cypress/cypress/helpers/b2b/b2b-checkout.ts
```diff
@@ -430,36 +452,34 @@ export function reviewB2bOrderConfirmation(
...
-    )}/users/**/carts/**/*?fields=deliveryAddress(FULL),deliveryMode(FULL),paymentInfo(FULL),costCenter(FULL),purchaseOrderNumber,paymentType(FULL)*`,
-  }).as(alias);
+  cy.intercept(
+    {
+      method: 'GET',
+      path: `${Cypress.env('OCC_PREFIX')}/${Cypress.env(
+        'BASE_SITE'
+      )}/users/**/carts/**/*?fields=deliveryAddress(FULL),deliveryMode(FULL),paymentInfo(FULL),costCenter(FULL),purchaseOrderNumber,paymentType(FULL)*`,
+    },
+    { fixture, statusCode: 200 }
```
RadhepS, line 481: Do we want to stub responses for e2es?

If the data returned from the call was invalid, we wouldn't know here right? (Again not sure if we care)
If our CI servers was slow or something we wouldn't know because the fixture is returned instantly (Don't know if we care)

We are headless so we are testing our entire product but we're closely coupled with Commerce Cloud right now, so should we be testing with the real data they return?

RadhepS (approved):

projects/storefrontapp-e2e-cypress/cypress/helpers/b2b/b2b-checkout.ts
RadhepS: Discussed this in a call, it will be used for the checkout tests

PR #15889, opened 2022-06-23
chore: Deprecate outlet static context property

Zeyber:
Closes: https://jira.tools.sap/browse/CXSPA-595


Zeyber:
In this fix, we make use of the `getLastValueSync()` method. This makes the usage of `context$` synchronous by nature in our implementation of Spartacus. `context` is used in many methods of our implementation already. However, because using `context$` is to encourage asynchronous usage, we have decided not to implement this method of deprecation for existing methods.

PR #15907, opened 2022-06-27
test: Improve currency checkout page core e2e

bgambocjaviniar:
2 things fixed:

1. fixed the plugin.js, where we had 2 module.export (should only have 1 and it took the latest export). This caused e2es using `cy.requireLoggedIn()` without an input param to have an undefined email, which can cause e2es to overlap. Basically, how it was before, the email that was generated when using `cy.requireLoggedIn() -> without input param` is `cypress_user_somethingsomething_undefined@sapcx.com`
- merging the module.export, it now displays the current timestamp instead of `_undefined`

2. Put the tests in one assertion instead of depending on cy.restoreStorage AND cy.saveStorage as each `it` should not depend on the previous states.

closes https://jira.tools.sap/browse/CXSPA-656

### DIFF
--- a/projects/storefrontapp-e2e-cypress/cypress/integration/regression/site-context/currency/currency-checkout-page.core-e2e-spec.ts
+++ b/projects/storefrontapp-e2e-cypress/cypress/integration/regression/site-context/currency/currency-checkout-page.core-e2e-spec.ts
@@ -1,5 +1,7 @@
-import { manipulateCartQuantity } from '../../../../helpers/cart';
+import { waitForCartPageData } from '../../../../helpers/b2b/b2b-saved-cart';
 import * as siteContextSelector from '../../../../helpers/site-context-selector';
+import { product } from '../../../../sample-data/checkout-flow';
+import { clearAllStorage } from '../../../../support/utils/clear-all-storage';
 
 describe('Currency switch - checkout page', () => {
   const checkoutShippingPath =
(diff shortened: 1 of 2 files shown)


projects/storefrontapp-e2e-cypress/cypress/integration/regression/site-context/currency/currency-checkout-page.core-e2e-spec.ts
```diff
@@ -9,28 +11,20 @@ describe('Currency switch - checkout page', () => {
...
   });
 
   siteContextSelector.stub(
     siteContextSelector.CURRENCY_REQUEST,
     siteContextSelector.CURRENCIES
   );
 
-  describe('populate cart, history, quantity', () => {
-    it('should have basic data', () => {
-      manipulateCartQuantity();
```
bgambocjaviniar, line 27: no need to have this as it's not even related to `currency`. That helper function just plays with the cart quantity to have some kind of cart, but I replaced it with `waitForCartPageData(product);`. Basically, it adds a product to the cart in the backend for the user/

projects/storefrontapp-e2e-cypress/cypress/plugins/index.js
```diff
@@ -3,16 +3,6 @@ const fs = require('fs');
...
-
-  /* Set exact timestamp to be shared in all spec files */
-  config.env.TIMESTAMP = Date.now() - 1535535333333;
-  return config;
-};
-
-/**
- * Read file but return null if no file found (instead of failing the test).
- */
-module.exports = (on, config) => {
```
bgambocjaviniar, line 15: this is what broke `cy.requireLoggedIn()` <-- without an input param, where the user's timestamp is always `_undefined`, therefore might have been breaking overall e2es

bgambocjaviniar: You can verify the facts by going to `projects/storefrontapp-e2e-cypress/cypress/support/require-logged-in.commands.ts` and append `console.log('username', username);` after line 90.

```
    const username =
      account.registrationData.email ||
      generateMail(account.user, options.freshUserOnTestRefresh);

    console.log('username', username);

```

You'll see the logged in user in develop will be `cypress_user_somethingsomething_undefined@sapcx.com` instead of a proper timestamp. `_undefined`

PR #15910, opened 2022-06-27
CXEC:14117 - Select first payment type by default in B2B even if one payment type enabled

I850322:
https://jira.tools.sap/browse/CXEC-14117


feature-libs/checkout/b2b/components/checkout-payment-type/checkout-payment-type.component.spec.ts
```diff
@@ -55,6 +71,57 @@ const mockActivatedRoute = {
   },
 };
 
+describe('CheckoutOnePaymentTypeComponent', () => {
```
ercultimate, line 74: This should be part of the existing unit test group for `CheckoutPaymentTypeComponent`

feature-libs/checkout/b2b/components/checkout-payment-type/checkout-payment-type.component.ts
```diff
@@ -42,22 +42,50 @@ export class CheckoutPaymentTypeComponent {
...
+            availablePaymentTypes[0].code as string,
+            this.poNumberInputElement?.nativeElement?.value
+          )
+          .subscribe({
+            complete: () => this.onSuccess(),
+            error: () => this.onError(),
+          });
+        return availablePaymentTypes[0];
+      } else {
+        return availablePaymentTypes[1];
```
ercultimate: This case seems to be incomplete or incorrect.  Can you please explain purpose of the case?

I am reading it as "if there is no previous selection or the previous selection is not in the list, and the available types is 0 or 2+, then use the 2nd item in the payment array"

feature-libs/checkout/b2b/components/checkout-payment-type/checkout-payment-type.component.spec.ts
```diff
@@ -55,6 +71,57 @@ const mockActivatedRoute = {
...
+    })
+  );
+
+  beforeEach(() => {
+    fixture = TestBed.createComponent(CheckoutPaymentTypeComponent);
+    component = fixture.componentInstance;
+    fixture.detectChanges();
+  });
+
+  it('get payment type selected by default if one payment type returned ', (done) => {
```
ercultimate:
```suggestion
  it('should get payment type selected by default if one payment type is returned', (done) => {
```

feature-libs/checkout/b2b/components/checkout-payment-type/checkout-payment-type.component.spec.ts
```diff
@@ -31,6 +31,19 @@ class MockCheckoutPaymentTypeService
   );
 }
 
+class MockCheckoutOnePaymentTypeService
+  implements Partial<CheckoutPaymentTypeFacade>
+{
+  getPaymentTypes = createSpy().and.returnValue(of(mockDisableOnePaymentTypes));
+  setPaymentType = createSpy().and.returnValue(of('setPaymentType'));
```
ercultimate: They type is listed as `unknown`, please use a more neutral value or a reasonable mock value for the default return.

```suggestion
  setPaymentType = createSpy().and.returnValue(of(undefined));
```

ercultimate (changes requested):

feature-libs/checkout/b2b/components/checkout-payment-type/checkout-payment-type.component.ts
I850322: There should be at least one payment type we will be getting back from backend. So there would be either one payment type or more than one payment type.

I850322: I changed the logic in which I am checking whether at-least a payment type is present. (Backend will return at-least one payment type) to make it more understandable and readable.

feature-libs/checkout/b2b/components/checkout-payment-type/checkout-payment-type.component.spec.ts
I850322: I created separate CheckoutPaymentTypeComponent because I have to use different class for provide: CheckoutPaymentTypeFacade if there is way to use difference class for same provider then I can change it.

feature-libs/checkout/b2b/components/checkout-payment-type/checkout-payment-type.component.spec.ts
I850322: Done.

feature-libs/checkout/b2b/components/checkout-payment-type/checkout-payment-type.component.spec.ts
I850322: Done.

feature-libs/checkout/b2b/components/checkout-payment-type/checkout-payment-type.component.spec.ts
ercultimate: Yes, it's possible.  For a given test
- use subjects that are manually controlled
- you can override the mock's return value using a `jasmine.Spy`
- override the TestBed configuration with `TestBed.overrideProvider()` https://angular.io/api/core/testing/TestBed#overrideprovider

Unfortunately, for this component, to do either requires a restructuring of the spec file.  I would recommend restructuring the spec file, but we should wait for word from a Spartacus team member first.

feature-libs/checkout/b2b/components/checkout-payment-type/checkout-payment-type.component.ts
```diff
@@ -42,22 +42,51 @@ export class CheckoutPaymentTypeComponent {
...
-          CheckoutStepType.PAYMENT_DETAILS,
-          selected?.code === B2BPaymentTypeEnum.ACCOUNT_PAYMENT
-        );
-      })
-    );
+      map((state) => state.data)
+    ),
+    this.paymentTypes$,
+  ]).pipe(
+    map(([prevSelected, availablePaymentTypes]) => {
```
Matejk00: Can we rename `prevSelected` to more informative variable name like: `selectedPaymentType`?

We can specify types for `map` argument here. The first one will be `PaymentType` or `undefined` as payment can be not selected and second will be an array of `PaymentType`.

```ts
 map(
      ([selectedPaymentType, availablePaymentTypes]: [
        PaymentType | undefined,
        PaymentType[]
      ]) => {
```

What do you think?

feature-libs/checkout/b2b/components/checkout-payment-type/checkout-payment-type.component.ts
```diff
@@ -42,22 +42,51 @@ export class CheckoutPaymentTypeComponent {
...
+            availablePaymentTypes[0].code as string,
+            this.poNumberInputElement?.nativeElement?.value
+          )
+          .subscribe({
+            complete: () => this.onSuccess(),
+            error: () => this.onError(),
+          });
+        return availablePaymentTypes[0];
+      }
+      // } else {
```
Matejk00: Do we need this part of code?

feature-libs/checkout/b2b/components/checkout-payment-type/checkout-payment-type.component.ts
```diff
@@ -42,22 +42,51 @@ export class CheckoutPaymentTypeComponent {
...
+      if (prevSelected) {
+        if (
+          availablePaymentTypes.find((availablePaymentType) => {
+            return availablePaymentType.code === prevSelected.code;
+          })
+        ) {
+          return prevSelected;
+        }
+      }
+      if (availablePaymentTypes.length !== 0) {
```
Matejk00: Not equal to zero you mean greater than 0 right? 

or shouldn't this to be enough:

```ts
if (availablePaymentTypes.length) {
....
```

feature-libs/checkout/b2b/components/checkout-payment-type/checkout-payment-type.component.ts
```diff
@@ -42,22 +42,51 @@ export class CheckoutPaymentTypeComponent {
...
-          selected?.code === B2BPaymentTypeEnum.ACCOUNT_PAYMENT
-        );
-      })
-    );
+      map((state) => state.data)
+    ),
+    this.paymentTypes$,
+  ]).pipe(
+    map(([prevSelected, availablePaymentTypes]) => {
+      if (prevSelected) {
```
Matejk00: This can be combined into one `if` statement I guess:

```ts
if (
        selectedPaymentType &&
        availablePaymentTypes.find(
          (availablePaymentType) =>
            availablePaymentType.code === selectedPaymentType?.code
        )
      ) {
        return selectedPaymentType;
      }
```

feature-libs/checkout/b2b/components/checkout-payment-type/checkout-payment-type.component.ts
```diff
@@ -42,22 +42,51 @@ export class CheckoutPaymentTypeComponent {
...
+          })
+        ) {
+          return prevSelected;
+        }
+      }
+      if (availablePaymentTypes.length !== 0) {
+        this.busy$.next(true);
+        this.checkoutPaymentTypeFacade
+          .setPaymentType(
+            availablePaymentTypes[0].code as string,
```
Matejk00: Do we need casting `availablePaymentTypes[0].code` value to the `string`?

feature-libs/checkout/b2b/components/checkout-payment-type/checkout-payment-type.component.ts
```diff
@@ -42,22 +42,51 @@ export class CheckoutPaymentTypeComponent {
...
+          availablePaymentTypes.find((availablePaymentType) => {
+            return availablePaymentType.code === prevSelected.code;
+          })
+        ) {
+          return prevSelected;
+        }
+      }
+      if (availablePaymentTypes.length !== 0) {
+        this.busy$.next(true);
+        this.checkoutPaymentTypeFacade
```
Matejk00: We are re-using a similar part of code in the file on lines 133:141.
What do you think to delegate this into a separate method? And call this method instead in both places but with different arguments.

Matejk00 (changes requested):

feature-libs/checkout/b2b/components/checkout-payment-type/checkout-payment-type.component.ts
I850322: Line 133:141 having below which goes to different route
` complete: () => this.checkoutStepService.next(this.activatedRoute),
        error: () => this.onError(),`
but Line 67:75
` complete: () => this.onSuccess(),
              error: () => this.onError(),`
this makes both call different.

feature-libs/checkout/b2b/components/checkout-payment-type/checkout-payment-type.component.ts
I850322: Yes, In payment type interface code attribute is defined as string | undefined and in setPaymentType method input must be of type string.

feature-libs/checkout/b2b/components/checkout-payment-type/checkout-payment-type.component.ts
I850322: Yes, I combined them. Thank you

feature-libs/checkout/b2b/components/checkout-payment-type/checkout-payment-type.component.ts
I850322: Yes, It should be enough, did the change.

feature-libs/checkout/b2b/components/checkout-payment-type/checkout-payment-type.component.ts
I850322: Nopes. I removed it.

feature-libs/checkout/b2b/components/checkout-payment-type/checkout-payment-type.component.ts
I850322: I think it's good to defined the types for map argument. so I added them.

Matejk00: Thanks and what about variable name like: `selectedPaymentType`?

feature-libs/checkout/b2b/components/checkout-payment-type/checkout-payment-type.component.ts
Matejk00: There is a possibility to create a method and pass specific functions as arguments of the function. But maybe it's a bit of overkill in this case. Let's keep it as you proposed.

feature-libs/checkout/b2b/components/checkout-payment-type/checkout-payment-type.component.ts
I850322: Yes, I changed the variable name to `selectedPaymentType`

Matejk00 (approved):

feature-libs/checkout/b2b/components/checkout-payment-type/checkout-payment-type.component.spec.ts
ercultimate: As there was no followup from the Spartacus reviewers, I will consider this accepted and resolved.

ercultimate (approved):

PR #15923, opened 2022-06-29
chore: fix "Public API Changes" bot (#15883) [backport to 4.3.x]

Platonn:
**Improvements**:
- make the "Public API changes" bot back working in CI (get rid of all the errors in the script)
  - side effect: reduce the time of executing "Public API changes" bot from 2h back to ~30min (see [test PR](https://github.com/SAP/spartacus/actions/runs/2580268027))
- reduced the warnings noise in logs:
  - disabled API-extractor warnings not relevant in our project, e.g. [warning for missing release tags like alpha/beta/internal/public in our TSDocs](https://api-extractor.com/pages/messages/ae-missing-release-tag/),
  - disabled API-extractor warnings about missing exports, because sometimes caused false-positive warnings e.g. when an interface is exported, but from a different library
  - disabled all warnings from TSDocs parser (checking the syntax of TSDocs is not in scope of our breaking changes detection bot; generated ~30k warnings per script execution before)
- reduced the warnings noise in the report comment:
  - API Extractor warnings are not printed to the report comment, but only in the console of the script - to avoid fake reports because of a different file path, for example:
     ```diff
     -// /github/workspace/branch-clone/dist/<...file path> - <warning message>
     +// /github/workspace/dist/<...file path> - <warning message>
    ```

**Changes**:
- analyze d.ts instead of TS files in dependent libraries (e.g. when finding `import {X} from '@spartacus/core'`, analyze `dist/core/.../.d.ts` instead of `/projects/core/.../.ts`)
- disable api-extractor warning of type `ae-missing-release-tag` and `ae-forgotten-export`
- disable all TSDoc warnings (of type `tsdocMessageReporting`)
- disable reporting warnings in the markdown report, but leave them in the script console output
- add explicit return type for `APP_INITIALIZER`s that returned `()=>Promise<Config>` - to avoid api-extractor error `ERROR: Failed to fetch entity for import() type node: import("@spartacus/core").Config`
  - by the way, I added also missing return types for other `APP_INITIALIZER`s, which was not essential for this PR
- by the way, removed some redundant JSDoc notices `@member` from `occ-models.ts` - which caused warnings in api-extractor and were redundant

**Implementation notes**:
Previously for API Extractor tool we used the tsconfig with path mappings to source code (e.g. for resolving an import from `@spartacus/core` it looked up the source code TS files in `projects/core/*`).

According to https://api-extractor.com/pages/messages/ae-wrong-input-file-type/ , it seems that we were using the api-extractor tool not in a way that the tool's authors assumed. We analyzed TS files (only in dependency packages), but api-extractor expects to analyze d.ts

Moreover, since API Extractor v7.24.0 (released 2022-05-14), analyzing TS files throws an error (see [changelog](https://github.com/microsoft/rushstack/blob/main/apps/api-extractor/CHANGELOG.md#7240) and [troubleshooting page](https://api-extractor.com/pages/messages/ae-wrong-input-file-type/)), which caused our CI script for api-extractor to fail since 2022-05-14.

In this PR, we start using the "production" version of `tsconfig` (`tsconfig.app.prod.json` instead of `tsconfig.json`), which uses `"paths"` mapping to the `/dist` folder, which contains d.ts files.

### DIFF
--- a/projects/core/src/cart/cart-persistence.module.ts
+++ b/projects/core/src/cart/cart-persistence.module.ts
@@ -2,13 +2,14 @@ import { APP_INITIALIZER, ModuleWithProviders, NgModule } from '@angular/core';
 import { ActionReducer, MetaReducer, META_REDUCERS } from '@ngrx/store';
 import { tap } from 'rxjs/operators';
 import { ConfigInitializerService } from '../config/config-initializer/config-initializer.service';
+import { Config } from '../config/config-tokens';
 import { MultiCartStatePersistenceService } from './services/multi-cart-state-persistence.service';
 import { activeCartInitialState } from './store/reducers/multi-cart.reducer';
 
 export function cartStatePersistenceFactory(
   cartStatePersistenceService: MultiCartStatePersistenceService,
   configInit: ConfigInitializerService
-) {
+): () => Promise<Config> {
   const result = () =>
     configInit
       .getStable('context')
--- a/.github/api-extractor-action/api-extractor.json
+++ b/.github/api-extractor-action/api-extractor.json
@@ -305,8 +305,19 @@
      */
     "extractorMessageReporting": {
       "default": {
-        "logLevel": "warning"
-        // "addToApiReportFile": false
+        "logLevel": "warning",
+        "addToApiReportFile": false
+      },
+
+      // Spartaus: currently we don't use the release tags @alpha/@beta/@internal/@public in our JSDocs:
+      "ae-missing-release-tag": {
+        "logLevel": "none"
+      },
+
+      // Spartacus: disabled due to false-positive warnings e.g. when an interface is exported, but from a different library
+      "ae-forgotten-export": {
+        "logLevel": "none",
+        "addToApiReportFile": false
       }
 
       // "ae-extra-release-tag": {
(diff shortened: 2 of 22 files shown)


projects/core/src/cart/cart-persistence.module.ts
```diff
@@ -2,13 +2,14 @@ import { APP_INITIALIZER, ModuleWithProviders, NgModule } from '@angular/core';
...
 import { ConfigInitializerService } from '../config/config-initializer/config-initializer.service';
+import { Config } from '../config/config-tokens';
 import { MultiCartStatePersistenceService } from './services/multi-cart-state-persistence.service';
 import { activeCartInitialState } from './store/reducers/multi-cart.reducer';
 
 export function cartStatePersistenceFactory(
   cartStatePersistenceService: MultiCartStatePersistenceService,
   configInit: ConfigInitializerService
-) {
+): () => Promise<Config> {
```
Platonn, line 12: Headsup: In `develop` this file is placed in the new cart lib. But in 4.3.x it still lives in the core lib.
See the [corresponding file](https://github.com/SAP/spartacus/pull/15923/files#diff-b7392b6530ed0e2819d57bac03344b40f9e03b3f2d7a6c5c5c7e3afa739b3819) in the original PR to develop.

No action needed.

bgambocjaviniar (approved):

PR #15927, opened 2022-06-29
decision: constructor parameters - use or not

Platonn:


Platonn:
Decided. moved to wiki

PR #15929, opened 2022-06-30
test: Fixed asm flaky e2es

bgambocjaviniar:
Notes:

- removed redundant registration that happens twice at the beginning and other duplicate tasks
- make sure to wait for asm input box to no longer be disabled before typing
- removed unused functions
- removed duplicates and reuse the intended function for customer emulation tests
- instead of visiting a page by using `cy.visit`, we use the my-account dropdown instead. Visiting pages is not a user journey on how a user would navigate the page. Also, visiting a page is like a refresh as it's changing the url and going to that route through the browser instead of making it a user journey (an end to end test).
- in addition, the visit caused most of the customer agent session timeout because of some existing bug in the feature

---
important: 
- reproducible asm agent session timeout as shown in videos by just waiting ~6 mins in a page. Needs fix in feature for token expiration instead, however, the videos in the issue shows flakiness as well (not related to session timeout), which should be fixed by this PR

closes https://jira.tools.sap/browse/CXSPA-661

### DIFF
--- a/projects/storefrontapp-e2e-cypress/cypress/helpers/asm.ts
+++ b/projects/storefrontapp-e2e-cypress/cypress/helpers/asm.ts
@@ -4,12 +4,15 @@ import * as checkout from '../helpers/checkout-flow';
 import { fillShippingAddress } from '../helpers/checkout-forms';
 import * as consent from '../helpers/consent-management';
 import * as profile from '../helpers/update-profile';
+import { SampleUser } from '../sample-data/checkout-flow';
 import { interceptGet, interceptPost } from '../support/utils/intercept';
+import { login } from './auth-forms';
 import * as loginHelper from './login';
 import {
   navigateToAMyAccountPage,
   navigateToCategory,
   navigateToHomepage,
+  waitForPage,
 } from './navigation';
 
 export function listenForAuthenticationRequest(): string {
@@ -88,6 +91,13 @@ export function startCustomerEmulation(customer): void {
   cy.get('cx-customer-emulation').should('be.visible');
 }
 
+export function loginCustomerInStorefront(customer) {
+  const authRequest = listenForAuthenticationRequest();
+
+  login(customer.email, customer.password);
+  cy.wait(authRequest).its('response.statusCode').should('eq', 200);
+}
+
 export function agentSignOut() {
   const tokenRevocationAlias = loginHelper.listenForTokenRevocationRequest();
   cy.get('button[title="Sign Out"]').click();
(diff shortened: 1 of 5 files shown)


projects/storefrontapp-e2e-cypress/cypress/helpers/asm.ts
```diff
@@ -175,4 +213,46 @@ export function testCustomerEmulation() {
     navigateToCategory('Brands', 'brands', false);
     cy.get('cx-product-list-item').should('exist');
   });
+
+  it('should verify data changed by the agent as a customer', () => {
```
bgambocjaviniar, line 217: included here because I believe it's a core e2e. Without this part, how would we know it actually worked

projects/storefrontapp-e2e-cypress/cypress/helpers/update-profile.ts
```diff
@@ -1,13 +1,22 @@
...
-export function updateProfile() {
+export function updateProfile(user?: SampleUser) {
+  if (user) {
+    cy.get('[formcontrolname="firstName"]').should(
+      'have.value',
+      user.firstName
+    );
+    cy.get('[formcontrolname="lastName"]').should('have.value', user.lastName);
+  }
+
```
plabadie, line 19: Is this change present in this pr intentionally?  Doesn't look asm related.

projects/storefrontapp-e2e-cypress/cypress/helpers/asm.ts
```diff
@@ -93,24 +110,15 @@ export function assertCustomerIsSignedIn() {
...
-
 export function testCustomerEmulation() {
+  let customer: SampleUser;
+
   it('should test customer emulation', () => {
-    let customer = getSampleUser();
-    checkout.registerUser(false, customer);
+    checkout.visitHomePage();
+
+    customer = checkout.registerUser(false);
```
plabadie, line 119: Not within the scope go this PR, but I guess we should move checkout.visitHomePage() in navigation.ts at some point in the future.

Also, I realize we should make it clear by the name of the function if a `visit` statement will be used.  Perhaps `visit` vs `navigate` is how we distinguish the internal behaviour.

projects/storefrontapp-e2e-cypress/cypress/helpers/update-profile.ts
plabadie: Nevermind, I just saw that updateProfile is called from the asm test.

plabadie (approved):

PR #15951, opened 2022-07-05
chore: fix a small bug in saved cart

WeizhengSap:
ticket cxspa-638


feature-libs/cart/saved-cart/core/store/effects/saved-cart.effect.ts
```diff
@@ -125,14 +125,13 @@ export class SavedCartEffects {
               },
               GlobalMessageType.MSG_TYPE_CONFIRMATION
             );
-
             return [
               ...actions,
-              new CartActions.SetActiveCartId(cartId),
               new CartActions.LoadCartSuccess({
```
WeizhengSap, line 130: To reviewer: We should not set `active cart id` before we set `active cart data`, otherwise, there is a moment that the active cart data is incorrect. In this `restore saved cart` effect, we set `active cart id` first, it points to a `saved cart`. so, in this moment, the active cart becomes saved cart.

feature-libs/cart/base/core/store/effects/multi-cart.effect.ts
```diff
@@ -58,8 +58,22 @@ export class MultiCartEffects {
           }
           break;
         }
-        case CartActions.LOAD_CART_SUCCESS:
-        // point to `temp-${uuid}` when we are creating/merging cart
+        case CartActions.LOAD_CART_SUCCESS: {
+          if (action?.payload?.extraData?.active) {
```
WeizhengSap, line 62: To reviewer: In here, we make sure the active cart is not saved cart, in case the cart's `saveTime` is set in backend.

plabadie (approved):

PR #15975, opened 2022-07-07
test: Adding E2E test for checking that out of stock products do not have Add-to-Cart buttons for Product Carousel

justinlee01:


projects/storefrontapp-e2e-cypress/cypress/integration/regression/product-carousel/product-carousel.e2e-spec.ts
```diff
@@ -34,6 +37,92 @@ context('Product carousel', () => {
 
         cart.checkAddedToCartDialog();
       });
+
+      it('should not show for out-of-stock products', () => {
```
Zeyber, line 41: Works good. I would just make the title a little more descriptive. eg. 'should not show add to cart button for out-of-stock products'

Zeyber (changes requested):

projects/storefrontapp-e2e-cypress/cypress/integration/regression/product-carousel/product-carousel.e2e-spec.ts
justinlee01: I think the `describe` on line 25 provides enough context - so the full sentence is "Product Carousel Add to Cart should not show for out-of-stock products". But I can still change it, I'm indifferent.

Zeyber: Good point. Okay.

Zeyber (approved):

bullfroggy-sap (approved):

PR #15985, opened 2022-07-11
feat: Support for better specificity on page meta resolvers

bgambocjaviniar:
closes https://jira.tools.sap/browse/CXSPA-757

Gave more weight to when pageUid is taken into consideration. Kept the logic isolated as such with the resolvers and did not leak it to the applicable and page meta service. Basically, the scoring logic should be isolated within the resolvers as the current logic is iterating through the list of resolvers and matching the page *resolver that corresponds to the current page


projects/core/src/cms/page/page-meta.resolver.spec.ts
```diff
@@ -60,4 +64,18 @@ describe('PageMetaResolver', () => {
     service.pageTemplate = 'any-template';
     expect(service.getScore(mockContentPageWithTemplate)).toEqual(2);
   });
+
+  it('should score 3 for ContentPage with page template and page id', () => {
```
WeizhengSap: score 3?

projects/core/src/cms/page/page-meta.resolver.ts
```diff
@@ -33,6 +36,11 @@ export abstract class PageMetaResolver implements Applicable {
     if (this.pageTemplate) {
       score += page.template === this.pageTemplate ? 1 : -1;
     }
+
+    if (this.pageUid) {
+      score += page.pageId === this.pageUid ? 100 : -100;
```
WeizhengSap: why 100?

bgambocjaviniar: jus a random higher weight

a use case is like how Patrick phrased it 
```
I might have been about these kind cases, like theoretically:
 title A could match page type and template and get a score of 2,
while title B could get a uid match and get a score of 1. 
```

bgambocjaviniar: More weight is given to pageUid because we're specifically trying to target a specific page

Moerover, from what I wrote in the description
```
Gave more weight to when pageUid is taken into consideration. Kept the logic isolated as such with the resolvers and did not leak it to the applicable and page meta service. Basically, the scoring logic should be isolated within the resolvers as the current logic is iterating through the list of resolvers and matching the page *resolver that corresponds to the current page
```

The logic would have been easier to work with IF we can just say in the resolver to just use this resolver if it exist
But the logic is separate to how it iterates through every resolvers to find the matching ones

WeizhengSap (approved):

PR #15999, opened 2022-07-14
Feature/cxec-15439

MarcinGanczorz:
https://jira.tools.sap/browse/CXEC-15439

### DIFF
--- a/feature-libs/order/components/order-history/order-history.component.ts
+++ b/feature-libs/order/components/order-history/order-history.component.ts
@@ -1,5 +1,6 @@
 import { ChangeDetectionStrategy, Component, OnDestroy } from '@angular/core';
 import {
+  CmsOrderHistoryComponent,
   isNotUndefined,
   RoutingService,
   TranslationService,
@@ -10,6 +11,7 @@ import {
   OrderHistoryList,
   ReplenishmentOrderHistoryFacade,
 } from '@spartacus/order/root';
+import { CmsComponentData } from '@spartacus/storefront';
 import { combineLatest, Observable } from 'rxjs';
 import { filter, map, take, tap } from 'rxjs/operators';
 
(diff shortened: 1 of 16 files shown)


feature-libs/order/styles/components/order-history/_order-history.scss
```diff
@@ -22,7 +22,7 @@
     }
 
     td {
-      width: 25%;
+      width: auto;
```
I539321, line 25: Hmm, I think auto is the default for width, if that's the case here do you think we should drop it, or keep it for clarity?

feature-libs/order/components/order-history/order-history.component.ts
```diff
@@ -23,14 +25,26 @@ export class OrderHistoryComponent implements OnDestroy {
...
 
   private PAGE_SIZE = 5;
   sortType: string;
+  showUnitOrders = false;
+
+  showUnitOrders$: Observable<boolean | undefined> = this.component.data$.pipe(
+    map((data) => data.showUnitOrders),
+    tap((value) => {
+      if (value !== this.showUnitOrders) {
+        this.showUnitOrders = value ? value : false;
```
I539321, line 40: Hmm did you mean assignment (=) or comparison (==) here?

MarcinGanczorz: Not a bug, just set the value if it's set, false otherwise

feature-libs/order/styles/components/order-history/_order-history.scss
MarcinGanczorz: I'm not sure what is the proper way to do here.

feature-libs/order/components/order-history/order-history.component.ts
I539321: Then I'd suggest using the nullish coalescing operator (??) here

MarcinGanczorz:
No longer valid

PR #16001, opened 2022-07-18
chore: Review Deleted Api

bgambocjaviniar:
closes https://jira.tools.sap/browse/CXSPA-718

### DIFF
--- a/projects/schematics/src/migrations/5_0/rename-symbol/rename-symbol.ts
+++ b/projects/schematics/src/migrations/5_0/rename-symbol/rename-symbol.ts
@@ -214,6 +214,12 @@ export const GENERATED_RENAMED_SYMBOLS_DATA: RenamedSymbol[] = [
     newImportPath: '@spartacus/checkout/base/components',
   },
   {
+    previousNode: 'PaymentMethodComponent',
+    previousImportPath: '@spartacus/checkout/components',
+    newNode: 'CheckoutPaymentMethodComponent',
+    newImportPath: '@spartacus/checkout/base/components',
+  },
+  {
     previousNode: 'PaymentMethodModule',
     previousImportPath: '@spartacus/checkout/components',
     newNode: 'CheckoutPaymentMethodModule',
--- a/projects/schematics/src/migrations/5_0/rename-symbol/rename-symbol.ts
+++ b/projects/schematics/src/migrations/5_0/rename-symbol/rename-symbol.ts
@@ -232,6 +232,12 @@ export const GENERATED_RENAMED_SYMBOLS_DATA: RenamedSymbol[] = [
     newImportPath: '@spartacus/checkout/base/components',
   },
   {
+    previousNode: 'PlaceOrderComponent',
+    previousImportPath: '@spartacus/checkout/components',
+    newNode: 'CheckoutPlaceOrderComponent',
+    newImportPath: '@spartacus/checkout/base/components',
+  },
+  {
     previousNode: 'PlaceOrderModule',
     previousImportPath: '@spartacus/checkout/components',
     newNode: 'CheckoutPlaceOrderModule',
--- a/projects/schematics/src/migrations/5_0/rename-symbol/rename-symbol.ts
+++ b/projects/schematics/src/migrations/5_0/rename-symbol/rename-symbol.ts
@@ -166,6 +166,12 @@ export const GENERATED_RENAMED_SYMBOLS_DATA: RenamedSymbol[] = [
     newImportPath: '@spartacus/checkout/base/components',
   },
   {
+    previousNode: 'GuestRegisterFormComponent',
+    previousImportPath: '@spartacus/checkout/components',
+    newNode: 'OrderGuestRegisterFormComponent',
+    newImportPath: '@spartacus/order/components',
+  },
+  {
     previousNode: 'NotCheckoutAuthGuard',
     previousImportPath: '@spartacus/checkout/components',
     newImportPath: '@spartacus/checkout/base/components',
--- a/projects/schematics/src/migrations/5_0/rename-symbol/rename-symbol.ts
+++ b/projects/schematics/src/migrations/5_0/rename-symbol/rename-symbol.ts
@@ -155,6 +155,12 @@ export const GENERATED_RENAMED_SYMBOLS_DATA: RenamedSymbol[] = [
     newImportPath: '@spartacus/checkout/base/components',
   },
   {
+    previousNode: 'DeliveryModeModule',
+    previousImportPath: '@spartacus/checkout/components',
+    newNode: 'CheckoutDeliveryModeModule',
+    newImportPath: '@spartacus/checkout/base/components',
+  },
+  {
     previousNode: 'ExpressCheckoutService',
     previousImportPath: '@spartacus/checkout/components',
     newImportPath: '@spartacus/checkout/base/components',
--- a/projects/schematics/src/migrations/5_0/rename-symbol/rename-symbol.ts
+++ b/projects/schematics/src/migrations/5_0/rename-symbol/rename-symbol.ts
@@ -226,6 +226,12 @@ export const GENERATED_RENAMED_SYMBOLS_DATA: RenamedSymbol[] = [
     newImportPath: '@spartacus/checkout/base/components',
   },
   {
+    previousNode: 'CARD_TYPE_NORMALIZER',
+    previousImportPath: '@spartacus/checkout/core',
+    newNode: 'PAYMENT_CARD_TYPE_NORMALIZER',
+    newImportPath: '@spartacus/checkout/base/core',
+  },
+  {
     previousNode: 'CHECKOUT_FEATURE',
     previousImportPath: '@spartacus/checkout/core',
     newImportPath: '@spartacus/checkout/base/root',
--- a/projects/schematics/src/migrations/5_0/rename-symbol/rename-symbol.ts
+++ b/projects/schematics/src/migrations/5_0/rename-symbol/rename-symbol.ts
@@ -220,6 +220,12 @@ export const GENERATED_RENAMED_SYMBOLS_DATA: RenamedSymbol[] = [
     newImportPath: '@spartacus/checkout/base/components',
   },
   {
+    previousNode: 'PaymentTypeModule',
+    previousImportPath: '@spartacus/checkout/components',
+    newNode: 'CheckoutPaymentTypeModule',
+    newImportPath: '@spartacus/checkout/base/components',
+  },
+  {
     previousNode: 'CHECKOUT_FEATURE',
     previousImportPath: '@spartacus/checkout/core',
     newImportPath: '@spartacus/checkout/base/root',
--- a/projects/schematics/src/migrations/5_0/rename-symbol/rename-symbol.ts
+++ b/projects/schematics/src/migrations/5_0/rename-symbol/rename-symbol.ts
@@ -214,6 +214,12 @@ export const GENERATED_RENAMED_SYMBOLS_DATA: RenamedSymbol[] = [
     newImportPath: '@spartacus/checkout/base/components',
   },
   {
+    previousNode: 'PaymentTypeComponent',
+    previousImportPath: '@spartacus/checkout/components',
+    newNode: 'CheckoutPaymentTypeComponent',
+    newImportPath: '@spartacus/checkout/base/components',
+  },
+  {
     previousNode: 'CHECKOUT_FEATURE',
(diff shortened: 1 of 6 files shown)


tools/breaking-changes/data/breaking-changes.json
```diff
@@ -8627,7 +8627,7 @@
         "apiElementKind": "Class",
         "entryPoint": "@spartacus/checkout/components",
         "deletedComment": "// TODO:Spartacus - Method 'handleUserRole' was removed from Class 'CheckoutAuthGuard'.",
-        "migrationComment": "",
+        "migrationComment": "It is not used in the base checkout entrypoint (b2c) as checkout has been decoupled into base (b2c), b2b, and replenishment. The method is found when b2b checkout is used instead, which uses the new class 'CheckoutB2BAuthGuard.",
```
znikola, line 8630:
```suggestion
        "migrationComment": "It is not used in the base checkout entrypoint (b2c) as checkout has been decoupled into base (b2c), b2b, and scheduled replenishment. The method is found when b2b checkout is used instead, which uses the new class 'CheckoutB2BAuthGuard.",
```

znikola (changes requested):

tools/breaking-changes/data/breaking-changes.json
```diff
@@ -8890,9 +8890,9 @@
         "returnType": "Observable<string>"
       }
     ],
-    "isDeleted": true,
-    "deletedComment": "Class CheckoutDetailsService has been removed and is no longer part of the public API.",
-    "migrationComment": "",
+    "isMoved": true,
+    "newName": "CheckoutQueryService",
+    "newEntryPoint": "@spartacus/checkout/base/core",
```
znikola, line 8895: Nice!

projects/schematics/src/migrations/5_0/rename-symbol/rename-symbol.ts
```diff
@@ -62,6 +62,12 @@ export const GENERATED_RENAMED_SYMBOLS_DATA: RenamedSymbol[] = [
     previousImportPath: '@spartacus/checkout/components',
     newImportPath: '@spartacus/checkout/base/components',
   },
+  {
+    previousNode: 'CheckoutDetailsService',
+    previousImportPath: '@spartacus/checkout/components',
+    newNode: 'CheckoutQueryService',
+    newImportPath: '@spartacus/checkout/base/core',
+  },
```
znikola, line 70: Does this handle the case when we renamed the property? 
E.g., previously we had

```ts
constructor(..., protected checkoutDetailsService: CheckoutDetailsService, ...) {}
```

but now we changed the type _and_ the property name: 

```ts
constructor(..., protected checkoutQueryService: CheckoutQueryService, ...) {}
```

tools/breaking-changes/data/breaking-changes.json
```diff
@@ -11614,7 +11614,7 @@
         "apiElementKind": "Class",
         "entryPoint": "@spartacus/checkout/components",
         "deletedComment": "// TODO:Spartacus - Property 'checkoutDeliveryService' was removed from Class 'ExpressCheckoutService'.",
-        "migrationComment": "",
+        "migrationComment": "'CheckoutDeliveryService' has been splitted and renamed to 'CheckoutDeliveryAddressFacde' and 'checkoutDeliveryModesFacade'. Use 'checkoutDeliveryAdressFacade' and 'checkoutDeliveryModesFacade' instead.",
```
znikola, line 11617:
```suggestion
        "migrationComment": "'CheckoutDeliveryService' has been splitted and renamed to 'CheckoutDeliveryAddressFacde' and 'CheckoutDeliveryModesFacade'. Use 'checkoutDeliveryAdressFacade' and 'checkoutDeliveryModesFacade' instead.",
```

tools/breaking-changes/data/breaking-changes.json
```diff
@@ -12198,9 +12198,10 @@
         "type": "PromotionLocation"
       }
     ],
-    "isDeleted": true,
-    "deletedComment": "Class OrderConfirmationItemsComponent has been removed and is no longer part of the public API.",
-    "migrationComment": "",
+    "isMoved": true,
+    "newName": "",
+    "newEntryPoint": "@spartacus/order/components",
+    "newNamespace": "",
```
znikola, line 12204: Just curious, do we have specify this?

tools/breaking-changes/data/breaking-changes.json
```diff
@@ -14522,7 +14522,7 @@
     "type": "",
     "isDeleted": true,
     "deletedComment": "Variable CheckoutActions.ADD_DELIVERY_ADDRESS_FAIL has been removed and is no longer part of the public API.",
-    "migrationComment": "It is no used anymore. Checkout library no longer uses NgRX to dispatch actions. Please take a look at Commands & Queries documentation on how we handle state https://sap.github.io/spartacus-docs/commands-and-queries",
+    "migrationComment": "It is no used anymore. Checkout library no longer uses NgRX to dispatch actions. Please take a look at Commands & Queries documentation on how we handle state https://sap.github.io/spartacus-docs/commands-and-queries. You can leverage the creation of delivery address 'state' using the returned observable from 'createAndSetAddress' in CheckoutDeliveryAddressFacade.",
```
znikola, line 14525:
```suggestion
    "migrationComment": "It is no used anymore. Checkout library no longer uses NgRX to dispatch actions. Please take a look at Commands & Queries documentation on how we handle state https://sap.github.io/spartacus-docs/commands-and-queries. You can leverage RxJS' error handling when calling the 'createAndSetAddress()' from CheckoutDeliveryAddressFacade.",
```

tools/breaking-changes/data/breaking-changes.json
```diff
@@ -14544,7 +14544,7 @@
     "type": "",
     "isDeleted": true,
     "deletedComment": "Variable CheckoutActions.ADD_DELIVERY_ADDRESS_SUCCESS has been removed and is no longer part of the public API.",
-    "migrationComment": "",
+    "migrationComment": "It is no used anymore. Checkout library no longer uses NgRX to dispatch actions. Please take a look at Commands & Queries documentation on how we handle state https://sap.github.io/spartacus-docs/commands-and-queries. You can use the EventService to listen for 'CheckoutDeliveryAddressCreatedEvent', which gets fired when the address has been successfully added.",
```
znikola, line 14547: Nice.

tools/breaking-changes/data/breaking-changes.json
```diff
@@ -14566,7 +14566,7 @@
     "type": "",
     "isDeleted": true,
     "deletedComment": "Variable CheckoutActions.ADD_DELIVERY_ADDRESS has been removed and is no longer part of the public API.",
-    "migrationComment": "",
+    "migrationComment": "It is no used anymore. Checkout library no longer uses NgRX to dispatch actions. Please take a look at Commands & Queries documentation on how we handle state https://sap.github.io/spartacus-docs/commands-and-queries. You can leverage the creation of delivery address 'state' using the returned observable from 'createAndSetAddress' in CheckoutDeliveryAddressFacade.",
```
znikola, line 14569:
```suggestion
    "migrationComment": "It is no used anymore. Checkout library no longer uses NgRX to dispatch actions. Please take a look at Commands & Queries documentation on how we handle state https://sap.github.io/spartacus-docs/commands-and-queries. You can leverage the 'createAndSetAddress()' from CheckoutDeliveryAddressFacade.",
```

tools/breaking-changes/data/breaking-changes.json
```diff
@@ -14610,7 +14610,7 @@
     ],
     "isDeleted": true,
     "deletedComment": "Class CheckoutActions.AddDeliveryAddress has been removed and is no longer part of the public API.",
-    "migrationComment": "",
+    "migrationComment": "It is no used anymore. Checkout library no longer uses NgRX to dispatch actions. Please take a look at Commands & Queries documentation on how we handle state https://sap.github.io/spartacus-docs/commands-and-queries. You can leverage the creation of delivery address 'state' using the returned observable from 'createAndSetAddress' in CheckoutDeliveryAddressFacade.",
```
znikola, line 14613: Same as for the previous.
I'm not sure what you mean by "You can leverage the creation of delivery address 'state' using"? I think we can just tell customers to use the new API method, without specifying if it's a state or not?
And to be even more precise - it's a command, therefore not a state.

tools/breaking-changes/data/breaking-changes.json
```diff
@@ -14650,7 +14650,7 @@
     ],
     "isDeleted": true,
     "deletedComment": "Class CheckoutActions.AddDeliveryAddressFail has been removed and is no longer part of the public API.",
-    "migrationComment": "",
+    "migrationComment": "It is no used anymore. Checkout library no longer uses NgRX to dispatch actions. Please take a look at Commands & Queries documentation on how we handle state https://sap.github.io/spartacus-docs/commands-and-queries. You can leverage the creation of delivery address 'state' using the returned observable from 'createAndSetAddress' in CheckoutDeliveryAddressFacade.",
```
znikola, line 14653: Won't be flagging these anymore. It would be good to align on the previous comments.
If we decide to change these messages, I think you can just do a search to find them all.

tools/breaking-changes/data/breaking-changes.json
```diff
@@ -14718,7 +14718,7 @@
     ],
     "isDeleted": true,
     "deletedComment": "TypeAlias CheckoutActions.CardTypesAction has been removed and is no longer part of the public API.",
-    "migrationComment": "",
+    "migrationComment": "It is no used anymore. Checkout library no longer uses NgRX to dispatch actions. Please take a look at Commands & Queries documentation on how we handle state https://sap.github.io/spartacus-docs/commands-and-queries. You can use the query 'getPaymentCardTypesState' in CheckoutPaymentFacade to get the data and state for the card types.",
```
znikola, line 14721: I'm thinking is the first part too much?
This: `It is no used anymore. Checkout library no longer uses NgRX to dispatch actions. Please take a look at Commands & Queries documentation on how we handle state https://sap.github.io/spartacus-docs/commands-and-queries. ` might repeat way too often, and it might just create noise in the code.
I think it should be obvious the clients should read the c&q docs, which already mentions the fact the checkout doesn't use ngrx anymore.

tools/breaking-changes/data/breaking-changes.json
```diff
@@ -14844,7 +14844,7 @@
     ],
     "isDeleted": true,
     "deletedComment": "TypeAlias CheckoutActions.CheckoutAction has been removed and is no longer part of the public API.",
-    "migrationComment": "",
+    "migrationComment": "It is no used anymore. Checkout library no longer uses NgRX to dispatch actions. Please take a look at Commands & Queries documentation on how we handle state https://sap.github.io/spartacus-docs/commands-and-queries.",
```
znikola:
```suggestion
    "migrationComment": "It is not used anymore. Checkout library no longer uses NgRX to dispatch actions. Please take a look at Commands & Queries documentation on how we handle state https://sap.github.io/spartacus-docs/commands-and-queries.",
```

znikola (changes requested):

tools/breaking-changes/data/breaking-changes.json
bgambocjaviniar: nay, just template things haha if ever a newNameSpace was actually made

projects/schematics/src/migrations/5_0/rename-symbol/rename-symbol.ts
bgambocjaviniar: Looking at the source code for it, it does not unfortunately, but they'd have the imports changed to use the correct one :d 

---
all in all, these were one of my pain points, where I asked Patrick that we should make a mgiration comment available for all migrations

tools/breaking-changes/data/breaking-changes.json
bgambocjaviniar: yeye it was already fixed in one of the latest commits

tools/breaking-changes/data/breaking-changes.json
bgambocjaviniar: yeye I understand it's a command, but they can derive the state by leveraging RxJS using that command, but yeah how I wrote it is confusing :P

projects/schematics/src/migrations/5_0/rename-symbol/rename-symbol.ts
znikola: Yes, the imports would change. 
A minor problem is that they would have two properties with different names that do the same thing.

znikola (approved):

PR #16005, opened 2022-07-19
(CXEC-14761) feat: add UI for customer ticket details

Parthlakhani:
JIRA ticket: https://jira.tools.sap/browse/CXEC-14761

- UI for customer ticket details on top of the page support-ticket
- Changes in the card to display different colour of a text
- Component name change to follow same standards


feature-libs/customer-ticketing/components/customer-ticketing-details/customer-ticketing-details.module.ts
```diff
@@ -0,0 +1,28 @@
...
+import { CustomerTicketingDetailsComponent } from './customer-ticketing-details.component';
+
+@NgModule({
+  imports: [CommonModule, I18nModule, UrlModule, CardModule],
+  providers: [
+    provideDefaultConfig(<CmsConfig>{
+      cmsComponents: {
+        SupportTicketDetailsComponent: {
+          component: CustomerTicketingDetailsComponent,
+          gurds: [AuthGuard],
```
adolfobee: guards

Parthlakhani: Ohh yes!

feature-libs/customer-ticketing/components/customer-ticketing-details/customer-ticketing-details.component.html
```diff
@@ -0,0 +1,44 @@
+<ng-container *ngIf="ticketDetails$ | async as ticket">
+  <div class="cx-ticket-details">
+    <div class="container">
+      <div class="cx-details-card">
+        <cx-card
+          [content]="
+            prepareCardContent(ticket?.id, 'customerTicketing.ticketId') | async
```
khristian23: in "ticket?.id"  is the compiler asking you to check for null?  It shouldn't be since the container checks for ticketDetails$

feature-libs/customer-ticketing/components/customer-ticketing-details/customer-ticketing-details.component.ts
```diff
@@ -0,0 +1,71 @@
...
+  ): Observable<Card> {
+    return this.translation.translate(titleTranslation).pipe(
+      filter(() => Boolean(entity)),
+      map((textTitle) => ({
+        title: textTitle,
+        text: [entity],
+        textColor: this.getStatusColor(entity),
+      }))
+    );
+  }
```
khristian23: This seems to be a very repeatable code, Can they be extracted to a parent ticketing component for all the components in the ticketing feature to reuse?

feature-libs/customer-ticketing/components/customer-ticketing-details/customer-ticketing-details.component.ts
```diff
@@ -0,0 +1,71 @@
...
+
+  getStatusColor(status: string): string {
+    return status === 'Open' ? '#38871F' : status === 'Close' ? '#6C7079' : '';
```
khristian23: Please extract the colors to Constants.

nit: Can we change the color setting to a CSS class?  
Inline style is hard to be overwritten by CSS custom classes

khristian23 (changes requested):

feature-libs/customer-ticketing/components/customer-ticketing-details/customer-ticketing-details.component.html
Parthlakhani: yes, I believe we can remove it.

feature-libs/customer-ticketing/components/customer-ticketing-details/customer-ticketing-details.component.ts
Parthlakhani: This code is used to prepare the content for the Card component that we are using to show the ticket details header.
In customer ticketing, it will be only used here in customer ticket details.. So I think we can keep it here.

feature-libs/customer-ticketing/components/customer-ticketing-details/customer-ticketing-details.component.ts
Parthlakhani: yes, adding constants.
and, yes we can.. adding that as well.

projects/storefrontlib/shared/components/card/card.component.html
```diff
@@ -38,7 +38,7 @@
           <div
             *ngIf="!truncateText; else truncate"
             class="cx-card-label"
```
khristian23, line 40: The change by the customCss is OK, but it is still inline.. I was thinking in something like:
class="cx-card-label color_open"

next to:
.color-open {
  color: #38871F;
}
.color-closed {
  color: #6C7079;
}

Parthlakhani: Do you mean something like this?
https://github.com/SAP/spartacus/pull/16005/commits/6af961183d9297e75ee911b590a972d9b802d87d#diff-c3b9ae880d0032f2fc3b0428a589360e6364ff4d5a195f6bc17bfbaa7d3b364d

Parthlakhani: Because that was my initial implementation but than I changed it because, using the current implementation someone can just pass the color without any changes in the shared component or css.
in the older implementation, we need to add css in the shared components's [css](https://github.com/SAP/spartacus/pull/16005/commits/6af961183d9297e75ee911b590a972d9b802d87d#diff-d05258fec2870a02438f36e3338ce82b7c226fb568aac9812b126d570e31948f) file.

khristian23: OK, we can still use inline colors

khristian23 (approved):

feature-libs/customer-ticketing/root/customer-ticketing-constants.ts
```diff
@@ -0,0 +1,4 @@
+export const COLOR_OPEN = '#38871F';
```
bgambocjaviniar: instead of storing it in /root/somethin-constants.ts, what do you think about creating `feature-libs/something/root/model/customer-ticketing.model.ts` instead

Reason why I'm opting out of the something-constants approach is because this was usually for feature names or ngrx processes, but as you can find in the code base, most don't use these conventions and places them in a something.model.ts file instead.

inside this something.model.ts file, you can create `two enums` 

one for 

```
export const enum COLOR {
  CLOSE = 'etc'
  OPEN = 'etc'
}
```

and
```
export const enum STATUS {
 OPEN = 'etc',
 CLOSE = 'etc'
}
```

reason for the 'const' keyword for the enums is for extendability purposes for customers to augment in the future without much complications. Remember to export them accordingly and wrap it in a comment as we have a script that helps with augmentation. It is imperative that the augmentation comments are at the bottom of the file. An example can be found under `feature-libs/checkout/base/root/public_api.ts`

feature-libs/customer-ticketing/components/customer-ticketing-details/customer-ticketing-details.component.html
```diff
@@ -0,0 +1,44 @@
+<ng-container *ngIf="ticketDetails$ | async as ticket">
```
bgambocjaviniar, line 1: Just a reminder, the 'overview'  and 'chat' are not separate cms components? 

If yes, then it's fine having this customer-ticketing-details.component.html as I believe you'll just include statically the chat component in this tempalte right?

If not, then this file should be more decoupled in a way that this is only for the top part of cms component

feature-libs/customer-ticketing/components/customer-ticketing-details/customer-ticketing-details.component.spec.ts
```diff
@@ -0,0 +1,36 @@
...
+  });
+
+  beforeEach(() => {
+    fixture = TestBed.createComponent(CustomerTicketingDetailsComponent);
+    component = fixture.componentInstance;
+    fixture.detectChanges();
+  });
+
+  it('should create', () => {
+    expect(component).toBeTruthy();
```
bgambocjaviniar, line 34: I'm assuming the unit tests will be done later right?

feature-libs/customer-ticketing/components/customer-ticketing-details/customer-ticketing-details.component.ts
```diff
@@ -0,0 +1,81 @@
...
+import { filter, map } from 'rxjs/operators';
+
+@Component({
+  selector: 'cx-customer-ticketing-details',
+  templateUrl: './customer-ticketing-details.component.html',
+})
+export class CustomerTicketingDetailsComponent implements OnInit {
+  constructor(protected translation: TranslationService) {}
+
+  ticketDetails$ = of({
```
bgambocjaviniar, line 15: 👍 

no need to mock data now that we have a way forward on what to do to get the data from the web server

bgambocjaviniar: maybe it's best to work on the core part first than the UI now. What do you think? 

If no, then continue on as I don't want to say how to work haha. Just a suggestion

feature-libs/customer-ticketing/components/customer-ticketing-details/customer-ticketing-details.component.ts
```diff
@@ -0,0 +1,81 @@
...
+      },
+      {
+        author: 'Mark Rivers',
+        createdAt: '2022-06-22T14:37:15+0000',
+        message: 'A message to consider',
+      },
+    ],
+  });
+
+  ngOnInit(): void {}
```
bgambocjaviniar: Do we really need to hook into Angular's lifecycle hook for ngOnInit?

feature-libs/customer-ticketing/components/customer-ticketing-details/customer-ticketing-details.component.ts
```diff
@@ -0,0 +1,81 @@
...
+
+  prepareCardContent(
```
bgambocjaviniar, line 53: 👍

bgambocjaviniar: to be honest, I'm fine making it generic like this and re-using it for the other card contents, but let me put this on the table.

What if the customer wants to override the specific data that's showing for 'x' column in your 'overview' component because right now it's made that any changes applied to this function will be by the customer will be applied to all. For example, what if they don't want the title for the 2nd column, or the 1st, etc.

But let me know what you think. Again, we can keep this approach.

feature-libs/customer-ticketing/components/customer-ticketing-details/customer-ticketing-details.component.ts
```diff
@@ -0,0 +1,81 @@
...
+      filter(() => Boolean(entity)),
+      map((textTitle) => ({
+        title: textTitle,
+        text: [entity],
+        customCss: this.getStatusColor(entity),
+      }))
+    );
+  }
+
+  getStatusColor(status: string): { color: string } | null {
```
bgambocjaviniar: related to the constants file

I see you went with the styling approach, but what do you think about classes instead. I believe it's more configurable in a sense that a customer can override the class in the css level

If going with classes, maybe no need for the enums anymore

projects/storefrontlib/shared/components/card/card.component.ts
```diff
@@ -104,6 +105,10 @@ export class CardComponent implements OnInit {
     return (action as CardLinkAction).link !== undefined;
   }
 
+  getCustomStyle() {
+    return this.content?.customCss;
```
bgambocjaviniar: related to my comment under `  getStatusColor(status: string): { color: string } | null {`

instead of styling, maybe allow to attach classes instead of passing styles. What do you think?

Moreover, you should explicitly put the type for the function

bgambocjaviniar (changes requested):

bgambocjaviniar (changes requested):
By the way, not really relevant to this PR, but can you merge pull changes to develop and merge it to your epic.

Reason being there is a fix for 2205 / 2211 servers, where it broke PLP page in Spartacus

feature-libs/customer-ticketing/root/customer-ticketing-constants.ts
Parthlakhani: Got it!

feature-libs/customer-ticketing/components/customer-ticketing-details/customer-ticketing-details.component.html
Parthlakhani: yes, overview and the chat are separate components.
SupportTicketDetailsComponent and SupportTicketUpdateComponent respectively.
Maybe we need to revise the names.

feature-libs/customer-ticketing/components/customer-ticketing-details/customer-ticketing-details.component.spec.ts
Parthlakhani: Yes, along with the core implementation.

feature-libs/customer-ticketing/components/customer-ticketing-details/customer-ticketing-details.component.ts
Parthlakhani: yes yes, I also prefer working on core first, but the UI is 90% done including the messaging and reopen/close request.
So, we thought we can merge this changes into epic and start on the core now.

feature-libs/customer-ticketing/components/customer-ticketing-details/customer-ticketing-details.component.ts
Parthlakhani: Nope, removing.

feature-libs/customer-ticketing/components/customer-ticketing-details/customer-ticketing-details.component.ts
Parthlakhani: I agree. But, is it possible to add a condition on the text or title based on the title?
if yes, I know it might be little bit difficult. What do you suggest?

feature-libs/customer-ticketing/components/customer-ticketing-details/customer-ticketing-details.component.ts
Parthlakhani: > By the way, not really relevant to this PR, but can you merge pull changes to develop and merge it to your epic.
> 
> Reason being there is a fix for 2205 / 2211 servers, where it broke PLP page in Spartacus

Yup. Noted.

Parthlakhani:
> By the way, not really relevant to this PR, but can you merge pull changes to develop and merge it to your epic.
> 
> Reason being there is a fix for 2205 / 2211 servers, where it broke PLP page in Spartacus

Yup. Noted.

bgambocjaviniar (approved):

PR #16010, opened 2022-07-22
fix: Remove remaining Ng-bootstrap 

morganm58:
fixes CXSPA-759
https://jira.tools.sap/browse/CXSPA-759


morganm58:
NgbModule is also being imported into these files. Do these imports also need to be removed? And if so I would need some help refactoring the files.
cart-base-components.module.ts
Cart.module.d.ts
cart-shared.module.ts
Consignment-tracking.component.spec.ts
Modal-options.ts
Modal.directive.ts
Modal.service.ts
Tracking-events.component.d.ts
tracking-events.component.spec.ts

PR #16021, opened 2022-07-26
feat: The 'Save cart for later' button in <cx-add-to-saved-cart> is disabled if its current cart has no entries

justinlee01:
#Closes https://jira.tools.sap/browse/CXSPA-527


feature-libs/cart/saved-cart/components/add-to-saved-cart/add-to-saved-cart.component.html
```diff
@@ -12,6 +12,7 @@
       <span>{{ 'addToSavedCart.savedCarts' | cxTranslate }}</span>
     </button>
     <button
+      *ngIf="showSaveCartForLater(cart)"
```
Zeyber: It would be better if we use an observable here. Otherwise the function will keep being recalled.

Zeyber (changes requested):

feature-libs/cart/saved-cart/components/add-to-saved-cart/add-to-saved-cart.component.html
```diff
@@ -12,6 +12,7 @@
       <span>{{ 'addToSavedCart.savedCarts' | cxTranslate }}</span>
     </button>
     <button
+      *ngIf="showSaveCartForLater$ | async"
```
justinlee01: Does this need to be feature flagged for 5.1? It's a simple fix, just was wondering.

Zeyber: Perhaps we can disable this button instead of removing it from the template. It is not necessarily a breaking change in this case.

Zeyber (changes requested):

feature-libs/cart/saved-cart/components/add-to-saved-cart/add-to-saved-cart.component.html
justinlee01: My only concern is that the customer may think "saving the cart for later" is a possible thing to do with an empty cart, however they have not fulfilled some invisible condition to do so.
Nevertheless I will update the PR with the proposed change. Perhaps my concern is minor.

feature-libs/cart/saved-cart/components/add-to-saved-cart/add-to-saved-cart.component.spec.ts
```diff
@@ -106,6 +108,75 @@ describe('AddToSavedCartComponent', () => {
     );
   });
 
+  it("should show the 'Save cart for later' button", (done) => {
```
Zeyber: Semantaics, but this should be `enable/disable` now instead of `show/hide`.

Zeyber (changes requested):
I think it could be a good call to add/modify e2e tests to make sure the user cannot save the cart in the case of "empty cart" as this is a UI improvement.

feature-libs/cart/saved-cart/components/add-to-saved-cart/add-to-saved-cart.component.spec.ts
justinlee01: Ah, slipped my mind. Thank you.

justinlee01:
@Zeyber I have spent the last few days trying to add an E2E test for this feature. I will explain the abnormal difficulty.

I have implementing this test plan:
1. Add a product to cart.
2. Go to cart. Check that "Save cart for later" is enabled.
3. Save the product for later. Check that "Save cart for later" is disabled.

When executing step 3, the code is trying to update the saved cart, which does not exist yet. Because of NgRX state management, this creates a malformed HTTP request causing the program to terminate early. I have not been able to reproduce it outside of Cypress.

I can continue pursuing this if it is absolutely necessary. A possible solution is to simply not allow the program to query for undefined saved carts.

justinlee01:
Finally found the issue. When the application loads, it also loads the selective (saved) cart.

Because Cypress is so quick, it creates a race hazard where UI is loaded and interacted with, but the saved cart has not been loaded yet, thus causing issues. The solution is to use Angular's router to navigate through routes instead of using `cy.visit` which triggers the application's loading code again.

Zeyber (changes requested):
I have found this test to fail on my local environment. Should something be done for setup?
image

Also, this should really be a part of `save-for-later` e2es rather than the main `cart` e2es as it is an additional feature.

justinlee01:
@Zeyber I'm not sure what's missing from the set-up; if you can share the full URL of the request that is responded with 400, that may help. That request + error looks familiar.

projects/storefrontapp-e2e-cypress/cypress/integration/regression/cart/cart.e2e-spec.ts
```diff
@@ -14,7 +14,7 @@ describe('Cart', () => {
     });
 
     context('Registered user', () => {
-      before(() => {
+      beforeEach(() => {
```
Zeyber: Leave as is?

Zeyber (changes requested):

Zeyber (approved):

kimhw0630 (approved):

PR #16024, opened 2022-07-27
feat: Customer ticketing

Parthlakhani:
This PR contains changes for Customer Ticketing feature:
[Ticketing Menu and Tickets List view](https://jira.tools.sap/browse/CXEC-9977)
[Create a Support Ticket](https://jira.tools.sap/browse/CXEC-9980)
[Support Ticket Details](https://jira.tools.sap/browse/CXEC-9981)

Following tasks are taken are of:
- [x] E2E tests
- [x] Browser compatibility
- [x] Accessibility validation
- [x] Running schematics from Customer POV
- [x] Handover of documentation
- [x] Threat modelling


.env-cmdrc
```diff
@@ -1,12 +1,12 @@
 {
   "dev": {
-    "CX_BASE_URL": "https://40.76.109.9:9002"
+    "CX_BASE_URL": "https://spartacus-dev0.eastus.cloudapp.azure.com:9002"
```
bgambocjaviniar: don't forget to revert before merge

giancorderoortiz:
Pend sec review. please DO NOT merge.

ci-scripts/unit-tests.sh
```diff
@@ -79,3 +79,11 @@ yarn --cwd feature-libs/tracking run test:schematics --coverage
 echo "Running unit tests and code coverage for schematics library"
 
 yarn --cwd projects/schematics run test --runInBand --coverage
+
+echo "Running schematics unit tests and code coverage for customer-ticketing"
```
bgambocjaviniar: nitpicking:

- for the sake of consistency in this file, make it run the unit test before schematics as done as all others with the exception of schematics lib.

docs/libs/creating-lib.md
```diff
@@ -206,6 +213,8 @@ Use the following template:
     "strictInjectionParameters": true,
     "enableResourceInlining": true,
     "strictTemplates": true,
+    "strictInputAccessModifiers": true,
+    "strictTemplates": true,
```
bgambocjaviniar: duplicate to line 215

docs/libs/creating-lib.md
```diff
@@ -59,33 +59,38 @@ Just copy paste the following and and make sure to rename `TODO:` to you lib's n
 module.exports = function (config) {
```
bgambocjaviniar, line 59: thank you for contributing to this doc 🙇

feature-libs/customer-ticketing/.release-it.json
```diff
@@ -0,0 +1,36 @@
+{
```
bgambocjaviniar: we no longer use release-it, please delete this file. This file for this feature wasn't hooked also, so there's nothing else to change for it

feature-libs/customer-ticketing/_index.scss
```diff
@@ -0,0 +1 @@
+@import './styles/index';
```
bgambocjaviniar, line 3: move the first two line in the file `feature-libs/customer-ticketing/styles/_index.scss` to the topmost of this file.

'convention' to styles has been normalized in this PR https://github.com/SAP/spartacus/pull/16323/files#diff-6b3120f56209c6d20d5c6c1d4d15378d100b263eaea837537de69c50b657fdfd

The gist for this convention is to help with style versioning as some styles do not get applied when the version number flag changes.

feature-libs/customer-ticketing/assets/public_api.ts
```diff
@@ -0,0 +1,8 @@
+/*
+ * SPDX-FileCopyrightText: 2022 SAP Spartacus team <spartacus-team@sap.com>
```
bgambocjaviniar: take other licensed headers and paste it over all the files that contains these duplicates. For example, copy the header comments from `feature-libs/asm/core/facade/asm.service.ts` and replace all instances in customer ticketing with the same duplicate.

1. find all 
2. replace with copied license

```
/*
 * SPDX-FileCopyrightText: 2023 SAP Spartacus team <spartacus-team@sap.com>
 *
 */
```

feature-libs/customer-ticketing/components/customer-ticketing-components.module.ts
```diff
@@ -0,0 +1,35 @@
...
+    ReactiveFormsModule,
+    FormErrorsModule,
+    CustomerTicketingDetailsModule,
+    CustomerTicketingCloseModule,
+    CustomerTicketingReopenModule,
+    CustomerTicketingListModule,
+    CustomerTicketingMessagesModule,
+    CustomerTicketingCreateModule,
+  ],
+  declarations: [],
```
bgambocjaviniar: remove declarations and exports. You don't have to state an empty array value for it.

feature-libs/customer-ticketing/components/details/customer-ticketing-close/customer-ticketing-close-dialog/customer-ticketing-close-dialog.component.ts
```diff
@@ -0,0 +1,70 @@
...
+    } else {
+      this.isDataLoading$.next(true);
+      this.subscription = this.customerTicketingFacade
+        .createTicketEvent(this.prepareTicketEvent())
+        .subscribe({
+          complete: () => {
+            this.isDataLoading$.next(false);
+            this.close('Ticket closed successfully');
+            this.routingService.go({ cxRoute: 'supportTickets' });
+            this.eventService.dispatch({}, GetTicketQueryResetEvent);
```
bgambocjaviniar: what do you think about delegating this dispatch to the event listener for the function `onTicketEventCreated`? 

reason being, you want to have the newly updated data from the backend when you use the create ticket api anyhow right?

feature-libs/customer-ticketing/components/details/customer-ticketing-close/customer-ticketing-close.module.ts
```diff
@@ -0,0 +1,54 @@
...
+          guards: [AuthGuard],
+        },
+      },
+    }),
+  ],
+  declarations: [
+    CustomerTicketingCloseComponent,
+    CustomerTicketingCloseDialogComponent,
+  ],
+  exports: [CustomerTicketingCloseComponent],
```
bgambocjaviniar: Don't we want to export `CustomerTicketingCloseDialogComponent` too

feature-libs/customer-ticketing/components/public_api.ts
```diff
@@ -0,0 +1,14 @@
+/*
+ * SPDX-FileCopyrightText: 2022 SAP Spartacus team <spartacus-team@sap.com>
+ * SPDX-FileCopyrightText: 2023 SAP Spartacus team <spartacus-team@sap.com>
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+export * from './customer-ticketing-components.module';
```
bgambocjaviniar, line 7: instead of manually putting all these at the root of the components folder, create a barrel file (index.ts) in each of the folders from components as well (list, details) and reference them instead.

so you would have 3 lines in this file

- details/index
- list/index
- shared/index

**you are missing `./list/customerticketing-create/index too`**

feature-libs/customer-ticketing/components/details/customer-ticketing-details/customer-ticketing-details.component.html
```diff
@@ -0,0 +1,51 @@
+<!-- TODO:CXEC-17553 (Brian / Parth) add spinner until the data is loaded  -->
```
bgambocjaviniar: I believe this comment should be removed as it's already done

feature-libs/customer-ticketing/components/details/customer-ticketing-details/customer-ticketing-details.component.ts
```diff
@@ -0,0 +1,89 @@
...
+  getStatusClass(id?: string): string {
+    if (id === STATUS.OPEN || id === STATUS.INPROCESS) {
+      return TEXT_COLOR_CLASS.GREEN;
+    } else if (id === STATUS.CLOSED) {
+      return TEXT_COLOR_CLASS.GREY;
+    }
+    return '';
+  }
+
+  protected reloadOnRedirection() {
```
bgambocjaviniar, line 69: good initiative. Will think about it because it's already too late to reload as you'll see the old state for a few seconds before it being updated due to the component initialization happens at the time it hits the page with the component, and not before.

feature-libs/customer-ticketing/components/details/customer-ticketing-details/customer-ticketing-details.component.html
```diff
@@ -0,0 +1,51 @@
...
+            prepareCardContent(ticket.id!, 'customerTicketing.ticketId') | async
+          "
+        ></cx-card>
+      </div>
+
+      <div class="cx-details-card">
+        <cx-card
+          [content]="
+            prepareCardContent(
+              (ticket.createdAt | cxDate: dateFormat) ?? dateFormat,
```
bgambocjaviniar, line 16: @Parthlakhani 

let's discuss these changes to solve strict typing of templates for line 17, 28 and 39. 

You should always avoid `! to assert that it will never be null/undefined`. Let's see the errors and discuss other ways to circumvent it.

bgambocjaviniar (changes requested):
first 33 files from top down out of 215 to go.
Should be faster the coming days as the entire morning was booked with meetings.

------
1 - unit tests are failing for 
- [CustomerTicketMessagesComponent](http://localhost:9876/context.html?spec=CustomerTicketMessagesComponent) > [should call uploadAttachment if the file is attached](http://localhost:9876/context.html?spec=CustomerTicketMessagesComponent%20should%20call%20uploadAttachment%20if%20the%20file%20is%20attached)
- [CustomerTicketMessagesComponent](http://localhost:9876/context.html?spec=CustomerTicketMessagesComponent) > [should not dispatch GetTicketQueryReloadEvent if the file is attached](http://localhost:9876/context.html?spec=CustomerTicketMessagesComponent%20should%20not%20dispatch%20GetTicketQueryReloadEvent%20if%20the%20file%20is%20attached)


2 - more of a question:
- what is associate to? Is it only the agent who can see what the user is associating the ticket to? I see that cart and orders are part of it, but are replenishment orders supposed to be part of it too as it isn't
- Should we hide associate to when it has nothing pull from (empty)?

3 - can I get the figma link as it's no longer available to me as it says this file may have been deleted.

feature-libs/customer-ticketing/components/details/customer-ticketing-messages/customer-ticketing-messages.component.ts
```diff
@@ -0,0 +1,125 @@
...
+            event.files instanceof FileList &&
+            event.files?.length &&
+            createdEvent.code
+          ) {
+            this.customerTicketingFacade.uploadAttachment(
+              event.files.item(0) as File,
+              createdEvent.code
+            );
+          } else {
+            this.eventService.dispatch({}, GetTicketQueryReloadEvent);
```
bgambocjaviniar: You can delegate this to the event listeners that would capture when you create a ticket event.
Basically, you already have a event for this specific api available `onTicketEventCreated`. You just delegate the reload there. It will basically say "Upon creation of any ticket event, I'd like to reload the ticket details"

bgambocjaviniar: In line 57 for `this.customerTicketingFacade.uploadAttachment(`, I noticed inside the service, you create a side effect to reload directly. I would say it's much better if you create a new event like `TicketUploadAttachmentSetEvent, where customers or you can hook into it to do 'xyz' things like when that event gets called, you reload the ticket details.

feature-libs/customer-ticketing/components/details/customer-ticketing-reopen/customer-ticketing-reopen-dialog/customer-ticketing-reopen-dialog.component.ts
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  reopenRequest(): void {
+    if (this.form.invalid) {
+      this.form.markAllAsTouched();
+      FormUtils.deepUpdateValueAndValidity(this.form);
+    } else {
+      this.isDataLoading$.next(true);
+      this.subscription = this.customerTicketingFacade
+        .createTicketEvent(this.prepareTicketEvent())
```
bgambocjaviniar: There is some inconsistency of how you do your streams as found here https://github.com/SAP/spartacus/pull/16024/files#diff-e2a15bc368ecf4c03733899e76aa589c0144aca3d563638108560991a15ce6f3R51.

The one above, you created the side effect in the `subscribe`, while this one is in the `tap`.

The difference between them is that if the createTicketEvent fails to emit, it won't ever try to upload the attachment, whereas if it was in the tap before the subscribe, then it will always try to upload even if it fails as a side effect.

My assumption is that you want it to call this attachment if and only if the createTicketEvent `completes` so that we know it didn't fail

bgambocjaviniar: Another suggestion is that I noticed these side effects exist in multiple places, where it fires the createTicketEvent function.

What do you think of putting the logic together in createTicketEvent as it's just repeated throughout and you can pass in a flag/object as a second param that determines whether or not to upload the attachment with the given file and code

feature-libs/customer-ticketing/components/details/customer-ticketing-reopen/customer-ticketing-reopen.module.ts
```diff
@@ -0,0 +1,54 @@
...
+          guards: [AuthGuard],
+        },
+      },
+    }),
+  ],
+  declarations: [
+    CustomerTicketingReopenComponent,
+    CustomerTicketingReopenDialogComponent,
+  ],
+  exports: [CustomerTicketingReopenComponent],
```
bgambocjaviniar: I believe `CustomerTicketingReopenDialogComponent` should be exported as well

feature-libs/customer-ticketing/components/list/customer-ticketing-create/customer-ticketing-create-dialog/customer-ticketing-create-dialog.component.html
```diff
@@ -0,0 +1,170 @@
...
+            </option>
+          </select>
+
+          <cx-form-errors
+            [control]="form.get('ticketCategory')"
+          ></cx-form-errors>
+        </label>
+
+        <label
+          *ngIf="ticketAssociatedObjects$ | async as ticketAssociatedObjects"
```
bgambocjaviniar, line 68: As mentioned previously, should this be displayed when the associated objects are empty. Moreover, this input is optional, so I feel like we do not need to show this input when it's empty. What do you think? 

If that's the case, check for the length instead

feature-libs/customer-ticketing/components/list/customer-ticketing-create/customer-ticketing-create.module.ts
```diff
@@ -0,0 +1,37 @@
...
+    KeyboardFocusModule,
+    ReactiveFormsModule,
+    FormErrorsModule,
+    FileUploadModule,
+  ],
+  declarations: [
+    CustomerTicketingCreateComponent,
+    CustomerTicketingCreateDialogComponent,
+  ],
+  exports: [CustomerTicketingCreateComponent],
```
bgambocjaviniar: the dialog should be exported too right?

feature-libs/customer-ticketing/components/list/customer-ticketing-create/customer-ticketing-create-dialog/customer-ticketing-create-dialog.component.html
```diff
@@ -0,0 +1,170 @@
...
+          </p>
+        </label>
+
+        <label *ngIf="ticketCategories$ | async as ticketCategories">
+          <span class="cx-customer-ticket-label label-content">
+            {{ 'createCustomerTicket.category' | cxTranslate }}
+          </span>
+          <select
+            class="form-control"
+            (change)="setSelectedCategory(ticketCategories, $event)"
```
bgambocjaviniar: we can remove this event listener by following my next comments in regards to this function.

feature-libs/customer-ticketing/components/list/customer-ticketing-create/customer-ticketing-create-dialog/customer-ticketing-create-dialog.component.ts
```diff
@@ -0,0 +1,150 @@
...
+      ticketAssociatedObjects[selectedAssociatedObjectIndex - 1];
+    this.form.controls.associatedTo.setValue(
+      this.selectedAssociatedObject.code
+    );
+  }
+
+  setSelectedCategory(ticketCategories: Category[], selectEvent: Event): void {
+    const selectedCategoryIndex = (selectEvent.target as HTMLSelectElement)
+      .selectedIndex;
+    this.selectedCategory = ticketCategories[selectedCategoryIndex - 1];
```
bgambocjaviniar: we can remove this function and simplify things.

1 - in the html for `ngValue="{{ category.id }}"`, let's change the syntax to the other alternative. Use `[ngValue]=category` instead.

2 - no reason to have this added complexity for selecting a category.
we can delete the event handler `(change)` and the function that binds with it.
We can make use of the `formControlName` to help bind the selected value so we don't have to 'hack' it this way, where we pass a list and re-iterate through it just to find the one that's selected.

For example,

in the html, append the formControlName and bind the 'object' as I mentioned above. Reason why I'm also opting for this is because we're created a formgroup, but we're not making the entire use of it. For example, the way it is now, we're creating the value for ticketCategory for the form on selection/change.

```
          <select class="form-control" formControlName="ticketCategory">
            <option value="" disabled selected>
              {{ 'createCustomerTicket.selectCategory' | cxTranslate }}
            </option>
            <option
              *ngFor="let category of ticketCategories"
              [ngValue]="category"
              [selected]="category.id === selectedCategory?.id"
            >
              {{ category.name }}
            </option>
          </select>

```

in the ts, we can remove the function in question (`setSelectedCategory`), and now in `getCreateTicketPayload` function, we can just pass in the form value as shown below. Basically, the ticketCategory will always have the selectedvalue due to the formControlName being binded to the select html tag

```
  protected getCreateTicketPayload(form: FormGroup): TicketStarter {
    return {
      message: form?.get('message')?.value,
      subject: form?.get('subject')?.value,
      associatedTo: this.selectedAssociatedObject,
      ticketCategory: form?.get('ticketCategory')?.value,
    };
  }
```

**NOTE:** this applies to the associated object and potentially other places in the customer ticketing codebase

feature-libs/customer-ticketing/components/list/customer-ticketing-create/customer-ticketing-create-dialog/customer-ticketing-create-dialog.component.ts
```diff
@@ -0,0 +1,150 @@
...
+  selectedAssociatedObject: AssociatedObject;
+
+  attachment: File;
+
+  protected getCreateTicketPayload(form: FormGroup): TicketStarter {
+    return {
+      message: form?.get('message')?.value,
+      subject: form?.get('subject')?.value,
+      associatedTo: this.selectedAssociatedObject,
+      ticketCategory: this.selectedCategory,
```
bgambocjaviniar: as mentioned, we can just do this after changing to what I mentioned

```suggestion
      ticketCategory: form?.get('ticketCategory')?.value,
```

Another point to change is to remove the parameter of the function. I don't see a reason passing the form when it's available to the `scope` of the class.

feature-libs/customer-ticketing/components/list/customer-ticketing-create/customer-ticketing-create-dialog/customer-ticketing-create-dialog.component.ts
```diff
@@ -0,0 +1,150 @@
...
+    this.customerTicketingFacade.getTicketAssociatedObjects();
+  subscription: Subscription;
+
+  @Input()
+  selectedCategory: Category;
+
+  @Input()
+  selectedAssociatedObject: AssociatedObject;
+
+  attachment: File;
```
bgambocjaviniar, line 39: any reason to have this as a class variable? I see it only being used when creating a ticket request, but other than that, I don't see its point. If there's no reason, please remove.

You can also keep the scope in the function if you want to just use it there.

feature-libs/customer-ticketing/components/list/customer-ticketing-list/customer-ticketing-list.component.html
```diff
@@ -0,0 +1,179 @@
+<ng-container *ngIf="tickets$ | async as ticketList; else spinner">
+  <span class="cx-ticketing-list-title">
+    <h2 class="cx-ticketing-list-title-text">
+      {{ 'customerTicketingList.requestTitle' | cxTranslate }}
+      ({{ ticketList.pagination?.totalResults || 0 }})
+    </h2>
+  </span>
+  <ng-container
+    *ngIf="
+      ticketList.tickets && ticketList.tickets.length > 0;
```
bgambocjaviniar: You can make use of optional chaining for catching undefined and null.

```suggestion
    ticketList?.tickets?.length > 0;
```

feature-libs/customer-ticketing/components/list/customer-ticketing-list/customer-ticketing-list.component.html
```diff
@@ -0,0 +1,179 @@
...
+          </th>
+          <th scope="col">{{ 'customerTicketing.createdOn' | cxTranslate }}</th>
+          <th scope="col">{{ 'customerTicketing.changedOn' | cxTranslate }}</th>
+          <th scope="col">{{ 'customerTicketing.status' | cxTranslate }}</th>
+        </tr>
+      </thead>
+      <tbody>
+        <tr
+          *ngFor="let ticket of ticketList.tickets"
+          (click)="goToTicketDetail(ticket.id!)"
```
bgambocjaviniar: I won't comment again on this, but let's fix this instead of using `!` to define that it really always be defined. We can get on a call with anyone and figure the strict typing solution.

feature-libs/customer-ticketing/components/list/customer-ticketing-list/customer-ticketing-list.component.ts
```diff
@@ -0,0 +1,100 @@
...
+  templateUrl: './customer-ticketing-list.component.html',
+})
+export class CustomerTicketingListComponent {
+  constructor(
+    protected customerTicketingFacade: CustomerTicketingFacade,
+    protected routing: RoutingService,
+    protected translation: TranslationService
+  ) {}
+  PAGE_SIZE = 5;
+  sortType: string = 'byId';
```
bgambocjaviniar: No reason to hardcode for the 'sort code' as that's given with the response of the api when getting the details of tickets.

for line 34 for `tickets$`, you can pipe and tap to assign to the `sortType` variable for any time the list of customer tickets is viewed/accessed.

feature-libs/customer-ticketing/components/list/customer-ticketing-list/customer-ticketing-list.component.ts
```diff
@@ -0,0 +1,100 @@
...
+  iconTypes = ICON_TYPE;
+  customerTicketsFlag: boolean = true;
```
bgambocjaviniar: I believe this is dead code, please remove.

feature-libs/customer-ticketing/components/list/customer-ticketing-list/customer-ticketing-list.component.ts
```diff
@@ -0,0 +1,100 @@
...
+  selector: 'cx-customer-ticketing-list',
+  templateUrl: './customer-ticketing-list.component.html',
+})
+export class CustomerTicketingListComponent {
+  constructor(
+    protected customerTicketingFacade: CustomerTicketingFacade,
+    protected routing: RoutingService,
+    protected translation: TranslationService
+  ) {}
+  PAGE_SIZE = 5;
```
bgambocjaviniar: Let's make this configurable. Please create a config so that customers can easily extend/update to a page size they want. 

We've gotten many complaints in the parts, where people wish it's configurable, so let's avoid creating more technical debt

bgambocjaviniar: maybe add to your existing config for an object called 'list', with a default page size property.

feature-libs/customer-ticketing/components/list/customer-ticketing-list/customer-ticketing-list.component.ts
```diff
@@ -0,0 +1,100 @@
...
+      currentPage: currentPage,
+      sortCode: sortCode,
+    };
+  }
+
+  private fetchTicketList(ticketListParams: {
+    sortCode: string;
+    currentPage: number;
+  }): void {
+    this.tickets$ = this.customerTicketingFacade.getTickets(
```
bgambocjaviniar, line 84: instead of 're-reassigning the stream', let's make use of the stream, and make use of its reactivity.
Only reason why it doesn't work without 're-assigning' is because you're not asking for the data to be reloaded. I see 2 options, but both options is more code than just re-assigning.

What do you think about 
1 - dispatching the reload
2 - or creating an event upon sorting, which you'll have an event listener that hooks on it, then dispatches the reload.

I prefer the 1st, but please hold off on this change. I would like to discuss this use case with a peer.

feature-libs/customer-ticketing/components/details/customer-ticketing-close/customer-ticketing-close-dialog/customer-ticketing-close-dialog.component.ts
```diff
@@ -0,0 +1,70 @@
...
+          },
+          error: () => {
+            this.close('Something went wrong while closing the ticket');
```
bgambocjaviniar: you need to call the `isDataLoading` and emit false as well on error. If you do not, and there is an error, then it will be forever loading.

In the example of checkout, we made them into functions so that it's configurable for the customers to do as they wish for the loading state. Reason being, anything inside the complete and error is all opinioated towards what we want done. If we  keep these things in complete and error, then the customer needs to extend and recreate there own stream.

This applies to alll complete and error on this PR as I won't throw comments on them all. Basically, you can move all 'logic' into a function for customers to hook into and change the behavior of things.

feature-libs/customer-ticketing/components/shared/customer-ticketing-dialog/default-customer-ticketing-form-layout.config.ts
```diff
@@ -0,0 +1,33 @@
+/*
+ * SPDX-FileCopyrightText: 2022 SAP Spartacus team <spartacus-team@sap.com>
+ * SPDX-FileCopyrightText: 2023 SAP Spartacus team <spartacus-team@sap.com>
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { DIALOG_TYPE, LayoutConfig } from '@spartacus/storefront';
+import { CustomerTicketingCloseDialogComponent } from '../../details/customer-ticketing-close';
```
bgambocjaviniar: we don't have to import the entire thing from the barrel as we want just the component. Please put the path to the component instead 

```suggestion
import { CustomerTicketingCloseDialogComponent } from '../../details/customer-ticketing-close/customer-ticketing-close.component';
```

again applicable to  others in the PR. Won't be throwing many comments for the same thing*

feature-libs/customer-ticketing/components/list/customer-ticketing-list/customer-ticketing-list.component.ts
```diff
@@ -0,0 +1,100 @@
...
+
+@Component({
+  selector: 'cx-customer-ticketing-list',
+  templateUrl: './customer-ticketing-list.component.html',
+})
+export class CustomerTicketingListComponent {
+  constructor(
+    protected customerTicketingFacade: CustomerTicketingFacade,
+    protected routing: RoutingService,
+    protected translation: TranslationService
```
bgambocjaviniar:
```suggestion
    protected routingService: RoutingService,
    protected translationService: TranslationService
```

feature-libs/customer-ticketing/core/customer-ticketing-core.module.ts
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { NgModule } from '@angular/core';
+import { HttpErrorHandler, PageMetaResolver } from '@spartacus/core';
+import { CustomerTicketingConnector } from './connectors';
+import { facadeProviders } from './facade/facade-providers';
+import { BadTicketRequestHandler } from './http-interceptors/handlers/bad-ticket-request.handler';
+import { CustomerTicketingPageMetaResolver } from './services';
```
bgambocjaviniar: put the full path to the resolver please

feature-libs/customer-ticketing/core/customer-ticketing-core.module.ts
```diff
@@ -0,0 +1,32 @@
+/*
+ * SPDX-FileCopyrightText: 2022 SAP Spartacus team <spartacus-team@sap.com>
+ * SPDX-FileCopyrightText: 2023 SAP Spartacus team <spartacus-team@sap.com>
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { NgModule } from '@angular/core';
+import { HttpErrorHandler, PageMetaResolver } from '@spartacus/core';
+import { CustomerTicketingConnector } from './connectors';
```
bgambocjaviniar: full path to the actual connector

bgambocjaviniar: not really a big problem with implicitness, but when it's explicit such as `./connectors/index`, we usually get errors when building the lib for production.

just say it once, use the path to the source file instead for this case.

the use case for the public_api is a difference story though. That can stay the same as it is.

feature-libs/customer-ticketing/core/facade/customer-ticketing.service.ts
```diff
@@ -0,0 +1,353 @@
...
+    return [GetTicketCategoryQueryResetEvent];
+  }
+  getTicketAssociatedObjectsQueryReloadEvents(): QueryNotifier[] {
+    return [GetTicketAssociatedObjectsQueryReloadEvent];
+  }
+  getTicketAssociatedObjectsQueryResetEvents(): QueryNotifier[] {
+    return [GetTicketAssociatedObjectsQueryResetEvent];
+  }
+  /**
+   * Returns the reload events for the getTicket query.
```
bgambocjaviniar: if you have comments for these event functions, wouldn't it be best that it's added to the others as well

feature-libs/customer-ticketing/core/facade/customer-ticketing.service.spec.ts
```diff
@@ -0,0 +1,423 @@
...
+import { CustomerTicketingService } from './customer-ticketing.service';
+import createSpy = jasmine.createSpy;
+
+const mockUserId = OCC_USER_ID_CURRENT;
+const mockRoutingParams = { ticketCode: '1' };
+const mockTicketDetails: TicketDetails = {
+  id: '1',
+  subject: 'MockTicket',
+};
+const mockTicketList: TicketList = {
```
bgambocjaviniar, line 31: not a big problem to me, but I just want to ask that was there a need to copy the entire response from the api to 'mock' as I don't see all these data being used in the end. My complaint is these mocks are long for no reason, but it's fine to keep :P

feature-libs/customer-ticketing/core/facade/customer-ticketing.service.ts
```diff
@@ -0,0 +1,353 @@
...
+} from '@spartacus/customer-ticketing/root';
+import { combineLatest, Observable } from 'rxjs';
+import {
+  distinctUntilChanged,
+  map,
+  switchMap,
+  take,
+  tap,
+} from 'rxjs/operators';
+import { CustomerTicketingConnector } from '../connectors';
```
bgambocjaviniar: path to source code please

feature-libs/customer-ticketing/core/facade/customer-ticketing.service.ts
```diff
@@ -0,0 +1,353 @@
...
+          this.customerTicketingConnector
+            .uploadAttachment(
+              customerId,
+              payload.ticketId ?? ticketId,
+              payload.eventCode,
+              payload.file
+            )
+            .pipe(
+              tap(() =>
+                this.eventService.dispatch({}, GetTicketQueryReloadEvent)
```
bgambocjaviniar: I mentioned in one of the previous comments, but create a new event that says something about a successful uploadattachment, and you can place that in the event listener so that it reloads when uploading.

feature-libs/customer-ticketing/core/http-interceptors/handlers/bad-ticket-request.handler.ts
```diff
@@ -0,0 +1,67 @@
...
+  HttpResponseStatus,
+  Priority,
+  RoutingService,
+  GlobalMessageService,
+} from '@spartacus/core';
+import { isTicketNotFoundError } from '../../utils/utils';
+@Injectable({
+  providedIn: 'root',
+})
+export class BadTicketRequestHandler extends HttpErrorHandler {
```
bgambocjaviniar: I don't think this is a BadRequestHandler, but a NotFoundRequestHandler

feature-libs/customer-ticketing/core/http-interceptors/handlers/bad-ticket-request.handler.spec.ts
```diff
@@ -0,0 +1,78 @@
...
+
+  it('should not handle response without errors', () => {
+    service.handleError(MockRequest, MockRandomResponse);
+    expect(globalMessageService.add).not.toHaveBeenCalled();
+  });
+
+  it('should handle ticket not found error', () => {
+    service.handleError(MockRequest, MockTicketNotFoundResponse);
+    expect(globalMessageService.add).toHaveBeenCalledWith(
+      { key: 'httpHandlers.ticketNotFound' },
```
bgambocjaviniar: what about test cases for both `NOT` that needs to actually be handled? Moreover, if it would redirect you as well.

feature-libs/customer-ticketing/core/public_api.ts
```diff
@@ -0,0 +1,11 @@
...
+ * SPDX-FileCopyrightText: 2022 SAP Spartacus team <spartacus-team@sap.com>
+ * SPDX-FileCopyrightText: 2023 SAP Spartacus team <spartacus-team@sap.com>
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+export * from './connectors/index';
+export * from './customer-ticketing-core.module';
+export * from './facade/index';
+export * from './services/index';
```
bgambocjaviniar, line 10: missing to export the `utils` and ` http-interceptors`

feature-libs/customer-ticketing/core/utils/utils.ts
```diff
@@ -0,0 +1,24 @@
...
+
+/**
+ * Check if the returned error is of type notFound.
+ */
+export function isTicketNotFoundError(error: ErrorModel): boolean {
+  return (
+    error.reason === OccHttpErrorReason.NOT_FOUND_ERROR &&
+    error.type === OccHttpErrorType.NOT_FOUND_ERROR &&
+    error.message !== undefined &&
+    error.message.toLowerCase().startsWith('ticket')
```
bgambocjaviniar: I don't think depending on the `message` would help define if it's for customer ticketing or not. Aren't these message configurable in the backend, and in hac and backoffice?

If that's the case , then I don't think we need this utils and just plug it directly in the handler.

What I do suggest though on how to tell if it's for ticketing is to hook into the `getRouterState` from routingservice to get the context of the page you're trying to access. If it's from the semantic route of  `supportTicketDetails`, then we return true or false together with verifying if it's really a not found error.

feature-libs/customer-ticketing/jest.schematics.config.js
```diff
@@ -0,0 +1,34 @@
+const { pathsToModuleNameMapper } = require('ts-jest');
```
bgambocjaviniar, line 1: update the jest configs accordingly to the changes here https://github.com/SAP/spartacus/pull/16532/files

Last month, changes towards schematics on all existing libs were updated.

What changed?

- new props in this file with new setup file
- setup-jest.ts has replaced test-jest
-added entry in tsconfig.lib to excluse the setup-jest file 
- changes towards tsconfig.schematics.json (do not extend nor set a ourDir prop as done in this PR). Basically, do not add the changes for tsconfig.schematics.json for the properties as found here https://github.com/SAP/spartacus/commit/f096901e7a1f5b2ada7f6011a4b55d891ecfcb95, where I removed 'additional' ones that broken our builds.

bgambocjaviniar (changes requested):
90/215 files

going to pass through customert-ticketing occ and below as they're sequential from topdown changes.

feature-libs/customer-ticketing/components/details/customer-ticketing-details/customer-ticketing-details.component.ts
Parthlakhani: This implementation came in to picture because when user opens a ticket, go back to ticket list and opens another ticket, the data was not reloading.

feature-libs/customer-ticketing/components/details/customer-ticketing-messages/customer-ticketing-messages.component.ts
Parthlakhani: I am dispatching this event here because, we don't want ticket data to reload if any attachment is attached. as we would end up ticket data twice with the attachment.

feature-libs/customer-ticketing/occ/adapters/occ-customer-ticketing.adapter.ts
```diff
@@ -0,0 +1,278 @@
...
+        tap((ticket) => ticket.ticketEvents?.reverse()),
+        this.converter.pipeable(CUSTOMER_TICKETING_DETAILS_NORMALIZER)
+      );
+  }
+
+  createTicket(
+    customerId: string,
+    ticket: TicketStarter
+  ): Observable<TicketStarter> {
+    ticket = this.converter.convert(
```
bgambocjaviniar: Any reason to place this 'convert'? 

Only reason is to 'serialize' the data before sending the payload to the backend. Therefore, the normalizer here isn't even a use case. When we say convert, we're referring to modifying the current payload before it's sent to the server in a way that it accepts it. That's what a serializer is for.

A normalizer, on the other hand, is for when we get the data from the server, but it doesn't meet our needs in terms of what it returns. So we 'normalize' the data to conform to our needs and modify the payload from the server for the code to use accordingly.

feature-libs/customer-ticketing/occ/adapters/occ-customer-ticketing.adapter.ts
```diff
@@ -0,0 +1,278 @@
...
+      .pipe(
+        catchError((errorResponse) => {
+          return throwError(normalizeHttpError(errorResponse));
+        }),
+        tap((ticket) => ticket.ticketEvents?.reverse()),
+        this.converter.pipeable(CUSTOMER_TICKETING_DETAILS_NORMALIZER)
+      );
+  }
+
+  createTicket(
```
bgambocjaviniar, line 106: nitpicking here, but swap the function `createTicket` with `getTicketEndpoint` to get function and helper function url builder as the current 'flow' in this file.

feature-libs/customer-ticketing/occ/adapters/occ-customer-ticketing.adapter.ts
```diff
@@ -0,0 +1,278 @@
...
+  }
+
+  createTicketEvent(
+    customerId: string,
+    ticketId: string,
+    ticketEvent: TicketEvent
+  ): Observable<TicketEvent> {
+    ticketEvent = this.converter.convert(
+      ticketEvent,
+      CUSTOMER_TICKETING_EVENT_NORMALIZER
```
bgambocjaviniar: same comment as mentioned for `createTicket`

feature-libs/customer-ticketing/occ/adapters/occ-customer-ticketing.adapter.ts
```diff
@@ -0,0 +1,278 @@
...
+    });
+  }
+
+  uploadAttachment(
+    customerId: string,
+    ticketId: string,
+    eventCode: string,
+    file: File
+  ): Observable<unknown> {
+    file = this.converter.convert(file, CUSTOMER_TICKETING_FILE_NORMALIZER);
```
bgambocjaviniar: same comment for `createTicket`

these aren't being 'serialized' in the end, so we can remove them.

bgambocjaviniar: Just to re-iterate

serializer = transforming payload that before it gets sent to server in a format it accepts
normalizer = transforming incoming payload from server to adhere to our data structure or to clean it up in a format that we want to use in the frontend

feature-libs/customer-ticketing/occ/adapters/occ-customer-ticketing.adapter.ts
```diff
@@ -0,0 +1,278 @@
...
+  }
+
+  getTicket(customerId: string, ticketId: string): Observable<TicketDetails> {
+    return this.http
+      .get<TicketDetails>(this.getTicketEndpoint(customerId, ticketId))
+      .pipe(
+        catchError((errorResponse) => {
+          return throwError(normalizeHttpError(errorResponse));
+        }),
+        tap((ticket) => ticket.ticketEvents?.reverse()),
```
bgambocjaviniar, line 92: this instance  here can be a use case to create the normalizer for this function.
What the function would do is mutate the list in a way you return the list with the reversed ticketEvents instead of directly doing it here.

Why?
- it's like an opinionated logic, where one will always have the reversed ticketEvents from the ticket. It can be fine, but we can always move this logic to the normalizer to 'normalize' the data into what we want, therefore they can just remove the normalizer or update that instead in order to keep the http client logic free and unopinionated

feature-libs/customer-ticketing/package.json
```diff
@@ -0,0 +1,46 @@
...
+    "@angular/forms": "^14.2.3",
+    "@angular/platform-browser-dynamic": "^14.2.3",
+    "@angular/router": "^14.2.3",
+    "@spartacus/cart": "4.1.0-next.0",
+    "@spartacus/core": "4.1.0-next.0",
+    "@spartacus/schematics": "4.1.0-next.0",
+    "@spartacus/storefront": "4.1.0-next.0",
+    "@spartacus/styles": "4.1.0-next.0",
+    "rxjs": "^6.6.0",
+    "zone.js": "~0.11.8"
```
bgambocjaviniar: I don't think `zone.js` and `@angular/platform-browser-dynamic` are required as peer deps as it is not in other libs

feature-libs/customer-ticketing/root/config/customer-ticketing-config.ts
```diff
@@ -0,0 +1,31 @@
...
+
+import { Injectable } from '@angular/core';
+import { Config } from '@spartacus/core';
+
+@Injectable({
+  providedIn: 'root',
+  useExisting: Config,
+})
+export abstract class CustomerTicketingConfig {
+  customerTicketing?: {
```
bgambocjaviniar, line 15: as mentioned in the past comments, make sure to include the defaultPageSize instead of hardcoding the  number in the component

feature-libs/customer-ticketing/root/config/default-support-ticket-routing-config.ts
```diff
@@ -0,0 +1,22 @@
+/*
```
bgambocjaviniar: I believe the name of the file should be `default-customer-ticketing-routing-config.ts` instead of support ticket.

This is the routing 'config' for the customer-ticketing feature. As it implies, update the variable as well.

feature-libs/customer-ticketing/root/events/customer-ticketing-event.listener.ts
```diff
@@ -0,0 +1,125 @@
...
+    protected globalMessageService: GlobalMessageService
+  ) {
+    this.onGetTicketQueryReload();
+    this.onGetTicketsQueryReload();
+    this.onLoginAndLogoutEvent();
+    this.onTicketCreatedEvent();
+    this.onTicketEventCreated();
+  }
+
+  onTicketCreatedEvent() {
```
bgambocjaviniar: let's mark this as protected as done with the others in this file

feature-libs/customer-ticketing/root/events/customer-ticketing.events.ts
```diff
@@ -0,0 +1,51 @@
+/*
+ * SPDX-FileCopyrightText: 2022 SAP Spartacus team <spartacus-team@sap.com>
+ * SPDX-FileCopyrightText: 2023 SAP Spartacus team <spartacus-team@sap.com>
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { CxEvent } from '@spartacus/core';
+
+export class GetTicketQueryResetEvent extends CxEvent {
```
bgambocjaviniar, line 9: not related to this line, but the way it is currently is fine with me, but if you want to group things to buckets, such as an abstract event for CustomerTicketListEvent and CustomerTicketDetailsEvent. 

All related 'event' should extend it, but again not necessary

feature-libs/customer-ticketing/root/feature-name.ts
```diff
@@ -0,0 +1,9 @@
+/*
+ * SPDX-FileCopyrightText: 2022 SAP Spartacus team <spartacus-team@sap.com>
+ * SPDX-FileCopyrightText: 2023 SAP Spartacus team <spartacus-team@sap.com>
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+export const CUSTOMER_TICKETING_FEATURE = 'customer-ticketing';
```
bgambocjaviniar: let's stay concise. On one hand you have the value as a `kebab` case, whereas the other is on `camel` case.

I believe throughout the project in spartacus, we tend to go for camel case,so please update.

feature-libs/customer-ticketing/root/model/index.ts
```diff
@@ -0,0 +1,10 @@
+/*
+ * SPDX-FileCopyrightText: 2022 SAP Spartacus team <spartacus-team@sap.com>
+ * SPDX-FileCopyrightText: 2023 SAP Spartacus team <spartacus-team@sap.com>
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+export * from './augmented-core.model';
+export * from './customer-ticketing.model';
+export * from './augmented-core.model';
```
bgambocjaviniar: delete this entry at line 10 as it is just a duplicate.

feature-libs/customer-ticketing/root/public_api.ts
```diff
@@ -0,0 +1,22 @@
...
+export * from './events/index';
+export * from './facade/index';
+export * from './feature-name';
+/** AUGMENTABLE_TYPES_START */
+export {
+  STATUS,
+  STATUS_NAME,
+  TEXT_COLOR_CLASS,
+} from './model/customer-ticketing.model';
+export * from './model/index';
```
bgambocjaviniar: move this entry above just before the comment that starts that augmentable_types_start.
Augmenting should always be the last in the file, and it is being used by `tools/build-lib/augmented-types/index.ts` as our builder

bgambocjaviniar: make sure to not format the file so it doesn't jumble the imports afterwards

feature-libs/customer-ticketing/styles/components/_customer-ticketing-list.scss
```diff
@@ -0,0 +1,166 @@
...
+      min-width: 140px;
+    }
+  }
+
+  .cx-ticketing-list-id {
+    text-decoration: underline;
+  }
+
+  .cx-ticketing-list-value {
+    color: var(--cx-color-text);
```
bgambocjaviniar, line 99: from what we talked about with Parth, the id and subject are not red because there are no sign of color red for them

feature-libs/customer-ticketing/test-jest.ts
```diff
@@ -0,0 +1,18 @@
...
+ */
+
+import 'zone.js/fesm2015/zone-testing-bundle.min.js';
+import { getTestBed } from '@angular/core/testing';
+import {
+  BrowserDynamicTestingModule,
+  platformBrowserDynamicTesting,
+} from '@angular/platform-browser-dynamic/testing';
+
+getTestBed().initTestEnvironment(
```
bgambocjaviniar: reminder to check the comment about updating to proper jest with the new setup as done in this PR https://github.com/SAP/spartacus/commit/422e4de4229f03bd2bbf20809f8a052de5e6ad5c#

feature-libs/order/tsconfig.schematics.json
```diff
@@ -151,6 +151,24 @@
       "@spartacus/checkout/scheduled-replenishment/root": [
         "../../feature-libs/checkout/scheduled-replenishment/root/public_api"
       ],
+      "@spartacus/customer-ticketing/assets": [
+        "../../feature-libs/customer-ticketing/assets/public_api"
```
bgambocjaviniar, line 155: just to clarify, these were automatically generated by the config script right as done in other non-relevant customer ticketing files ?

projects/assets/src/translations/en/common.ts
```diff
@@ -151,6 +151,7 @@ export const common = {
       'This coupon has exceeded the number of times it can be used',
     unknownIdentifier: 'Item not found.',
     otherCartErrors: 'Cart errors occurred.',
+    ticketNotFound: 'Ticket not found.',
```
bgambocjaviniar: I don't think this is a common key/value that would get thrown. Wouldn't you agree that it is specific to customer ticketing as a 'ticket'? 

What do you think moving it there specifically in `customerTicketingDetails`

bgambocjaviniar (changes requested):

projects/schematics/src/shared/lib-configs/customer-ticketing-schematics-config.ts
```diff
@@ -0,0 +1,56 @@
...
+  },
+  i18n: {
+    resources: CUSTOMER_TICKETING_TRANSLATIONS,
+    chunks: CUSTOMER_TICKETING_TRANSLATION_CHUNKS_CONFIG,
+    importPath: SPARTACUS_CUSTOMER_TICKETING_ASSETS,
+  },
+  styles: {
+    scssFileName: CUSTOMER_TICKETING_SCSS_FILE_NAME,
+    importStyle: SPARTACUS_CUSTOMER_TICKETING,
+  },
```
bgambocjaviniar, line 55: thinking out loud, but this feature is not 'strictly' dependent on cart or order right? From the top of my head, I don't think so because you don't need orders or cart to function.

However, this feature requires someone to login to access the 'feature'. in this case, we want to include a `dependencyFeatures` property that says to have `USER_PROFILE_FEATURE_NAME`. Basically, our schematics, will include the user profile library before installing customer ticketing

projects/schematics/src/shared/schematics-config-mappings.ts
```diff
@@ -113,6 +114,7 @@ export const SCHEMATICS_CONFIGS: SchematicConfig[] = [
   EPD_SCHEMATICS_CONFIG,
 
   S4OM_SCHEMATICS_CONFIG,
+  CUSTOMER_TICKETING_SCHEMATICS_CONFIG,
```
bgambocjaviniar: put this above the comment for `// integration libraries start` as these aren't integration libs like third party features (iss / epd / etc)

projects/storefrontapp-e2e-cypress/cypress.ci.json
```diff
@@ -11,7 +11,7 @@
   },
   "videoUploadOnPasses": false,
   "env": {
-    "API_URL": "https://20.83.184.244:9002",
+    "API_URL": "https://spartacus-dev0.eastus.cloudapp.azure.com:9002",
```
bgambocjaviniar: don't forget to remove before merging. Moreover, I mentioned earlier today to update the sample data so that we can merge it in the default branch `release/5.x` of the sample data repo.

Afterwards, we will deploy it to the main dev and ci servers right away.

projects/storefrontapp-e2e-cypress/cypress/helpers/customer-ticketing/customer-ticketing-helpers/create-tickets.ts
```diff
@@ -0,0 +1,119 @@
+/*
+ * SPDX-FileCopyrightText: 2022 SAP Spartacus team <spartacus-team@sap.com>
+ * SPDX-FileCopyrightText: 2023 SAP Spartacus team <spartacus-team@sap.com>
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import * as commons from './customer-ticketing-commons';
```
bgambocjaviniar: If you're going to import everything, you don't have to 're-import for TestTicketDetails, TICKET_MESSAGE_MAX_LENGTH, TICKET_SUBJECT_MAX_LENGTH`

projects/storefrontapp-e2e-cypress/cypress/integration/regression/customer-ticketing/create-ticket.e2e-spec.ts
```diff
@@ -0,0 +1,204 @@
+import { viewportContext } from '../../../helpers/viewport-context';
+import * as customerTicketing from '../../../helpers/customer-ticketing/customer-ticketing';
```
bgambocjaviniar, line 1: same case as the previous comment. No need to re-import `TestTicketDetails` and `TestCategory` if you are already importing everything

projects/storefrontapp-e2e-cypress/cypress/helpers/customer-ticketing/customer-ticketing.ts
```diff
@@ -0,0 +1,11 @@
+/*
```
bgambocjaviniar, line 1: random mention:

I like how this was setup for e2e :) 

Thank you for this as we were going to fix our 'everywhere' structure that people has been putting in the past years into 'domain' specific. This will help with choosing e2e runs and stuff

bgambocjaviniar: Another mention:

very nice e2es. I haven't thoroughly looked at them  as I would want the server to be updated to test the repeatability, but overall, it is very nice. 
1 - idempotent (not dependent on previous test runs)
2 - organized (not like any others I've seen so far - this is exceptional :) )


one thing I would like to say though is to drop the `viewportContext` so that it will just run for desktop. Test are long which is not really a con, but I don't think there are benefits for testing mobile for this feature as there's no significant changes that would hinder the feature 'unusable' on mobile I believe.

projects/storefrontlib/shared/components/card/card.component.html
```diff
@@ -35,14 +35,19 @@
...
+            [ngClass]="content.customClass"
+          >
             {{ line }}
           </div>
 
           <ng-template #truncate>
             <cx-truncate-text-popover
               [charactersLimit]="charactersLimit"
               [content]="line"
+              [customClass]="content.customClass"
```
bgambocjaviniar, line 50: Something for me to look into

I don't think this is a breaking change, but if it is, you would need to create duplicates and wrap them in feature flags. For example, an html snippet taken from `feature-libs/cart/quick-order/components/quick-order/quick-order.component.html` 

changing the html semantic from h3 to h2 is a breaking change so we want to show the h3 as usual for 5.0 and below while it only shows the h2 for 5.1 and above.

```
      <h3 *cxFeatureLevel="'!5.1'">
        {{ 'quickOrderList.header' | cxTranslate }}
      </h3>
      <h2 *cxFeatureLevel="'5.1'">
        {{ 'quickOrderList.header' | cxTranslate }}
      </h2>
```

projects/storefrontlib/shared/components/chat-messaging/messaging/messaging.component.ts
```diff
@@ -0,0 +1,237 @@
+/*
```
bgambocjaviniar, line 1: for me : 

Brian double check later

projects/storefrontlib/shared/components/form/file-upload/file-upload.component.html
```diff
@@ -7,13 +7,22 @@
...
 <button
+  *ngIf="!customButton"
   [attr.aria-label]="'common.selectFile' | cxTranslate"
   class="btn btn-action"
   type="button"
   (click)="fileInput.click()"
 >
   <ng-content></ng-content>
 </button>
+<!-- TODO(Brian-Parth): CXEC-17196 -->
```
bgambocjaviniar, line 18: Can you remind me why we put this comment here?

There's no context in the jira issue

projects/storefrontstyles/scss/components/content/navigation/_breadcrumb.scss
```diff
@@ -38,7 +38,7 @@ html[dir='rtl'] cx-breadcrumb nav span:not(:last-child):after {
       margin: 0;
 
       li {
-        display: inline;
+        display: flex;
```
bgambocjaviniar, line 41: as mentioned, I won't take a look at styling since Mikel went over it and made changes for this lbirary.

bgambocjaviniar (changes requested):

feature-libs/customer-ticketing/core/http-interceptors/handlers/bad-ticket-request.handler.ts
ishrarhussain: Thank you, @bgambocjaviniar .. The intention was to have a placeholder for handling other bad request error as well, including the `ticketNotFound` error.

bgambocjaviniar: Thanks for the comment! But I do recall sharing an example to derive it from, which is `feature-libs/cart/base/core/http-interceptors/handlers/bad-cart-request.handler.ts` 

In this sense, 'bad request' for cart makes sense since it's a handler specifically for `400 - bad requests`. Similarly, this class for the ticket handler is targeting specifically the `404 - not found error`. Currently, they  are 1 to 1 in terms of handler creation and what it handles

feature-libs/customer-ticketing/occ/adapters/occ-customer-ticketing.adapter.ts
Parthlakhani: Okay got it.
But, it's okay to leave this as it is, right?

feature-libs/customer-ticketing/root/config/customer-ticketing-config.ts
Parthlakhani: yup, that's done

feature-libs/customer-ticketing/styles/components/_customer-ticketing-list.scss
Parthlakhani: Yes! that is fine as it was changed by Mikel to support accessibility

feature-libs/order/tsconfig.schematics.json
Parthlakhani: Not sure, let's discuss

projects/schematics/src/shared/lib-configs/customer-ticketing-schematics-config.ts
Parthlakhani: Yup, right. not dependent on Cart or Order.

projects/storefrontlib/shared/components/form/file-upload/file-upload.component.html
Parthlakhani: It was a reminder to check file upload changes again as I modified the shared component.

projects/storefrontapp-e2e-cypress/cypress/helpers/customer-ticketing/customer-ticketing.ts
sushant-stha: Deleted viewport and pushed the change

feature-libs/customer-ticketing/core/http-interceptors/handlers/bad-ticket-request.handler.ts
ishrarhussain: Thank you so much @bgambocjaviniar for your comment.. But the backend throws other errors too including 400 for customer ticketing. It is true that we do not have any bad request other than ticket not found, but if we choose to handle other bad request errors, they can be included here in this same class, keeping the similarity with other modules, like the 'cart'. I can surely change the class name, if you still think so.

feature-libs/customer-ticketing/core/utils/utils.ts
ishrarhussain: Thank you, @bgambocjaviniar . Actually, the backend OCC extension for customer ticketing does not currently have configurable error messages. Hence, we went forward with this solution. But yes, I agree with your suggestion, as the error messages may change in future, and it is better not to depend on their text.

I've now added the changes, as per our discussion on slack. Thank you again so much for your help on the solution.

feature-libs/customer-ticketing/core/http-interceptors/handlers/bad-ticket-request.handler.ts
bgambocjaviniar: We will discuss this during our call for your recent push for fixing the error that determines whether or not it's a not found or not.



-----------
For simplicity, an example of what I'm mentioning is that these handlers have a single responsibility. Either it handles a not found request, a bad request, a forbidden request, etc. 

As seen in the screenshot, we inject several handlers, and customers can as well. As explained and seen in the code, the specific 'handling' is specifically looking only at the '404' errors. This is a 1-1 relationship and how it's being made and handled. 

In the end, anyone can inject several handlers for 'x' different errors. But the case you're currently looking at is specifically `404`

Screenshot 2023-01-26 at 11 26 10 AM

feature-libs/customer-ticketing/components/list/customer-ticketing-list/customer-ticketing-list.component.ts
bgambocjaviniar: Going back to this. I have discussed with a peer, and I recalled that this was not supposed to be handled this way through a query. What was supposed to be done specifically for the get ticket list request was to call the `connector` directly instead of wrapping it in a query. https://sap-my.sharepoint.com/:w:/p/parth_lakhani/EX96033xTSdMsG_zhI41Hh4Bn43TOKuOb1gpKNGmss_ABQ?e=k11ABj @adolfobee @Parthlakhani

feature-libs/customer-ticketing/core/utils/utils.ts
bgambocjaviniar: > I've now added the changes, as per our discussion on slack. Thank you again so much for your help on the solution.

We did discuss, but you shouldn't have pushed your solution as it wasn't an agreed solution.

projects/storefrontlib/shared/components/card/card.component.html
bgambocjaviniar: not a problem

projects/storefrontlib/shared/components/form/file-upload/file-upload.component.html
bgambocjaviniar: we'll keep it for now 

future improvements can be made if ever

feature-libs/customer-ticketing/components/details/customer-ticketing-messages/customer-ticketing-messages.component.ts
bgambocjaviniar: as discussed, logic was decoupled to event listeners

feature-libs/customer-ticketing/components/list/customer-ticketing-list/customer-ticketing-list.component.ts
bgambocjaviniar: after several discussions with Parth and a peer, we'll keep it for now

feature-libs/customer-ticketing/occ/adapters/occ-customer-ticketing.adapter.ts
bgambocjaviniar: fine

projects/storefrontlib/shared/components/chat-messaging/messaging/messaging.component.ts
bgambocjaviniar: future revisit

bgambocjaviniar (approved):

PR #16030, opened 2022-07-29
Feature/cxspa 905 change console message

ChristophHi:
That is not breaking, as the protected method was not released


schuppell (approved):
Looks good

PR #16036, opened 2022-08-01
fix: update test visa to 12/2125 expiry

SVendittelli:
The cypress test began failing as the expiry date was set to 07/2022.
When we reached 2022-07-01 the ests began to fail. Setting the expiry to
the far future fixes this.


giancorderoortiz:
Please set expire year to 2027 (to avoid scrolling) or 2031, which is the last option in dropdown menu. 
Screen Shot 2022-08-01 at 11 22 53 AM

WeizhengSap (approved):

PR #16042, opened 2022-08-03
feat: CXSPA-590 Quotes actions by role

pawelfras:


feature-libs/commerce-quotes/components/commerce-quotes-actions-by-role/commerce-quotes-actions-by-role.module.ts
```diff
@@ -0,0 +1,47 @@
...
+        actionsOrderByState: {
+          BUYER_DRAFT: [QuoteAction.CANCEL, QuoteAction.SUBMIT],
+          CANCELLED: [QuoteAction.REQUOTE],
+          SELLER_REQUEST: [QuoteAction.EDIT, QuoteAction.SUBMIT],
+          BUYER_OFFER: [
+            QuoteAction.CANCEL,
+            QuoteAction.EDIT,
+            QuoteAction.CHECKOUT,
+          ],
+        },
```
psmul: What do you think about wrapping this into an object to make it more context-precise, because now there will be two separate fields relating to the same configuration? So my proposition would be to wrap it into "actions". Thoughts?

feature-libs/commerce-quotes/components/details/overview/commerce-quotes-details-overview.component.html
```diff
@@ -1,68 +1,80 @@
...
-        <cx-card
-          [content]="
-            getCardContent(
-              quote?.expirationTime | cxDate,
-              'commerceQuotes.details.expiryDate'
-            ) | async
-          "
-        ></cx-card>
+<ng-container *ngIf="quoteDetails$ | async as quoteDetailsState">
+  <ng-container *ngIf="!quoteDetailsState.loading; else loading">
```
psmul, line 2: I think those should be replaced because in this case, the spinner will never show up as the data gets loaded and rendered so no point in showing spinner then

feature-libs/commerce-quotes/core/facade/commerce-quotes.service.ts
```diff
@@ -1,59 +1,45 @@
...
-            this.commerceQuotesConnector.getQuotes(userId, {
-              currentPage,
-              sort,
-              pageSize: this.config.view?.defaultPageSize,
-            })
-          )
-        ),
-      { reloadOn: [CommerceQuotesListReloadQueryEvent] }
-    );
+  protected isLoading$ = new BehaviorSubject<boolean>(false);
```
psmul: I'm not sure if I'm a fan of this solution. We're putting information for a single component in general service. In this case, we should have a separate service as it was done in `CommerceQuotesActionLinksService` - a dedicated service for such cases to avoid putting contextual data into general services.

feature-libs/commerce-quotes/components/commerce-quotes-actions-by-role/commerce-quotes-actions-by-role.component.ts
```diff
@@ -0,0 +1,61 @@
...
+        }
+        return {
+          ...detailsState,
+          data: {
+            ...data,
+            allowedActions: this.getOrderedList(state, allowedActions),
+          },
+        };
+      }
+    })
```
psmul: If we're using some model mapping here along with config for feature, I think we should create a normalizer for such a case. This way we provide a clear and readable way of transforming/mapping backend response into our local model type.

psmul: Side note: maybe it's worth double checking the above comment with Nik / Kris about whether this case requires a normalizer.

feature-libs/commerce-quotes/core/facade/commerce-quotes.service.ts
```diff
@@ -256,7 +282,19 @@ export class CommerceQuotesService implements CommerceQuotesFacade {
...
-    return this.quoteDetailsState$.get();
+  requote(quoteCode: string): Observable<Quote> {
+    return this.requoteCommand.execute({ quoteStarter: { quoteCode } });
+  }
+
+  getQuoteDetails(): Observable<QueryState<Quote | undefined>> {
+    return combineLatest([
+      this.isLoading$,
+      this.quoteDetailsState$.getState(),
+    ]).pipe(
```
psmul: I think instead of combining an additional/external flag, maybe we can put the filter into the pipe so we pass data only if loading for the state is false?

psmul (changes requested):

feature-libs/commerce-quotes/core/facade/commerce-quotes.service.ts
pawelfras: It wouldn't solve my issue and I think that we shouldn't filter data that way (otherwise the `loading` flag would not make sense and couldn't be used in template to show loading spinner in different scenarios than mine). I put additional flag because action like requote takes some time and only after that details are reloaded, so in the meantime user is not informed that something is happening under the hood. As you mentioned in comment before, this should be moved to component service.

feature-libs/commerce-quotes/components/commerce-quotes-actions-by-role/commerce-quotes-actions-by-role.component.ts
pawelfras: Good Idea! Going further, we could add an indicator to each action to have information if it's primary or not, instead of another filtering in component. I've asked Nik and I'm waiting for his thoughts about this :)

feature-libs/commerce-quotes/core/facade/commerce-quotes.service.ts
pawelfras: Yep, I agree it should be moved to separate component service 👍

feature-libs/commerce-quotes/components/commerce-quotes-actions-by-role/commerce-quotes-actions-by-role.module.ts
pawelfras: Consider it done!

feature-libs/commerce-quotes/components/details/overview/commerce-quotes-details-overview.component.html
pawelfras: Sorry, old name stayed and probably makes some misunderstanding - name of the stream should be `quoteDetailsState$`. I've changed return type to `Observable<QueryState<Quote | undefined>>` and as we're returning query state, first thing that I'm checking is if state is defined (what doesn't mean that data are loaded as well), then based on `loading` indicator I'm showing loading spinner or not. I'm pretty sure it reacts on each data reloading etc.

feature-libs/commerce-quotes/core/facade/commerce-quotes.service.ts
psmul: If we want to inform users about stuff happening in the background, using spinner component, I would take path with adding such flag on component level, not service

pawelfras: "As you mentioned in comment before, this should be moved to component service." - pawelfras 😂

feature-libs/commerce-quotes/core/facade/commerce-quotes.service.ts
pawelfras: Well, it turned out, that it cannot be moved to component-specific service, as this flag probably should be set each time some action is performed (not only requote) to inform user that something is loading under the hood. We can discuss whether there is a different approach to handle such case.

feature-libs/commerce-quotes/core/facade/commerce-quotes.service.ts
pawelfras: Finally it stays after discussion we had. Flag is necessary for all CMS-driven components.

feature-libs/commerce-quotes/core/facade/commerce-quotes.service.ts
pawelfras: Finally it stays after discussion we had. Flag is necessary for all CMS-driven components.

feature-libs/commerce-quotes/components/commerce-quotes-actions-by-role/commerce-quotes-actions-by-role.component.ts
pawelfras: Logic has been moved to normalizer with separate unit tests. Component is now clean and shiny 👌

psmul (approved):

PR #16046, opened 2022-08-04
feature/cxsecur5-9376/icon-loader

meeque:
Trying to rewrite `IconComponent` without bypassing Angular's `DomSanitizer`. Ideally by using Angular templating rather than HTML source code or DOM.

In order to use Angular templates, I moved the `TEXT` and `SVG` icon rendering logic from `IconLoaderService` to `IconComponent`. This worked quite well.

The stylesheet links for `LINK` items need to go into the header though. This turns out to be rather tricky to accomplish with Angular templating. Therefore I've left this part in `IconLoaderService`. It was already using safe DOM functions for most part. This is consistent with Angular's own [Meta](https://github.com/angular/angular/blob/14.1.2/packages/platform-browser/src/browser/meta.ts#L42-L201) service.

For backward compatibility with custom code, I've left `IconLoaderService.getHtml()` in place. But it's not used elsewhere in in Spartacus, apart from specs. Therefore I marked it as deprecated. Nevertheless, I fixed potential XSS flaws in `getHtml()`, by making better use of the DOM and running Angular's `DomSanitizer` on all URLs.

I tried to keep most specs, as long as they still apply. And I added heaps more. I've successfully tested most of them against the original implementation.

Also had to adjust some mocks in unrelated tests.


projects/storefrontlib/cms-components/misc/icon/icon.component.html
```diff
@@ -1 +1,3 @@
-<i [outerHTML]="icon"></i><ng-content></ng-content>
+<ng-container *ngIf="isTextIcon()">{{ icon }}</ng-container
+><svg *ngIf="isSvgIcon()"><use [attr.xlink:href]="icon"></use></svg
+><ng-content></ng-content>
```
meeque, line 3: Formatting is not nice, because there is no space allowed between either icon type and the projected content. This mimics the behavior of the old implementation. If whitespace is ok after all, we can form at this a lot nicer.

projects/storefrontlib/cms-components/misc/icon/icon-loader.service.ts
```diff
@@ -23,69 +24,76 @@ export class IconLoaderService {
...
   /**
    * Returns an html fragment which can be added to the DOM in a safe way.
+   *
+   * @deprecated only exists for backwards compatibility, html generation has been moved to icon component
    */
-  getHtml(type: ICON_TYPE | string): SafeHtml | undefined {
-    if (this.isResourceType(type, IconResourceType.SVG)) {
-      return this.sanitizer.bypassSecurityTrustHtml(
-        `<svg><use xlink:href="${this.getSvgPath(type)}"></use></svg>`
+  public getHtml(type: ICON_TYPE_STRING): SafeHtml | undefined {
```
meeque: If `getHtml()` is not needed for backward compatibility with custom code, it can be removed.

projects/storefrontlib/cms-components/misc/icon/icon.model.ts
```diff
@@ -43,6 +43,8 @@ export enum ICON_TYPE {
   EYE_SLASH = 'EYE_SLASH',
 }
 
+export type ICON_TYPE_STRING = ICON_TYPE | string;
```
meeque, line 52: Added this, because both `IconLoaderService` and `IconComponent` where already using something similar. Either inline or as declared type.

Not sure about the name. Maybe `ICON_TYPE_OR_STRING` or `ICON_TYPISH` instead?

projects/storefrontlib/cms-components/misc/icon/icon-loader.service.spec.ts
```diff
@@ -7,7 +8,8 @@ import { IconConfig, IconResourceType, ICON_TYPE } from './icon.model';
 
 const FONT_AWESOME_RESOURCE =
   'https://use.fontawesome.com/releases/v5.8.1/css/all.css';
-const MockFontIconConfig: IconConfig = {
+
+export const MockIconConfig: IconConfig = {
```
meeque, line 12: Exported this, so it can be reused in the IconComponent spec.

Not sure, if this is good practice. But it seems preferable to use a common `IconConfig` rather than setting up custom mocks or spies for each individual test.

Maybe move it to somewhere else? E.g. to the icon-testing module?

projects/storefrontlib/cms-components/misc/icon/icon.component.ts
```diff
@@ -41,87 +42,97 @@ export class IconComponent {
...
     this.setIcon(type);
   }
 
   /**
    * The type input parameter is bound to the icon type. You can feed the `ICON_TYPE` to
    * accomplish a configurable button in the UI.
    */
-  @Input() set type(type: ICON_TYPE) {
+  @Input() set type(type: ICON_TYPE_STRING) {
+    // XXX this clashes with the type attribute of some html form elements, e.g. input or button
```
meeque, line 61: I guess it'd be preferable to prefix all html stuff with `cx`. Maybe deprecate this and introduce a `cxType` setter instead? (Has little to do with this PR though.)

projects/storefrontlib/cms-components/misc/icon/icon.component.ts
```diff
@@ -41,87 +42,97 @@ export class IconComponent {
...
-
-  /**
-   * The `flip-at-rtl` class is added to the DOM for the style layer to flip the icon in RTL direction.
-   */
-  @HostBinding('class.flip-at-rtl') flipAtRtl: boolean;
-
-  /**
-   * The `flip-at-ltr` class is added to the DOM for the style layer to flip the icon in LTR direction.
-   */
-  @HostBinding('class.flip-at-ltr') flipAtLtr: boolean;
```
meeque, line 75: Replaced these with a single `@HostBinding` for `class`. Seems to make it easier.

projects/storefrontlib/cms-components/misc/icon/icon.component.ts
```diff
@@ -41,87 +42,97 @@ export class IconComponent {
...
+   * Adds the style classes for the icon.
    */
-  protected addStyleClasses(type: ICON_TYPE): void {
-    this.renderer.addClass(this.host, 'cx-icon');
-
-    this.styleClasses?.forEach((cls) =>
-      this.renderer.removeClass(this.host, cls)
-    );
+  protected addStyleClasses(type: ICON_TYPE_STRING): void {
+    this.styleClasses = [CSS_CLASSES.CX_ICON];
```
meeque, line 135: For some reason it works best to assign a brand new array to `styleClasses`. Safes us the trouble of deleting individual classes.

For some reason, Angular still retains the classes that are declared in the html template of the host component. I have not idea why. But tests indicate it, see `'host icon components'` specs.

projects/storefrontlib/cms-components/misc/icon/icon.component.spec.ts
```diff
@@ -67,113 +56,412 @@ describe('IconComponent', () => {
...
+        component.type = 'HAPPY';
+        fixture.detectChanges();
+
+        expect(nativeDebugElement.textContent).toBe('😊');
+        expect(nativeDebugElement.childElementCount).toBe(0);
+
+        const hostClassList = nativeDebugElement.classList;
+        expect(hostClassList.length).toBe(2);
+        expect(hostClassList).toContain('cx-icon');
+        expect(hostClassList).toContain('😊');
```
meeque, line 238: Not sure, if it makes sense to add the symbol of a TEXT icon to the CSS classes.  But afaict, the old implementation did it, too.

projects/storefrontlib/cms-components/misc/icon/icon-loader.service.ts
```diff
@@ -23,69 +24,76 @@ export class IconLoaderService {
...
+   * Get the `IconResourceType` that is configured for the given `ICON_TYPE`
+   * Defaults to `IconResourceType.LINK`, if an unknown `IconResourceType` is configured or none is configured.
+   */
+  public getResourceType(iconType: ICON_TYPE_STRING): IconResourceType {
+    const iconResourceType = this.config?.resources?.find((res) =>
+      res.types?.includes(iconType)
+    )?.type;
+    if (!Object.values<any>(IconResourceType).includes(iconResourceType)) {
+      return IconResourceType.LINK;
+    }
```
meeque, line 69: Looks like the old implementation defaulted to `LINK`. Or should we fail differently? In particular, when the config contains an unexpected IconResourceType?

Btw, I hope there is a better way to phrase this `if`-condition in TypeScript. But I couldn't figure it out. I tried to change the return type to `keyof typeof IconResourceType`, and let TypeScript do the rest. But that didn't work, because we need the values of the enum, not the keys.

projects/storefrontlib/cms-components/misc/icon/icon.component.ts
```diff
@@ -41,87 +42,97 @@ export class IconComponent {
...
+    if (this.isSvgIcon()) {
+      // looks like angular does not sanitize URLs when binding to the xlink:href attribute of svg elements
+      // so we need to sanitze the URL manually
+      this.iconValue =
+        this.sanitizer.sanitize(
+          SecurityContext.URL,
+          this.iconLoader.getSvgPath(type) || null
+        ) || undefined;
+    } else if (this.isTextIcon()) {
+      this.iconValue = this.iconLoader.getSymbol(type);
```
meeque, line 94: Fyi, `TEXT` icons are now treated as pure text. If the value happens to contain html markup, it will not be executed by the browser. As a result:

* Storefront visitors will see any html markup from `TEXT` icons verbatim.
* `TEXT` icons will never cause XSS.

This is different from the previous implementation. But there were no specs that used HTML code snippets as `TEXT` icon symbols, so I assume that this is not an intended use-case.

projects/storefrontlib/cms-components/misc/icon/icon-loader.service.ts
```diff
@@ -99,51 +107,51 @@ export class IconLoaderService {
...
-      link.type = 'text/css';
-      link.href = resource.url;
-      head.appendChild(link);
+      // using DOM APIs, so need to sanitize our URLs manually
+      const sanitizedUrl = this.sanitizer.sanitize(
+        SecurityContext.URL,
+        resource.url
+      );
+      if (sanitizedUrl) {
+        const head = this.winRef.document.getElementsByTagName('head')[0];
```
meeque, line 126: Or should we use `Renderer2` rather than using `document` directly?

meeque:
I've added some comments with questions to individual files. But I'm not sure if anyone else can see them yet? I think I may need to start a an official review to share my comment with you. But I don't have permissions:-(

projects/storefrontlib/cms-components/misc/icon/icon.component.ts
plabadie: If ever we rename things it will have to wait to a major version release because it's a breaking change.

projects/storefrontlib/cms-components/misc/icon/icon.component.ts
```diff
@@ -47,87 +47,100 @@ export class IconComponent {
    * The cxIcon directive is bound to the icon type. You can feed the `ICON_TYPE` to
    * accomplish a configurable button in the UI.
    */
-  @Input() set cxIcon(type: ICON_TYPE) {
+  @Input()
+  public set cxIcon(type: ICON_TYPE_STRING) {
```
plabadie: I think that as a convention we don't declare public members explicitely.

projects/storefrontlib/cms-components/misc/icon/icon.model.ts
plabadie: Replacing existing types with this new type creates a series of breaking changes accross the PR (among them are numerous method signature changes).  We have to revert this change if we are aiming to release in a minor version.

projects/storefrontlib/cms-components/misc/icon/icon-loader.service.ts
plabadie: Let's not return a different resource type.  That would be a non backwards compatible change.

projects/storefrontlib/cms-components/misc/icon/icon-loader.service.ts
plabadie: I don't know the pros and cons of one or the other.  However, using the Renderer2 would require a breaking change in the constructor to add a new dependency.

projects/storefrontlib/cms-components/misc/icon/icon.component.ts
```diff
@@ -47,87 +47,100 @@ export class IconComponent {
...
+   * If necessary, adds CSS classes for flipping for some icons to support rtl and ltr directions.
    */
-  protected flipIcon(type: ICON_TYPE) {
-    // TODO: this can be dropped with the next major release.
+  protected flipIcon(type: ICON_TYPE_STRING) {
+    // TODO: this check can be dropped with the next major release.
     if (!this.iconLoader.getFlipDirection) {
       return;
     }
     const iconDirection = this.iconLoader.getFlipDirection(type);
```
plabadie, line 123: I don't know if in practice this could happen, but it seems that calling flipIcon() multiple times with different types can accumulate both direction classes in the array.  Could that be the case givent that styleClasses is a class attribute and classes are always added to it?

With the previous logic it looks like only one or the other could be applied.

projects/storefrontlib/cms-components/misc/icon/icon.component.ts
```diff
@@ -47,87 +47,100 @@ export class IconComponent {
...
+
+  private iconValue?: string;
+  private iconResourceType: IconResourceType;
 
   constructor(
     protected iconLoader: IconLoaderService,
-    protected elementRef: ElementRef<HTMLElement>,
-    protected renderer: Renderer2
+    protected sanitizer: DomSanitizer
   ) {}
```
plabadie, line 78: Removing a constructor parameter is a breaking change.

projects/storefrontlib/cms-components/misc/icon/icon.component.ts
plabadie: We can't remove public class attributes in a minor release since it introduces a breaking change.

projects/storefrontlib/cms-components/misc/icon/icon.component.ts
meeque: ok, I'll fix that. I've seen both in the old code. and the linter didn't seem to complain

projects/storefrontlib/cms-components/misc/icon/icon.component.ts
meeque: I think that Angular's `@HostBinding('class')` will ultimately use DOM's `classList` under the hood. this is a  [DOMTokenList](https://developer.mozilla.org/en-US/docs/Web/API/DOMTokenList) which operates as a set.

Just call `document.body.classList.add('foo')` multiple times and observe the result.

But I'll see if any tests cover this. if not, I'll add new ones....

projects/storefrontlib/cms-components/misc/icon/icon.model.ts
meeque: I don't think so. I've only used `ICON_TYPE_STRING` in locations that where already using `ICON_TYPE | string`, either directly or via a type alias. TS does not care how the type is declared, it only cares that it's `ICON_TYPE | string`.

In `icon.component.ts` it's a little tricky to see, because the old code redefined `ICON_TYPE`:
```
import { ICON_TYPE as DEFAULT_ICON_TYPE } from './icon.model';
type ICON_TYPE = DEFAULT_ICON_TYPE | string;
```

So in that class, the old `ICON_TYPE` is the same as the new `ICON_TYPE_STRING`.

I thought redefining the meaning of a type would be very confusing. Or is this a common practice?

projects/storefrontlib/cms-components/misc/icon/icon.component.ts
meeque: Same for adding a non-optional constructor parameter. I'll definitely have to fix this.

projects/storefrontlib/cms-components/misc/icon/icon.component.ts
meeque: should be fixed now

PR #16048, opened 2022-08-05
add unit tests results XML generator

FollowTheFlo:
To be compliant with Sirius. Unit tests needs to be generated in XML files.
using Karma-Junit-reporter: https://www.npmjs.com/package/karma-junit-reporter
20 karma test suites are setup with XML report (storefrontapp test suite is excluded)

each xml files are set with name of its test suite eg 'unit-test-core-.xml unit-test-smartedit.xml' .
based on this file name shape, build.yml is setup with pattern **/unit-test-*.xml to collect unit test report files.
https://github.tools.sap/cx-commerce-storefronts/piper-pipeline-azure/blob/main/stages/build.yml
change from --filePattern **/TEST-*.xml to --filePattern **/unit-test-*.xml

bash files (unit-test.sh and unit-test-core-libs.sh) were modified to be able to run on both pipelines (GH and Azure), -azure flag needs to be added (unit-test.sh -azure) to target azure pipeline.

XML content:
In the PR, no need specific setup of xml attribute.
Note that in future to be FC-2 Compliant, attributes 'name' and 'classname' will need to be appropriately filled with rules defined in
https://wiki.one.int.sap/wiki/display/DevFw/Option+B%3A+Establish+Traceability+with+a+requirement.mapping+file#OptionB:EstablishTraceabilitywitharequirement.mappingfile-wildcard_example

testing:
- 19 test files generated, 10 for core, 9 for Integrations  -> expected as 'asset' test suite is not in bash files, so not running
- Those 19 files are visible in Sirius and interpreted in FC-3 'Regression' Portlet with Success, Skipped, Failed test cases numbers.


hackergil (dismissed):
I think all of the test reports have the same name and go to the same folder. Is that correct? If yes, it needs to be changed.
answer:
files have different name (eg:unit-test-storefront.xml), all in same folder ('../../unit-tests-reports')

projects/storefrontapp/karma.conf.js
```diff
@@ -11,11 +11,18 @@ module.exports = function (config) {
       require('karma-chrome-launcher'),
       require('karma-jasmine-html-reporter'),
       require('@angular-devkit/build-angular/plugins/karma'),
+      require('karma-junit-reporter'),
```
hackergil: We don't need reports on this one since we don't ship it

projects/storefrontapp-e2e-cypress/package.json
```diff
@@ -10,7 +10,7 @@
     "cy:open:epd-visualization": "cypress open --env BASE_SITE=powertools-epdvisualization-spa",
     "cy:run": "cypress run",
     "cy:run:b2b": "cypress run --env BASE_SITE=powertools-spa,OCC_PREFIX_USER_ENDPOINT=orgUsers",
-    "cy:run:core": "cypress run --spec \"cypress/integration/**/*.core-e2e-spec.ts\"",
+    "cy:run:core": "cypress run --reporter junit --reporter-options mochaFile=results/test-results-[hash].xml --spec \"cypress/integration/**/*.core-e2e-spec.ts\"",
```
bgambocjaviniar: I don't think this is part of this PR since this is for unit tests only and the other PR is for cypress right? If so, please remove

bgambocjaviniar (commented):
related to https://github.tools.sap/cx-commerce-storefronts/piper-pipeline-azure/commit/49195ecc75ea6a62cd495412620c3edfb97e6be8

core-libs/setup/karma.conf.js
```diff
@@ -11,11 +11,18 @@ module.exports = function (config) {
...
     ],
     client: {
       clearContext: false, // leave Jasmine Spec Runner output visible in browser
     },
-    reporters: ['progress', 'kjhtml', 'dots'],
+    reporters: ['progress', 'kjhtml', 'dots', 'junit'],
+    junitReporter: {
+      outputFile: 'unit-test-results.xml',
+      outputDir: require('path').join(__dirname, '../../unit-tests/storefront'),
+      classNameFormatter: undefined,
```
bgambocjaviniar: is it required to include this property as it shows that is the default https://github.com/karma-runner/karma-junit-reporter/blob/master/README.md

projects/storefrontapp/karma.conf.js
bgambocjaviniar: very good catch, and now I'm wondering why do we have this file and the 'test' section in the angular.json for storefrontapp

bgambocjaviniar (changes requested):
tiny questions from below/above and would like to go over together during the morning

projects/storefrontapp-e2e-cypress/package.json
FollowTheFlo: correct, will be removed

core-libs/setup/karma.conf.js
FollowTheFlo: correct, classNameFormatter can be removed

projects/storefrontapp/karma.conf.js
FollowTheFlo: revert done

ci-scripts/unit-tests-core-lib.sh
```diff
@@ -1,130 +1,124 @@
 #!/usr/bin/env bash
-set -e
 set -o pipefail
 
+# -azure argument used by azure hyperspace pipeline
```
FollowTheFlo: azure pipeline: all tests run even if one fails (as 'set-e' not present). Azure does not support stream target '/dev/fd/5'.
GH pipeline: keep the same behavior as before -> unit test process stops when 1 test fails

ci-scripts/unit-tests-core-lib.sh
```diff
@@ -1,130 +1,124 @@
...
 if [[ -n "$coverage" ]]; then
     echo "Error: Tests did not meet coverage expectations"
     exit 1
 fi
 echo "Running schematics unit tests and code coverage for smartedit library"
-exec 5>&1
-output=$(yarn --cwd feature-libs/smartedit run test:schematics --coverage=true | tee /dev/fd/5)
-
 
-if [[ $1 == '-h' ]]; then
```
FollowTheFlo, line 119: agreed with Brian to remove this old code intended to test sonar

ci-scripts/unit-tests-core-lib.sh
```diff
@@ -1,130 +1,124 @@
...
+    target=" | tee /dev/fd/5"
+fi
+
 echo "-----"
 
 echo "Running unit tests and code coverage for core"
-exec 5>&1
-output=$(ng test core --watch=false --sourceMap --code-coverage --browsers=ChromeHeadless | tee /dev/fd/5)
+
+output=$(eval "ng test core --watch=false --sourceMap --code-coverage --browsers=ChromeHeadless"$target)
```
FollowTheFlo: The only way to have the pipe '|' concat to a string then being interpreted as a command, is to use 'eval' afterwards.

ci-scripts/unit-tests.sh
```diff
@@ -1,130 +1,125 @@
 #!/usr/bin/env bash
-set -e
 set -o pipefail
 
+# -azure argument used by azure hyperspace pipeline
+if [[ $1 == '-azure' ]]; then
```
bgambocjaviniar: let's call it --azure

I would associate single dashes with 'aliased' names, so like -a

Basically, just change to --azure or support --azure OR -a

docs/libs/creating-lib.md
```diff
@@ -71,8 +71,8 @@ module.exports = function (config) {
     },
    reporters: ['progress', 'kjhtml', 'dots', 'junit'],
     junitReporter: {
-      outputFile: 'unit-test-results.xml',
-      outputDir: require('path').join(__dirname, '../../unit-tests/storefront'),      
+      outputFile: 'unit-test-storefront.xml',
+      outputDir: require('path').join(__dirname, '../../unit-tests-reports/storefront'),
```
bgambocjaviniar: from the direction that I'm now seeing, you you'll create a 1-1 relationship, where you create a directory and put that specific junit xml output to it.

Couldn't we have kept it in one folders with all these different outputFile instead?

for example,

junitReporter: {
 outputFile: 'unit-test-TODO:INSERTNAME.xml'
 outputDir: require(...).join(..., '../../unit-tests-reports'))
}

Basically, 1 folder, which contains all those output files

docs/libs/creating-lib.md
```diff
@@ -64,11 +64,17 @@ module.exports = function (config) {
...
     ],
     client: {
       clearContext: false, // leave Jasmine Spec Runner output visible in browser
     },
-    reporters: ['progress', 'kjhtml', 'dots'],
+   reporters: ['progress', 'kjhtml', 'dots', 'junit'],
+    junitReporter: {
+      outputFile: 'unit-test-<lib-name>.xml',
+      outputDir: require('path').join(__dirname, '../../unit-tests-reports'),      
+       useBrowserName: false,
```
bgambocjaviniar: remove that extra space

ci-scripts/unit-tests.sh
bgambocjaviniar: nvm it was already changed lol

I look through commits from top down when new things come

ci-scripts/unit-tests.sh
```diff
@@ -1,130 +1,125 @@
 #!/usr/bin/env bash
-set -e
 set -o pipefail
 
+# -azure argument used by azure hyperspace pipeline
```
bgambocjaviniar: it's still -azure instead of --release :d

bgambocjaviniar (approved):

bgambocjaviniar (changes requested):
discussed with Gil

the approach on just updating it in the private repo may be the route we want to take instead

bgambocjaviniar (dismissed):

bgambocjaviniar (approved):

PR #16056, opened 2022-08-08
CXEC-14759: Integrate api for ticket details

Parthlakhani:
https://jira.tools.sap/browse/CXEC-14759

Integrated getTickt API for getting ticket details by ID
added unit tests for core and component for ticket details


feature-libs/customer-ticketing/components/customer-ticketing-details/customer-ticketing-details.component.spec.ts
```diff
@@ -33,4 +36,38 @@ describe('CustomerTicketingDetailsComponent', () => {
...
+    const mockCardContent: Card = {
+      text: ['1'],
+      title: 'ID',
+      customClass: '',
+    };
+    component
+      .prepareCardContent(mockTicketId, 'ID')
+      .pipe(take(1))
+      .subscribe((result) => {
+        expect(result).toEqual(mockCardContent);
```
bgambocjaviniar, line 68: This is awesome where you used take(1). I recall this is referenced using the checkout lib. If you look at the checkout lib, some includes param of `done` to include the callback for `done()`, which helps alleviate async code.

```
it('sohuld prepare content for card', (done) => {
    const mockCardContent: Card = {
      text: ['1'],
      title: 'ID',
      customClass: '',
    };
    component
      .prepareCardContent(mockTicketId, 'ID')
      .pipe(take(1))
      .subscribe((result) => {
        expect(result).toEqual(mockCardContent);
       done();
     });
});
   
```

It's good practice including the done callback to let jasmine know the subscription actually got triggered and we can finish this assertion as without the done callback, it can happen that the test may be 'false positive' in a way that the async code didn't happen to finish, but still the test says it passes.

Make sense?

feature-libs/customer-ticketing/components/customer-ticketing-details/customer-ticketing-details.component.spec.ts
```diff
@@ -33,4 +36,38 @@ describe('CustomerTicketingDetailsComponent', () => {
   it('should create', () => {
     expect(component).toBeTruthy();
   });
+
+  it('sohuld prepare content for card', () => {
```
bgambocjaviniar:
```suggestion
  it('should prepare content for card', () => {
```

feature-libs/customer-ticketing/components/customer-ticketing-details/customer-ticketing-details.component.ts
```diff
@@ -1,73 +1,47 @@
...
-        message: 'This is the way',
-      },
-      {
-        author: 'Mark Rivers',
-        createdAt: '2022-06-22T14:37:15+0000',
-        message: 'A message to consider',
-      },
-    ],
-  });
+  ticketDetails$: Observable<TicketDetails | undefined> =
```
bgambocjaviniar: can you move properties above constructors and functions below the constructor please

just nitpicking

feature-libs/customer-ticketing/components/customer-ticketing-details/customer-ticketing-details.component.ts
```diff
@@ -1,73 +1,47 @@
...
-        customClass: this.getStatusClass(entity),
+        customClass: this.getStatusClass(id),
       }))
     );
   }
 
-  getStatusClass(status: string): string {
-    return status === STATUS.OPEN
+  getStatusClass(id?: string): string {
+    return id === STATUS.OPEN
```
bgambocjaviniar, line 43: to verify (Brian)

feature-libs/customer-ticketing/components/customer-ticketing-details/customer-ticketing-details.component.spec.ts
bgambocjaviniar: should be applied to any async testing

feature-libs/customer-ticketing/core/facade/customer-ticketing.service.ts
```diff
@@ -1,6 +1,73 @@
...
+        if (!userId || !routingParams) {
+          throw new Error('Customer tickets pre conditions not met');
+        }
+        return [userId, routingParams.ticketCode];
+      })
+    );
+  }
+
+  getTicket(): Observable<TicketDetails | undefined> {
+    return this.getTicketQuery$.getState().pipe(map((state) => state.data));
```
bgambocjaviniar: would you like to expose the 'loading state' and data? If so, no need to map to data

If you would like to just expose the data, then call the .get() function instead of getState() from the query

----
My question is geared towards how would you use it and if it would benefit the customer.
Will you have a spinner upon landing in the details page (flexible as this is not what's happening in other details page, but it should have it)?

feature-libs/customer-ticketing/root/model/customer-ticketing.model.ts
```diff
@@ -4,6 +4,33 @@ export const enum CUSTOM_CLASS {
...
+  ticketCategory?: Categoty;
+  ticketEvents?: Array<Event>;
+}
+
+export interface Status {
+  id: string;
+  name: string;
+}
+
+export interface Categoty {
```
bgambocjaviniar:
```suggestion
export interface Category {
```

feature-libs/customer-ticketing/core/facade/customer-ticketing.service.ts
bgambocjaviniar: If you would like to use the state and data, then the type will be of Observable<QueryState<TicketDetails | undefined>> instead

feature-libs/customer-ticketing/core/facade/customer-ticketing.service.ts
```diff
@@ -1,6 +1,73 @@
...
+    );
+
+  constructor(
+    protected queryService: QueryService,
+    protected userIdService: UserIdService,
+    protected customerTicketingConnector: CustomerTicketingConnector,
+    protected routingService: RoutingService
+  ) {}
+
+  protected checkTicketPreConditions(): Observable<[string, string]> {
```
bgambocjaviniar: customerTicketingPreconditions 

--> similarly to what checkout had as checkoutPreconditions

feature-libs/customer-ticketing/core/facade/customer-ticketing.service.ts
```diff
@@ -1,6 +1,73 @@
...
+    return combineLatest([
+      this.userIdService.getUserId(),
+      this.routingService.getParams(),
```
bgambocjaviniar: instead of having it re-emit for different params if ever that becomes the case, then maybe we can compose this differently as follows:

```
 combineLatest([
this.routingService..getRouterState().pipe(
      map((routingData) => routingData.state.params.orderCode),
      distinctUntilChanged()
)
])
```

What do you think?

bgambocjaviniar (changes requested):

feature-libs/customer-ticketing/components/customer-ticketing-details/customer-ticketing-details.component.spec.ts
Parthlakhani: yes, makes sense. I will add it

feature-libs/customer-ticketing/components/customer-ticketing-details/customer-ticketing-details.component.ts
Parthlakhani: yes, the check was based on status name before, I changed it to status id because status name changes if the language is changed

feature-libs/customer-ticketing/core/facade/customer-ticketing.service.ts
Parthlakhani: yes, I kept the state earlier, but there was no use of it so I changes it to this. 
let's just use the get method instead

feature-libs/customer-ticketing/core/facade/customer-ticketing.service.ts
Parthlakhani: yes we do this. but, why can't we use the getParams method instead?

Parthlakhani: like this:
```
combineLatest([
      this.routingService.getParams().pipe(
        map((params) => params.ticketCode),
        distinctUntilChanged()
      ),
    ])
```

bgambocjaviniar: that's fine too

feature-libs/customer-ticketing/core/connectors/customer-ticketing.connector.spec.ts
```diff
@@ -0,0 +1,43 @@
...
+import { take } from 'rxjs/operators';
+import { CustomerTicketingAdapter } from './customer-ticketing.adapter';
+import { CustomerTicketingConnector } from './customer-ticketing.connector';
+
+import createSpy = jasmine.createSpy;
+
+class MockCustomerTicketingAdapter
+  implements Partial<CustomerTicketingAdapter>
+{
+  getTicket = createSpy().and.returnValue(of({}));
```
bgambocjaviniar: the getTicket is supposed to return 'something'

let's assert that the subscribed value returns that 'something' for this spec file as well.

feature-libs/customer-ticketing/core/facade/customer-ticketing.service.ts
bgambocjaviniar: let's get in a call

bgambocjaviniar: -------
reminder

UX/UI:
- ask for if there should be a spinner after visiting the details page / refresh the details page
- ask what should happen when a support ticket does not exist (should it rely redirect back to the listing page with a glboal message alert?)

Additional review from your team

feature-libs/customer-ticketing/components/customer-ticketing-details/customer-ticketing-details.component.spec.ts
```diff
@@ -33,4 +54,39 @@ describe('CustomerTicketingDetailsComponent', () => {
...
+      let result = component.getStatusClass('CLOSE');
+
+      expect(result).toEqual('cx-text-gray');
+    });
+
+    it('should return empty if the id is not passed', () => {
+      let result = component.getStatusClass('');
+
+      expect(result).toEqual('');
+    });
```
khristian23:
```suggestion
    function assertStatusClassByStatusId(expectedClass: string, statusId: string) {
      const result = component.getStatusClass(statusId);
      expect(result).toEqual(expectedClass);
    }

    it('should return open class when the status is open', () => {
      assertStatusClassByStatusId('cx-text-green', 'OPEN');
    });

    it('should return close class when the status is close', () => {
      assertStatusClassByStatusId('cx-text-gray', 'CLOSE');
    });

    it('should return empty if the id is not passed', () => {
      assertStatusClassByStatusId('', '');
    });
    
    it('should return empty if the id is undefined', () => {
      assertStatusClassByStatusId('', undefined);
    });
```

feature-libs/customer-ticketing/root/facade/customer-ticketing.facade.ts
```diff
@@ -0,0 +1,20 @@
...
+import { CUSTOMER_TICKETING_FEATURE } from '../feature-name';
+import { TicketDetails } from '../model';
+
+@Injectable({
+  providedIn: 'root',
+  useFactory: () =>
+    facadeFactory({
+      facade: CustomerTicketingFacade,
+      feature: CUSTOMER_TICKETING_FEATURE,
+      methods: ['getTicket'],
```
bgambocjaviniar: need to add getTicketState as well 

make it the same order of it the same as it is in the abstract methods for my eyes 👁️ 👁️ 

-----
for further explanation on why we're doing this:
- you can lazy load a feature by calling one of it's methods as well instead of relying on the cms component to exist in a page.

Parthlakhani: ohh, good to know. adding

feature-libs/customer-ticketing/occ/adapters/occ-customer-ticketing.adapter.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+        return (
+          req.method === 'GET' &&
+          req.url === `users/${mockCustomerId}/tickets/${mockTicketId}`
+        );
+      });
+      expect(mockReq.cancelled).toBeFalsy();
+      expect(mockReq.request.responseType).toEqual('json');
+      mockReq.flush(mockTicketiDetails);
+    });
+  });
```
khristian23, line 73: What happens when no data is retrieved? like on 404

feature-libs/customer-ticketing/root/events/customer-ticketing-event.listener.spec.ts
```diff
@@ -0,0 +1,79 @@
...
+  });
+
+  describe('onGetTicketQueryReload', () => {
+    it('LanguageSetEvent should dispatch GetTicketQueryReloadEvent', () => {
+      mockEventStream$.next(new LanguageSetEvent());
+
+      expect(eventService.dispatch).toHaveBeenCalledWith(
+        {},
+        GetTicketQueryReloadEvent
+      );
```
khristian23: nit:
These lines are repeated: can we factorize them by:
function assertServiceDispatchForEvent(event, dispatchMethod) {
     mockEventStream$.next(event);

      expect(eventService.dispatch).toHaveBeenCalledWith(
        {},
        dispatchMethod
      );
}

feature-libs/customer-ticketing/root/events/customer-ticketing.events.ts
```diff
@@ -0,0 +1,15 @@
+import { CxEvent } from '@spartacus/core';
+
+export class GetTicketQueryResetEvent extends CxEvent {
+  /**
+   * Event's type
+   */
```
khristian23: Are these comments needed?

feature-libs/customer-ticketing/root/model/customer-ticketing.model.ts
```diff
@@ -4,6 +4,33 @@ export const enum CUSTOM_CLASS {
...
+
+export interface Category {
+  id: string;
+  name: string;
+}
+
+export interface Event {
+  author?: string;
+  createdAt?: string;
+  message?: string;
```
khristian23, line 35: missing: toStatus: Status

khristian23 (changes requested):

feature-libs/customer-ticketing/root/events/customer-ticketing-event.listener.spec.ts
Parthlakhani: yes. changing

feature-libs/customer-ticketing/occ/adapters/occ-customer-ticketing.adapter.spec.ts
Parthlakhani: do you mean in this test case or in the adapter?

feature-libs/customer-ticketing/root/events/customer-ticketing.events.ts
Parthlakhani: not really, but it was there for all others events. should I remove it?

khristian23: There are not adding any value.  Please remove

feature-libs/customer-ticketing/occ/adapters/occ-customer-ticketing.adapter.spec.ts
khristian23: in the adapter :)

feature-libs/customer-ticketing/root/events/customer-ticketing.events.ts
Parthlakhani: got it

feature-libs/customer-ticketing/occ/adapters/occ-customer-ticketing.adapter.spec.ts
Parthlakhani: It would catch the error, and we should be able to use that to show error page

feature-libs/customer-ticketing/root/model/customer-ticketing.model.ts
```diff
@@ -14,6 +14,7 @@ export interface TicketDetails {
   createdAt?: string;
   modifiedAt?: string;
   status?: Status;
+  toStatus?: Status;
```
khristian23: toStatus should be part of the  Event interface, not the TicketDetails :)

khristian23 (changes requested):

feature-libs/customer-ticketing/occ/adapters/occ-customer-ticketing.adapter.spec.ts
khristian23: OK

feature-libs/customer-ticketing/root/model/customer-ticketing.model.ts
Parthlakhani: ohh, let me correct

khristian23 (approved):

bgambocjaviniar (approved):

PR #16044, opened 2022-08-03
CXEC-15848: UI for reopen/close request

Parthlakhani:
https://jira.tools.sap/browse/CXEC-15848

CMS component for reopen request button
CMS component for close request button
Reopen request dialog box with file upload and message
Changes in file upload to support custom button

PR #16059, opened 2022-08-09
Adding ticketing list, styles and table - Feature/CXEC-15925

adolfobee:
Adding ticketing list, styles and table


feature-libs/customer-ticketing/assets/translations/en/customer-ticketing.ts
```diff
@@ -15,5 +17,12 @@ export const customerTicketing = {
     uploadFile: 'Upload File',
     fileSizeLimit: 'File Size Limit: {{count}}MB',
     maximumAttachment: 'Maximum of one attachment per message',
+    noTickets: `You don't have any request`,
+    requestTitle: 'All Requests',
+    sortSubtitle: 'Sort by',
+    // Mobile view
+    ticketIdFull: 'Ticket ID',
+    dateCreated: 'Date Created',
+    dateUpdated: 'Date Updated',
```
khristian23: Currently discussing with So.  We need to follow Desktop View.  IMO It should be "Created On" and "Changed On"
I am asking So to update Figma wires

feature-libs/customer-ticketing/components/customer-ticketing-list/customer-ticketing-list.component.html
```diff
@@ -1,13 +1,110 @@
...
+      </span>
+      <span class="cx-ticketing-list-title-count">
+        ({{ tickets.length }})
+      </span>
+    </h3>
+    <div class="cx-ticketing-list-subtitle">
+      {{ 'customerTicketing.sortSubtitle' | cxTranslate }}
+    </div>
+    <select
+      class="cx-ticketing-list-select"
```
khristian23: Is cx-ticketing-list-select following spartacus rules?  Why do we need to create a new class?
Same question for:
cx-ticketing-list-title
cx-ticketing-list-title-text
cx-ticketing-list-subtitle

feature-libs/customer-ticketing/styles/components/_customer-ticketing-list.scss
```diff
@@ -0,0 +1,76 @@
...
+    .cx-tickting-list-ticket-mobile-icon {
+      margin-top: auto;
+      margin-bottom: auto;
+    }
+
+    cx-icon {
+      color: #a0aabf;
+    }
+
+    .cx-ticketing-list-ticket-item-mobile {
```
khristian23: Similar question, isn't there already a class that achieves the same?

feature-libs/customer-ticketing/components/customer-ticketing-list/customer-ticketing-list.component.html
```diff
@@ -1,13 +1,110 @@
...
+            <th scope="col">{{ 'customerTicketing.changedOn' | cxTranslate }}</th>
+            <th scope="col">{{ 'customerTicketing.status' | cxTranslate }}</th>
+          </tr>
+        </thead>
+        <tbody>
+          <tr *ngFor="let ticket of tickets">
+            <td>
+              <a [routerLink]="['/my-account/customer-ticketing', ticket.id]">{{ ticket.id }}</a>
+            </td>
+            <td>{{ ticket.subject }}</td>
```
khristian23: Subject is also clickable

feature-libs/customer-ticketing/components/customer-ticketing-list/customer-ticketing-list.component.html
```diff
@@ -1,13 +1,110 @@
...
+            <span class="col-4 cx-ticketing-list-ticket-item-mobile">
+            {{ 'customerTicketing.ticketCategory' | cxTranslate }}
+            </span>
+            <span class="col-8 cx-ticketing-list-ticket-item-mobile">
+            {{ ticket.ticketCategory }}
+            </span>
+          </div>
+          <div>
+            <span class="col-4 cx-ticketing-list-ticket-item-mobile">
+            {{ 'customerTicketing.dateCreated' | cxTranslate }}
```
khristian23:
```suggestion
            {{ 'customerTicketing.createdOn' | cxTranslate }}
```

feature-libs/customer-ticketing/components/customer-ticketing-list/customer-ticketing-list.component.html
```diff
@@ -1,13 +1,110 @@
...
+            </span>
+            <span class="col-8 cx-ticketing-list-ticket-item-mobile">
+            {{ ticket.createdAt }}
+            </span>
+          </div>
+          <div>
+            <span class="col-4 cx-ticketing-list-ticket-item-mobile">
+            {{ 'customerTicketing.dateUpdated' | cxTranslate }}
```
khristian23:
```suggestion
            {{ 'customerTicketing.changedOn' | cxTranslate }}
```

feature-libs/customer-ticketing/components/customer-ticketing-list/customer-ticketing-list.component.ts
```diff
@@ -1,11 +1,135 @@
...
+  iconTypes = ICON_TYPE;
+
   constructor() {}
 
-  ngOnInit() {}
+  // Create table Headers
+  tableHeaders = [
+    { key: 'ticketId', label: 'ID' },
+    { key: 'subject', label: 'Subject' },
+    { key: 'ticketCategory', label: 'Category' },
```
khristian23: tableHeader will be removed in a later commit?

feature-libs/customer-ticketing/styles/components/_customer-ticketing-list.scss
```diff
@@ -0,0 +1,76 @@
...
+          font-size: 14px;
+          line-height: 20px;
+          color: #000000;
+          padding: 16px 0;
+        }
+      }
+    }
+  }
+
+  .cx-ticketing-list-mobile {
```
khristian23: Do we really need these?  I will be surprised if we do not have mobile breakpoints already defined.

khristian23 (changes requested):

feature-libs/customer-ticketing/assets/translations/en/customer-ticketing.ts
adolfobee: I have changed it to the way figma reflects now, using changed on and created on

feature-libs/customer-ticketing/components/customer-ticketing-list/customer-ticketing-list.component.html
adolfobee: I'm just creating this classes for css styling

feature-libs/customer-ticketing/styles/components/_customer-ticketing-list.scss
adolfobee: There is not, each of this classes must have different styling

feature-libs/customer-ticketing/components/customer-ticketing-list/customer-ticketing-list.component.html
adolfobee: I will modify this when I have created the link to the ticket in the next pr

feature-libs/customer-ticketing/components/customer-ticketing-list/customer-ticketing-list.component.ts
adolfobee: Not really, they are being used to display the options in the dropdown

feature-libs/customer-ticketing/styles/components/_customer-ticketing-list.scss
adolfobee: this type of breakpoints are the ones used across the code in different places, if there is a better implementation we can change it too!

bgambocjaviniar (changes requested):
at a first glance and looking at it in the web browser without looking at the code/PR,

Maybe we had some confusion, but I believe this is one of those 'spartacus' way on how the table should look like 

for example:

- the line outer padding / line-height is different
- the headers should be all uppercase
- sort by and input box should be beside each other (input selection box is looking weird)
- left most table cell data should be underlined
- each table data should be clickable
- each table row should be clickable
- last table header should be right aligned

Screen Shot 2022-08-26 at 3 25 57 PM


----
looking at the PR soon

feature-libs/customer-ticketing/assets/translations/en/customer-ticketing.ts
```diff
@@ -15,5 +17,12 @@ export const customerTicketing = {
     uploadFile: 'Upload File',
     fileSizeLimit: 'File Size Limit: {{count}}MB',
     maximumAttachment: 'Maximum of one attachment per message',
+    noTickets: `You don't have any request`,
+    requestTitle: 'All Requests',
+    sortSubtitle: 'Sort by',
+    // Mobile view
```
bgambocjaviniar: instead of putting comments to separate things, what do you think about below:

```
export const customerTicketing = {
 customerTicketingDetails: {
 anyKeyValueForThisPage
}

customerTicketingList: {
anyKeyvaluePair or even complex objects

desktop: {
 anyKeyValuePair
}

mobile: {
anyKeyValuePair
}
}

}
```

does that make sense?
Let's split this entire translations into chunks basically

feature-libs/customer-ticketing/components/customer-ticketing-list/customer-ticketing-list.component.html
```diff
@@ -1,13 +1,110 @@
-<h2 class="cx-title d-none d-lg-block d-xl-block">
-  <!-- {{ 'title' | cxTranslate }} -->
-</h2>
-
 <ng-container>
```
bgambocjaviniar: I don't think this outer ng-container is necessary

feature-libs/customer-ticketing/components/customer-ticketing-list/customer-ticketing-list.component.ts
bgambocjaviniar: It should be gone.

An example would be the OrderHistoryComponent. There's no hardcoded things here as the paginated object comes from the backend and we'll be using that to populate the dropdown

feature-libs/customer-ticketing/styles/components/_customer-ticketing-list.scss
```diff
@@ -0,0 +1,76 @@
+cx-customer-ticketing-list {
```
bgambocjaviniar: You should be using placeholders selectors. An example can be found here https://github.com/SAP/spartacus/pull/16044/commits/5cb1d7acaf03997ab209878961e2cfa8fe056a42.

An explaination can be found here https://github.com/SAP/spartacus/pull/16044#discussion_r941734729

feature-libs/customer-ticketing/components/customer-ticketing-list/customer-ticketing-list.component.ts
adolfobee: Yes, this change will be implemented with the next PR that adds backend callls!

feature-libs/customer-ticketing/styles/components/_customer-ticketing-list.scss
adolfobee: Awesome will change this!

feature-libs/customer-ticketing/assets/translations/en/customer-ticketing.ts
khristian23: If changes were done, we no longer need these dateCreated nor dateUpdated. Please remove

feature-libs/customer-ticketing/components/customer-ticketing-list/customer-ticketing-list.component.html
```diff
@@ -1,13 +1,110 @@
...
+    </h3>
+    <div class="cx-ticketing-list-subtitle">
+      {{ 'customerTicketing.sortSubtitle' | cxTranslate }}
+    </div>
+    <select
+      class="cx-ticketing-list-select"
+      [ngModel]="selected"
+      (ngModelChange)="selected = $event"
+    > 
+      <option *ngFor="let heading of tableHeaders" [ngValue]="heading">
```
khristian23: We just need TicketId and ChangedOn

khristian23 (changes requested):

feature-libs/customer-ticketing/styles/components/_customer-ticketing-list.scss
```diff
@@ -0,0 +1,76 @@
...
+    }
+  }
+
+  .cx-ticketing-list-mobile {
+    @include media-breakpoint-down(lg) {
+      border-top: 1px solid var(--cx-color-light);
+      display: flex;
+    }
+
+    @include media-breakpoint-up(lg) {
```
bgambocjaviniar: The table is invisible from >= 992 to <= 1200 width


Screen Shot 2022-08-31 at 4 53 24 PM

@morganm58  / @developpeurweb  
This is another point of having a 'different' look and feel table without different styles and not really the same template. Problems like these will occur

bgambocjaviniar: All in all, I would say the template should be the similar and width similar applied styles to all our listing pages

bgambocjaviniar (changes requested):
@developpeurweb / @morganm58 

what do you think about having a different 'look and feel' table. 
Basically, about this statement below, which ties to this request https://github.com/SAP/spartacus/pull/16059#pullrequestreview-1087447665

```
! I haven’t changed the way the table looks because we were told to give priority to how it looks in the mockups, so things like all uppercase headers and the thickness of the table I tried to make them look as close as possible to what I perceive in the mockups
```
Screen Shot 2022-08-31 at 4 50 12 PM

I for one will think it's weird because we have these 'listing' pages, which has all the same 'table' (look and feel wise), but just for customer ticketing, it will be different. What do you think?

Can @morganm58 / @developpeurweb  one of you work with Adolfo if really requires the same 'table' (look and feel wise)

morganm58:
> @developpeurweb / @morganm58
> 
> what do you think about having a different 'look and feel' table. Basically, about this statement below, which ties to this request [#16059 (review)](https://github.com/SAP/spartacus/pull/16059#pullrequestreview-1087447665)
> 
> ```
> ! I haven’t changed the way the table looks because we were told to give priority to how it looks in the mockups, so things like all uppercase headers and the thickness of the table I tried to make them look as close as possible to what I perceive in the mockups
> ```
> 
> Screen Shot 2022-08-31 at 4 50 12 PM
> 
> I for one will think it's weird because we have these 'listing' pages, which has all the same 'table' (look and feel wise), but just for customer ticketing, it will be different. What do you think?
> 
> Can @morganm58 / @developpeurweb one of you work with Adolfo if really requires the same 'table' (look and feel wise)

The table should follow the look and feel of the existing tables. When in doubt always refer to the spartacus core UX designers and also Miguel( @developpeurweb ) and I (@morganm58) to ask questions.

@adolfobee You can ping me on slack if you have any questions.

bgambocjaviniar (changes requested):
- tables still doesn’t look the same as there’s no table applied to it
- this issue is still a problem https://github.com/SAP/spartacus/pull/16059#discussion_r960018493
- mobile should be the ‘same’
- is the add button supposed to be below the table or on top?
-------
solution:
- copy paste saved cart listing page, and add/remove as many columns as you need to have all your necessary data in place.
- copy paste the styles of saved cart list.scss and change the class name to say customer ticketing and just adapt the styles if needed
^ it would alleviate all the problems for look and feel

OR

you can Mikel to do your UI table and styles

Screen Shot 2022-09-06 at 3 15 21 PM
Screen Shot 2022-09-06 at 3 18 15 PM

bgambocjaviniar (changes requested):
As posted in the DM between your front end team and us, I'm just reposting for visibility.

**This applies to everyone starting today
For any new change / commits, please have someone in your team  to take a look at it and if they think it’s okay, you can send it to me for review.**

------ 

Screen Shot 2022-09-16 at 10 26 23 AM

The look and feel of the desktop is getting there, but some improvements can be done as I looked at this rejuvenated template that you have for the look and feel. All things below are relevant to the what the PR and issue entails as you are creating the UI before working with the data as done with your peers as well

- any reason to technically have 2 ‘tables’ (mobile and desktop) --> which is basically just a repeat of displayed data and extra style classes. Moreover think about it as if ever some sort of selector needs to be done to choose between them, it would need to be applied to two places instead of one 'table' 
- Let’s make *ngIf="tickets.length > 0" to *ngIf="tickets.length > 0; noCustomerTickets" and bind it to the ngTemplate of that reference. Reason for that, it may cover cases if ever the backend return <=0.
- due to the first bullet point, when you hover in a ‘row’ for your mobile, it doesn’t have the hover highlight 
- the dropdown, you should have been using cx-sorting . Again the references were saved cart / order history.
- I know I said to forget your pagination object as this PR was taking too long, but since we’re already here, can you include it. You should be using cx-pagination. They don’t exist in saved cart as the backend doesn’t send back the pagination, so take a look at order history for this. You can user mark rivers in powertools and look at order history. I set it up for you to ‘look’ at it
- screenshot 1. the data for id should be aligned ----- padding-inline-start
- You can apply the colors as well on your ‘table datas’ as some of our listing tables do have them like the replenishment order history and order approval
- **Finally and more importantly**, your table is not accessible. If you try to tab through the tables, you won’t be able to. Not like all our other ‘listing’ tables (saved cart, replenishment, order history, order approval, etc)

feature-libs/customer-ticketing/components/customer-ticketing-list/customer-ticketing-list.component.html
```diff
@@ -1,13 +1,125 @@
...
+          (viewPageEvent)="pageChange($event)"
+        ></cx-pagination>
+      </div>
+    </div>
+  </ng-container>
+  <ng-container *ngIf="!customerTicketsFlag">
+    <div class="text-center">
+      <h3>{{ 'customerTicketingList.noTickets' | cxTranslate }}</h3>
+    </div>
+  </ng-container>
```
Parthlakhani: need to close main `ng-container` here.

Parthlakhani:
some unit tests are failing, we are not adding any new unit tests for this PR. but let's just fix the failing ones.

feature-libs/customer-ticketing/components/customer-ticketing-list/customer-ticketing-list.component.ts
```diff
@@ -1,11 +1,173 @@
...
+
+  getStatusClass = (status: string): string => {
+    switch (status) {
+      case 'OPEN':
+        return 'cx-text-green';
+      case 'CLOSED':
+        return 'cx-text-red';
+      default:
+        return '';
+    }
```
Parthlakhani: As per the mockups,
color for the `Closed` should be gray. unless it was changes later.

Also, we need to consider `In Process` status as well.

Similar to what we have in customer-ticketing-details, it could be something like this:

```
  getStatusClass(id?: string): string {
    return id === STATUS.OPEN || id === STATUS.INPROCESS
      ? CUSTOM_CLASS.OPEN
      : id === STATUS.CLOSE
      ? CUSTOM_CLASS.CLOSE
      : '';
  }
```

feature-libs/customer-ticketing/components/customer-ticketing-list/customer-ticketing-list.component.html
```diff
@@ -1,13 +1,125 @@
...
+    </label>
+    <div
+      *ngIf="ticketList.ticketsObs.pagination.totalPages > 1"
+      class="cx-ticketing-list-pagination"
+    >
+      <cx-pagination
+        [pagination]="ticketList.ticketsObs.pagination"
+        (viewPageEvent)="pageChange($event)"
+      ></cx-pagination>
+    </div>
```
Parthlakhani: We do not need pagination twice, right? 
if yes, we can get rid of this.

feature-libs/customer-ticketing/styles/components/_customer-ticketing-list.scss
```diff
@@ -0,0 +1,149 @@
...
+  }
+
+  .cx-ticketing-list-thead-mobile {
+    @include media-breakpoint-down(sm) {
+      display: none;
+    }
+  }
+
+  .cx-ticketing-list-label {
+    padding-right: 15px;
```
Parthlakhani: validation check is failing because of this, `padding-inline-end` can be used here instead.

please refer to this comment from Mikel for better understanding:
https://github.com/SAP/spartacus/pull/16085#discussion_r977730561

feature-libs/customer-ticketing/styles/components/_customer-ticketing-list.scss
```diff
@@ -0,0 +1,149 @@
...
+
+    @include media-breakpoint-down(sm) {
+      font-size: var(--cx-font-size, 0.875rem);
+      font-weight: var(--cx-font-weight-bold);
+      min-width: 140px;
+    }
+  }
+
+  .cx-ticketing-list-value {
+    padding-left: 15px;
```
Parthlakhani: similarly, `padding-inline-start` here.

feature-libs/customer-ticketing/styles/components/_customer-ticketing-list.scss
```diff
@@ -0,0 +1,149 @@
...
+  }
+
+  .cx-text-green {
+    color: var(--cx-color-success);
+  }
+
+  .cx-text-red {
+    color: var(--cx-color-danger);
+  }
+}
```
Parthlakhani: This can be removed.

Parthlakhani:
Screen Shot 2022-09-27 at 12 27 32 PM

- `CHANGED ON` column and spacing overall looks a bit weird as compared to mockups. maybe.. it is because of the added padding on left. it should be added for headings as well.
- Table is not accessible as saved cart and order history, all fields of the tables should be accessible while tabbing.

Parthlakhani:
Screen Shot 2022-09-27 at 12 41 52 PM

Access assistant is throwing few errors, you can check from adding this [extension](https://chrome.google.com/webstore/detail/access-assistant/ojiighldhdmahfdnhfdebnpmlbiemdfm) to chrome.

Parthlakhani (approved):

feature-libs/customer-ticketing/components/customer-ticketing-list/customer-ticketing-list.component.ts
```diff
@@ -1,11 +1,159 @@
...
+    );
+  }
+
+  changeSortCode(sortCode: string): void {
+    // const event: { sortCode: string; currentPage: number } = {
+    //   sortCode,
+    //   currentPage: 0,
+    // };
+    this.sortType = sortCode;
+    // this.fetchTicketList(event);
```
khristian23: Please consider extracting the event creation to a different function
this.sortType = buildSortingEvent(sortCode)

feature-libs/customer-ticketing/components/customer-ticketing-list/customer-ticketing-list.component.ts
```diff
@@ -1,11 +1,159 @@
...
+  }
+
+  getStatusClass = (status: string): string => {
+    switch (status) {
+      case 'OPEN' || 'INPROCESS':
+        return 'cx-text-green';
+      case 'CLOSED':
+        return 'cx-text-grey';
+      default:
+        return '';
```
khristian23: Could OPEN, INPROCESS, CLOSED be constants values?
You could use:
import { STATUS, TEXT_COLOR_CLASS } from '@spartacus/customer-ticketing/root';

```suggestion
    
    switch (status) {
      case STATUS.OPEN || STATUS.INPROCESS:
        return TEXT_COLOR_CLASS.ACTIVE;
      case 'CLOSED':
        return TEXT_COLOR_CLASS.INACTIVE;
      default:
        return TEXT_COLOR_CLASS.DEFAULT
```

feature-libs/customer-ticketing/root/model/customer-ticketing.model.ts
```diff
@@ -1,3 +1,5 @@
+import { SortModel, PaginationModel } from '@spartacus/core';
+
 export const enum CUSTOM_CLASS {
   CLOSE = 'cx-text-gray',
   OPEN = 'cx-text-green',
```
khristian23:
```suggestion
export const enum TEXT_COLOR_CLASS {
  INACTIVE = 'cx-text-gray',
  ACTIVE = 'cx-text-green',
  DEFAULT = '';
```

khristian23 (changes requested):

feature-libs/customer-ticketing/components/customer-ticketing-list/customer-ticketing-list.component.ts
adolfobee: I considered this before but to keep it consistent with the rest of the components (Order history and saved cart) I preferred to do it the same way as the other tables

feature-libs/customer-ticketing/root/model/customer-ticketing.model.ts
adolfobee: I used Grey and Green because I don't want to add new variable names and assign them a color

feature-libs/customer-ticketing/components/customer-ticketing-list/customer-ticketing-list.component.ts
adolfobee: Enums cannot be assigned an empty string, and the switch case was added separately

feature-libs/customer-ticketing/components/customer-ticketing-list/customer-ticketing-list.component.ts
```diff
@@ -1,11 +1,161 @@
...
+              name: 'Open',
+            },
+          },
+        ],
+      },
+    ],
+  });
+
+  tickets = this.tickets$.pipe(
+    map((tickets) => {
```
khristian23: How about changing the pipe and subscribe returning variable to "result" or "ticketList" in order to avoid "tickets.tickets..."

feature-libs/customer-ticketing/components/customer-ticketing-list/customer-ticketing-list.component.ts
khristian23: If you considered it before it means you found it fishy as well :)
Let's keep the code DRY. We don't care how other components do in terms of quality.
Please have a look at the following recommendation:
https://github.com/ryanmcdermott/clean-code-javascript#functions-should-do-one-thing

khristian23 (changes requested):

feature-libs/customer-ticketing/components/customer-ticketing-list/customer-ticketing-list.component.ts
```diff
@@ -130,23 +130,25 @@ export class CustomerTicketingListComponent implements OnInit {
...
-    // const event: { sortCode: string; currentPage: number } = {
-    //   sortCode: this.sortType,
-    //   currentPage: page,
-    // };
-    console.log('pageChange', page);
+    this.createTicketListEvent(this.sortType, page);
     // this.fetchTicketList(event);
   }
 
+  createTicketListEvent(
```
khristian23, line 146:
```suggestion
  createTicketListSortEvent(
```

khristian23 (changes requested):

feature-libs/customer-ticketing/components/customer-ticketing-list/customer-ticketing-list.component.ts
```diff
@@ -1,11 +1,163 @@
...
+  }
+
+  changeSortCode(sortCode: string): void {
+    this.sortType = sortCode;
+    this.pageChange(0);
+  }
+
+  pageChange(page: number): void {
+    this.createTicketListEvent(this.sortType, page);
+    // this.fetchTicketList(event);
```
khristian23, line 143: Please remove

feature-libs/customer-ticketing/components/customer-ticketing-list/customer-ticketing-list.component.ts
```diff
@@ -1,11 +1,163 @@
...
+            createdAt: '2021-01-14T10:06:57+0000',
+            message: 'Door received broken',
+            toStatus: {
+              id: 'OPEN',
+              name: 'Open',
+            },
+          },
+        ],
+      },
+    ],
```
khristian23, line 97: Could this be replaced by a real call?

khristian23 (changes requested):

feature-libs/customer-ticketing/components/customer-ticketing-list/customer-ticketing-list.component.ts
adolfobee: This event is not only for sorting but also for changing pages so I prefer to leave the name as it is

feature-libs/customer-ticketing/components/customer-ticketing-list/customer-ticketing-list.component.ts
adolfobee: This will be used in the next PR with backend calls

feature-libs/customer-ticketing/components/customer-ticketing-list/customer-ticketing-list.component.ts
adolfobee: In the backend calls PR this will be removed

feature-libs/customer-ticketing/components/customer-ticketing-list/customer-ticketing-list.component.html
```diff
@@ -1,13 +1,143 @@
-<h2 class="cx-title d-none d-lg-block d-xl-block">
-  <!-- {{ 'title' | cxTranslate }} -->
-</h2>
+<ng-container
+  *ngIf="{
+    ticketsObs: tickets$ | async,
```
bgambocjaviniar: Why do you have these two in your ngIf?

Couldn't you just use `<ng-container *ngIf="ticket$ | async as ticketList">`, and use any properties from pagination, sorting, and tickets

bgambocjaviniar: the convention for observables is to suffix it with `$`, so you don't need to explicitly say Obs. I believe Parth mentioned that somewhere as well

feature-libs/customer-ticketing/components/customer-ticketing-list/customer-ticketing-list.component.html
```diff
@@ -1,13 +1,143 @@
...
+        (viewPageEvent)="pageChange($event)"
+      ></cx-pagination>
+    </div>
   </div>
-</ng-container>
 
-<ng-template>
-  <div class="cx-"></div>
-</ng-template>
+  <ng-container *ngIf="customerTicketsFlag">
```
bgambocjaviniar: You don't have to say `Flag`, and no need to hardcode it.  This should be taken from ticketList.tickets.length > 0; noCustomerTickets

bgambocjaviniar: same point here https://github.com/SAP/spartacus/pull/16059#pullrequestreview-1110829679

```
Let’s make *ngIf="tickets.length > 0" to *ngIf="tickets.length > 0; noCustomerTickets" and bind it to the ngTemplate of that reference. Reason for that, it may cover cases if ever the backend return <=0.
```

feature-libs/customer-ticketing/components/customer-ticketing-list/customer-ticketing-list.component.html
```diff
@@ -1,13 +1,143 @@
...
+    <table id="ticketing-list-table" class="table cx-ticketing-list-table">
```
bgambocjaviniar, line 36: any reason for adding the id? I don't see it being used as well. Would need to be prefixed with cx- if it was needed

feature-libs/customer-ticketing/components/customer-ticketing-list/customer-ticketing-list.component.html
```diff
@@ -1,13 +1,143 @@
...
+      <thead class="cx-ticketing-list-thead-mobile">
+        <tr>
```
bgambocjaviniar, line 38: nice 👍

feature-libs/customer-ticketing/components/customer-ticketing-list/customer-ticketing-list.component.html
```diff
@@ -1,13 +1,143 @@
...
+          <th scope="col">
+            {{ 'customerTicketingList.ticketCategory' | cxTranslate }}
+          </th>
+          <th scope="col">{{ 'customerTicketing.createdOn' | cxTranslate }}</th>
+          <th scope="col">{{ 'customerTicketing.changedOn' | cxTranslate }}</th>
+          <th scope="col">{{ 'customerTicketing.status' | cxTranslate }}</th>
+        </tr>
+      </thead>
+      <tbody>
+        <tr *ngFor="let ticket of ticketList.tickets">
```
bgambocjaviniar: don't forget to add a click event listener to the row as well as mentioned here https://github.com/SAP/spartacus/pull/16059#pullrequestreview-1087447665

feature-libs/customer-ticketing/components/customer-ticketing-list/customer-ticketing-list.component.html
```diff
@@ -1,13 +1,143 @@
...
+            <div class="cx-ticketing-list-label">
+              {{ 'customerTicketingList.mobile.ticketIdFull' | cxTranslate }}
+            </div>
+            <a
+              [routerLink]="
+                { cxRoute: 'supportTicketDetails', params: ticket.id } | cxUrl
+              "
+              class="cx-ticketing-list-value"
+            >
+              <span tabindex="0">{{ ticket.id }}</span>
```
bgambocjaviniar: do you really have to add a span tag that is being used entirely and wrapped under an a tag?

bgambocjaviniar: Moreover, tabindex were added in some a tag and some in tabindex, and sometimes in both.

You don't have to add the tab index if you just leave it wrapped under an a tag as it will focus there for being an a tag. All in all, remove the span and drop the tabindex in places, especially inconsistent ones like line 82 to 84 vs line 68 to 70

feature-libs/customer-ticketing/components/customer-ticketing-list/customer-ticketing-list.component.html
```diff
@@ -1,13 +1,143 @@
...
+              tabindex="0"
+            >
+              <span tabindex="0">{{ ticket.subject }}</span>
+            </a>
+          </td>
+          <td class="cx-ticketing-list-data">
+            <div class="cx-ticketing-list-label">
+              {{ 'customerTicketingList.ticketCategory' | cxTranslate }}
+            </div>
+            <span class="cx-ticketing-list-value" tabindex="0">
```
bgambocjaviniar: instead of a span, you should be using an a tag as mentioned here, all table data should be clickable to go to the details page https://github.com/SAP/spartacus/pull/16059#pullrequestreview-1087447665

feature-libs/customer-ticketing/components/customer-ticketing-list/customer-ticketing-list.component.html
```diff
@@ -1,13 +1,143 @@
...
+            <div class="cx-ticketing-list-label">
+              {{ 'customerTicketing.changedOn' | cxTranslate }}
+            </div>
+            <span class="cx-ticketing-list-value" tabindex="0">
+              {{ ticket.modifiedAt | cxDate }}
+            </span>
+          </td>
+          <td
+            [ngClass]="
+              getStatusClass(ticket.status.id) + ' cx-ticketing-list-data'
```
bgambocjaviniar: instead of binding additional classes like `cx-ticketing-list-data` like this, can't you just put it under class="cx-ticketing-list-data", and have the ngClass binding as `[ngClass]="(getStatusClass(ticket.status.id)"`

feature-libs/customer-ticketing/components/customer-ticketing-list/customer-ticketing-list.component.spec.ts
```diff
@@ -1,12 +1,139 @@
...
+  Output,
+  Pipe,
+  PipeTransform,
+} from '@angular/core';
 import { CustomerTicketingListComponent } from './customer-ticketing-list.component';
+import { I18nTestingModule } from '@spartacus/core';
+import { TicketList } from '@spartacus/customer-ticketing/root';
+import { Observable, of } from 'rxjs';
+
+const mockTicketList: TicketList = {
```
bgambocjaviniar, line 15: If specs aren't being worked on, you didn't need to add this mock data test as it's not even being used

feature-libs/customer-ticketing/components/customer-ticketing-list/customer-ticketing-list.component.ts
```diff
@@ -1,11 +1,163 @@
...
+          createdAt: ticket.createdAt,
+          modifiedAt: ticket.modifiedAt,
+        };
+      });
+    })
+  );
+
+  ngOnInit(): void {
+    this.tickets$.subscribe((tickets) => {
+      this.customerTicketsFlag = tickets && tickets.tickets?.length > 0;
```
bgambocjaviniar: you're already subscribing in template by using the async pipe, so this is just creating another subscription.

Moreover, if you were to go this route, you would need to handle the unsubscription to avoid memory leaks when the component gets destroyed. Utilizing the async pipe handles all these problems for you.

feature-libs/customer-ticketing/styles/components/_customer-ticketing-list.scss
```diff
@@ -0,0 +1,147 @@
...
+    @include media-breakpoint-down(sm) {
+      display: none;
+    }
+  }
+
+  .cx-ticketing-list-label {
+    color: var(--cx-color-secondary);
+
+    @include media-breakpoint-up(md) {
+      padding: 0 15px 0 0;
```
bgambocjaviniar: is this necessary when it's being 'hidden'?

feature-libs/customer-ticketing/styles/components/_customer-ticketing-list.scss
```diff
@@ -0,0 +1,147 @@
...
+
+  .cx-ticketing-list-data {
+    text-align: center;
+
+    @include media-breakpoint-down(sm) {
+      text-align: start;
+    }
+  }
+
+  .cx-ticketing-list-sort {
```
bgambocjaviniar, line 88: you can pull out 

```
    display: flex;
    margin: 0;
    @include media-breakpoint-down(sm) {
      flex-direction: column;
    }
```

to `.cx-ticketing-list-sort` 

as they're common to both .top and .bottom and being used for them

bgambocjaviniar (changes requested):

feature-libs/customer-ticketing/components/customer-ticketing-list/customer-ticketing-list.component.html
adolfobee: it is being used as `ariaControls` for the `cx-sorting`, the same way it is done in the order history table, I use the same naming convention as the other files to keep it consistent

feature-libs/customer-ticketing/components/customer-ticketing-list/customer-ticketing-list.component.html
adolfobee: The whole idea behind this was that only the first 2 fields would serve as links to the ticket details, I will replace it so the whole row is the link

feature-libs/customer-ticketing/components/customer-ticketing-list/customer-ticketing-list.component.spec.ts
adolfobee: it will be used in the next PR coming after this one! was simply adding it here

feature-libs/customer-ticketing/components/customer-ticketing-list/customer-ticketing-list.component.ts
adolfobee: Everything including the OnInit has been removed

feature-libs/customer-ticketing/styles/components/_customer-ticketing-list.scss
adolfobee: if not this causes layout problems

feature-libs/customer-ticketing/components/customer-ticketing-list/customer-ticketing-list.component.html
adolfobee: I'm not sure I understand this

feature-libs/customer-ticketing/components/customer-ticketing-list/customer-ticketing-list.component.html
bgambocjaviniar: it's the same 'idea' and concept as mentioned here https://github.com/SAP/spartacus/pull/16059#discussion_r992990153

where you should have it clickable

feature-libs/customer-ticketing/components/customer-ticketing-list/customer-ticketing-list.component.html
bgambocjaviniar:
```
<td
      class="cx-ticketing-list-data"
      [ngClass]="getStatusClass(ticket.status.id)"
>
...
...
</td>
```

feature-libs/customer-ticketing/styles/components/_customer-ticketing-list.scss
bgambocjaviniar: Hmmm have you tried it? 

I tried removing it and even adding padding: 100px to it. It won't do anything since you're literally 'hiding' the element for desktop

feature-libs/customer-ticketing/components/customer-ticketing-list/customer-ticketing-list.component.html
bgambocjaviniar: got it

feature-libs/customer-ticketing/styles/components/_customer-ticketing-list.scss
```diff
@@ -84,28 +84,39 @@
     }
   }
 
+  .cx-ticketing-list-data-extra {
```
bgambocjaviniar: There's no need to have this 'extra' class.

You can re-use the cx-ticketing-list-data as done before, but approach it using the way I mentioned here in terms of what you had with all the spans everywhere before https://github.com/SAP/spartacus/pull/16059#discussion_r993686347

feature-libs/customer-ticketing/styles/components/_customer-ticketing-list.scss
```diff
@@ -138,10 +149,12 @@
   }
 
   .cx-text-green {
+    text-decoration: none;
```
bgambocjaviniar: Will only flag this, but applies to any hovered table data. 

When hovering on the table data, it is normal for our table to show an underline, so please remove.

bgambocjaviniar (changes requested):
looks almost good about 95% there

- go through the comments I left again earlier

new things I just noticed:
- you should be able to navigate to the details page. (looking into this right now, but you should too)
- when it's empty, it's missing the title and count as seen in figma. Forget the sorting one (ignore it) as it doesn't make sense to display it when there's nothing. I'd like to see the title and count though
Screen Shot 2022-10-12 at 3 28 59 PM

feature-libs/customer-ticketing/styles/components/_customer-ticketing-list.scss
bgambocjaviniar: > Hmmm have you tried it?
> 
> I tried removing it and even adding padding: 100px to it. It won't do anything since you're literally 'hiding' the element for desktop

feature-libs/customer-ticketing/components/customer-ticketing-list/customer-ticketing-list.component.html
```diff
@@ -1,13 +1,176 @@
...
+      *ngIf="ticketList.pagination.totalPages > 1"
+      class="cx-ticketing-list-pagination"
+    >
+      <cx-pagination
+        [pagination]="ticketList.pagination"
+        (viewPageEvent)="pageChange($event)"
+      ></cx-pagination>
+    </div>
   </div>
+  <ng-container *ngIf="ticketList.tickets?.length > 0; else noCustomerTickets">
```
bgambocjaviniar, line 10: when it's empty, it's missing the title and count as seen in figma. `**Forget the sorting one (ignore it) as it doesn't make sense to display it when there's nothing. I'd like to see the title and count though**`

bgambocjaviniar (approved):

PR #16063, opened 2022-08-10
Add initial changes for create ticket feature

manpreetlamba4:
Jira Link: https://jira.tools.sap/browse/CXEC-9980

This PR contains UI changes required for Create New Ticket feature


feature-libs/customer-ticketing/components/customer-ticketing-create/customer-ticketing-create-dialog/customer-ticketing-create-dialog.component.ts
```diff
@@ -0,0 +1,106 @@
...
+      new FormControl('', [
+        Validators.required,
+        Validators.maxLength(this.inputCharactersLimitForCreateMessage),
+      ])
+    );
+    form.setControl(
+      'file',
+      new FormControl('', [
+        this.filesFormValidators.maxSize(this.maxSize),
+        // this.filesFormValidators.maçççxEntries(this.maxEntries),
```
khristian23: If not needed, please remove

feature-libs/customer-ticketing/components/customer-ticketing-create/customer-ticketing-create-dialog/customer-ticketing-create-dialog.component.ts
```diff
@@ -0,0 +1,106 @@
...
+
+  get maxSize(): number | undefined {
+    return this.customerTicketingConfig.customerTicketing
+      ?.attachmentRestrictions?.maxSize;
+  }
+
+  get maxEntries(): number | undefined {
+    return this.customerTicketingConfig.customerTicketing
+      ?.attachmentRestrictions?.maxEntries;
+  }
```
khristian23: nit: I feel these methods will be repeated in full or partially on all pages using the AddMessage and the Attachment
Don't you believe it will be a better approach to create a separate component or it  is not worthed? 
Since I don't know Angular, I don't know what would be the implications
@Parthlakhani

feature-libs/customer-ticketing/components/customer-ticketing-create/customer-ticketing-create.component.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  openCreateNewTicketDialog() {
+    const dialog = this.launchDialogService.openDialog(
+      LAUNCH_CALLER.CUSTOMER_TICKETING_CREATE,
+      this.element,
+      this.vcr
+    );
+    if (dialog) {
+      this.subscription.add(dialog.pipe(take(1)).subscribe());
+    }
```
khristian23, line 34: Is OpenDialog synchronous?
Asking because you are using dialog right after

feature-libs/customer-ticketing/root/config/default-customer-ticketing-config.ts
```diff
@@ -15,6 +15,7 @@ export const defaultCustomerTicketingConfig: CustomerTicketingConfig = {
         '.pdf',
       ],
     },
+    inputCharactersLimitForCreateMessage: 5000,
     inputCharactersLimit: 2000,
```
khristian23: Why don't use the constants defined in "feature-libs/customer-ticketing/core/customer-ticketing-constants.ts"
Otherwise, in case of change we need to modify two files.
@Parthlakhani

khristian23 (changes requested):

feature-libs/customer-ticketing/components/customer-ticketing-create/customer-ticketing-create-dialog/customer-ticketing-create-dialog.component.ts
Parthlakhani: yes, this methods will be used at 3 places excluding the shared messaging component.
It would be good to create a mechanism that returns these properties altogether from config.
Maybe, it can be done in the component service. `customer-ticketing/components/customer-ticketing.service.ts`

feature-libs/customer-ticketing/root/config/default-customer-ticketing-config.ts
Parthlakhani: yes, I think we can use the constants here as well.
I have one doubt though, would it affect matter that the configs are in root and the constants are in core?
@bgambocjaviniar

feature-libs/customer-ticketing/core/customer-ticketing-constants.ts
```diff
@@ -1 +1,2 @@
+export const MAX_INPUT_CHARACTERS_FOR_MESSAGE = 5000;
 export const MAX_INPUT_CHARACTERS = 2000;
```
Parthlakhani: Limit is 5000 on each place in mockups, I didn't check and put 2000 initially.
You can just modify `MAX_INPUT_CHARACTERS` to 5000.

Unless, we think we should provide different limit for input message and reopen/close message.
IMO, it should be same as we are treating it as just another message to create/close/reopen ticket.

Parthlakhani: Same thing for the input characters as well

khristian23: So, we need to remove one of the constants

feature-libs/customer-ticketing/components/customer-ticketing-create/customer-ticketing-create-dialog/customer-ticketing-create-dialog.component.ts
bgambocjaviniar: What we can do here is create a 'general customer ticketing dialog' that is abstract and that other components will extend to have these functionalities

feature-libs/customer-ticketing/root/config/default-customer-ticketing-config.ts
bgambocjaviniar: root are things that are made available from the get-go (eagerly loaded). So if this eagerly-loaded code that is available right away imports something from a lazy loaded part of the code, then it would break lazy loading

feature-libs/customer-ticketing/core/customer-ticketing-constants.ts
bgambocjaviniar: since these are just constants and will be re-used in root, let us move this in root. A few more variable won't save us more bytes

bgambocjaviniar: related to https://github.com/SAP/spartacus/pull/16063#discussion_r955323891

feature-libs/customer-ticketing/components/customer-ticketing-create/customer-ticketing-create-dialog/customer-ticketing-create-dialog.component.ts
Parthlakhani: correction to my comment:
It will be used at 2 places excluding the shared messaging component

feature-libs/customer-ticketing/root/config/default-customer-ticketing-config.ts
Parthlakhani: yes, so either we should not use the constants or we should include the constants in the root. right?

feature-libs/customer-ticketing/components/customer-ticketing-create/customer-ticketing-create-dialog/customer-ticketing-create-dialog.component.ts
bgambocjaviniar: An example would be `BaseMessageComponent`  ---> it's not related to chat component at all if ever :d

feature-libs/customer-ticketing/components/customer-ticketing-reopen/customer-ticketing-reopen-dialog/customer-ticketing-reopen-dialog.component.ts
```diff
@@ -11,7 +11,7 @@ import {
 } from '@spartacus/storefront';
 
 @Component({
-  selector: 'cx-customer-ticketing-reopen-dialog',
+  selector: 'cx-customer-ticketing-dialog',
```
bgambocjaviniar: this should have not been changed. Revert back to the original distinct selector

and in your create component, put it as -create-dialog.

You can make your stylesheet similar to this https://github.com/SAP/spartacus/pull/16063/files#diff-e4ba2751cc8990d33682351a9eec88aa6247136c383656b1290abd8150c25f4aR2

feature-libs/customer-ticketing/components/customer-ticketing-create/customer-ticketing-create-dialog/customer-ticketing-create-dialog.component.ts
bgambocjaviniar: Should be kept, but will be normalized into a 'base' dialog that everyone will derive from 

@Parthlakhani

feature-libs/customer-ticketing/styles/components/_customer-ticketing-dialog.scss
```diff
@@ -1,4 +1,4 @@
-%cx-customer-ticketing-reopen-dialog {
+%cx-customer-ticketing-dialog {
```
bgambocjaviniar, line 1: I understand re-using the stylesheet, but I don't think the button should be taking the full width based of the wireframes
Screen Shot 2022-08-25 at 3 33 16 PM

feature-libs/customer-ticketing/styles/components/_customer-ticketing-actions.scss
```diff
@@ -1,16 +1,18 @@
-%cx-customer-ticketing-close,
-%cx-customer-ticketing-reopen {
-  display: flex;
-  justify-content: flex-end;
+%cx-customer-ticketing-add {
```
bgambocjaviniar: what is `cx-customer-ticketing-add` as I don't see any component with that selector 

Moreover, it is wrapping the other placeholder selectors, which I don't think make sense. I believe this was changed accidentally?

feature-libs/customer-ticketing/core/index.ts
```diff
@@ -1,3 +1,5 @@
+export * from '../root/config/customer-ticketing-config';
```
bgambocjaviniar: Why is this being exported here in the core?

This is already being exported in `feature-libs/customer-ticketing/root/config/index.ts`

If ever it's needed to use something from that config, you would import the config as import { something } from '@customer-ticketing/root'

feature-libs/customer-ticketing/root/model/augmented-core.model.ts
```diff
@@ -2,6 +2,7 @@ import '@spartacus/storefront';
 
 declare module '@spartacus/storefront' {
   const enum LAUNCH_CALLER {
+    CUSTOMER_TICKETING_CREATE = 'CUSTOMER_TICKETING_CREATE',
```
bgambocjaviniar: 👍 

line 6 is something I missed that should be handled in the refactor of Parth's reopen/close PR

feature-libs/customer-ticketing/core/customer-ticketing-constants.ts
```diff
@@ -1 +1,2 @@
+export const MAX_INPUT_CHARACTERS_FOR_MESSAGE = 5000;
```
bgambocjaviniar: as mentioned by someone in a comment, let's remove this constant and set it to 5000

bgambocjaviniar: therefore removing the needing to have different limits https://github.com/SAP/spartacus/pull/16063/files#diff-a9e6c1c63dd6600ad490edb4ecfb7b650dffc06530b5abe6d7ed1526ed77a415R18

bgambocjaviniar (changes requested):
Thanks @khristian23 for https://github.com/SAP/spartacus/pull/16063/files#r954195915

I've always been contemplating, but decided to go ahead with the following

create under components a shared folder, which will contain a customerticketingdialogcomponent. That component will be an abstract class in which the other 3 dialogs will extend from in order to have the implementation and to ahere to DRY.

feature-libs/customer-ticketing/styles/components/_customer-ticketing-actions.scss
bgambocjaviniar: I believe it was intended for the -create component

bgambocjaviniar: I believe this was added with the intention of adding the -create component button.

Maybe it's better to create your own stylesheet and extending one of these as I don't think this button should be taking the full width of the container as shown in the wirefrrames

Screen Shot 2022-08-25 at 3 33 16 PM

manpreetlamba4: This is fixed with https://github.com/SAP/spartacus/pull/16063/commits/cfc3a2fd847515d99677e7b566ee3fa7b3e000a5

feature-libs/customer-ticketing/components/customer-ticketing-create/customer-ticketing-create.component.ts
manpreetlamba4: Yes..and also this is how it is being used at various other places too.

feature-libs/customer-ticketing/components/customer-ticketing-create/customer-ticketing-create.component.ts
```diff
@@ -0,0 +1,40 @@
...
+  OnDestroy,
+  ViewChild,
+  ViewContainerRef,
+} from '@angular/core';
+import { LaunchDialogService, LAUNCH_CALLER } from '@spartacus/storefront';
+import { Subscription } from 'rxjs';
+import { take } from 'rxjs/operators';
+
+@Component({
+  selector: 'cx-customer-ticketing-add',
```
bgambocjaviniar: the selector name is different from the component name

usually when you create a component through ng cli, they're uniform/similar. So I believe this should have been cx-customer-ticketing-create

feature-libs/customer-ticketing/styles/components/_customer-ticketing-actions.scss
```diff
@@ -1,3 +1,4 @@
+%cx-customer-ticketing-add,
```
bgambocjaviniar: Don't forget to change the selector here too

feature-libs/customer-ticketing/styles/components/_customer-ticketing-dialog.scss
```diff
@@ -38,6 +39,13 @@
         .cx-customer-ticket-input-hint {
           font-size: 0.75rem;
           font-style: italic;
+          text-align: end;
+          margin-bottom: 0;
+        }
+        .cx-customer-ticket-file-upload-hint {
```
bgambocjaviniar, line 45: I believe we can remove `.cx-customer-ticket-input-hint` that was done in Parth's PR as I noticed in the wireframes that they're all aligned at the start and not at the end. Therefore, update the class in the reopen component as well please! 
Screen Shot 2022-09-01 at 4 27 35 PM

manpreetlamba4: Have made changes in reopen component to use cx-customer-ticket-file-upload-hint class. We require '.cx-customer-ticket-input-hint' as well for 'Characters left' text at the bottom of message.

bgambocjaviniar (approved):

khristian23 (approved):

PR #16070, opened 2022-08-11
migration to yarn 3

pawelfras:


pawelfras:
Regarding thousand files in `.yarn/cache`, based on yarn [documentation](https://yarnpkg.com/getting-started/qa#which-files-should-be-gitignored):
```
.yarn/cache and .pnp.* may be safely ignored, but you'll need to run yarn install to regenerate them between each branch switch - which would be optional otherwise, cf [Zero-Installs](https://yarnpkg.com/features/zero-installs).
```
We have to decide whether we want to keep those files in repo (probably not).

PR #16075, opened 2022-08-15
CXSPA-752 [BOPIS] Implement Pickup functionality on PDP

SVendittelli:


feature-libs/pickup-in-store/components/presentational/store/store.component.html
```diff
@@ -42,19 +42,23 @@
     <div>
       <cx-icon
         class="cx-store-stock-icon"
-        *ngIf="!!storeDetails?.stockInfo?.stockLevel"
+        *ngIf="isInStock()"
```
JedidiahAwuku: This should be a variable instead of function call.

feature-libs/pickup-in-store/components/presentational/store/store.component.ts
```diff
@@ -25,4 +25,11 @@ export class StoreComponent {
   toggleOpenHours(): void {
     this.openHoursOpen = !this.openHoursOpen;
   }
+
+  isInStock(): boolean {
```
JedidiahAwuku: This should be replaced with a variable.

projects/storefrontapp/src/environments/environment.ts
```diff
@@ -23,4 +23,5 @@ export const environment: Environment = {
   cpq: buildProcess.env.CX_CPQ ?? false,
   digitalPayments: buildProcess.env.CX_DIGITAL_PAYMENTS ?? false,
   epdVisualization: buildProcess.env.CX_EPD_VISUALIZATION ?? false,
+  pickupInStore: buildProcess.env.CX_PICKUP_IN_STORE ?? false,
```
JedidiahAwuku: Default to true but leave a todo to change to false.

JedidiahAwuku (changes requested):

feature-libs/pickup-in-store/components/presentational/store/store.component.html
SVendittelli: Done

feature-libs/pickup-in-store/components/presentational/store/store.component.ts
SVendittelli: Done

projects/storefrontapp/src/environments/environment.ts
SVendittelli: Done

JedidiahAwuku (approved):

PR #16076, opened 2022-08-15
CXSPA-823: Improve breaking change migration handling automation for major versions

plabadie:
Closes: https://jira.tools.sap/browse/CXSPA-823


tools/breaking-changes/common.ts
```diff
@@ -4,6 +4,135 @@
...
+export const MAJOR_VERSION_DOC_HOME = `../../docs/migration/${NEW_MAJOR_VERSION}_0`;
+export const DELETED_API_COMMENTS_FILE_PATH = `${MAJOR_VERSION_DOC_HOME}/deleted-api.json`;
+export const DELETED_MEMBERS_COMMENTS_FILE_PATH = `${MAJOR_VERSION_DOC_HOME}/deleted-renamed-api-members.json`;
+export const RENAMED_API_LOOKUP_FILE_PATH = `${MAJOR_VERSION_DOC_HOME}/renamed-api.json`;
+export const MIGRATION_SCHEMATICS_HOME = `../../projects/schematics/src/migrations/${NEW_MAJOR_VERSION}_0`;
+
+// Shared Functions
+export function readBreakingChangeFile(): any {
+  const breakingChangesData = JSON.parse(
+    fs.readFileSync(BREAKING_CHANGES_FILE_PATH, 'utf-8')
```
bgambocjaviniar: this can be pulled out into one function instead for `readBreakingChangeFile, readDeletedApiCommentsFile, readDeletedMembersCommentsFile, readRenamedApiLookupFile` by passing the *_FILE_PATH(s)

tools/breaking-changes/generate-deleted.ts
```diff
@@ -51,9 +46,20 @@ fs.writeFileSync(
  */
 
 function getSchematicsData(apiElement: any): any {
+  const breakingChangeEntry = common.getTopLevelBreakingChangeEntry(
```
bgambocjaviniar: we got an entry of changeType deleted, but what is it supposed to do? (it's unused)

tools/breaking-changes/readme.md
```diff
@@ -2,6 +2,16 @@
 
 To make sure we get an accurate diff, do not update the dependencies of this tool unless you are sure that we don't need to compare the results with data generated with a previous version of the dependencies.
 
+# Configuration
```
bgambocjaviniar, line 5: not totally related to this file, but when I tried all steps and having the NEW_MAJOR_VERSION set to 5, I'm getting errors saying `data/5_0/breaking-changes.json`, but I believe `writeFileSync` should have generated it right? 

(I have the fake manually imported data-set in `docs/migration/5_0/renamed-api.json`

tools/breaking-changes/common.ts
bgambocjaviniar: but my assumption is you rather have it separated for semantics? 
- same case for (isElement*<etc>*) right?

tools/breaking-changes/generate-constructors.ts
```diff
@@ -5,14 +5,13 @@
...
-import * as fs from 'fs';
-import stringifyObject from 'stringify-object';
+import * as common from './common';
 
 /**
  * This script generated the constructor deprecation schematics entries.
  *
- * Input: A breaking changes file, likely `./data/breaking-changes.json`
- * Output: A file, `generate-constructors.out.ts`, that contains a ConstructorDeprecation[] array to paste over in the migration schematics code.
+ * Input: Breaking change data returned by readBreakingChangeFile().  Likely is is ./data/X_0/breaking-change.json.  The folder depends on the major version config.`
```
bgambocjaviniar, line 13: nit picking, but I wouldn't say the 'script' necessarily consumes an input. Not like before when it was taking the arguments right after calling the 'script', therefore this is just the default behavior of what the 'script' does, but does not consume an input directly

tools/breaking-changes/common.ts
```diff
@@ -87,7 +216,9 @@ export function getParameterDoc(
   if (functonElement.parameters?.length) {
     let parameterDoc = lineEnding;
     functonElement.parameters.forEach((parameter: any, index: number) => {
-      parameterDoc += `  ${parameter.name}: ${parameter.type}${
+      parameterDoc += `  ${parameter.name}${parameter.isOptional ? '?' : ''}: ${
```
bgambocjaviniar, line 204: 👍  I see what you did there for catering to optional params

tools/breaking-changes/common.ts
```diff
@@ -112,3 +243,147 @@ export function getElementCategory(apiElement: any): string {
...
+  outputData: any
+): void {
+  const templateHeader = fs.readFileSync(templateFilePath, 'utf-8');
+  const outputDataString = stringifyObject(outputData);
+  const outputFileContent = templateHeader + outputDataString + ';\n';
+  createFoldersForFilePath(outputFilePath);
+  fs.writeFileSync(outputFilePath, outputFileContent);
+}
+
+export function writeTextDataOutput(
```
bgambocjaviniar: nitpicking, but I'll just say it

I believe this is only used in the generate doc instead of being a 'common' function

Moreover, other instances in this file are helper functions to be used in the main common functions, so the 'helpers', I would set them as private or pull it out in a utils file 

ALL in all, this is just nitpicking, but not necessary to do

tools/breaking-changes/generate-deleted.ts
bgambocjaviniar: just a clean up

tools/breaking-changes/readme.md
bgambocjaviniar: needs a fix with the recursive option

tools/breaking-changes/generate-constructors.ts
bgambocjaviniar: no need `don't worry` :)

tools/breaking-changes/common.ts
plabadie: I agree.  I moved it to gen-doc.ts

tools/breaking-changes/common.ts
plabadie: I extracted the repeated code in a function, and the 4 existing read* functions delegate the call to it with the right path to keep the convenience of just calling the read function without having to know the right path constant to use.

tools/breaking-changes/readme.md
plabadie: Agreed.  Fixed.  The compare step now creates the output folder when it does not exist.

tools/breaking-changes/generate-doc.ts
```diff
@@ -128,17 +160,29 @@ function getMembersDoc(apiElement: any): string {
...
-          doc += `\n### ${memberBreakingChange.deletedMember.kind} ${memberBreakingChange.deletedMember.name} is removed.\n`;
-          doc += `\n${memberBreakingChange.migrationComment}\n`;
+          const memberMigrationComment = common.findDeletedMemberComment(
+            apiElement,
+            memberBreakingChange.changeElementName,
+            deletedMembersCommentData
+          );
+
+          doc += `\n### ${memberBreakingChange.old.kind} ${memberBreakingChange.old.name} is removed.\n`;
+          doc += `\n${memberMigrationComment}\n`;
```
bgambocjaviniar, line 185: what do you think about prefixing these 'additional' comments as shown below. (Valid to all instances in this generated doc automation - reason is it gets appended below the generated 'automatic' comments, but then there's this random text that appears below, which is weird (in my opinion))
```suggestion
          doc += `\nAdditional comments: ${memberMigrationComment}\n`;
```

docs/migration/6_0/renamed-api.json
```diff
@@ -0,0 +1,14 @@
+[
+  {
+    "apiElementName": "ExampleService",
+    "entryPoint": "@spartacus/feature/core",
+    "newApiElementName": "ExampleRenamedService",
+    "migrationComment": "This optional comment will appear in the generated markdown doc.  You can use it to add more info beyond what is automatically generated if needed.  As an example, it can be used if it's usefult to say why a api element was renamed.  If left empty, or if not present, this field will be ignored.  (note: the examples in this file won't appear in the migration material because the corresponding api element do not exist)"
```
bgambocjaviniar: 👍 not being a police on this, but let's change this `if it's usefult to say why a api element was renamed` to `if it's useful to say why an api element was renamed`

that's all for me flagging things like this

tools/breaking-changes/generate-doc.ts
bgambocjaviniar, line 182: what do you think about prefixing these 'additional' comments as shown below. 

(Valid to all instances in this generated doc automation - reason why I'm asking for the change is because it gets appended below the generated 'automatic' comments, but then there's this random text that appears below, which is weird (in my opinion))
```suggestion
          doc += `\nAdditional Comments: ${memberMigrationComment}\n`;
```

tools/breaking-changes/compare.ts
```diff
@@ -6,8 +6,7 @@
 
 import deepEqual from 'deep-equal';
 import * as fs from 'fs';
-import { getElementCategory, getSignatureDoc, printStats } from './common';
-
+import * as common from './common';
```
bgambocjaviniar, line 9: not related to this line, but on line 309 for the function `paramDiff`, it is no longer being used.

tools/breaking-changes/generate-doc.ts
bgambocjaviniar: Don't have to prefix those in the generated schematics comment though as they're in one line.

tools/breaking-changes/generate-moved.out.template
```diff
@@ -0,0 +1,5 @@
+import { RenamedSymbol } from '../../../../shared/utils/file-utils';
+//
+// Generated file, don't update manually.
+//
+export const GENERATED_RENAMED_SYMBOLS_DATA: RenamedSymbol[] =
```
bgambocjaviniar, line 5: related to 3. from request change

```
import { RenamedSymbol } from '../../../../shared/utils/file-utils';
//
// Generated file, don't update manually.
//

export function migrate(): Rule {
  return (tree: Tree) => {
    return migrateRenamedSymbols(tree, GENERATED_RENAMED_SYMBOLS_DATA);
  };
}

export const GENERATED_RENAMED_SYMBOLS_DATA: RenamedSymbol[] = 
```

bgambocjaviniar: applies to all other .out.template

tools/breaking-changes/compare.ts
```diff
@@ -16,6 +15,7 @@ const newApiFile = process.argv[3];
 
 const newApiData = JSON.parse(fs.readFileSync(newApiFile, 'utf-8'));
 const oldApiData = JSON.parse(fs.readFileSync(oldApiFile, 'utf-8'));
+const renamedApiLookupData = common.readRenamedApiLookupFile();
```
bgambocjaviniar, line 18: Just to confirm this is not a valid use case.

1. someone forgets to include the breaking change in the renamed-api.json as part of there DoD in the PR
2. we already generated the migrations, so like we ran yarn compare to generate the breaking-change.json
3. after some time the person who forgot to include the BR includes it, and then includes it in the renamed-api.json

even if we `yarn gen-doc`, it won't work due to the breaking-change.json not having the component/class marked as one that is moved, but still shown as 'deleted'.

So the question is, is it fine for us to `yarn compare` again manually to get the truthy breaking-change.json file.
I was thinking if we do `yarn gen-doc`, it should take into account 'new' added changes instead of having to manually run the `yarn compare` command before running the schematics migration package script again.

What do you think of running `yarn compare` at every schematics/docs package script?

----
Reason why I'm saying this is because if we add changes to deleted-api.json, deleted-renamed-api-members.json, then those are taken into account from the get-go. If those are taken into account right away, then the ones that were marked as deleted, but are supposed to be moved, should have the same behavior. What do you think?

tools/breaking-changes/generate-moved.out.template
bgambocjaviniar: applies to all other .out.template

bgambocjaviniar (changes requested):
1. some cleanup from the escaped single quote is needed, which appends the 'backslash \\' as shown in the screenshots
Screen Shot 2022-09-11 at 11 06 13 PM
Screen Shot 2022-09-11 at 11 45 51 PM

2. not really a problem, but just letting you know. Old return types that were implicit are being put in the generated docs due to the fact that an explicit type was set to it now.
Screen Shot 2022-09-12 at 12 09 41 AM

3. As these 'generated'-schematics/docs are created, they are not 'linked/mapped' to the actual schematics migration mechanism. 
- I believe to fix this problem, you should include them as part of the templates per 'schematics'. For instance, 
```
import { RenamedSymbol } from '../../../../shared/utils/file-utils';
//
// Generated file, don't update manually.
//

export function migrate(): Rule {
  return (tree: Tree) => {
    return migrateRenamedSymbols(tree, GENERATED_RENAMED_SYMBOLS_DATA);
  };
}

export const GENERATED_RENAMED_SYMBOLS_DATA: RenamedSymbol[] = 
```

docs/migration/6_0/renamed-api.json
plabadie: No worries.  And it's fixed.

tools/breaking-changes/compare.ts
plabadie: It's important for the renamed lookup to be performed by the compare script because it is a pre-requisite for comparison.  When an api element is renamed, it's only via the rename lookup info that the compare script can compare the previous and current versions of the renamed api element.   Therefore we can't delay taking into account the rename lookup to after the compare.  If I didn't quite understand your point let's discuss.

Also, in general, it would be nice if we can eliminate manual interventions in breaking-changes.json.  This way we could re-generate the breaking changes file to be able to react to a change in scope of the release.  Perhaps we find a way to add the new random migration comments in a separate file.

tools/breaking-changes/compare.ts
plabadie: Fixed.  paramDiff removed.

tools/breaking-changes/generate-doc.ts
plabadie: For the deleted migration comments, I think the title above the comment provides enough context for the migration comment not to look randomly placed:

```
### Method removeEntry is removed.

The method is not used anymore.  Use that other method instead.
```

With the prefix it would look like this:
```
### Method removeEntry is removed.

Additional comment: The method is not used anymore.  Use that other method instead.
```

In the end, it's not a change that is long to implement, but let's discuss what we want to do about it.

docs/migration/6_0/deleted-api.json
```diff
@@ -0,0 +1,12 @@
...
+    "apiElementName": "SomeService",
+    "entryPoint": "@spartacus/feature/core",
+    "migrationComment": "Use class OtherService instead.  (note: this example won't appear in the migration material because the SomeService class does not exist)"
+  },
+  {
+    "apiElementName": "AwesomeClass",
+    "entryPoint": "@spartacus/feature/components",
+    "migrationComment": "This class is not needed anymore.  Things are done in this way instead.  (note: this example won't appear in the migration material because the AwesomeClass class does not exist)"
+  }
+]
```
plabadie: > some cleanup from the escaped single quote is needed, which appends the 'backslash \' as shown in the screenshots

The quotes are escaped because the string is delimited by the same kind of quotes.  It's the choice of the api used to stringily the object array.

plabadie: > not really a problem, but just letting you know. Old return types that were implicit are being put in the generated docs due to the fact that an explicit type was set to it now.

Indeed.  This is one of the limits of the tool.  If return types are different they will be listed amongst breaking changes.  To set correct expectations, at the top of the md doc file we announce that we have a list of breaking changes or _potentially_ breaking changes.

plabadie: > As these 'generated'-schematics/docs are created, they are not 'linked/mapped' to the actual schematics migration mechanism.

It's not a bad idea.  However, we usually bootstrap the migration schematics one per major release.  It's a short one shot manual operation.  Moreover we still need to bootstrap the schematics manually because there are more schematics and more files involved than these.
The scope of the breaking change tool was rather to focus on the output it is directly responsible for producing and eliminate the annoying copy/past that needed to happen for the generated schematics to be where they should.  
Moreover, the files you mention can potentially be updated by multiple devs if they need to add manually written schematics.
It doesn't seem it's worth it to generate these files automatically, but we can discuss it.
Also I might not have understood the suggestion properly.

bgambocjaviniar (dismissed):

bgambocjaviniar (approved):
LGTM~

PR #16077, opened 2022-08-16
CXEC-15457: add fields to order details page for buyer and unit

I565452:


giancorderoortiz:
I don't see Cypress e2e tests for this feature

PR #16085, opened 2022-08-16
CXEC-14762: UI for messaging and file upload

Parthlakhani:
https://jira.tools.sap/browse/CXEC-14762

Created shared messaging and avatar component
Using with customer ticketing messaging component
Added validation for file upload


feature-libs/customer-ticketing/components/customer-ticketing.service.ts
```diff
@@ -1,12 +1,14 @@
...
 import { Observable, of } from 'rxjs';
 import { map } from 'rxjs/operators';
+import { CustomerTicketingConfig } from '../root';
 
 @Injectable({
   providedIn: 'root',
 })
 export class CustomerTicketingService {
-  constructor() {}
+  constructor(protected customerTicketingConfig: CustomerTicketingConfig) {}
```
bgambocjaviniar: I understand this is mock data to be used for the UI, but shouldn't you include this in `feature-libs/customer-ticketing/core/facade/customer-ticketing.service.ts` instead, so that we can delete this file

feature-libs/customer-ticketing/components/customer-ticketing.service.ts
```diff
@@ -72,6 +74,22 @@ export class CustomerTicketingService {
   }
 
   getTicketStatus(): Observable<string> {
-    return this.ticketDetails$.pipe(map((details) => details.status.name));
+    return this.ticketDetails$.pipe(map((details) => details.status.id));
   }
+
+  prepareMessageDetails = (): Observable<MessageDetails> =>
```
bgambocjaviniar: out of curiosity, why the change of 'written' function as done above to now an arrow function?

again, I understand this is mock data, but why? and in the end, it will be put in the core/facade

projects/storefrontlib/shared/components/chat-messaging/avatar/avatar.component.html
```diff
@@ -0,0 +1,15 @@
+<!-- TODO: Verify current user to display left aligned messages -->
+<div
+  *ngIf="event"
+  class="cx-avatar"
+  [ngClass]="event.author !== 'Mark Rivers' ? 'right-align' : ''"
```
bgambocjaviniar: instead of hardcoding, would you like to fix this logic in this PR or more todos for later?

I would think it would be done in this PR since it's the UI part and not necessarily related to data

projects/storefrontlib/shared/components/chat-messaging/avatar/avatar.component.spec.ts
```diff
@@ -0,0 +1,24 @@
+import { ComponentFixture, TestBed } from '@angular/core/testing';
+
+import { AvatarComponent } from './avatar.component';
+
+describe('AvatarComponent', () => {
```
bgambocjaviniar, line 13: I'm assuming the specs will be done during the 'core' services implementation right?

projects/storefrontlib/shared/components/chat-messaging/avatar/avatar.component.html
bgambocjaviniar: Is there a distinct value given from the events of customer ticketing that says the person is support or customer because looking at the mocked data you provided as a test, I don't see any. But, maybe the actual response from the server provides greater details?

If not, you should potentially look into the leveraging a 'user' service (profile/account) to match the current logged in user. If he isn't, then he's support

projects/storefrontlib/shared/components/chat-messaging/avatar/avatar.component.ts
```diff
@@ -0,0 +1,21 @@
...
+  @Input() event: MessageEvent;
+  iconTypes = ICON_TYPE;
+
+  constructor() {}
+
+  ngOnInit(): void {}
+
+  getIntitials(name: string) {
+    const names = name.split(' ');
+    return `${names[0]?.split('')[0]}${names[1] ? names[1]?.split('')[0] : ''}`;
```
bgambocjaviniar: this is pretty hard to read 

what do you think of this apparoach

```
 return name.split(' ') --> to separate any spaced names as that is the norm I believe
     .map(name => name[0]) --> transform the array containing the names to only have the first letter
    .join(''); --> to put the 'array' into a string

```

projects/storefrontlib/shared/components/chat-messaging/messaging/messaging.component.html
```diff
@@ -0,0 +1,80 @@
+<ng-container *ngIf="messageDetails$ | async as messageDetails">
+  <div class="container">
+    <div class="cx-avatar-line"></div>
+    <div class="cx-messages" id="cx-messages">
+      <ng-container *ngFor="let event of messageDetails.messages">
+        <div
+          class="cx-message-card"
+          [ngClass]="{ 'cx-message-display': event.author !== 'Mark Rivers' }"
```
bgambocjaviniar: 🔴  (hardcoded name) even if it's using mocks

feature-libs/customer-ticketing/components/customer-ticketing-messages/customer-ticketing-messages.component.ts
```diff
@@ -0,0 +1,81 @@
...
+  });
+
+  messageDetails$: Observable<MessageDetails> =
+    this.customerTicketingService.prepareMessageDetails();
+
+  messagingConfigs: MessagingConfigs =
+    this.customerTicketingService.prepareMessagingConfigs();
+
+  onSend(_event: { files: FileList | undefined; message: string }) {
+    // call to submit new event and upload attachment
```
bgambocjaviniar, line 86: I understand you're using 'mocks', but couldn't this be done in a way that when you send, you basically append to the current string temporarily.

just nitpicking instead of having to add additional entries myself :d and to test things out

projects/storefrontstyles/scss/components/misc/chat-messaging/_avatar.scss
```diff
@@ -0,0 +1,25 @@
+%cx-avatar {
```
bgambocjaviniar, line 1: the styles looks fine to me

I don't see a problem, but since it's for a shared component, where it's going to be used across different featuress (potentially), then send this PR to Miguel/Mikel if they like what they see

bgambocjaviniar (changes requested):
Not related to the PR, but maybe let's restructure the components folder in an another ticket or in this one (up to u)

Because if someone were to go to the code, I don't think they would understand where the messaging component is being used and such.

An example to what I'm leaning towards is the order-details filestructure. It's laid out and 'packaged' in a proper way, where one can tell that all those sub folders is part of the order details


-----
I didn't finish reviewing as there's still some things I need to look into in the PR, but overall, `great job` Parth!

feature-libs/customer-ticketing/components/customer-ticketing.service.ts
Parthlakhani: The purpose of this service is to put helper functions for the customer ticketing. Similar to what's done in saved cart details. Also, I renamed the service to customer-ticketing-data-service in the next PR.

Parthlakhani:
> Not related to the PR, but maybe let's restructure the components folder in an another ticket or in this one (up to u)
> 
> Because if someone were to go to the code, I don't think they would understand where the messaging component is being used and such.
> 
> An example to what I'm leaning towards is the order-details filestructure. It's laid out and 'packaged' in a proper way, where one can tell that all those sub folders is part of the order details
> 
> I didn't finish reviewing as there's still some things I need to look into in the PR, but overall, `great job` Parth!

Yes, I agree! and thank you!! :) 

I was thinking we can divide the feature into 2 parts,

Ticket List
- CustomerTicketingList
- CustomerTicketingCreate

Ticket Details
- CustomerTIcketingReopen
- CustomerTIcketingClose
- CustomerTIcketingDetails
- CustomerTIcketingMessages
- CustomerTicketingDialog

what do you think?

feature-libs/customer-ticketing/components/customer-ticketing.service.ts
Parthlakhani: haha, no particular reason. I changed other two methods as well in the next PR

projects/storefrontlib/shared/components/chat-messaging/avatar/avatar.component.html
Parthlakhani: yes, I was thinking the same to use the user profile but there were some confusions related to what you mentioned. need to discuss this with Christian and you. and yes, let's add this in this PR only.

projects/storefrontlib/shared/components/chat-messaging/avatar/avatar.component.spec.ts
Parthlakhani: I think we can at least add the specs for shared component with this PR

projects/storefrontlib/shared/components/chat-messaging/avatar/avatar.component.ts
Parthlakhani: yup, that's better.

projects/storefrontlib/shared/components/chat-messaging/messaging/messaging.component.html
Parthlakhani: it will be handled with the flag that matches the author with current user

projects/storefrontlib/shared/components/chat-messaging/avatar/avatar.component.html
Parthlakhani: Had a discussion with Christian and found that `addedByAgent` flag will be there when the message is from admin.
We can define the flag based on that and we won't need to retrieve user profile.

feature-libs/customer-ticketing/components/customer-ticketing-messages/customer-ticketing-messages.component.html
```diff
@@ -0,0 +1,5 @@
+<cx-messaging
```
khristian23, line 1: Is the name "customer-ticketing-messages" plural because it handles multiple messages? If not, maybe singular would be better

projects/storefrontlib/shared/components/chat-messaging/avatar/avatar.component.html
```diff
@@ -0,0 +1,12 @@
+<!-- TODO: Verify current user to display left aligned messages -->
+<div
+  *ngIf="event"
```
khristian23: This is a shared component that uses properties from a Customer Ticketing Event -> too coupled.
Let's change "event" by  "message"

projects/storefrontlib/shared/components/chat-messaging/avatar/avatar.component.html
```diff
@@ -0,0 +1,12 @@
...
+  class="cx-avatar"
+  [ngClass]="{ 'right-align': event.addedByAgent }"
```
khristian23: Let's give the power to the hosting app to set its alignment value
```suggestion
  [ngClass]="{ 'right-align': message.rightAlign }"
```

projects/storefrontlib/shared/components/chat-messaging/avatar/avatar.component.html
```diff
@@ -0,0 +1,12 @@
...
+>
+  <h1 *ngIf="!event.addedByAgent">
+    {{ getIntitials(event?.author) }}
+  </h1>
+
+  <cx-icon *ngIf="event.addedByAgent" [type]="iconTypes.HEADSET"></cx-icon>
```
khristian23:
```suggestion
  <cx-icon *ngIf="message.rightAlign" [type]="iconTypes.HEADSET"></cx-icon>
```

projects/storefrontlib/shared/components/chat-messaging/avatar/avatar.component.html
```diff
@@ -0,0 +1,12 @@
+<!-- TODO: Verify current user to display left aligned messages -->
+<div
+  *ngIf="event"
+  class="cx-avatar"
+  [ngClass]="{ 'right-align': event.addedByAgent }"
+>
+  <h1 *ngIf="!event.addedByAgent">
+    {{ getIntitials(event?.author) }}
```
khristian23: no need to add "?"  
Again, let's give the hosting component to provide the name or the email
```suggestion
    {{ getIntitials(message.author) }}
```

projects/storefrontlib/shared/components/chat-messaging/avatar/avatar.component.ts
```diff
@@ -0,0 +1,23 @@
+import { Component, Input, OnInit } from '@angular/core';
+import { ICON_TYPE } from '@spartacus/storefront';
+import { MessageEvent } from '../messaging';
+
+@Component({
+  selector: 'cx-avatar',
+  templateUrl: './avatar.component.html',
+})
+export class AvatarComponent implements OnInit {
+  @Input() event: MessageEvent;
```
khristian23: I suggest changing the model name.  Quotes feature will use this component and they do not know the concept of Event
```suggestion
  @Input() message: Message;
```

projects/storefrontlib/shared/components/chat-messaging/avatar/avatar.component.ts
```diff
@@ -0,0 +1,23 @@
...
+  iconTypes = ICON_TYPE;
+
+  constructor() {}
+
+  ngOnInit(): void {}
+
+  getIntitials(name: string) {
+    return name
```
khristian23: What if the input is undefined?

projects/storefrontlib/shared/components/chat-messaging/avatar/avatar.component.ts
```diff
@@ -0,0 +1,23 @@
...
+})
+export class AvatarComponent implements OnInit {
+  @Input() event: MessageEvent;
+  iconTypes = ICON_TYPE;
+
+  constructor() {}
+
+  ngOnInit(): void {}
+
+  getIntitials(name: string) {
```
khristian23: It could be a name, fullname or email ( depends on the hosting component )
+ typo "Initials"
```suggestion
  getInitials(author: string) {
```

projects/storefrontlib/shared/components/chat-messaging/messaging/messaging.component.html
```diff
@@ -0,0 +1,80 @@
+<ng-container *ngIf="messageDetails$ | async as messageDetails">
+  <div class="container">
+    <div class="cx-avatar-line"></div>
+    <div class="cx-messages" id="cx-messages">
+      <ng-container *ngFor="let event of messageDetails.messages">
```
khristian23:
```suggestion
      <ng-container *ngFor="let message of messageDetails.messages">
```

projects/storefrontlib/shared/components/chat-messaging/messaging/messaging.component.html
```diff
@@ -0,0 +1,80 @@
...
+        <div
+          class="cx-message-card"
+          [ngClass]="{ 'cx-message-display': event.addedByAgent }"
+        >
+          <cx-avatar [event]="event"> </cx-avatar>
```
khristian23:
```suggestion
          <cx-avatar [message]="message"> </cx-avatar>
```

projects/storefrontlib/shared/components/chat-messaging/messaging/messaging.component.html
```diff
@@ -0,0 +1,80 @@
...
+
+          <div>
+            <label>{{
+              event.createdAt | cxDate: 'MMMM d, YYYY h:mm aa'
+            }}</label>
+            <div
+              [ngClass]="
+                event.addedByAgent
```
khristian23:
```suggestion
                message.rightAlign
```

projects/storefrontlib/shared/components/chat-messaging/messaging/messaging.component.html
```diff
@@ -0,0 +1,80 @@
...
+                  ? 'cx-message-right-align-text'
+                  : 'cx-message-left-align-text'
+              "
+            >
+              {{ event.message }}
+            </div>
+            <p *ngIf="event.attachments" class="cx-attachment">
+              <cx-icon [type]="iconTypes.ATTACHMENT"></cx-icon>
+              {{ event.attachments[0].filename }}
+            </p>
```
khristian23: Let's code it so that we receive multiple attachments and not only one

projects/storefrontlib/shared/components/chat-messaging/messaging/messaging.component.html
```diff
@@ -0,0 +1,80 @@
...
+          multiple="maxEntries > 1"
+        >
+          <ng-template>
+            <cx-icon [type]="iconTypes.UPLOAD"></cx-icon>
+            <span class="cx-message-footer-text">{{
+              'chatMessaging.uploadFile' | cxTranslate
+            }}</span>
+          </ng-template>
+        </cx-file-upload>
+
```
khristian23: Quotes do not provide a file upload option:
https://www.figma.com/file/NkJedfiEuQtpes9n8mnpsQ/SPA---Customer-Quotes?node-id=2452%3A15241
Can we have a configuration flag to decide whether or not to show the upload option?

projects/storefrontlib/shared/components/chat-messaging/messaging/messaging.component.ts
```diff
@@ -0,0 +1,113 @@
...
+
+  @Output()
+  send = new EventEmitter<{ files: FileList | undefined; message: string }>();
+
+  iconTypes = ICON_TYPE;
+  form: FormGroup;
+
+  MAX_INPUT_CHARACTERS: number = 2000;
+  MAX_SIZE: number = 10;
+  MAX_ENTRIES: number = 1;
```
khristian23, line 26: nit:  Shouldn't these be taken from a constant file?
I am OK is these are hardcoded here

projects/storefrontlib/shared/components/chat-messaging/messaging/messaging.component.ts
```diff
@@ -0,0 +1,113 @@
...
+      const element = this.windowRef.document.getElementById('cx-messages');
+      element?.scroll({
+        top: element?.scrollHeight,
+        behavior: 'auto',
+      });
+      setTimeout(() => {
+        this.windowRef.document
+          .getElementById('cx-message-footer')
+          ?.scrollIntoView({ behavior: 'smooth', block: 'end' });
+      }, 500);
```
khristian23: What guarantees that after 500ms the list of messages is loaded?
Is there a better way?  like executing after a promise is resolved?

projects/storefrontlib/shared/components/chat-messaging/messaging/messaging.component.ts
```diff
@@ -0,0 +1,113 @@
...
+    }
+  }
+
+  getIntitials(name: string) {
+    const names = name.split(' ');
+    return `${names[0]?.split('')[0]}${names[1] ? names[1]?.split('')[0] : ''}`;
+  }
+
```
khristian23: Is this still used? or it has been moved inside the avatar component?

projects/storefrontlib/shared/components/chat-messaging/messaging/messaging.model.ts
```diff
@@ -0,0 +1,28 @@
...
+}
+
+export interface MessageEvent {
+  author?: string;
+  addedByAgent?: boolean;
+  createdAt?: string;
+  message?: string;
+  attachments?: Array<Attachment>;
+}
+
```
khristian23: It is too coupled in concept with Ticketing Event.
How about:
```suggestion
export interface Message {
  author?: string;
  leftAlign?: boolean;
  createdAt?: string;
  text?: string;
  attachments?: Array<Attachment>;
}

```

khristian23 (changes requested):

feature-libs/customer-ticketing/components/customer-ticketing-messages/customer-ticketing-messages.component.html
Parthlakhani: yeah, that's what I thought. since it's messages and not just one message

projects/storefrontlib/shared/components/chat-messaging/avatar/avatar.component.html
Parthlakhani: yes, discussed something similar with Brian as well.
We can do this, but if the messageEvent is standard at SAP level, and the messageEvent would be similar in other cases as well.
We can keep the check for addedByAgent as well, which makes the rightAlign optional to pass.
What do you think @khristian23 ?

khristian23: Let me ask the Quotes team how they receive their entities in the UI.  If they also have the concept of an Event, then it will be ok.  Otherwise, we need to find a more generic naming

projects/storefrontlib/shared/components/chat-messaging/avatar/avatar.component.html
Parthlakhani: got it

projects/storefrontlib/shared/components/chat-messaging/avatar/avatar.component.html
Parthlakhani: Alright

projects/storefrontlib/shared/components/chat-messaging/avatar/avatar.component.ts
Parthlakhani: adding a check for undefined

Parthlakhani: should we return anything specific instead of null or ' '?

projects/storefrontlib/shared/components/chat-messaging/avatar/avatar.component.ts
Parthlakhani: Yup, seems better.

projects/storefrontlib/shared/components/chat-messaging/avatar/avatar.component.ts
khristian23: if we return an empty string, then we will be having an empty circle as avatar.
Since this information is only for messages created by the customer, then we could return "C" as a customer or better if you find an icon that represents a customer

Parthlakhani: any related icon is not there, let's just show C as these seems like a very rare case rather than adding new icon

projects/storefrontlib/shared/components/chat-messaging/messaging/messaging.component.html
Parthlakhani: Yes!

projects/storefrontlib/shared/components/chat-messaging/messaging/messaging.component.html
Parthlakhani: yes, adding it.

projects/storefrontlib/shared/components/chat-messaging/messaging/messaging.component.ts
Parthlakhani: We can not use the customer ticketing constant file. I decided to declare it here as it's not being used outside of this component. what do you suggest?

projects/storefrontlib/shared/components/chat-messaging/messaging/messaging.component.ts
Parthlakhani: It's hard to say,
but we can try doing something when we integrate the API by passing the state maybe.

projects/storefrontlib/shared/components/chat-messaging/messaging/messaging.component.ts
Parthlakhani: it's already removed. need to push the changes.

projects/storefrontlib/shared/components/chat-messaging/messaging/messaging.component.ts
khristian23: I am Ok with it.  the hosting component can inject their values using the message config anyways 👍

projects/storefrontlib/shared/components/chat-messaging/messaging/messaging.model.ts
Parthlakhani: yup, seems more generic.

projects/storefrontlib/shared/components/chat-messaging/avatar/avatar.component.ts
```diff
@@ -1,23 +1,23 @@
...
-      .split(' ')
-      .map((name) => name[0])
-      .join('');
+  getInitials(author: string) {
+    return author
+      ? author
+          ?.split(' ')
+          .map((author) => author[0])
+          .join('')
+      : 'C';
```
khristian23: const FROM_CUSTOMER = 'C';

feature-libs/customer-ticketing/components/customer-ticketing.service.ts
```diff
@@ -69,27 +73,33 @@ export class CustomerTicketingService {
...
       )
     );
+  }
 
   prepareMessagingConfigs = (): MessagingConfigs => ({
     attachmentRestrictions:
       this.customerTicketingConfig.customerTicketing?.attachmentRestrictions,
     charactersLimit:
       this.customerTicketingConfig.customerTicketing?.inputCharactersLimit,
+    uploadFile: true,
```
khristian23:
```suggestion
    enableFileUploadOption: true,
```

projects/storefrontlib/shared/components/chat-messaging/messaging/messaging.model.ts
```diff
@@ -19,6 +15,7 @@ export interface MessagingConfigs {
   attachmentRestrictions?: AttachmentRestrictions;
   charactersLimit?: number;
   newMessagePlaceHolder?: string;
+  uploadFile?: boolean;
```
khristian23:
```suggestion
  enableFileUploadOption?: boolean;
```

khristian23 (changes requested):

projects/storefrontlib/shared/components/chat-messaging/avatar/avatar.component.ts
```diff
@@ -0,0 +1,25 @@
...
+export class AvatarComponent {
+  @Input() message: MessageEvent;
+  iconTypes = ICON_TYPE;
+
+  constructor() {}
+
+  getInitials(author: string) {
+    return author
+      ? author
+          ?.split(' ')
```
khristian23: We no longer need it to be optional, since it is being validated in the first part
```suggestion
          .split(' ')
```

projects/storefrontlib/shared/components/chat-messaging/messaging/messaging.component.html
```diff
@@ -0,0 +1,84 @@
...
+      <ng-container *ngFor="let message of messageEvents">
+        <div
+          class="cx-message-card"
+          [ngClass]="{ 'cx-message-display': message.rightAlign }"
+        >
+          <cx-avatar [message]="message"> </cx-avatar>
+
+          <div>
+            <label>{{
+              message.createdAt | cxDate: 'MMMM d, YYYY h:mm aa'
```
khristian23: Does this format need to be used on all customer ticketing pages?
If so, then can we extract to a central customer ticketing formatter?
Even mobile view needs the same format
@adolfobee @manpreetlamba4

khristian23 (changes requested):

projects/storefrontlib/shared/components/chat-messaging/messaging/messaging.component.html
Parthlakhani: Yes, it will be used at 2 places in customer ticketing. we can use a constant for that.
I am adding that into the PR for abstract dialog component.

Parthlakhani: we can declare the same constant here as well.

Parthlakhani: Also, adding dateFormat to config so it can be passed from parent

projects/storefrontlib/shared/components/chat-messaging/messaging/messaging.component.ts
```diff
@@ -31,6 +31,7 @@ export class MessagingComponent implements OnInit, AfterViewInit {
   MAX_INPUT_CHARACTERS: number = 2000;
   MAX_SIZE: number = 10;
   MAX_ENTRIES: number = 1;
+  dateFormat: string = 'MMMM d, YYYY h:mm aa';
```
khristian23, line 27:
```suggestion
  DATE_FORMAT: string = 'MMMM d, YYYY h:mm aa';
```

projects/storefrontlib/shared/components/chat-messaging/messaging/messaging.model.ts
```diff
@@ -16,6 +16,7 @@ export interface MessagingConfigs {
   charactersLimit?: number;
   newMessagePlaceHolder?: string;
   enableFileUploadOption?: boolean;
+  dateFormat?: string;
```
khristian23, line 19: Can the hosting component set the format?
It will be preferable if that is the case so that customer ticketing can set a general date format that will affect all its pages.

khristian23 (changes requested):

projects/storefrontlib/shared/components/chat-messaging/messaging/messaging.model.ts
Parthlakhani: Yes, hosting component can pass the `dateFormat`. if it's not passed, we are using the constant we defined in messaging component.

Parthlakhani: We did all the changes required for dateFormat in this [PR](https://github.com/SAP/spartacus/pull/16113/commits/b2a443e3ce8ac05feef5cb8793c25d679deaa197)

projects/storefrontlib/shared/components/chat-messaging/messaging/messaging.component.ts
khristian23: Is the Ticket State already implemented?

Parthlakhani: no, not yet. We can add this note to this ticket: [CXEC-14760](https://jira.tools.sap/browse/CXEC-14760): Implement core logic to add new messages

developpeurweb (changes requested):
1. A keyboard pattern is missing. I suggest using _Slack_ or _Teams_ chat keyboard pattern as guidance. E.g. Tab goes in/out chatbot area, up/down arrows move between messages.
2. The “Upload File” button is also unreachable with keyboard.
3. Once keyboard navigation works, please test with a screen reader to make sure it sounds right, or reach out to me again for JAWS/VO testing.
4. When the character limit is reached, either a sr-only alert (`cx-visually-hidden`) is needed, or it could also be a visual alert that is picked up by the screen reader as well.
5. We’re trying to avoid using Bootstrap layout, eg. `col-xs-6 col-md-3 col-lg-2`, if you can fix this with flexbox, that would be great; otherwise pls reach out to Mikel Morgan for solution, if he’s not available then leave as is, this is low priority.
6. Placeholders are not labels, so we need to add a label associated with the message field (if this is as per mockup, pls ping designers and have them ping me, to elaborate. This causes an AMP bug). 
7. There are other AMP bugs, that are easy to fix. Let me know if you need AMP accounts to be setup.
Screen Shot 2022-08-29 at 4 52 51 PM


Screen Shot 2022-08-29 at 4 53 34 PM

developpeurweb (commented):
LGTM a11y-wise. Tested with JAWS/NVDA/VO and AMP. No issues found.

feature-libs/customer-ticketing/components/customer-ticketing-messages/customer-ticketing-messages.component.ts
```diff
@@ -0,0 +1,81 @@
...
+import { CustomerTicketingService } from '../customer-ticketing.service';
+
+@Component({
+  selector: 'cx-customer-ticketing-messages',
+  templateUrl: './customer-ticketing-messages.component.html',
+})
+export class CustomerTicketingMessagesComponent {
+  constructor(protected customerTicketingService: CustomerTicketingService) {}
+
+  ticketDetails$ = of({
```
adolfobee: We're getting rid of this in another pr?

feature-libs/customer-ticketing/components/customer-ticketing-messages/customer-ticketing-messages.component.ts
```diff
@@ -0,0 +1,81 @@
...
+      },
+      {
+        author: 'Admin',
+        createdAt: '2022-06-22T14:37:15+0000',
+        message: 'A message to consider',
+      },
+    ],
+  });
+
+  messageEvents$: Observable<Array<MessageEvent> | undefined> =
```
adolfobee, line 80: why are we using $ after the variable, are you using it for events?

feature-libs/customer-ticketing/components/customer-ticketing-messages/customer-ticketing-messages.component.ts
Parthlakhani: Yes, this PR is for UI only, and it takes data from here.

feature-libs/customer-ticketing/components/customer-ticketing-messages/customer-ticketing-messages.component.ts
Parthlakhani: I believe that's a common practice followed when we use Observables.

projects/assets/src/translations/en/common.ts
```diff
@@ -210,4 +210,14 @@ export const common = {
     showPassword: 'Show password',
     hidePassword: 'Hide password',
   },
+  chatMessaging: {
+    charactersLeft: 'characters left: {{count}}',
+    addNewMessage: 'Add New Message',
+    send: 'Send',
+    uploadFile: 'Upload File',
+    textLabel: '{{author}}. {{text}} at {{date}}',
```
adolfobee: maybe rename this to title label or informationLabel, to specify its function rather than type

projects/storefrontlib/shared/components/chat-messaging/avatar/avatar.component.ts
```diff
@@ -0,0 +1,25 @@
+import { Component, Input } from '@angular/core';
+import { ICON_TYPE } from '@spartacus/storefront';
+import { MessageEvent } from '../messaging';
+
+export const FROM_CUSTOMER = 'C';
```
adolfobee: we getting rid of this later?

Parthlakhani: nope, it's used for showing in Avatar when there's no author for message found.

projects/assets/src/translations/en/common.ts
Parthlakhani: sure.

projects/storefrontlib/shared/components/chat-messaging/messaging/messaging.component.html
```diff
@@ -0,0 +1,119 @@
...
+  <div class="container">
+    <div class="cx-avatar-line"></div>
+    <div
+      class="cx-messages"
+      id="cx-messages"
+      (keydown.arrowdown)="focusNextChild($event)"
+      (keydown.arrowup)="focusPreviousChild($event)"
+      [attr.aria-label]="'chatMessaging.messages' | cxTranslate"
+    >
+      <ng-container *ngFor="let message of messageEvents; let i = index">
```
adolfobee, line 11: if this is still an array of events, I think it's better to call it `messageEvent of messageEvents`

Parthlakhani: I kept the name `message` so that it can be more readable. what do you think?

projects/storefrontlib/shared/components/chat-messaging/messaging/messaging.model.ts
```diff
@@ -0,0 +1,26 @@
...
+  author?: string;
+  rightAlign?: boolean;
+  createdAt?: string;
+  text?: string;
+  attachments?: Array<Attachment>;
+}
+
+export interface Attachment {
+  filename: string;
+  URL: string;
```
adolfobee: any reason why this is not lowercase like other attributes?

projects/storefrontlib/shared/components/form/file-upload/file-upload.component.html
```diff
@@ -16,9 +16,14 @@
   <ng-content></ng-content>
 </button>
 <!-- TODO(Brian-Parth): CXEC-17196 -->
-<div *ngIf="customButton" (click)="fileInput.click()">
+<button
+  class="link"
+  *ngIf="customButton"
+  (click)="fileInput.click()"
+  type="button"
```
adolfobee: I think this is redundant

adolfobee (approved):

projects/storefrontlib/shared/components/chat-messaging/avatar/avatar.component.ts
adolfobee: maybe in the future we will use initials like they do in other apps, this is fine for first release

projects/storefrontlib/shared/components/chat-messaging/messaging/messaging.component.html
adolfobee: yes you can keep it, we can print the event and see the text inside

projects/storefrontlib/shared/components/chat-messaging/messaging/messaging.model.ts
Parthlakhani: this needs to be replaced with `id`

projects/storefrontlib/shared/components/form/file-upload/file-upload.component.html
Parthlakhani: do you mean the button? or click?

adolfobee: let's remove the type button

projects/storefrontlib/shared/components/chat-messaging/avatar/avatar.component.ts
Parthlakhani: We are using initials, this is when there is no user name provided. since it is shared component, we decided to keep something like this.

adolfobee (approved):

projects/storefrontlib/shared/components/chat-messaging/avatar/avatar.component.ts
```diff
@@ -0,0 +1,25 @@
+import { Component, Input } from '@angular/core';
+import { ICON_TYPE } from '@spartacus/storefront';
```
bgambocjaviniar: need to fix this import into a relative import to the path as recall, when you are within the same lib, use relative path. But when using things from outside the 'library', then you can use the aliased package.

For example, since this is `storefrontlib`, which is @spartacus/storefront, you should not do @import xxx from @spartacus/storefront. You should be using the relative path which is `import { ICON_TYPE } from '../../../../cms-components/misc/icon/icon.model';`

projects/storefrontlib/shared/components/chat-messaging/avatar/avatar.component.ts
```diff
@@ -0,0 +1,25 @@
...
+import { MessageEvent } from '../messaging';
```
bgambocjaviniar: let's complete the path explicitly, so either `import { MessageEvent } from '../messaging/messaging.model';` or `import { MessageEvent } from '../messaging/index';`

projects/storefrontlib/shared/components/chat-messaging/messaging/messaging.component.ts
```diff
@@ -0,0 +1,157 @@
...
+  AfterViewInit,
+  Component,
+  EventEmitter,
+  Input,
+  OnInit,
+  Output,
+} from '@angular/core';
+import { FormControl, FormGroup, Validators } from '@angular/forms';
+import { WindowRef } from '@spartacus/core';
+import { ICON_TYPE } from 'projects/storefrontlib/cms-components';
```
bgambocjaviniar: fix this absolute path for both ICON_TYPE and FilesFormValidators. 

Make sure it's relative and to not use @spartacus/storefront

bgambocjaviniar: import { ICON_TYPE } from '../../../../cms-components/misc/icon/icon.model';
import { FilesFormValidators } from '../../../services/file/files-form-validators';

projects/storefrontlib/shared/components/chat-messaging/avatar/avatar.component.spec.ts
```diff
@@ -0,0 +1,58 @@
+import { ComponentFixture, TestBed } from '@angular/core/testing';
+import { By } from '@angular/platform-browser';
+import { I18nTestingModule } from '@spartacus/core';
+import { IconModule } from 'projects/storefrontlib/cms-components';
```
bgambocjaviniar: Testing (spec files) are more lenient to the aliased path within the same lib.

Use relative or the aliased path in spec files instead of absolute path. We have checks in the pipeline for kind of things, where absolute paths are used which makes it fail.

projects/storefrontlib/shared/components/chat-messaging/messaging/messaging.component.spec.ts
```diff
@@ -0,0 +1,67 @@
+import { ComponentFixture, TestBed } from '@angular/core/testing';
+import { ReactiveFormsModule } from '@angular/forms';
+import { By } from '@angular/platform-browser';
+import { I18nTestingModule } from '@spartacus/core';
+import { IconModule } from 'projects/storefrontlib/cms-components';
```
bgambocjaviniar, line 5: Testing (spec files) are more lenient to the aliased path within the same lib.

Use relative or the aliased path in spec files instead of absolute path. We have checks in the pipeline for these kind of things, where absolute paths are used which makes it fail.

projects/storefrontlib/shared/components/chat-messaging/messaging/messaging.module.ts
```diff
@@ -0,0 +1,24 @@
+import { NgModule } from '@angular/core';
+import { CommonModule } from '@angular/common';
+import { MessagingComponent } from './messaging.component';
+import { I18nModule } from '@spartacus/core';
+import { IconModule } from 'projects/storefrontlib/cms-components/misc/icon/icon.module';
```
bgambocjaviniar: Here too, use relative path

`import { IconModule } from '../../../../cms-components/misc/icon/icon.module';`

projects/storefrontlib/shared/components/chat-messaging/messaging/messaging.module.ts
```diff
@@ -0,0 +1,24 @@
...
+import { FormsModule, ReactiveFormsModule } from '@angular/forms';
+import { AvatarModule } from '../avatar';
```
bgambocjaviniar: **Going to stop throwing these kind of fixes, but make sure in the PR, the imports does not implicitly let index be used**. For `AvartarModule` and `FileUploadModule, FormErrorsModule`, you can do the following


`import { AvatarModule } from '../avatar/avatar.module';`

AND

`import { FileUploadModule, FormErrorsModule } from '../../form/index';`

projects/storefrontlib/shared/components/chat-messaging/avatar/avatar.component.ts
bgambocjaviniar: if somehow author, is not available, let's use an icon, such as `fas fa-user` - user

What do you think? So in the template, you'd have an  *ngIf="user"; else anonymous"

feature-libs/customer-ticketing/components/customer-ticketing-messages/customer-ticketing-messages.component.ts
bgambocjaviniar: Just to clarify, this 'mocked' data is pretty redundant since your taking the mocks from here instead `customerTicketingService`, which I believe that service will also be 'gone' once your working with the data with the actual 'facade'

feature-libs/customer-ticketing/components/customer-ticketing.service.ts
```diff
@@ -1,14 +1,16 @@
 import { Injectable } from '@angular/core';
```
bgambocjaviniar, line 1: let's delete this file actually since this will not exist anyhow in the future for any upcoming PRs.

Just move the mocked data here https://github.com/SAP/spartacus/pull/16085/files#r968614886

Don't worry about injecting a facade as this is only for the UI anyhow

projects/storefrontlib/shared/components/chat-messaging/messaging/messaging.component.html
```diff
@@ -0,0 +1,119 @@
+<ng-container *ngIf="messageEvents$ | async as messageEvents">
```
bgambocjaviniar, line 1: @morganm58 what do you think of the UI? I believe the accessibility was covered by @developpeurweb
Tbh it looks good to me. 

How to test and get to this page?

1. go to this branch
2. enable b2b by doing `yarn start:b2b`
3. login as mark rivers (mark.rivers@rustic-hw.com) or any other b2b users
4. visit http://localhost:4200/powertools-spa/en/USD/my-account/support-ticket/00000001 


--------
Not relevant to this comment above, but the data flow for this 'generic' component. I'll need to get approval for. Will pull another developer or bring in catchup for Wednesday, September 21st because the more I look at it, it seems to me like a specific component just for customer ticketing. 

One can say, we will re-use this styles across different other components/features, but maybe not? Moreover, if ever the data structure that comes from the backend for 'other' messaging component is different then they'd have to serialize it to conform to our data, which is fine, but please hold on for a second.

projects/storefrontlib/shared/components/chat-messaging/messaging/messaging.component.ts
```diff
@@ -0,0 +1,157 @@
...
+        results[results.length - 1].focus();
+      } else {
+        results[focusedIndex - 1].focus();
+      }
+    }
+  }
+
+  private getResultElements(): HTMLElement[] {
+    return Array.from(
+      this.windowRef.document.querySelectorAll('.cx-message-card > div > div')
```
bgambocjaviniar: instead of going through the hierarchy, you can do this querySelectorAll for this `'[role="listitem"]'`. Also instead of list item

bgambocjaviniar: Moreover, if we're accessing the `document` object, we need to wrap it only for browsers as `document` object is not available in ssr

bgambocjaviniar: You can verify using the `isBrowser` from winRef

projects/storefrontlib/shared/components/chat-messaging/messaging/messaging.component.ts
```diff
@@ -0,0 +1,157 @@
...
+    protected windowRef: WindowRef,
+    protected filesFormValidators: FilesFormValidators
+  ) {}
+
+  ngOnInit(): void {
+    this.buildForm();
+  }
+
+  ngAfterViewInit(): void {
+    if (this.scrollToInput) {
```
bgambocjaviniar: let's remove this logic until it will be implemented. So let's remove ngAfterViewInit, so we can merge this as soon as  possible

projects/storefrontlib/shared/components/chat-messaging/avatar/avatar.component.spec.ts
```diff
@@ -0,0 +1,58 @@
...
+    fixture.detectChanges();
+  });
+
+  it('should create', () => {
+    expect(component).toBeTruthy();
+  });
+
+  it('should return initials', () => {
+    let result = component.getInitials('Mark Rivers');
+
```
bgambocjaviniar, line 37: If the icon for 'anonmyous' (ghost writer) - missing author,

then add the test case for that

bgambocjaviniar (changes requested):

feature-libs/customer-ticketing/components/customer-ticketing-messages/customer-ticketing-messages.component.spec.ts
```diff
@@ -0,0 +1,24 @@
...
+  });
+
+  beforeEach(() => {
+    fixture = TestBed.createComponent(CustomerTicketingMessagesComponent);
+    component = fixture.componentInstance;
+    fixture.detectChanges();
+  });
+
+  it('should create', () => {
+    expect(component).toBeTruthy();
```
bgambocjaviniar, line 22: should add test cases here in the future when you get data + sending data

projects/storefrontstyles/scss/components/misc/chat-messaging/_avatar.scss
morganm58: The styling is missing the css variables for font and colors. We use variables because spartacus has the ability to switch themes. If you got0 your console in the browser and add the class `santorini` to the `cx-storefront` tag you'll see the theme change. I'll add the comments in the scss with the different vars.

projects/storefrontstyles/scss/components/misc/chat-messaging/_avatar.scss
```diff
@@ -0,0 +1,18 @@
...
+  .cx-avatar {
+    height: 4.375rem;
+    min-width: 4.375rem;
+    background-color: #fff1f1;
+    border-radius: 50%;
+    justify-content: center;
+    align-items: center;
+    display: flex;
+    border: 1px solid;
```
morganm58: this should be `bord: 1px solid var(--cx-color-light)`

projects/storefrontstyles/scss/components/misc/chat-messaging/_avatar.scss
```diff
@@ -0,0 +1,18 @@
...
+    border-color: #d3d6db;
```
morganm58: this. can be removed and consolidated. in. the line above

projects/storefrontstyles/scss/components/misc/chat-messaging/_avatar.scss
```diff
@@ -0,0 +1,18 @@
+%cx-avatar {
+  .cx-avatar {
+    height: 4.375rem;
+    min-width: 4.375rem;
+    background-color: #fff1f1;
```
morganm58: this should be `background-color:mix(theme-color('inverse'), theme-color('primary'), 90%);`

projects/storefrontstyles/scss/components/misc/chat-messaging/_avatar.scss
```diff
@@ -0,0 +1,18 @@
...
+    justify-content: center;
+    align-items: center;
+    display: flex;
+    border: 1px solid;
+    border-color: #d3d6db;
+    font-size: 2rem;
+  }
+
+  .right-align {
+    background-color: #f1f1f1;
```
morganm58: this should be background-color: var(--cx-color-background)

projects/storefrontstyles/scss/components/misc/chat-messaging/_avatar.scss
```diff
@@ -0,0 +1,18 @@
...
+    height: 4.375rem;
+    min-width: 4.375rem;
+    background-color: #fff1f1;
+    border-radius: 50%;
+    justify-content: center;
+    align-items: center;
+    display: flex;
+    border: 1px solid;
+    border-color: #d3d6db;
+    font-size: 2rem;
```
morganm58: font-size: var(--cx-font-size, 2rem);

projects/storefrontstyles/scss/components/misc/chat-messaging/_messaging.scss
```diff
@@ -0,0 +1,139 @@
...
+    display: flex;
+    flex-direction: row;
+    padding: 15px 0;
+    min-height: 400px;
+    padding-bottom: 2rem;
+
+    .cx-avatar-line {
+      border: 0px solid;
+      margin: 0 0 0 2.1875rem;
+      color: #d3d6db;
```
morganm58: `color: var(--cx-color-light)`

projects/storefrontstyles/scss/components/misc/chat-messaging/_messaging.scss
```diff
@@ -0,0 +1,139 @@
...
+        @include media-breakpoint-up(md) {
+          padding: 0 0 0 4.875rem;
+        }
+      }
+
+      .cx-message-left-align-text {
+        padding: 20px 28px;
+        border: 1px solid;
+        background-color: #fff1f1;
+        border-color: #d3d6db;
```
morganm58: `color: var(--cx-color-light)`

projects/storefrontstyles/scss/components/misc/chat-messaging/_messaging.scss
```diff
@@ -0,0 +1,139 @@
...
+        border-radius: 0px 16px 16px 16px;
+      }
+
+      .cx-message-right-align-text {
+        padding: 20px 28px;
+        border: 1px solid;
```
morganm58: `border: 1px solid var(--cx-color-light)`

projects/storefrontstyles/scss/components/misc/chat-messaging/_messaging.scss
```diff
@@ -0,0 +1,139 @@
...
+        background-color: #f1f1f1;
+        border-color: #d3d6db;
```
morganm58: this can be removed

projects/storefrontstyles/scss/components/misc/chat-messaging/_messaging.scss
```diff
@@ -0,0 +1,139 @@
...
+        border: 1px solid;
+        background-color: #fff1f1;
+        border-color: #d3d6db;
+        border-radius: 0px 16px 16px 16px;
+      }
+
+      .cx-message-right-align-text {
+        padding: 20px 28px;
+        border: 1px solid;
+        background-color: #f1f1f1;
```
morganm58: `background: color: var(--cx-color-background)`

projects/storefrontlib/shared/components/form/file-upload/file-upload.component.html
```diff
@@ -16,9 +16,9 @@
   <ng-content></ng-content>
 </button>
 <!-- TODO(Brian-Parth): CXEC-17196 -->
-<div *ngIf="customButton" (click)="fileInput.click()">
+<button class="link" *ngIf="customButton" (click)="fileInput.click()">
```
morganm58: action links should also have this class `class= "link cx-action-link"`

projects/storefrontstyles/scss/components/misc/chat-messaging/_messaging.scss
```diff
@@ -0,0 +1,139 @@
...
+        gap: 0.875rem;
+
+        div {
+          width: 100%;
+        }
+      }
+
+      .cx-message-display {
+        @include media-breakpoint-up(md) {
+          padding: 0 0 0 4.875rem;
```
morganm58: This should be `padding-top: 0; padding-inline-end: 0; padding-bottom: 0; padding-inline-start: 4.875rem`
any padding or margin that has left or right values should be `padding-inline-start:`, `padding-inline-end` or `margin-inline-end` or `margin-inline-start`. We do this to make our padding and margin work with directionality for languages that read right to left or left to right.

projects/storefrontstyles/scss/components/misc/chat-messaging/_messaging.scss
```diff
@@ -0,0 +1,139 @@
...
+      color: #d3d6db;
+      border-right-width: 1px;
+    }
+
+    .cx-messages {
+      display: flex;
+      flex-direction: column;
+      width: 100%;
+      gap: 1rem;
+      margin: 0 0 0 -2.1875rem;
```
morganm58: `margin-top: 0; margin-inline-end: 0; margin-bottom: 0; margin-inline-start: -2.1875rem`

projects/storefrontstyles/scss/components/misc/chat-messaging/_messaging.scss
```diff
@@ -0,0 +1,139 @@
...
+        padding: 20px 28px;
+        border: 1px solid;
+        background-color: #f1f1f1;
+        border-color: #d3d6db;
+        border-radius: 0px 16px 16px 16px;
+      }
+
+      .cx-attachment {
+        cx-icon {
+          color: #a1aabf;
```
morganm58: `color: var(--cx-color-secondary)`

projects/storefrontstyles/scss/components/misc/chat-messaging/_messaging.scss
```diff
@@ -0,0 +1,139 @@
...
+          width: auto;
+        }
+        .cx-message-footer-text {
+          text-decoration: underline;
+          cursor: pointer;
+          padding: 0rem 0.5rem;
+          white-space: nowrap;
+        }
+        cx-icon {
+          color: #a1aabf;
```
morganm58: `color: var(--cx-color-secondary)`

projects/storefrontstyles/scss/components/misc/chat-messaging/_messaging.scss
```diff
@@ -0,0 +1,139 @@
...
+  .container {
+    display: flex;
+    flex-direction: row;
+    padding: 15px 0;
+    min-height: 400px;
+    padding-bottom: 2rem;
+
+    .cx-avatar-line {
+      border: 0px solid;
+      margin: 0 0 0 2.1875rem;
```
morganm58: `margin-top: 0; margin-inline-end: 0; margin-bottom: 0; margin-inline-start: 2.1875rem`

projects/storefrontstyles/scss/components/misc/chat-messaging/_messaging.scss
```diff
@@ -0,0 +1,139 @@
...
+      }
+    }
+  }
+
+  .cx-message-footer {
+    display: flex;
+    gap: 1rem;
+
+    @include media-breakpoint-up(md) {
+      padding: 0 0 0 4.875rem;
```
morganm58: padding-top: 0; padding-inline-end: 0; padding-bottom: 0; padding-inline-start: 4.875rem

projects/storefrontstyles/scss/components/misc/chat-messaging/_messaging.scss
```diff
@@ -0,0 +1,139 @@
...
+        }
+        cx-icon {
+          color: #a1aabf;
+          cursor: pointer;
+          font-size: 1.25rem;
+        }
+      }
+
+      .cx-footer-label {
+        font-weight: 600;
```
morganm58: `font-weight: var(--cx-font-weight-bold);`

projects/storefrontstyles/scss/components/misc/chat-messaging/_messaging.scss
```diff
@@ -0,0 +1,139 @@
...
+        .cx-message-footer-text {
+          text-decoration: underline;
+          cursor: pointer;
+          padding: 0rem 0.5rem;
+          white-space: nowrap;
+        }
+        cx-icon {
+          color: #a1aabf;
+          cursor: pointer;
+          font-size: 1.25rem;
```
morganm58: `font-size: var(--cx-font-size, 1.25rem);`

projects/storefrontstyles/scss/components/misc/chat-messaging/_messaging.scss
```diff
@@ -0,0 +1,139 @@
...
+        font-weight: 600;
+        .cx-message-input {
+          display: flex;
+          gap: 1rem;
+        }
+      }
+
+      .cx-message-footer-info {
+        display: flex;
+        padding: 0 7.5rem 0 0;
```
morganm58: `padding-top: 0; padding-inline-end: 7.5rem; padding-bottom: 0; padding-inline-start: 0`

projects/storefrontstyles/scss/components/misc/chat-messaging/_messaging.scss
```diff
@@ -0,0 +1,139 @@
...
+        justify-content: space-between;
+        .cx-message-footer-info-text {
+          width: 100%;
+        }
+        p {
+          font-size: 0.75rem;
```
morganm58: `font-size: var(--cx-font-size, 0.75rem);`

projects/storefrontstyles/scss/components/misc/chat-messaging/_messaging.scss
```diff
@@ -0,0 +1,139 @@
...
+      .cx-message-display {
+        @include media-breakpoint-up(md) {
+          padding: 0 0 0 4.875rem;
+        }
+      }
+
+      .cx-message-left-align-text {
+        padding: 20px 28px;
+        border: 1px solid;
+        background-color: #fff1f1;
```
morganm58: `background-color:mix(theme-color('inverse'), theme-color('primary'), 90%);`

projects/storefrontlib/shared/components/chat-messaging/avatar/avatar.component.ts
Parthlakhani: Understood

projects/storefrontlib/shared/components/chat-messaging/messaging/messaging.module.ts
Parthlakhani: got it

Parthlakhani:
> if somehow author, is not available, let's use an icon, such as fas fa-user - user
> 
> What do you think? So in the template, you'd have an *ngIf="user"; else anonymous" https://github.com/SAP/spartacus/pull/16085#discussion_r975748589

I hesitated before adding a icon and used this before, but think that's better, added the changes

Parthlakhani:
> Just to clarify, this 'mocked' data is pretty redundant since your taking the mocks from here instead customerTicketingService, which I believe that service will also be 'gone' once your working with the data with the actual 'facade'  https://github.com/SAP/spartacus/pull/16085#discussion_r975764523

Yes, it's not being used anyway, removing it. this logic would be properly fixed in the next pr.

feature-libs/customer-ticketing/components/customer-ticketing.service.ts
Parthlakhani: It will require changes in other components and unit tests as well, I already removed this in other PR. so I think we can keep this one for now. I am moving all logic related to messaging in the component to save future conflicts.

projects/storefrontlib/shared/components/chat-messaging/messaging/messaging.component.ts
Parthlakhani: got it

projects/storefrontlib/shared/components/chat-messaging/messaging/messaging.component.ts
Parthlakhani: okay!

projects/storefrontlib/shared/components/chat-messaging/avatar/avatar.component.spec.ts
Parthlakhani: sure

projects/storefrontstyles/scss/components/misc/chat-messaging/_avatar.scss
Parthlakhani: got it

projects/storefrontlib/shared/components/form/file-upload/file-upload.component.html
Parthlakhani: Got it

projects/storefrontstyles/scss/components/misc/chat-messaging/_messaging.scss
Parthlakhani: Ohh, I see.
`padding-inline-start: 4.875rem` is sufficient in that case.
I added it this was because `padding-left` was failing pipeline.

projects/storefrontstyles/scss/components/misc/chat-messaging/_messaging.scss
Parthlakhani: adding `font-weight: var(--cx-font-weight-semi);` as it matches more with the mockup

morganm58 (approved):

khristian23 (approved):

PR #16095, opened 2022-08-18
CXSPA-574: Customer ID Http Header for ASM emulation

justinlee01:
Closes https://jira.tools.sap/browse/CXSPA-574 , https://jira.tools.sap/browse/CXSPA-149 , https://jira.tools.sap/browse/CXSPA-151

- Creating UserIdInterceptor, used to append a "sap-commerce-cloud-user-id" header to certain HTTP requests that provide a HttpContext. This is so that there is no ambiguity as to who (the agent or the emulated user) is fetching certain resources.
- Covering GET /costcenters and GET /product/search
- Changing some E2E utils to be more stable (ex. use in-house `interceptGet` rather than cy.get with an incorrect BASE_SITE)
- Changing some E2E utils to support the B2B case


feature-libs/asm/root/interceptors/user-id.interceptor.ts
```diff
@@ -0,0 +1,76 @@
...
+  ): Observable<HttpEvent<unknown>> {
+    const asmContext = httpRequest.context.get(OCC_ASM_TOKEN);
+
+    let userIdObservable: Observable<string | undefined>;
+
+    if (typeof asmContext.sendUserIdAsHeader === 'string') {
+      userIdObservable = of(asmContext.sendUserIdAsHeader);
+    } else if (asmContext.sendUserIdAsHeader) {
+      userIdObservable = this.userIdService
+        .getUserId()
```
plabadie: consider calling `takeUserId()` instead of `getUserId()` if we just need the current userId and move on.

projects/core/src/occ/services/asm-context.service.ts
```diff
@@ -0,0 +1,26 @@
...
+/**
+ * Used to provide an ASM context for certain endpoints.
+ * For example, if an endpoint needs to pass in the emulated user's ID to resolve access issues,
+ * you should create a context.
+ * @see feature-libs/asm/root/interceptors/user-id.interceptor.ts
+ */
+@Injectable({
+  providedIn: 'root',
+})
+export class AsmContextService {
```
plabadie: For the sake of the discussion, would there be a way to make this context mechanism not ASM specific?  Ideally we'd avoid bringing back ASM logic back in the core lib and spreading ASM related dependencies in classes.

feature-libs/asm/root/interceptors/user-id.interceptor.ts
```diff
@@ -0,0 +1,76 @@
...
+} from '@spartacus/core';
+import { Observable, of } from 'rxjs';
+import { concatMap, map } from 'rxjs/operators';
+
+/**
+ * Looks for a specific key in the HttpRequest's context (OCC_ASM_TOKEN) to decide when to
+ * configure a request with 'sap-commerce-cloud-user-id' header.
+ */
+@Injectable({ providedIn: 'root' })
+export class UserIdInterceptor implements HttpInterceptor {
```
plabadie: Could this logic become non ASM specific and make it's way in the core lib?  Let's discuss.

feature-libs/asm/root/interceptors/user-id.interceptor.ts
```diff
@@ -0,0 +1,76 @@
...
+          });
+
+          return next.handle(request);
+        } else {
+          return next.handle(httpRequest);
+        }
+      })
+    );
+  }
+}
```
plabadie: Overall we might investigate to see if in some cases, it's best to bail out of the interceptor logic.  For example, perhaps it doesn't make sense to pass `anonymous` or `current` as a customer id in the header.  Figuring out that rule might require a bit of investigation.

feature-libs/asm/root/interceptors/user-id.interceptor.ts
justinlee01: ! I didn't realize the difference. Thank you.

projects/core/src/occ/services/asm-context.service.ts
justinlee01: Hmm. I can rename the method `createASMContext` (same implementation) and thus rename the service to `HttpContextService`.

feature-libs/asm/root/interceptors/user-id.interceptor.ts
justinlee01: Line 55 serves that purpose -
```
this.uniqueUserIdConstants.has(userId) ? undefined : userId
```
This filters out any hardcoded userId (that is not a real UUID) as that will break the request, as I've found out.

feature-libs/asm/root/interceptors/user-id.interceptor.ts
justinlee01: That's an interesting discussion. It was moved to the ASM module based on this premise:
1. 'sap-commerce-cloud-user-id' only needs to be sent when the user needs to be explicit i.e. in customer emulation case.
2. If it's only for customer emulation, then we can bolster the code's performance, if ever so slightly, by removing this interceptor from the chain, and by needing to load the code in the first place.

Is there an incentive to move it to core?

projects/core/src/occ/services/asm-context.service.ts
armanozak: Or completely remove it maybe? After all, it doesn’t do much and `HttpContext` is chainable. It’s the same as doing `new HttpContext().set(OCC_ASM_TOKEN, contextObj)` and the `OCC_ASM_TOKEN` is already exposed.

feature-libs/asm/root/interceptors/user-id.interceptor.ts
plabadie: In OCC/backend, is this header used exclusively for ASM?

projects/core/src/occ/utils/index.ts
```diff
@@ -4,6 +4,17 @@
...
+export { OCC_USER_IDS };
+
 export * from './interceptor-util';
 export * from './occ-constants';
 export * from './occ-url-util';
+export * from './occ-asm-token';
+
+export const OCC_USER_ID_CONSTANTS = new InjectionToken<{
+  [identifier: string]: string;
+}>('List of OCC constants that pass for user IDs.');
```
plabadie, line 20: Couldn't we let other classes import the constants instead of requiring dependency injection for them to use the constants?

projects/core/src/occ/services/asm-context.service.ts
plabadie: I agree that we should rather remove it.

projects/core/src/occ/base-occ.module.ts
```diff
@@ -27,6 +29,10 @@ export class BaseOccModule {
           useExisting: WithCredentialsInterceptor,
           multi: true,
         },
+        {
+          provide: OCC_USER_ID_CONSTANTS,
+          useValue: OCC_USER_IDS,
+        },
```
plabadie: Another option would be to provide the token as we instantiate it like so:
```
export const OCC_USER_ID_CONSTANTS = new InjectionToken('OCC_USER_ID_CONSTANTS', {
  providedIn: 'root',
  factory() {
    return OCC_USER_IDS;
  }
});
```

Got this idea from @armanozak and I thought it might be more concise to go with this.  What do you think?

justinlee01: Sure, I'll implement that. I didn't even know it accepted a second argument.

feature-libs/asm/root/interceptors/user-id.interceptor.ts
justinlee01: The header essentially overrides any session context that may have been used. For example, passing in the emulated customer ID will always override the user agent's authorization with the emulated user's. It's a very straightforward, if blunt, tool.

I can only imagine this would be used for ASM _for now_. I can't imagine any other scenario, at least given the current use of SPA.

projects/storefrontapp-e2e-cypress/cypress/helpers/asm.ts
```diff
@@ -37,8 +37,12 @@ export function listenForCustomerSearchRequest(): string {
...
 }
 
-export function listenForUserDetailsRequest(): string {
-  return interceptGet('userDetails', '/users/*');
+export function listenForUserDetailsRequest(b2b = false): string {
+  if (!b2b) {
+    return interceptGet('userDetails', '/users/*');
+  } else {
+    return interceptGet('userDetails', '/orgUsers/*');
+  }
```
plabadie, line 45: Suggestion to improve readability, do :
```
if (b2b) {
     ...
} else {
     ...
}
```

projects/core/src/occ/utils/occ-asm-token.ts
```diff
@@ -0,0 +1,19 @@
...
+  /**
+   * Indicates to the interceptor that the emulated customer ID should be sent to the OCC
+   * to avoid ambiguity.
+   * 'string' if the request has the user ID at hand.
+   * 'boolean' (or, 'true') if the request does not have the user ID but still needs to send the emulated ID.
+   */
+  sendUserIdAsHeader?: string | boolean;
+}
+
+export const OCC_ASM_TOKEN = new HttpContextToken<OccAsmContext>(() => ({}));
```
plabadie: I think it would help readability to call the token `OCC_ASM_CONTEXT_TOKEN` or even `OCC_ASM_HTTP_CONTEXT_TOKEN`.

Or even something more generic than ASM as mentioned above.

projects/core/src/occ/adapters/user/occ-user-cost-centers.adapter.ts
```diff
@@ -25,8 +26,15 @@ export class OccUserCostCenterAdapter implements UserCostCenterAdapter {
   ) {}
 
   loadActiveList(userId: string): Observable<EntitiesModel<CostCenter>> {
+    // @see feature-libs/asm/root/interceptors/user-id.interceptor.ts
+    const context = new HttpContext().set(OCC_ASM_TOKEN, {
+      sendUserIdAsHeader: true,
+    });
+
```
plabadie, line 32: Is there unit test coverage we could add for this?  (depending on how complex that would be, I don't know what it requires)

projects/core/src/occ/adapters/product/occ-product-search.adapter.ts
```diff
@@ -38,8 +39,13 @@ export class OccProductSearchAdapter implements ProductSearchAdapter {
     query: string,
     searchConfig: SearchConfig = DEFAULT_SEARCH_CONFIG
   ): Observable<ProductSearchPage> {
+    // @see feature-libs/asm/root/interceptors/user-id.interceptor.ts
+    const context = new HttpContext().set(OCC_ASM_TOKEN, {
+      sendUserIdAsHeader: true,
+    });
```
plabadie, line 44: Is there unit test coverage we could add for this?  (depending on how complex that would be, I don't know what it requires)

projects/core/src/occ/utils/occ-asm-token.ts
```diff
@@ -0,0 +1,19 @@
+/*
+ * SPDX-FileCopyrightText: 2022 SAP Spartacus team <spartacus-team@sap.com>
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { HttpContextToken } from '@angular/common/http';
+
+export interface OccAsmContext {
```
plabadie: Perhaps the "core lib" side of the logic could be called something else than "Asm" specific names and make it more generic.  It's really just about sending a flag or the user id to ask for the user id to be added as a header.

I don't have specific names in mind, but it could be something like `UserIdHttpContext`.

Also, while we are discussing naming, the file could be named `occ-userid-http-context.ts` or something like that as long as it it consistent with the naming inside the file.

feature-libs/asm/root/interceptors/user-id.interceptor.ts
```diff
@@ -0,0 +1,76 @@
...
+import { Observable, of } from 'rxjs';
+import { concatMap, map } from 'rxjs/operators';
+
+/**
+ * Looks for a specific key in the HttpRequest's context (OCC_ASM_TOKEN) to decide when to
+ * configure a request with 'sap-commerce-cloud-user-id' header.
+ */
+@Injectable({ providedIn: 'root' })
+export class UserIdInterceptor implements HttpInterceptor {
+  private readonly userIdHeader = 'sap-commerce-cloud-user-id';
```
plabadie: Can we change the private members to `protected`?

`protected` is the default visibility we use over private.  This helps for extensibility (customers can extend the class and see/use the variables.)

feature-libs/asm/root/interceptors/user-id.interceptor.ts
```diff
@@ -0,0 +1,76 @@
...
+
+  private readonly uniqueUserIdConstants: Set<string>;
+
+  constructor(
+    protected userIdService: UserIdService,
+    @Inject(OCC_USER_ID_CONSTANTS)
+    userIdConstants: { [identifier: string]: string }
```
plabadie: Just for consistency, could the userIdConstants property be protected as well?

feature-libs/asm/root/interceptors/user-id.interceptor.ts
plabadie: I'm ok to leave the interceptor in asm.  I agree that for the foreseeable future, the purpose of the interceptor is to enable customer emulation for endpoints that don't have other ways to pass the user id parameter.

I think however, that the logic of assigning a context and the required objects in "core" should be named in a more generic, non asm way.  We never know what future use case we might have for it or how customers might use it.  Plus this keeps the core lib free of apparent asm specific logic.  I left a comment in the file where the token and interface are declared for the context.

plabadie: I think that precise naming improves the code readability, may I suggest we give the interceptor a name that communicates its purpose.  Perhaps `UserIdHeaderInterceptor` or `UserIdHttpHeaderInterceptor`

projects/core/src/occ/utils/occ-asm-token.ts
justinlee01: The original idea was that this object could be extended for other flags needed for customer emulation. However, there is currently nothing else to send.
I am inclining toward your suggestion. Any thoughts @armanozak ?

projects/core/src/occ/adapters/user/occ-user-cost-centers.adapter.ts
```diff
@@ -25,8 +26,15 @@ export class OccUserCostCenterAdapter implements UserCostCenterAdapter {
   ) {}
 
   loadActiveList(userId: string): Observable<EntitiesModel<CostCenter>> {
+    // @see feature-libs/asm/root/interceptors/user-id.interceptor.ts
```
plabadie: These comments are nice, but they come with a maintenance overhead.
Perhaps looking up the token references in the ide is enough for devs to understand the usage?

If we decide to keep them, lets update the path.

justinlee01:
@armanozak @plabadie As the feature requires a corresponding change to Commerce Cloud, I have hidden it behind a feature flag, implemented in "UserIdHttpHeaderInterceptor". I have also skipped the written e2e tests for the same reason. Let me know your thoughts.

projects/core/src/occ/utils/occ-asm-token.ts
armanozak: How about calling the interface just `OccHttpContext` and the token `OCC_HTTP_CONTEXT`? Maybe later we can even consider making the context interface augmentable.

projects/storefrontlib/feature-toggles.ts
```diff
@@ -26,5 +26,16 @@ export interface FeatureToggles {
...
+    /**
+     * To fix certain features when a customer agent is emulating a user, we must send a header called
+     * "sap-commerce-cloud-user-id" to the backend fetching certain data as the user, rather than the
+     * agent. However, the header will break instances of Commerce Cloud who do not allow requests with
+     * this header. (For example, the configuration "corsfilter.commercewebservices.allowedHeaders" will
+     * need to allow it.)
+     *
+     * Enabling this feature will send the emulated user's IDs on requests that need it as context.
+     */
+    enableCommerceCloudUserIdHeader?: boolean;
```
plabadie: Good catch for the breaking change due to core filter settings.

I get the main reason for the feature flag is to prevent issues to customers when they upgrade to 5.1.  They would get errors form the requests that send the new header.
However it doesn't seem that this is a feature in of itself that customers want to toggle.

Do you think this behaviour could be hidden behind a 5.1 feature level flag instead?  Whenever a feature level flag is possible, we prefer that vs accumulating feature toggles.  (we have very few feature toggles.). Let's discuss.

In 6.0, it should become enabled by default.

projects/storefrontapp-e2e-cypress/cypress/integration/regression/asm/asm.e2e-spec.ts
```diff
@@ -72,4 +84,41 @@ context('Assisted Service Module', () => {
...
+    before(() => {
+      Cypress.env('BASE_SITE', APPAREL_BASESITE);
+    });
+
+    after(() => {
+      Cypress.env('BASE_SITE', ELECTRONICS_BASESITE);
+    });
+
+    // This test only works if "sap-commerce-cloud-user-id" is added to the allowed headers of "corsfilter.commercewebservices.allowedHeaders" on the Commerce Cloud side
+    it.skip("should fetch products in a category based on the emulated user's authentication", () => {
```
plabadie, line 157: We could update the CI backends to accept this header and enable the tests.
If we decide to go for a feature level flag, we might not have a choice but to update the CI backends before we merge this PR.  Let's discuss.

feature-libs/asm/root/interceptors/user-id-http-header.interceptor.ts
```diff
@@ -0,0 +1,83 @@
...
+
+  intercept(
+    httpRequest: HttpRequest<unknown>,
+    next: HttpHandler
+  ): Observable<HttpEvent<unknown>> {
+    if (!isFeatureEnabled(this.config, 'enableCommerceCloudUserIdHeader')) {
+      return next.handle(httpRequest);
+    }
+
+    const asmContext = httpRequest.context.get(OCC_HTTP_TOKEN);
```
plabadie, line 53: perhaps rename `asmContext` to something like `occHttpContext`

projects/core/src/occ/utils/occ-http-token.ts
```diff
@@ -0,0 +1,20 @@
...
+
+import { HttpContextToken } from '@angular/common/http';
+
+export interface OccHttpContext {
+  /**
+   * Indicates to the interceptor that the emulated customer ID should be sent to the OCC
+   * to avoid ambiguity.
+   * 'string' if the request has the user ID at hand.
+   * 'boolean' (or, 'true') if the request does not have the user ID but still needs to send the emulated ID.
+   * @see feature-libs/asm/root/interceptors/user-id-http-header.interceptor.ts
```
plabadie, line 15: I think this comment is also generated and published online, in which case the file path won't mean much.  Perhaps we use the class name and import path to refer to other classes.  Not sure what we do elsewhere for these cases.

feature-libs/asm/root/interceptors/user-id-http-header.interceptor.ts
```diff
@@ -0,0 +1,83 @@
...
+        );
+    } else {
+      return next.handle(httpRequest);
+    }
+
+    return userIdObservable.pipe(
+      concatMap((userId) => {
+        if (userId) {
+          const request = httpRequest.clone({
+            headers: httpRequest.headers.set(this.userIdHeader, userId),
```
plabadie, line 75: There should be documentation updates that go with this PR.
We need to update the installation/configuration guides for the backend to mention the new allowed header that needs to be added in the config.

projects/storefrontlib/feature-toggles.ts
justinlee01: Aren't there cases where the customer upgrades Spartacus but does not upgrade Commerce Cloud?
This type of toggle seems more pragmatic.

feature-libs/asm/root/interceptors/user-id-http-header.interceptor.ts
justinlee01: Tbh I don't see a difference. In this case this context is still acting as an ASM context.

projects/core/src/occ/utils/occ-http-token.ts
justinlee01: I can remove this line, but I'd like to see where this type of documentation gets published. This is an unusual obstacle for writing notes.
If the solution is to remove the JSDoc tag, that works for me.

justinlee01:
As a note, I have created a ticket for updating the backend: https://jira.tools.sap/browse/CXSPA-1355
I myself am not sure how this is done - through updating `spartacussampledata` or other.

feature-libs/asm/root/interceptors/user-id-http-header.interceptor.ts
justinlee01: Draft here: https://github.com/SAP/spartacus-docs/pull/1492

projects/core/src/occ/utils/occ-http-token.ts
plabadie: It's [this doc](https://sap.github.io/spartacus/injectables/AsmService.html), where each element of the Public API is documented.

However, I just saw that _all_ the api elements in this doc are referenced from their source code file path, so forget about my objections.

projects/storefrontlib/feature-toggles.ts
plabadie: I'm not sure I understand.  Let's discuss.

feature-libs/asm/root/interceptors/user-id-http-header.interceptor.spec.ts
```diff
@@ -0,0 +1,226 @@
...
+import { provideConfig } from '@spartacus/core';
+import { Observable, of } from 'rxjs';
+import { take } from 'rxjs/operators';
+
+import { OCC_USER_ID_CURRENT } from '../../../../projects/core/src/occ/utils/occ-constants';
+import { UserIdService } from '../../../../projects/core/src/auth/user-auth/facade';
+import {
+  OCC_HTTP_TOKEN,
+  OCC_USER_ID_CONSTANTS,
+} from '../../../../projects/core/src/occ/utils';
```
ercultimate: Are these imports appropriate?

justinlee01: Ah, no. I suppose these are ignored for compilation because they're in unit tests, but they ought to be fixed. Ty.

feature-libs/asm/core/config/asm-config.ts
```diff
@@ -11,6 +11,9 @@ import { Config, OccConfig } from '@spartacus/core';
   providedIn: 'root',
   useExisting: Config,
 })
+/**
+ * @deprecated. Please use feature-libs/asm/root/asm-config.ts
```
plabadie: The deprecated message should use the import path instead of the file path.  The import path is what the customers need to use the newer/non-deprecated class.

feature-libs/asm/core/config/default-asm-config.ts
```diff
@@ -6,6 +6,9 @@
 
 import { AsmConfig } from './asm-config';
 
+/**
+ * @deprecated. Please use feature-libs/asm/root/default-asm-config.ts
+ */
```
plabadie: Was defaultAsmConfig part go the public API?

plabadie: If not, perhaps we can move this one directly without deprecating.

justinlee01: Hmm. `AsmConfig` will live in `core`, but `defaultAsmConfig` will live in `root`. Let me see if the dependency framework is ok with that (as in, we `provideDefaultConfig` in root).

plabadie (dismissed):

plabadie (approved):

kimhw0630 (approved):

PR #16101, opened 2022-08-22
refactor: move unit-level orders feature to organization library

MarcinGanczorz:
https://jira.tools.sap/browse/CXEC-17288


feature-libs/organization/unit-order/styles/_order-history.scss
```diff
@@ -83,8 +83,8 @@
   }
```
I539321: Naming could be improved, even though it's not easy to do. Maybe the only name I'd like to see changed is the name of this file, back to how it was, because currently we have two files named `_order-history.scss`.

I539321 (changes requested):

feature-libs/organization/unit-order/styles/_order-history.scss
MarcinGanczorz: thanks i missed that

I539321 (approved):

I565452 (approved):

I539321 (approved):

PR #16103, opened 2022-08-22
GH-15783: add testrunner for WTF E2E cypress tests

schuppell:


projects/storefrontapp-e2e-cypress/package.json
```diff
@@ -18,6 +18,7 @@
...
     "cy:run:ci:ccv2": "cypress run --config-file cypress.ci.json --config baseUrl=$ENDPOINT_URL_PUBLIC_SPARTACUS --reporter junit --reporter-options mochaFile=results/spartacus-test-results-[hash].xml --env API_URL=$ENDPOINT_URL_PUBLIC_API --spec \"cypress/integration/cx_ccv2/regression/b2c/**/*.e2e-spec.ts\"",
     "cy:run:ci:ccv2-b2b": "cypress run --config-file cypress.ci.json --config baseUrl=$ENDPOINT_URL_PUBLIC_SPARTACUS --reporter junit --reporter-options mochaFile=results/spartacus-test-results-[hash].xml --env API_URL=$ENDPOINT_URL_PUBLIC_API,BASE_SITE=powertools-spa,OCC_PREFIX_USER_ENDPOINT=orgUsers --spec \"cypress/integration/cx_ccv2/regression/b2b/**/*.e2e-spec.ts\"",
+    "cy:run:ci:ccv2-cpq": "cypress run --config-file cypress.ci.json --config baseUrl=$ENDPOINT_URL_PUBLIC_SPARTACUS --reporter junit --reporter-options mochaFile=results/spartacus-test-results-[hash].xml --env API_URL=$ENDPOINT_URL_PUBLIC_API,BASE_SITE=powertools-spa,OCC_PREFIX_USER_ENDPOINT=orgUsers --spec \"cypress/integration/**/cpq-configuration.ccv2-e2e-spec.ts\"",
```
bgambocjaviniar, line 21: where will this be tested on ccv2? My assumption from the thread that you dm'd me is for p3 right?

If it's for p3, is the base site really powertools-spa or is `powertools-epdvisualization-spa` as done in the package script `cy:run:ci:epd-visualization`

projects/storefrontapp-e2e-cypress/cypress/integration/vendor/cpq/cpq-configuration.ccv2-e2e-spec.ts
```diff
@@ -0,0 +1,589 @@
...
+      ovContent.forEach((line, idx) => {
+        configurationOverviewCpq.checkAttrDisplayed(line.name, line.value, idx);
+        configurationOverviewCpq.checkAttrPriceDisplayed(line.price, idx);
+        configurationOverviewCpq.checkAttrType(line.type, idx);
+      });
+    });
+  });
+
+  /*
+  describe('Configuration Process', () => {
```
bgambocjaviniar: What is the reason for commenting it out? 

From my understanding, I believe it was commented out since it did not work ootb on the p3 env on ccv2? Therefore, I'm assuming this new file was created to just comment this `describe` out, which is the only difference

Maybe it's best to fix the e2e test case instead?

---
For a general information on why there is a new ccv2 folder

- Giancarlo created them in favor for getting them quick and stable as some people were waiting for results for Spartacus testing on ccv2

run-spartacus-cpq-e2es.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+# install cypress standlone dependencies
+(cd projects/storefrontapp-e2e-cypress && yarn install)
```
bgambocjaviniar, line 4: I guess this can be improved where we pass in args to determine what package script to run.

You don't have to fix it and add that switch statement or conditional statements for the args, but I'm just mentioning that this won't scale well as more teams test in ccv2.

I believe the team will 'upgrade' this pattern when we test 5.0 in ccv2

bgambocjaviniar (changes requested):

projects/storefrontapp-e2e-cypress/package.json
schuppell: Yes we want to test on P3.
We test on powertools-spa

projects/storefrontapp-e2e-cypress/cypress/integration/vendor/cpq/cpq-configuration.ccv2-e2e-spec.ts
schuppell: I will delete the comment block in the test class.
The test is in the vendor folder, because it is dependent on a third party system. Therefor we do not want them to be executed when ccv2-b2c or ccv2-b2b tests are executed.

run-spartacus-cpq-e2es.sh
schuppell: I agree with this statement. Should I change it or do we leave it for now?

projects/storefrontapp-e2e-cypress/cypress/integration/vendor/cpq/cpq-configuration.ccv2-e2e-spec.ts
bgambocjaviniar: I understand you don't want to execute it for b2c/b2b test suite

my main question was more towards, why is it commented out? I believe the file is exactly the same as the other file in the same folder, but was just removed because it didn't work well with ccv2 p3. If it does not work well on ccv2 p3, maybe it's best to update that test case to adhere to what p3 has.

run-spartacus-cpq-e2es.sh
bgambocjaviniar: We can leave it for now, and someone in Spartacus team will update.

However, if you feel like updating it yourself, please feel free to do so! Help is always much appreciated!

projects/storefrontapp-e2e-cypress/cypress/integration/vendor/cpq/cpq-configuration.ccv2-e2e-spec.ts
schuppell: The problem is more of a user/test data type.
We use specific users in our own infrastructure tests, which are not available on P3.

bgambocjaviniar (approved):

PR #16110, opened 2022-08-26
chore: Copy code from public repository to private repository

bgambocjaviniar:
closes https://jira.tools.sap/browse/CXSPA-852

cron job that run at midnight every monday and thursday

depends on https://jira.tools.sap/browse/CXSPA-1365

-----

can see it work in action using this branch in this repo - develop-fake-test
which will be pushed here https://github.tools.sap/cx-commerce-storefronts/pwa-storefront-toolset/tree/develop-fake-test


.github/workflows/repo-sync.yml
```diff
@@ -0,0 +1,64 @@
...
+    name: Force sync a branch to the private repository with current git history
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.inputs.branch_to_sync }}
+          fetch-depth: 0
+      - name: Push a branch to the private repository
+        run: | 
+          git push -u https://${{ secrets.GHT_USER }}:${{ secrets.GHT_PRIVATE_REPO_TOKEN }}@github.tools.sap/cx-commerce-storefronts/${{ secrets.GHT_SPARTACUS_REPO}}.git ${{ github.event.inputs.branch_to_sync }} -f
```
RadhepS: {{ secrets.GHT_SPARTACUS_REPO}} just needs space at end to be consistent
{{ secrets.GHT_SPARTACUS_REPO }}

.github/workflows/repo-sync.yml
```diff
@@ -0,0 +1,64 @@
...
+  schedule:
+    - cron:  '0 0 * * 1,4'
+  workflow_dispatch:
+    inputs:
+      branch_to_sync:
+        description: Branch to sync to the private repository
+        default: develop
+        required: true
+
+name: Sync a branch to private PWA repo
```
RadhepS: headless storefront toolset repo (can also just write private repo)

.github/workflows/repo-sync.yml
```diff
@@ -0,0 +1,64 @@
+on:
+  schedule:
+    - cron:  '0 0 * * 1,4'
+  workflow_dispatch:
+    inputs:
+      branch_to_sync:
+        description: Branch to sync to the private repository
+        default: develop
+        required: true
```
RadhepS, line 9: required is true but when this runs automatically the input won't be passed. will it still work?
or does it work because the default value is set to develop? meaning that required:true doesn't really do anything?

bgambocjaviniar: explained as if there is not required, then the user can delete the input and run it with having the input being empty, whereas if this flag is included, it will prompt the user to write a value to the input

RadhepS (approved):

PR #16113, opened 2022-08-26
Create abstract dialog component for customer ticketing

Parthlakhani:


feature-libs/customer-ticketing/components/customer-ticketing-dialog/customer-ticketing-dialog.component.ts
```diff
@@ -0,0 +1,86 @@
+import { Directive, ElementRef, HostListener } from '@angular/core';
```
bgambocjaviniar: let's create it under ../customer-ticketing/components/shared/customer-ticketing-dialog

feature-libs/customer-ticketing/components/customer-ticketing-dialog/customer-ticketing-dialog.component.ts
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+  get allowedTypes(): string[] | undefined {
+    return this.customerTicketingConfig.customerTicketing
+      ?.attachmentRestrictions?.allowedTypes;
+  }
+
+  get getInputCharactersLimit(): number {
+    return (
+      this.customerTicketingConfig.customerTicketing?.inputCharactersLimit ||
```
bgambocjaviniar: would like you think using `nullish coalescing` is better than the or operator for this case?

Let's say the customer sets it at 0, do you want it to evaluate to the lefthand side or the right handside (default value)

reason being, the OR operator will evaluate 0 as `falsy`, whereas the nullish opeartor evaluates as truthy.

feature-libs/customer-ticketing/components/customer-ticketing-dialog/customer-ticketing-dialog.component.ts
```diff
@@ -0,0 +1,86 @@
...
+    protected launchDialogService: LaunchDialogService,
+    protected el: ElementRef,
+    protected customerTicketingConfig: CustomerTicketingConfig,
+    protected filesFormValidators: FilesFormValidators
+  ) {}
+
+  close(reason: string): void {
+    this.launchDialogService.closeDialog(reason);
+  }
+}
```
bgambocjaviniar: not related to this line

you can include an abstract method, so the extended class will have to implement the form


abstract buildForm(): void

feature-libs/customer-ticketing/components/customer-ticketing-dialog/default-customer-ticketing-form-layout.config.ts
```diff
@@ -1,9 +1,9 @@
 import { DIALOG_TYPE, LayoutConfig } from '@spartacus/storefront';
-import { CustomerTicketingReopenDialogComponent } from './customer-ticketing-reopen-dialog.component';
+import { CustomerTicketingReopenDialogComponent } from '../customer-ticketing-reopen';
 
 export const defaultCustomerTicketingFormLayoutConfig: LayoutConfig = {
```
bgambocjaviniar: instead of putting this 'dialog' within the shared dialog component, you can just place it in the individual folders to come like the create, reopen, and close as part of it's small 'domain'

feature-libs/customer-ticketing/components/customer-ticketing-reopen/customer-ticketing-reopen-dialog/customer-ticketing-reopen-dialog.component.ts
```diff
@@ -1,81 +1,20 @@
...
     this.buildForm();
   }
 
-  close(reason: string): void {
-    this.launchDialogService.closeDialog(reason);
-  }
-
   reopenRequest(): void {
     if (this.form.invalid) {
       this.form.markAllAsTouched();
```
bgambocjaviniar, line 19: line 25 `protected buildForm() {`

let's explicitly put the 'type' as `:void`

feature-libs/customer-ticketing/components/customer-ticketing-dialog/customer-ticketing-dialog.component.ts
```diff
@@ -0,0 +1,86 @@
...
+
+  get messagesCharacterLeft(): number {
+    return (
+      this.inputCharactersLimit - (this.form.get('message')?.value?.length || 0)
+    );
+  }
+
+  get allowedTypes(): string[] | undefined {
+    return this.customerTicketingConfig.customerTicketing
+      ?.attachmentRestrictions?.allowedTypes;
```
bgambocjaviniar: should we have a default constant and value for allowed types as well?

bgambocjaviniar (changes requested):

feature-libs/customer-ticketing/components/customer-ticketing-dialog/customer-ticketing-dialog.component.ts
Parthlakhani: Yes, let's use nullish coalescing.

feature-libs/customer-ticketing/components/customer-ticketing-dialog/customer-ticketing-dialog.component.ts
Parthlakhani: got it

feature-libs/customer-ticketing/components/customer-ticketing-dialog/customer-ticketing-dialog.component.ts
Parthlakhani: If the value is null or empty array, it will allow all types of files to be uploaded.
So, I think it's better if we don't defined the default constant.

feature-libs/customer-ticketing/components/customer-ticketing-dialog/customer-ticketing-dialog.component.ts
Parthlakhani: I added the whole method in abstract component instead as 2 fields are common in all components and that way only create component will need to add the form fields.

feature-libs/customer-ticketing/components/customer-ticketing-dialog/default-customer-ticketing-form-layout.config.ts
bgambocjaviniar: @Parthlakhani I think this was missed

the default 'layout' config would be placed in the /customer-ticketing-reopen folder, and you would rename the file as `default-customer-ticketing-reopen-layout.config.ts`

bgambocjaviniar: basically, each of these dialog will have it's own layout config within it's of its own folder

bgambocjaviniar (approved):

khristian23 (approved):

PR #16116, opened 2022-08-29
chore: Update the feature level.

plabadie:
closes https://jira.tools.sap/browse/CXSPA-845


ercultimate:
What about the CSS versioning in storefrontstyles?  Doesn't that need updated with the new major version or somethign set in storefrontapp SCSS?

plabadie:
> What about the CSS versioning in storefrontstyles? Doesn't that need updated with the new major version or somethign set in storefrontapp SCSS?


CSS already uses `$useLatestStyles: true;`  as configured in `projects/storefrontapp/src/styles.scss`

Although we need to validate it works as intended.  If it need fixing it will be part of a separate ticket.

Dalvir-Personal (approved):

PR #16124, opened 2022-08-31
chore: Make base branches that runs the pipeline to have distinct values

bgambocjaviniar:


justinlee01 (approved):
Out of curiosity - does `github.event.pull_request.head.sha` evaluate early with `undefined` if none of these properties are defined? Or are they all defined? Not familiar with Git workflows.

hackergil (approved):

bgambocjaviniar:
> Out of curiosity - does `github.event.pull_request.head.sha` evaluate early with `undefined` if none of these properties are defined? Or are they all defined? Not familiar with Git workflows.

sorry mis-spoke in the thread, but it evaluates to an empty string, which is why `OR` operator is used instead of `??` nullish coalescing.

You can find the payload of github context (which include events) here 
Screen Shot 2022-08-31 at 12 57 59 PM


Screen Shot 2022-08-31 at 12 59 44 PM

PR #16126, opened 2022-08-31
Feature/cxec 17290

I565452:


feature-libs/order/assets/translations/en/order.i18n.ts
```diff
@@ -14,7 +14,7 @@ export const order = {
     emptyPurchaseOrderId: 'None',
     none: 'None {{value}}',
     placed: 'Placed',
-    placedBy: 'Placed By',
+    placedBy: 'Buyer',
```
I539321: This does not seem like a good idea as it might break existing usages. I'd rather add a new entry buyer with the 'Buyer' value. Having a key named placedBy and its value is 'Buyer; does not feel right.

Perhaps I'm missing something, in either case let me know.

I539321 (changes requested):

feature-libs/order/assets/translations/en/order.i18n.ts
I565452: actually it is not used anywhere, but you are right - I updated

feature-libs/order/components/order-details/order-overview/order-overview.component.ts
```diff
@@ -16,15 +29,27 @@ import { filter, map } from 'rxjs/operators';
...
 
-  constructor(protected translation: TranslationService) {}
+  constructor(
+    protected translation: TranslationService,
+    protected userAccountFacade: UserAccountFacade
+  ) {}
+
+  ngOnInit(): void {
+    this.userAccountFacade.get().subscribe((user: User | undefined) => {
+      this.isUserOrgUnitViewer = !!user?.roles?.includes(
```
I539321: Is the !! needed here? The `includes` method returns a boolean. Maybe your intent is to use it for the user object?

I539321 (changes requested):

I539321 (approved):

feature-libs/order/components/order-details/order-overview/order-overview.component.ts
I565452: it is needed since `user?` returns `boolean | undefined` and this causes a type check error
I changed the variable type

feature-libs/order/components/order-details/order-overview/order-overview.component.spec.ts
```diff
@@ -103,6 +103,13 @@ const mockOrder: Order = {
       name: 'Rustic',
     },
   },
+  orgCustomer: {
+    uid: 'gi.sun@rustic-hw.com',
+    name: 'Gi Sun',
+  },
+  orgUnit: {
```
MarcinGanczorz, line 122: we could fix typecheck for orgUnit

feature-libs/order/components/order-details/order-overview/order-overview.component.spec.ts
```diff
@@ -486,4 +493,48 @@ describe('OrderOverviewComponent', () => {
...
+          expect(data.title).toEqual('test');
+          expect(data.text).toEqual([mockOrder.orgUnit.name]);
+        })
+        .unsubscribe();
+
+      expect(component.getUnitNameCardContent).toHaveBeenCalledWith(
+        mockOrder.orgUnit.name
+      );
+    });
+  });
```
MarcinGanczorz: i don't see any test that use new ngOnInit method for order-overview component. Test code coverage is missing there

feature-libs/order/components/replenishment-order-details/replenishment-order-cancellation/replenishment-order-cancellation.component.ts
```diff
@@ -44,6 +44,7 @@ export class ReplenishmentOrderCancellationComponent implements OnDestroy {
       this.vcr
     );
 
+    /* istanbul ignore if */
```
MarcinGanczorz: Do we really want to ignore code coverage here? it's not our code.

feature-libs/order/components/replenishment-order-history/replenishment-order-history.component.ts
```diff
@@ -109,6 +109,7 @@ export class ReplenishmentOrderHistoryComponent implements OnDestroy {
       replenishmentOrderCode
     );
 
+    /* istanbul ignore if */
```
MarcinGanczorz: Do we really want to ignore code coverage here? it's not our code.

I565452: I think yes, the same explanation as above

feature-libs/order/components/replenishment-order-details/replenishment-order-cancellation/replenishment-order-cancellation.component.ts
I565452: I think yes, this conditions only subcribe the dialog instance for release and there is no point to cover it with the test, but it decreases the test coverage significantly

feature-libs/order/components/order-details/order-overview/order-overview.component.spec.ts
I565452: updated, I added a test for ngOnInit, now it is covered

feature-libs/order/components/order-details/order-overview/order-overview.component.spec.ts
I565452: what do you mean?

PR #16133, opened 2022-09-06
CXSPA-1261: Changes to support Angular 14 in a backwards compatible way.

plabadie:
Fixes: https://jira.tools.sap/browse/CXSPA-1261


plabadie:
Replaced by #16547

PR #16139, opened 2022-09-09
Feature/cxspa 303

kimhw0630:
Disabled automatic scroll restoration when using Spartacus scroll position restoration feature.


projects/storefrontapp-e2e-cypress/cypress/integration/regression/scrolling/scroll-position-restoration.e2e-spec.ts
```diff
@@ -0,0 +1,36 @@
+context('scroll Position Restoration', () => {
+  const PRODUCT_NAME = 'DC Car Battery Adapter';
```
giancorderoortiz, line 2: Can default product be used instead? 
Less maintenance costs afterwards.

kimhw0630: @giancorderoortiz 
We need to select a product in the middle of the screen to check scrolling.
Thanks

Zeyber (changes requested):
This PR needs to be updated with the latest develop branch.

projects/storefrontlib/router/on-navigate.service.ts
```diff
@@ -51,6 +51,8 @@ export class OnNavigateService {
     this.subscription?.unsubscribe();
 
     if (enable) {
+      // Disable automatic scroll restoration to avoid race conditions
+      this.viewportScroller.setHistoryScrollRestoration('manual');
```
Zeyber: Even though this change is small and works well, we need to add feature flags for `5.1` and mention it in change logs as it is still breaking. As well as leaving cleanup comments to remove the logic for `6.0`.

Zeyber (changes requested):

projects/storefrontlib/router/on-navigate.service.ts
```diff
@@ -51,6 +52,11 @@ export class OnNavigateService {
     this.subscription?.unsubscribe();
 
     if (enable) {
+      // Disable automatic scroll restoration to avoid race conditions
+      // TODO: this feature level check should be removed in Major release
```
Zeyber: Please use the `@deprecated since 5.1` annotation as well so that deprecated changes are searchable when upgrading to the next major.

projects/storefrontlib/router/on-navigate.service.ts
kimhw0630: @Zeyber 
added feature flag .. thanks

Zeyber (changes requested):

projects/storefrontlib/router/on-navigate.service.ts
Zeyber: Also, changelog `6_0.md` should be updated.

kimhw0630: done thanks

Zeyber (approved):

PR #16142, opened 2022-09-12
Feature/cxec 9980 Integrate APIs in create new ticket dialog

manpreetlamba4:
This PR contains changes for adding 'Categories' API and 'Associated Objects' API in the customer ticketing create dialog

https://jira.tools.sap/browse/CXEC-16778
https://jira.tools.sap/browse/CXEC-16779

### DIFF
--- a/feature-libs/customer-ticketing/components/customer-ticketing-create/customer-ticketing-create-dialog/customer-ticketing-create-dialog.component.html
+++ b/feature-libs/customer-ticketing/components/customer-ticketing-create/customer-ticketing-create-dialog/customer-ticketing-create-dialog.component.html
@@ -41,7 +41,9 @@
             {{ 'customerTicketing.category' | cxTranslate }}
           </span>
           <select class="form-control">
-            <option></option>
+            <option *ngFor="let category of ticketCategories">
+              {{ category.name }}
+            </option>
           </select>
 
           <cx-form-errors [control]="form.get('message')"></cx-form-errors>
@@ -52,7 +54,12 @@
             {{ 'customerTicketing.associateTo' | cxTranslate }}
           </span>
           <select class="form-control">
-            <option></option>
+            <option
+              *ngFor="let objects of ticketAssociatedObjects"
+              [ngValue]="c"
+            >
+              {{ objects.type }} {{ objects.code }}
+            </option>
           </select>
         </label>
 
--- a/feature-libs/customer-ticketing/components/customer-ticketing-create/customer-ticketing-create-dialog/customer-ticketing-create-dialog.component.html
+++ b/feature-libs/customer-ticketing/components/customer-ticketing-create/customer-ticketing-create-dialog/customer-ticketing-create-dialog.component.html
@@ -41,7 +41,7 @@
             {{ 'customerTicketing.category' | cxTranslate }}
           </span>
           <select class="form-control">
-            <option *ngFor="let category of ticketCategories">
+            <option *ngFor="let category of ticketCategories$ | async">
               {{ category.name }}
             </option>
           </select>
@@ -54,7 +54,9 @@
             {{ 'customerTicketing.associateTo' | cxTranslate }}
           </span>
           <select class="form-control">
-            <option *ngFor="let associatedObject of ticketAssociatedObjects">
+            <option
+              *ngFor="let associatedObject of ticketAssociatedObjects$ | async"
+            >
               {{ associatedObject.type }} {{ associatedObject.code }}
             </option>
           </select>
--- a/feature-libs/customer-ticketing/components/customer-ticketing-create/customer-ticketing-create-dialog/customer-ticketing-create-dialog.component.html
+++ b/feature-libs/customer-ticketing/components/customer-ticketing-create/customer-ticketing-create-dialog/customer-ticketing-create-dialog.component.html
@@ -54,11 +54,8 @@
             {{ 'customerTicketing.associateTo' | cxTranslate }}
           </span>
           <select class="form-control">
-            <option
-              *ngFor="let objects of ticketAssociatedObjects"
-              [ngValue]="c"
-            >
-              {{ objects.type }} {{ objects.code }}
+            <option *ngFor="let associatedObject of ticketAssociatedObjects">
+              {{ associatedObject.type }} {{ associatedObject.code }}
             </option>
           </select>
         </label>
--- a/feature-libs/customer-ticketing/components/customer-ticketing-create/customer-ticketing-create-dialog/customer-ticketing-create-dialog.component.ts
+++ b/feature-libs/customer-ticketing/components/customer-ticketing-create/customer-ticketing-create-dialog/customer-ticketing-create-dialog.component.ts
@@ -1,6 +1,7 @@
 import { Component, OnInit } from '@angular/core';
 import { AssociatedObject, Category } from '@spartacus/customer-ticketing/root';
 import { FormUtils } from '@spartacus/storefront';
+import { Observable } from 'rxjs';
 import { CustomerTicketingDialogComponent } from '../../shared/customer-ticketing-dialog/customer-ticketing-dialog.component';
 @Component({
   selector: 'cx-customer-ticketing-create-dialog',
@@ -10,20 +11,12 @@ export class CustomerTicketingCreateDialogComponent
   extends CustomerTicketingDialogComponent
   implements OnInit
 {
-  ticketCategories: Array<Category>;
-  ticketAssociatedObjects: Array<AssociatedObject>;
+  ticketCategories$: Observable<Category[]> =
+    this.customerTicketingFacade.getTicketCategories();
+  ticketAssociatedObjects$: Observable<AssociatedObject[]> =
+    this.customerTicketingFacade.getTicketAssociatedObjects();
 
   ngOnInit(): void {
-    this.customerTicketingFacade
-      .getTicketCategories()
-      .subscribe((categories) => {
-        this.ticketCategories = categories;
-      });
-    this.customerTicketingFacade
-      .getTicketAssociatedObjects()
-      .subscribe((associatedObjects) => {
-        this.ticketAssociatedObjects = associatedObjects;
-      });
     this.buildForm();
   }
 
--- a/feature-libs/customer-ticketing/components/customer-ticketing-create/customer-ticketing-create-dialog/customer-ticketing-create-dialog.component.ts
+++ b/feature-libs/customer-ticketing/components/customer-ticketing-create/customer-ticketing-create-dialog/customer-ticketing-create-dialog.component.ts
@@ -1,7 +1,10 @@
 import { Component, OnInit } from '@angular/core';
+import {
+  AssociatedObjects,
+  Category,
+} from '@spartacus/customer-ticketing/root';
 import { FormUtils } from '@spartacus/storefront';
 import { CustomerTicketingDialogComponent } from '../../shared/customer-ticketing-dialog/customer-ticketing-dialog.component';
-
 @Component({
   selector: 'cx-customer-ticketing-create-dialog',
   templateUrl: './customer-ticketing-create-dialog.component.html',
@@ -10,7 +13,18 @@ export class CustomerTicketingCreateDialogComponent
   extends CustomerTicketingDialogComponent
   implements OnInit
 {
+  ticketCategories: Array<Category>;
+  ticketAssociatedObjects: Array<AssociatedObjects>;
+
   ngOnInit(): void {
+    this.customerTicketingFacade.getTicketCategories().subscribe((category) => {
+      this.ticketCategories = category;
+    });
+    this.customerTicketingFacade
+      .getTicketAssociatedObjects()
+      .subscribe((associatedObject) => {
+        this.ticketAssociatedObjects = associatedObject;
+      });
     this.buildForm();
   }
 
--- a/feature-libs/customer-ticketing/components/customer-ticketing-create/customer-ticketing-create-dialog/customer-ticketing-create-dialog.component.ts
+++ b/feature-libs/customer-ticketing/components/customer-ticketing-create/customer-ticketing-create-dialog/customer-ticketing-create-dialog.component.ts
@@ -1,8 +1,5 @@
 import { Component, OnInit } from '@angular/core';
-import {
-  AssociatedObjects,
-  Category,
-} from '@spartacus/customer-ticketing/root';
+import { AssociatedObject, Category } from '@spartacus/customer-ticketing/root';
 import { FormUtils } from '@spartacus/storefront';
 import { CustomerTicketingDialogComponent } from '../../shared/customer-ticketing-dialog/customer-ticketing-dialog.component';
 @Component({
@@ -14,16 +11,18 @@ export class CustomerTicketingCreateDialogComponent
   implements OnInit
 {
   ticketCategories: Array<Category>;
-  ticketAssociatedObjects: Array<AssociatedObjects>;
+  ticketAssociatedObjects: Array<AssociatedObject>;
 
   ngOnInit(): void {
-    this.customerTicketingFacade.getTicketCategories().subscribe((category) => {
-      this.ticketCategories = category;
-    });
+    this.customerTicketingFacade
+      .getTicketCategories()
+      .subscribe((categories) => {
+        this.ticketCategories = categories;
+      });
     this.customerTicketingFacade
       .getTicketAssociatedObjects()
-      .subscribe((associatedObject) => {
-        this.ticketAssociatedObjects = associatedObject;
+      .subscribe((associatedObjects) => {
+        this.ticketAssociatedObjects = associatedObjects;
       });
     this.buildForm();
   }
--- a/feature-libs/customer-ticketing/components/shared/customer-ticketing-dialog/customer-ticketing-dialog.component.spec.ts
+++ b/feature-libs/customer-ticketing/components/shared/customer-ticketing-dialog/customer-ticketing-dialog.component.spec.ts
@@ -1,5 +1,6 @@
 import { Component } from '@angular/core';
 import { ComponentFixture, TestBed } from '@angular/core/testing';
+import { I18nTestingModule } from '@spartacus/core';
 import { CustomerTicketingDialogComponent } from './customer-ticketing-dialog.component';
 
 @Component({
@@ -13,6 +14,7 @@ describe('CustomerTicketingDialogComponent', () => {
 
   beforeEach(async () => {
     await TestBed.configureTestingModule({
+      imports: [I18nTestingModule],
       declarations: [CustomerTicketingDialogComponent],
     }).compileComponents();
   });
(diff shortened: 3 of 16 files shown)


adolfobee (approved):

feature-libs/customer-ticketing/components/customer-ticketing-create/customer-ticketing-create-dialog/customer-ticketing-create-dialog.component.html
```diff
@@ -52,7 +54,12 @@
             {{ 'customerTicketing.associateTo' | cxTranslate }}
           </span>
           <select class="form-control">
-            <option></option>
+            <option
+              *ngFor="let objects of ticketAssociatedObjects"
```
khristian23:
```suggestion
              *ngFor="let associatedObject of ticketAssociatedObjects"
```
should be in singular

feature-libs/customer-ticketing/components/customer-ticketing-create/customer-ticketing-create-dialog/customer-ticketing-create-dialog.component.html
```diff
@@ -52,7 +54,12 @@
...
+              [ngValue]="c"
```
khristian23: What does this mean ngValue = c?

feature-libs/customer-ticketing/components/customer-ticketing-create/customer-ticketing-create-dialog/customer-ticketing-create-dialog.component.ts
```diff
@@ -10,7 +13,18 @@ export class CustomerTicketingCreateDialogComponent
   extends CustomerTicketingDialogComponent
   implements OnInit
 {
+  ticketCategories: Array<Category>;
+  ticketAssociatedObjects: Array<AssociatedObjects>;
+
   ngOnInit(): void {
+    this.customerTicketingFacade.getTicketCategories().subscribe((category) => {
```
khristian23: shouldn't it be categories in plural?

feature-libs/customer-ticketing/components/customer-ticketing-create/customer-ticketing-create-dialog/customer-ticketing-create-dialog.component.ts
```diff
@@ -10,7 +13,18 @@ export class CustomerTicketingCreateDialogComponent
...
+      this.ticketCategories = category;
+    });
+    this.customerTicketingFacade
+      .getTicketAssociatedObjects()
+      .subscribe((associatedObject) => {
```
khristian23: Same here, it should be plural

feature-libs/customer-ticketing/root/model/customer-ticketing.model.ts
```diff
@@ -29,6 +29,20 @@ export interface Category {
   name: string;
 }
 
+export interface CategoryList {
+  ticketCategories?: Category[];
+}
+
+export interface AssociatedObjects {
```
khristian23:
```suggestion
export interface AssociatedObject {
```
This interface represents only one Object.  It should be in singular

feature-libs/customer-ticketing/root/model/customer-ticketing.model.ts
```diff
@@ -29,6 +29,20 @@ export interface Category {
   name: string;
 }
 
+export interface CategoryList {
```
khristian23:
```suggestion
export interface CategoriesList {
```
Or ListOfCategories
Or categoryListing

feature-libs/customer-ticketing/core/facade/customer-ticketing.service.ts
```diff
@@ -19,6 +25,18 @@ import { CustomerTicketingConnector } from '../connectors';
...
+  }
+  getTicketCategoriesQueryResetEvents(): QueryNotifier[] | undefined {
+    return [GetTicketCatQueryResetEvent];
+  }
+  getTicketAssociatedObjectsQueryReloadEvents(): QueryNotifier[] | undefined {
+    return [GetTicketAssociatedObjectsQueryReloadEvent];
+  }
+  getTicketAssociatedObjectsQueryResetEvents(): QueryNotifier[] | undefined {
+    return [GetTicketAssociatedObjectsQueryResetEvent];
+  }
```
khristian23: Why do these methods return. QueryNotifier[] or undefined?
Shouldn't be enough to return QueryNotifier[], since their implementations are hardcoded to return an array of one element?

feature-libs/customer-ticketing/core/facade/customer-ticketing.service.ts
```diff
@@ -70,6 +130,19 @@ export class CustomerTicketingService implements CustomerTicketingFacade {
       })
     );
   }
+  protected customerTicketingAssociatedObjectsPreConditions(): Observable<string> {
+    return this.userIdService.getUserId().pipe(
+      take(1),
+      map((userId) => {
+        if (!userId) {
+          throw new Error(
+            'Customer ticketing associated objects pre conditions not met'
```
khristian23, line 139: Is this a message that can be seen by the customer?  If so, you need to make it transalatable

feature-libs/customer-ticketing/occ/adapters/occ-customer-ticketing.adapter.ts
```diff
@@ -20,6 +28,44 @@ export class OccCustomerTicketingAdapter implements CustomerTicketingAdapter {
...
+      },
+    });
+  }
+
+  getTicketCategories(): Observable<Category[]> {
+    return this.http.get<CategoryList>(this.getTicketCategoriesEndpoint()).pipe(
+      catchError((error) => throwError(normalizeHttpError(error))),
+      map((categoryList) => categoryList.ticketCategories ?? []),
+      this.converter.pipeableMany(CUSTOMER_TICKETING_CATEGORY_NORMALIZER)
+    );
```
khristian23: Cna it be pretty printed?

khristian23 (changes requested):
I understand changing interface AssociatedObjects by its correspondent singular AssociatedObject will imply changing many places, but it should not be plural since it represents a single object

feature-libs/customer-ticketing/components/customer-ticketing-create/customer-ticketing-create-dialog/customer-ticketing-create-dialog.component.html
bgambocjaviniar: I'd like to know to know as well  +1 


My current assumption is that it is dead code as this form is not doing anything. 'c' was supposed to be objects, I believe as well

First, you would need to attach an event listener, specifically the change event (`select (change)='etcFunctionToupdateTheForm(passingTheValueSelected)'`) to the select tag, which you would create in order to update the form

---
If you search for `<option` in the codebase, you can see how to attach an event listener, and what to do with it

bgambocjaviniar: Just to add more to Christian's other comments, it would have to be singular, similarly to how you named things in line 44 of this fine.

feature-libs/customer-ticketing/components/customer-ticketing-create/customer-ticketing-create-dialog/customer-ticketing-create-dialog.component.html
```diff
@@ -41,7 +41,9 @@
             {{ 'customerTicketing.category' | cxTranslate }}
           </span>
           <select class="form-control">
-            <option></option>
+            <option *ngFor="let category of ticketCategories">
```
bgambocjaviniar: similarly to https://github.com/SAP/spartacus/pull/16142/files#r988362362

it doesn't do anything. It just 'displays' a dropdown, but the values aren't being passed to the APIs

feature-libs/customer-ticketing/components/customer-ticketing-create/customer-ticketing-create-dialog/customer-ticketing-create-dialog.component.ts
bgambocjaviniar: If you were to subscribe in the component .ts file, you would need to always unsubscribe on component destroy as it will cause memory leaks.

You have options:

1 - let angular handle everything by using the `async` pipe in the template ('.html'). What this does for you is it subscribes to the observable, and unsubscribes to it whenever the component gets destroyed.


2 - Do what you're doing here, but you would need to create a subscription object, such as `const subscriptions = new Subscription()` from `rxjs`. and add every subscribed things in the component level. You would need to implement the `OnDestroy` to handle the lifecycle when the component gets destryoed, and therefore unsubscribe in one shot.

bgambocjaviniar: I would prefer, you would go with the 1st approach as it's less code and more compact. 

That would also mean, you would make the variables of type Observable< insert type here >

feature-libs/customer-ticketing/components/customer-ticketing-create/customer-ticketing-create-dialog/customer-ticketing-create-dialog.component.ts
bgambocjaviniar: same subscription problem here https://github.com/SAP/spartacus/pull/16142/files#r988362872

feature-libs/customer-ticketing/components/shared/customer-ticketing-dialog/customer-ticketing-dialog.component.spec.ts
```diff
@@ -13,6 +14,7 @@ describe('CustomerTicketingDialogComponent', () => {
 
   beforeEach(async () => {
     await TestBed.configureTestingModule({
+      imports: [I18nTestingModule],
       declarations: [CustomerTicketingDialogComponent],
     }).compileComponents();
```
bgambocjaviniar, line 19: Is there a dedicated ticket for unit tests here? I thought it would be done once the apis will be integrated in conjunction with the template that was created a while back

feature-libs/customer-ticketing/components/customer-ticketing-create/customer-ticketing-create-dialog/customer-ticketing-create-dialog.component.ts
```diff
@@ -10,7 +13,18 @@ export class CustomerTicketingCreateDialogComponent
...
+        this.ticketAssociatedObjects = associatedObject;
+      });
     this.buildForm();
   }
```
bgambocjaviniar: createTicketRequest  --- line 17

you would need to utilize this function to pass to the newly integrated service to use your apis

feature-libs/customer-ticketing/components/shared/customer-ticketing-dialog/customer-ticketing-dialog.component.spec.ts
bgambocjaviniar: Applies to all specs here so I don't flag every single one of them

feature-libs/customer-ticketing/core/facade/customer-ticketing.service.spec.ts
```diff
@@ -93,4 +93,62 @@ describe('CustomerTicketingService', () => {
         });
     });
   });
+  describe('getTicketCategories', () => {
```
bgambocjaviniar, line 96: nice 👍

feature-libs/customer-ticketing/core/facade/customer-ticketing.service.ts
bgambocjaviniar: You can re-use the `customerTicketingPreConditions` as it returns a tuple.

The current logic says if ticketId doesn't exist, throw an error. We can just remove that condition and make it able to return a tuple of [userId, ticketId || undefined]. The way to do this is to make sure your second stream that is being combined can startWith `undefined`

bgambocjaviniar: Because the way I see it, we don't always need the ticketId. Only for certain cases when it's within the ticket details page.

feature-libs/customer-ticketing/root/events/customer-ticketing-event.listener.ts
```diff
@@ -45,6 +49,56 @@ export class CustomerTicketingEventListener implements OnDestroy {
     );
   }
 
+  protected onGetTicketCatQueryReset(): void {
```
bgambocjaviniar: 1- for all these created functions, they needed to be called in the constructor

2 - we don't need all these functions, you can place this in one function that has the merges of the stream of `LogoutEvent` and `LoginEvent`, which will fire those events all in one because at this point, it is just DRY

-----
The 1- would make sense if we're making multiple event listeners for each of the `facade`, but it looks like the direction we want to go is to have one facade that encapsulates the whole functionality. I don't have much preference. It is entirely up to you guys on whether you want to have decouple the facade into separate ones to keep the service logic different. If not, then let's go with 2.

bgambocjaviniar (changes requested):
1 - missing a lot of functionality in terms of integrating the apis for creating a new ticket.
- the component and apis aren't fully integrated. User can't create a ticket, nor the api to create the ticket is done

2 - dialog only contains to display data, but not working with it, for example, updating the form based on what is selected and doing something with it

bgambocjaviniar:
> 1 - missing a lot of functionality in terms of integrating the apis for creating a new ticket.
> 
> * the component and apis aren't fully integrated. User can't create a ticket, nor the api to create the ticket is done
> 
> 2 - dialog only contains to display data, but not working with it, for example, updating the form based on what is selected and doing something with it

scratch 1-

The title misled me to believe this PR is supposed to integrate the entire API for creating a dialog, but apparently it's just to get the categories and associated objects in the dropdown. But, the 2- still applies, where the form is not going to be updated as it's not fully integrated with api.

I am confused though as the branch name is referencing this jira, which would mean it 'should' include all the apis and fully integrated with your template. https://jira.tools.sap/browse/CXEC-9980.

I would create the appropriate branch that corresponds to what this PR is actually for and change the title of the PR

feature-libs/customer-ticketing/core/facade/customer-ticketing.service.ts
```diff
@@ -45,13 +63,55 @@ export class CustomerTicketingService implements CustomerTicketingFacade {
...
+    QueryState<AssociatedObject[]>
+  > {
+    return this.getTicketAssociatedObjectsQuery.getState();
+  }
+  getTicketAssociatedObjects(): Observable<AssociatedObject[]> {
+    return this.getTicketAssociatedObjectsState().pipe(
+      map((state) => state.data ?? [])
+    );
+  }
+  getTicketCategories(): Observable<Category[]> {
```
bgambocjaviniar, line 106: Let's switch the function just for the sake of reading the 'service', where it's first the state then the 'data'.

Basically swap line 106 function with line 1123 function

```
  getTicketCategoriesState(): Observable<QueryState<Category[]>> {
    return this.getTicketCategoriesQuery.getState();
  }

getTicketCategories(): Observable<Category[]> {
....
}
```

feature-libs/customer-ticketing/core/facade/customer-ticketing.service.ts
```diff
@@ -63,13 +123,26 @@ export class CustomerTicketingService implements CustomerTicketingFacade {
...
       map(([userId, ticketId]) => {
-        if (!userId || !ticketId) {
+        if (!userId) {
           throw new Error('Customer ticketing pre conditions not met');
         }
         return [userId, ticketId];
       })
     );
   }
+  protected customerTicketingAssociatedObjectsPreConditions(): Observable<string> {
```
bgambocjaviniar, line 133: needs to be removed as it's not longer used

feature-libs/customer-ticketing/core/facade/customer-ticketing.service.ts
```diff
@@ -63,13 +123,26 @@ export class CustomerTicketingService implements CustomerTicketingFacade {
...
+      map((userId) => {
+        if (!userId) {
+          throw new Error(
+            'Customer ticketing associated objects pre conditions not met'
+          );
+        }
+        return userId;
+      })
+    );
+  }
```
bgambocjaviniar, line 145: nit, can you move getTicketState and getTicket above the precondition function

feature-libs/customer-ticketing/root/events/customer-ticketing.events.ts
```diff
@@ -7,3 +7,19 @@ export class GetTicketQueryResetEvent extends CxEvent {
 export class GetTicketQueryReloadEvent extends CxEvent {
   static readonly type = 'GetTicketQueryReloadEvent';
 }
+
+export class GetTicketCatQueryResetEvent extends CxEvent {
```
bgambocjaviniar, line 11: Let's explicitly say Category instead of Cat

bgambocjaviniar: same for other instances in this PR

feature-libs/customer-ticketing/core/facade/customer-ticketing.service.spec.ts
bgambocjaviniar: These tests are actually failing

have you ran `ng test customer-tickerting`. I said nice above because I saw tests, but they're not passing. 
What needs to be done is that it requires the mock connector to be aware of these functions, so include them here https://github.com/SAP/spartacus/pull/16142/files#diff-450d65a6cbacd13bee796c96a548940c8ee202110c969f6fe9e25a335fda26ddR32

bgambocjaviniar (changes requested):
1 - first thing first. Let's fix the convention on creating branches that relates to a jira ticket. Let's change this head branch to feature/CXEC-16778 instead (which would include changes from 16779, which is fine), but it's part of a subtask. If we were to merge this branch, it's like saying  you fully integrated the API to create a new ticket dialog, but that is false as 9980 is not done. I mentioned here to create a new branch and change the title https://github.com/SAP/spartacus/pull/16142#issuecomment-1275549610

2 - I un-resolved this https://github.com/SAP/spartacus/pull/16142#discussion_r992958938. What I want is a function that handles firing all the expected reset/reload on login / reload. 

Basically, something like below

```
onLoginAndLogoutEvent(): {
this.subscriptions.add(
      merge(
        this.eventService.get(LogoutEvent),
        this.eventService.get(LoginEvent)
      ).subscribe(() => {
dispatch all reload / reset of everything that hooks into login/logout in this file to avoid DRY. <---- important
}
}
```

3 - the service unit tests aren't passing, look at the comments I wrote on how to fix. I also left other tiny fixes to do

manpreetlamba4:
All comments are resolved in https://github.com/SAP/spartacus/pull/16301

PR #16144, opened 2022-09-12
chore: Spartacus CCv2 repo update on demand

bgambocjaviniar:
closes https://jira.tools.sap/browse/CXSPA-1247


.github/workflows/deploy-ccv2.yml
```diff
@@ -1,11 +1,9 @@
 on:
-  schedule:
-    - cron:  '0 0 * * *'
   workflow_dispatch:
     inputs:
       ccv2_branch:
         description: Branch on ccv2 to update
-        default: spa_p4_dist_test
+        default: spa_p4_it_works
```
bgambocjaviniar: will change to `spa_p4` before we merge

ci-scripts/deploy-ccv2.sh
```diff
@@ -42,49 +44,99 @@ else
...
+    echo "Dist folder has been updated"
+else
+    echo "Dist folder has NOT been updated"
+    exit 1
+fi
+
+echo "---------------------------------------------------------------------------------------------------------------------------"
+echo "Build SSR for b2b storefront"
+
+export SPA_ENV='b2b'
```
bgambocjaviniar: to help clarify why this is here is because in the package script, we allow to consume $SPA_ENV for any 'added/modified' env. We need to build for b2b so it was either using that mechanism or create a package script for b2b:build and buid:b2b:ssr etc

ci-scripts/deploy-ccv2.sh
```diff
@@ -0,0 +1,149 @@
...
+
+cp -a dist/storefrontapp/. $CCV2_B2B_STOREFRONT_PATH/dist/$B2B_STORE/browser/
+cp -a dist/storefrontapp-server/. $CCV2_B2B_STOREFRONT_PATH/dist/$B2B_STORE/server/
+
+echo "---------------------------------------------------------------------------------------------------------------------------"
+echo "Push to remote repository"
+
+cd $GHT_REPO
+
+git config --global user.email louis.pierrestiger@sap.com
```
bgambocjaviniar: reason why it is louis email is because that's what is set in the global ccv2 script

.github/workflows/update-ccv2.yml
```diff
@@ -0,0 +1,51 @@
...
+      - name: Cache node_modules
+        id: cache-node-modules
+        uses: actions/cache@v2
+        with:
+          path: | 
+            node_modules
+          key: nodemodules-${{ steps.branch-to-deploy-commit-id.outputs.commit_hash }}
+          restore-keys: nodemodules-${{ steps.branch-to-deploy-commit-id.outputs.commit_hash }}
+      - name: Install angular CLI
+        run: npm install -g @angular/cli@v13-lts
```
plabadie: Is it using angular v13 on purpose or should it be bumped to 14?

ci-scripts/update-ccv2.sh
```diff
@@ -0,0 +1,177 @@
...
+
+verify_branch_exist "origin" "$SOURCE_BRANCH_TO_DEPLOY"
+
+echo "------------------------------------------------------------------"
+echo "Verify CCv2 branch exist"
+
+verify_branch_exist "$GH_BASE_URL" "$CCV2_BRANCH"
+
+echo "------------------------------------------------------------------"
+echo "Comment out occBaseUrl from configration to allow index.html meta tag to set the occBaseUrl"
```
plabadie, line 53: Doesn't the index.html meta tag have precedence over any "internal" config?

ci-scripts/update-ccv2.sh
```diff
@@ -0,0 +1,177 @@
...
+    cat "$1"
+
+    if grep -Fq "addToHomeScreen: true" "$1"
+    then
+        echo "PWA config has NOT been removed"
+        exit 1
+    else
+        echo "PWA config has SUCCESSFULLY been removed"
+    fi
+}
```
plabadie, line 35: Overall, it would be nice if at some point in the future, we had a way to provide a deployment specific config without having to use `sed` to modify the default files.  Just food for thought, but no action point required in this PR.

ci-scripts/update-ccv2.sh
```diff
@@ -0,0 +1,177 @@
...
+echo "------------------------------------------------------------------"
+echo "Push to remote repository"
+
+cd "$GHT_REPO"
+
+git config --global user.email dl_61c08cf5a3cac30261c7e88c@global.corp.sap
+git config --global user.name cx-cc-automation-serviceuser
+
+git add .
+git commit --allow-empty -m "Update with $SOURCE_BRANCH_TO_DEPLOY branch from source of Spartacus"
```
plabadie, line 176: Not really an issue, but it doesn't seem like we need `--allow-empty` here.

bgambocjaviniar: I believe there were times when there's nothing to update, for example, redeploying a branch twice when there was no change, so it would fail the action saying there is nothing to push.

Me putting the allow-empty is to just know that someone triggered it if ever it's empty, and to make the action green instead of red

ci-scripts/update-ccv2.sh
bgambocjaviniar: nay internal takes precedence

bgambocjaviniar: https://sap.github.io/spartacus-docs/configuring-base-url/

.github/workflows/update-ccv2.yml
bgambocjaviniar: I can bump to 14, but there's no issue. It's just getting an ng version on the machine.

But we still just clone spartacus and the repo is on ng 14 if it's develop / release/4.3.x


Should we? If yes, then we need to bump the other pipeline files as it's on v13-lts. I can open the PR up for right now

bgambocjaviniar: bam https://github.com/SAP/spartacus/pull/16214

ci-scripts/update-ccv2.sh
plabadie: Agreed.  If it is there to add robustness, then we keep it for sure..

ci-scripts/update-ccv2.sh
plabadie: Thanks!  I would have thought the precedence is the other way around.

plabadie (approved):

PR #16148, opened 2022-09-12
Rename scope

RadhepS:


plabadie (changes requested):
This has tremendous impact on both the product and the customer custom code.  We need to discuss the requirement.

PR #16152, opened 2022-09-14
refactor: remove ngbNav from storefinder library

Matejk00:
Closes: CXSPA-1260

### DIFF
--- a/feature-libs/storefinder/components/store-finder-search-result/store-finder-list/store-finder-list.component.html
+++ b/feature-libs/storefinder/components/store-finder-search-result/store-finder-list/store-finder-list.component.html
@@ -64,13 +64,13 @@
 
     <!-- mobile tabs for column set only -->
     <div *ngIf="locations?.stores" class="cx-columns-mobile">
-      <ul role="tablist" class="nav">
-        <li class="nav-item" *ngFor="let mode of displayModes | keyvalue">
+      <ul class="cx-nav" role="tablist">
+        <li class="cx-nav-item" *ngFor="let mode of displayModes | keyvalue">
           <button
             [id]="'tab-' + mode?.value"
             role="tab"
             [ngClass]="{
-              'nav-link': true,
+              'cx-nav-link': true,
               active: isDisplayModeActive(mode?.value)
             }"
             [attr.aria-controls]="'tab-' + mode?.value + '-panel'"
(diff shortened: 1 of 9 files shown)


feature-libs/storefinder/components/store-finder-search-result/store-finder-list/store-finder-list.component.html
```diff
@@ -63,66 +63,89 @@
...
-            {{ 'storeFinder.listView' | cxTranslate }}
-          </a>
-          <ng-template ngbNavContent>
-            <div class="cx-address-col">
-              <div class="cx-store-details" *ngIf="isDetailsModeVisible">
-                <cx-store-finder-store-description
-                  [location]="storeDetails"
-                  [disableMap]="true"
-                ></cx-store-finder-store-description>
+      <ul role="tablist" class="nav">
```
morganm58: We should add the `cx-` prefix to the nav class

feature-libs/storefinder/components/store-finder-search-result/store-finder-list/store-finder-list.component.html
```diff
@@ -63,66 +63,89 @@
...
+        <li class="nav-item" *ngFor="let mode of displayModes | keyvalue">
```
morganm58: We should add the `cx-` prefix to the nav-item class

feature-libs/storefinder/components/store-finder-search-result/store-finder-list/store-finder-list.component.html
Matejk00: @morganm58 this class depends on bootstrap styling and is provided by the bootstrap library. 
If we add such a prefix, all default styling will gone.

It's a similar usage like `btn` or `btn-primary` for example.

feature-libs/storefinder/components/store-finder-search-result/store-finder-list/store-finder-list.component.html
Matejk00: Same case as above.

feature-libs/storefinder/styles/components/_store-finder-list.scss
```diff
@@ -23,6 +23,13 @@
     .cx-map-col {
       height: 70vh;
     }
+
+    .nav-item {
```
morganm58: we should use a `cx-` class to add the styling here

psmul (approved):

feature-libs/storefinder/styles/components/_store-finder-list.scss
Matejk00: done

morganm58 (approved):

PR #16160, opened 2022-09-15
chore: PR improvements

znikola:


integration-libs/cdc/user-profile/forgot-password/cdc-forgot-password-component.service.ts
```diff
@@ -48,12 +48,14 @@ export class CDCForgotPasswordComponentService
           // Reset password using CDC Gigya SDK
           this.cdcJsService
             .resetPasswordWithoutScreenSet(this.form.value.userEmail)
-            .subscribe((isResetSuccess) => {
-              if (isResetSuccess) {
+            .subscribe((isResetResponse) => {
```
Platonn: nitpick: let's rename it to just `response`. it's not a boolean, so let's not start it with `is...`

integration-libs/cdc/user-profile/register/cdc-register-component.service.ts
```diff
@@ -11,27 +11,20 @@ import { User } from '@spartacus/user/account/root';
...
-                } else {
-                  userRegistered.error(null);
-                }
-              });
+    this.command.create(({ user }) =>
+      // Registering user through CDC Gigya SDK
+      this.cdcJSService.registerUserWithoutScreenSet(user).pipe(
+        tap((result) => {
+          if (result.status !== 'OK') {
+            throw new Error(`Received an unexpected status: ${result.status}`);
```
Platonn: Do we really need this code here?
When the status is different than `OK`, don't we already throw an error from the lower level method `registerUserWithoutScreenSet()`?

integration-libs/cdc/user-profile/register/cdc-register-component.service.ts
```diff
@@ -50,24 +43,26 @@ export class CDCRegisterComponentService extends RegisterComponentService {
    * @param user as UserSignUp
    */
   register(user: UserSignUp): Observable<User> {
+    if (!user.firstName || !user.lastName || !user.uid || !user.password) {
+      throw new Error(`The provided user is not valid: ${user}`);
```
Platonn: I believe, here we should `return throwError(new Error(...))`, because we're not in the scope of rxjs operator. I believe we want to emit an rxjs error instead of simply throw JS Error

integration-libs/cdc/user-profile/forgot-password/cdc-forgot-password-component.service.ts
znikola: Done

integration-libs/cdc/user-profile/register/cdc-register-component.service.ts
znikola: You're right, removed.

integration-libs/cdc/user-profile/register/cdc-register-component.service.ts
znikola: Great catch, fixed

Platonn (approved):

PR #16172, opened 2022-09-17
feat: bump Angular version from 14.1 to 14.2 and fix some command arguments

WeizhengSap:
Changes include:

- bump Angular version from 14.1 to 14.2
- remove warning of "Usage of ‘~’ in imports is deprecated"
- fix install scripts
- change some command line arguments


scripts/install/functions.sh
```diff
@@ -85,43 +85,44 @@ function create_shell_app {
 
 function add_b2b {
     if [ "${ADD_B2B_LIBS}" = true ] ; then
-        ng add --skip-confirmation @spartacus/organization@${SPARTACUS_VERSION} --interactive false
-        ng add --skip-confirmation @spartacus/checkout@${SPARTACUS_VERSION} --interactive false --features="Checkout-B2B" --features="Checkout-Scheduled-Replenishment"
+        ng add @spartacus/organization@${SPARTACUS_VERSION} --skip-confirmation --no-interactive
```
plabadie, line 88: Overall, we might need to update the spartacus schematics command line examples throughout the documentation.  Perhaps we need to make a full text search to catch instances that use parameters we need to change.

scripts/install/config.default.sh
```diff
@@ -41,7 +41,7 @@ SPARTACUS_PROJECTS=(
         )
 
 SPARTACUS_REPO_URL="https://github.com/SAP/spartacus.git"
-BRANCH='develop'
+BRANCH='release/5.0.x'
```
plabadie, line 44: That's ok, just remember not to port this change to develop when ng 14 is added there.

scripts/install/functions.sh
```diff
@@ -85,43 +85,44 @@ function create_shell_app {
...
+        ng add @spartacus/checkout --skip-confirmation --no-interactive --features="Checkout-B2B" --features="Checkout-Scheduled-Replenishment"
```
plabadie: Smoke test comment for PR:  The SSR starts, but the CSR does not start.
PM2 says 
```
Error [ERR_REQUIRE_ESM]: require() of ES Module .../.nvm/versions/node/v16.14.2/lib/node_modules/serve/build/main.js not supported.
main.js is treated as an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which declares all .js files in that package scope as ES modules.
Instead rename main.js to end in .cjs, change the requiring code to use dynamic import() which is available in all CommonJS modules, or change "type": "module" to "type": "commonjs" in .../.nvm/versions/node/v16.14.2/lib/node_modules/serve/package.json to treat all .js files as CommonJS (using .mjs for all ES modules instead).

    at Object.<anonymous> (.../.nvm/versions/node/v16.14.2/lib/node_modules/pm2/lib/ProcessContainerFork.js:33:23) {
  code: 'ERR_REQUIRE_ESM'
}
```

scripts/install/functions.sh
```diff
@@ -85,43 +85,44 @@ function create_shell_app {
...
     fi
 }
 
 function add_cdc {
   if [ "$ADD_CDC" = true ] ; then
-        ng add --skip-confirmation @spartacus/cdc@${SPARTACUS_VERSION} --interactive false
+        ng add @spartacus/cdc@${SPARTACUS_VERSION} --skip-confirmation --no-interactive
```
plabadie, line 95: Smoke test comment 2:  When running the CSR app directly via `ng serve` it works.  However there are many deprecation warnings like this one:
```
Warning: '/.../apps/csr/node_modules/@spartacus/qualtrics/_index.scss' imports '~bootstrap/scss/_mixins' with a tilde. Usage of '~' in imports is deprecated.
```

feature-libs/cart/_index.scss
```diff
@@ -1,12 +1,12 @@
...
+@import '@spartacus/styles/scss/core';
+@import '@spartacus/styles/scss/cxbase/blocks/modal';
+@import '@spartacus/styles/scss/misc/table';
 
-@import '~bootstrap/scss/functions';
-@import '~bootstrap/scss/variables';
-@import '~bootstrap/scss/_mixins';
+@import 'bootstrap/scss/functions';
+@import 'bootstrap/scss/variables';
+@import 'bootstrap/scss/_mixins';
```
plabadie, line 9: It seems like `~`  says to import from the `node_modules`  folder.
Can you think of a way that removing the `-` can change the import behaviour and break a custom app?

feature-libs/organization/administration/components/shared/item-active.directive.ts
```diff
@@ -15,7 +15,9 @@ import { BaseItem } from './organization.model';
 @Directive({
   selector: '[cxOrgItemActive]',
 })
-export class ItemActiveDirective<T = BaseItem> implements OnInit, OnDestroy {
+export class ItemActiveDirective<T extends BaseItem = BaseItem>
```
plabadie, line 18: Adding a type to the generic type looks like a potential (compile time) breaking change.

feature-libs/organization/administration/components/shared/sub-list/assign-cell.component.ts
```diff
@@ -30,7 +31,7 @@ import { SubListService } from './sub-list.service';
   `,
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
-export class AssignCellComponent<T> extends CellComponent {
+export class AssignCellComponent<T extends BaseItem> extends CellComponent {
```
plabadie, line 34: Adding a type to the generic type looks like a potential (compile time) breaking change.

Same for all the other type changes in this PR

feature-libs/cart/_index.scss
WeizhengSap: https://github.com/webpack-contrib/sass-loader#resolving-import-at-rules
https://github.com/angular/components/commit/f2ff9e31425f0e395e6926bcaf48f876688000d8

feature-libs/organization/administration/components/shared/item-active.directive.ts
WeizhengSap: We are not adding a type, we just make the generic type more accurate. The new Angular seems more strict for type checking.

projects/core/src/cms/store/reducers/components.reducer.ts
```diff
@@ -14,7 +15,7 @@ export const initialState: ComponentsContext = {
   pageContext: {},
 };
 
-function componentExistsReducer<T>(
+function componentExistsReducer<T extends CmsComponent>(
```
plabadie, line 18: potential (compile time) breaking change.

projects/core/src/state/utils/get-state-slice.ts
```diff
@@ -30,7 +30,7 @@ export function createShellObject<T, E>(
   }
 
   const shell = key.split(OBJECT_SEPARATOR).reduceRight((acc, previous) => {
-    return { [previous]: acc } as unknown as T;
+    return { [previous]: acc } as unknown as NonNullable<T>;
```
plabadie, line 33: Potential (compile time) breaking change.

projects/core/src/util/create-from.ts
```diff
@@ -12,6 +12,6 @@ import { Type } from '@angular/core';
  * @param type reference to the class
  * @param data object with properties to be copied to the class
  */
-export function createFrom<T>(type: Type<T>, data: T): T {
+export function createFrom<T extends object>(type: Type<T>, data: T): T {
```
plabadie, line 15: Potential (compile time) breaking change.

projects/storefrontlib/cms-structure/page/component/services/component-data.provider.ts
```diff
@@ -32,7 +32,7 @@ export class ComponentDataProvider {
    * merged into the component data. The static data is complemented and
    * overridden with data retrieved from the cms service.
    */
-  get<T>(uid: string, type?: string): Observable<T> {
+  get<T extends CmsComponent>(uid: string, type?: string): Observable<T> {
```
plabadie, line 35: Potential (compile time) breaking change.

package.json
```diff
@@ -153,7 +153,7 @@
     "@ngrx/store-devtools": "^14.3.0",
```
plabadie, line 133: Overall didn't we want to upgrade the minimum angular version to 14.2?  Is there a point to leave it at 14.1.3 ?

scripts/install/functions.sh
plabadie: This error still happens.  Is it caused by changes from this PR?

scripts/install/functions.sh
plabadie: These warnings are now gone.

tools/config/manage-dependencies.ts
```diff
@@ -291,11 +292,37 @@ function filterLocalRelativeImports(
...
+        }
+        if (imp.importPath.startsWith('@')) {
+          return true;
+        }
+        // whether imports can be resolved from relative path
+        let folder = path.dirname([...imp.files][0]);
+        let file;
+        if (imp.importPath.includes('/')) {
+          const parts = imp.importPath.split('/');
+          file = parts.pop();
```
plabadie: Just to leverage the path api (and readability), could the file name perhaps be retrieved via `path.basename(imp.importPath)`

tools/config/manage-dependencies.ts
```diff
@@ -291,11 +292,37 @@ function filterLocalRelativeImports(
...
+          folder = folder + '/' + parts.join('/');
```
plabadie: Instead of `parts.join('/')`, the foler path could perhaps be retrieved with `path.dirname(imp.importPath)`

tools/config/manage-dependencies.ts
```diff
@@ -291,11 +292,37 @@ function filterLocalRelativeImports(
...
+        if (fs.existsSync(folder)) {
+          const fileList = fs.readdirSync(folder);
+          if (
+            fileList.includes(file + '.scss') ||
+            fileList.includes('_' + file + '.scss')
+          ) {
+            return false;
+          }
+        }
+        return true;
```
plabadie: Instead of creating and searching the file list of the folder we could also directly test the paths of the files that we want to know if they exist:
```
if ( fs.existSync(`${folder}/${file}.scss`)  || 
     fs.existSync(`${folder}/_${file}.scss`) ) 
{
  return false;
}
```

tools/config/manage-dependencies.ts
```diff
@@ -291,11 +292,37 @@ function filterLocalRelativeImports(
         return acc;
       }, {} as LibraryWithDependencies['tsImports']);
     lib.scssImports = Object.values(lib.scssImports)
-      .filter(
-        (imp) =>
-          imp.importPath.startsWith('node_modules/') ||
-          imp.importPath.startsWith('~')
-      )
+      .filter((imp) => {
```
plabadie, line 295: What script can I call to test this code?

scripts/install/functions.sh
plabadie: In the en this is a know issue and it is unrelated to the changes in the PR.
https://jira.tools.sap/browse/CXSPA-788

plabadie (dismissed):

package.json
plabadie: We agreed to do this in a follow up ticket if we go forward with this.

scripts/install/functions.sh
```diff
@@ -80,48 +80,49 @@ function update_projects_versions {
...
 function create_shell_app {
-    ( cd ${INSTALLATION_DIR} && ng new ${1} --style=scss --routing=false)
+    ( cd ${INSTALLATION_DIR} && ng new ${1} --style scss --no-routing)
 }
 
 function add_b2b {
     if [ "${ADD_B2B_LIBS}" = true ] ; then
-        ng add --skip-confirmation @spartacus/organization@${SPARTACUS_VERSION} --interactive false
-        ng add --skip-confirmation @spartacus/checkout@${SPARTACUS_VERSION} --interactive false --features="Checkout-B2B" --features="Checkout-Scheduled-Replenishment"
+        ng add @spartacus/organization@${SPARTACUS_VERSION} --skip-confirmation --no-interactive
```
plabadie, line 88: Some ng add command modules were removed this suffix `@${SPARTACUS_VERSION}` and some others no.  Is that intentional?

tools/config/manage-dependencies.ts
```diff
@@ -22,6 +22,7 @@ import chalk from 'chalk';
 import { execSync } from 'child_process';
 import fs, { readFileSync } from 'fs';
 import glob from 'glob';
+import * as path from 'path';
```
plabadie, line 25: This comment is for another file:

Need to bump the angular version in `scripts/install/config.default.sh`

plabadie (dismissed):

plabadie (approved):

PR #16176, opened 2022-09-19
refactor: change modal handling in add-to-cart-dialog

Matejk00:
Closes: CXSPA-1221


feature-libs/cart/base/components/added-to-cart-dialog/added-to-cart-dialog.component.html
```diff
@@ -1,124 +1,129 @@
-<div #dialog>
```
psmul, line 1: Let's include information about removing one layer of wrappers from this component in the migration docs. You can follow the pattern and provide `#### Template changes` as a paragraph to mention those changes.

psmul (changes requested):

PR #16180, opened 2022-09-21
Add release files to release branch

RadhepS:
closes: https://jira.tools.sap/browse/CXSPA-1365


sonar-project.properties
```diff
@@ -1,17 +1,20 @@
-sonar.projectKey=sap_cloud-commerce-spartacus-storefront
-sonar.projectName=Cloud commerce spartacus storefront
+sonar.host.url=https://sonar.tools.sap
 
+sonar.projectKey=pwa-storefront-toolset
+sonar.projectName=pwa-storefront-toolset
+sonar.branch.name=release/5.0-next.2
```
bgambocjaviniar: related to where I'm pointing here. I don't think we should specify a specific branch name here as we would need to change this many times. Should we just run sonar on the 'default' branch, therefore not specifying a branch

projects/storefrontstyles/.npmignore
```diff
@@ -2,3 +2,6 @@ yarn.lock
 importer.js
 temp-scss/
 .release-it.*
+.pipeline/
+sub-folder/
+vault*
```
bgambocjaviniar: I'm assuming vault is where we store the secrets in hyperspace. This is only generated under the hood when running the pipeline right as I can't find that folder in the private repo

sonar-project.properties
bgambocjaviniar: What do you think? I believe develop to be the source of truth in the private repo

projects/storefrontstyles/.npmignore
RadhepS: Hmm not sure why I put it there, I think I saw files related to that path but I don't see anything after removing it from the npmignore. I'll remove it :)

bgambocjaviniar: now I have to ask about these 

```
staging_output_deprecated.json
stage-config.json
errorDetails.json
```

RadhepS: not sure, originally saw those files in the artifacts, thats why i included them. but i dont see them anymore... removed them

sonar-project.properties
RadhepS: Thing is it will override the develop sonar scan no? Like we will run the sonar scan nightly for develop but if we then run this 5.0 branch for release, then it will show the 5.0 branch issues on the develop branch (default) which can be confusing?

RadhepS: I think each release branch should have its own sonar scan branch too but maybe im thinking of it wrong.
When this PR gets merged into develop, I can set develop there (or leave it empty cause it'll default to that).

But I see I should rename the branch, from release/5.0-next.2 to release/5.0-next.x most likely. Think that makes more sense

projects/storefrontstyles/.npmignore
bgambocjaviniar: consensus that we agreed upon is to delete it from source, and publish it automatically during the sync with the proper branch value

bgambocjaviniar (approved):

PR #16187, opened 2022-09-22
Feature/cxspa 1213

mono424:


scripts/install/run.sh
```diff
@@ -46,17 +46,26 @@ fi
 
 CLONE_DIR="${BASE_DIR}/${CLONE_DIR}"
 INSTALLATION_DIR="${BASE_DIR}/${INSTALLATION_DIR}"
+SKIP_SANITY=false
```
WeizhengSap: maybe these new settings should be in `config.default.sh`.

scripts/install/functions.sh
```diff
@@ -26,13 +44,16 @@ function cmd_clean {
 
     delete_dir ${BASE_DIR}
     delete_dir storage
-
-    yarn cache clean
+    if [[ "${KEEP_YARN_CACHE}" = false ]]; then
```
WeizhengSap, line 47: first, not sure whether we want to keep yarn cache, because I think install script wants to install everything from scratch; second, using `yarn cache clean` to remove the shared cache files is better than force-removed some folder `~/.yarn-cache`.

scripts/install/functions.sh
```diff
@@ -398,15 +415,399 @@ function stop_apps {
...
+    if [ $EXIT_CODE -eq 0 ]; then
+        echo "✅ SSR is working."
+    else
+        echo "🚫 SSR is NOT working."
+    fi
+}
+
+function run_e2e {
+    if [[ "$SKIP_E2E" = true ]] ; then
+        echo "⏩️ B2B E2E skipped (Option: --skip-e2e)."
```
WeizhengSap: "B2B E2E skipped" -- it is not B2B

scripts/install/run.sh
mono424: yes much nicer! :)

scripts/install/functions.sh
mono424: good catch!

scripts/install/run.sh
```diff
@@ -46,17 +46,26 @@ fi
...
+CHECK_AFTER_START=false
+CHECK_B2B_AFTER_START=false
+FORCE_E2E=false
+SKIP_E2E=false
+MAX_PARALLEL="6"
```
WeizhengSap: please move them to default.confg.sh

WeizhengSap (approved):

PR #16190, opened 2022-09-23
fix: add XSS security for icon-loader

FollowTheFlo:
same PR as https://github.com/SAP/spartacus/pull/16182 , now targeting 5.0.x
Url sanitizer has been added for required url used by svg and link icon types. Html symbol is now converted into text for text icon type.


giancorderoortiz (approved):
Same code changes as https://github.com/SAP/spartacus/pull/16182/files, which were approved by @RadhepS.

PR #16212, opened 2022-09-29
feat: automate the process of bumping package versions for publishing

RadhepS:
closes: https://jira.tools.sap/browse/CXSPA-853

**Basic test cases**

Test 1:

1. Set a value for PUBLISHING_VERSION in spartacus/tools/config/const.ts
2. Run yarn config:update -> it should update the publishing version as well as dependency versions to the value you set

Test 2:

1. Keep the value empty for PUBLISHING_VERSION in spartacus/tools/config/const.ts
2. Run yarn config:update -> it should not update the publishing version or dependency versions

Test 3:

1. Update the publishing version for @spartacus/core in its package.json (update line 3 to any different number)
2. Run yarn config:update -> it should update @spartacus/core in other package.jsons such as checkout (this makes sure the previous feature still works)

### DIFF
--- a/tools/config/manage-dependencies.ts
+++ b/tools/config/manage-dependencies.ts
@@ -222,6 +222,7 @@ export function manageDependencies(
       return acc;
     }, {});
 
+    // If publishing version is defined, update the publishing versions of packages
     if (PUBLISHING_VERSION) {
       updatePublishingVersions(libraries, PUBLISHING_VERSION);
     }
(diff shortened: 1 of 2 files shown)


tools/config/manage-dependencies.ts
```diff
@@ -221,6 +222,11 @@ export function manageDependencies(
       return acc;
     }, {});
 
+    // If publishing version is defined, update the publishing versions of packages
+    if (PUBLISHING_VERSION) {
```
FollowTheFlo: extra indent that shouldn't be here?

RadhepS: fixed :)

FollowTheFlo (approved):
looks good

PR #16213, opened 2022-09-29
add private npm capability for npm_install

FollowTheFlo:
Modified functions.sh with:
- create npmrc file before the install and put it in apps/csr, apps/ssr (+pwa) folders
- then remove npmrc file and  token value in config.sh
only impact install_npm


scripts/install/config.default.sh
```diff
@@ -71,3 +71,9 @@ ADD_EPD_VISUALIZATION=false
 
 # The base URL (origin) of the SAP EPD Fiori launchpad
 EPD_VISUALIZATION_BASE_URL=
+
+#NPM connection info
+NPM_TOKEN=
+NPM_USER=
+NPM_URL=
```
FollowTheFlo, line 78: I removed default URL https://common.repositories.cloud.sap/artifactory/api/npm/deploy-releases-hyperspace-npm/
for security purpose, like checkmarx vulnerability.

scripts/install/config.default.sh
```diff
@@ -71,3 +71,10 @@ ADD_EPD_VISUALIZATION=false
 
 # The base URL (origin) of the SAP EPD Fiori launchpad
 EPD_VISUALIZATION_BASE_URL=
+
+#NPM connection info
+#NPM_URL must start by 'https://' and end with '/' char
+NPM_TOKEN=
+NPM_USER=
+NPM_URL=
```
FollowTheFlo, line 79: no common.repositories url by default for security purpose like checkmarx scan

FollowTheFlo, line 78: no common.repositories url by default to avoid checkmarx warning. so user is forced to add it in config.sh

scripts/install/functions.sh
```diff
@@ -144,11 +154,19 @@ function add_spartacus_ssr {
     add_cdc
     add_epd_visualization
     add_product_configurator
+    remove_npmrc
+    if [ ! -z "$IS_NPM_INSTALL" ] ; then
+        remove_npmrc
```
RadhepS: remove npmrc appears twice. we can remove the if condition and just keep the "remove_npmrc" outside because that function does a check to see if the .npmrc file exists. we shouldn't get any errors

scripts/install/functions.sh
```diff
@@ -159,30 +177,35 @@ function add_spartacus_ssr_pwa {
     add_cdc
     add_epd_visualization
     add_product_configurator
+    if [ ! -z "$IS_NPM_INSTALL" ] ; then
```
RadhepS: i think we can remove the if statement like i mentioned on the other comment

scripts/install/config.default.sh
```diff
@@ -71,3 +71,10 @@ ADD_EPD_VISUALIZATION=false
 
 # The base URL (origin) of the SAP EPD Fiori launchpad
 EPD_VISUALIZATION_BASE_URL=
+
+#NPM connection info
+#NPM_URL must start by 'https://' and end with '/' char
+NPM_TOKEN=
+NPM_EMAIL=
```
RadhepS: we can remove email, its not needed

scripts/install/functions.sh
```diff
@@ -382,3 +413,30 @@ function cmd_help {
     echo " stop"
     echo " help"
 }
+
+function create_npmrc {   
+    local NPMRC_CONTENT="always-auth=${NPM_ALWAYS_AUTH}\n@spartacus:registry=${NPM_URL}\n$(echo ${NPM_URL} | sed 's/https://g'):_auth=${NPM_TOKEN}\nemail=${NPM_EMAIL}\n"
+    if [ -z "$NPM_TOKEN" ] ; then
+        echo "NPM_TOKEN is empty"
+    fi
+    if [ -z "$NPM_EMAIL" ] ; then
```
RadhepS: remove email, not needed for npmrc

scripts/install/functions.sh
```diff
@@ -129,11 +131,19 @@ function add_spartacus_csr {
     add_cdc
     add_epd_visualization
     add_product_configurator
+    if [ ! -z "$IS_NPM_INSTALL" ] ; then
```
RadhepS: think we can remove if statement (check other comment)

scripts/install/functions.sh
```diff
@@ -382,3 +413,30 @@ function cmd_help {
...
+}
+
+function remove_npmrc {   
+    if [ -f "./.npmrc" ]; then
+        echo 'removing .npmrc file'
+        rm -rf .npmrc
+    fi
+}
+
+function remove_npm_token {
```
RadhepS, line 427: maybe some people might the token in the config.default.sh? should we check and remove it from there too just in case?

RadhepS (changes requested):

scripts/install/functions.sh
FollowTheFlo: good idea!

scripts/install/functions.sh
FollowTheFlo: will be done

scripts/install/functions.sh
FollowTheFlo: yes it should not be there twice, forgot to remove one. I'll keep the one outside the if condition and  test it first.

scripts/install/config.default.sh
FollowTheFlo: nice

RadhepS (approved):

PR #16242, opened 2022-10-04
replace sanitizer by html escape

FollowTheFlo:
- JsonLdScriptFactory: sanitizer replaced by escapeHtml
- breaking change added on 5_0.md

test done fro StructuredDataFactory: on pdp, plp, grab generated json-ld from DOM, verified validity on https://validator.schema.org/
test for jsonLdDirective: insert html tags, rab generated json-ld from DOM, make sure html get escaped, verified validity on https://validator.schema.org/


projects/storefrontlib/cms-structure/seo/structured-data/json-ld-script.factory.ts
```diff
@@ -78,16 +75,15 @@ export class JsonLdScriptFactory {
...
    * into the json-ld script.
    */
-  sanitize(schema: {}): string {
-    return JSON.stringify(schema, (_key, value) =>
-      typeof value === 'string'
-        ? this.sanitizer.sanitize(SecurityContext.HTML, value)
-        : value
-    );
+  escapeHtml(schema: {}): string {
+    const renderer: Renderer2 = this.rendererFactory.createRenderer(null, null);
```
FollowTheFlo: not sure creating a renderer to create a detached div element is the best solution in terms of memory and performance. Could be an issue when you use with high frequency like https://github.com/angular-translate/angular-translate/issues/1843

other solutions to improve i could think: 
- create htmlElement from this.winRef.document.createElement
- create an html encoder with regex (challenging to cover all use cases)
- destroy() the local renderer if not already done after return

projects/storefrontlib/cms-structure/seo/structured-data/json-ld.directive.ts
```diff
@@ -30,22 +30,19 @@ export class JsonLdDirective {
   }
 
   constructor(
-    private renderer: Renderer2,
+    protected renderer: Renderer2,
     protected jsonLdScriptFactory: JsonLdScriptFactory,
-    private element: ElementRef
+    protected element: ElementRef
```
znikola, line 35: Nice! 👍

projects/storefrontlib/cms-structure/seo/structured-data/json-ld-script.factory.ts
znikola: You can either:
- create the renderer using the factory in the constructor, and "cache" it as a `protected` member. We do this in some places in Spartacus.
- pass the actual renderer to the method as a parameter. But this approach is a bit problematic, as it requires all the callers to provide the render instance. What if a service calls it? We would be back to square one. So I suggest to go with the option from above.

In regards to the performance/memory concern - we can check if the garbage collector cleans up the detached DOM elements. I can do that tomorrow

docs/migration/5_0.md
```diff
@@ -775,6 +775,10 @@ Due to i18next migration, certain set of keys have been migrated (from `_plural`
 
 - Text type icon symbols are now converted into text, meaning DOM does not render HTML symbols anymore.
 
+### json-ld
```
znikola: Just a reminder - we should add the schematics migrations for the breaking changes done in the previous PR, as well as the ones done in this PR.

docs/migration/5_0.md
```diff
@@ -775,6 +775,10 @@ Due to i18next migration, certain set of keys have been migrated (from `_plural`
...
+
+- For security improvement, html sanitizer has been removed from `JsonLdScriptFactory`, now replaced by html characters escape. It impacts `JsonLdDirective` and `StructuredDataFactory`.
+
```
znikola, line 798: Did you check if the JSON LD is rendered in SSR mode?

znikola (changes requested):

projects/storefrontlib/cms-structure/seo/structured-data/json-ld-script.factory.ts
FollowTheFlo: good idea! i saw it done on LaunchRenderStrategy.

docs/migration/5_0.md
FollowTheFlo: it is fine on SSR, json-ld correctly rendered, tested on ssr localhost.

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/json-ld.script.factory.migration.ts
```diff
@@ -0,0 +1,17 @@
+import {
+  JSON_LD_SCRIPT_FACTORY, SANITIZE_METHOD,
+  TODO_SPARTACUS
+} from '../../../../shared/constants';
+import { SPARTACUS_STOREFRONTLIB } from '../../../../shared/libs-constants';
+import { MethodPropertyDeprecation } from '../../../../shared/utils/file-utils';
+
+// fil path
```
znikola: oops :)

znikola (changes requested):

projects/schematics/src/migrations/5_0/methods-and-properties-deprecations/data/json-ld.script.factory.migration.ts
FollowTheFlo: added file path, copyright header, prettier applied

docs/migration/5_0.md
FollowTheFlo: schematics for previous PR will be handled in [CXSPA-1491](https://jira.tools.sap/browse/CXSPA-1491)
Add schematics for icon-loader service change

projects/storefrontlib/cms-structure/seo/structured-data/json-ld-script.factory.ts
```diff
@@ -12,27 +12,28 @@ import {
...
 } from '@angular/core';
-import { DomSanitizer } from '@angular/platform-browser';
 import { WindowRef } from '@spartacus/core';
 import { SeoConfig } from '../config';
 
 @Injectable({
   providedIn: 'root',
 })
 export class JsonLdScriptFactory {
+  protected renderer: Renderer2;
```
znikola: Nice! :)

Just to be more consistent with other classes, let's init this property inline:

```ts
  protected renderer: Renderer2 = this.rendererFactory.createRenderer(
    null,
    null
  );

  constructor(
    @Inject(PLATFORM_ID) protected platformId: string,
    protected winRef: WindowRef,
    protected rendererFactory: RendererFactory2,
    protected config: SeoConfig
  ) {}
```

effectively, there won't be any difference when the code is compiled down to .js.

znikola (changes requested):

projects/storefrontlib/cms-structure/seo/structured-data/json-ld-script.factory.ts
FollowTheFlo: good idea, thx

znikola (approved):

PR #16246, opened 2022-10-05
fix: Change provideConfig to provideDefaultConfig in modules

Zeyber:
Closes: https://jira.tools.sap/browse/CXSPA-289


feature-libs/.eslintrc.json
```diff
@@ -0,0 +1,16 @@
+{
+  "overrides": [
+    {
+      "files": ["*.module.ts"],
+      "rules": {
+        "no-restricted-syntax": [
+          "error",
+          {
+            "selector": "CallExpression[callee.name=provideConfig]",
+            "message": "[Spartacus] provideConfig() is reserved for library consumers. For providing default configs inside Spartacus libraries please use provideDefaultConfig() instead."
```
znikola:
```suggestion
            "message": "[Spartacus] provideConfig() is intended for library consumers. To allow for better extensibility, please provide default configs inside Spartacus libraries using provideDefaultConfig()"
```

.eslintrc.json
```diff
@@ -93,4 +93,4 @@
       }
     }
   ]
-}
+}
```
znikola, line 113: Is it possible to specify the lint rule in the root `.eslintrc.json`, since all the other `.eslintrc.json` extend from it?

feature-libs/.eslintrc.json
znikola: 👆 just a suggestion

znikola (changes requested):

PR #16268, opened 2022-10-07
design update and localizations

kimhw0630:


justinlee01 (changes requested):
We're no longer using Fragments; see this branch: https://github.com/SAP/spartacus/tree/task/asm-360-data-layer

PR #16274, opened 2022-10-10
Feature/CXSPA-1350: Implement user rights assignment functionality for unitorderviewergroup

T0bas-2:
Ticket: https://jira.tools.sap/browse/CXSPA-1350


feature-libs/organization/administration/assets/translations/en/units.i18n.ts
```diff
@@ -167,7 +167,7 @@ export const orgUnitUsers = {
 };
 
 export const orgUnitUserRoles = {
-  title: 'User roles',
+  title: 'User roles and rights',
```
I539321, line 170: For a constant named `orgUnitUserRoles` I'd expect its title to be `User roles` as it is and not `User roles and rights`, perhaps we can create another constant named `orgUnitUserRights` or so? Or maybe create another constant `orgUnitUserRolesAndRights`?

MarcinGanczorz:
Changes are looking good to me. The only thing left is to fix the e2e tests

MarcinGanczorz (approved):
LGTM, remaining e2e tests will be fixed in separate PR

PR #16275, opened 2022-10-10
feature/cxspa-1474 add email address to the unit-level order history list

I565452:
this is how it's look like with email
image

In addition, I changed the details page to display the email field instead of uid.

### DIFF
--- a/feature-libs/organization/unit-order/styles/_unit-order.scss
+++ b/feature-libs/organization/unit-order/styles/_unit-order.scss
@@ -23,7 +23,7 @@
 
     td {
       width: 16.6%;
-      padding: 1.625rem 0;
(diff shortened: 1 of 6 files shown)


feature-libs/organization/unit-order/styles/_unit-order.scss
```diff
@@ -47,6 +47,11 @@
           padding-bottom: 1.25rem;
         }
       }
+
+      span {
+        color: #a1aabf;
```
MarcinGanczorz, line 52: Maybe we could use a variable color instead of hardcoded one.

MarcinGanczorz (changes requested):

Issue #15566, opened 2022-04-07
Adapt configuration related additions in cart to refactored tables


steinsebastian:
The cart view has been refactored to use real table-tags. Thus our additions don't fit (e.g. bundle items).
<img width="803" alt="Screenshot 2022-04-06 at 16 43 25" src="https://user-images.githubusercontent.com/54309425/162176587-87f2b451-2f91-42aa-b5b2-2055c0cd1fe3.png">

developpeurweb:
@Larisa-Staroverova pls be aware of this ticket for grouping notifications, GH-15483 we are still waiting for a decision on mockups.

JayEmKay-8:
For reference: this is what it used to look like before the table refactoring. Much easier to check and to grasp at a glance:
![Spartacus_CPQCameraBundle_Cart-mit-Items-und-issue](https://user-images.githubusercontent.com/54537803/179513106-6e9d60ad-983b-4037-b811-19cc5c09006d.png)

PR #16285, opened 2022-10-12
v6 POC #15566 cxCartItemContext, fixed styling

Platonn:
Same as [v3 POC](https://github.com/SAP/spartacus/pull/16272/) plus:
- issue-notification component moved inside the cart item row component, into the "Description" column 
- added a gradient fade out of the red color on the right

related to #15566 

Result:
> image


Platonn:
This approach was rejected by visual designers as not consistent with other views. Moreover it doesn't make sense, when we'll change the design of alerts, to have a solid border around. See the upcoming future design (perhaps to be released only in 6.0?):

image

PR #16287, opened 2022-10-12
fix: Change provideConfig to provideDefaultConfig in modules and add lint

### LINKED ISSUE #289: Improve the CmsPageGuard logic and its unit tests
Investigation required.

This ticket blocks GH-290

We have a lookup pattern in CmsPageGuard that looks roughly like this:

```
Is data present in store?
  if yes, great!  Return it.
  if no, try to load the data in the store and wait for the result 
    if now the data is successfully loaded, great! Return it.
    if the data can't be loaded, react accordingly.
```

Right now, the issues we suspect our code to have are:
- It is more complex than it should.
- The code is hard to understand, debug and evolve.
- It's difficult to write unit tests for this piece of code.
- We are not sure if we can get in intfinite loop situations ( Looping on: "if we don't find it, try to load it" )
- We are not sure if we should deal with concurrency and the 'loading' state of the data we are looking up.
- We are not sure which part or the store calls are synchronous and which are asynchronous.
- The logic to handle retries should probably not be inside the route guard, Addis unnecessary complexity.

The goal of ticket is to improve the code according to the above criteria.  We need simple, reliable and testable code in the route guard logic.

Below here is the code that needs improvements.  The code is directly in the route guard at the moment of writing this description, but the code will soon be moved to a facade service.  The problem will remain and the logic in the facade services should be improved/fixed.  To find the code just look which facade service function is called from the guard's canActivate function.

The hasPage function in the CmsPageGuard
```
  hasPage(): Observable<boolean> {
    let tryTimes = 0;

    return this.routingService.routerState$.pipe(
      map(routerState => routerState.state.context),
      take(1),
      mergeMap(pageContext =>
        this.store.pipe(
          select(fromStore.getPageEntities),
          map((entities: { [key: string]: Page }) => {
            let key = pageContext.id + '_' + pageContext.type;
            let found = !!entities[key];
            if (!found) {
              const defaultPageIds = this.defaultPageService.getDefaultPageIdsBytype(
                pageContext.type
              );
              if (defaultPageIds) {
                for (let i = 0, len = defaultPageIds.length; i < len; i++) {
                  key = defaultPageIds[i] + '_' + pageContext.type;
                  found =
                    entities[key] &&
                    entities[key].seen.indexOf(pageContext.id) > -1;
                  if (found) {
                    break;
                  }
                }
              }
            }
            // found page directly from store
            if (found && tryTimes === 0) {
              this.store.dispatch(new fromStore.UpdateLatestPageKey(key));
            }
            return found;
          }),
          tap(found => {
            // if not found, load this cms page
            if (!found) {
              tryTimes = tryTimes + 1;
              this.store.dispatch(new fromStore.LoadPageData(pageContext));
            }
          }),
          filter(found => found || tryTimes === 3),
          take(1)
        )
      )
    );
  }
```

Zeyber:
Closes: #289


projects/core/src/config/config.module.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @spartacus-eslint/use-default-provide-config */
```
znikola: Not sure how the script (if any) works for adding the license. Maybe it checks the very first line in a file and adds the license again if it doesn't find a particular string (this is me just guessing).
For that reason, maybe it's better to specify the rule inline?

znikola (changes requested):

znikola (dismissed):

znikola (approved):

PR #16294, opened 2022-10-12
chore: add capability to install spartacus from private packages

FollowTheFlo:
Spartacus can now be installed from sap private npm packages when using install_npm command. Config vars have been added: NPM_TOKEN, NPM_URL,NPM_ALWAYS_AUTH When the install starts, .npmrc file gets created then removed once finished. For security purpose, NPM_TOKEN var gets cleared from config.sh and config.default.sh when install is finished.

BREAKING CHANGE

New config vars have been introduced: NPM_TOKEN, NPM_URL are required to be filled by the user in config.sh file. It is needed only when using install_npm command.


giancorderoortiz (approved):
Same changes as in https://github.com/SAP/spartacus/pull/16213/files.

PR #16299, opened 2022-10-13
doc: Add note about false-positives during migration

znikola:
closes CXSPA-1508

### DIFF
--- a/docs/migration/5_0.md
+++ b/docs/migration/5_0.md
@@ -1,7 +1,7 @@
 # Technical Changes in Spartacus 5.0
 
 Spartacus migration schematics tries to find in your codebase usages of Spartacus' public API (e.g. functions/methods/classes) which were changed (in a breaking way) or removed in version 5.0. The migration schematics will insert a code comment with a hint "TODO:Spartacus" above such lines of code, in order to assist you with the migration. 
-Note: Please beware that if by coincidence you have in your codebase a usage of a method/function/class that is named the same as the changes or removed Spartacus' public APIs, there might be a chance that it may be incorrectly classified as Spartacus' API and the migration schematics might insert a comment there as well (even though it shouldn't). In that case, please ignore and remove such a comment.
+Note: Please beware that if by coincidence you have in your codebase a usage of a method/function/class that is named the same as the changed or removed Spartacus' public APIs, there might be a chance that it may be incorrectly classified as Spartacus' API and the migration schematics might insert a comment there as well (even though it shouldn't). In that case, please ignore and remove such a comment.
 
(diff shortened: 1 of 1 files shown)


docs/migration/5_0.md
```diff
@@ -1,5 +1,8 @@
 # Technical Changes in Spartacus 5.0
 
+This page contains all the technical changes introduced in 5.0 release.
+Please note the automated migration process might yield some false-positive results. For example, after running the Spartacus upgrade schematics, you might see a comment in your code, where Spartacus is informing you about a breaking change, even though there is none. This is a false-positive, which likely happened because Spartacus uses the same API method / property name. In cases like these, it is recommended to just ignore the comment and remove it.
+
```
Platonn: terms like "False-positives" are more familiar to us than customers. It would be good if the narration is more customer-oriented. What do you think about this:
```suggestion
Spartacus migration schematics tries to find in your codebase usages of Spartacus' public API (e.g. function/method/class) that was changed (in a breaking way) or removed in 5.0. Then the migration schematics will insert a code comment with a hint "TODO" over such a line of code, to assist you with the migration. 

Note: Please beware that if by accident you have in your codebase an usage of a method/function/class that is named the same as some of Spartacus' public API, there is a chance that it may be incorrectly classified as Spartacus' API and the migration schematics might insert a comment there as well (even though it shouldn't). In that case, please ignore and remove such a comment.
```

docs/migration/5_0.md
```diff
@@ -1,5 +1,8 @@
 # Technical Changes in Spartacus 5.0
 
+Spartacus migration schematics tries to find in your codebase usages of Spartacus' public API (e.g. functions/methods/classes) which were changed (in a breaking way) or removed in version 5.0. The migration schematics will insert a code comment with a hint "TODO:Spartacus" above such lines of code, in order to assist you with the migration. 
+Note: Please beware that if by coincidence you have in your codebase a usage of a method/function/class that is named the same as the changes or removed Spartacus' public APIs, there might be a chance that it may be incorrectly classified as Spartacus' API and the migration schematics might insert a comment there as well (even though it shouldn't). In that case, please ignore and remove such a comment.
```
Platonn:
```suggestion
Note: Please beware that if by coincidence you have in your codebase a usage of a method/function/class that is named the same as the changed or removed Spartacus' public APIs, there might be a chance that it may be incorrectly classified as Spartacus' API and the migration schematics might insert a comment there as well (even though it shouldn't). In that case, please ignore and remove such a comment.
```

Platonn (approved):

PR #16309, opened 2022-10-14
Task/asm 360 rebase develop

justinlee01:


feature-libs/asm/root/config/default-asm-config.ts
```diff
@@ -49,5 +50,52 @@ export const defaultAsmConfig: AsmConfig = {
...
+        {
+          i18nNameKey: 'asm.customer360.overviewTab',
+          components: [
+            {
+              component: 'AsmCustomer360OverviewComponent',
+            },
+          ],
+        },
+        {
+          i18nNameKey: 'asm.customer360.profileTab',
```
justinlee01: Should use `CmsConfig`

justinlee01:
Make Asm 360 its own package / lazy loaded

justinlee01:
Use queries for 360 fetch

justinlee01:
For 360 - look at `ProductScope`

feature-libs/asm/root/config/default-asm-config.ts
justinlee01: Having reviewed `cms-config.ts`, there is not much to re-use - the `ComponentWrapperDirective` API is rather sparing.

PR #16310, opened 2022-10-14
feat: CXSPA-1502 Migration improvements for NgBootstrap removal

pawelfras:
closes CXSPA-1502


projects/schematics/src/migrations/5_0/constructor-deprecations/data/generated-constructor.migration.ts
```diff
@@ -3800,40 +3800,6 @@ export const GENERATED_CONSTRUCTOR_MIGRATIONS: ConstructorDeprecation[] =
         },
       ],
     },
-    {
```
plabadie: Updating the generated files is not going to hold. There is a good change they get re-generated.

plabadie (approved):

PR #16321, opened 2022-10-18
chore: Remove changelog script and everything related to it

znikola:
closes CXSPA-892


package.json
```diff
@@ -198,8 +199,7 @@
     "ts-morph": "^9.1.0",
     "ts-node": "^10.6.0",
     "typescript": "^4.7.2",
-    "webpack-cli": "^4.10.0",
     "webpack": "^5.73.0",
-    "@babel/runtime": "^7.18.9"
```
Platonn, line 203: Is the removal of `@babel/runtime` really related to this PR?
Please double check with @WeizhengSap , who added it when upgrading to Angular 14 https://github.com/SAP/spartacus/blame/6c9b4ae3c40837158ba213dccb6caf793a7ad2d0/package.json#L202

Zeyber: You'll find its not removed. But rather just moving itself to another position in the `package.json`. ;)

znikola: Exactly - not removed, just moved :)

Platonn: I'm sorry. My bad!

package.json
```diff
@@ -129,12 +129,12 @@
     "@angular/cli": "^14.2.3",
```
bgambocjaviniar, line 128: the package script `generate:changelog` from this file needs to be removed

package.json
```diff
@@ -129,12 +129,12 @@
...
     "@angular/compiler-cli": "^14.2.3",
     "@angular/language-service": "^14.2.3",
+    "@babel/runtime": "^7.18.9",
     "@ngrx/store-devtools": "^14.3.0",
```
bgambocjaviniar, line 132: not related to this file, but remove the references of changelog adaptation changes from `creating-lib.md`. Starting from line 270

package.json
```diff
@@ -129,12 +129,12 @@
...
     "@nguniversal/builders": "^14.2.0",
     "@sapui5/ts-types-esm": "1.98.0",
```
bgambocjaviniar, line 134: not related to this file, but `scripts/package.json` has be a dangling/dead config for a while. File for tsconfig existed in the past, which can be found here https://github.com/SAP/spartacus/pull/4485/files, but was no longer used. Especially since we deleted the .spec file, we do not need it anyhow

bgambocjaviniar: Pretty sure it's safe to remove as 'most' dependencies related to this file can be found in the root package.

znikola: Was thinking the same, but wasn't _bold_ enough. Done!

bgambocjaviniar (approved):

PR #16327, opened 2022-10-19
Filtering options on Overview Page

Larisa-Staroverova:


Larisa-Staroverova:
After analysis we came to the conclusion that existing `cx-product-facet-navigation` component cannot be reused. A new one should be created for filtering options on the overview page.

PR #16328, opened 2022-10-19
CXSPA-1474: update the color for email field

I565452:
the color for email on the `unit-level order history` page was changed after agreement with UX designer


MarcinGanczorz (approved):
LGTM

PR #16339, opened 2022-10-19
disabling assertion of message on user registration 

KateChuen:


KateChuen:
Do not need this anymore. Issue is fixed in https://jira.tools.sap/browse/CXSPA-1557

PR #16342, opened 2022-10-19
cxec-14760: add new message and upload/download attachment

Parthlakhani:
Closes [CXEC-14760](https://jira.tools.sap/browse/CXEC-14760): Implement core logic to add new messages
Closes [CXEC-17408](https://jira.tools.sap/browse/CXEC-17408): Implement API for Upload and Download Attachment


feature-libs/customer-ticketing/components/customer-ticketing-messages/customer-ticketing-messages.component.ts
```diff
@@ -1,112 +1,110 @@
...
 
-  prepareMessageEvents(): Observable<Array<MessageEvent> | undefined> {
+  protected prepareMessageEvents(): Observable<
+    Array<MessageEvent> | undefined
+  > {
     return this.ticketDetails$.pipe(
       map((ticket) =>
-        ticket.ticketEvents?.map(
-          (event): MessageEvent => ({
+        ticket?.ticketEvents?.reverse().map(
```
khristian23: How about sorting the list by creation time,  because no matter the order in which the events arrive, you need to ensure that they are always presented in ascending order

projects/storefrontlib/shared/components/chat-messaging/messaging/messaging.model.ts
```diff
@@ -17,6 +20,7 @@ export interface MessagingConfigs {
   newMessagePlaceHolder?: string;
   enableFileUploadOption?: boolean;
   dateFormat?: string;
+  displayInput?: Observable<boolean>;
```
khristian23:
```suggestion
  displayAddMessageSection?: Observable<boolean>;
```
Input is too vague

feature-libs/customer-ticketing/components/customer-ticketing-messages/customer-ticketing-messages.component.ts
```diff
@@ -1,112 +1,110 @@
...
 
-  prepareMessagingConfigs(): MessagingConfigs {
+  protected prepareMessagingConfigs(): MessagingConfigs {
     return {
       attachmentRestrictions:
         this.customerTicketingConfig.customerTicketing?.attachmentRestrictions,
       charactersLimit:
         this.customerTicketingConfig.customerTicketing?.inputCharactersLimit,
       enableFileUploadOption: true,
+      displayInput: this.ticketDetails$.pipe(
```
khristian23:
```suggestion
      displayAddMessageSection: this.ticketDetails$.pipe(
```

feature-libs/customer-ticketing/components/customer-ticketing-messages/customer-ticketing-messages.component.ts
```diff
@@ -1,112 +1,110 @@
...
+        map((ticket) => ticket?.status?.id !== STATUS.CLOSED)
```
khristian23, line 96: According to this: "ticket" or its "status" could be undefined and therefore different to CLOSED, which as a consequence will evaluate displayAddMessageSection to TRUE

feature-libs/customer-ticketing/core/facade/customer-ticketing.service.ts
```diff
@@ -70,12 +70,13 @@ export class CustomerTicketingService implements CustomerTicketingFacade {
...
-                    TicketEventCreatedEvent
-                  )
-                )
+                tap(() => {
+                  if (ticketEvent.toStatus?.id)
+                    this.eventService.dispatch(
+                      { status: ticketEvent.toStatus?.id },
+                      TicketEventCreatedEvent
+                    );
+                })
```
khristian23: This seems that it only sends a call if a status is present.
Nevertheless, the OCC API supports adding just a message without changing the ticket status
Why are you adding now that validation?

feature-libs/customer-ticketing/occ/adapters/occ-customer-ticketing.adapter.ts
```diff
@@ -122,4 +125,80 @@ export class OccCustomerTicketingAdapter implements CustomerTicketingAdapter {
       },
     });
   }
+
+  uploadAttachment(
+    customerId: string,
+    ticketId: string,
+    eventCode: string,
+    file: File
+  ): Observable<unknown> {
```
khristian23, line 135: Upload Attachment endpoint returns a TicketEventAttachment entity.  It looks like you don't use it, but do you think it should be added? Or you are reloading the ticket details upon event creation?

https://spartacus-dev0.eastus.cloudapp.azure.com:9002/occ/v2/swagger-ui/index.html#/Tickets/createTicketEventAttachment

feature-libs/customer-ticketing/components/customer-ticketing-messages/customer-ticketing-messages.component.ts
```diff
@@ -1,112 +1,110 @@
...
+
+  downloadAttachment(event: {
+    messageCode: string;
+    attachmentId: string;
+    fileName: string;
+  }) {
+    this.subscription.add(
+      this.customerTicketingFacade
+        .downloadAttachment(event.messageCode, event.attachmentId)
+        .subscribe((data) => {
```
khristian23, line 61: I'm confused here, please clarify.
Isn't the DownloadAttachment returning a Blob? how that is transformed into the downloadURL?
The download URL must be built from the input params:
downloadURL = ...adapter. getDownloadAttachmentEndpoint(customerId, ticketId, EventCode, AttachmentId)

feature-libs/customer-ticketing/occ/adapters/occ-customer-ticketing.adapter.ts
```diff
@@ -122,4 +125,80 @@ export class OccCustomerTicketingAdapter implements CustomerTicketingAdapter {
...
+          eventCode,
+          attachmentId
+        ),
+        httpOptions
+      )
+      .pipe(
+        catchError((error) => throwError(normalizeHttpError(error))),
+        this.converter.pipeable(CUSTOMER_TICKETING_FILE_NORMALIZER)
+      );
+  }
```
khristian23, line 188: I don't get why you are explicitly doing the download, it should be the browser doing it.

khristian23 (changes requested):

feature-libs/customer-ticketing/components/customer-ticketing-messages/customer-ticketing-messages.component.ts
Parthlakhani: Yeah, but we only want to hide the addMessageSection when the status is closed.

feature-libs/customer-ticketing/core/facade/customer-ticketing.service.ts
Parthlakhani: TicketEventCreatedEvent triggers the getTicket API and reloads the data.
I am only triggering it when the status is changed, as in other cased when message is added or attachment is uploaded, we are taking care of this from the component level.

Reason: If we reload data here, it will not wait for the upload attachment. And, if we add this to uploadAttachmentCommand, it will not be triggered when there is no attachment.
So, it's added at the component level making sure that it only triggers once.

feature-libs/customer-ticketing/occ/adapters/occ-customer-ticketing.adapter.ts
Parthlakhani: yes, TicketEventAttachment is not needed as ticket details is reloaded upon event creation.

feature-libs/customer-ticketing/components/customer-ticketing-messages/customer-ticketing-messages.component.ts
Parthlakhani: We need to do this In order to download the blob in a desirable format, I tried this way and it opens the raw dat in unreadable format.

feature-libs/customer-ticketing/components/customer-ticketing-messages/customer-ticketing-messages.component.ts
Parthlakhani: It is already sorted from the backend, but it is in the reversed order. So, I am reversing it again.

khristian23: Please try adding the sorting in the adapter.

khristian23 (approved):

adolfobee (approved):

PR #16350, opened 2022-10-20
CXSPA-1463: add total result to unit-level orders history

MarcinGanczorz:
https://jira.tools.sap/browse/CXSPA-1463

### DIFF
--- a/projects/storefrontlib/shared/components/list-navigation/total/total.component.ts
+++ b/projects/storefrontlib/shared/components/list-navigation/total/total.component.ts
@@ -18,19 +18,15 @@ export class TotalComponent {
   get pagination(): PaginationModel {
     return this._pagination;
   }
-  @Input() set pagination(value: PaginationModel | undefined) {
-    if (value) {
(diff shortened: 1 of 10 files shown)


projects/storefrontlib/shared/components/list-navigation/total/total.component.ts
```diff
@@ -0,0 +1,54 @@
...
+  templateUrl: './total.component.html',
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class TotalComponent {
+  private _pagination: PaginationModel;
+
+  get pagination(): PaginationModel {
+    return this._pagination;
+  }
+  @Input() set pagination(value: PaginationModel | undefined) {
```
I565452: I would name the `value` parameter with a name that explains its role or content
You might also want to use the syntax `elvis` here? :)  Up to you - I simply like it :)

I565452 (commented):
Review done, one small comment

I565452 (approved):

PR #16354, opened 2022-10-21
fix: log user out when closing account

pebarron15:
Closes: https://jira.tools.sap/browse/CXSPA-741

Deprecation cleanup: https://jira.tools.sap/browse/CXSPA-2056


Zeyber (changes requested):
Can we add checks that the user has correctly logged out in the e2e test too (`projects/storefrontapp-e2e-cypress/cypress/integration/regression/my-account/close-account.core-e2e-spec.ts`)?

feature-libs/user/profile/components/close-account/components/close-account-modal/close-account-modal.component.ts
```diff
@@ -80,7 +80,7 @@ export class CloseAccountModalComponent implements OnInit {
           GlobalMessageType.MSG_TYPE_CONFIRMATION
         );
       });
-    this.routingService.go({ cxRoute: 'home' });
+    this.routingService.go({ cxRoute: 'logout' });
```
Zeyber: This logs the user out but the user appears to stay on the close account page after the account is closed. Shouldn't we redirect back to the home or login after the account is closed? The e2e test should verify this behaviour.
image

Zeyber (changes requested):

feature-libs/user/profile/components/close-account/components/close-account-modal/close-account-modal.component.ts
pebarron15: @Zeyber I'm seeing the correct behavior the user is logged out and redirected to the homepage 
image

projects/storefrontapp-e2e-cypress/cypress/integration/regression/my-account/close-account.core-e2e-spec.ts
```diff
@@ -69,6 +69,8 @@ describe('My Account - Close Account', () => {
         alerts
           .getSuccessAlert()
           .should('contain', 'Account closed with success');
+
```
pebarron15: @Zeyber the e2e test does cover verify that the page is redirected to the homepage above on line 67

Zeyber: This line `cy.location('pathname').should('contain', '/');` will pass with any url that contains a `/` character which doesn't really mean anything. Instead we should look for elements unique to the home page.

projects/storefrontapp-e2e-cypress/cypress/integration/regression/my-account/close-account.core-e2e-spec.ts
```diff
@@ -69,6 +69,8 @@ describe('My Account - Close Account', () => {
...
+        cy.get('cx-login .cx-login-greet').should('not.exist');
```
Zeyber: The login component should be existing in some form when a page has successfully loaded. This does not test that we have successfully logged out but only that the DOM is changing and the element has disappeared temporarily.

Instead, we should look for the button to sign in.

feature-libs/user/profile/components/close-account/components/close-account-modal/close-account-modal.component.ts
Zeyber: This may be the case with manual testing but it is not apparent with the current e2e test. This test is known to be flaky as-is which is why we need to be meticulous in verifying correct behaviour.

FYI: Timing is a common issue with causing test flakes so if a test completes too quickly without verifying all steps completed, requests may be cancelled before the test finishes and cause issues for the next test.

Zeyber (changes requested):

feature-libs/user/profile/components/close-account/components/close-account-modal/close-account-modal.component.ts
Zeyber: You will also need to handle the breaking change here with the `featureConfigService.isLevel()` method.

projects/storefrontapp-e2e-cypress/cypress/integration/regression/my-account/close-account.core-e2e-spec.ts
pebarron15: This is checking for the greeting message where it states "Hi 'user'" when a user is logged in

projects/storefrontapp-e2e-cypress/cypress/integration/regression/my-account/close-account.core-e2e-spec-flaky.ts
```diff
@@ -64,11 +65,15 @@ describe('My Account - Close Account', () => {
...
-        cy.location('pathname').should('contain', '/');
+        const homePageAlias = waitForPage('homepage', 'getHomePage');
 
         alerts
           .getSuccessAlert()
           .should('contain', 'Account closed with success');
+
+        cy.wait(`@${homePageAlias}`);
+
+        cy.get('cx-login .cx-login-greet').should('not.exist');
```
Zeyber, line 76: Lets also check for the `Sign In / Register` link here.

Zeyber (changes requested):
You should add a note to the breaking changes doc `6_0.md` about this method change as its useful after a version upgrade to know what exactly has changed and might cause unexpected behaviour or require intervention.

You will also need to create a ticket for the deprecation cleanup so we can assign it later. You can check #16388 for examples on this.

Also, please change the PR name to use the standard before we commit this: `fix: log user out when closing account`

Zeyber (approved):

justinlee01 (approved):

PR #16360, opened 2022-10-21
chore: Run entire E2E suite on demand for testing purposes

bgambocjaviniar:
closes https://jira.tools.sap/browse/CXSPA-1588


.github/workflows/ci.yml
```diff
@@ -151,14 +154,14 @@ jobs:
       - name: Run e2es
         env: 
           SPA_ENV: ci,b2c
-          BUILD_NUMBER: ci-build-number-${{ github.event.pull_request.head.sha || github.sha }}
+          BUILD_NUMBER: ci-build-number-${{ github.event.pull_request.head.sha || github.run_id }}
```
bgambocjaviniar, line 157: build_number needed to be unique for parallelization 

PR is not a problem, but when we trigger a workflow_dispatch to run e2es on demand on 'develop' or 'any-other-branch', the github.sha is not unique as it will be the same as the previous run.

replacing it with run_id is fine as the build number will be really unique. It's already unique per commit for PRs, but now when re-triggering the entire e2e on demand for testing purposes, it will be unique due to the run_id. I was reluctant in to use run_id in the past at first creation when migrating from travis is because it's not unique when re-triggering a build, but we do not allow re-trigger, so not a problem here :)

RadhepS (approved):

PR #16361, opened 2022-10-21
feat: timeout backend requests in SSR and add error handling in SSR

### LINKED ISSUE #14559: Investigate ways to avoid hanging SSR renders (or to recover from them)
The SSR render might never finish when some async task is forever pending, e.g. http call to backend API. [In Angular Universal there is no way by default to abort the hanged render](https://indepth.dev/posts/1437/angular-universal-real-app-problems#7-inability-to-abort-rendering). Then the resources of the rendering are not released, which can cause memory leaks. When the memory leaks repeat this might eventually cause server to restart due to out of memory.

We already made some steps to improve the monitoring experience when the render hangs - we added config `SsrOptimizationOptions.maxRenderTime`. After the maxRenderTime we log in the console that the render takes way to long and it's probably hanging but we cannot do anything about it.

Ideally, we should avoid the hanging renders, when possible. First it would be good to investigate what are most common causes of the never-ending async operations in SSR. And then try to prevent them.

I only guess in most cases it's due to a never-responding HTTP API call.
If this is really the case (to be checked!), in SSR mode we could introduce a generic http timeout for all API calls, so they eventually return the control (with an error), so the SSR render can finish (even though with error).

-- 
Alternatively we might research alternative ways to do SSR (e.g. different from Angular Universal?), which would allow for programatically aborting the hanging renders and releasing the allocated resources.

Platonn:
- feature: configurable timeouts of HTTP requests - different per platform (server/browser). Configurable via global config 'config.backend.timeout' or per request via HttpContextToken.
- feature: in OptimizedSsrEngine prevent caching or sending rendered HTML to a client, if encountered http errors during rendering. Instead, fallback to CSR. In particular: during SSR intercept and collect all http errors (including timeout errors). Then transfer those errors to 'OptimizedSsrEngine' via object 'Response.locals'. 

closes #14559
closes CXSPA-529


Platonn:
closing in favor of https://github.com/SAP/spartacus/pull/17056

PR #16365, opened 2022-10-24
fix: Prerendering

znikola:
This PR fixes the prerendering functionality by making sure both the `SERVER_REQUEST_ORIGIN` and `SERVER_REQUEST_URL` injection tokens are provided, as these were not present outside of the express layer.

closes CXSPA-1555


tsconfig.json
```diff
@@ -9,8 +9,13 @@
     "moduleResolution": "node",
     "experimentalDecorators": true,
     "target": "es2020",
-    "typeRoots": ["node_modules/@types"],
```
znikola, line 12: Mostly formatting changes after I ran `yarn config:update`

projects/schematics/src/add-ssr/prerender.ts
```diff
@@ -0,0 +1,196 @@
...
+  export { renderModule, renderModuleFactory } from '@angular/platform-server';
+  export { AppServerModule } from './app/app.prerender';  
+  `;
+  return (tree: Tree) => {
+    tree.create(mainPrerenderTs, content);
+  };
+}
+
+function provideAppPrerenderModule(): Rule {
+  const appPrerenderModule = './src/app/app.prerender.ts';
```
Platonn: What about `app-prerender.module.ts` (to have `module.ts` as the suffix)?

core-libs/setup/prerender/providers/prerender-providers.ts
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+import { inject, StaticProvider } from '@angular/core';
+import { INITIAL_CONFIG } from '@angular/platform-server';
+import { SERVER_REQUEST_ORIGIN, SERVER_REQUEST_URL } from '@spartacus/core';
+
+/**
+ * Returns the providers for the prerender.
+ */
+export function getPrerenderProviders(options: {
```
Platonn: For consistency, let's prefix providers util functions with `provideXxx`. Examples:
- Spartacus: `provideConfig`, `provideConfigFactory`
- Angular 14 (ngmodule-less): `provideHttpClient`, `provideRoutes`

projects/storefrontapp/src/app/app.module.ts
```diff
@@ -43,7 +40,6 @@ if (!environment.production) {
 @NgModule({
   imports: [
     BrowserModule.withServerTransition({ appId: 'spartacus-app' }),
-    BrowserTransferStateModule,
```
Platonn, line 46: Is this change really related to this PR?

core-libs/setup/prerender/providers/prerender-providers.ts
```diff
@@ -0,0 +1,27 @@
...
+  requestOrigin: string;
```
Platonn: Let's add JSDoc - so customers know what to put here, with an example `https://example.com` (with ending slash? without ending slash?)

projects/storefrontapp/tsconfig.app.prod.json
```diff
@@ -3,6 +3,7 @@
   "compilerOptions": {
     "paths": {
       "@spartacus/setup": ["dist/setup"],
+      "@spartacus/setup/prerender": ["dist/setup/prerender"],
       "@spartacus/setup/ssr": ["dist/setup/ssr"],
```
Platonn: Let's discuss the structure. What about:
`setup/srr`
`setup/ssr/prerender`
`setup/ssr/app`

ideally:
`setup/server/ssr` - optimized ssr engine
`setup/server/prerender` - prerendering
`setup/server/app` - providers needed in the app
?

Why I ask:
In https://github.com/SAP/spartacus/pull/16361/files I plan to introduce a `BaseServerModule` to be imported in client's `AppServerModule`. My new module would be responsible for any core functionalities that are needed for SSR. For example it would provide BEFORE_APP_SERIALIZED hook that transfers errors from the App to the OptimizedSsrEngine. I guess I will need to put this Angular code in a different entry point than `/setup/ssr`, becasue it already contains native Node code.

projects/storefrontapp/src/app/app.module.ts
znikola: No, it's not. I figured to do the cleanup here, since I'm testing both ssr and prerendering.
Do you want to remove it from this PR?

projects/storefrontapp/tsconfig.app.prod.json
znikola: As discussed offline, we decide to leave it as is.
The import for customers will be `@spartacus/setup/prerendering`

core-libs/setup/prerender/providers/prerender-providers.ts
znikola: Done.

extra-webpack.setup.config.js
```diff
@@ -0,0 +1,10 @@
+module.exports = {
+  resolve: {
+    fallback: {
+      url: false,
+    },
+  },
+  node: {
+    global: true,
+  },
+};
```
Platonn, line 10: Instead of creating a new extra-webpack.config.js, that is used for all test runs,
can put fallback `url: false` only in [`karma.conf.js` of the setup lib](https://github.com/SAP/spartacus/blob/develop/core-libs/setup/karma.conf.js#L46-L49)?

package.json
```diff
@@ -21,7 +21,7 @@
...
+    "build:libs": "ng build core --configuration production && ng build storefrontlib --configuration production && concurrently --kill-others-on-fail yarn:build:schematics yarn:build:user && yarn build:cart && yarn build:order && concurrently --kill-others-on-fail yarn:build:checkout yarn:build:asm yarn:build:tracking && concurrently --kill-others-on-fail yarn:build:organization yarn:build:product yarn:build:product-configurator yarn:build:storefinder yarn:build:cdc && concurrently --kill-others-on-fail yarn:build:smartedit yarn:build:qualtrics yarn:build:assets yarn:build:cds yarn:build:digital-payments yarn:build:epd-visualization && yarn build:setup",
```
Platonn: Is this change really related to current PR?

core-libs/setup/ssr/providers/ssr-providers.ts
```diff
@@ -4,12 +4,93 @@
...
+        if (serverRequestUrl) {
+          if (options?.serverRequestOrigin) {
+            return serverRequestUrl.replace(
+              serverRequestOrigin,
+              options.serverRequestOrigin
+            );
+          }
+
+          return serverRequestUrl;
+        }
```
Platonn: I have a problem to find a usecase for replacing the domain for in the live SSR request. Therefore I'm not sure we should implement and maintain such feature.

(I perceive it as an edge case, so I'm ok with customers implementing it on their side with custom `{provide: SERVER_REQUEST_URL, use...<whatever edge case they need>}`.)


If there is a serverRequestOrigin already provided, IMHO let's leave it. If not, let's fallback to the configured serverRequestOrigin.

core-libs/setup/package.json
```diff
@@ -17,6 +17,7 @@
   },
   "peerDependencies": {
     "@angular/core": "^14.2.3",
+    "@angular/platform-server": "^14.2.3",
```
Platonn: Here we add a new dependency to the `@spartacus/setup` library, which can be perceived as a breaking change, at the first sight. When a customer bumps Spartacus version from 5.0 to 5.1, but didn't have `"@angular/platform-server"` package installed before, we should check if it will raise compilation errors for them.

core-libs/setup/ssr/providers/ssr-providers.ts
```diff
@@ -4,12 +4,93 @@
...
+import { INITIAL_CONFIG, PlatformConfig } from '@angular/platform-server';
 import { REQUEST } from '@nguniversal/express-engine/tokens';
 import { SERVER_REQUEST_ORIGIN, SERVER_REQUEST_URL } from '@spartacus/core';
 import { getRequestOrigin } from '../util/request-origin';
 import { getRequestUrl } from '../util/request-url';
 
+/**
+ * Returns the providers used for SSR and pre-rendering processes.
+ */
+export function provideSsrAndPrerendering(options?: {
```
Platonn: To simplify it, what about `provideServer()`?

core-libs/setup/package.json
Platonn: I did an experiment and confirmed that adding a new dependency `@angular/platform-server` that is used inside `@spartacus/setup/ssr` sub-entry point **is not a breaking change**, due to very specific circumstances (happy coincidence).

The `@spartacus/setup` library serves 2 purposes: 
a) `@spartacus/setup` main entry point - exports default configs for b2b occ endpoints. It's used only when a customer used any B2B-specific feature.
b) `@spartacus/setup/ssr` sub-entry point - exports SSR-related stuff. It's used only when a customer uses Angular/Spartacus SSR.

In situation (a) a customer doesn't import anything from `@spartacus/setup/ssr`. Therefore they don't import anything transitively from `@angular/platform-server`. So they don't need `@angular/platform-server` to be installed. **No errors.**

In situation (b) a customer already imports something from `@spartacus/setup/ssr`. Since they are using SSR, they must have installed `@angular/platform-server` before (becuase they had to import `ServerModule` from `@angular/platform-server` in their `AppServerModule`). So **no errors** due to the fact we started dependning in our `@spartacus/setup/ssr` on `@angular/platform-server`.

How I did an experiment
- created a fresh app with Spartacus v4.3.x, that doesn't use SSR, but uses a MyOrganization B2b feature (so during installation it imports B2B configs from `@spartacus/setup`, but doesn't install `@angular/platform-server` nor any other SSR deps). 
- Then added locally in the Spartacus repo in branch `release/4.3.x` an import of `INITIAL_CONFIG` token from `@angular/platform-server` inside our `OptimizedSsrEngine` source code:
   ```ts
   import { INITIAL_CONFIG, PlatformConfig } from '@angular/platform-server';
   ...
    protected spike: InjectionToken<PlatformConfig> = INITIAL_CONFIG;
    ```
- compiled the `@spartacus/setup` lib locally and published to verdaccio
- bumped in the app the Spartacus version to the latest one (the one from verdaccio)
- compiled and run the app - checked there were no erros

To double check, I also added in the app: `/// <reference types="@spartacus/setup/ssr" />`, then it caused compilation errors about missing dependencies `@nguniversal/express-engine`, `@angular/platform-server`, etc. Which was expected and is not an issue.

Platonn: **For consistency with other SSR-related dependencies, can we move this peerDep to `optionalDependencies`?**

tools/schematics/testing.ts
```diff
@@ -84,16 +82,16 @@ function beforeExit(): void {
...
+    newVersion = semver.parse(
       JSON.parse(fs.readFileSync('projects/core/package.json', 'utf-8')).version
-    );
+    )?.version;
   }
-
   // Bump version to publish
-  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-  semver.inc(currentVersion!, 'patch');
+  newVersion = semver.inc(newVersion ?? '', 'patch') ?? '';
```
Platonn: IMHO using `!` (non null assertion) was deliberate in this case. Our intention really isn't falling back to `''`. _We just know_ it won't be undefined. But typescript doesn't know.

Alternative way to fix the typesafety - never let `currentVersion` be undefined:
```ts
function getCurrentVersion(): string {
   const result = semver.parse(
      JSON.parse(fs.readFileSync('projects/core/package.json', 'utf-8')).version
    )?.version;
    if(!result) {
       throw new Error(`File 'projects/core/package.json' doesn't contain a valid field "version"`);
    }
    return result;
}

let currentVersion: string = getCurrentVersion();

function publishLibs(reload = false): void {
  if(reload) {
    currentVersion = getCurrentVersion();
  }
  currentVersion = semver.inc(currentVersion, 'patch');
  ...
}
```

docs/migration/6_0.md
```diff
@@ -1,5 +1,25 @@
...
+    ...provideSsrAndPrerendering({
+      serverRequestOrigin: process.env['SERVER_REQUEST_ORIGIN'],
+    }),
+    ...
+  ],
+})
+export class AppServerModule {}
+```
+
+The domain you specify `serverRequestOrigin` will be used as the origin when prerendering.
```
Platonn: Let's emphasize in this docs: 
When running live SSR (e.g. with ExpressJS), you don't want to configure `serverRequestOrigin`. It will be derived automatically from the client's request. Static configuration is needed ONLY for prerendering

angular.json
```diff
@@ -629,8 +635,11 @@
           }
         },
         "test": {
-          "builder": "@angular-devkit/build-angular:karma",
+          "builder": "@angular-builders/custom-webpack:karma",
           "options": {
+            "customWebpackConfig": {
+              "path": "./extra-webpack.setup.config.js"
+            },
```
Platonn, line 642: Instead of creating a new ~global~ `extra-webpack.config.js`, ~that is used for all test runs~,
can we simply add a `fallback` section with  `url: false` only in [`karma.conf.js` of the setup lib](https://github.com/SAP/spartacus/blob/develop/core-libs/setup/karma.conf.js#L46-L49)?

(this comment is valid only until we move SSR tests from Karma to Jest)

core-libs/setup/ssr/providers/ssr-providers.ts
```diff
@@ -4,12 +4,93 @@
...
+  /**
+   * Specify a domain (origin) from which the HTTP requests are being made.
+   * Should be without the trailing slash, e.g. "https://my.domain.com"
+   */
+  serverRequestOrigin?: string;
+}): StaticProvider[] {
+  return [
+    {
+      provide: SERVER_REQUEST_ORIGIN,
+      useFactory: (serverRequestOrigin?: string): string => {
```
Platonn: - For readability of the providers array, and for testability, I suggest to move factories to separate functions and add unit tests for them. 
- For type safety and ergonomics, what about using `inject()` function instead of `deps` array?

angular.json
Platonn: Update: As discussed offline, the `karma.conf.js` doesn't seem to respect the additional property `node: {global: true}`.

And my bad, I thought this extra-webpack.config.js is used for all unit tests, but only now I realized it's used only for `setup` libary

package.json
znikola: no, but I was experiencing the build failing due to the fact that `order` lib must be built before `checkout`.

core-libs/setup/ssr/providers/ssr-providers.ts
znikola: Levent pointed it out.
Look at it from the dev's experience - if you passed `https://my.domain.com` to `provideSsrAndPrerendering()`, you would expect your ORIGIN and URLs to be resolved using the domain you explicitly provided.
Otherwise, you would be confused why your ORIGIN and URLs are different in prerendering and SSR modes.

core-libs/setup/package.json
znikola: Awesome investigation, thank you for doing that!

znikola: Done.

core-libs/setup/ssr/providers/ssr-providers.ts
znikola: Sounds good to me 👍🏻

docs/migration/6_0.md
znikola: Related to https://github.com/SAP/spartacus/pull/16365#discussion_r1030243703 - it depends on how you look at it.
I would expect that my _explicitly_ provided value in `provideServer()` _will be used_, instead of silently ignored.
It will likely cause confusion.

I'll hold off with any action until we come to a conclusion

angular.json
znikola: Yep, and it will be remove in https://github.com/SAP/spartacus/pull/16532/files

core-libs/setup/ssr/providers/ssr-providers.ts
znikola: Thanks, good suggestion. I extracted the factories to separate files, please take a look.
Will proceed to testing, after we settle on https://github.com/SAP/spartacus/pull/16365#discussion_r1030243703

tools/schematics/testing.ts
znikola: Thanks! Done.

package.json
Platonn: Strange other folks don't experience it. Can you reproduce the same problem also in fresh `develop`?
If it's a _real issue_, for github history clarity, it would be nice to fix it in another PR.

docs/migration/6_0.md
Platonn: Ok, let's go with your proposal of implementation. It might be considered a more predictable. However IMHO hardcoded domain in SSR is not a common usecase, but rather unintended usage. Therefore I recommend to emphasize in .md docs and in JSDocs of this config property that it's recommended to use it in prerendering, but not in live SSR (as there the value will be figured out from the handled HTTP Request).

---
Side note:
IMHO we wouldn't have to discuss the implementation details of `provideServer()` for this edge case, if only we had a clear separation of concerns: SSR vs prerendering. E.g. have separate modules: AppServerModule and AppPrerenderingModule, which would contain different providers, only appropriate for their needs.

We now have an abstraction over them (SSR and prerendering), a shared `provideServer()` provider, and a shared AppServerModule. And hence we have to debate on unlikely edge-case handling and increasing the complexity of our internal code. 
On the other hand, having one module helps avoiding a complexity of setting up customer's repo 2 different modules: AppServerModule and AppPrerenderingModule (+ angular.json configurations).

In other words:
We loose:
- simplicity of our implmentation, but not _that much_
- clarity of our config (allows for unintended usecases - hardcoded domain even for live SSR)

We win:
- simplicty of customer's repo setup

I agree with your solution, as I prefer simplicity of customer's setup in this case. (as long as we emphasize in docs, that providing a hardcoded domain option is intended mainly for prerendering)

core-libs/setup/ssr/providers/ssr-providers.ts
Platonn: Agreed. Your implementation suggestion is more predictable for the user of the API.

core-libs/setup/ssr/providers/server-request-origin.ts
```diff
@@ -0,0 +1,48 @@
...
+
+    // SSR mode
+    if (serverRequestOrigin) {
+      return serverRequestOrigin;
+    }
+
+    // likely prerendering mode
+    if (options?.serverRequestOrigin) {
+      return options.serverRequestOrigin;
+    }
```
Platonn: We agreed on your solution, so we should prioritize value `options.serverRequestOrigin` over `serverRequestOrigin` injected from parent platform injector. Right?

core-libs/setup/ssr/providers/server-request-origin.ts
```diff
@@ -0,0 +1,48 @@
...
+    throw new Error(
+      `The request origin is not set. Please specify it through environment variables when initiating the process.
+    
+    E.g.
+    > SERVER_REQUEST_ORIGIN=https://my.domain.com yarn prerender
+    > SERVER_REQUEST_ORIGIN=http://localhost:4200 yarn serve:ssr
+    
+    
+    Alternatively, you can pass it as an argument to provideServer
+    function, but beware it will be used for server-side rendering as well.
```
Platonn: This message is confusing to me. Primarily the request origin can be passed only via `provideServer()`.
It's only a customer's possible implementation of their AppServerModule that maps the Node env variable `SERVER_REQUEST_ORIGIN` to the option of `provideServer`...

New interesting idea from top of my head:
Driving it REALLY by env variable, without any mapping? Like Angular does here https://github.com/angular/angular-cli/blob/503682766fa80bf324462ee700b86b2f477413ef/packages/angular_devkit/build_angular/src/utils/environment-options.ts#L81? I'm not sure if feasible though, would need to check...

Platonn: Regarding above interesting idea, I did a PoC that it's possible to "leak" the phrase `process.env["SPIKE"]` into the output compiled code of our library (I've run `ng build <library>` and checked the /dist/<library> code). It requires having  `types: ["node"]` in the tsconfig of this library.

Then... we might consider not to put any boilerplate in customer's AppServerModule like `provideServer({serverRequestOrgin: process.env['SERVER_REQUEST_ORIGIN'])`!
But instead just read the `process.env['CX_SERVER_REQUEST_ORIGIN']` inside our library's source code. 

IMHO this would help avoid much confusion for users of Live SSR (which are majority) - so they don't have   notion of passing the option `serverRequestOrgin` in their AppServerModule. 

And for prerendering (when the SERVER REQUEST ORIGIN is missing) -> we can provide docs and the descriptive error message (as you wrote above) with the instructions.

Platonn: Although technically it works to use Node env variables to configure Spartacus, it might not be a best idea. It would be a precedence in our codebase so should be first be justified enough.

It seems [that env variables are used in Angular CLI]((https://github.com/angular/angular-cli/blob/503682766fa80bf324462ee700b86b2f477413ef/packages/angular_devkit/build_angular/src/utils/environment-options.ts)) in compile-time (not in Angular runtime), and:
- the are quite unofficial:
   - they are not mentioned in the official [angular.io](http://angular.io/) docs
   - they are mentioned in community-driven blog posts
- the have a very limited purpose (scoped to the build time):
   1) enforcing no optimizations (for debugging purposes) - `NG_BUILD_DEBUG_OPTIMIZE` and `NG_BUILD_MANGLE`
   1) control how many CPU threads to use for compilation - `NG_BUILD_MAX_WORKERS`
   1) temporary backward compatibility with old SASS syntax (don’t confuse with @import deprecation): `NG_BUILD_LEGACY_SASS`
   1) debug performance of the build: `NG_BUILD_DEBUG_PERF`

No of above use cases look any similar to our use case for configuring the domain for prerendering (which is configuration of the application, not the CLI). 
And I agree with you it looks more natural to configure it explicitly via our public API.

package.json
znikola: I don't remember what I was doing at that moment to experience it. 
If you insist, we can revert this line?

core-libs/setup/ssr/providers/server-request-origin.ts
znikola: Correct, thanks for pointing it out. Done.

core-libs/setup/ssr/providers/server-request-origin.ts
znikola: yep, it's more convenient mechanism to "build once, deploy everywhere", but customers don't have to use it, ofc.

docs/migration/6_0.md
znikola: Thanks, I clarified in the docs.
And you're absolutely right, there are pros and cons to this approach, but I believe the balance is good in this case.

package.json
Platonn: If possible, let's revert it please

Platonn (approved):

PR #16378, opened 2022-10-27
fix: ASM Customer list header style(CXSPA-1551)

kimhw0630:
updated ASM customer list header according to mocks


justinlee01 (approved):

kimhw0630:
This is a new component and no incompatibility is introduced.

PR #16388, opened 2022-10-31
fix: Fix lost url paramaters and improve internal link recognition on…

Zeyber:
Closes: https://jira.tools.sap/browse/CXSPA-778

Deprecation cleanup: https://jira.tools.sap/browse/CXSPA-2015


projects/storefrontlib/cms-components/content/paragraph/paragraph.component.ts
```diff
@@ -26,15 +26,30 @@ export class ParagraphComponent {
...
-        event.preventDefault();
-        this.router.navigate([`/${href}`]);
+      /**
+       * @deprecated in 5.1
+       * TODO: Remove old method of resolving internal links.
+       */
+      if (this.featureConfigService?.isLevel('5.1')) {
+        const documentHost =
+          element.ownerDocument.URL.split('://')[1].split('/')[0];
+        console.log(documentHost, element.host);
```
pawelfras: `query:` do we want to keep this console log?

projects/storefrontlib/cms-components/content/paragraph/paragraph.component.ts
```diff
@@ -26,15 +26,30 @@ export class ParagraphComponent {
       const href = element?.getAttribute('href');
 
       // Use router for internal link navigation
-      if (href?.indexOf('/') === 0) {
-        event.preventDefault();
-        this.router.navigate([`/${href}`]);
+      /**
+       * @deprecated in 5.1
+       * TODO: Remove old method of resolving internal links.
```
pawelfras: Actually, such change is rather breaking change of `ParagraphComponent` as the constructor has been changed due to the new DI. I believe schematics rule for this breaking change should be part of this PR as well as proper info in documentation.

projects/storefrontlib/cms-components/content/paragraph/paragraph.component.spec.ts
```diff
@@ -46,6 +60,10 @@ describe('CmsParagraphComponent in CmsLib', () => {
             provide: CmsComponentData,
             useValue: MockCmsComponentData,
           },
+          {
+            provide: FeatureConfigService,
+            useClass: MockFeatureConfigService,
```
pawelfras, line 65: `should:` Because `FeatureConfigService` is optional, I believe it's worth adding unit tests that cover such a scenario.

projects/storefrontlib/cms-components/content/paragraph/paragraph.component.ts
```diff
@@ -25,16 +25,33 @@ export class ParagraphComponent {
...
   }
 
+  /**
+   * TODO: (#CXSPA-778) Remove featureConfigService from constructor in 6.0.
+   */
   constructor(
     public component: CmsComponentData<CmsParagraphComponent>,
-    protected router: Router
+    protected router: Router,
+    protected featureConfigService?: FeatureConfigService
```
pawelfras: `should:` Please double check if it works without the `Optional()` decorator, as we want to handle optional dependency injection

projects/storefrontlib/cms-components/content/paragraph/paragraph.component.spec.ts
Zeyber: Seems a bit odd  to test since the behaviour should be the same or we would test that a bug occurs in case the feature config service is not imported. This mock service will also be removed in 6.0 after cleanup as a result of the earlier cleanup TODO.

pawelfras: We could check whether `navigate` has been called instead of `navigateByUrl` but yes, such a test will have to be removed for 6.0. Adding this is totally up to you 🙂

pawelfras (approved):

PR #16389, opened 2022-10-31
fix: fix the bug in style version comparison

WeizhengSap:
ticket: CXSPA-1467


projects/storefrontstyles/scss/_versioning.scss
```diff
@@ -64,6 +64,81 @@ $useLatestStyles: false !default;
...
+  $v2Major: map-get($v2, major);
+  $v2Minor: map-get($v2, minor);
+
+  @if ($v1Major > $v2Major) {
+    @return 1;
+  }
+  @if ($v1Major < $v2Major) {
+    @return -1;
+  }
+  @if ($v1Major == $v2Major) {
```
plabadie: Perhaps this comparision can be removed.  If none is greater of lower, they must be equal

projects/storefrontstyles/scss/_versioning.scss
```diff
@@ -64,6 +64,81 @@ $useLatestStyles: false !default;
...
+    @if ($v1Minor > $v2Minor) {
+      @return 1;
+    }
+    @if ($v1Minor < $v2Minor) {
+      @return -1;
+    }
+    @if ($v1Minor == $v2Minor) {
```
plabadie: Perhaps this comparision can be removed too . If none is greater of lower, they must be equal

projects/storefrontstyles/scss/_versioning.scss
```diff
@@ -64,6 +64,81 @@ $useLatestStyles: false !default;
...
+
+@function toNumber($string) {
+  $strings: '0' '1' '2' '3' '4' '5' '6' '7' '8' '9';
+  $numbers: 0 1 2 3 4 5 6 7 8 9;
+
+  $result: 0;
+  @for $i from 1 through str-length($string) {
+    $character: str-slice($string, $i, $i);
+    $index: index($strings, $character);
+    $number: nth($numbers, $index);
```
plabadie: I think the index is already the number we want.  Maybe there is no need for `$numbers: 0 1 2 3 4 5 6 7 8 9;`

projects/storefrontstyles/scss/_versioning.scss
```diff
@@ -64,6 +64,81 @@ $useLatestStyles: false !default;
...
+
+  @return (major: $major, minor: $minor);
+}
+
+@function toString($value) {
+  @if type-of($value) == 'number' {
+    @return #{$value};
+  } @else if type-of($value) == 'string' {
+    @return $value;
+  } @else if type-of($value) == 'list' {
```
plabadie, line 103: In which case would there be a `list`?

projects/storefrontstyles/scss/_versioning.scss
```diff
@@ -54,7 +54,7 @@ $useLatestStyles: false !default;
 
 @mixin forVersion($from, $to: 0) {
   // We never create content that is part of future releases, this should not really happen.
-  @if ($from <= $_fullVersion) {
+  @if (compareVersion($from, $_fullVersion) <= 0) {
```
plabadie, line 64: Overall, what is the expected behaviour for patch versions?

projects/storefrontstyles/scss/_versioning.scss
WeizhengSap: index() starts from 1, not from 0. So, e.g. for character `5`, `index$` is 6. `nth` gets the 6th number from `$numbers`, so 5 is the result. Yes, we can get `$number` from `$index - 1`.

projects/storefrontstyles/scss/_versioning.scss
WeizhengSap: `1.2.3` will be `[1.2, 0.3]`; `1.2.3.4` will be `[1.2, 0.3, 0.4]`. So, customer can set any dot numbers, but we only get the first item in list `toString(nth($value, 1));`, that is `1.2`.

projects/storefrontstyles/scss/_versioning.scss
WeizhengSap: patch version is ignored. `1.2.3` will be treated as `1.2`. I can add the logic to handle patch version (not hard), the thing is, do we want to do it? From my understanding, patch is only for bug fix. There should be no breaking changes involved.

plabadie (dismissed):

plabadie (approved):

PR #16398, opened 2022-11-01
CXSPA-1421: Backporting UserIdHttpHeaderInterceptor

justinlee01:
… use it


projects/storefrontapp-e2e-cypress/cypress/helpers/b2b/my-company/features/assignments.ts
```diff
@@ -271,6 +271,8 @@ export function assignmentsTest(config: MyCompanyConfig) {
           .contains(ASSIGNMENT_LABELS.MANAGE)
           .click();
         cy.wait('@getAssignable');
+        // waiting for the UI element to finish loading. TODO: This is not ideal. there has to be a way to work around loading states.
+        cy.wait(1000);
```
plabadie: Agreed that waits should not be required.  On the other hand I am aware we are struggling with test stability/reliability.  If you think this improves things until we find a better solution, then it's ok since it's on the maintenance branch.

projects/storefrontapp-e2e-cypress/cypress.ci.json
```diff
@@ -11,7 +11,7 @@
   },
   "videoUploadOnPasses": false,
   "env": {
-    "API_URL": "https://20.55.1.150:9002",
+    "API_URL": "https://spartacus-demo.eastus.cloudapp.azure.com:8443",
```
plabadie: Is this change intentional?

projects/storefrontapp-e2e-cypress/cypress/helpers/b2b/my-company/features/assignments.ts
justinlee01: For the life of me I can't recall how precisely this change fixed the test. The code looks quite different in current `develop`; I could copy that over, but I'm not sure if anything else had changed.

`develop` code as reference:
https://github.com/SAP/spartacus/blob/develop/projects/storefrontapp-e2e-cypress/cypress/helpers/b2b/my-company/features/assignments.ts

feature-libs/asm/core/config/asm-config.ts
```diff
@@ -13,6 +13,18 @@ export abstract class AsmConfig extends OccConfig {
...
+      /**
+       * To fix certain features when a customer agent is emulating a user, we must send a header called
+       * "sap-commerce-cloud-user-id" to the backend fetching certain data as the user, rather than the
+       * agent. However, the header will break instances of Commerce Cloud who do not allow requests with
+       * this header. (For example, the configuration "corsfilter.commercewebservices.allowedHeaders" will
+       * need to allow it.)
+       *
+       * Enabling this feature will send the emulated user's IDs on requests that need it as context.
+       */
+      enable?: boolean;
```
plabadie, line 26: Is a default value set for this config?
I guess it should be falsy to be backwards compatible.

projects/storefrontapp-e2e-cypress/cypress.ci.json
plabadie: We need to revert the api url to what is was before in all the files of this PR.

feature-libs/asm/core/config/asm-config.ts
justinlee01: There's no default value; this is omitted. Omission has the same behavior, though I can add a default if it makes more sense to you.

plabadie: If the falsy default works, then this is fine.

plabadie (approved):

PR #16405, opened 2022-11-02
CXEC-17550: Navigate to Ticket Listing in case of 404 Not Found Ticket Detail

ishrarhussain:
Jira Issue # CXEC-17550 (https://jira.tools.sap/browse/CXEC-17550)

### DIFF
--- a/feature-libs/customer-ticketing/occ/adapters/occ-customer-ticketing.adapter.ts
+++ b/feature-libs/customer-ticketing/occ/adapters/occ-customer-ticketing.adapter.ts
@@ -4,6 +4,9 @@ import {
   ConverterService,
   normalizeHttpError,
   OccEndpointsService,
+  RoutingService,
+  GlobalMessageService,
+  GlobalMessageType
 } from '@spartacus/core';
 import {
   CustomerTicketingAdapter,
@@ -33,7 +36,9 @@ export class OccCustomerTicketingAdapter implements CustomerTicketingAdapter {
   constructor(
     protected http: HttpClient,
     protected occEndpoints: OccEndpointsService,
-    protected converter: ConverterService
+    protected converter: ConverterService,
+    protected routingService: RoutingService,
+    protected globalMessageService: GlobalMessageService
   ) {}
   getTicketAssociatedObjects(
     customerId: string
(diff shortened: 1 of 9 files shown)


feature-libs/customer-ticketing/occ/adapters/occ-customer-ticketing.adapter.ts
```diff
@@ -80,7 +85,18 @@ export class OccCustomerTicketingAdapter implements CustomerTicketingAdapter {
     return this.http
       .get<TicketDetails>(this.getTicketEndpoint(customerId, ticketId))
       .pipe(
-        catchError((error) => throwError(normalizeHttpError(error))),
+        catchError((error) => {
+          if(error.error.errors[0].type === 'NotFoundError'
+              && error.error.errors[0].message.toLowerCase().startsWith('ticket'))
```
khristian23: How about doing something to avoid "error.error.error"
```suggestion
        catchError((errorResponse) => {
          const errorDetails = errorResponse.error.errors[0];
          if(errorDetails.type === 'NotFoundError'
              && errorDetails.message.toLowerCase().startsWith('ticket'))
```
In addition, why do you need to validate the message? Wouldn't be enough to maybe verify the return status:  if it is 404, then we know for sure a ticket with "ticketId" was not found.
But what happens if for XYZ reason: the customerId is not the owner of the ticket?

khristian23 (changes requested):
I am OK with the code in the fix. Nevertheless, I have doubts about the fact the fix is done in the Adapter.  I was expecting to see it at the component level after the error bubbles up from the adapter to the ticket details component.
@Parthlakhani what is your opinion?

feature-libs/customer-ticketing/occ/adapters/occ-customer-ticketing.adapter.ts
```diff
@@ -80,7 +85,18 @@ export class OccCustomerTicketingAdapter implements CustomerTicketingAdapter {
     return this.http
       .get<TicketDetails>(this.getTicketEndpoint(customerId, ticketId))
       .pipe(
-        catchError((error) => throwError(normalizeHttpError(error))),
+        catchError((error) => {
+          if(error.error.errors[0].type === 'NotFoundError'
```
adolfobee: put this error in a constant

feature-libs/customer-ticketing/occ/adapters/occ-customer-ticketing.adapter.ts
adolfobee: same as above

Parthlakhani: It is easier to add in the adapter as we are catching the error there, we might be able to pass the error to the component level as well.

But, as discussed with Ishrar, we can create request-handler as well, we are just not sure which if it would be an ideal solution in our scenario. to be on the safe side, we should take the opinion from Brian as he will be reviewing at the end.

feature-libs/customer-ticketing/occ/adapters/occ-customer-ticketing.adapter.ts
ishrarhussain: Done. I've now added a new module for constants, as no such module existed earlier.

feature-libs/customer-ticketing/occ/adapters/occ-customer-ticketing.adapter.ts
ishrarhussain: Thanks guys.. I have now added the check on http status code and also added the constant strings in error details into a new module, which didn't previously exist.

adolfobee (approved):

khristian23 (approved):

PR #16408, opened 2022-11-02
CXEC-20553: fix file upload restrictions

Parthlakhani:
Add validation check for allowed file types
Remove file and reset form on send message

### DIFF
--- a/feature-libs/customer-ticketing/components/customer-ticketing-messages/customer-ticketing-messages.component.ts
+++ b/feature-libs/customer-ticketing/components/customer-ticketing-messages/customer-ticketing-messages.component.ts
@@ -1,6 +1,10 @@
-import { Component, OnDestroy } from '@angular/core';
+import { Component, OnDestroy, ViewChild } from '@angular/core';
 import { EventService } from '@spartacus/core';
-import { MessageEvent, MessagingConfigs } from '@spartacus/storefront';
+import {
+  MessageEvent,
+  MessagingComponent,
+  MessagingConfigs,
+} from '@spartacus/storefront';
 import {
   CustomerTicketingConfig,
   CustomerTicketingFacade,
@@ -17,6 +21,8 @@ import { map } from 'rxjs/operators';
   templateUrl: './customer-ticketing-messages.component.html',
 })
 export class CustomerTicketingMessagesComponent implements OnDestroy {
+  @ViewChild(MessagingComponent) messagingComponent: MessagingComponent;
+
   ticketDetails$: Observable<TicketDetails | undefined> =
     this.customerTicketingFacade.getTicket();
 
@@ -43,8 +49,10 @@ export class CustomerTicketingMessagesComponent implements OnDestroy {
               event.files.item(0),
               createdEvent.code
             );
+            this.messagingComponent.resetForm();
           } else {
             this.eventService.dispatch({}, GetTicketQueryReloadEvent);
+            this.messagingComponent.resetForm();
           }
         })
     );
--- a/feature-libs/customer-ticketing/components/customer-ticketing-messages/customer-ticketing-messages.component.ts
+++ b/feature-libs/customer-ticketing/components/customer-ticketing-messages/customer-ticketing-messages.component.ts
@@ -49,11 +49,10 @@ export class CustomerTicketingMessagesComponent implements OnDestroy {
               event.files.item(0),
               createdEvent.code
             );
-            this.messagingComponent.resetForm();
           } else {
             this.eventService.dispatch({}, GetTicketQueryReloadEvent);
-            this.messagingComponent.resetForm();
           }
+          this.messagingComponent.resetForm();
         })
     );
   }
--- a/projects/storefrontlib/shared/components/chat-messaging/messaging/messaging.component.ts
+++ b/projects/storefrontlib/shared/components/chat-messaging/messaging/messaging.component.ts
@@ -1,16 +1,26 @@
-import { Component, EventEmitter, Input, OnInit, Output } from '@angular/core';
+import {
+  Component,
+  EventEmitter,
+  Input,
+  OnInit,
+  Output,
+  ViewChild,
+} from '@angular/core';
 import { FormControl, FormGroup, Validators } from '@angular/forms';
 import { WindowRef } from '@spartacus/core';
 import { Observable } from 'rxjs';
 import { MessageEvent, MessagingConfigs } from './messaging.model';
 import { ICON_TYPE } from '../../../../cms-components/misc/icon/icon.model';
 import { FilesFormValidators } from '../../../services/file/files-form-validators';
+import { FileUploadComponent } from '../../form';
 
 @Component({
   selector: 'cx-messaging',
   templateUrl: './messaging.component.html',
 })
 export class MessagingComponent implements OnInit {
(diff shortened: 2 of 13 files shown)


adolfobee (approved):

projects/storefrontlib/shared/components/chat-messaging/messaging/messaging.component.ts
```diff
@@ -54,6 +64,10 @@ export class MessagingComponent implements OnInit {
     );
   }
 
+  get allowedTypes(): Array<string> {
+    return this.messagingConfigs?.attachmentRestrictions?.allowedTypes || [];
```
osahtout, line 68: Would there be a case where an empty list would be returned and allow types that were never supposed to be allowed?

osahtout (approved):

projects/storefrontlib/shared/components/chat-messaging/messaging/messaging.component.ts
Parthlakhani: empty list would mean that no files are allowed.

osahtout: Ahh nevermind I read it wrong

feature-libs/customer-ticketing/components/customer-ticketing-messages/customer-ticketing-messages.component.ts
```diff
@@ -43,8 +49,10 @@ export class CustomerTicketingMessagesComponent implements OnDestroy {
               event.files.item(0),
               createdEvent.code
             );
+            this.messagingComponent.resetForm();
           } else {
             this.eventService.dispatch({}, GetTicketQueryReloadEvent);
+            this.messagingComponent.resetForm();
           }
         })
```
khristian23:
```suggestion
              event.files.item(0),
              createdEvent.code
            );
          } else {
            this.eventService.dispatch({}, GetTicketQueryReloadEvent);
          }
          this.messagingComponent.resetForm();
        })
```

projects/storefrontlib/shared/services/file/files-form-validators.ts
```diff
@@ -56,4 +56,30 @@ export class FilesFormValidators {
...
+
+  /**
+   * Checks allowed types
+   *
+   * @param {Array<string>} allowedTypes Allowed types of files
+   * @returns Uses 'notParsable' validator error with allowedTypes property
+   * @memberOf FilesFormValidators
+   */
+  allowedTypes(allowedTypes?: Array<string>): ValidatorFn {
+    return (control: AbstractControl): ValidationErrors | null => {
```
khristian23, line 71: What if the allowedTypes function is misused? For Example is control is not a fileUpload control, should line 71 be protected or an exception thrown?
Maybe add Unit tests to be sure of the expectations.

projects/storefrontlib/shared/services/file/files-form-validators.ts
```diff
@@ -56,4 +56,30 @@ export class FilesFormValidators {
...
+      const errors: ValidationErrors = {};
+      if (allowedTypes && control.value) {
+        const files: File[] = Array.from(control.value);
+        files.forEach((file: File) => {
+          if (!allowedTypes.includes(this.getExtension(file.name))) {
+            errors.notParsable = true;
+          }
```
khristian23:
```suggestion
        errors.notParsable = files.some({name} => {
          return !allowedTypes.includes(this.getExtension(name));
```
or
```suggestion
        errors.notParsable = files.some({name} => {
          const fileExtensionIsNotAllowed = !allowedTypes.includes(this.getExtension(name));
          return fileExtensionIsNotAllowed;
```

projects/storefrontlib/shared/services/file/files-form-validators.ts
```diff
@@ -56,4 +56,30 @@ export class FilesFormValidators {
...
+        });
+      }
+      return Object.keys(errors).length === 0 ? null : errors;
+    };
+  }
+
+  protected getExtension(filename?: string): string {
+    return `.${filename?.split('.').pop()}` || '';
+  }
```
khristian23: Why filename is optional?
The return will never fallback in. ->.  || '';
Since the worst case scenario `.${filename?.split('.').pop()}` will be '.undefined'

In addition, if the filename does not have a '.' character the result of `.${filename?.split('.').pop()}` will be the same filename provided that the filename is not undefined
```  this.getExtension('filewithoutextension');   ==>  '.filewithoutextension' ```

File: projects/storefrontlib/shared/services/file/files-form-validators.ts [hunk 6]
```diff
@@ -21,7 +21,7 @@ export class FilesFormValidators {
   maxSize(maxSize?: number): ValidatorFn {
     return (control: AbstractControl): ValidationErrors | null => {
       const errors: ValidationErrors = {};
-      if (maxSize) {
+      if (maxSize && control.value) {
         const files: File[] = Array.from(control.value);
         files.forEach((file: File) => {
```
khristian23: file type can be inferred by TS.  Moreover, you could destructure to get {size}

projects/storefrontlib/shared/services/file/files-form-validators.ts
```diff
@@ -21,7 +21,7 @@ export class FilesFormValidators {
...
           if (file.size > maxSize * 1000000) {
```
khristian23: Please replace magic number

khristian23 (changes requested):

projects/storefrontlib/shared/services/file/files-form-validators.ts
Parthlakhani: It won't be an issue as restriction is from backend as well. This is for the better use experience.
And, protected methods can not be accessed from other components.

projects/storefrontlib/shared/services/file/files-form-validators.ts
Parthlakhani: changed it to 
```
errors.notParsable = files.some(
          (file) => !allowedTypes.includes(this.getExtension(file.name))
        );
```

projects/storefrontlib/shared/services/file/files-form-validators.ts
Parthlakhani: let's discuss in a call

projects/storefrontlib/shared/services/file/files-form-validators.ts
```diff
@@ -11,6 +11,9 @@ import { AbstractControl, ValidationErrors, ValidatorFn } from '@angular/forms';
   providedIn: 'root',
 })
 export class FilesFormValidators {
+  multiplier: number = 1000000;
```
khristian23:
```suggestion
  CONVERT_TO_MB = 1000000;
  or
  ONE_MB  = 1000000
```
It should be 10 MBs, right?   If you don't know the number leave it as it was

khristian23 (changes requested):

projects/storefrontlib/shared/services/file/files-form-validators.ts
Parthlakhani: Yes, that is right

khristian23 (approved):

PR #16422, opened 2022-11-04
Feature/CXSPA-1611: Truncate email address with ellipsis in unit-level order list

T0bas-2:
Truncate email address
Ticket: https://jira.tools.sap/browse/CXSPA-1611 

Bildschirmfoto 2022-11-04 um 16 29 41

*mobile view not affected


feature-libs/organization/unit-order/styles/_unit-order.scss
```diff
@@ -24,6 +24,15 @@
     td {
       width: 16.6%;
       padding: 1rem 0;
+      max-width: 1px;
```
MarcinGanczorz: setting `max-width: 1px;` is messing up the mobile view. Since we need this only on desktop and tablet view we should wrap it in media-breakpoint
```suggestion
@include media-breakpoint-up(md) {
  max-width: 1px;
}
```

feature-libs/organization/unit-order/styles/_unit-order.scss
```diff
@@ -24,6 +24,15 @@
...
+
+      .text {
```
MarcinGanczorz: maybe change the name of the class `.text` to something like `.text-ellipsis`

MarcinGanczorz (changes requested):

feature-libs/organization/unit-order/styles/_unit-order.scss
T0bas-2: Thanks that's correct, I adapted this part.

MarcinGanczorz (approved):
LGTM

PR #16435, opened 2022-11-07
CXSPA-1639: Cypress deleteFolder task mimics Unix rm -rf in that it does not error if the folder does not exist

justinlee01:
Fixes many cart import tests which seem to impact one another
https://nodejs.org/api/fs.html\#fspromisesrmdirpath-options

NOTE: This will mess up tests which depend on the "deleteFolder" throwing an error; though this seems unlikely to me, I would not be aware of the possibility.


giancorderoortiz:
Localhost tests continue to fail. Please check.

PR #16446, opened 2022-11-09
feat: menu for configurator OV page

ChristophHi:


feature-libs/product-configurator/common/assets/translations/en/configurator-common.ts
```diff
@@ -100,6 +100,12 @@ export const configurator = {
       noAttributeText: 'Remove filter(s) to see Overview content',
       itemPrice: 'Item Price',
     },
+    overviewMenu: {
+      title: 'Menu',
```
steinsebastian, line 104: Is a generic title "Menu" helpful for users? Wouldn't it be better to have something like "Groups" or "Attribute Groups"? Yes, it's in the wireframes but does it make sense?

ChristophHi: Hmhh agree. @JayEmKay-8 What do you think? Or maybe 'Group structure'?

feature-libs/product-configurator/rulebased/components/overview-menu/configurator-overview-menu.component.html
```diff
@@ -0,0 +1,111 @@
...
+  <ng-container *ngFor="let group of overviewGroups">
+    <div [ngClass]="getGroupLevelStyleClasses(level)">
+      <span class="cx-visually-hidden">
+        {{
+          'configurator.a11y.group'
+            | cxTranslate
+              : {
+                  group: group.groupDescription
+                }
+        }}
```
steinsebastian: This visually hidden span is not read during tabbing. It would only be read when using reading mode and it would lead to multiple times of reading the group information as this would be read, then the title of the span inside the button would be read and finally the content of the span itself. I suggest to remove it. Instead the button tag should be enhanced with an aria-label (see below).

feature-libs/product-configurator/rulebased/components/overview-menu/configurator-overview-menu.component.html
```diff
@@ -0,0 +1,111 @@
...
+      </span>
+      <button
+        class="cx-menu-item"
+        role="tab"
+        (click)="navigateToGroup(idPrefix, group.id)"
+      >
+        <span title="{{ group.groupDescription }}">
```
steinsebastian: Is there a reason for adding a title attribute? It would lead to multiple times of reading the same info in screen readers.

feature-libs/product-configurator/rulebased/components/overview-menu/configurator-overview-menu.component.html
```diff
@@ -0,0 +1,111 @@
...
+          {{ group.groupDescription }}</span
+        >
+      </button>
+      <ng-container *ngIf="group.subGroups?.length > 0">
```
steinsebastian, line 47: Suggested code for the button tag:

`      <button
        class="cx-menu-item"
        role="tab"
        [attr.aria-label]="
        'configurator.a11y.groupName'
          | cxTranslate: { group: group.groupDescription}
        "      
        (click)="navigateToGroup(idPrefix, group.id)"
      >
        <span aria-hidden="true">
          {{ group.groupDescription }}</span
        >
      </button>`

steinsebastian, line 47:
```suggestion
            <button
        class="cx-menu-item"
        role="tab"
        [attr.aria-label]="
        'configurator.a11y.groupName'
          | cxTranslate: { group: group.groupDescription}
        "      
        (click)="navigateToGroup(idPrefix, group.id)"
      >
        <span aria-hidden="true">
          {{ group.groupDescription }}</span
        >
      </button>
```

feature-libs/product-configurator/rulebased/components/overview-menu/configurator-overview-menu.component.html
```diff
@@ -0,0 +1,111 @@
...
+      <span class="cx-visually-hidden">
+        {{
+          'configurator.a11y.group'
+            | cxTranslate
+              : {
+                  group: group.groupDescription
+                }
+        }}
+      </span>
+      <button
```
steinsebastian, line 29:
```suggestion
            <button
        class="cx-menu-item"
        role="tab"
        [attr.aria-label]="
        'configurator.a11y.groupName'
          | cxTranslate: { group: group.groupDescription}
        "      
        (click)="navigateToGroup(idPrefix, group.id)"
      >
        <span aria-hidden="true">
          {{ group.groupDescription }}</span
        >
      </button>
```

steinsebastian: This suggestion would add an aria-label for the button and would omit the reading of the content of the button. It would read something like "Group Visual Entertainment"

feature-libs/product-configurator/rulebased/components/overview-menu/configurator-overview-menu.component.html
```diff
@@ -0,0 +1,111 @@
+<ng-container *ngIf="ovGroups$ | async as ovGroups; else ghostMenu">
+  <div class="cx-sticky-menu">
+    <div class="cx-menu-bar">
+      <button role="tab" class="active">
```
steinsebastian, line 4: Is there a reason for making the heading of the menu a button? There is no action that could be triggered.
If it is only for the sake of making clear for SR-users that they are in the group-menu I would suggest a hidden span-tag that is referenced by aria-describedby (see configurator-group-menu.html -> span with id "inListOfGroups").

feature-libs/product-configurator/common/assets/translations/en/configurator-common.ts
JayEmKay-8: Anthony is currently suggesting "Sections", but he wants to do a short hallway test before giving a definitive answer. So bear with me.

steinsebastian (changes requested):
Please have a look at the suggestions for accessibility

feature-libs/product-configurator/rulebased/components/overview-menu/configurator-overview-menu.component.html
ChristophHi: The reason is that soon there will be an action (switching between menu and filter component). This navigation will be moved to yet another component then, still I think we can leave this as it is (filter will be merged in 1 or 2 weeks)

steinsebastian (approved):

PR #16453, opened 2022-11-09
feat: Invalidate user upon password change

bgambocjaviniar:
~~closes https://jira.tools.sap/browse/CXSPA-1242~~

closnig in favor for https://github.com/SAP/spartacus/pull/16472


feature-libs/user/profile/components/update-password/update-password-component.service.ts
```diff
@@ -78,7 +102,15 @@ export class UpdatePasswordComponentService {
     );
     this.busy$.next(false);
     this.form.reset();
-    this.routingService.go({ cxRoute: 'home' });
+
+    // sets the redirect url after login
+    this.authRedirectService?.setRedirectUrl(
```
bgambocjaviniar, line 107: as the issue may indirectly imply, the logic would be similar to when someone changes the email.

PR #16464, opened 2022-11-14
CXSPA-805 Add Captcha for registration screen

ArkadiySukharenko:
[CXSPA-805](https://jira.tools.sap/browse/CXSPA-805)


projects/storefrontlib/shared/components/captcha/captcha.component.ts
```diff
@@ -0,0 +1,63 @@
...
+  constructor(
+    protected config: CaptchaApiConfig,
+    protected captchaService: CaptchaService,
+    private renderer: Renderer2
+  ) {}
+
+  ngAfterViewInit(): void {
+    if (this.config?.fields) {
+      Object.keys(this.config.fields).forEach((key) => {
+        if (this.config.fields) {
```
justinlee01: I think this is what was intended?
```suggestion
        if (this.config.fields[key]) {
```

projects/storefrontlib/shared/components/captcha/captcha.component.ts
```diff
@@ -0,0 +1,63 @@
...
+          this.renderer.setAttribute(
+            this.captchaRef.nativeElement,
+            key,
+            this.config.fields[key]
+          );
+        }
+      });
+    }
+
+    this.captchaService.getCaptchaConfig().subscribe((captchaConfig) => {
```
justinlee01: Destroy this subscription.

projects/storefrontlib/shared/components/captcha/captcha.service.ts
```diff
@@ -0,0 +1,119 @@
...
+
+/**
+ * Global function to be passes as "onload" url param for captcha <script>, to be
+ * triggered once script and dependencies are loaded
+ */
+declare global {
+  interface Window {
+    onCaptchaLoad: () => void;
+  }
+}
```
justinlee01: Can this declaration be moved to its own file?

projects/storefrontlib/shared/components/captcha/captcha.service.ts
```diff
@@ -0,0 +1,119 @@
...
+
+  getCaptchaConfig(): Observable<CaptchaConfig> {
+    return this.captchaConfigSubject$.asObservable();
+  }
+
+  renderCaptcha(
+    element: HTMLElement,
+    key: string,
+    callbackFunction?: () => void
+  ): void {
```
justinlee01: I think it would be nice if this returned an Observable, rather than accept a callback function.
So the code can be rewritten as, starting from line 90:
```
const retVal = new Subject<void>();

grecaptcha.render(element, {
  sitekey: key,
  callback: (response: string) => {
    this.token = response;
    this.retVal.next();
    this.retVal.complete();
  },
});

return retVal.asObservable();
```

projects/storefrontlib/shared/components/captcha/captcha.service.ts
```diff
@@ -0,0 +1,119 @@
...
+    if (element && key) {
```
justinlee01: This conditional is unnecessary, as Typescript checks that these are defined.
Or change the parameter types to be possibly undefined.

projects/storefrontlib/shared/components/captcha/captcha.service.ts
ArkadiySukharenko: Doesn't seem to work if I move it.

feature-libs/user/profile/components/register/register.component.ts
```diff
@@ -38,6 +38,7 @@ export class RegisterComponent implements OnInit, OnDestroy {
 
   isLoading$ = new BehaviorSubject(false);
 
+  private isCaptchaEnabled = false;
```
Zeyber: Should be `protected` instead since `private` is very difficult to modify if needbe.

feature-libs/user/profile/components/register/register.component.ts
```diff
@@ -139,10 +142,19 @@ export class RegisterComponent implements OnInit, OnDestroy {
   }
 
   submitForm(): void {
-    if (this.registerForm.valid) {
-      this.registerUser();
-    } else {
+    let isAllValid = true;
```
Zeyber: Can we append captcha to the `registerForm` instead of adding extra validation steps?

feature-libs/user/profile/occ/adapters/occ-user-profile.adapter.ts
```diff
@@ -25,13 +26,15 @@ import {
...
+import { CaptchaService } from '@spartacus/storefront';
 
 @Injectable()
 export class OccUserProfileAdapter implements UserProfileAdapter {
   constructor(
     protected http: HttpClient,
     protected occEndpoints: OccEndpointsService,
-    protected converter: ConverterService
+    protected converter: ConverterService,
+    protected captchaService: CaptchaService
```
Zeyber: This parameter must be optional if we are introducing it in 5.1

feature-libs/user/profile/occ/adapters/occ-user-profile.adapter.ts
```diff
@@ -51,6 +54,12 @@ export class OccUserProfileAdapter implements UserProfileAdapter {
       'Content-Type': 'application/json',
     });
     headers = InterceptorUtil.createHeader(USE_CLIENT_TOKEN, true, headers);
+    if (this.captchaService.getToken()) {
```
Zeyber: Since this logic is repeated, can we create a function for it? eg. `headers = appendCaptchaToken()`

projects/storefrontlib/shared/components/captcha/captcha.service.ts
```diff
@@ -0,0 +1,122 @@
...
+
+  getCaptchaConfig(): Observable<CaptchaConfig> {
+    return this.captchaConfigSubject$.asObservable();
+  }
+
+  renderCaptcha(element: HTMLElement, key: string): Observable<void> {
+    const retVal = new Subject<void>();
+
+    // @ts-ignore Global object created when captcha script is loaded
+    grecaptcha.render(element, {
```
Zeyber: This is google-specific code so we can't really keep it in the service as using any other provider would break it. Any way we could move it to the config?

feature-libs/user/profile/components/register/register.component.html
```diff
@@ -184,6 +184,11 @@
               </label>
             </div>
           </div>
+          <cx-captcha
```
Zeyber, line 187: Can we transform this component to one with form control embedded? See `DatePickerComponent` for an example.

Zeyber (changes requested):

projects/storefrontlib/shared/components/captcha/captcha.service.ts
ArkadiySukharenko: I think it was decided to "hardcode" it to recaptcha v2 as the only provider we support, but I'll try to move it to the config.

projects/storefrontlib/shared/components/captcha/config/default-captcha-api-config.ts
```diff
@@ -0,0 +1,27 @@
+/*
+ * SPDX-FileCopyrightText: 2022 SAP Spartacus team <spartacus-team@sap.com>
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { CaptchaApiConfig } from './captcha-api-config';
+import { Subject } from 'rxjs';
+
+export const defaultCaptchaApiConfig: CaptchaApiConfig = {
```
Zeyber, line 10: Lets try and move this to a service like `GoogleV2CaptchaService` that extends our `CaptchaService`. 

You can find examples of how this can be done in our `TagManagementService` or `TMS` (see `feature-libs/tracking`) which does something similar with third-party providers allowing users to import `Google Tag Manager` or  `Adobe Experience Platform`

Zeyber (changes requested):

projects/storefrontlib/shared/components/captcha/captcha.component.html
```diff
@@ -0,0 +1,2 @@
+<div #captcha class="g-recaptcha"></div>
```
Zeyber, line 1: `g-recaptcha` class is specific to google's recaptcha. Are we able to take a more generic approach? (eg. Insertable template using ng-container from GoogleV2Config)

projects/storefrontlib/shared/components/captcha/google-recaptchaV2/google-recaptchaV2.service.ts
```diff
@@ -0,0 +1,153 @@
...
+  getToken(): string {
+    return this.token;
+  }
+
+  /**
+   * Load external script with dependencies to be added to <head>.
+   * @param {string} lang - Language used by api in the script
+   */
+  loadScript(params?: { [key: string]: string }): void {
+    if (this.apiConfig?.apiUrl) {
```
Zeyber, line 132: Needs some tidy up?

projects/storefrontlib/shared/components/captcha/config/captcha-api-config.ts
```diff
@@ -0,0 +1,23 @@
...
+import { Injectable, Type } from '@angular/core';
+import { Config } from '@spartacus/core';
+import { CaptchaProvider } from '../captcha.model';
+
+@Injectable({
+  providedIn: 'root',
+  useExisting: Config,
+})
+export abstract class CaptchaApiConfig {
+  apiUrl?: string;
```
Zeyber, line 16: I think having `apiUrl` and `fields` can still be relevant only to specific captchas (google's in this case). Lets try to be as agnostic as we can at this generic captcha layer.
We can move these to a `GoogleRecaptcha2` options object that can apply them as `captchaOptions`.

projects/storefrontlib/shared/components/captcha/captcha.component.html
```diff
@@ -0,0 +1,2 @@
...
+<cx-form-errors [control]="control"></cx-form-errors>
```
Zeyber, line 2: Lets not have `cx-form-errors` here within the template to allow the user to add the errors tag where they like externally (eg. under the input like on the registration form).

projects/core/src/model/misc.model.ts
```diff
@@ -122,4 +122,10 @@ export interface BaseSite {
   urlEncodingAttributes?: string[];
   baseStore?: BaseStore;
   requiresAuthentication?: boolean;
+  captchaConfig?: CaptchaConfig;
+}
+
+export interface CaptchaConfig {
+  enabled: boolean;
+  publicKey?: string;
```
Zeyber, line 130: Can we leave a helpful jsdoc comment to explain a bit more about this config? eg. Public key is provided by the backend...

feature-libs/user/profile/occ/adapters/occ-user-profile.adapter.ts
```diff
@@ -159,4 +164,16 @@ export class OccUserProfileAdapter implements UserProfileAdapter {
       this.converter.pipeableMany(TITLE_NORMALIZER)
     );
   }
+
+  protected appendCaptchaToken(currentHeaders: HttpHeaders): HttpHeaders {
```
Zeyber, line 168: Instead of doing this in each adapter where we want to use captcha, lets move this to a "captcha" interceptor instead to provide the public key in the request (see `global-message` interceptors for examples).

Is it possible to tell if an endpoint is going to be captcha protected pre-emptively before the request? If so, we would be able to tell when we want to use captcha in the header without configuring each endpoint on the frontend. But I'm not sure if we could get away with it...

feature-libs/user/profile/components/register/register.component.html
```diff
@@ -184,6 +184,11 @@
...
+            [control]="registerForm.get('captcha')"
+            (enabled)="captchaEnabled($event)"
```
Zeyber, line 189: I think we could manipulate the form control from within the captcha component so that we do not need to have extra logic to handle `true` or `false` in the registration component.

Zeyber (changes requested):

PR #16469, opened 2022-11-14
Disable all Product Configurator Tests

giancorderoortiz:
Disable all product configurator tests pending cx upgrade and further investigation of tests flakiness


FollowTheFlo (approved):
fine to me

PR #16472, opened 2022-11-15
feat: Invalidate user upon password change

bgambocjaviniar:
closes https://jira.tools.sap/browse/CXSPA-1242

from https://github.com/SAP/spartacus/pull/16453


feature-libs/user/profile/components/update-password/update-password-component.service.ts
```diff
@@ -78,7 +102,15 @@ export class UpdatePasswordComponentService {
     );
     this.busy$.next(false);
     this.form.reset();
-    this.routingService.go({ cxRoute: 'home' });
+
+    // sets the redirect url after login
+    this.authRedirectService?.setRedirectUrl(
```
bgambocjaviniar, line 107: as the issue may indirectly imply, the logic would be similar to when someone changes the email.

feature-libs/user/profile/components/update-password/update-password-component.service.spec.ts
```diff
@@ -12,22 +14,35 @@ import { of } from 'rxjs';
 import { UpdatePasswordComponentService } from './update-password-component.service';
 import createSpy = jasmine.createSpy;
 
-class MockUserPasswordService implements Partial<UserPasswordFacade> {
+class MockUserPasswordFacade implements Partial<UserPasswordFacade> {
```
RadhepS, line 17: I think the name should be changed to MockUserUpdatePasswordService since that is what is being tested, what do you think?

RadhepS (approved):

PR #16477, opened 2022-11-15
fix: fix the issue of SmartEdit with Spartacus in SSR not working properly

WeizhengSap:
ticket: CXSPA-776


feature-libs/smartedit/root/services/smart-edit-launcher.service.ts
```diff
@@ -35,9 +55,16 @@ export class SmartEditLauncherService {
   load(): void {
     if (
       this.isLaunchedInSmartEdit() &&
-      this.featureModules.isConfigured('smartEdit')
+      this.featureModules.isConfigured(SMART_EDIT_FEATURE)
     ) {
-      this.featureModules.resolveFeature('smartEdit').subscribe();
+      if (this.featureConfigService?.isLevel('5.1')) {
+        if (this.platformId && isPlatformBrowser(this.platformId)) {
+          // we don't want to process smartedit when doing SSR
```
kacperknapik: Does it mean that Smartedit will not be available at all when SSR is used?

WeizhengSap: No, it just make smartedit contract added during client apps render,. So, you can still start your app in SSR.

bgambocjaviniar (changes requested):
new changes doesn't work properly on the server this time

using the setTimeout trick to simulate 'slowness' causes an infinite loop instead of it working. The previous solution was fine, but local was not. 
This solution local is fine, but server is not.

Screenshot 2023-01-09 at 9 12 01 AM

I had it deployed on p4 if you want to double check

projects/core/src/routing/store/reducers/router.reducer.ts
```diff
@@ -153,8 +153,13 @@ export class CustomSerializer
     semanticRoute = semanticRoute || this.lookupSemanticRoute(urlString);
 
     const { params } = state;
+    console.log('state', state.queryParams.cmsTicketId);
```
bgambocjaviniar: remove the log

feature-libs/smartedit/core/services/smart-edit.service.ts
```diff
@@ -35,11 +35,11 @@ export class SmartEditService {
     protected winRef: WindowRef,
     protected rendererFactory: RendererFactory2,
     protected config: SmartEditConfig,
+    /**
```
bgambocjaviniar, line 41: before I forget, put a TODO(CXPSA-ticketnumber): Remove constructor comment and reference a created issue for this

bgambocjaviniar (dismissed):

bgambocjaviniar (commented):
suite runs perfectly

https://cloud.cypress.io/projects/k3nmep/runs?branches=%5B%7B%22label%22%3A%22feature%2FCXSPA-776%22%2C%22suggested%22%3Afalse%2C%22value%22%3A%22feature%2FCXSPA-776%22%7D%5D&committers=%5B%5D&flaky=%5B%5D&page=1&status=%5B%5D&tags=%5B%5D&timeRange=%7B%22startDate%22%3A%221970-01-01%22%2C%22endDate%22%3A%222038-01-19%22%7D

note:
check for b2b is failing since recent build from manual trigger flaked for a b2b test

bgambocjaviniar (approved):

PR #16479, opened 2022-11-15
fix: icons in group menu shall have a tooltip (CXEC-16787)

Uli-Tiger:


feature-libs/product-configurator/common/assets/translations/en/configurator-common.ts
```diff
@@ -76,6 +76,12 @@ export const configurator = {
       cancelConfiguration: 'Cancel Configuration',
       cancelConfigurationMobile: 'Cancel',
     },
+    icon: {
+      groupComplete: 'Complete',
+      groupIncomplete: 'Incomplete',
+      groupConflict: 'Conflict',
+      subgroup: 'Sub-group',
```
steinsebastian: As the ">" icon means there is a sub-group, shouldn't be the text of the tooltip something like "has sub-groups" or "navigate to subgroup" similar?

Uli-Tiger: changed to "has sub-groups"

steinsebastian (approved):

PR #16483, opened 2022-11-16
fix: long text for exit button for screen readers

steinsebastian:
Provided aria-label for exit button. Thus "Exit Configuration" or "Cancel Configuration" is read by screen readers although the button contains only a short text ("Exit" or "Cancel").


Uli-Tiger:
@steinsebastian : a component test would be appreciated

Uli-Tiger (approved):

PR #16490, opened 2022-11-17
fix: bypass auth header for basesites request

FollowTheFlo:
add context with skipAuthorizationHeader flag


projects/core/src/auth/user-auth/services/auth-http-header.service.ts
```diff
@@ -158,6 +159,14 @@ export class AuthHttpHeaderService implements OnDestroy {
     return url.includes(this.occEndpoints.getBaseUrl());
   }
 
+  protected isBaseSitesRequest(request: HttpRequest<any>): boolean {
+    return request.url.includes(
+      this.occEndpoints.getRawEndpointValue('baseSites')
```
WeizhengSap, line 164: `baseSites` or `basesites`? or `/basesites`?

projects/core/src/auth/user-auth/services/auth-http-header.service.ts
```diff
@@ -158,6 +159,14 @@ export class AuthHttpHeaderService implements OnDestroy {
...
+    )
+      ? true
```
WeizhengSap: I think no need `? true : false`, right?

projects/core/src/auth/user-auth/services/auth-http-header.service.ts
FollowTheFlo: it needs to be 'baseSites' to match the property in default-occ-site-context-config.ts -> backend.occ.endpoint.baseSites. It is the way getRawEndpointValue fct works, it picks property value of backend.occ.endpoint[baseSites]. so fails if 'basesites' or '/basesites' are used.
To clarify, the target string we search is the one from the config so 'basesites?fields=baseSites(uid,defaultLanguage(isocode),urlEncodingAttributes,urlPatterns,stores(currencies(isocode),defaultCurrency(isocode),languages(isocode),defaultLanguage(isocode)),theme,defaultPreviewCatalogId,defaultPreviewCategoryCode,defaultPreviewProductCode)'

projects/core/src/auth/user-auth/services/auth-http-header.service.ts
FollowTheFlo: correct, it is redundant, i'll remove them.

feature-libs/asm/root/services/asm-auth-http-header.service.spec.ts
```diff
@@ -52,6 +52,11 @@ class MockOccEndpointsService implements Partial<OccEndpointsService> {
   getBaseUrl() {
     return 'some-server/occ';
   }
+  getRawEndpointValue(endpoint: string): string {
```
FollowTheFlo: needs to be dded in asm spec file to avoid unit tests failures as also using auth-http-header.service.ts as super class
super.alterRequest(request, token);

WeizhengSap: it can just return some random string
```
getRawEndpointValue(): string {
   return 'some value';
  }
```

WeizhengSap (dismissed):

WeizhengSap (approved):

PR #16505, opened 2022-11-18
Run e2e tests with Chrome instead of Electron

giancorderoortiz:
Compare performance between Chrome and Electron in CI


ci-scripts/e2e-cypress.sh
```diff
@@ -93,9 +93,9 @@ else
     echo '-----'
     echo "Running Cypress end to end tests"
 
-    if [ "${GITHUB_EVENT_NAME}" == "pull_request" ]; then
-        yarn e2e:run:ci:core"${SUITE}"
-    else
+#    if [ "${GITHUB_EVENT_NAME}" == "pull_request" ]; then
+#        yarn e2e:run:ci:core"${SUITE}"
+#    else
```
plabadie: This looks like a temp change that should be reverted.

giancorderoortiz: Correct. Thanks for pointing it out.

KateChuen (approved):

PR #16509, opened 2022-11-21
feat: CXSPA-1656 cypress tests for unit-level order history page - filtering and sorting

I565452:
I have combined cypress tests for filtering and sorting in one test file since both of them are using the same test data. 
Now test data is created only using the API - without using user interface elements.


projects/storefrontapp-e2e-cypress/cypress/integration/b2b/regression/unit-level-order/b2b-unit-level-order-history.e2e-spec.ts
```diff
@@ -0,0 +1,183 @@
...
+    doPlaceB2BOrder(sampleData.product6);
+    unitLevelOrderHistory.loginB2bUnitOrderViewerManager();
+    cy.visit(`/my-account/unitLevelOrders`).wait(500);
+  });
+
+  describe('Check sorting of unit-level orders by buyer name', () => {
+    it('should display unit-level orders sorted by Buyer (Ascending)', () => {
+      cy.get('.ng-input').first().click();
+      cy.get('[class=ng-option-label]').contains('Buyer (Ascending)').click();
+      checkThatOrdersAreSorted(fieldBuyer);
```
MarcinGanczorz, line 54: We should wait for the response from the server, before we validate the view.

MarcinGanczorz (changes requested):

projects/storefrontapp-e2e-cypress/cypress/integration/b2b/regression/unit-level-order/b2b-unit-level-order-history.e2e-spec.ts
```diff
@@ -0,0 +1,203 @@
...
+    cy.get('#order-history-table').should((tableData) => {
+      expect(tableData).to.not.be.empty;
+    });
+  }
+
+  function checkThatOrdersAreSorted(
+    dataCellSelector: string,
+    ascending: boolean = true
+  ) {
+    cy.get(dataCellSelector, { timeout: 500 }).then((items) => {
```
T0bas-2: Why does the timeout need to be reduced to 500, isn't a higher one (or the default) more reliable for this operation?

I565452: ups, it is my oversight, I have tested this options and left it unintentionally
updated

MarcinGanczorz (approved):

PR #16511, opened 2022-11-21
OV group headers should be sticky only for mobile widgets

Larisa-Staroverova:
To be able to test this PR, you should set `$useLatestStyles` to `true !default` in `_versioning.scss file`.
Before 5.1 version subgroups have not been supported, that's why the styling for it looks weird.

### DIFF
--- a/feature-libs/product-configurator/rulebased/styles/_configurator-overview-form.scss
+++ b/feature-libs/product-configurator/rulebased/styles/_configurator-overview-form.scss
@@ -3,10 +3,6 @@
   flex-direction: column;
   justify-content: flex-start;
   max-width: 1140px;
-
-  padding-inline-start: 0px;
-  padding-inline-end: 0px;
-
   padding-block-start: 5px;
   padding-block-end: 5px;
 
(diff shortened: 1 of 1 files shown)


feature-libs/product-configurator/rulebased/styles/_configurator-overview-form.scss
```diff
@@ -228,105 +192,50 @@
...
-          border: solid 1px var(--cx-color-light);
-          background-color: var(--cx-color-background);
           font-size: 1rem;
           font-weight: normal;
           text-transform: uppercase;
-          position: sticky;
+          border: solid 1px var(--cx-color-light);
+          background-color: var(--cx-color-background);
+
+          @include forVersion(5.1) {
```
ChristophHi: This is already nested into a forVersion(5.1) include, so we don't need it, right?

ChristophHi (changes requested):

feature-libs/product-configurator/rulebased/styles/_configurator-overview-form.scss
Larisa-Staroverova: You are right! I will remove it.

Larisa-Staroverova: Done!

ChristophHi (approved):

PR #16514, opened 2022-11-21
Feature/cxspa 1759: Show text with the selected filters

I539321:
Jira ticket: https://jira.tools.sap/browse/CXSPA-1759

### DIFF
--- a/feature-libs/organization/unit-order/components/unit-level-order-history/filter/unit-level-order-history-filter.component.html
+++ b/feature-libs/organization/unit-order/components/unit-level-order-history/filter/unit-level-order-history-filter.component.html
@@ -289,13 +289,11 @@
   >
     <div class="cx-selected-filters">
       {{ 'unitLevelOrderHistory.yourFilters' | cxTranslate }}
-      <span *ngIf="unitFilterMobileValue">"{{ unitFilterMobileValue }}"</span
-      ><span *ngIf="unitFilterMobileValue && buyerFilterMobileValue">{{
+      <span *ngIf="buyerFilterMobileValue">"{{ buyerFilterMobileValue }}"</span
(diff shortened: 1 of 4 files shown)


feature-libs/organization/unit-order/components/unit-level-order-history/filter/unit-level-order-history-filter.component.html
```diff
@@ -282,9 +282,32 @@
...
+
+  <div class="cx-unit-level-order-history-filters-show-and-remove">
+    <div
+      #showSelectedFilters
+      class="cx-selected-filters"
+      *ngIf="unitFilterMobileId.value || buyerFilterMobileId.value"
+    >
+      {{ 'unitLevelOrderHistory.yourFilters' | cxTranslate }}
+      <span *ngIf="unitFilterMobileId.value"
+        >"{{ unitFilterMobileId.value }}"</span
```
MarcinGanczorz: We shouldn't use control values here because they'll be updated even without request call. We should maybe store those values and update them only if we are about to send filter change event.

MarcinGanczorz (changes requested):

feature-libs/organization/unit-order/components/unit-level-order-history/filter/unit-level-order-history-filter.component.html
I539321: That is a good suggestion, will update it.

MarcinGanczorz (approved):

feature-libs/organization/unit-order/components/unit-level-order-history/filter/unit-level-order-history-filter.component.html
I539321: Resolving conversation

PR #16532, opened 2022-11-23
chore: Use Jest in Setup library

znikola:
closes CXSPA-1750


core-libs/setup/tsconfig.spec.json
```diff
@@ -1,12 +1,404 @@
...
   "compilerOptions": {
+    "baseUrl": "./",
     "outDir": "../../out-tsc/spec",
     "strict": false,
-    "target": "es2020",
-    "module": "es2020",
-    "types": ["jasmine", "node"]
+    "module": "CommonJs",
+    "types": ["jest"],
+    "paths": {
```
Platonn, line 9: Do we _really_ need all those "paths" mappings declared here, just for the specs of the `setup` library? Most of them are not used in this lib.

(Same comment applies for all other similar files in this PR)

feature-libs/checkout/jest.schematics.config.js
```diff
@@ -1,20 +1,25 @@
...
     'ts-jest': {
+      // our naming convention deviates from the standard "tsconfig.spec.json"
       tsconfig: '<rootDir>/tsconfig.schematics.json',
     },
   },
+
   collectCoverage: false,
   coverageReporters: ['json', 'lcov', 'text', 'clover'],
-  coverageDirectory: '<rootDir>/../../coverage/checkout/schematics',
+  coverageDirectory: '<rootDir>/../../coverage/cart/schematics',
```
Platonn: `checkout` vs. `cart` ? I guess it's a copy-paste typo?

feature-libs/order/package.json
```diff
@@ -29,6 +29,7 @@
     "@angular/common": "^14.2.3",
     "@angular/core": "^14.2.3",
     "@angular/forms": "^14.2.3",
+    "@angular/platform-browser-dynamic": "^14.2.3",
```
Platonn: Why adding `"@angular/platform-browser-dynamic"` and `zone.js` as a peer deps here in `order` library? The source code of this library doesn't import anything from those dependencies.

The fact that we use those dependencies for our tests only IMHO doesn't justify adding them as peer deps in the public package.json that we ship to customers.

(this comment applies also to other package.json files of libraries in this PR)

projects/schematics/src/wrapper-module/__snapshots__/index_spec.ts.snap
```diff
@@ -2,13 +2,13 @@
...
 exports[`Spartacus Wrapper Module Schematics: ng g @spartacus/schematics:wrapper-module Checkout Scheduled Replenishment should create the checkout wrapper module and import Checkout features 1`] = `
 "import { NgModule } from '@angular/core';
-import { checkoutB2BTranslationChunksConfig, checkoutB2BTranslations } from \\"@spartacus/checkout/b2b/assets\\";
-import { CheckoutB2BRootModule } from \\"@spartacus/checkout/b2b/root\\";
-import { checkoutTranslationChunksConfig, checkoutTranslations } from \\"@spartacus/checkout/base/assets\\";
-import { CheckoutRootModule, CHECKOUT_FEATURE } from \\"@spartacus/checkout/base/root\\";
-import { checkoutScheduledReplenishmentTranslationChunksConfig, checkoutScheduledReplenishmentTranslations } from \\"@spartacus/checkout/scheduled-replenishment/assets\\";
-import { CheckoutScheduledReplenishmentRootModule } from \\"@spartacus/checkout/scheduled-replenishment/root\\";
-import { CmsConfig, I18nConfig, provideConfig } from \\"@spartacus/core\\";
+import { checkoutB2BTranslationChunksConfig, checkoutB2BTranslations } from "@spartacus/checkout/b2b/assets";
```
Platonn, line 5: Just a headsup: we should also apply similar jest configuration (formatting) changes in the PRs that are about to be merged soon to develop (with new libraries, with their schematics and jest snapshots).

core-libs/setup/tsconfig.spec.json
znikola: They are necessary in jest setups. Otherwise, the [import](https://github.com/SAP/spartacus/pull/16532/files#diff-a8e35d359e755e7aa63e750b9b96184549bc9fd60c820cf986bcbfa178491c6bR9) from `@spartacus/core` wouldn't work in the setup's test.

feature-libs/checkout/jest.schematics.config.js
znikola: Yep, thanks for catching it! 
I double-checked the others and this is the only case.

feature-libs/order/package.json
znikola: It was outdated, please check again now.

core-libs/setup/tsconfig.spec.json
Platonn: Wouldn't it suffice to add just the _necessary_ path mapping, e.g. only for `@spartacus/core`?
Or to extend some ts-config where all those mapping mappings are already defined? 

My intention is not to repeat ourselves so much with those mapping here, and in every library's schematics.tsconfig. It's hardly maintainable.

tools/config/tsconfig-paths.ts
```diff
@@ -191,13 +212,66 @@ function handleSchematicsConfigs(
       }
     }
   });
+
+  if (showAllGood) {
+    success();
+  }
+}
+
+/**
```
Platonn: Why is setup library so special that it requires a separate config handler function?

znikola: Answered offline

Platonn (approved):

PR #16534, opened 2022-11-23
CXSPA-1696: ASM Bind cart enhancements

ercultimate:
Refactors the bind cart component to open a dialog when there is an existing cart to select how to handle the binding.  Currently only supports "replace", which saves the active cart before starting the binding process.


justinlee01 (dismissed):

justinlee01 (approved):
Concerning the old Lighthouse score, we don't have to export `AsmBindCartDialogComponent` from `AsmComponentsModule`.

PR #16536, opened 2022-11-23
Cypress 10: Evidently there is a new role from the backend that is not translated or identified in the BE. Adding it but read-only

justinlee01:


justinlee01:
@FollowTheFlo is fixing this. This is unfortunately failing in a known quirky test - 'apparel-checkout-as-guest' - and a new one - 'consignment-tracking.core-e2e.cy.ts'.

PR #16537, opened 2022-11-24
fix: recognize proper error description even if belongs to the same error type

Matejk00:
Considering this solution as a breaking change.
Closes: CXSPA-1760


projects/core/src/global-message/http-interceptors/handlers/http-error.handler.ts
```diff
@@ -44,6 +44,15 @@ export abstract class HttpErrorHandler implements Applicable {
     return errorResponse.status === this.responseStatus;
   }
 
+  /**
+   * Converts error description to translation key format.
+   *
+   * Example: 'User is disabled' will be transformed to 'user_is_disabled'.
+   */
+  createErrorTranslationKey(reason: string): string {
```
Zeyber: `getErrorTranslationKey` could be a better name for this method as it does not really create anything.

projects/core/src/global-message/http-interceptors/handlers/bad-request/bad-request.handler.ts
```diff
@@ -39,7 +39,9 @@ export class BadRequestHandler extends HttpErrorHandler {
     ) {
       this.globalMessageService.add(
         {
-          key: 'httpHandlers.badRequestPleaseLoginAgain',
+          key: `httpHandlers.badRequest.${this.createErrorTranslationKey(
```
Zeyber: Could be have a condition to revert to the old message in case `error_description` does not exist? (I don't know if we need to worry about this case.)

We can have a `getBadRequestTranslationKey` method return the new key if the `error_description` exists or return `httpHandlers.badRequestPleaseLoginAgain`. We can also handle the breaking change in this method if this is going into 5.2.

Zeyber (changes requested):

projects/core/src/global-message/http-interceptors/handlers/http-error.handler.ts
Matejk00: Right, I changed the method name.

projects/core/src/global-message/http-interceptors/handlers/bad-request/bad-request.handler.ts
Matejk00: I added a statement with a feature flag check and if a parameter is not a string. Is it what you meant?

Matejk00: And yes, hopefully, it should go into 5.2

projects/core/src/global-message/http-interceptors/handlers/http-error.handler.ts
```diff
@@ -49,8 +51,15 @@ export abstract class HttpErrorHandler implements Applicable {
    *
    * Example: 'User is disabled' will be transformed to 'user_is_disabled'.
    */
-  createErrorTranslationKey(reason: string): string {
-    return reason.toLowerCase().replace(/ /g, '_');
+  getErrorTranslationKey(reason: string): string {
+    const translationPrefix = `httpHandlers.badRequest`;
+
+    // TODO: Remove feature flag `isLevel` check for 6.0 (CXSPA-2413)
+    if (this.featureConfigService?.isLevel('5.2') || !reason) {
```
Zeyber: This will always return the `PleaseLoginAgain` for feature 5.2+ right? Is the logic backwards?

Zeyber (changes requested):

projects/core/src/global-message/http-interceptors/handlers/http-error.handler.ts
Matejk00: I changed the if statement to affect versions below 5.2.

Zeyber (approved):

PR #16550, opened 2022-11-28
Intersection service, enable adding an additional parameters

Larisa-Staroverova:


projects/storefrontlib/layout/loading/intersection.service.ts
```diff
@@ -27,26 +27,37 @@ export class IntersectionService {
...
+   * @param element - HTML element
+   * @param options - Allows to specify an optional root margin, in order to fire before the element shows up in the viewport
+   * @param intersectingCondition - Allows to specify an intersecting condition
+   * @returns Element intersects?
    */
   isIntersected(
     element: HTMLElement,
-    options?: IntersectionOptions
+    options?: IntersectionOptions,
+    intersectingCondition?: (entry: IntersectionObserverEntry) => boolean
```
znikola: Since we reuse this fn type in a couple of places, I suggest creating a type out of it

projects/storefrontlib/layout/loading/intersection.service.ts
```diff
@@ -27,26 +27,37 @@ export class IntersectionService {
    *
    * The returned observable will only emit the first value. The
    * observable must be cleaned up either way, since the value might never emit; it
-   *  depends on whether the element appears in the view port.
+   * depends on whether the element appears in the view port.
+   *
+   * @param element - HTML element
+   * @param options - Allows to specify an optional root margin, in order to fire before the element shows up in the viewport
+   * @param intersectingCondition - Allows to specify an intersecting condition
```
znikola: Let's make it clear what happens if `intersectingCondition` is not provided

projects/storefrontlib/layout/loading/intersection.service.ts
```diff
@@ -27,26 +27,37 @@ export class IntersectionService {
...
 
   /**
-   * Returns an observable that emits for every change of intersection of a given element
-   * @param  element - HTML element
-   * @param  options - Allows to specify an optional root margin, in order to fire before the element shows up in the viewport
+   * Returns an observable that emits for every change of intersection of a given element.
+   *
+   * @param element - HTML element
+   * @param options - Allows to specify an optional root margin, in order to fire before the element shows up in the viewport
+   * @param intersectingCondition - Allows to specify an intersecting condition
```
znikola: The same as above - can we make it a bit clearer to customers what happens if this is provided, and what happens if it is not provided ?

znikola (changes requested):

projects/storefrontlib/layout/loading/intersection.service.ts
Larisa-Staroverova: Done!

projects/storefrontlib/layout/loading/intersection.service.ts
Larisa-Staroverova: Done!

projects/storefrontlib/layout/loading/intersection.service.ts
Larisa-Staroverova: Done!

znikola (approved):

PR #16553, opened 2022-11-28
fix: asm customer list tab problem (CXSPA-1886)

kimhw0630:
closes https://jira.tools.sap/browse/CXSPA-1886

added tabindex to asm customer list link


feature-libs/asm/components/asm-main-ui/asm-main-ui.component.html
```diff
@@ -18,12 +18,15 @@
...
         <a
           #customerListLink
+          tabindex="0"
+          role="button"
           class="cx-asm-customer-list-link"
           (click)="showCustomList()"
-          ><span>{{ 'asm.customers' | cxTranslate }}</span></a
+        >
+          <cx-icon [type]="iconTypes.USER_FRIENDS"></cx-icon>
+          <span>{{ 'asm.customers' | cxTranslate }}</span></a
```
justinlee01, line 29: To double-check - do we need an aria-label here?

kimhw0630: I don't think so because usually, we use aria-label for non-text links and buttons... also AMP doesn't complain.

justinlee01 (approved):

PR #16562, opened 2022-11-29
chore: include SonarQube scans for pull requests

RadhepS:
closes: https://jira.tools.sap/browse/CXSPA-1641

Changes:
- Add workflow to run SonarQube scans only on PRs to identify if new code changes raise issues
- Add exclusions/inclusions directly on the SonarQube dashboard so the release and public pipelines will have the same files there resulting in a single source of truth
- Removed the exclusions/inclusions from the sap-pipelines branch on the private repository since it will now be retrieved from the SonarQube dashboard

Things tested/verified:
- Files scanned are only new files introduced in the PR
- Exclusions/Inclusions work
- PR is correctly compared to the branch its trying to merge into
- New code issues are raised correctly (intentionally added a problematic file and it was flagged)
- Made sure SonarQube scans are only run against develop/release branches
- SonarQube scans on release pipeline still work correctly even with the exclusion/inclusion changes


.github/workflows/sonarqube-scan.yml
```diff
@@ -0,0 +1,23 @@
+on:
+  pull_request:
```
bgambocjaviniar: discussion for team on push for branches like develop / release / epic / etc

.github/workflows/sonarqube-scan.yml
```diff
@@ -0,0 +1,23 @@
...
+    types: [opened, synchronize]
+
+name: SonarQube scan
+jobs:
+  sonarqube_scan:
+    name: SonarQube Scan
+    runs-on: ubuntu-latest
+    steps:
```
bgambocjaviniar: good to have extra step for this job, where if the sonar step fails, then log a static message explaining why it's blocking the merge

.github/workflows/sonarqube-scan.yml
```diff
@@ -0,0 +1,23 @@
...
+      - name: Scanning files using SonarQube
+        uses: sonarsource/sonarqube-scan-action@master
+        with:
+          args: >
+            -Dsonar.qualitygate.wait=true
+            -Dsonar.projectKey=composable-storefront
+        env:
+          SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
+          SONAR_HOST_URL: https://sonar.tools.sap
+    if: ${{ github.base_ref == 'develop' || startsWith(github.base_ref, 'release') }}
```
bgambocjaviniar: quick question

even if it's not develop / release, wouldn't it be better that it's applied to any PRs whether or not it's the above base branch or not? 

it will help secure code quality instead of fixing things later once it gets into develop / release

.github/workflows/sonarqube-scan.yml
RadhepS: won't do as discussed

.github/workflows/sonarqube-scan.yml
RadhepS: discussed with team in stand up and we won't run them on push

.github/workflows/sonarqube-scan.yml
RadhepS: changed, will run on all PRs

bgambocjaviniar (approved):

PR #16575, opened 2022-12-01
Bug Fix - CXSPA-1766 - Account Summary attachment timeout

kuhris-sap:
Fix for bug: https://jira.tools.sap/browse/CXSPA-1766


projects/core/src/auth/user-auth/http-interceptors/auth.interceptor.ts
```diff
@@ -100,4 +101,48 @@ export class AuthInterceptor implements HttpInterceptor {
   protected isExpiredToken(resp: HttpErrorResponse): boolean {
     return resp.error?.errors?.[0]?.type === 'InvalidTokenError';
   }
+
+  /**
+   * If the input is a json blob of type HttpErrorResonse: extract the json error from the blob
+   * Otherwise: rethrow the error
+   */
```
WeizhengSap: I think you can create some blob error handler inside "account summary" instead of modify the existing AuthInterceptor. Please take a look of `http-interceptors` in `/organization/administration` (e.g. OrganizationBadRequestHandler). And in "account summary", you can reuse the functions in "storefrontlib".

WeizhengSap (approved):

PR #16576, opened 2022-12-01
fix: checkout payment type sonar issue

RadhepS:
closes: https://jira.tools.sap/browse/CXSPA-1893


feature-libs/checkout/b2b/components/checkout-payment-type/checkout-payment-type.component.ts
```diff
@@ -81,6 +81,7 @@ export class CheckoutPaymentTypeComponent {
             });
           return availablePaymentTypes[0];
         }
+        return undefined;
```
bgambocjaviniar, line 84: instead of undefined, what do you think about an `empty object` to adhere to it the observable expecting an object

bgambocjaviniar: think of it like this https://github.com/SAP/spartacus/pull/16576/files#diff-054e45bd3f923f32810253d7cd96e7b5710824575c3a475549694902485632d1R48

in the service for `getPaymentTypes`, we set the default as an empty array for when the data is undefined. It's not expecting Observable<someType | undefined>

RadhepS: We would need to change the logic further because we do a check to filter out all undefined values, with the default returning an object, the logic would also need to be modified to exclude empty objects. Is it common practice for us to return an empty object? The reason I ask that is because an empty object is not falsy which could cause issues if not handled properly.

https://masteringjs.io/tutorials/fundamentals/empty-object-truthy#:~:text=There%20are%20only%20seven%20values,You%20can%20use%20the%20Object.

RadhepS: We would also need to change the filter to also check if an empty object is returned? Because right now for this use case the tap() would not be executed since undefined is filtered out. Just returning an empty object would cause the tap() to be expected since its not filtered?

bgambocjaviniar: > Just returning an empty object would cause the tap() to be expected since its not filtered?

I actually didn't look at the whole lol I thought it stopped after the map, but yes

bgambocjaviniar (approved):

PR #16593, opened 2022-12-03
Task/asm e2e fix

kimhw0630:


projects/storefrontapp-e2e-cypress/cypress/helpers/asm.ts
```diff
@@ -245,7 +245,8 @@ export function loginCustomerInStorefront(customer) {
 export function agentSignOut() {
   const tokenRevocationAlias = loginHelper.listenForTokenRevocationRequest();
   cy.get('button[title="Sign Out"]').click();
-  cy.wait(tokenRevocationAlias).its('response.statusCode').should('eq', 200);
+  // due to flaky not to check 200 status because revocation get called twice and some time one of call fails
```
justinlee01, line 248: `// When the agent signs out, there are two revocations made simultaneously - the second one occasionally fails, though it is not necessary. We therefore are not interested in the status code of the second one.`

projects/storefrontapp-e2e-cypress/cypress/integration/b2b/regression/checkout/asm-cost-center.e2e-spec.ts
```diff
@@ -55,19 +65,23 @@ context('B2B - ASM Account Checkout', () => {
...
-      alerts.getErrorAlert().within(() => {
+    cy.log('--> should not show error on valid cost cente');
+    alerts.getErrorAlert().then((alert) => {
+      cy.wrap(alert).within(() => {
         cy.get('button').click();
       });
     });
-    cy.intercept('PUT', '*costcenter?costCenterId=*').as('costCenterReq');
+    // flaky:  added timer between intercepters
+    cy.wait(1000);
```
justinlee01: The same request is intercepted twice, causing errors. A solution possibly would be to remove this and line 77, though it may cause issues. Or branch it off into two tests.

PR #16605, opened 2022-12-06
Instead of redirecting to logout route, logout internally then redire…

justinlee01:
…ct to home


pebarron15:
@Zeyber can you review this solution? Using the preferred logout method did not work as desired for this bug.

pebarron15 (approved):

PR #16608, opened 2022-12-07
chore: basic configuration for OPF integration with both b2c and b2b

psmul:
Closes: [CSOCO-2808](https://jira.tools.sap/browse/CCSOCO-2808)


integration-libs/opf/root/model/augmented-types.ts
```diff
@@ -0,0 +1,13 @@
+/*
+ * SPDX-FileCopyrightText: 2022 SAP Spartacus team <spartacus-team@sap.com>
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+// import '@spartacus/checkout/base/root';
+
```
pawelfras, line 9: `could:` consider adding a TODO comment to track this augmentation issue.

pawelfras (approved):

integration-libs/opf/root/model/augmented-types.ts
psmul: Thanks, great catch!

PR #16609, opened 2022-12-07
test: Fix flaky close account e2e test

Zeyber:
Closes: [CXSPA-1926](https://jira.tools.sap/browse/CXSPA-1926)


projects/storefrontapp-e2e-cypress/cypress/integration/regression/my-account/close-account.core-e2e-spec.ts
```diff
@@ -92,5 +95,20 @@ describe('My Account - Close Account', () => {
...
       });
     });
+
+    /**
+     * Waits for the background work to complete after a cy.visit() call and the browser adds base parameters.
+     *
+     * Handling of authentication revocation can slow down page loads to the point that tests fail due to timeouts.
+     * When the background work completes after a cy.visit(), the base paramaters are added to the url and the page loads normally.
+     */
+    function waitForRedirections() {
```
Matejk00: What about moving this function into level up? I mean instead creating it exactly in a spec file, maybe the better will be to add it to some helper file. Then other tests can re-use such method if needed.

Matejk00: Also I am thinking if 'contain' value can be passed as an argument.

Matejk00: What do you think @Zeyber?

Zeyber: I have added the function to an appropriate helper file and given it a path value.

I would prefer not to add a contain value since this function is to solve the issue of waiting for background work that relies specifically on base params being added to the path to indicate this work is done.

Matejk00 (approved):

PR #16612, opened 2022-12-07
Fix customer list dialog 400% zoom behavior: CXSPA-1904

bullfroggy-sap:


justinlee01 (dismissed):

justinlee01 (dismissed):

kimhw0630:
@bullfroggy-sap  Jeremy, your change will create a scroll bar for the entire dialog...
(change customerList.pageSize = 25 in default-asm-config.ts:)
left side is before your change and right side is after your change... I don't think UX wants entire scroll bar
image

kimhw0630 (dismissed):

feature-libs/asm/styles/components/_customer-list.component.scss
```diff
@@ -4,7 +4,8 @@ $sapFioriContent_LabelColor: #6a6d70;
 
 %cx-customer-list {
   .cx-modal-content {
-    min-height: 424px;
+    max-height: 100%;
+    overflow-y: auto;
```
justinlee01, line 8: Not to be a nit, but should we remove this `overflow-y` too? Does it cause double scroll bars?

bullfroggy-sap: need the double scroll bars else the ui at 100% gets messed up

justinlee01 (dismissed):

ercultimate (dismissed):

ercultimate (dismissed):

plabadie (dismissed):

ercultimate (approved):

PR #16615, opened 2022-12-08
test: add voucher request interceptor to coupons e2e 

FollowTheFlo:
From video error: the POST occ vouchers  call never happened when press coupons Apply btn
Added an interceptor for POST occ vouchers before Apply click
Added wait command after click
make the test more robust to avoid race condition and failing request
image


projects/storefrontapp-e2e-cypress/cypress/helpers/coupons/cart-coupon.ts
```diff
@@ -142,9 +142,11 @@ export function claimCoupon(couponCode: string) {
 const cartCouponInput = 'input.input-coupon-code';
 const cartCouponButton = 'button.apply-coupon-button';
 const applyCartCoupon = (code: string) => {
+  interceptVoucherPostRequest();
   cy.get('cx-cart-coupon').within(() => {
     cy.get(cartCouponInput).type(code);
     cy.get(cartCouponButton).click();
```
FollowTheFlo, line 148: Click on 'Apply' Coupon always generated POST occ vouchers call, adding a wait command to avoid race condition.

RadhepS (approved):

PR #16620, opened 2022-12-08
CXSPA-2005: Choose only the most recent, stable version of Chromium

justinlee01:


projects/storefrontapp-e2e-cypress/cypress/plugins/index.js
```diff
@@ -28,5 +28,18 @@ module.exports = (on, config) => {
 
   /* Set exact timestamp to be shared in all spec files */
   config.env.TIMESTAMP = Date.now() - 1535535333333;
+
+  /* Filter only stable Chromium browsers, then sort by the browsers by version (which is in string) and choose the most recent. */
+  const latestStableChromeBrowsers = config.browsers
+    .filter(
+      (browser) => browser.channel === 'stable' && browser.family === 'chromium'
```
RadhepS: While testing I noticed something.

Even though we pass --browser chrome in the package.json, when we go to the list of browsers it still includes non-chrome browsers such as Electron and FireFox (for me when running locally).

So it might be possible that we'll select non-chrome browsers in this case? For example Electron is also on stable and its family is chromium.

What I'm thinking of to be on the safe side:
```suggestion
      (browser) => browser.channel === 'stable' && browser.family === 'chromium' && browser.name === 'chrome'
```

This always means we can remove `--browser chrome` on the package.json so we have one source of truth.

justinlee01: Oh, Electron is chromium? When I ran it locally on my end, I believe it eliminated Electron. ~If it is as you say though, you can commit this suggestion.~ Actually, I'll stick to the safe side and make the changes you recommended.

justinlee01: I think we still need to specify `--browser chrome` - the config only controls which ones are valid, but "electron" is still considered the default.
https://github.com/cypress-io/cypress/issues/6646

RadhepS: > Oh, Electron is chromium? When I ran it locally on my end, I believe it eliminated Electron. ~If it is as you say though, you can commit this suggestion.~ Actually, I'll stick to the safe side and make the changes you recommended.

You're right here it did eliminate Electron but that's only because Electron has a lower version. It's only removed by your filter which takes the highest version, that's why Chrome gets returned. If for whatever reason Electron had a higher version, it would be returned. This happens even if the flag in the package.json is `--browser chrome` which is why I suggested putting chrome in the if statement. This is what I observed when I tested it anyways.
image

RadhepS: > I think we still need to specify `--browser chrome` - the config only controls which ones are valid, but "electron" is still considered the default. [cypress-io/cypress#6646](https://github.com/cypress-io/cypress/issues/6646)

You're right about needing the flag, I re-tested it

RadhepS (approved):

PR #16621, opened 2022-12-08
CXCDS-6543 added additional scope "price" to refresh value after logi…

szegi:
added additional scope "price" to refresh value after login,logout


giancorderoortiz (approved):
e2e tests pass

PR #16632, opened 2022-12-12
fix: Schematics installer for PWA and SSR

znikola:
fixes CXSPA-2013


projects/schematics/src/add-pwa/index_spec.ts
```diff
@@ -64,6 +64,64 @@ describe('Spartacus Schematics: add-pwa', () => {
...
+      const appModule = tree.readContent(
+        '/projects/schematics-test/src/app/app.module.ts'
+      );
+      expect(appModule).toMatchSnapshot();
+
+      const packageJson = tree.readContent('/package.json');
+      expect(packageJson).toMatchSnapshot();
+    });
+
+    it('should modify package.json', async () => {
```
pawelfras, line 87: `query:` is it not already checked in the previous test?

znikola: Good catch, thanks!
It was a leftover, I removed the one from above.

pawelfras (approved):

PR #16634, opened 2022-12-12
CXSPA-2003: Fix search for first group with attributes

steinsebastian:


feature-libs/product-configurator/rulebased/core/state/effects/configurator-basic-effect.service.ts
```diff
@@ -14,35 +14,57 @@ import { Configurator } from '../../model/configurator.model';
...
+   * Throws error if such a group does not exist, as this is an illegal state
    * @param configuration
+   * @param includeConflicts (optional) if true it includes also conflict groups in the search
    * @returns Group id
+   *
    */
   getFirstGroupWithAttributes(
-    configuration: Configurator.Configuration
+    configuration: Configurator.Configuration,
+    includeConflicts: boolean = false
```
ChristophHi:
```suggestion
    includeConflicts = false
```
(Sonar)

feature-libs/product-configurator/rulebased/core/state/effects/configurator-basic.effect.ts
```diff
@@ -269,7 +269,8 @@ export class ConfiguratorBasicEffects {
                 map((currentGroupId) => {
                   const groupIdFromPayload =
                     this.configuratorBasicEffectService.getFirstGroupWithAttributes(
-                      payload
+                      payload,
+                      currentGroupId?.startsWith('CONFLICT')
```
ChristophHi, line 277: Please add a test for updateConfigurationSuccess and current group as conflict group, I think this is not covered

ChristophHi, line 277: Please add comment that commerce will always have group ID starting with 'CONFLICT' for group type Configurator.GroupType CONFLICT_GROUP

feature-libs/product-configurator/rulebased/core/state/effects/configurator-basic-effect.service.spec.ts
```diff
@@ -68,6 +119,28 @@ describe('ConfiguratorBasicEffectService', () => {
...
+    });
+
+    it('should find attribute group as first group in multi level config although conflicts exist if includeConflicts is set to false', () => {
+      productConfiguration.groups = groupListWithConflicts;
+      expect(
+        classUnderTest.getFirstGroupWithAttributes(productConfiguration, false)
+      ).toBe(GROUP_ID_8);
+    });
+
+    it('should find attribute group as first group in multi level config although conflicts exist if includeConflicts is set to true', () => {
```
ChristophHi: Please fix description of test (you expect conflict group to be found)

ChristophHi (changes requested):

feature-libs/product-configurator/rulebased/core/state/effects/configurator-basic-effect.service.ts
steinsebastian: done

feature-libs/product-configurator/rulebased/core/state/effects/configurator-basic-effect.service.spec.ts
steinsebastian: Done.

feature-libs/product-configurator/rulebased/core/state/effects/configurator-basic.effect.ts
steinsebastian: Done

steinsebastian: done

feature-libs/product-configurator/rulebased/core/state/effects/configurator-basic-effect.service.ts
```diff
@@ -14,35 +14,62 @@ import { Configurator } from '../../model/configurator.model';
...
-    const groupWithAttributes: Configurator.Group | undefined = groups
-      .filter(
-        (currentGroup) =>
-          currentGroup.attributes && currentGroup.attributes.length > 0
-      )
-      .shift();
+    let groupWithAttributes: Configurator.Group | undefined;
+    if (
+      includeConflicts &&
+      groups[0].groupType === Configurator.GroupType.CONFLICT_HEADER_GROUP
```
ChristophHi, line 54: There is no check that groups contains at least one record

ChristophHi (changes requested):

feature-libs/product-configurator/rulebased/core/state/effects/configurator-basic-effect.service.ts
steinsebastian: done

ChristophHi (changes requested):
I did another round of testing. It seems that with your changes, on fixing the first conflict, the second one gets the current group. Was this intended?

steinsebastian:
@ChristophHi 
It is more or less intended.
The original issue was to avoid jumping from standard group to conflict group as soon as a conflict appeared.
After solving this, the behavior for setting values inside a conflict group was that even if the value change inside the conflict group was not solving the conflict it was jumping to the standard group.
That is because the group id changes with the new value assignment thus triggers a group change although for the user the conflict appears to be the same. (Only technically it is a new group not semantically).

To avoid this glitch it was now implemented that if the current group is a conflict group it should first search for existing conflict groups as new current group.
But as a side effect it would always find conflict groups as new current group as long as the user is doing value assignments inside a conflict group and other conflict groups exist.

ChristophHi (approved):

PR #16635, opened 2022-12-12
fix: storefrontlib sonar issues

RadhepS:
closes: https://jira.tools.sap/browse/CXSPA-1891


projects/storefrontlib/router/on-navigate.service.ts
```diff
@@ -123,4 +122,13 @@ export class OnNavigateService {
       currentRoute.routerEvent.urlAfterRedirects.split('?')[0]
     );
   }
+
+  /**
+   * Set history scroll restoration to manual if it is feature enabled
+   */
+  private setHistoryScrollRestoration() {
```
bgambocjaviniar, line 129: I'd say to just add the type (void)

but also, I do not see this in one of the warnings in sonar. I'll move it back todo so that you can find it and show it to me.

But I do see other thrown warnings that are not related to this found here:
- https://sonar.tools.sap/project/issues?resolved=false&branch=develop&id=composable-storefront&open=AYToK8USTnlbuCy4HmtF
- https://sonar.tools.sap/project/issues?resolved=false&branch=develop&id=composable-storefront&open=AYTn07QCTnlbuCy4Spwt

bgambocjaviniar (changes requested):

PR #16636, opened 2022-12-13
Fix/cxspa 2016 check for empty

ChristophHi:


feature-libs/product-configurator/rulebased/components/overview-form/configurator-overview-form.component.spec.ts
```diff
@@ -229,40 +234,77 @@ describe('ConfigurationOverviewFormComponent', () => {
...
+        overview: {
+          configId: ConfigurationTestData.CONFIG_ID,
+          productCode: ConfigurationTestData.PRODUCT_CODE,
+        },
+      };
+      expect(component.hasAttributes(configWOOverviewGroups)).toBe(false);
+    });
+
+    it('should detect that a configuration w/o groups that carry attributes does not provide OV attributes', () => {
+      initialize();
```
Uli-Tiger, line 259: does the name fit? - i would phrase it like this
should detect that a configuration **with** groups, **which do not carry any attributes,** does not provide OV attributes

**PS:** ok i know got it - the 'that' refers to groups and not to configuration
i leave it up to you, whether you rename it or not

Uli-Tiger (changes requested):

Uli-Tiger (approved):

PR #16642, opened 2022-12-13
refactor: Disable shipping address step if there is no delivery entries in cart

WeizhengSap:
ticket: CXSPA-1798


feature-libs/cart/base/core/facade/active-cart.service.spec.ts
```diff
@@ -799,4 +799,34 @@ describe('ActiveCartService', () => {
       });
     });
   });
+
+  describe('hasPickupItems and hasDeliveryItems', () => {
```
bgambocjaviniar, line 803: can we also get the test case when `pickupItemsQuantity` and `deliveryItemsQuantity` are undefined, which will test the false branch part.

I'm mentioning this now because we're lacking in branches for our test coverage in the cart lib. Threshold is at 75%, but we're currently at 80% 

Screenshot 2023-01-05 at 1 52 39 PM

feature-libs/checkout/base/components/guards/checkout-steps-set.guard.ts
```diff
@@ -19,22 +20,43 @@ import {
...
   constructor(
     protected checkoutStepService: CheckoutStepService,
     protected routingConfigService: RoutingConfigService,
     protected checkoutDeliveryAddressFacade: CheckoutDeliveryAddressFacade,
     protected checkoutPaymentFacade: CheckoutPaymentFacade,
     protected checkoutDeliveryModesFacade: CheckoutDeliveryModesFacade,
     protected router: Router
-  ) {}
+  ) {
+    this.activeCartFacade = inject(ActiveCartFacade);
```
bgambocjaviniar: Mentioned in our internal slack about opinions on going forward with this

feature-libs/checkout/base/components/guards/checkout-steps-set.guard.ts
```diff
@@ -134,4 +156,8 @@ export class CheckoutStepsSetGuard implements CanActivate {
       this.routingConfigService.getRouteConfig(routeName)?.paths?.[0] as string
     );
   }
+
+  ngOnDestroy(): void {
+    this.subscription.unsubscribe();
```
bgambocjaviniar, line 159:
```suggestion
    this.subscription?.unsubscribe();
```

feature-libs/cart/base/core/facade/active-cart.service.ts
```diff
@@ -527,6 +527,22 @@ export class ActiveCartService implements ActiveCartFacade, OnDestroy {
       .subscribe();
   }
 
+  hasPickupItems(): Observable<boolean> {
```
bgambocjaviniar, line 530: not being used, I'm assuming it will help bopis later?

feature-libs/checkout/base/components/guards/checkout-steps-set.guard.spec.ts
```diff
@@ -122,6 +130,18 @@ describe(`CheckoutStepsSetGuard`, () => {
     );
     checkoutDeliveryModesFacade = TestBed.inject(CheckoutDeliveryModesFacade);
     checkoutPaymentFacade = TestBed.inject(CheckoutPaymentFacade);
+    activeCartFacade = TestBed.inject(ActiveCartFacade);
+    checkoutStepService = TestBed.inject(CheckoutStepService);
+  });
+
+  describe('should be able to disable/enable delivery address step', () => {
```
bgambocjaviniar, line 136: nitpicking here 
```
describe('there is no delivery items set', () {
....
it('should disable delivery address ste', () {
```

bgambocjaviniar: should also add the same test case to the b2b unit test variant `checkout-b2b-steps-set.guard`

feature-libs/checkout/base/components/guards/checkout.guard.ts
```diff
@@ -19,7 +19,15 @@ import { ExpressCheckoutService } from '../services/express-checkout.service';
   providedIn: 'root',
 })
 export class CheckoutGuard implements CanActivate {
-  private readonly firstStep$: Observable<UrlTree>;
+  private readonly firstStep$: Observable<UrlTree> =
```
bgambocjaviniar, line 22: just wondering, but I'm fine with it to keep

2 questions:
1- the output is the same. Before it utilizes the 'helper' function to get the steps, but now you're composing it
2- why move it to the property init vs constructor init 
any reason to move it to property initialization vs constructor init?
Moreover, the changes in the specs?

again, don't mind to keep both, but just wondering

**If the answer is, you prefer this way, then I'm fine with it 👍**

feature-libs/checkout/base/components/checkout-delivery-mode/checkout-delivery-mode.component.ts
```diff
@@ -105,8 +98,11 @@ export class CheckoutDeliveryModeComponent implements OnInit, OnDestroy {
             this.mode.controls['deliveryModeId'].setValue(code);
             this.changeMode(code);
           }
-        })
-    );
+        }),
+        map(([deliveryModes]) =>
+          deliveryModes.filter((mode) => mode.code !== 'pickup')
```
bgambocjaviniar, line 103: I'm sure there are xyz different codes. What do about creating a model (enum) in checkout root to re-use for future extensibility. The enums can be something like DeliveryMode.PICKUP and DeliveryMode.DELIVERY. I'm not sure what the other code is, but maybe you can put that? 

What is the other code(s) though

bgambocjaviniar (changes requested):

bgambocjaviniar (changes requested):
Also big questions to this PR are:

1 - are you going to fix the side effects caused by disabling the delivery address step in this PR? 
---> when we just add a product for pickup, it disables it as intended, but an empty delivery mode
---> we get an empty mini cart number when that specific setup is  done
**---> when an actual product for delivery + pickup are in the cart, it will show delivery address, which is fine to me - but is that intentional?** 

I'm more curious about the 3rd point from `--->`

feature-libs/checkout/base/components/guards/checkout-steps-set.guard.ts
WeizhengSap: subscription is not optional

feature-libs/checkout/base/components/guards/checkout-steps-set.guard.ts
WeizhengSap: confirm with Bill, the refactor work will be in 6.0.

feature-libs/cart/base/core/facade/active-cart.service.ts
WeizhengSap: yes, will be used in bopis.

feature-libs/checkout/base/components/guards/checkout-steps-set.guard.spec.ts
WeizhengSap: It tests the codes in constructor. `checkout-b2b-steps-set.guard` call `super(...)` in its constructor. So, I think no need add the same test in b2b.

feature-libs/checkout/base/components/guards/checkout.guard.ts
WeizhengSap: 1. The output is not the same. Before it calls `this.checkoutStepService.getFirstCheckoutStepRoute()`, it read the first step ("delivery address") even it is disabled. Now, it uses `this.checkoutStepService.steps$`. 
2. https://stackoverflow.com/questions/51494083/initializing-variables-inline-during-declaration-vs-in-the-constructor-in-angula

The changes is in spec. (actually, the change is: `this.checkoutStepService.steps$` replace `this.checkoutStepService.getFirstCheckoutStepRoute()`)

feature-libs/checkout/base/components/checkout-delivery-mode/checkout-delivery-mode.component.ts
WeizhengSap: The other codes are "standard-gross" and "premium-gross". I don't think we can hard-code these codes in enum. It read from backend DB. But, "pickup" is a special one. :)

WeizhengSap:
1. "when we just add a product for pickup, it disables it as intended, but an empty delivery mode" -- the delivery mode will not be empty. BOPIS adds a new component in it. We have another ticket https://jira.tools.sap/browse/CXSPA-1799 for it.
2. mini cart is fixed in BOPIS branch
3. The delivery mode will look like this (but in this ticket, we focus on disable/enable shipping address step)
Screenshot 2023-01-05 at 11 08 09 AM

feature-libs/checkout/base/components/guards/checkout.guard.ts
bgambocjaviniar: 1. yup I did say that as well, but why I say they're the same is because isn't `getFirstCheckoutStepRoute` === `steps[0].routeName` === `this.checkoutStepService.steps$(**emitted** - steps[0].routeName)`
2. Yup, I understand, but in the end is just preference right? Just wanted to know why you changed it

feature-libs/checkout/base/components/guards/checkout-steps-set.guard.ts
bgambocjaviniar: unresolved this, and but I believe Nik shared one of my concerns on introducing it in a lib. https://sap-cx.slack.com/archives/C02L8BUATM5/p1672948206227529?thread_ts=1672945495.312019&cid=C02L8BUATM5

bgambocjaviniar (changes requested):
should be good to go after fixing unit test for `CheckoutStepsSetGuard should be able to disable/enable delivery address step should enable delivery address step FAILED`

bgambocjaviniar (approved):

feature-libs/checkout/base/components/guards/checkout.guard.ts
WeizhengSap: `steps[0].routeName` is not dynamically changed, but `steps$` is.

PR #16644, opened 2022-12-13
Fix: Customer list trap focus does not work properly - CXSPA 1887 

bullfroggy-sap:
Add config option for TrapFocusConfig to specify when up and down keyboard events should be ignored
Implement ignoreArrows config in trap-focus directive
Use ignoreArrows config in customer-list component


ercultimate (dismissed):

feature-libs/asm/components/customer-list/customer-list.component.ts
```diff
@@ -39,6 +39,7 @@ export class CustomerListComponent implements OnInit, OnDestroy {
     trap: true,
     block: true,
     autofocus: 'customer-list-selector',
+    ignoreArrows: true,
```
justinlee01: Can this be changed to `ignoreArrowKeys`?
Or, more precisely, `ignoreUpDownArrowKeys`?

projects/storefrontlib/layout/a11y/keyboard-focus/keyboard-focus.model.ts
```diff
@@ -113,6 +113,11 @@ export type TrapFocusType =
 export interface TrapFocusConfig extends TabFocusConfig {
   /** traps the focus */
   trap?: TrapFocusType;
+
+  /**
+   * Indicates that up and down arrow keyboard events should be ignored
```
justinlee01:
```suggestion
   * Indicates that up and down arrow keyboard events should be ignored. Works only if 'trap' is truthy.
```

justinlee01 (changes requested):

feature-libs/asm/components/customer-list/customer-list.component.ts
bullfroggy-sap: yea, ok

projects/storefrontlib/layout/a11y/keyboard-focus/keyboard-focus.model.ts
bullfroggy-sap: Ok

justinlee01 (approved):

ercultimate (approved):

plabadie (approved):

PR #16656, opened 2022-12-14
fix: Add customer isolation support

Matejk00:
Closes: CXSPA-1237


projects/core/src/auth/user-auth/services/auth-multisite-isolation.service.ts
```diff
@@ -0,0 +1,50 @@
...
+
+import { Injectable } from '@angular/core';
+import { take } from 'rxjs/operators';
+import { BaseSiteService } from '../../../site-context/facade/base-site.service';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class AuthMultisiteIsolationService {
+  protected static MULTISITE_SEPARATOR = '|';
```
Platonn: Nitpick: why `static`? 
I don't mind too much, but IMHO `readonly` would suffice. ;)

projects/core/src/auth/user-auth/services/auth-multisite-isolation.service.ts
```diff
@@ -0,0 +1,50 @@
...
+      .pipe(take(1))
+      .subscribe((baseSite) => {
+        if (baseSite?.isolated) {
+          baseSiteUid =
+            AuthMultisiteIsolationService.MULTISITE_SEPARATOR + baseSite?.uid;
+        }
+      });
+
+    return baseSiteUid;
+  }
```
Platonn: Would you mind changing the return type to an Observable, instead of a static value?

We're trying to obtain the observable's value synchronously. By coincidence it works, because the Observable returned from `baseSiteService.get()` emits a value synchronously. However, if in the future we change the logic of the `baseSiteService.get()` to emit asunchronoulsy (e.g. by piping: `debounceTime(0)` or `scheduler(asapScheduler)`, etc.), then the static variable `let baseSiteUid` would remain `undefined`.

(In general, it would be good not to mix observable and static logic, unless really justified)

---
Then in the Promise-based AuthService method we could call e.g.
```ts
someVariable = await this.authMultisiteIsolationService.method().toPromise()
```

projects/core/src/auth/user-auth/facade/auth.service.ts
```diff
@@ -87,10 +90,18 @@ export class AuthService {
    */
   async loginWithCredentials(userId: string, password: string): Promise<void> {
     try {
+      let uId: string = userId;
+
+      if (this.authMultisiteIsolationService?.getBaseSiteDecorator()) {
```
Platonn: Wouldn't it suffice to check just if the `this.authMultisiteIsolationService` is injected?
```ts
if(this.authMultisiteIsolationService){ ... }
```

projects/core/src/auth/user-auth/facade/auth.service.ts
```diff
@@ -87,10 +90,18 @@ export class AuthService {
...
+        uId =
+          userId + this.authMultisiteIsolationService?.getBaseSiteDecorator();
+      }
```
Platonn: It's great we moved a lot of logic to a separate service! This avoids unnecessary responsibilities in the AuthService.

IMHO the implementation detail of building the uid (that it's concatenated _in this order_: userId+decorator and e.g. not the other way around) should not be responsibility of the AuthService. Would you mind moving it to the `authMultisiteIsolationService` method too?

In other words, what about having a method:
```ts
this.authMultisiteIsolationService.getUserId(userId);
//or
this.authMultisiteIsolationService.decorateUserId(userId);
```
?

projects/core/src/auth/user-auth/facade/auth.service.ts
```diff
@@ -87,10 +90,18 @@ export class AuthService {
    */
   async loginWithCredentials(userId: string, password: string): Promise<void> {
     try {
+      let uId: string = userId;
```
Platonn, line 93: Do we _really_ need to wrap our new logic inside the `try catch` block? If not, let's move it up, outside of the block.

feature-libs/asm/root/services/asm-auth.service.ts
```diff
@@ -37,15 +38,17 @@ export class AsmAuthService extends AuthService {
     protected authStorageService: AsmAuthStorageService,
     protected authRedirectService: AuthRedirectService,
     protected globalMessageService: GlobalMessageService,
-    protected routingService: RoutingService
+    protected routingService: RoutingService,
+    protected authMultisiteIsolationService: AuthMultisiteIsolationService
```
Platonn: Let's make it an optional dep, not to introduce a breaking change.
```suggestion
    protected authMultisiteIsolationService?: AuthMultisiteIsolationService
```

btw. no need for `@Optional` decorator, because `AuthMultisiteIsolationService` is `providedIn: 'root'`

projects/core/src/auth/user-auth/facade/auth.service.ts
```diff
@@ -42,7 +43,9 @@ export class AuthService {
     protected oAuthLibWrapperService: OAuthLibWrapperService,
     protected authStorageService: AuthStorageService,
     protected authRedirectService: AuthRedirectService,
-    protected routingService: RoutingService
+    protected routingService: RoutingService,
+    @Optional()
```
Platonn: no need for `@Optional` decorator, because `AuthMultisiteIsolationService` is `providedIn: 'root'`

projects/storefrontapp/src/app/spartacus/spartacus-b2b-configuration.module.ts
```diff
@@ -16,9 +16,11 @@ import {
 } from '@spartacus/storefront';
 import { environment } from '../../environments/environment';
 
+const defaultBaseSite = ['powertools-spa', 'powertools-standalone'];
+
 const baseSite = environment.epdVisualization
-  ? ['powertools-epdvisualization-spa', 'powertools-spa']
-  : ['powertools-spa'];
+  ? ['electronics-epdvisualization-spa'].concat(defaultBaseSite)
+  : defaultBaseSite;
```
Platonn: Nitpick: what about:
```ts
let baseSite = ['powertools-spa', 'powertools-standalone'];

if(environment.epdVisualization) {
  baseSite.unshift('powertools-epdvisualization-spa')
}
```

projects/core/src/auth/user-auth/services/auth-multisite-isolation.service.ts
```diff
@@ -0,0 +1,50 @@
...
+   * The strategy how to handle isolation is to use an additional decorator
+   * passed as a suffix to the uid field.
+   *
+   * Example uid value for baseSiteA will be email@example.com|baseSiteA
+   *
+   * This method checks if currently active baseSite is "isolated". If so,
+   * it returns specific `uid` suffix.
+   *
+   * Example: `|electronics-standalone`.
+   */
```
Platonn: Great comment, very informative.

btw. if we decide to change this method to return the already-concatenated userId with the decorator, then I think we can skip this part:

```
   * This method checks if currently active baseSite is "isolated". If so,
   * it returns specific `uid` suffix.
   *
   * Example: `|electronics-standalone`.
```

But again, very nice TSDoc!

projects/storefrontapp-e2e-cypress/cypress/integration/regression/multisite-isolation/multisite-isolation.core-e2e-spec.ts
```diff
@@ -0,0 +1,109 @@
...
+
+      cy.cxConfig({
+        context: {
+          baseSite: [ELECTRONICS_STANDALONE_BASESITE],
+          currency: [ELECTRONICS_CURRENCY],
+        },
+      });
+    });
+
+    it.skip('should register customer on standalone environment', () => {
```
Platonn, line 39: Are tests in this suite `skip`ped deliberately?

projects/storefrontapp-e2e-cypress/cypress/integration/regression/multisite-isolation/multisite-isolation.core-e2e-spec.ts
```diff
@@ -0,0 +1,109 @@
...
+      cy.visit('/login/register');
+      register(user);
```
Platonn, line 41: Do we _really_ want to "register" an user credentials that is shared among possibly more test suites (that may also want to possibly register this user)?

Although maybe everything works fine, the consequences of using a shared user credentials are not clear at the first sight. For clarity, what about creating fresh user credentials that are shared only in the scope of this test suite?

projects/core/src/auth/user-auth/facade/auth.service.ts
```diff
@@ -87,10 +90,18 @@ export class AuthService {
    */
   async loginWithCredentials(userId: string, password: string): Promise<void> {
     try {
+      let uId: string = userId;
+
+      if (this.authMultisiteIsolationService?.getBaseSiteDecorator()) {
+        uId =
+          userId + this.authMultisiteIsolationService?.getBaseSiteDecorator();
+      }
+
```
Platonn: Do we _really_ need to wrap our whole new logic inside the `try catch` block? If not, let's move it up, outside of the block.

projects/core/src/auth/user-auth/services/auth-multisite-isolation.service.ts
Matejk00: Thanks, I changed to `readonly`.

projects/core/src/auth/user-auth/services/auth-multisite-isolation.service.ts
Matejk00: Thank you for clarification, that makes sense, I switched to `Observable` then.

projects/core/src/auth/user-auth/facade/auth.service.ts
Matejk00: I think service can be injected, but if site is non isolated I would like to do nothing. So that's why I'm checking if the method returns any value.

projects/core/src/auth/user-auth/facade/auth.service.ts
Matejk00: Cool, I added separated method.

feature-libs/asm/root/services/asm-auth.service.ts
Matejk00: Changed.

projects/core/src/auth/user-auth/facade/auth.service.ts
Matejk00: Changed.

projects/storefrontapp/src/app/spartacus/spartacus-b2b-configuration.module.ts
Matejk00: LGTM

projects/core/src/auth/user-auth/services/auth-multisite-isolation.service.ts
Matejk00: I removed the last part.

projects/storefrontapp-e2e-cypress/cypress/integration/regression/multisite-isolation/multisite-isolation.core-e2e-spec.ts
Matejk00: It requires change in CI sample data to working properly, so that's why I decided to skip them for now.
I have a ticket in progress: CXSPA-2035 related to this issue.

projects/storefrontapp-e2e-cypress/cypress/integration/regression/multisite-isolation/multisite-isolation.core-e2e-spec.ts
Matejk00: For check authentication we need first to create a user on specific standalone basesite. I am not sure if we can use pre-defined for testing such cases (users for electronics-spa and powertools-spa won't work). I need to investigate.

As far as I checked `uid` is shared through all scenarios I prepared in the test suite.

projects/core/src/auth/user-auth/facade/auth.service.ts
Matejk00: I think is not necessary to have it in try/catch.

Platonn (dismissed):

Platonn (approved):

projects/core/src/auth/user-auth/facade/auth.service.ts
```diff
@@ -86,11 +88,20 @@ export class AuthService {
    * @param password
    */
   async loginWithCredentials(userId: string, password: string): Promise<void> {
+    let uid = userId;
+
+    if (this.authMultisiteIsolationService) {
+      uid = await this.authMultisiteIsolationService
```
Platonn, line 94: nitpick: we could just simply overwrite the `userId`

projects/core/src/auth/user-auth/services/auth-multisite-isolation.service.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+}
+
+describe('AuthMultisiteIsolationService', () => {
+  let service: AuthMultisiteIsolationService;
+  let baseSiteService: BaseSiteService;
+
+  beforeEach(() => {
+    TestBed.configureTestingModule({
+      providers: [
+        AuthMultisiteIsolationService,
```
Platonn, line 27: nitpick: this service is providedIn: root. so no need to provide it explicitly here

projects/core/src/auth/user-auth/services/auth-multisite-isolation.service.ts
```diff
@@ -0,0 +1,52 @@
...
+   * can only log into baseSiteA, not baseSiteB.
+   * To login into baseSiteB customer have to use the same e-mail and register an account
+   * on baseSiteB.
+   *
+   * The strategy how to handle isolation is to use an additional decorator
+   * passed as a suffix to the uid field.
+   *
+   * Example uid value for baseSiteA will be email@example.com|baseSiteA
+   */
+  getBaseSiteDecorator(): Observable<string> {
```
Platonn, line 31: nitpick: can be protected. but then unit tests need adjusting

PR #16658, opened 2022-12-14
feat: configurable timeouts for backend API calls

### LINKED ISSUE #14559: Investigate ways to avoid hanging SSR renders (or to recover from them)

Platonn:
This PR introduces a mechanism of configurable timeout for backend HTTP requests - can be configured different per platform (server/browser). It can be configured via the global config `config.backend.timeout = { browser: ..., server: ...} `. Or it can be configured more specifically - per concrete request by passing a `HTTP_TIMEOUT_CONFIG` `HttpContextToken` to the Angular `HttpClient`'s method.

Context:
Timing out outgoing http calls that take too long is especially important improvement in SSR (Node.js), because in Node.js there is no default timeout of outgoing http calls (in contrary to browsers, which usually timeout long http calls after long time, e.g. 1 minute).

This configurable timeout logic is now implemented in Spartacus, in the layer of Angular Http Interceptor. That said, prior to this feature being released, customers can implement similar logic themselves, e.g. by implementing an Angular Http Interceptor themselves. 

Notes:
- by default the mechanism is disabled, not to introduce a behavioral breaking change in a minor release 5.x

closes #14559
closes https://jira.tools.sap/browse/CXSPA-529

---
**Environments for testing:**
- SSR (`yarn dev:ssr` AND use `curl` in another terminal  e.g.: `curl http://localhost:4200`)
- browser (`yarn start`)

**How can be tested:**
- configure some timeout in the global spartacus config, e.g. (uncomment appropriate:)
    ```ts
    provideConfig({ 
      backend: { 
        timeout: { 
          server: 3_000,
          browser: 3_000
        }
      }
    })
    ```

- Artificially delay the backend responses, e.g. with a custom local proxy:
  - Install `http-proxy` tool
  ```
  npm install -g http-proxy
  ```

  - Write a `http-proxy.js` server file   
  ```js
  // http-proxy.js

  // SPIKE TODO: DON'T MERGE IT

  // PROXY SERVER FOR DELAYING SOME OCC BACKEND RESPONSES
  // ONLY FOR TESTING PURPOSES

  const httpProxy = require('http-proxy');
  const http = require('http');

  const proxy = httpProxy.createProxyServer({ secure: false });

  const ENDPOINT_FOR_DELAY = 'consenttemplates';
  const BACKEND_BASE_URL = 'https://40.76.109.9:9002';
  const DELAY = 3000; // <-------------------------- PLEASE MANIPULATE IT: e.g.  2_500 or 3_000 

  /** custom predicate, whether we should delay a request */
  const shouldDelay = (req) => {
    // Note: In browser there are 2 requests: preflight (OPTIONS) and actual request (GET).

    const result = req.url.includes(ENDPOINT_FOR_DELAY);
    result && console.log({ delay: DELAY, url: req.url, method: req.method });
    return result;
  };

  http
    .createServer(function (req, res) {
      const forwardRequest = () =>
        proxy.web(req, res, { target: BACKEND_BASE_URL });
      const delay = shouldDelay(req) ? DELAY : 0;
      setTimeout(forwardRequest, delay);
    })
    .listen(9002);
  ```

  - Execute the server in a terminal:   
  ```
  node http-proxy.js
  ```
  
- set the OCC backend url to the `http://localhost:9002` (note: NOT HTTPS!) in `.env-cmdrc` file:
```
  "dev": {
    "CX_BASE_URL": "http://localhost:9002"
  },
```


projects/core/src/http/http-timeout/http-timeout.interceptor.ts
```diff
@@ -0,0 +1,113 @@
...
+} from '@angular/common/http';
+import { Injectable } from '@angular/core';
+import { NEVER, Observable, of, TimeoutError } from 'rxjs';
+import { catchError, startWith, switchMap, timeout } from 'rxjs/operators';
+import { OccConfig } from '../../occ/config/occ-config';
+import { WindowRef } from '../../window/window-ref';
+import { HTTP_TIMEOUT_CONFIG } from './http-timeout.config';
+
+/**
+ * Timeouts requests that take longer than the specified timeout, by throwing
```
Zeyber: This reads a little strange. Is this what you mean?
`Timeouts requests that take longer than the specified timeout throw an error of type HttpErrorResponse with http code 504 (Gateway Timeout).`

Zeyber (changes requested):

projects/core/src/http/http-timeout/http-timeout.interceptor.ts
Platonn: I'm sorry, I might be using the `timeout` incorrectly in English.
Sometimes I use it as noun (the expected max time) and sometimes as verb (to kill something after specified max time).
Please let me know how I can improve. I'm happy to learn.

PS. There is a leftover in this sentence (from past version of the draft PR, where I returned 504 http status code in the error object; but I reverted it).

Zeyber: Hmm... I think you might be looking for the verb `Times out` in this instance?

projects/core/src/http/http-timeout/http-timeout.interceptor.ts
```diff
@@ -0,0 +1,112 @@
...
+} from '@angular/common/http';
+import { Injectable } from '@angular/core';
+import { NEVER, Observable, of, TimeoutError } from 'rxjs';
+import { catchError, startWith, switchMap, timeout } from 'rxjs/operators';
+import { OccConfig } from '../../occ/config/occ-config';
+import { WindowRef } from '../../window/window-ref';
+import { HTTP_TIMEOUT_CONFIG } from './http-timeout.config';
+
+/**
+ * It throws an error when a request takes longer than the specified time.
```
Zeyber, line 23: I think the previous explanation was better here (ie. specifying gateway timeout and status code).

Platonn: I no longer return artificial status code nor the statusText, but only the Error object. See the [recent commit](https://github.com/SAP/spartacus/pull/16658/commits/ac8216e28aac30287476e2d6957b9802658bb7a2)

Zeyber (approved):

Platonn:
Here's a doc input, which should be edited and moved later to the official documentation of Spartacus:

-------

# Configurable timeouts of backend API calls
Since Spartacus ~v5.2~ v6.0 the timeout time values for outgoing http calls can be configured via the Spartacus configuration.
​
The timeout value can be configured globally for all http calls, or differently for each http call. Moreover, the timeout value can be configured different for when Spartacus is running in the browser vs. running in the server. 
​
That said, **the most important thing is to configure _any_ timeout for outgoing http requests, when Spartacus is running in the server (SSR)**. Otherwise, the server might get out of stability, when the backend API responses arrive very late or never (e.g. due to temporary networking issues).
​
## How to configure backend API timeouts globally (for browser and server runtime respectively)
```typescript
provideConfig({
  backend: {
    timeout: {
      // values in milliseconds:
      browser: 60_000
      server: 10_000
    }
  }
})
```
Note: it's most important to configure the timeout for the **server** runtime, because browsers often already have a built-in timeout mechanism for long running http requests.
​
## How to configure a backend API timeouts when making a specific http call
Configuring a global http timeouts once for all requests via the main Spartacus config should suffice for most users. But if you need to configure a specific timeout only for certain http requests, you can define it in the special `HttpContext` token of Spartacus: `HTTP_TIMEOUT_CONFIG` (from `@spartacus/core`) and pass it directly to the `HttpClient`, when making a call. See the following example below:

```typescript
import { HTTP_TIMEOUT_CONFIG, HttpTimeoutConfig } from `@spartacus/core`;
​
/* ... */
​
​
return this.httpClient.get('/some/api', {
  context: new HttpContext().set(
    HTTP_TIMEOUT_CONFIG, 
    { server: 15_000 } // value in milliseconds 
  )
})
```
​
## Timed out requests print a warning to the console
When a request times out, the following warning is logged to the console:
```
Request to URL '${request.url}' exceeded expected time of ${timeoutValue}ms and was aborted.
```
​
## Why configuring a timeout for outgoing http requests in `server` runtime is crucial for stability of the Server-Side-Rendering
The calls to backend APIs sometimes might take a long time to complete. This can be caused by a number of reasons, such as slow network or slow backend. Browsers by default time out such long http calls, e.g. after 1 minute. But in the server runtime `NodeJS`, no timeout is not set by default for outgoing http requests by the `NodeJS` platform.
​
Therefore, it's crucial for the stability for the Server-Side-Rendered applications running in `NodeJS` to set an explicit timeout for each outgoing http call. Otherwise, if a backend API responds very slow (or never), the Server-Side-Rendered application will keep waiting for the response very long (or forever). In that case, the memory allocated for this application won't be released, which is a memory leak by itself.
​
## Future outlook:  default http timeout in `server` runtime
Currently, the backend API timeouts need to be configured explicitly in Spartacus. In the future major version we plan to introduce a default timeout value for the server runtime, so that the developers don't need to configure it explicitly.
​
## Spartacus versions ~<5.2~ <6.0
​
The configurable http timeouts were introduced only in Spartacus ~v5.2~ v6.0. If you're using older Spartacus version, you need to implement a `HttpTimeoutInterceptor` yourself and provide it in your application. 
​
Feel free to copy-paste the original implementation from the ~v5.2~ v6.0 to your app (and possibly adapt the timeout values to your needs):
https://github.com/SAP/spartacus/blob/907d7897dba6add3ce2b56aa194f71596b9afb77/projects/core/src/http/http-timeout/http-timeout.interceptor.ts
​
Then provide the copied (and adapted) interceptor in your app (e.g. in your app.module):
​
```typescript
providers: [
  {
    provide: HTTP_INTERCEPTORS,
    useClass: HttpTimeoutInterceptor,
    multi: true,
  },
]
```

PR #16679, opened 2022-12-19
Feature/cxec 17126

steinsebastian:


feature-libs/product-configurator/rulebased/components/tab-bar/configurator-tab-bar.component.ts
```diff
@@ -44,6 +52,37 @@ export class ConfiguratorTabBarComponent {
     )
   );
 
+  getTabIndexConfigTab(): number {
```
ChristophHi, line 60: Please add JS doc for new public methods

feature-libs/product-configurator/rulebased/components/tab-bar/configurator-tab-bar.component.ts
```diff
@@ -44,6 +52,37 @@ export class ConfiguratorTabBarComponent {
...
+      if (!isOvPage) {
+        tabIndex = -1;
+      }
+    });
+    return tabIndex;
+  }
+
+  switchTabOnArrowPress(event: KeyboardEvent, currentTab: string): void {
+    if (event.code === 'ArrowLeft' || event.code === 'ArrowRight') {
+      event.preventDefault();
```
ChristophHi, line 92: Why is that needed? I can see no change in behaviour if that is omitted (or is there a default handling for ArrowLeft and ArrowRight I am not aware of?)

ChristophHi (changes requested):

feature-libs/product-configurator/rulebased/components/tab-bar/configurator-tab-bar.component.ts
steinsebastian: done

feature-libs/product-configurator/rulebased/components/tab-bar/configurator-tab-bar.component.ts
steinsebastian: It is necessary to restrict the keypress to ArrowLeft and ArrowRight.
If it is omitted then any keypress will lead to focus change between the tabs (e.g. if you press any letter-key like "g", etc.).
But here only ArrowLeft and ArrowRight should trigger the focus change.

ChristophHi (approved):

PR #16686, opened 2022-12-21
feat: SSR - log the options upon initialization

znikola:
closes CXSPA-2107

### DIFF
--- a/core-libs/setup/ssr/optimized-engine/optimized-ssr-engine.spec.ts
+++ b/core-libs/setup/ssr/optimized-engine/optimized-ssr-engine.spec.ts
@@ -111,6 +111,34 @@ const getCurrentConcurrency = (
 };
 
 describe('OptimizedSsrEngine', () => {
+  describe('logOptions', () => {
+    it('should log the provided options', () => {
+      new TestEngineRunner({
+        timeout: 50,
+        renderingStrategyResolver: () => RenderingStrategy.ALWAYS_SSR,
+      });
+
+      expect(console.log).toMatchInlineSnapshot(`
+        [MockFunction] {
+          "calls": [
+            [
+              "[spartacus] SSR optimization engine initialized with the following options: {
+          "timeout": 50,
+          "renderingStrategyResolver": "() => ssr_optimization_options_1.RenderingStrategy.ALWAYS_SSR"
+        }",
+            ],
+          ],
+          "results": [
+            {
+              "type": "return",
+              "value": undefined,
+            },
+          ],
+        }
+      `);
+    });
+  });
+
   describe('timeout option', () => {
     it('should fallback to CSR if rendering exceeds timeout', fakeAsync(() => {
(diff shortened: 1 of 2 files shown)


core-libs/setup/ssr/optimized-engine/optimized-ssr-engine.ts
```diff
@@ -59,7 +59,13 @@ export class OptimizedSsrEngine {
   constructor(
     protected expressEngine: NgExpressEngineInstance,
     protected ssrOptions?: SsrOptimizationOptions
-  ) {}
+  ) {
+    this.log(
```
Platonn, line 79: I guess, we want to pass `false` as a second argument?
Otherwise the options are not logged, unless you configure `debug:true` option.

znikola: I believe it should be logged only when `debug` is set to `true` :)

core-libs/setup/ssr/optimized-engine/optimized-ssr-engine.ts
```diff
@@ -59,7 +59,28 @@ export class OptimizedSsrEngine {
   constructor(
     protected expressEngine: NgExpressEngineInstance,
     protected ssrOptions?: SsrOptimizationOptions
-  ) {}
+  ) {
+    this.logOptions();
+  }
+
+  protected logOptions(): void {
```
pawelfras, line 66: `query:` thinking out loud - maybe it's worth considering covering such a method by unit tests.  I know that it's only for logging options, nevertheless, its behaviour depends on some conditions and maybe it's worth covering it in terms of any refactoring in the future.

pawelfras (dismissed):
Everything works as expected 👌. Just one comment to consider regarding the addition of tests 🙂

core-libs/setup/ssr/optimized-engine/optimized-ssr-engine.spec.ts
```diff
@@ -17,6 +17,7 @@ const host = 'my.shop.com';
 jest.mock('fs', () => ({
   readFileSync: () => '',
 }));
+jest.spyOn(console, 'log').mockImplementation();
```
pawelfras, line 20: cool move 😎 👌

core-libs/setup/ssr/optimized-engine/optimized-ssr-engine.spec.ts
```diff
@@ -110,22 +111,44 @@ const getCurrentConcurrency = (
...
 
 describe('OptimizedSsrEngine', () => {
+  describe('logOptions', () => {
+    it('should log the provided options', () => {
+      new TestEngineRunner({
+        timeout: 50,
+        renderingStrategyResolver: () => RenderingStrategy.ALWAYS_SSR,
+      });
+
+      expect(console.log).toMatchInlineSnapshot(`
```
pawelfras, line 121: 💪

pawelfras (approved):

PR #16694, opened 2022-12-27
fix: unit order feature not working when installed using libraries

RadhepS:
closes: https://jira.tools.sap/browse/CXSPA-2116


szczepaniak-michal (dismissed):

feature-libs/organization/schematics/add-organization/schema.json
```diff
@@ -29,15 +29,17 @@
           "Administration",
           "Order-Approval",
           "Account-Summary",
-          "Organization-User-Registration"
+          "Organization-User-Registration",
+          "Unit-Order"
```
bgambocjaviniar, line 33: 👍

feature-libs/organization/unit-order/core/unit-order-core.module.ts
```diff
@@ -25,10 +23,6 @@ export class UnitOrderCoreModule {
           useExisting: UnitOrderService,
         },
         UnitOrderConnector,
-        {
```
bgambocjaviniar, line 28: 👍

feature-libs/organization/unit-order/core/model/index.ts
```diff
@@ -4,5 +4,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+export * from './augmented-core.model';
```
bgambocjaviniar: move `export interface OrderHistoryQueryParams {` from `augmented-core.model.ts` to the empty file at the same level `unit-order.model.ts` 

and then export that instead of this augmented-core.model

bgambocjaviniar (changes requested):
going to stop thumbings up a lot of things, but change the model reference

looking and testing manually still, but that's what I see upfront right now in the PR

feature-libs/organization/unit-order/components/ng-package.json
```diff
@@ -0,0 +1,6 @@
+{
+  "$schema": "../../../node_modules/ng-packagr/ng-package.schema.json",
```
bgambocjaviniar: need to go up one more folder 'level'

```suggestion
  "$schema": "../../../../node_modules/ng-packagr/ng-package.schema.json",
```

feature-libs/organization/unit-order/components/unit-level-order-history/filter/unit-level-order-history-filter.component.spec.ts
```diff
@@ -16,7 +16,7 @@ import { ReactiveFormsModule } from '@angular/forms';
 import { By } from '@angular/platform-browser';
 import { PaginationModel } from '@spartacus/core';
 import { ICON_TYPE } from '@spartacus/storefront';
-import { OrderHistoryQueryParams } from '../../../core/model/augmented-core.model';
+import { OrderHistoryQueryParams } from '@spartacus/organization/unit-order/core';
```
bgambocjaviniar, line 19: future reference, the specs doesn't matter in terms of absolute path or relative, but 👍

feature-libs/organization/unit-order/components/unit-order-components.module.ts
```diff
@@ -6,6 +6,8 @@
 
 import { NgModule } from '@angular/core';
 import { RouterModule } from '@angular/router';
+import { UnitOrderDetailsOrderEntriesContextToken } from '@spartacus/organization/unit-order/root';
+import { UnitOrderDetailsOrderEntriesContext } from './page-context';
```
bgambocjaviniar: instead of delegating it to the index to export it all, let's specify it to `./page-context/unit-order-details-order-entries.context'

reason being:
- I recall when things are in a barrel file (index.ts), we get some compilation issues when it's not specified instead, or if it's not ./page-context/index. But I rather specify it now.

feature-libs/organization/unit-order/core/ng-package.json
```diff
@@ -0,0 +1,6 @@
+{
+  "$schema": "../../../node_modules/ng-packagr/ng-package.schema.json",
```
bgambocjaviniar: need to go up one more folder 'level'

```suggestion
  "$schema": "../../../../node_modules/ng-packagr/ng-package.schema.json",
```

projects/schematics/src/shared/lib-configs/organization-schematics-config.ts
```diff
@@ -163,6 +167,48 @@ export const ORGANIZATION_USER_REGISTRATION_SCHEMATICS_CONFIG: SchematicConfig =
...
+  i18n: {
+    resources: ORGANIZATION_UNIT_ORDER_TRANSLATIONS,
+    chunks: ORGANIZATION_UNIT_ORDER_TRANSLATION_CHUNKS_CONFIG,
+    importPath: SPARTACUS_ORGANIZATION_UNIT_ORDER_ASSETS,
+  },
+  styles: {
+    scssFileName: ORGANIZATION_SCSS_FILE_NAME,
+    importStyle: SPARTACUS_ORGANIZATION,
+  },
+  dependencyFeatures: [USER_PROFILE_FEATURE_NAME],
```
bgambocjaviniar: this feature depends on orders as well

How?
- you look at the @spartacus reference in the lib

```suggestion
  dependencyFeatures: [USER_PROFILE_FEATURE_NAME, ORDER_FEATURE_NAME],
```

bgambocjaviniar: also need to update the snaps I believe to make sure these files gets generated

bgambocjaviniar (changes requested):

projects/schematics/src/shared/lib-configs/organization-schematics-config.ts
bgambocjaviniar: add a test case for unit-order in `feature-libs/organization/schematics/add-organization/index_spec.ts`

bgambocjaviniar: unresolved this in case you missed above @RadhepS

feature-libs/organization/schematics/add-organization/index_spec.ts
```diff
@@ -420,4 +427,69 @@ describe('Spartacus Organization schematics: ng-add', () => {
       });
     });
   });
+
```
bgambocjaviniar, line 433: for the sake of 'completion', put an assertion for the unit order feature module after line 140

bgambocjaviniar:
for anyone wondering, above has been tested with a test branch called `test-5.1.0`. The reason why we had to check in another branch is because develop is currently broken, which fails to build the schematics of each libs + main schematics itself.

It won't work on this branch, but it is working on that test branch, which does not include the broken changes to schematics.

bgambocjaviniar (approved):

PR #16701, opened 2022-12-29
feat: CXSPA-2132 Translation - fallback keys

pawelfras:
Closes https://jira.tools.sap/browse/CXSPA-2132


docs/migration/6_0/html.md
```diff
@@ -73,3 +73,60 @@
...
+
+  to
+  
+  ```ts
+    translate(key: string | string[], options?: any, whitespaceUntilLoaded?: boolean): Observable<string>;
+  ```
+
+### TranslationPipe
+
+- Method 'translate' changed from
```
Platonn: typo. Pipe has a `transfrom` method
```suggestion
- Method 'transform' changed from
```

docs/migration/6_0/html.md
```diff
@@ -73,3 +73,60 @@
...
 - Changed `inline: true` to `inlineRoot: true` for keyboard tabbing and VO to work correctly.
+
+### ParagraphComponent
+
+- The `handleClick()` method now uses the condition `documentHost === element.host` to recognise external links.
+- The `handleClick()` method now uses `router.navigateByUrl()` to navigate internal links.
+
+### CloseAccoutModalComponent
+
+- The `onSuccess()` method now uses `authService.coreLogout()` to log user out before routing to homepage.
```
Platonn: Are those 2 classes really related to this PR? or are copy-paste leftovers?

btw. those 2 paragraphs contain 2 typos: `CloseAccoutModalComponent` and `recognise`

projects/core/src/i18n/i18next/i18next-translation.service.ts
```diff
@@ -73,16 +91,23 @@ export class I18nextTranslationService implements TranslationService {
...
   /**
    * Returns a fallback value in case when the given key is missing
-   * @param key
+   * @param keys
    */
-  protected getFallbackValue(key: string): string {
-    return isDevMode() ? `[${key}]` : this.NON_BREAKING_SPACE;
+  protected getFallbackValue(keys: string[]): string {
+    return isDevMode()
+      ? keys.map((key) => `[${key}]`).join(this.NON_BREAKING_SPACE)
```
Platonn: What about simply separate keys with `,`  _inside_ one pair of brackets? By example:
```
[common:helloWorld, common:hello, common:hi]
```
instead of
```
[common:helloWorld] [common:hello] [common:hi]
                   ^ --- nbsp --- ^
```

projects/core/src/i18n/translatable.ts
```diff
@@ -5,11 +5,15 @@
...
   params?: TranslatableParams;
   raw?: string;
 }
 
 export interface TranslatableParams {
   [param: string]: any;
 }
+
+export function isTranslatable(input: any): input is Translatable {
+  return typeof input !== 'string' && ('key' in input || 'raw' in input);
```
Platonn: Why do we check `typeof input !== string`? in the type guard for `Translatable`?
Shouldn't we check it somewhere else (outside `isTranslatable()`)?

projects/core/src/i18n/testing/mock-translate.ts
```diff
@@ -4,10 +4,14 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-export function mockTranslate(key: string | undefined, options: any = {}) {
+export function mockTranslate(
+  key: string | string[] | undefined,
+  options: any = {}
+) {
+  const keyString = Array.isArray(key) ? key[0] : key;
```
Platonn: This function is a simple but comprehensive and bullet-proof util for testing to be used by many libs. To keept it so, what about printing all keys:
```suggestion
  const keyString = Array.isArray(key) ? key.join(',') : key;
```
?

(I didn't include space after comma deliberately, because other options are separated with spaces)

Example output:
```
common:helloWorld,common:hello,common:hi firstname:Pawel surname:Fras
```

projects/core/src/i18n/i18next/i18next-translation.service.ts
```diff
@@ -36,28 +36,46 @@ export class I18nextTranslationService implements TranslationService {
     // Otherwise, we the will trigger additional deferred change detection in a view that consumes the returned observable,
     // which together with `switchMap` operator may lead to an infinite loop.
 
-    const chunkName = this.translationChunk.getChunkNameForKey(key);
-    const namespacedKey = this.getNamespacedKey(key, chunkName);
+    const translationKeys = Array.isArray(key) ? key : [key];
+
+    const chunkNamesByKeys: Map<string, string> = new Map();
+    const chunksNames: string[] = [];
```
Platonn: nitpick: Can't it be retrieved from `chunkNamesByKeys.values()`?

projects/core/src/i18n/i18next/i18next-translation.service.ts
```diff
@@ -73,16 +91,23 @@ export class I18nextTranslationService implements TranslationService {
...
+      : this.NON_BREAKING_SPACE;
   }
 
-  private reportMissingKey(key: string, chunkName: string) {
+  private reportMissingKey(chunkNamesByKeys: Map<string, string>) {
+    const keys = [...chunkNamesByKeys.keys()];
     if (isDevMode()) {
       console.warn(
-        `Translation key missing '${key}' in the chunk '${chunkName}'`
+        `Translation keys missing [${keys.join(', ')}]. Attempt to load ${[
```
Platonn:
```suggestion
        `Translation keys missing [${keys.join(', ')}]. Attempted to load ${[
```

Platonn:
It would be good to also document the new feature in the official docs of i18n of Spartacus: https://help.sap.com/docs/SAP_COMMERCE_COMPOSABLE_STOREFRONT/eaef8c61b6d9477daf75bff9ac1b7eb4/775e61ed219c4999852d43be5244e94a.html?q=i18n
e.g.

### Fallback keys
This feature allows for providing multiple variations of a key or multiple keys that may be used interchangeably. In case the primary key is not available, the fallback keys will be attempted in the order they are provided. The first key that can be resolved will be used for the translation.

```
< example(s) of usage here >
```

Note: When using lazy-loading of translation chunks, the function will first load all the chunks for the corresponding keys before attempting to resolve the value. This ensures that all the necessary data is available before the translation is performed.

projects/core/src/i18n/translation.service.ts
```diff
@@ -12,12 +12,12 @@ export abstract class TranslationService {
    * If key is missing, it tries to load the chunk and emits a value when chunk is loaded.
    * If key is missing after loaded chunk, a fallback value is emitted
```
Platonn: Let's change this JSDoc.

> It accepts also an array of keys. It allows for providing multiple variations of a key or multiple keys that may be used interchangeably as fallback. In case the primary key is not available, the fallback keys will be attempted in the order they are provided. The first key that can be resolved will be used for the translation.
> 
> Note: When using lazy-loading of translation chunks, it will first load all the chunks for the corresponding keys before attempting to resolve the value. This ensures that all the necessary data is available before the translation is performed.

---- 

Btw. it would be good to also to add a new paragraph "Fallback Keys" to the [official i18n docs of spartacus](https://help.sap.com/docs/SAP_COMMERCE_COMPOSABLE_STOREFRONT/eaef8c61b6d9477daf75bff9ac1b7eb4/775e61ed219c4999852d43be5244e94a.html?q=i18n), with the information above + code examples of using the fallback keys array.

docs/migration/6_0/html.md
pawelfras: Yeah, they are copy-paste leftovers. I'll remove them from this file but also fix typos where they can be found.

projects/core/src/i18n/i18next/i18next-translation.service.ts
pawelfras: Good point, thank you!

projects/core/src/i18n/i18next/i18next-translation.service.ts
pawelfras: Yes, I agree. Looks better that way 👌

projects/core/src/i18n/testing/mock-translate.ts
pawelfras: Sure, good catch!

projects/core/src/i18n/translatable.ts
pawelfras: Yes, you are right. I left that check because I had in my mind where to use that guard. Since I wanted to distinguish the value from the string with it, I added this condition to the middle of the guard. In the bigger picture, it doesn't make sense. I should rather check if `typeof input === 'object' && ('key' in input || 'raw' in input)` (typeof check has to be here, otherwise there will be an error `Cannot use 'in' operator to search for 'key' in input` object)

projects/core/src/i18n/translation.service.ts
pawelfras: Very good idea! I'll have that in mind and I'll contact Matthew in that matter.

projects/core/src/i18n/testing/mock-translate.ts
pawelfras: I've also added 2 tests covering the scenario array of keys. to `mock-translate.spec.ts`

projects/core/src/i18n/i18next/i18next-translation.service.spec.ts
```diff
@@ -49,142 +68,156 @@ describe('I18nextTranslationService', () => {
...
+          beforeEach(() => {
+            spyOn(i18next, 'exists').and.returnValue(true);
+          });
+
+          it('should emit result of i18next.t', () => {
+            spyOn(i18next, 't').and.returnValue('value');
+            let result;
+            service
+              .translate(key, testOptions)
+              .pipe(take(namespacedKeys.length))
```
Platonn: Do we really want to `take` more than 1 emission from this service?

I believe it should emit just once, regardless the size of the fallback keys array.

projects/core/src/i18n/i18next/i18next-translation.service.ts
```diff
@@ -36,28 +36,46 @@ export class I18nextTranslationService implements TranslationService {
...
+              subscriber.next(this.getFallbackValue(namespacedKeys));
             }
-          });
+          }
+        );
+
+        // if namespaces are not loaded yet, we can emit a non-breaking space
+        if (!namespacesLoaded && whitespaceUntilLoaded) {
+          subscriber.next(this.NON_BREAKING_SPACE);
         }
```
Platonn, line 79: Summary of the refactor, for future reference:

## Purpose
This refactor should not change the behavior seen by an external observer of this Observable.
It allowed us to more eaisly **introduce fallback keys** and **preload namespaces for all the fallback keys upfront**

## Comparison

### Before
- let's check if a key can be resolved (aka "if exists"). if yes then let's emit it
- if not, then let's load a namespace chunk for this key. and in the meantime let's emit temporarily the NON_BREAKING_SPACE
- when namespace is eventually loaded (can be sync or async) and if the key still can't be resolved let's emit a fallback value

### After
- **first, let's load the namespace (can be sync or async)**
- when the namespace is eventually loaded, check if the key can be resolved (aka "if exists"). if yes, let's emit it
- if not, let's emit the fallback value

Note: If retrieving the namespace wasn't instant (meaning: it was asynchronous), then in the meantime let's emit temporarily the NON_BREAKING_SPACE.

projects/core/src/i18n/i18next/i18next-translation.service.ts
```diff
@@ -73,16 +91,21 @@ export class I18nextTranslationService implements TranslationService {
 
   /**
    * Returns a fallback value in case when the given key is missing
-   * @param key
+   * @param keys
    */
-  protected getFallbackValue(key: string): string {
-    return isDevMode() ? `[${key}]` : this.NON_BREAKING_SPACE;
+  protected getFallbackValue(keys: string[]): string {
```
Platonn, line 96: Shouldn't we mention this method's signature change in the 6.0 migration docs as well?

Platonn:
✅  Tested with combination of many valid keys, and combination of missing keys, and those keys in various order. Tested all those things using 4 combinations of the following setups:
- static translations vs. dynamic translations
- CSR vs. SSR

Platonn:
closing in favor of https://github.com/SAP/spartacus/pull/18874

PR #16703, opened 2022-12-30
Update Open Source Libraries. Compliance. SDOL-005. 

giancorderoortiz:
Compliance purposes. 
OSS libraries versions used must not be older than 1 year.

### DIFF
--- a/projects/core/package.json
+++ b/projects/core/package.json
@@ -22,7 +22,7 @@
     "@ngrx/effects": "^14.3.0",
     "@ngrx/router-store": "^14.3.0",
     "@ngrx/store": "^14.3.0",
-    "angular-oauth2-oidc": "^13.0.1",
+    "angular-oauth2-oidc": "^14.0.1",
     "i18next": "^21.9.1",
     "i18next-http-backend": "^1.4.1",
     "rxjs": "^6.6.0"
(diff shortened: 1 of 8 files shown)


projects/core/package.json
```diff
@@ -22,7 +22,7 @@
     "@ngrx/effects": "^14.3.0",
     "@ngrx/router-store": "^14.3.0",
     "@ngrx/store": "^14.3.0",
-    "angular-oauth2-oidc": "^13.0.1",
+    "angular-oauth2-oidc": "^14.0.1",
```
plabadie, line 25: I think that updating a major version of a dependency in a minor release might be considered non backwards compatible (a breaking change).  Let's discuss.

WeizhengSap: When we upgrade angular to v14, we should update this lib as well. But, `angular-oauth2-oidc` is not ready yet at that time.

giancorderoortiz: See https://wiki.one.int.sap/wiki/display/PSSEC/SDOL-005

bgambocjaviniar:
changes will be included in https://github.com/SAP/spartacus/pull/16912

the existing version for this package is the same as latest release 15 for this package, but he just mentions in a doc for this 14 version that there is no standalone apis for HttpClient ootb (not like ng15).

All in all, does not have any difference between the package version 14 and 15 other than the angular engine changes.

Anyhow, I will close this PR.

PR #16706, opened 2022-12-30
chore(deps): bump decode-uri-component from 0.2.0 to 0.2.2 (#16676)

bgambocjaviniar:
backporting using https://github.com/SAP/spartacus/pull/16676


giancorderoortiz (approved):
Updated Copyright headers also.

PR #16720, opened 2023-01-04
fix: Fix duplicate string literals

Zeyber:
Closes: https://jira.tools.sap/browse/CXSPA-1194


feature-libs/organization/account-summary/components/details/account-summary-mock-data.ts
```diff
@@ -10,11 +10,18 @@ import {
   DocumentStatus,
 } from '@spartacus/organization/account-summary/root';
 
+const PURCHASE_ORDER = 'Purchase Order';
+const JAN_24 = '2022-01-24';
+const JAN_26 = '2022-01-26';
+const FEB_25 = '2022-02-25';
```
Matejk00: I am thinking if wouldn't be better to create an array of dates, instead of separate variables.

Matejk00 (changes requested):
There is one conflict, which needs to be resolved.

Matejk00 (approved):

PR #16721, opened 2023-01-04
Feature/gh 15991 state

ChristophHi:


Uli-Tiger:
Based on our discussion i would have expected that the code for open/close the the conflict solver dialog within the conflict-solver-dialog module would get simplified, but i'm not able to see this in the PR?

ChristophHi:
> Based on our discussion i would have expected that the code for open/close the the conflict solver dialog within the conflict-solver-dialog module would get simplified, but i'm not able to see this in the PR?

It's a bit more clear, see [configurator-groups.service.ts]. We request the new attribute in interactionState directly. But agree, we should revisit the names of the new facade methods and reconsider what to put in facade and what into the new components. Will first merge the branch though so that we have a complete picture

PR #16727, opened 2023-01-05
fix: ensure ids used for name attributes and keys for cxFocus directive are unquie

Uli-Tiger:
In case of immediate conflict resolution an attribute may get rendered twice, once in the conflict resolution popup and once still in background. As the id generation assumes that an attribute is only rendered once, this leads to duplicate ids. In most cases this does not cause any issues, however there are 2 situations where ids need to be unique.

- key of cxFocus directive
- name attributes of controls

for these cases there is an additional uiContextKey injected into the generated ids.


feature-libs/product-configurator/rulebased/components/attribute/types/radio-button/configurator-attribute-radio-button.component.ts
```diff
@@ -20,6 +25,8 @@ export class ConfiguratorAttributeRadioButtonComponent
   extends ConfiguratorAttributeSingleSelectionBaseComponent
   implements OnInit
 {
+  @Input() uiKeyPrefix: string = '';
```
Larisa-Staroverova: `@Input() uiKeyPrefix: string = '';` should be moved to `ConfiguratorAttributeSingleSelectionBaseComponent`

feature-libs/product-configurator/rulebased/components/attribute/types/checkbox-list/configurator-attribute-checkbox-list.component.ts
```diff
@@ -30,6 +30,7 @@ export class ConfiguratorAttributeCheckBoxListComponent
   attributeCheckBoxForms = new Array<UntypedFormControl>();
 
   @Input() group: string;
+  @Input() uiKeyPrefix: string = '';
```
Larisa-Staroverova: `@Input() uiKeyPrefix: string = '';` should be moved to` ConfiguratorAttributeMultiSelectionBaseComponent`

Uli-Tiger: this component is the only implementation of ConfiguratorAttributeMultiSelectionBaseComponent consuming/using the uiKeyPrefix

feature-libs/product-configurator/rulebased/components/attribute/types/radio-button/configurator-attribute-radio-button.component.ts
Uli-Tiger: not all implemtations of ConfiguratorAttributeSingleSelectionBaseComponent use the uiKeyPrefix

PR #16731, opened 2023-01-05
fix: Fix Loop Counter Manipulation on Cart Item List Component (#16710)

Zeyber:
Closes: https://jira.tools.sap/browse/CXSPA-1195


szczepaniak-michal (approved):

Zeyber:
Not backporting.

PR #16732, opened 2023-01-05
fix: Same Value Function Return on Custom Form Validators (#16711)

Zeyber:
Closes: https://jira.tools.sap/browse/CXSPA-908


giancorderoortiz (approved):
Approval based on https://github.com/SAP/spartacus/pull/16711

PR #16739, opened 2023-01-06
refactor: Display pickup in store items or delivery method conditionly

WeizhengSap:
ticket cxspa-1799


feature-libs/checkout/base/components/checkout-delivery-mode/checkout-delivery-mode.component.spec.ts
```diff
@@ -188,36 +202,40 @@ describe('CheckoutDeliveryModeComponent', () => {
   });
 
   it('should get deliveryModeInvalid()', () => {
-    const invalid = component.deliveryModeInvalid;
+    fixture.detectChanges();
 
-    expect(invalid).toBe(true);
+    const invalid = component.deliveryModeInvalid;
+    expect(invalid).toBe(false);
```
RadhepS, line 208: we're changing the assertion from toBe(true) to toBe(false), is this intended?

feature-libs/checkout/base/components/checkout-delivery-mode/checkout-delivery-mode.module.ts
```diff
@@ -8,18 +8,34 @@ import { CommonModule } from '@angular/common';
...
+    PageComponentModule,
+  ],
   providers: [
     provideDefaultConfig(<CmsConfig>{
       cmsComponents: {
         CheckoutDeliveryMode: {
           component: CheckoutDeliveryModeComponent,
+          data: {
+            composition: {
+              inner: ['PickupInStoreDetails'],
```
RadhepS: Should this be CheckoutPickupInStoreDetails?

feature-libs/checkout/base/components/checkout-delivery-mode/checkout-delivery-mode.component.spec.ts
WeizhengSap: yes, by default, the invalid value is true. After `fixture.detectChanges()`, the value should be false. Before, it just check the default value.

RadhepS (approved):

PR #16742, opened 2023-01-09
CXCDS/6825 added additional scope "price" to refresh value after login, logout

ThreadFool:
Back-port


giancorderoortiz (approved):
Based on https://github.com/SAP/spartacus/pull/16621

PR #16743, opened 2023-01-09
CXCDS/6825 

ThreadFool:
Back-port


giancorderoortiz (approved):
Based on https://github.com/SAP/spartacus/pull/16621

PR #16754, opened 2023-01-11
feat: CXSPA-1977 Migration to NPM package manager

pawelfras:
Closes CXSPA-1977


tools/schematics/testing.ts
```diff
@@ -119,29 +112,27 @@ function publishLibs(reload = false): void {
...
   [...files, ...distFiles].forEach((packagePath) => {
-    // Update version in package
     const content = JSON.parse(fs.readFileSync(packagePath, 'utf-8'));
-    content.version = newVersion;
-    fs.writeFileSync(packagePath, JSON.stringify(content, undefined, 2));
 
     // Publish package
     const dir = path.dirname(packagePath);
     console.log(`\nPublishing ${content.name}`);
+    console.log(dir, packagePath);
```
bgambocjaviniar: was this a test log or it's intentional to show the directory and package path

.gitignore
```diff
@@ -20,7 +20,6 @@ docs.zip
 
 # dependencies
 node_modules/
-package-lock.json
```
bgambocjaviniar, line 23: do we want to push yarn lock files? If not, replace with the yarn lock

with npm 7+, npm doesn't mind having yarn lock files though.

Just a food for thought, but I think let's not mix and match in our source code even if it doesn't mind it anymore.

README.md
```diff
@@ -49,7 +49,8 @@ If you are working with Spartacus 5.x, your Angular development environment shou
 
 - [Angular CLI](https://angular.io/): Version **14.2.3** or newer, < 15.
 - [Node.js](https://nodejs.org/en/): Version **14.15** or newer (but **less than** version 15), or **16.10** or newer.
-- [Yarn](https://classic.yarnpkg.com/lang/en/): Version **1.15** or newer.
+- [Yarn](https://classic.yarnpkg.com/lang/en/): Version **1.15** or newer (**OR** from Spartacus 5.2 [npm](https://www.npmjs.com/): Version **6.14** or newer.)
```
bgambocjaviniar, line 52: maybe put it in another entry such as below

```suggestion
- [Yarn](https://classic.yarnpkg.com/lang/en/): Version **1.15** or newer (**OR** from Spartacus 5.2 
- [Npm](https://www.npmjs.com/): Version **6.14** or newer.)
```

nitpicking: Also, just to follow this doc 'convention', let's make it capital

bgambocjaviniar: also, should we give recommendations for previous versions as well? 

 for like when we were at ng 12, 10, 9, 8, etc.

Since we put npm as a recommendation now, maybe it's best to put for the rest as well based on the supported npm version with it's node  counterpart.

Note:
- maybe out of scope? since it needs to be tested in  a customer's 'perspective'.

core-libs/setup/jest.config.js
```diff
@@ -20,4 +20,9 @@ module.exports = {
       lines: 90,
     },
   },
+  //Will become default in v29, and can be removed. See: https://jestjs.io/docs/upgrading-to-jest29
+  snapshotFormat: {
```
bgambocjaviniar, line 24: just a quick question

I noticed you downgraded it to v28.1.3 according to the root package and lock file, whereas it was in at v29 already.

What is the reason for downgrading to this version + putting the default behavior as done in v29?

scripts/install/functions.sh
```diff
@@ -41,7 +41,7 @@ function cmd_clean {
     delete_dir ${BASE_DIR}
     delete_dir storage
 
-    yarn cache clean
+    npm cache verify
```
bgambocjaviniar: I don't think this is a direct 'translation' 

before it used to clean the cache, but now you're just 'verifying' without cleaning. I believe the intent of this script step is to make sure it's in a vanilla state, therefore removing any leftover cache if any.

I believe you went this route because you got an error when putting `npm cache clean` right? I know it doesn't sound nice to `--force` things, but apparently with npm 5 and above, if you want to clean the cache, you always have to force it. 

What do you think, and is my assumption correct?

scripts/install/functions.sh
```diff
@@ -84,10 +84,9 @@ function update_projects_versions {
     fi
 
     printh "Updating all library versions to ${SPARTACUS_VERSION}"
-    for i in ${projects}
-        do
-            (cd "${CLONE_DIR}/${i}" && pwd && sed -i -E 's/"version": "[^"]+/"version": "'"${SPARTACUS_VERSION}"'/g' package.json);
-        done
+    (cd "${CLONE_DIR}/tools/config" && pwd && sed -i -E 's/PUBLISHING_VERSION = '\'\''/PUBLISHING_VERSION = '\'"${SPARTACUS_VERSION}"\''/g' const.ts);
```
bgambocjaviniar, line 86: nice alternative 👍

tools/build-lib/augmented-types/index.js
```diff
@@ -1,4 +1,9 @@
 "use strict";
+/*
```
bgambocjaviniar: Let's remove it here

we have a global license in root which covers licenses. 
Only reason, we have the 'per files' as well is that sonar complains that it's missing these headers, which we only added for .ts. In the end, it was just to 'fix' sonar thrown errors.

tools/config/index.ts
```diff
@@ -148,6 +149,10 @@ export type ProgramOptions = {
    * Sets if versions should be bumped. Use for majors only.
    */
   bumpVersions: boolean | undefined;
+  /**
+   * Sets if dependencies should be re-generated for all libraries.
+   */
+  generateDeps: boolean | undefined;
```
bgambocjaviniar, line 155: noice 👍

scripts/install/functions.sh
bgambocjaviniar: but now we do not need projects in this function right (line 81)

```
function update_projects_versions {
    projects=$@
```

bgambocjaviniar (changes requested):
1 - not to sure, but I don't get prompted in shellapp when installing through schematics from verdaccio now in all combinations (npm as package manager and as yarn with develop and your branch). Not sure what's happening
I will test on someone's machine (TBD)

2 - private pipeline fails 
need to fix on the private pipeline side as it uses yarn and it breaks for the build.yml in the private repo. Let's test again once we change it @RadhepS 

Won't post a link since it's in the private domain, but here's a snippet
Screenshot 2023-01-23 at 1 49 46 AM

3 - ootb cloned repo + installation script (./run.sh install) fails
can be reproduced in our public pipeline when the job runs for `
Installation / Build libs and install (push) `, which basically runs the command above
Screenshot 2023-01-23 at 1 51 23 AM

pawelfras:
Hi @bgambocjaviniar,

> 3 - ootb cloned repo + installation script (./run.sh install) fails can be reproduced in our public pipeline when the job runs for `Installation / Build libs and install (push)`, which basically runs the command above Screenshot 2023-01-23 at 1 51 23 AM

Script `run.sh` by default is cloning the `develop` branch where migration changes are missing. To test the current branch you need to change the `BRANCH` value to `feature/CXSPA-1977` in `scripts/install/config.default.sh` file.

scripts/install/functions.sh
pawelfras: True, good catch!

core-libs/setup/jest.config.js
pawelfras: We had to downgrade Jest to version 28 because `jest-preset-angular` in version 12.2.3 (which is the latest btw.) doesn't support Jest 29.
Due to different formatting of snapshot files between v28 and v29, we decided to set formatting that is the default for v29 to not re-generate snapshots twice - now and in the future after bumping the version pack to 29 as soon as `jest-preset-angular` will support it.
This change has been consulted with @znikola

README.md
pawelfras: Referring to your proposal - Well, I'm not quite sure how to approach this topic yet. We can't just post about `npm` support under section for version 5.x because it won't be true. The version we currently have on the `develop` branch won't install without using the `--legacy-peer-deps` flag due to dependency issues that are being fixed in this PR, so in theory we only support `npm` as of version 5.2

.gitignore
pawelfras: Good point, thanks! I'll replace it with `yarn.lock`

tools/schematics/testing.ts
pawelfras: Good catch, it was used for testing purposes and it's no longer needed.

scripts/install/functions.sh
pawelfras: Yeah, I'm aware that it's not a direct translation, but I was wondering whether we have to clean the cache due to how `npm` is validating it. But that's right - assuming that we want a vanilla state, we should force `npm` to clean the cache. Thanks!

bgambocjaviniar:
successful test and deployments on ccv2 can be found here https://github.tools.sap/cx-commerce/ec-automate-pipelines/pull/29

Will align with someone from CCv2 to have this merge as there's a change required to make it work with the automation script.

This will most likely be merged on Friday.

bgambocjaviniar (approved):

PR #16764, opened 2023-01-12
fix: Reduce cognitive complexity to pass sonar checks

Zeyber:
Closes: https://jira.tools.sap/browse/CXSPA-1118


projects/core/src/routing/services/url-matcher.service.ts
```diff
@@ -129,6 +124,36 @@ export class UrlMatcherService {
...
+      ) {
+        return null;
+      }
+      return { consumed: [], posParams: {} };
+    }
+
+    /**
+     * Logical function to reduce sonar complexity score as matcher scope is limited to inner function.
+     */
+    function or(a: boolean, b: boolean) {
```
Zeyber, line 147: Logical operations in functions are a less than ideal solution to reducing the sonar score. A better way would be to return the function logic without each condition getting `+2` for being nested in another function but I could not find a way to do this.

giancorderoortiz (approved):
I see a significant reduction of cognitive complexity warnings in Sonar thanks to this pr.
It suffices for now. 
What remains should be handled in separate tickets.

PR #16770, opened 2023-01-12
fix: changed heading level for correct document heading hierarchy(CXSPA-1067)

kimhw0630:
Closes: https://jira.tools.sap/browse/CXSPA-1067

change `h4` to `h2`

### DIFF
--- a/projects/storefrontstyles/scss/components/product/list/_active-facets.scss
+++ b/projects/storefrontstyles/scss/components/product/list/_active-facets.scss
@@ -4,7 +4,7 @@
   padding: 10px 0;
 
   color: var(--cx-color-text);
-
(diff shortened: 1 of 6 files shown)


projects/storefrontstyles/scss/components/product/list/_active-facets.scss
```diff
@@ -15,6 +15,17 @@
     }
   }
 
+  h2 {
```
Zeyber, line 18: We can leave a comment to remove the `h4` tag from styles too and include it in our migration docs.

Zeyber (changes requested):

justinlee01 (dismissed):

Zeyber (approved):

PR #16779, opened 2023-01-16
fix: add guard to validate a baseSite switch

Matejk00:
Closes: CXSPA-2062


projects/core/src/auth/user-auth/services/auth-multisite-isolation.service.ts
```diff
@@ -49,4 +54,17 @@ export class AuthMultisiteIsolationService {
...
+    const auth = this.statePersistenceService.readStateFromStorage({
+      key: 'auth',
+    }) as SyncedAuthState;
+
+    let baseSite = await this.getBaseSiteDecorator().toPromise();
+    baseSite = baseSite.replace(this.MULTISITE_SEPARATOR, '');
+
+    const tokenBaseSite = auth.redirectUrl?.split('/')[0];
+
+    return baseSite && baseSite !== tokenBaseSite ? true : false;
```
Zeyber: I think `return baseSite && baseSite !== tokenBaseSite` is enough here.

projects/core/src/auth/user-auth/facade/auth.service.ts
```diff
@@ -52,6 +52,10 @@ export class AuthService {
    */
   async checkOAuthParamsInUrl(): Promise<void> {
     try {
+      if (await this.authMultisiteIsolationService?.isDifferentBaseSite()) {
+        this.logout();
```
Zeyber, line 56: I'm just wondering that if we log out after switching a base site, our credentials would be lost and we'd need to log in again if we switched back immediately after. Is this actually what happens? 
Is it in the scope to maintain sessions between basesites?

projects/core/src/auth/user-auth/facade/auth.service.ts
```diff
@@ -52,6 +52,10 @@ export class AuthService {
...
+      }
```
Zeyber, line 57: An e2e test could be nice to make sure this issue is fixed.

Zeyber (changes requested):

projects/core/src/auth/user-auth/services/auth-multisite-isolation.service.ts
Matejk00: With this I am having an error, as we should return a `boolean`, but with this solution, a `string` seems to be possible. However, I simplified this line.

projects/core/src/auth/user-auth/facade/auth.service.ts
Matejk00: > I'm just wondering that if we log out after switching a base site, our credentials would be lost and we'd need to log in again if we switched back immediately after. Is this actually what happens?

Actually yes, but only if you switch from non-isolated into isolated basesite. If you switch from `electronics-spa` to `powertools-spa` the session can be shared between basesite in my opinion (as it works currently).

> Is it in the scope to maintain sessions between basesites?

Only if you switch from non-isolated to another non-isolated, as far as I know.
There is no specific requirement for how it should work, but `isDifferentBaseSite` method runs only for isolated baseSites. So if there is any active session on non-isolated baseSite the customer should be always logged out to not mix up the sessions between isolated and non-isolated base sites. Because isolated baseSites contains a different database of users.

projects/core/src/auth/user-auth/facade/auth.service.ts
Matejk00: I Added e2e scenario.

projects/storefrontapp-e2e-cypress/cypress/integration/regression/multisite-isolation/multisite-isolation.core-e2e-spec.ts
```diff
@@ -1,109 +1,50 @@
...
-    });
-  });
-
-  beforeEach(() => {
-    clearAllStorage();
+    cy.window().then((win) => win.sessionStorage.clear());
   });
 
-  describe('Standalone authorization', () => {
+  describe('Non-Standalone authorization', () => {
```
Zeyber, line 16: What is the difference between non-standalone and standalone authorization more specifically? Can we describe this in comments?

projects/storefrontapp-e2e-cypress/cypress/integration/regression/multisite-isolation/multisite-isolation.core-e2e-spec.ts
```diff
@@ -1,109 +1,50 @@
...
-      login.loginUser();
-
-      cy.wait('@tokenAuthentication')
-        .its('response.statusCode')
-        .should('eq', 200);
-
-      cy.get(login.userGreetSelector).should('exist');
+    it('should authenticate the customer on the isolated baseSite', () => {
+      isolation.verifyUserSession();
     });
```
Zeyber, line 48: I see that the test cases have been reorganized. Can we have some verification that an account from one basesite cannot be used to log into another? It looks like we have dropped this scenario.

Zeyber (changes requested):

PR #16788, opened 2023-01-18
feat: OPF occ logic with payment render options component

psmul:


integration-libs/opf/package.json
```diff
@@ -28,13 +28,15 @@
...
+    "@angular/platform-browser-dynamic": "^14.2.3",
     "@spartacus/cart": "4.1.0-next.0",
     "@spartacus/checkout": "4.1.0-next.0",
     "@spartacus/core": "4.1.0-next.0",
     "@spartacus/schematics": "4.1.0-next.0",
     "@spartacus/styles": "4.1.0-next.0",
     "bootstrap": "^4.6.1",
-    "rxjs": "^6.6.0"
+    "rxjs": "^6.6.0",
+    "zone.js": "~0.11.8"
```
pawelfras: `query:` is this peer dependency necessary for OPF?

integration-libs/opf/package.json
```diff
@@ -28,13 +28,15 @@
     "@angular-devkit/schematics": "^14.2.3",
     "@angular/common": "^14.2.3",
     "@angular/core": "^14.2.3",
+    "@angular/platform-browser-dynamic": "^14.2.3",
```
pawelfras: `query:` is this peer dependency necessary for OPF?

integration-libs/opf/root/facade/opf-checkout.facade.ts
```diff
@@ -0,0 +1,33 @@
...
+})
+export abstract class OpfCheckoutFacade {
+  /**
+   * Get checkout payment active configurations
+   */
+  abstract getActiveConfigurations(): Observable<ActiveConfiguration[]>;
+  /**
+   * Get checkout payment active configurations
+   */
+  abstract getActiveConfigurationsState(): Observable<
```
pawelfras, line 26: Double-check if this method is needed and if it should be part of the public API. It is used only in the implementation of this facade inside `getActiveConfigurations()` to get data from the state.

integration-libs/opf/styles/components/_opf-checkout-payments.scss
```diff
@@ -0,0 +1,26 @@
...
+    font-weight: 600;
+    font-size: 18px;
+  }
+
+  .form-check {
+    margin-bottom: 0;
+    padding-top: 1rem;
+    padding-bottom: 1rem;
+
+    &:not(:first-child) {
```
pawelfras: As we checked, this line won't work because there are no first child with `.form-check` in code below:
```html
<ng-container *ngIf="activeConfiguratons$ | async as activeConfiguratons">
  <div>
    <div class="title">{{ 'opf.checkout.paymentOption' | cxTranslate }}</div>
    <div class="form-check" *ngFor="let configuration of activeConfiguratons">
    ...
  </div>
</ng-container>
```

integration-libs/opf/components/opf-checkout-payments/opf-checkout-payments.component.html
```diff
@@ -0,0 +1,26 @@
+<ng-container *ngIf="activeConfiguratons$ | async as activeConfiguratons">
+  <div>
+    <div class="title">{{ 'opf.checkout.paymentOption' | cxTranslate }}</div>
```
pawelfras, line 3: consider to use `h3` element to improve a11y

integration-libs/opf/package.json
psmul: Actually good question, because those dependencies seems to be in other places as well, but I don't see point to it. Regardless, removed it here.

integration-libs/opf/package.json
psmul: Actually good question, because those dependencies seems to be in other places as well, but I don't see point to it. Regardless, removed it here.

integration-libs/opf/root/facade/opf-checkout.facade.ts
psmul: Actually, as times we might want to use facade instead of core method. I see this pattern in other places as well. Would leave it.

integration-libs/opf/styles/components/_opf-checkout-payments.scss
psmul: Fixed, thanks!

integration-libs/opf/components/opf-checkout-payments/opf-checkout-payments.component.html
psmul: Fixed in other PR extending logic of this one.
https://github.com/SAP/spartacus/pull/16802

integration-libs/opf/root/facade/opf-checkout.facade.ts
pawelfras: If so, please change JSDoc, because both methods have the same description.

integration-libs/opf/package.json
psmul: Fixed. Fault of outdated jest configuration in files.

integration-libs/opf/package.json
psmul: Fixed. Fault of outdated jest configuration in files.

integration-libs/opf/root/facade/opf-checkout.facade.ts
psmul: Fixed. Removed explicit method as it was redundant.

pawelfras (approved):

PR #16791, opened 2023-01-18
feat: in SSR load JSON translations from a file system

### LINKED ISSUE #13460: In SSR: translation chunk JSON files should be read directly from the filesystem
SSR currently follows the client behaviour for loading translation chunk JSON files: each is requested individual via an HTTP request.  In CCv2 with the 2.x and 3.0 versions this goes via loopback on the localhost address, from 3.1 onwards a fix to the host name handling means it goes to the jsapp endpoint.  Either way it still creates multiple HTTP requests for a single SSR page request (2 per chunk, for main + fallback language).  This amplifies the number of requests coming into the node.js server or jsapps endpoint (depending on the version) & that potentially has implications for denial of service style attacks.

Spartacus versions: 2.x and above
Reproduction: Setup Spartacus SSR with translation chunks, request SSR page

The underlying package that is used here is i18next, which has multiple different backends that it can use for loading the files,  see [https://www.i18next.com/overview/plugins-and-utils#backends](https://www.i18next.com/overview/plugins-and-utils#backends)  The Spartacus code is currently using i18next-xhr-backend (replaced with i18next-http-backend in Spartacus 4.0 I think).  However there are also backends for direct filesystem access in node.js, see [https://github.com/i18next/i18next-fs-backend](https://github.com/i18next/i18next-fs-backend) & also there is a capability to chain different backends, see [https://github.com/i18next/i18next-chained-backend](https://github.com/i18next/i18next-chained-backend).

Spartacus should include a capability to switch the i18n backend to filesystem when in SSR mode - perhaps with a chained backend so that it can fallback to HTTP requests if necessary.  This is both a performance (remove unnecessary HTTP request) and security (potential for DoS) fix in my view.

Platonn:
Previously, when the string config `i18n.backend.loadPath` was configured, the translations were loaded via Angular `HttpClient` from the given path.

Now this option is not recommended anymore due to SSR performance issues. It still works, but under special conditions.  Instead, it's recommended to use the new config property `i18n.backend.loader` (which is a **function**). 

The old option `i18n.backend.loadPath` will work as before only if a customer explicitly provides in their app the class `I18nextHttpBackendService` for the injection token `I18nextBackendService`:
```ts
{ provide: I18nextBackendService, useExisting: I18nextHttpBackendService }
```
This class is not provided/imported anywhere by default, to allow for tree-shaking it (together with it's 3rd party dependency `i18next-http-backend`) when using the new recommended config `i18n.backend.loader`.

The new config property `i18n.backend.loader` is a **function** accepting arguments `langauge` and `namespace` and returning a `Promise` of the translation resources. Instead of making http request, we can now simply use a **dynamic import** to load the translation resources, which will be performant both in CSR and SSR. See the following example:
```ts
{ loader: (lng: string, ns: string) => import(`../../assets/i18n-assets/${lng}/${ns}.json`) }
```

TODO:
- add unit tests
- remove spike todos

**CONSEQUENCES**:
- Different filenames of the lazy loaded translations.
    The dynamically-imported JSON translations resources are bundled by webpack and the output files have different filenames and extensions (`.js`) than the original JSON files. 
    In production build the filenames are uglified by default and look like below:
    image

    In development build the file names look familiar though:
    image

**BREAKING CHANGES**:
- HTTP client is not used anymore for loading translation resources (Angular `HttpClient`, interceptors etc.). Instead translation files are now loaded by the `loader` function, e.g. ```(lng, ns) => import(`../assets/i18n-assets/${lng}/${ns}.json`)``` 
- Previously we setup any i18next backend plugin only when the config property `i18n.backend.loadPath` was present. Now we check for the existence of `i18n.backend` key.
- `I18nextDefaultBackendService` is used by default (instead of `I18nextHttpBackendService`). To bring the old http-driven behavior, please provide the old service: `{ provide: I18nextBackendService, useExisting: I18nextHttpBackendService }`
- `@spartacus/core` has now a new peerDependency `i18next-resources-to-backend` (which is used by default to load translations via dynamic imports)

closes https://jira.tools.sap/browse/CXSPA-2060
closes #13460


projects/core/src/i18n/i18next/i18next-backend/i18next-default-backend.service.ts
```diff
@@ -0,0 +1,32 @@
...
+export class I18nextDefaultBackendService implements I18nextBackendService {
+  constructor(
+    protected config: I18nConfig,
+    @Inject(I18NEXT_INSTANCE) protected i18next: i18n
+  ) {}
+
+  initialize(): InitOptions {
+    const loader = this.config.i18n?.backend?.loader;
+
+    if (!loader || typeof loader !== 'function') {
```
znikola: just a nitpick:

```suggestion
    if (typeof loader !== 'function') {
```

works as well - you don't need to check if `loader` is falsy :)

core-libs/setup/ssr/providers/server-app-dist-folder.ts
```diff
@@ -0,0 +1,36 @@
...
+import { REQUEST } from '@nguniversal/express-engine/tokens';
+
+/**
+ * The absolute directory on the server pointing to the `/dist` folder of the application.
+ *
+ * It can be used for example to determine the path to the application's assets on the server,
+ * so to load them directly from the file system (e.g. using Node's `fs` package).
+ */
+export const SERVER_APP_DIST_FOLDER = new InjectionToken<string>(
+  'BROWSER_APP_DIST_FOLDER',
```
znikola, line 17:
```suggestion
  'SERVER_APP_DIST_FOLDER',
```
Shouldn't it be `SERVER_APP_DIST_FOLDER`?

core-libs/setup/ssr/providers/server-app-dist-folder.ts
```diff
@@ -0,0 +1,36 @@
...
+  { providedIn: 'root', factory: expressServerAppDistFolderFactory }
+);
+
+/**
+ * Returns the absolute directory on the server pointing to the `/dist` folder of the application.
```
znikola:
```suggestion
 * Returns the absolute path to the directory on the server pointing to the `/dist` folder of the application.
```

core-libs/setup/ssr/providers/server-app-dist-folder.ts
```diff
@@ -0,0 +1,36 @@
...
+ * It's retrieved from the `views` setting of the ExpressJS server instance
+ *
+ * Note: it assumes that the setting `views` has been set for the ExpressJS server,
+ *       e.g. `server.set('views', distFolder)`, likely in the file `server.ts`.
+ */
+function expressServerAppDistFolderFactory() {
+  const server = inject(REQUEST).app;
+  const distFolderKey = 'views';
```
znikola, line 30: I guess this WIP. 
Maybe the `SERVER_APP_DIST_FOLDER` token should be an object instead of a `string`, so that it can contain both the dist folder path _and_ the `views` key. 
The `views` key could be optional, and we could fallback to value `views` if it's not provided.

core-libs/setup/ssr/providers/server-app-dist-folder.ts
znikola: Btw, it's not used anywhere atm. I guess it's because this is a draft PR.

projects/core/src/i18n/config/i18n-config.ts
```diff
@@ -23,18 +23,41 @@ export abstract class I18nConfig {
...
+       * for `I18nextBackendService`:
+       * ```ts
+       * { provide: I18nextBackendService, useExisting: I18nextHttpBackendService }
+       * ```
        */
-
       loadPath?: string;
+
+      /**
+       * Function that loads asynchronously the translation resources, based on the provided language and namespace.
```
znikola:
```suggestion
       * Function that asynchronously loads the translation resources, based on the provided language and namespace.
```

projects/core/src/i18n/config/i18n-config.ts
```diff
@@ -23,18 +23,41 @@ export abstract class I18nConfig {
...
+       *
+       *
+       * If you really want to use this config, you'll need to additionally provide `I18nextHttpBackendService`
+       * for `I18nextBackendService`:
+       * ```ts
+       * { provide: I18nextBackendService, useExisting: I18nextHttpBackendService }
+       * ```
        */
-
       loadPath?: string;
```
znikola, line 47: shall we deprecate it? 
do we know any use cases when would customers choose `loadPath` over the more performant `loader`?

projects/core/src/i18n/config/i18n-config.ts
```diff
@@ -23,18 +23,41 @@ export abstract class I18nConfig {
...
-       * `assets/i18n-assets/{{lng}}/{{ns}}.json`
+       * ```ts
+       * { loadPath: 'assets/i18n-assets/{{lng}}/{{ns}}.json' }
+       * ```
+       *
+       * **CAUTION: It's recommended to use the `loader` config property instead, for SSR performance reasons
+       * (for more, see https://github.com/SAP/spartacus/issues/13460).**
+       *
+       *
+       * If you really want to use this config, you'll need to additionally provide `I18nextHttpBackendService`
```
znikola: [nitpick] maybe we can rephrase it to:

```suggestion
       * If you still decide to use this config, you can additionally provide `I18nextHttpBackendService`
```

znikola (changes requested):

core-libs/setup/ssr/providers/server-app-dist-folder.ts
Platonn: Yea, it can be a left over after my previous implementation that was using `fs` and had to know the project's absolute dist folder path in the system.

projects/core/src/i18n/config/i18n-config.ts
Platonn: 1. yes there are usecases - when translations are loaded from 3rd party service (e.g. convenient tool for translator people, that exposes data via API)
2. This config key will work only if additionally customer provides a module for Http backend connector for i18next. In other words, this `loadPath` config doesn't work out of the box. Which can be confusing.... I'm wondering if not remove it at all. And put the configuration directly in the custom backend connector.

projects/core/src/i18n/config/i18n-config.ts
Platonn: Thanks!
not `can` but  `have to`

```suggestion
       * If you still decide to use this config, you have to additionally provide `I18nextHttpBackendService`
```

Platonn:
Changed the architectural design. Closed in favor of https://github.com/SAP/spartacus/pull/17085

PR #16798, opened 2023-01-20
fix: flickering effect on the order details page

Matejk00:
Closes: CXPSA-1980


feature-libs/order/core/facade/order-history.service.ts
```diff
@@ -208,4 +208,11 @@ export class OrderHistoryService implements OrderHistoryFacade {
   resetCancelOrderProcessState(): void {
     return this.store.dispatch(new OrderActions.ResetCancelOrderProcess());
   }
+
+  /**
+   * Returns the order details loading flag
+   */
+  getOrderDetailsLoading(): Observable<boolean> {
+    return this.store.pipe(select(OrderSelectors.getOrderDetailsLoading));
```
Zeyber: I'm thinking about handling the breaking changes in the most convenient way. If we add the feature flag here as returning an Observable with value `true` when we have not updated our feature level, that should handle all cases we use `isOrderLoading$`?

Zeyber (changes requested):

feature-libs/order/core/facade/order-history.service.ts
Matejk00: I think it should return and Observable with a `false` value. Then it's not breaking anything below 5.2

Zeyber (approved):

PR #16802, opened 2023-01-23
feat: toc component implementation with removal of default toc and order button

psmul:


integration-libs/opf/components/opf-checkout-payment-and-review/opf-checkout-payment-and-review.module.ts
```diff
@@ -25,6 +39,7 @@ import { OpfCheckoutPaymentsModule } from '../opf-checkout-payments/opf-checkout
           component: OPFCheckoutPaymentAndReviewComponent,
           guards: [CheckoutAuthGuard, CartNotEmptyGuard],
         },
+        CheckoutPlaceOrder: {},
```
pawelfras: `could:` maybe it's worth consulting this with some architect even at this stage of implementation?

integration-libs/opf/components/opf-checkout-payment-and-review/opf-checkout-payment-and-review.component.spec.ts
```diff
@@ -66,18 +94,26 @@ describe('OPFCheckoutPaymentReviewComponent', () => {
...
 
     fixture.detectChanges();
 
-    expect(
-      fixture.debugElement.query(By.css('cx-opf-checkout-payments'))
-    ).toBeFalsy();
+    expect(el.query(By.css('cx-opf-checkout-payments'))).toBeFalsy();
+  });
+
+  it('should change form value when checkbox get selected / change state', () => {
```
pawelfras, line 108: This test seems to be unstable because it's failing occasionally (3 times for 10 runs on local env). Worth to investigate.
image

integration-libs/opf/components/opf-checkout-payments/opf-checkout-payments.component.ts
```diff
@@ -26,6 +26,9 @@ export class OpfCheckoutPaymentsComponent {
       })
     );
 
+  @Input()
+  disabled = false;
```
pawelfras: If I understand it correctly, the user cannot select a payment option before accepting terms and conditions - in this case `disabled` should have a `true` value by default.

pawelfras:
In addition, I'm not sure what's wrong, but `h2` and `h3` has the same font size 22px which cause they look the same.
image
image

integration-libs/opf/components/opf-checkout-payments/opf-checkout-payments.component.ts
psmul: Right, thanks!

integration-libs/opf/components/opf-checkout-payment-and-review/opf-checkout-payment-and-review.module.ts
psmul: After discussing the technical difficulty of this matter, it will be done in separate task / pull request

pawelfras (approved):

PR #16812, opened 2023-01-25
Epic/future stock

szczepaniak-michal:
Closes CXSPA-236


feature-libs/product/future-stock/components/future-stock-accordion/future-stock-accordion.component.html
```diff
@@ -0,0 +1,27 @@
+<button
+  class="cx-future-stock-accordion-header"
+  [attr.aria-expanded]="expanded"
+  (click)="toggle()"
+>
+  <span>
```
Zeyber: Do we need this `span` here?

feature-libs/product/future-stock/components/future-stock-accordion/future-stock-accordion.component.html
```diff
@@ -0,0 +1,27 @@
...
+    {{ header }}
+    <cx-icon *ngIf="expanded" [type]="iconType.CARET_UP"></cx-icon>
+    <cx-icon *ngIf="!expanded" [type]="iconType.CARET_DOWN"></cx-icon>
+  </span>
+</button>
+
+<div *ngIf="expanded">
```
Zeyber: We could use ng-container here to avoid additional DOM nestings.

feature-libs/product/future-stock/components/future-stock-accordion/future-stock-accordion.component.html
```diff
@@ -0,0 +1,27 @@
+<button
+  class="cx-future-stock-accordion-header"
+  [attr.aria-expanded]="expanded"
```
Zeyber: I think expanded should be an observable.

feature-libs/product/future-stock/components/future-stock-accordion/future-stock-accordion.component.html
```diff
@@ -0,0 +1,27 @@
...
+  (click)="toggle()"
+>
+  <span>
+    {{ header }}
+    <cx-icon *ngIf="expanded" [type]="iconType.CARET_UP"></cx-icon>
```
Zeyber: Maybe we can use a condition to get the arrow direction instead of two `*ngIf`s. Eg. `[type]="expanded ? CARET_UP : CARET_DOWN"`. If expanded is an observable, the icon should respond on change.

feature-libs/product/future-stock/core/connectors/future-stock.adapter.ts
```diff
@@ -0,0 +1,30 @@
...
+   *
+   * Abstract method used to get the future product availability for the specified product
+   */
+  abstract getFutureStock(
+    userId: string,
+    productCode: string
+  ): Observable<ProductFutureStock>;
+  /**
+   *
+   * Abstract method  used to get the future product availability for the list of specified products
```
Zeyber, line 22: Could be nice to have an example of how a list of product codes can be separated in the string.

feature-libs/product/future-stock/core/connectors/future-stock.adapter.ts
```diff
@@ -0,0 +1,30 @@
...
+
+import { Observable } from 'rxjs';
+import {
+  ProductFutureStock,
+  ProductFutureStockList,
+} from '../model/future-stock.model';
+
+export abstract class FutureStockAdapter {
+  /**
+   *
```
Zeyber: Extra lines in these comments.

feature-libs/product/future-stock/components/future-stock/future-stock.component.ts
```diff
@@ -0,0 +1,49 @@
...
+  ChangeDetectionStrategy,
+  Component,
+  OnDestroy,
+  OnInit,
+} from '@angular/core';
+import { FutureStock } from '@spartacus/core';
+import { FutureStockFacade } from '@spartacus/product/future-stock/root';
+import { Subscription } from 'rxjs';
+
+export const NO_FUTURE_STOCK =
```
Zeyber: This should be a translation key.

feature-libs/product/future-stock/components/future-stock/future-stock.component.ts
```diff
@@ -0,0 +1,49 @@
...
+  futureStocks$ = this.futureStockService.getFutureStock();
+  futureStocks: FutureStock[] | string;
+
+  futureStockSubscription: Subscription;
+
+  constructor(protected futureStockService: FutureStockFacade) {}
+
+  ngOnInit() {
+    if (!this.futureStocks) {
+      this.futureStockSubscription = this.futureStocks$.subscribe(
```
Zeyber: Are we able to achieve this result in the template using only the observable?

feature-libs/product/future-stock/components/future-stock/future-stock.module.ts
```diff
@@ -0,0 +1,28 @@
...
+        FutureStockComponent: {
+          component: FutureStockComponent,
+        },
+      },
+    }),
+  ],
+  declarations: [FutureStockComponent],
+  exports: [FutureStockComponent],
+})
+export class FutureStockModule {}
```
Zeyber: Its a bit confusing having multiple modules named the same `FutureStockModule` like here in components and also in root.

projects/storefrontapp-e2e-cypress/cypress/integration/regression/future-stock/future-stock.e2e-spec.ts
```diff
@@ -0,0 +1,74 @@
...
+
+describe('Future Stock', () => {
+  viewportContext(['desktop'], () => {
+    before(() => {
+      cy.window().then((win) => {
+        win.sessionStorage.clear();
+      });
+    });
+
+    it('when not logged in, future stock dropdown should not be visible (CXSPA-236)', () => {
```
Zeyber: Why do we have ticket numbers in the scenarios?

projects/storefrontapp-e2e-cypress/cypress/integration/regression/future-stock/future-stock.e2e-spec.ts
```diff
@@ -0,0 +1,74 @@
...
+const productIdWithFutureStock = '3318057';
+const productIdWithoutFutureStock = '3755219';
+
+const b2bUser = {
+  email: 'mark.rivers@pronto-hw.com',
+  password: 'pw4all',
+};
+
+describe('Future Stock', () => {
+  viewportContext(['desktop'], () => {
```
Zeyber: Are we not testing mobile?

feature-libs/product/future-stock/components/future-stock-accordion/future-stock-accordion.component.ts
```diff
@@ -0,0 +1,29 @@
...
+
+@Component({
+  selector: 'cx-future-stock-accordion',
+  templateUrl: './future-stock-accordion.component.html',
+})
+export class FutureStockAccordionComponent {
+  @Input() header: string;
+  @Input() content: FutureStock[] | string;
+
+  expanded: boolean = false;
```
Zeyber, line 17: Lets handle this with a behaviour subject so that the template can respond more efficiently to changes.

feature-libs/product/future-stock/components/future-stock/future-stock.component.html
```diff
@@ -0,0 +1,7 @@
+<ng-container *ngIf="futureStocks">
+  <cx-future-stock-accordion
+    [header]="'futureStockDropdown.header' | cxTranslate"
```
Zeyber: We shouldn't need to pass this translation key if its always the same for a future stock-specific component.

feature-libs/product/future-stock/components/future-stock/future-stock.component.html
```diff
@@ -0,0 +1,7 @@
...
+    [content]="futureStocks"
```
Zeyber: Can we pass the future stocks observable instead and have the logic handled inside the accordian component?

feature-libs/product/future-stock/occ/adapters/occ-future-stock.adapter.ts
```diff
@@ -0,0 +1,84 @@
...
+    });
+  }
+
+  protected getFutureStocksEndpoint(
+    userId: string,
+    productCodes: string
+  ): string {
+    const params = <any>{};
+
+    if (productCodes) {
```
Zeyber: If this method requires productCodes then we shouldn't need this condition.

Zeyber (changes requested):

feature-libs/product/future-stock/components/future-stock-accordion/future-stock-accordion.component.html
szczepaniak-michal: you are right, not needed, removed it

feature-libs/product/future-stock/components/future-stock-accordion/future-stock-accordion.component.html
szczepaniak-michal: yes, good point, changed div to ng-container

feature-libs/product/future-stock/components/future-stock-accordion/future-stock-accordion.component.html
szczepaniak-michal: changed separate if's to if / else syntax

feature-libs/product/future-stock/components/future-stock/future-stock.component.html
szczepaniak-michal: I would rather keep the container component smart, and accordion presentation, here is why:
https://blog.angular-university.io/angular-2-smart-components-vs-presentation-components-whats-the-difference-when-to-use-each-and-why/

feature-libs/product/future-stock/occ/adapters/occ-future-stock.adapter.ts
szczepaniak-michal: you are right, removed

projects/storefrontapp-e2e-cypress/cypress/integration/regression/future-stock/future-stock.e2e-spec.ts
szczepaniak-michal: dev4 has problems with mobile, will check on develop

feature-libs/product/future-stock/components/future-stock/future-stock.component.html
szczepaniak-michal: good point, removed header as input, it's now translated directly in accordion

feature-libs/product/future-stock/components/future-stock-accordion/future-stock-accordion.component.html
szczepaniak-michal: this would break toggle behavior, you would need to subscribe to behaviorsubject in the component

feature-libs/product/future-stock/core/connectors/future-stock.adapter.ts
szczepaniak-michal: ok, example added

feature-libs/product/future-stock/components/future-stock-accordion/future-stock-accordion.component.html
Zeyber:
```
I believe we could achieve this with an `ng-container *ngIf="expanded$ | async as expanded"` enclosing the component so that nothing else needs to change in the template.
```

Edit:  Just realized this won't work for any false expanded value.

feature-libs/product/future-stock/components/future-stock-accordion/future-stock-accordion.component.html
Zeyber: I think we could do this in the one line if we observe `expanded` with an initial `ng-container` wrapping the component.

Zeyber (dismissed):

Zeyber (approved):

PR #16814, opened 2023-01-26
fix: update tabindex value(CXSPA-948) 

kimhw0630:
Closes: https://jira.tools.sap/browse/CXSPA-948

tabindex value will get update by input value


feature-libs/storefinder/components/store-finder-search/store-finder-search.component.html
```diff
@@ -27,7 +27,7 @@
         ></cx-icon>
         <cx-icon
           *cxFeatureLevel="'5.2'"
-          tabindex="-1"
+          [attr.tabindex]="(queryInput.value && queryInput.value.length) ? 0 : -1"
```
justinlee01: I believe this is allowed in recent versions of Angular

```suggestion
          [attr.tabindex]="(queryInput.value?.length) ? 0 : -1"
```

justinlee01 (commented):
Not a breaking change?

feature-libs/storefinder/components/store-finder-search/store-finder-search.component.html
kimhw0630: i will update the code and it already has feature level check

justinlee01 (approved):

developpeurweb (approved):
Works well with VO and JAWS

PR #16816, opened 2023-01-26
CXEC-23624: code review changes

Parthlakhani:


feature-libs/customer-ticketing/components/details/customer-ticketing-reopen/customer-ticketing-reopen-dialog/customer-ticketing-reopen-dialog.component.ts
```diff
@@ -46,23 +45,29 @@ export class CustomerTicketingReopenDialogComponent
...
     }
   }
 
+  protected onComplete(): void {
+    this.isDataLoading$.next(false);
+    this.close('Ticket reopened successfully');
+  }
+
+  protected onError(): void {
+    this.close('Something went wrong while reopening ticket');
```
bgambocjaviniar, line 68: don't forget, it needs the `this.isDataLoading$.next(false);` too. If an error were to occur, they would forever be in a loading state, therefore it needs to be reset back.

feature-libs/customer-ticketing/components/details/customer-ticketing-details/customer-ticketing-details.component.html
```diff
@@ -1,11 +1,10 @@
-<!-- TODO:CXEC-17553 (Brian / Parth) add spinner until the data is loaded  -->
 <ng-container *ngIf="ticketDetails$ | async as ticket; else spinner">
   <div class="cx-ticket-details">
     <div class="container">
       <div class="cx-details-card">
         <cx-card
           [content]="
-            prepareCardContent(ticket.id!, 'customerTicketing.ticketId') | async
+            prepareCardContent(ticket.id, 'customerTicketing.ticketId') | async
```
bgambocjaviniar, line 7: 👍 , not related to this file, but there still exist `!` in the customer ticketing lib under `feature-libs/customer-ticketing/components/list/customer-ticketing-list/customer-ticketing-list.component.html`

feature-libs/customer-ticketing/components/details/index.ts
```diff
@@ -0,0 +1,11 @@
+/*
+ * SPDX-FileCopyrightText: 2022 SAP Spartacus team <spartacus-team@sap.com>
```
bgambocjaviniar, line 2: don't forget to have these fix from the library to not have both, but just the 2023 one. Just copy paste an existing one from a ts file to have the same 'format'

feature-libs/customer-ticketing/core/facade/customer-ticketing.service.ts
```diff
@@ -314,26 +303,31 @@ export class CustomerTicketingService implements CustomerTicketingFacade {
...
+        )
+      )
     );
   }
 
-  createTicketEvent(ticketEvent: TicketEvent): Observable<TicketEvent> {
-    return this.createTicketEventCommand.execute(ticketEvent);
+  createTicketEvent(
+    ticketEvent: TicketEvent,
+    containsAttachment?: boolean | false
```
bgambocjaviniar: this is a pretty weird typing. I'm assuming you want to have a default value set if no 2nd param is in right?

Then the following is how you should do it 
```suggestion
    containsAttachment?: boolean = false
```

bgambocjaviniar (changes requested):

feature-libs/customer-ticketing/core/facade/customer-ticketing.service.ts
Parthlakhani: It does not accept `=` as a intializer

feature-libs/customer-ticketing/components/details/index.ts
Parthlakhani: Yup, will be doing it in the separate PR after this one is merged.

feature-libs/customer-ticketing/components/details/customer-ticketing-messages/customer-ticketing-messages.component.spec.ts
```diff
@@ -68,9 +66,12 @@ describe('CustomerTicketMessagesComponent', () => {
...
     component.onSend(mockSendEvent);
 
-    expect(customerTicketingFacade.createTicketEvent).toHaveBeenCalledWith({
-      message: 'mock message',
-    });
+    expect(customerTicketingFacade.createTicketEvent).toHaveBeenCalledWith(
+      {
+        message: 'mock message',
+      },
+      false
```
khristian23: What does the second parameter mean?
for understanding purposes can a const be created:
const mustWaitForAttachment = false;  (it could also be includesAttachment)
```suggestion
      mustWaitForAttachment
```

feature-libs/customer-ticketing/components/details/customer-ticketing-messages/customer-ticketing-messages.component.ts
```diff
@@ -44,22 +43,21 @@ export class CustomerTicketingMessagesComponent implements OnDestroy {
 
   messagingConfigs: MessagingConfigs = this.prepareMessagingConfigs();
 
-  onSend(event: { files: FileList | File | undefined; message: string }) {
+  onSend(event: { files: FileList | undefined; message: string }) {
+    const attachmentFlag =
```
khristian23:
```suggestion
    const mustWaitForAttachment =
```

feature-libs/customer-ticketing/components/details/customer-ticketing-reopen/customer-ticketing-reopen-dialog/customer-ticketing-reopen-dialog.component.spec.ts
```diff
@@ -79,7 +79,8 @@ describe('CustomerTicketingReopenDialogComponent', () => {
       component.reopenRequest();
 
       expect(customerTicketingFacade.createTicketEvent).toHaveBeenCalledWith(
-        mockEvent
+        mockEvent,
+        false
```
khristian23:
```suggestion
        mustWaitForAttachment
```

feature-libs/customer-ticketing/components/details/customer-ticketing-reopen/customer-ticketing-reopen-dialog/customer-ticketing-reopen-dialog.component.ts
```diff
@@ -36,33 +34,39 @@ export class CustomerTicketingReopenDialogComponent
       this.form.markAllAsTouched();
       FormUtils.deepUpdateValueAndValidity(this.form);
     } else {
+      const attachmentFlag = this.form.get('file')?.value?.length > 0 ?? false;
```
khristian23:
```suggestion
      const mustWaitForAttachment = this.form.get('file')?.value?.length > 0 ?? false;
```

feature-libs/customer-ticketing/core/facade/customer-ticketing.service.ts
```diff
@@ -103,29 +107,39 @@ export class CustomerTicketingService implements CustomerTicketingFacade {
...
-                  }
-                })
-              )
-          )
-        ),
-      {
-        strategy: CommandStrategy.Queue,
-      }
-    );
+  protected createTicketEventCommand: Command<
```
khristian23, line 104: Using flag parameters goes against clean code:
https://github.com/ryanmcdermott/clean-code-javascript#dont-use-flags-as-function-parameters

The best case would be to pass the attachments and let createTicketEvent decide what to do.
Like this you don't need to repeat the line of code that calculates the 'attachmentFlag' to true to false

Another option would be to have two functions:
createTicketEvent and createTicketEventWithAttachments

feature-libs/customer-ticketing/core/facade/customer-ticketing.service.ts
```diff
@@ -103,29 +107,39 @@ export class CustomerTicketingService implements CustomerTicketingFacade {
...
+                  this.eventService.dispatch({}, TicketClosedEvent);
+                } else if (
+                  (payload.ticketEvent.toStatus?.id === STATUS.OPEN ||
+                    payload.ticketEvent.toStatus?.id === STATUS.INPROCESS) &&
+                  !payload.containsAttachment
+                ) {
+                  this.eventService.dispatch({}, TicketReopenedEvent);
+                } else if (!payload.containsAttachment) {
+                  this.eventService.dispatch({}, NewMessageEvent);
+                }
```
khristian23: Although we should avoid negative conditions, this reads better:

```suggestion
                } else if (!payload.containsAttachment){
                   if (payload.ticketEvent.toStatus?.id === STATUS.OPEN ||
                    payload.ticketEvent.toStatus?.id === STATUS.INPROCESS) {
                        this.eventService.dispatch({}, TicketReopenedEvent);
                    } else {
                        this.eventService.dispatch({}, NewMessageEvent);
                    }
                }
```

feature-libs/customer-ticketing/core/facade/customer-ticketing.service.ts
khristian23:
```suggestion
    containsAttachment = false
```

ishrarhussain (approved):
Looks fine to me.

khristian23 (changes requested):

feature-libs/customer-ticketing/core/facade/customer-ticketing.service.ts
khristian23: In any case, I am OK with leaving the parameter flag, but tests calling the createTicketEvent function must have proper naming.

Parthlakhani: yes, we changed the name and keeping the flag as it would be a big refactor.

feature-libs/customer-ticketing/core/facade/customer-ticketing.service.ts
bgambocjaviniar: only reason why it didn't work is because the `?` optional was there. Only needed to remove `?` and should work. Can't have both a default value and it being optional because it's indirectly optional with the default value

bgambocjaviniar (approved):
awesome :) 

Thanks Parth. Don't forget to update the license headers

PR #16820, opened 2023-01-27
Display prices for drop-down list and read-only

Larisa-Staroverova:
- Display positive and negative numbers with correct sign (+ or -) 
- Display price for read-only attributes
- Display Prices inside DDLB
- fixed old tests and added new tests


feature-libs/product-configurator/rulebased/components/attribute/types/base/configurator-attribute-base.component.ts
```diff
@@ -146,16 +146,33 @@ export class ConfiguratorAttributeBaseComponent {
...
+    value: Configurator.Value | undefined
+  ): string {
+    if (value?.valuePrice?.value && !value.selected) {
+      if (value.valuePrice.value < 0) {
+        title += ` [${value.valuePrice?.formattedValue}]`;
+      } else if (value.valuePrice.value > 0) {
+        title += ` [+${value.valuePrice?.formattedValue}]`;
+      }
+    }
     return title;
```
Uli-Tiger: why not remove title as parameter from getValuePrice and instead just concatenate the result to title in the getLabel Method?
`title += this.getValuePriceLabel(value);`

feature-libs/product-configurator/rulebased/components/price/configurator-price.component.ts
```diff
@@ -31,7 +31,14 @@ export class ConfiguratorPriceComponent {
     if (this.formula.priceTotal) {
       return this.priceTotal;
     } else {
-      return '+ ' + this.formula.price?.formattedValue;
+      if (this.formula.price?.value) {
+        if (this.formula.price?.value > 0) {
+          return '+' + this.formula.price?.formattedValue;
+        } else if (this.formula.price?.value < 0) {
+          return this.formula.price?.formattedValue || '';
```
Uli-Tiger:
```suggestion
          return this.formula.price?.formattedValue ?? '';
```
please use the  [Nullish Coalescing Operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing) to set defaults instead of logical OR, which can be cumbersome.

feature-libs/product-configurator/rulebased/components/price/configurator-price.component.ts
```diff
@@ -31,7 +31,14 @@ export class ConfiguratorPriceComponent {
...
+        }
+      }
+      return '';
```
Uli-Tiger: I'm a bit unhappy with the complexity of this method. For exmple, this code also fullfills all tests.
```
  get price(): string {
    if (this.formula.priceTotal) {
      return this.priceTotal;
    } else if ((this.formula.price?.value ?? 0) > 0) {
      return '+' + this.formula.price?.formattedValue;
    } else {
      return this.formula.price?.formattedValue ?? '';
    }
  }
```

feature-libs/product-configurator/rulebased/components/price/configurator-price.component.ts
```diff
@@ -41,7 +48,11 @@ export class ConfiguratorPriceComponent {
    * @return {string} - total price formula
    */
   get priceTotal(): string {
-    return '+ ' + this.formula.priceTotal?.formattedValue;
+    if (this.formula.priceTotal && this.formula.priceTotal.value >= 0) {
+      return '+' + this.formula.priceTotal?.formattedValue;
+    } else {
+      return this.formula.priceTotal?.formattedValue || '';
```
Uli-Tiger:
```suggestion
      return this.formula.priceTotal?.formattedValue ?? '';
```

feature-libs/product-configurator/rulebased/components/price/configurator-price.component.ts
```diff
@@ -41,7 +48,11 @@ export class ConfiguratorPriceComponent {
    * @return {string} - total price formula
    */
   get priceTotal(): string {
-    return '+ ' + this.formula.priceTotal?.formattedValue;
+    if (this.formula.priceTotal && this.formula.priceTotal.value >= 0) {
```
Uli-Tiger:
```suggestion
    if ((this.formula.priceTotal?.value ?? -1) >= 0) {
```

feature-libs/product-configurator/rulebased/components/price/configurator-price.component.ts
```diff
@@ -41,7 +48,11 @@ export class ConfiguratorPriceComponent {
...
+      return '+' + this.formula.priceTotal?.formattedValue;
+    } else {
+      return this.formula.priceTotal?.formattedValue || '';
+    }
   }
```
Uli-Tiger, line 108: I'm a bit confused - `get price` and `get priceTotal` behave different for prices with value '0'.
 `get price` returns  '0'
 `get priceTotal` returns  '+0'
is this intended?

feature-libs/product-configurator/rulebased/components/price/configurator-price.component.spec.ts
```diff
@@ -57,92 +50,129 @@ describe('ConfiguratorPriceComponent', () => {
...
+        component.formula.price?.formattedValue +
+        ')';
+      if (qty) {
+        expect(component.quantityWithPrice(qty.toString())).toEqual(result);
+      } else {
+        fail();
+      }
+      expect(component.priceTotal).toEqual(
+        '+' + component.formula.priceTotal?.formattedValue
+      );
```
Uli-Tiger: please do not use dynamic assertions. just expect a string literal or a const such as
`expect(component.priceTotal).toEqual('+$420.00')`
this way it is much easier to understand the test and its expectation. Additionally these dynamic expectations just duplicate the business logic they are supposed to test into the test itself.

Also instead of 
```
   const qty = component.formula.quantity;
   if (qty) {
        expect(component.quantityWithPrice(qty.toString())).toEqual(result);
      } else {
        fail();
      }
```
just do
```
const qty = component.formula.quantity;
expect(qty).toBeDefined();
expect(component.quantityWithPrice(String(qty))).toEqual('3 x $220');
```

however for the second assert it would even better if it would get the formula from HTML instead of calling a component method, which seems to me like a unit test inlined into a component test.

Uli-Tiger (changes requested):

feature-libs/product-configurator/rulebased/components/attribute/types/base/configurator-attribute-base.component.ts
Larisa-Staroverova: Agree! Done!

feature-libs/product-configurator/rulebased/components/price/configurator-price.component.spec.ts
Larisa-Staroverova: Agree! Done!

feature-libs/product-configurator/rulebased/components/price/configurator-price.component.ts
Larisa-Staroverova: Absolutely agree! Done!

feature-libs/product-configurator/rulebased/components/price/configurator-price.component.ts
Larisa-Staroverova: Sorry, I do not agree. It is less readable

feature-libs/product-configurator/rulebased/components/price/configurator-price.component.ts
Larisa-Staroverova: Agree!

feature-libs/product-configurator/rulebased/components/price/configurator-price.component.ts
Larisa-Staroverova: Agree!

feature-libs/product-configurator/rulebased/components/price/configurator-price.component.ts
Larisa-Staroverova: Agree!

Uli-Tiger (dismissed):

feature-libs/product-configurator/rulebased/components/attribute/types/read-only/configurator-attribute-read-only.component.ts
```diff
@@ -16,4 +20,88 @@ export class ConfiguratorAttributeReadOnlyComponent extends ConfiguratorAttribut
   @Input() attribute: Configurator.Attribute;
   @Input() group: String;
   @Input() expMode: boolean;
+
+  constructor(protected translation: TranslationService) {
```
ChristophHi: Add old constructor, and let translationService be optional

feature-libs/product-configurator/rulebased/components/price/configurator-price.component.ts
```diff
@@ -30,8 +30,10 @@ export class ConfiguratorPriceComponent {
   get price(): string {
     if (this.formula.priceTotal) {
       return this.priceTotal;
+    } else if ((this.formula.price?.value ?? 0) > 0) {
+      return '+' + this.formula.price?.formattedValue;
```
ChristophHi: Does this work for RTL languages?

ChristophHi (changes requested):

feature-libs/product-configurator/rulebased/components/price/configurator-price.component.ts
Larisa-Staroverova: Agree! Done!

feature-libs/product-configurator/rulebased/components/attribute/types/read-only/configurator-attribute-read-only.component.ts
Larisa-Staroverova: @Stofftiger, could you have a look at it?

Stofftiger: done

feature-libs/product-configurator/rulebased/components/price/configurator-price.component.ts
```diff
@@ -44,8 +92,23 @@ export class ConfiguratorPriceComponent {
...
-      return '+' + this.formula.priceTotal?.formattedValue;
+      if (this.isRTLDirection()) {
+        return this.addSing(
+          this.formula.priceTotal?.formattedValue,
+          '+',
+          false
+        );
+      } else {
+        return this.addSing(this.formula.priceTotal?.formattedValue, '+', true);
+      }
```
Uli-Tiger:
```suggestion
        return this.addSing(
          this.formula.priceTotal?.formattedValue,
          '+',
          !this.isRTLDirection()
        );
```

feature-libs/product-configurator/rulebased/components/price/configurator-price.component.ts
```diff
@@ -22,6 +28,38 @@ export interface ConfiguratorPriceComponentOptions {
...
+  protected addSing(
+    value: string | undefined,
+    sign: string,
+    before: boolean
+  ): string {
+    if (value) {
+      return before ? sign + value : value + sign;
+    }
+    return '';
+  }
```
Uli-Tiger: can we have just one method, that gets all required params:
- formattedValue
- value
- isRTL

and the returns the updated formatted Value.
So the method can then check internally, whether its a positive or a negative value and then add or remove the sign as necessary. So we have all the logic in one place.

feature-libs/product-configurator/rulebased/components/price/configurator-price.component.ts
```diff
@@ -31,8 +69,18 @@ export class ConfiguratorPriceComponent {
     if (this.formula.priceTotal) {
       return this.priceTotal;
     } else if ((this.formula.price?.value ?? 0) > 0) {
-      return '+' + this.formula.price?.formattedValue;
+      if (this.isRTLDirection()) {
+        return this.addSing(this.formula.price?.formattedValue, '+', false);
+      }
+      return this.addSing(this.formula.price?.formattedValue, '+', true);
```
Uli-Tiger:
```suggestion
        return this.addSing(this.formula.price?.formattedValue, '+', !isRTLDirection);
```

Uli-Tiger (changes requested):

Uli-Tiger:
@Larisa-Staroverova: the sign handling is now distributed over 4 methods: get price, get totalPrice, addSign andRemove sign. I believe it would be better understandable if there would be just one method that handler all cases. So it would get the value, the formatted Value, the RTL mode flag and than return an updated formatted value for display.

ChristophHi (dismissed):

feature-libs/product-configurator/rulebased/components/price/configurator-price.component.ts
Stofftiger: changed to: 
return this.addSing(this.formula.price?.formattedValue, '+', !this.isRTLDirection());

ChristophHi:
@Uli-Tiger fixed your last comments, please check!

Uli-Tiger (approved):

PR #16832, opened 2023-01-30
Feature: Tab A11Y directive

justinlee01:
Reusing tab accessibility code by using directives - a parent directive `[cxTabList]` is placed on the tabs' container and child directives `[cxTab]` are placed on the tabs themselves. They handle `aria-selected` and `tabindex` for the user.

Based on this spec: https://www.w3.org/WAI/ARIA/apg/example-index/tabs/tabs-manual.html

A possible enhancement is to automatically set HTML attribute role to value "tab".


steinsebastian:
@justinlee01 When is it planned to release this PR?

Larisa-Staroverova:
Hi @justinlee01, is there any timeline to release this PR?
We would like to bind the tab directive for the product configuration.
Thanks a lot for your response in advance!
Best regards, Larisa

PR #16841, opened 2023-02-01
refactor: add sample data

WeizhengSap:
revert back some changes after adding some new sample data
ticket cxspa-2329


feature-libs/order/components/order-confirmation/order-confirmation-shipping/order-confirmation-shipping.component.html
```diff
@@ -1,12 +1,39 @@
...
-    [cxOutlet]="cartOutlets.DELIVERY_ITEM_LIST"
-    [cxOutletContext]="{
-      entries: entries,
-      readonly: true,
-      deliveryAddress: order.deliveryAddress,
-      deliveryMode: order.deliveryMode
-    }"
-  >
-  </ng-template>
+  <ng-container *ngIf="!showItemList || entries.length > 0">
```
RadhepS, line 2: i want to confirm this logic makes sense?

we want to show this only when showItemFlag is false or there are entries?

feature-libs/order/components/order-details/order-detail-billing/order-detail-billing.component.html
```diff
@@ -1,11 +1,15 @@
...
-    *ngIf="order.paymentInfo"
-    [cxOutlet]="cartOutlets.PAYMENT_INFO"
-    [cxOutletContext]="{
-      readonly: true,
-      paymentDetails: order.paymentInfo
-    }"
-  >
-  </ng-template>
+  <div class="cx-review-summary" *ngIf="order.paymentInfo as paymentDetails">
+    <div class="cx-review-summary-card cx-review-summary-payment-card">
```
RadhepS, line 3: i'm not sure if styling like cx-review-summary-card is accessible here. we can check when sampledata is released

WeizhengSap: I leave styling to UI devs.

feature-libs/order/components/order-confirmation/order-confirmation-shipping/order-confirmation-shipping.component.html
WeizhengSap: if `showItemList` is `true`, then need make sure `entries.length >0`. If not `showItemList` (only display address and delivery mode, e.g. for consignment), no need check `entries` (because entries are not used).

projects/storefrontapp/src/environments/environment.ts
```diff
@@ -21,7 +21,7 @@ import { Environment } from './models/environment.model';
 
 export const environment: Environment = {
   production: false,
-  occBaseUrl: buildProcess.env.CX_BASE_URL,
+  occBaseUrl: 'https://spartacus-dev0.eastus.cloudapp.azure.com:9002',
```
WeizhengSap, line 24: will revert it back after sample data merged.

RadhepS (approved):

PR #16860, opened 2023-02-04
fix: fix accessibility problem in customer search(CXSPA-1292)

kimhw0630:
Closes: https://jira.tools.sap/browse/CXSPA-1292


feature-libs/asm/components/customer-selection/customer-selection.component.ts
```diff
@@ -117,10 +124,92 @@ export class CustomerSelectionComponent implements OnInit, OnDestroy {
...
+   * @param {number} selectedIndex - current selected item index
+   */
+  updateItemIndex(selectedIndex: number): void {
+    this.resultItems.toArray()?.[selectedIndex]?.nativeElement.focus();
+  }
+
+  setFocusOnInput(): void {
+    setTimeout(() => {
+      this.searchTerm.nativeElement.focus();
+    });
```
justinlee01: I don't think you should use `setTimeout` - if the issue is change detection, use `ChangeDetectorRef:detectChanges`.

feature-libs/asm/components/customer-selection/customer-selection.component.ts
```diff
@@ -42,6 +44,9 @@ export class CustomerSelectionComponent implements OnInit, OnDestroy {
 
   @ViewChild('resultList') resultList: ElementRef;
   @ViewChild('searchTerm') searchTerm: ElementRef;
+  @ViewChildren('resultItem') resultItems: QueryList<ElementRef<HTMLElement>>;
+
+  activeIndex = -1;
```
justinlee01: If I understand correctly -
either change the name to make this more clear (`activeFocusedButtonIndex`) or leave a comment explaining this is for ARIA.

feature-libs/asm/components/customer-selection/customer-selection.component.ts
```diff
@@ -117,10 +124,92 @@ export class CustomerSelectionComponent implements OnInit, OnDestroy {
...
+    }
+    this.updateItemIndex(this.activeIndex);
+  }
+  /**
+   * set focus to input search text
+   * @param event keyboard event
+   * @param {boolean} moveOneLeft - flag to move cursor one left
+   * @param {boolean} moveBegin - flag to move cursor to front
+   */
+  focusInputText(event: UIEvent, moveOneLeft = false, moveBegin = false): void {
```
ercultimate: Can these optional arguments be specified together?

feature-libs/asm/components/customer-selection/customer-selection.component.ts
```diff
@@ -42,6 +44,9 @@ export class CustomerSelectionComponent implements OnInit, OnDestroy {
 
   @ViewChild('resultList') resultList: ElementRef;
   @ViewChild('searchTerm') searchTerm: ElementRef;
+  @ViewChildren('resultItem') resultItems: QueryList<ElementRef<HTMLElement>>;
```
justinlee01: The name for this variable can be more descriptive i.e. `searchResultItems`.

justinlee01 (commented):
For my sanity - are we concerned about breaking changes for this PR? Because I perceive a number of changes may be _for a minor version_.

feature-libs/asm/components/customer-selection/customer-selection.component.ts
kimhw0630: no and we will change to enum for these last parameters ~~~

feature-libs/asm/components/customer-selection/customer-selection.component.ts
kimhw0630: the problem was without setTimeout it fires the form's submit event

feature-libs/asm/components/customer-selection/customer-selection.component.ts
kimhw0630: ok will do tomorrow ^^

feature-libs/asm/components/customer-selection/customer-selection.component.ts
kimhw0630: ok will update

feature-libs/asm/components/customer-selection/customer-selection.component.ts
kimhw0630: I will remove last two parameters because from the event object i could get which key has pressed

feature-libs/asm/components/customer-selection/customer-selection.component.ts
kimhw0630: $event.preventDefault() help on Enter is clicked but not ESC is clicked so need to keep setTimeout for closing the result popup

kimhw0630:
@justinlee01  this PR is for 6.0

feature-libs/asm/components/customer-selection/customer-selection.component.ts
kimhw0630: found a solution... 
we need both event handlings for ESC key
event.preventDefault();
event.stopPropagation();

feature-libs/asm/components/customer-selection/customer-selection.component.ts
```diff
@@ -86,12 +96,15 @@ export class CustomerSelectionComponent implements OnInit, OnDestroy {
...
-  selectCustomerFromList(customer: User) {
+  selectCustomerFromList(event: UIEvent, customer: User) {
     this.selectedCustomer = customer;
     this.customerSelectionForm.controls.searchTerm.setValue(
       this.selectedCustomer.name
     );
     this.asmService.customerSearchReset();
+    this.searchTerm.nativeElement.focus();
+    event.preventDefault();
+    event.stopPropagation();
```
justinlee01, line 107: These are usually invoked in the beginning of the function.
Also, I recall `event.stopImmediatePropagation` does both.

feature-libs/asm/components/customer-selection/customer-selection.component.spec.ts
```diff
@@ -240,4 +245,140 @@ describe('CustomerSelectionComponent', () => {
     ).toBeFalsy();
     expect(asmService.customerSearchReset).toHaveBeenCalled();
   });
+
+  describe('Search result navigation', () => {
+    beforeEach(() => {
+      spyOn(asmService, 'getCustomerSearchResults').and.returnValue(
+        of(mockCustomerSearchPage)
+      );
+      component.ngOnInit();
```
justinlee01, line 254: You shouldn't need to invoke `ngOnInit` directly. `fixture.detectChanges` should trigger this

feature-libs/asm/components/customer-selection/customer-selection.component.html
```diff
@@ -7,33 +7,49 @@
...
+    [attr.aria-selected]="activeFocusedButtonIndex === i"
+    [class.active]="activeFocusedButtonIndex === i"
+    (keydown.arrowup)="focusPreviousChild($event)"
+    (keydown.arrowdown)="focusNextChild($event)"
+    (keydown.arrowright)="focusInputText($event)"
+    (keydown.arrowleft)="focusInputText($event)"
+    (keydown.enter)="selectCustomerFromList($event, result)"
+    (keydown.escape)="closeResults($event)"
+    (keydown.home)="focusInputText($event)"
+    (keydown.end)="focusInputText($event)"
```
justinlee01, line 36: Feels like it would be cleaner to put these listeners on the parent element (`div.asm-results`).

justinlee01 (approved):

feature-libs/asm/components/customer-selection/customer-selection.component.html
kimhw0630: this button element is in *ngFor and button itself has focus and click so we need to handle separately if we move to parent so  i think it is better to stay here

feature-libs/asm/components/customer-selection/customer-selection.component.spec.ts
kimhw0630: i will re-test it...   I think i tested with fixture.detectChanges before but i will re-test

feature-libs/asm/components/customer-selection/customer-selection.component.ts
kimhw0630: i will test with stopImmediatePropagation

feature-libs/asm/components/customer-selection/customer-selection.component.html
justinlee01: Fair point on the button itself having focus, but the click event would propagate to the parent element. It's not uncommon to write Javascript this way.

kimhw0630: also, no search result button needs to be moved out from the parent so i will keep what i have

feature-libs/asm/components/customer-selection/customer-selection.component.spec.ts
kimhw0630: i need to keep component.ngOnInit() it fails all test

feature-libs/asm/components/customer-selection/customer-selection.component.ts
kimhw0630: stopImmediatePropagation doesn't work for Enter key.. it fires submit event

ercultimate (approved):

PR #16868, opened 2023-02-07
chore: force-bump transitive dependency 'ua-parser-js' to 1.0.33

Platonn:
related to https://jira.tools.sap/browse/CXSPA-2368


giancorderoortiz (approved):
ok

PR #16873, opened 2023-02-07
chore: update SAPUI5 version to 1.108.5

sam-garland:
1.108 is a long term maintenance version and will be available on the CDN until Q4 2028.

closes: https://jira.tools.sap/browse/CXSPA-2385


sam-garland:
Moved changes into https://github.com/SAP/spartacus/pull/16873

PR #16885, opened 2023-02-08
fix: update config:update script to not crash on release

RadhepS:
closes: https://jira.tools.sap/browse/CXSPA-2406

It failed because the script runs `yarn generate:deps --compare=true` which compares the file with the previous state and fails if they are different. When it comes to releasing, the file states will be different because the version was updated. So we can add a check to see that if we are releasing, run `yarn generate:deps` instead, which does no comparison, and updates dependencies.json.

### DIFF
--- a/tools/config/index.ts
+++ b/tools/config/index.ts
@@ -17,7 +17,7 @@ import { execSync } from 'child_process';
 import { Command } from 'commander';
 import { readFileSync } from 'fs';
 import glob from 'glob';
-import { NG_PACKAGE_JSON, PACKAGE_JSON } from './const';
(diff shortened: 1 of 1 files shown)


tools/config/index.ts
```diff
@@ -278,7 +278,11 @@ manageDependencies(repository, options);
 manageTsConfigs(repository, options);
 
 // collect and generate dependencies.json file.
+if (PUBLISHING_VERSION) {
```
plabadie, line 281: Would there be a way to make this more deliberate, like, for example passing a param to the script?

RadhepS:
I created a PR for this because I thought it can be a quick fix. Patrick suggested a better/proper way to resolve this but I won't spend time on that solution because of other priorities so I will close this PR and move the ticket the reserve board until it is prioritized.

PR #16892, opened 2023-02-10
CXCDS-7119: Retrieve CR from local-storage on page refresh to avoid blank CR headers

ShankariGiri-I:
This change allows cds spa-integration to pick up the CR from local-storage to initialize the variable used in setting X-Consent-Reference headers to occ calls (including search). 

This is crucial when the user refreshes the page as the initial requests sent had no CR header in the previous version - leading to lack of personalization of the search results.


robertstrache (dismissed):

integration-libs/cds/src/profiletag/services/profiletag-event.service.ts
```diff
@@ -44,6 +45,23 @@ export class ProfileTagEventService {
...
+  }
+
+  private setConsentReferenceFromLocalStorage(): void {
+    const profileTagMetadata = JSON.parse(
+      localStorage.getItem('profiletag') || '{"cr":{}}'
+    );
+    this.baseSiteService
+      .getActive()
+      .pipe(take(1))
+      .subscribe((baseSite) => {
```
plabadie: It's best practice to unsubscribe any subscription in the onDestroy method.
There are many examples in the code base.  The subscription management is done with a Subscription object.  Something like:

```
      this.subscription.add(
        example.subscribe()
        );
```

```
  ngOnDestroy(): void {
    this.subscription.unsubscribe();
  }
```

plabadie (changes requested):
See small comment on the PR

integration-libs/cds/src/profiletag/services/profiletag-event.service.ts
ShankariGiri-I: Fixed it. Could you have another look? Thanks! :)

plabadie (approved):

PR #16900, opened 2023-02-13
refactor: remove some duplicate codes

WeizhengSap:
ticket: feature/CXSPA-2422


feature-libs/checkout/base/components/checkout-review-submit/checkout-review-submit.component.ts
```diff
@@ -26,7 +26,8 @@ import {
   FeatureConfigService,
   TranslationService,
 } from '@spartacus/core';
-import { Card, getAddressNumbers, ICON_TYPE } from '@spartacus/storefront';
+import { deliveryAddressCard, deliveryModeCard } from '@spartacus/order/root';
```
RadhepS, line 29: checkout library now has a dependency on the order library, is this fine/worth it?

RadhepS (approved):

PR #16903, opened 2023-02-14
test: make sure to wait for pricing calls when selecting a value to address flakiness (CXSPA-2421)

Uli-Tiger:
In case config pricing is enabled, tests will now wait for the config pricing call to finish after changing a value. This might improve test stability especially if back-end is under load so that responses are slow.

### DIFF
--- a/projects/storefrontapp-e2e-cypress/cypress/helpers/product-configurator-vc.ts
+++ b/projects/storefrontapp-e2e-cypress/cypress/helpers/product-configurator-vc.ts
@@ -4,7 +4,6 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { PRODUCT_NO_PRICING } from '../sample-data/b2b-bulk-pricing';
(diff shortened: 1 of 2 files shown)


projects/storefrontapp-e2e-cypress/cypress/helpers/product-configurator-vc.ts
```diff
@@ -4,6 +4,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+import { PRODUCT_NO_PRICING } from '../sample-data/b2b-bulk-pricing';
```
Larisa-Staroverova: I get the following notification: 'PRODUCT_NO_PRICING' is declared but its value is never read.
Please, remove unused import.

Larisa-Staroverova (changes requested):

Larisa-Staroverova (approved):

PR #16912, opened 2023-02-17
feat: Upgrade to Angular 15 - 15.2.0

bgambocjaviniar:
## PACKAGE UPDATES
- updated `@angular/` and `@angular-devkit/` from 14.2.3 to 15.2.0

- updated `@angular-eslint/` from 14.1.2 to 15.2.1 https://github.com/angular-eslint/angular-eslint/blob/v15.2.1/package.json#L56

- updated `angular-oauth2-oidc` from 13.0.1 to 15.0.1 - due to https://github.com/SAP/spartacus/pull/16703. 14.x and 15.x are literally the same. It's just the major version is the engine number for the angular version. No big changes done between them as 15 was created first and cutoff to 14 as well while docs were provided on what is not supported.

- `ngx-infinite-scroll` needed to be updated from 14 to 15. But I believe in the future we want to remove this dependency entirely.

- updated `typescript` version to a minimum of ^4.8.2 from ^4.7.2.

- updated `jest` to v29 from v28 since `jest-preset-angular` with v13 (from v12) now supports jest 29.

## BREAKING CHANGE / DEPRECATION UPDATES
- schematics 
> - SchematicTestRunner has deprecated two functions `runExternalSchematicAsync` and `runSchematicAsync`.
> 1. `runExternalSchematicAsync` was an observable which we had to convert to a promise, but now they removed this repetition and created `runExternalSchematic` that returns a promise. runExternalSchematicAsync -> runExternalSchematic
> 2. `runSchematicAsync` was an observable which we had to convert to a promise, but now they removed this repetition and created `runSchematic` that returns a promise. runSchematicAsync -> runSchematic

> - externalSchematics for @schematics/angular all requires a project property https://github.com/SAP/spartacus/pull/16912/commits/f5d0dc4abfb500134fa428bed91a735d06397285

- out of the box angular
> - an out of the box ng15 angular vanilla project no longer comes with ngOnInit and an empty constructor. **However, we depend on this constructor to exist for creating a cms component with schematics** https://github.com/SAP/spartacus/pull/16912/commits/f12c1f32eeb6ee876c63f774df1d030485cf467d
> 1. a template was created to put back the empty constructor. This will most likely be updated and removed in the future to update our schematics to adhere to inject function instead of constructor for dependency injection.

> updated our angular skeleton to adhere to the new structure
> 1. instead of a path to polyfills and an import to zone.js, we can specify them in angular.json https://github.com/SAP/spartacus/pull/16912/commits/93a3e0681c85b1140067d125e59bc07aaf0b20d9

> `test.ts` is no longer part of the angular project by default https://github.com/SAP/spartacus/pull/16912/commits/e8ae1925e188c33aac7322fea089e0108242ef0d

-  fileReplacement for environment became optional

-  browserlist configurations file is part of the lib by default

-  removed all deprecated usage for `InjectFlags` to use `InjectOptions`. Basically, instead of using straight boolean flags, Angular team made it into objects.

-  relativeLinkResolution is no longer exposed, therefore removed for the `RouterModule.forRoot`

## MISC

- compiler is more stricter which caught a mis-typings
> https://github.com/SAP/spartacus/pull/16912/commits/672ee66e1563075382025eea64a0d09910a84e46
> https://github.com/SAP/spartacus/pull/16912/commits/7bbe6cfc5162327d84a673f335cfe8f9eaeef5b8
> https://github.com/SAP/spartacus/pull/16912/commits/14ee84af4ad55cc5418970c4c0046c759b7fa964

- updated CI, scripts, helpers to run on ng15

- updated remaining NgRx `@Effect` decorator, which is deprecated to use `createEffect()` function instead.

- removed all 'extra' redundancy from all `test.ts` that are no longer required as they're provided by default
> 1. removed the imports for zone
> 2. removed the require declaration
> 3. removed the context 

- removal of Spartacus 'wrapped' angular pwa schematics
> 1. This was not an issue for 15.1.x. This change that causes our spartacus pwa schematics test to fail was introduced in 15.2.x
> 2. JIRA - TBD, however it is related to [angular pwa being an ESM](https://github.com/angular/angular-cli/blob/15.2.0/packages/angular/pwa/pwa/index.ts#L33) while we're compiling our jest in CommonJs.
> 3. ticket will be related to either bringing it back once we're able to properly configure the jest config to compile in es than commonjs. We're able to configure to run on ES6 as it is currently experimental, but it's a matter or configuring it properly in our project.

closes https://jira.tools.sap/browse/CXSPA-2278


feature-libs/organization/administration/core/store/effects/org-unit.effect.ts
```diff
@@ -464,33 +464,37 @@ export class OrgUnitEffects {
...
-  //           new OrgUnitActions.LoadAddressesSuccess({ page, orgUnitId }),
-  //         ];
-  //       }),
-  //       catchError(error =>
-  //         of(
-  //           new OrgUnitActions.LoadAddressesFail({
-  //             orgUnitId,
-  //             error: normalizeHttpError(error),
-  //           })
+  // > = createEffect(() =>
```
plabadie: The dead code should probably just be removed.

plabadie:
Likely the main Spartacus readme `./README.md` needs updating.
This file is a maybe: `./docs/self-publishing-spartacus-libraries.md`.  It talks about ng14 bu also about spartacus 5.0.  Perhaps it's a different case.

plabadie:
Let's add a mention of the pea schematics removal in the migration doc https://github.com/SAP/spartacus/blob/develop/docs/migration/6_0/typescript-manual-doc.md

feature-libs/organization/administration/core/store/effects/org-unit.effect.ts
bgambocjaviniar: removed

plabadie (approved):

PR #16918, opened 2023-02-20
Cypress12 Upgrade

FollowTheFlo:


ci-scripts/e2e-cypress.sh
```diff
@@ -93,9 +93,9 @@ else
...
 
-    if [ "${GITHUB_EVENT_NAME}" == "pull_request" ]; then
-        yarn e2e:run:ci:core"${SUITE}"
-    else
+    # if [ "${GITHUB_EVENT_NAME}" == "pull_request" ]; then
+    #     yarn e2e:run:ci:core"${SUITE}"
+    # else
         yarn e2e:run:ci"${SUITE}"
-    fi
+    # fi
```
plabadie: Just a reminder to rollback before merging.

projects/storefrontapp-e2e-cypress/cypress/helpers/b2b/b2b-checkout.ts
```diff
@@ -495,7 +495,7 @@ export function reviewB2bOrderConfirmation(
     });
 
     if (!replenishment) {
-      cy.get('.cx-summary-card:nth-child(2) .cx-card').within(() => {
+      cy.get('.cx-summary-card:nth-child(2)').within(() => {
```
plabadie, line 498: Why remove cx-card in the context of a cy upgrade?

projects/storefrontapp-e2e-cypress/cypress/helpers/b2b/b2b-replenishment-order-history.ts
```diff
@@ -102,7 +102,7 @@ export function verifySorting() {
 export function cancelReplenishmentInHistory() {
   const replenishmentHistoryAlias = visitReplenishmentHistory();
 
-  cy.get(replenishmentHistoryAlias).should((xhr) => {
+  cy.get(replenishmentHistoryAlias).then((xhr) => {
```
plabadie, line 105: Why change should for then?

projects/storefrontapp-e2e-cypress/cypress/helpers/checkout-flow.ts
```diff
@@ -553,12 +564,12 @@ export function verifyOrderConfirmationPageWithCheapProduct(
...
       });
     });
-    cy.get('.cx-summary-card:nth-child(2) .cx-card').within(() => {
+    cy.get('.cx-summary-card:nth-child(2)').within(() => {
       cy.contains(sampleUser.fullName);
       cy.contains(sampleUser.address.line1);
       cy.contains('Standard Delivery');
     });
-    cy.get('.cx-summary-card:nth-child(3) .cx-card').within(() => {
+    cy.get('.cx-summary-card:nth-child(3)').within(() => {
```
plabadie, line 561: Why change the selector?

projects/storefrontapp-e2e-cypress/cypress/helpers/checkout-forms.ts
```diff
@@ -174,7 +174,7 @@ export function fillPaymentDetails(
        * It takes time for the delivery address to set.
        * Was reported in the ec-spartacus-release https://sap-cx.slack.com/archives/GLJ5MR1LL/p1586937731001500
        */
-      //cy.wait(3000);
+      cy.wait(3000);
```
plabadie: Was the wait required to pass the test under cy 12?

projects/storefrontapp-e2e-cypress/cypress/helpers/product-configurator-cart.ts
```diff
@@ -13,12 +13,12 @@ export function clickOnEditConfigurationLink(cartItemIndex: number): void {
...
-        .click()
-        .then(() => {
-          cy.location('pathname').should('contain', '/cartEntry/entityKey/');
-        });
+    .as('aElement');
+  cy.get('@aElement')
+    .find('a:contains("Edit")')
+    .click()
+    .then(() => {
+      cy.location('pathname').should('contain', '/cartEntry/entityKey/');
```
plabadie, line 21: What is the purpose of this change?

projects/storefrontapp-e2e-cypress/cypress/helpers/product-configurator-vc.ts
```diff
@@ -278,13 +279,14 @@ function clickOnConflictSolverLink(attribute: string, linkName: string): void {
...
 export function isConflictLinkAttached(attribute: string): void {
-  cy.get('cx-configurator-attribute-header').within(() => {
-    cy.get(`#cx-configurator--attribute-msg--${attribute}`, {
+  cy.get('cx-configurator-attribute-header').as('conf');
+  cy.get('@conf')
+    .find(`#cx-configurator--attribute-msg--${attribute}`, {
       timeout: 10000,
-    }).within(() => {
+    })
+    .within(() => {
```
plabadie, line 290: What is the purpose of this refactor?

projects/storefrontapp-e2e-cypress/cypress/plugins/index.js
```diff
@@ -15,13 +14,17 @@ module.exports = (on, config) => {
 
     deleteFolder(folderName) {
       return new Promise((resolve, reject) => {
-        rmdir(folderName, { maxRetries: 3, recursive: true }, (err) => {
-          if (err) {
-            console.error(err);
-            return reject(err);
+        fs.rm(
+          folderName,
+          { maxRetries: 3, recursive: true, force: true },
```
plabadie, line 19: what is the purpose of `force: true`?

projects/storefrontapp-e2e-cypress/cypress/helpers/b2b/b2b-checkout.ts
FollowTheFlo: yes, was failing due to several cx-card elements inside 'cx-summary-card:nth-child(2)'.
It is a new restriction for 'within' on Cy12 https://docs.cypress.io/guides/references/migration-guide#within

because 'cx-summary-card:nth-child(2)' is ensure to be a signle element, it fixes the prob, the DOM area is now slighly larger but does not cause any side effect with the inner 'within' logic:
  cy.get('.cx-card-title').should('contain', 'Placed on');
  cy.get('.cx-card-label').should('not.be.empty');

projects/storefrontapp-e2e-cypress/cypress/helpers/b2b/b2b-replenishment-order-history.ts
FollowTheFlo: Restriction on 'should' on Cy12, No cy commands allowed inside a 'should', see: https://docs.cypress.io/guides/references/migration-guide#should
It needs to be replaced by 'then'.
I belive 'should' was use to have a 'replay' cabability:'special logic to rerun the callback function until no assertions throw within it.'
https://docs.cypress.io/api/commands/should#Whats-the-difference-between-then-and-shouldand

projects/storefrontapp-e2e-cypress/cypress/helpers/checkout-forms.ts
FollowTheFlo: was a desperate attempt to fix the intermittent prob as it looks like this tempo was already as a workaround in the past. Will be reverted.

ci-scripts/e2e-cypress.sh
FollowTheFlo: will be done, thx

projects/storefrontapp-e2e-cypress/cypress/helpers/checkout-flow.ts
FollowTheFlo: Same reason as above, it is due to new restriction on 'within' command. There are about 6 or 7 occurrences of this change. 
Some other solutions exists as explained in https://docs.cypress.io/guides/references/migration-guide#within
like using 'alias+find'
I believe the more simple and less intrusive is to apply 'within' on parent element '.cx-summary-card:nth-child(3)', it is just slightly enlarges the DOM area, not causing any side effect.

projects/storefrontapp-e2e-cypress/cypress/helpers/product-configurator-cart.ts
FollowTheFlo: Due to restriction on 'within' no supporting multiple subjects anymore, I applied the fix mention in Migration Guide: https://docs.cypress.io/guides/references/migration-guide#within
which consists of adding alias then "find" command so we can handle multiple subjects.'

this code was throwing an error on Cy12 due to .find('cx-configure-cart-entry') returning several elements.
```
.find('cx-configure-cart-entry')
.within(() => {
      cy.get('a:contains("Edit")')

```
is replaced by:
```
.find('cx-configure-cart-entry')
.as('aElement');
  cy.get('@aElement')
    .find('a:contains("Edit")')
```

projects/storefrontapp-e2e-cypress/cypress/helpers/product-configurator-vc.ts
FollowTheFlo: Same refactor as above to handle 'within' command  Cy12 restriction, following workaround by migration guide, 'within + inner get' was replaced by 'get alias + find' 
```
 cy.get('cx-configurator-attribute-header').within(() => {
    cy.get(`#cx-configurator--attribute-msg--${attribute}`, {
      timeout: 10000,
    })
```
replaced by
```
 cy.get('cx-configurator-attribute-header').as('conf');
  cy.get('@conf')
    .find(`#cx-configurator--attribute-msg--${attribute}`, {
      timeout: 10000,
    })
```

The second 'within' (line 287) is not throwing any error, so stays as it is
```
.within(() => {
      cy.get('a.cx-action-link').wait(1000);
    });
```

projects/storefrontapp-e2e-cypress/cypress/plugins/index.js
FollowTheFlo: from guide https://nodejs.org/api/fs.html#fsrmpath-options-callback
When true, exceptions will be ignored if path does not exist. Default: false.

projects/storefrontapp-e2e-cypress/cypress/e2e/regression/checkout/express-checkout.e2e.cy.ts
```diff
@@ -1,11 +1,19 @@
...
 import * as checkout from '../../../helpers/checkout-flow';
 import * as expressCheckout from '../../../helpers/express-checkout';
 import { viewportContext } from '../../../helpers/viewport-context';
 import { clearAllStorage } from '../../../support/utils/clear-all-storage';
+import { isolateTestsBefore } from '../../../support/utils/test-isolation';
 
-context('Express checkout', () => {
+context('Express checkout', { testIsolation: false }, () => {
   viewportContext(['mobile', 'desktop'], () => {
+    isolateTestsBefore();
```
FollowTheFlo, line 16: Using isolateTestsBefore() and not isolateTests().
isolateTestsBefore() -> Clearing localStorage and cookies only before the 'context' block, not before each 'it'.
Reason: cookie 'checkout express true' (set in 'before' hook) needs to exist all along 'it' blocks
If using 'isolateTests()', 1 test is failing

projects/storefrontapp-e2e-cypress/cypress/e2e/accessibility/aux-key.e2e.cy.ts
```diff
@@ -1,6 +1,14 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { isolateTests } from '../../support/utils/test-isolation';
+
 context('Auxiliary Keys', () => {
   describe('Category Navigation', () => {
-    before(() => {
+    beforeEach(() => {
```
FollowTheFlo, line 11: rather than flagging it as {TestIsolation=false}, I made this describe block compatible with TestIsolation by changing the hook to run before each 'it' block. tested, no side effects.

projects/storefrontapp-e2e-cypress/cypress/e2e/accessibility/aux-key.e2e.cy.ts
```diff
@@ -44,7 +52,7 @@ context('Auxiliary Keys', () => {
   });
 
   describe('My Account Navigation', () => {
-    before(() => {
+    beforeEach(() => {
```
FollowTheFlo, line 55: rather than flagging it as {TestIsolation=false}, I made this describe block compatible with TestIsolation by changing the hook to run before each 'it' block. tested, no side effects.

projects/storefrontapp-e2e-cypress/cypress/e2e/regression/consignment-tracking/consignment-tracking.core-e2e.cy.ts
```diff
@@ -21,6 +27,7 @@ describe('consignment tracking', () => {
       cy.get('.close').click();
       cy.get('.cx-list')
         .next()
+        .first()
```
FollowTheFlo, line 30: Fixing the new restriction on 'within', force single subject.
tested, no side-effect.

projects/storefrontapp-e2e-cypress/cypress/e2e/regression/my-account/order-history-orders-flow.e2e.cy.ts
```diff
@@ -36,8 +43,9 @@ describe('Order History with orders', () => {
   });
 });
 
-describe('Order details page', () => {
+describe('Order details page', { testIsolation: false }, () => {
   viewportContext(['mobile', 'desktop'], () => {
+    isolateTestsBefore();
```
FollowTheFlo, line 48: using isolateTestsBefore() instead of isolateTests().
getting a token error if clearing localStorage between 'it' blocks.

projects/storefrontapp-e2e-cypress/cypress/e2e/regression/my-account/payment-methods.e2e.cy.ts
```diff
@@ -108,3 +116,6 @@ describe('Payment Methods', () => {
     });
   });
 });
+function clearCacheTestIsolation() {
```
FollowTheFlo: generated noise, will be removed

projects/storefrontapp-e2e-cypress/cypress/support/utils/test-isolation.ts
```diff
@@ -0,0 +1,36 @@
...
+For existing Spartacus e2e tests not supporting TestIsolation, utils functions can simulates pre-Cy12 behavior:
+https://docs.cypress.io/guides/references/migration-guide#Simulating-Pre-Test-Isolation-Behavior
+*/
+
+function clearLocalStorageAndCookies() {
+  cy.clearLocalStorage();
+  cy.clearCookies();
+}
+
+export function isolateTests(): void {
```
FollowTheFlo, line 23: function containing hooks to simulate pre-cy12 behavior, see guide:
https://docs.cypress.io/guides/references/migration-guide#Simulating-Pre-Test-Isolation-Behavior
in guide, only 'beforeEach' hook is present.
However 'before' also needs to be present to ensure logic in existing e2e test 'before' is not corrupted by cache of previous test.

Many e2e tests have a logic in 'Before' hook (eg: register/login/pageNavigation) without previously clearing cooking/storage.
if previous test (test1) finishes with a diff BasedSite (eg:apparel-uk stored in cookie) or with logged-in user (token in local storage)
The logic inside 'before' hook of next test (test2) will be affected - unwanted basedSite - or unwanted loggedIn user
eg: test1 -> apparel-checkout-flow.e2e : store 'apparel-uk-spa' baseSite get stored in cookie
then
test2-> home-page.core.e2e -> existing 'before' hook navigate to homePage -> Homepage will show as 'apparel-uk-spa' domain

projects/storefrontapp-e2e-cypress/cypress/support/utils/test-isolation.ts
```diff
@@ -0,0 +1,36 @@
...
+  before(() => {
+    clearLocalStorageAndCookies();
+  });
+  beforeEach(() => {
+    clearLocalStorageAndCookies();
+  });
+}
+
+export function isolateTestsBefore(): void {
```
FollowTheFlo, line 32: function isolateTestsBefore() is used in 3 exceptions where cache only needs to be cleared when before starting the test but leaving localStorage/cookies persistent between 'it' blocks.

projects/storefrontapp-e2e-cypress/cypress/helpers/checkout-as-guest.ts
```diff
@@ -98,5 +98,5 @@ export function createAccountFromGuest(password: string) {
     cy.get('[formcontrolname="passwordconf"]').clear().type(password);
     cy.get('button[type=submit]').click();
   });
-  cy.wait(`@${homePage}`).its('response.statusCode').should('eq', 200);
+  cy.wait(`@${homePage}`);
```
FollowTheFlo, line 101: fix for random failure.(CXSPA-2457)
Server logs show the 200 response is returned by server, however intermittently never received by client (for unknown reason, prob network related), the fix is to ignore the 200 status code so tests does not fail.

projects/storefrontapp-e2e-cypress/cypress/e2e/regression/cart/cart.e2e.cy.ts
```diff
@@ -47,6 +53,7 @@ describe('Cart', () => {
       });
 
       it('should keep cart on page refresh', () => {
+        cy.visit('/cart');
```
FollowTheFlo, line 56: just needed the navigation to /cart to be compatible with TestIsolation. so no need to flag the full 'describe' block

plabadie (dismissed):

plabadie (approved):

PR #16933, opened 2023-02-23
feat: improve test stability (CXSPA-2619)

Uli-Tiger:
This should reduce test flakiness.
Pricing requests are fired when the get/update config returns. A `cy.wait()` will wait for the last fired request, matching the given route, that has not yet been waited for. In case of pricing we do not wait for each pricing request in all test cases.
So the following might happen. We typically start to wait for the pricing request, after the update request returns. Because the pricing is fired when the update returns, it might be that pricing request actually fires a little bit after the cypress test started to wait for it. In this case it will most likely find an 'old' pricing request that has not been waited for and the wait immediately returns as it only waits for the 'old' pricing request that did already return.
By just delaying (100ms) the wait for pricing requests, we ensure that the pricing requests was fired, so that the wait will wait for the correct request.
For one test class 'product-configurator-tabbing.core-e2e-spec.ts'  this PR ensures that we wait for all pricing requests, hence in this case the delay of 100ns could be removed. However this is not implemented for all product config test classes, yet. hence the delay is still required.


projects/storefrontapp-e2e-cypress/cypress/helpers/product-configurator-vc.ts
```diff
@@ -309,17 +317,6 @@ export function clickOnViewInConfigurationAndWait(attribute: string): void {
...
- * Assuming the given attribute is involved in the conflict, it navigates from the conflict group to standard group
- * containing the corresponding attribute and waits for request to finish.
- *
- * @param attribute - Attribute name
- */
-export function clickOnViewInConfigurationAndWait(attribute: string): void {
-  clickOnViewInConfiguration(attribute);
-  cy.wait(GET_CONFIG_ALIAS);
-}
-
```
Uli-Tiger, line 322: just in case: This method was duplicated (error in vs code). Guess this happened during a manual merge accidentally. hence i removed it here.

projects/storefrontapp-e2e-cypress/cypress/helpers/product-configurator-vc.ts
```diff
@@ -612,28 +609,50 @@ export function selectAttributeAndWait(
   configuration.selectAttribute(attributeName, uiType, valueName, false);
   cy.wait(UPDATE_CONFIG_ALIAS);
   if (isPricingEnabled) {
-    cy.wait(CONFIG_PRICING_ALIAS);
+    waitForPricing();
   }
 }
 
+function waitForPricing() {
```
Stofftiger: why do we not create one function to handle all the different wait conditions?
something like:
wait(){
  cy.wait(GET_CONFIG_ALIAS);
  if (isPricingEnabled) {
     waitForPricing();
   }
}
this would make it shorter (line 635ff, line 652ff)and easier to maintain for future adaptions  since we call only 

wait() 

here instead of 

cy.wait(GET_CONFIG_ALIAS);
   if (isPricingEnabled) {
     waitForPricing();
   }

Uli-Tiger: good point

Stofftiger (approved):

PR #16947, opened 2023-02-28
changes for User Guard

anjana-bl:


integration-libs/cdc/root/guards/cdc-user.guard.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+import { UrlTree } from '@angular/router';
+import { RouterTestingModule } from '@angular/router/testing';
+import { StoreModule } from '@ngrx/store';
+
+import {
+  GlobalMessageService,
+  GlobalMessageType,
+  SemanticPathService,
+} from '@spartacus/core';
+import { CdcB2BUserService } from 'integration-libs/cdc/organization/administration/cdc-b2b-user.service';
```
Matejk00: Let's import this service directly from separate sub-library `@spartacus/cdc/organization/administration`.

integration-libs/cdc/root/guards/cdc-user.guard.ts
```diff
@@ -0,0 +1,44 @@
...
+
+import { Injectable } from '@angular/core';
+import { CanActivate, Router, UrlTree } from '@angular/router';
+import {
+  GlobalMessageService,
+  GlobalMessageType,
+  SemanticPathService,
+} from '@spartacus/core';
+import { UserGuard } from '@spartacus/organization/administration/core';
+import { CdcB2BUserService } from '../../organization/administration/public_api';
```
Matejk00: Shouldn't we import this service from `@spartacus/cdc/organization/administration` instead of the relative path?

integration-libs/cdc/root/guards/cdc-user.guard.ts
```diff
@@ -0,0 +1,44 @@
...
+  constructor(
+    protected globalMessageService: GlobalMessageService,
+    protected cdcb2bUserService: CdcB2BUserService,
+    protected semanticPathService: SemanticPathService,
+    protected router: Router
+  ) {
+    super(globalMessageService, cdcb2bUserService, semanticPathService, router);
+  }
+
+  canActivate(): boolean | UrlTree {
```
Matejk00, line 16: If the logic of `canActivate` method is the same as in the original `UserGuard` I think there is no need to duplicate the code.

What about to use something like this?
```
  canActivate(): boolean | UrlTree {
    return super.canActivate();
  }
```

or even I hope we can remove ` canActivate()` implementation from that file and leave only such part:

```
@Injectable({
  providedIn: 'root',
})
export class CdcUserGuard extends UserGuard {
  constructor(
    protected cdcb2bUserService: CdcB2BUserService,
    protected semanticPathService: SemanticPathService,
    protected router: Router
  ) {
    super(globalMessageService, cdcb2bUserService, semanticPathService, router);
  }
}
```

Matejk00 (changes requested):

integration-libs/cdc/root/guards/cdc-user.guard.ts
```diff
@@ -0,0 +1,44 @@
...
+import { UserGuard } from '@spartacus/organization/administration/core';
+import { CdcB2BUserService } from '../../organization/administration/public_api';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class CdcUserGuard extends UserGuard implements CanActivate {
+  constructor(
+    protected globalMessageService: GlobalMessageService,
+    protected cdcb2bUserService: CdcB2BUserService,
```
Matejk00: Maybe let's rename this parameter to `cdcB2bUserService`. Just to keep naming convention consistency.

integration-libs/cdc/root/guards/cdc-user.guard.spec.ts
anjana-bl: done

integration-libs/cdc/root/guards/cdc-user.guard.ts
anjana-bl: done

integration-libs/cdc/root/guards/cdc-user.guard.ts
anjana-bl: it will give Entry point @spartacus/cdc/organization/administration has a circular dependency on @spartacus/cdc/root error during build, hence not importing like @spartacus/cdc/organization/administration

integration-libs/cdc/root/guards/cdc-user.guard.ts
anjana-bl: I just realized injecting CdcB2BUserService into this guard is creating a circular dependency within cdc modules during build. hence i have changed the implementation. Can you check it again

Matejk00: This guard fires only on administration/organization pages (b2b). Personally, I am not convenienced if we should keep this guard responsible for such logic outside `@spartacus/cdc/organization/administration` sub-library.

As you mentioned it also produces some potential circular dependencies between libs.

My idea is to move this guard into `organization/administration/guards` folder and then try to overwrite the config 
 in `cdc-administration.module.ts` to use `CdcUserGuard` instead of `UserGuard`.

```
      cmsComponents: {
        ManageUsersListComponent: {
          providers: [
            {
              provide: UserListService,
              useExisting: CdcUserListService,
            },
            userCmsConfig.cmsComponents?.ManageUsersListComponent?.providers ||
              [],
          ],
          childRoutes: {
            children: [
              {
                path: 'create',
                component: UserFormComponent,
                canActivate: [CdcUserGuard],
              },
              {
                path: `:${ROUTE_PARAMS.userCode}`,
                component: UserDetailsComponent,
                data: {
                  cxPageMeta: { breadcrumb: 'orgUser.breadcrumbs.details' },
                },
                children: [
                  {
                    path: `edit`,
                    component: UserFormComponent,
                    canActivate: [CdcUserGuard],
                  },
                ],
              },
            ],
          },
        },
      },
    }),
```

By this `root` still will be responsible for stuff not related with any other sub-library.

Matejk00 (changes requested):

integration-libs/cdc/root/guards/cdc-user.guard.ts
Matejk00: But also I am open for other ideas @anjana-bl

anjana-bl: Hi Mateauz, thanks for the sample code. But if we do like this, then we have to copy the entire content of user.config.ts . Else, the remaining navigations wont work, right? eg: we cannot navigate to 'approvers' from user details screen. Is it okay if I copy the entire content of user.config.ts into a new file, say, cdc-user.config.ts and make necessary changes (eg: replace UserGuard with CdcUserGuard) and use it?

anjana-bl:
Closing this PR, as the issue is fixed in the core module itself via https://github.com/SAP/spartacus/pull/17009. So this extra piece of code is not necessary and the same functionality is working fine with existing code.

PR #16953, opened 2023-02-28
refactor: Standardize LOGs output in SSR for better debuggability in production

WeizhengSap:
ticket: cxspa-2251


core-libs/setup/ssr/optimized-engine/ssr-optimization-options.ts
```diff
@@ -123,5 +128,6 @@ export const defaultSsrOptimizationOptions: SsrOptimizationOptions = {
   forcedSsrTimeout: 60_000,
   maxRenderTime: 300_000,
   reuseCurrentRendering: true,
-  debug: false,
+  debug: true,
```
WeizhengSap: will revert it back

core-libs/setup/ssr/optimized-engine/augmented-request.ts
```diff
@@ -0,0 +1,16 @@
...
+ */
+
+export {};
+
+declare global {
+  namespace Express {
+    export interface Request {
+      uuid?: string; // request uuid for debug tracking
+      startTime?: number; // start timestamp of renderring
+    }
```
Platonn, line 14: For data related only to the _render_ (but not to Request), I'd suggest to provide them to the Angular App with ordinary providers. Or if we _really_ need to store it in the Request, object, I suggest to use a dedicated property for it: Response.locals https://expressjs.com/en/api.html#res.locals . However, I'd put data there only if _really_ justified.

Please note that Requests are not mapped to renders 1:1. In case when `reuseCurrentRendering` is enabled, then many **parallel** Requests for the **same** URL might reuse the same 1 pending render that was started when the first request in the row came.

core-libs/setup/ssr/optimized-engine/logging.interceptor.ts
```diff
@@ -0,0 +1,57 @@
...
+import { Inject, Injectable, Optional } from '@angular/core';
+import { RESPONSE } from '@nguniversal/express-engine/tokens';
+import { SSR_LOG_BEFORE_TIMEOUT, SSR_REQUEST_LOGGING } from '@spartacus/core';
+import { Response } from 'express';
+import { Observable } from 'rxjs';
+import { tap } from 'rxjs/operators';
+import { RequestLoggingService } from './request-logging.service';
+
+@Injectable()
+export class LoggingInterceptor implements HttpInterceptor {
```
Platonn, line 23: This interceptor is a separate feature, not related to just providing a structured JSON logging format.
So for clarity of git history, I'm in favor of introducing this interceptor in a separate feature PR.

core-libs/setup/ssr/optimized-engine/logging.interceptor.ts
```diff
@@ -0,0 +1,57 @@
...
+    return next.handle(request).pipe(
+      tap(
+        (_event) => {
+          if (_event.type === HttpEventType.Response) {
+            if (this.res?.headersSent) {
+              this.logger?.log(request.url + ' FINISHED AFTER TIMEOUT');
+            } else {
+              if (this.logBeforeTimeout) {
+                this.logger?.log(request.url + ' FINISHED BEFORE TIMEOUT');
+              }
```
Platonn, line 47: IMHO it's shouldn't be the responsibility of the http interceptor of the outgoing request to recognize and print whether the original SSR request did already timeout. It's too much coupling of unrelated concepts. This interceptor should at most log the outgoing requests, by using a special Logger. 
And it's the Logger who should have the capabilities of attaching useful context (e.g. whether the orignial SSR request has already timed out or not). It could be potentially a responsibility of the core Logger to attach more context to the logged message, (alternatively it could be responsibility of an extension feature of Logger - IMHO Logger should be extendable/composable to allow for eaisly adding more context data in the future, done by us or by customers)

```ts
// logger.ts
console.log(
  {
     message: ...
     request: {
        url: ...
     },
     response: {
        // this is an information whether the SSR reuqest has been already handled
        // (e.g. with a CSR fallback):
        headersSent: ... 
     }
  }
)
```

Thanks to moving the responsibility of the logger's context population to the Logger itself (or it's extensions), the logger will provide this ambient context also when used in other places than this interceptor.

In the end of the day, when we remove the context-providing responsibility from this interceptor's code, it will be responsible only for printing the fact that the outgoing request has succeeded or failed. IMHO such a verbose interceptor should be hidden by some feature flag like `debug`

core-libs/setup/ssr/optimized-engine/request-logging.service.ts
```diff
@@ -0,0 +1,24 @@
...
+  constructor(@Optional() @Inject(REQUEST) private req?: Request) {}
+
+  log(message: string) {
+    const time =
+      this.req?.startTime !== undefined
+        ? Date.now() - this.req.startTime
+        : new Date();
+
+    // eslint-disable-next-line no-console
+    console.log(this.req?.uuid + ' - ' + message + ' (ms:' + time + ')');
```
Platonn, line 22: My intention was to print logs in a standardized JSON format, so the logs are structurized, machine-readable and extensible (open for future additions of contextual information).

I'm not sure how logs are presented in Kibana, when logs are multi-line. So to be discussed with Florian. If printing `console.log(object)` prints many lines of text and is splitted into separate log entries in Kibana (not sure if it's the case!), then we can potentially avoid multi-line outputs, by simply printing a stringified JSON  `console.log(JSON.stringify(object))`

### Note about backward comaptibility
The new format of printed logs should be hidden by a feature flag (if released in minor), because it's a breaking changes for customers that already have some custom automation/alert tools that parse those messages (e.g. with regular expressions).

The existing flag `ssrOptions.debug` has a different purpose - verbosity and bigger amount of logged messages. But the formatting of messages is a distinct feature, independent from the `ssrOptions.debug` flag.

core-libs/setup/ssr/optimized-engine/request-logging.service.ts
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { Inject, Injectable, Optional } from '@angular/core';
+import { REQUEST } from '@nguniversal/express-engine/tokens';
+import { Request } from 'express';
+
+@Injectable()
+export class RequestLoggingService {
```
Platonn, line 12: I'd like to build one common Logger that could be used both in OptimizedSrrEngine and inside Angular app.
It should allow for logging/warning/erroring anything, just as console.log would do.

IMHO the logged message should be ideally in the JSON format. But to be verified with Florian how it would behave in Kibana logs.

On top of it, I'd like to be able to provide additional context to the message with composable extensions of the Logger.

E.g. when running in SSR, we could plug an extension of the Logger that populates the context of each logged message with the data specific to the Request (e.g. URL and possibly requestID in the future for distributed tracing) and the particular render job (e.g. uuid of the render, and its timeStart timestamp).

Example JSON properties from top of my head:
- message
- request
   - url
   - (add in the future: request ID for distributed tracing, For more see https://jira.tools.sap/browse/CXSPA-2067. This data is not yet available in Request)
- render
    - timeStart (timestamp)
    -  uuid (unique UUID of the Angular SSR render; don't confuse with request ID meant for distributed tracing). This is might be useful for customers only as long as we don't have a better property available - request ID

WeizhengSap: This PR is just like a POC. We can create a mature `LoggingService` in core, and inject it in this `RequestLoggingService`. I didn't consider the output format, because it depends on requirements. JSON is fine, but it may make the log output very big.

core-libs/setup/ssr/optimized-engine/augmented-request.ts
WeizhengSap: `uuid` and `startTime` is for request, not for render.
```
  protected renderResponse(
    filePath: string,
    options: any,
    callback: SsrCallbackFn
  ): void {
    const request: Request = options.req;
    const response: Response = options.res || options.req.res;

    request.uuid = crypto.randomUUID();
    request.startTime = Date.now();
    const logger = new RequestLoggingService(request);
...
```

core-libs/setup/ssr/optimized-engine/logging.interceptor.ts
WeizhengSap: In the ticket, we want to log the occ requests as well, right? I think, it is the requirement.

PR #16955, opened 2023-02-28
CXSPA-2114: Reduce CI workload.

plabadie:
https://jira.tools.sap/browse/CXSPA-2114

### DIFF
--- a/.github/workflows/ci-continuous-integration.yml
+++ b/.github/workflows/ci-continuous-integration.yml
@@ -1,6 +1,10 @@
 on:
   pull_request:
     types: [opened, synchronize]
+    branches:
+      - develop
+      - release/**
+      - epic/**
   workflow_dispatch:
     # empty as it is used only to manually trigger the workflow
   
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-continuous-integration.yml
```diff
@@ -1,6 +1,10 @@
 on:
   pull_request:
     types: [opened, synchronize]
+    branches:
```
RadhepS, line 4: if i understand this correctly we only run e2es when PRs are going into develop/release/epic?
is this as intended and is there any way to run E2Es when trying to merge a PR into a non develop branch? 
for example this PR: https://github.com/SAP/spartacus/pull/16962

plabadie: The purpose of this change is not for e2es.  They already only run for PRs to develop and maintenance.  
Rather, this change is to only trigger the CI checks on PRs that target develop, maintenance branches and epic branches.  

What you can observe in your test PR, is that indeed, the CI checks were not triggered.
On this PR, the CI checks were triggered since the PR targets develop.

RadhepS (approved):

PR #16979, opened 2023-03-03
test: reduce flakiness by waiting for pricing requests (CXSPA-2745)

Uli-Tiger:


ChristophHi (changes requested):
I don't think this is correct

`commerceRelease.isPricingEnabled = responseBody.pricingEnabled;`
    
in last weeks PR https://github.com/SAP/spartacus/pull/16933 (merged, approved by @Stofftiger ).
In 22.05, when pricing was always active, this would evaluate to false, since the OCC property was only introduced for 22.11.
Please fix although not part of this PR

Uli-Tiger:
> I don't think this is correct
> 
> `commerceRelease.isPricingEnabled = responseBody.pricingEnabled;`
> 
> in last weeks PR #16933 (merged, approved by @Stofftiger ). In 22.05, when pricing was always active, this would evaluate to false, since the OCC property was only introduced for 22.11. Please fix although not part of this PR

ah ok - changed the default to true

ChristophHi (approved):

PR #16994, opened 2023-03-06
fix: update how disabled action works on anchor tag to not break build

RadhepS:
closes: https://jira.tools.sap/browse/CXSPA-2755


FollowTheFlo (approved):
looks good.

PR #16998, opened 2023-03-06
CXSPA-1236: Embedded script compliance

plabadie:


feature-libs/storefinder/core/service/google-map-renderer.service.ts
```diff
@@ -35,18 +36,31 @@ export class GoogleMapRendererService {
...
-            this.drawMap(mapElement, locations, selectMarkerHandler);
-          },
-        });
-      } else {
-        this.drawMap(mapElement, locations, selectMarkerHandler);
+    if (this.config.googleMaps?.apiKey) {
+      if (Object.entries(locations[Object.keys(locations)[0]]).length > 0) {
+        if (this.googleMap === null) {
+          const apiKey =
+            this.config.googleMaps.apiKey === GOOGLE_MAPS_DEVELOPMENT_KEY_CONFIG
```
FollowTheFlo, line 43: from discussion with Patrick: we keep the dev key available also when Spartacus run in prod mode. It allows our demo instances to display the map.
Compliancy requirement to 'NOT call any third-party server when api key is NOT provided' is fulfilled.

FollowTheFlo (approved):
looks good, tested it sucefully, some code logic points has been discussed.
Note that our demo instances won't have the google map displayed unless the config is added manually in app.module.ts, that could eventually be adjusted at install script level

PR #17009, opened 2023-03-08
fix(CDC,MyCompany): provide UserGuard in child injector and resolve UserGuard using CmsGuardsService

Platonn:
### Root cause of the bug 
When Angular's `Router` tries to inject a guard's instance it only performs lookup inside the root injector (for the guard as well as it's dependencies). See [Angular Router's source code](https://github.com/angular/angular/blob/61e69c87e6b577f1c5a71fb5c39c3bd6125c4b91/packages/router/src/operators/check_guards.ts#L65) - for injecting guards, the `EnvironmentInjector` is used, which in our case is simply the root injector.

In our code, when we tried to over-provide the `B2BUserService` only inside the child injector of a lazy loaded module, this over-provider was invisible to Angular's `Router`, because Angular's `Router` is not able to look up the child injector, but only the root injector.

It's is depcited in the following diagram:

image

### Solution
2 steps needed in conjunction:
1. Wrap `UserGuard` within a function that is able to lookup the child injector, and only then pass it to the `Route` object. Under the hood wrapper function it uses `CmsGuardsService` to inject the guard, because this service is capable of injecting the guards instances the child injectors of lazy-loaded modules.
2. Moreover, now we provide `UserGuard` in the child injector of a lazy-loaded module, instead of `providedIn: 'root'`. Thanks to that, the over-provider of `B2BUserService` (siting in the same child injector) is taken into account.

It's is depcited in the following diagram:

image

### Changes by the way
Similarly to `UserGuard`, also provided `AdminGuard` in the child injector of lazy-loaded module, instead of `providedIn: 'root'`.

related to https://jira.tools.sap/browse/CXSPA-2377


projects/storefrontlib/cms-structure/services/cms-guards.service.spec.ts
```diff
@@ -136,5 +138,68 @@ describe('CmsGuardsService', () => {
...
+      guards.push(PositiveGuard, NotGuard, PositiveGuardObservable);
+
+      expect(() => {
+        service.cmsPageCanActivate(
+          [],
+          mockActivatedRouteSnapshot,
+          mockRouterStateSnapshot
+        );
+      }).toThrowError('Invalid CanActivate guard in cmsMapping');
+    });
```
Platonn, line 152: Here just added one missing unit tests for the existing behavior. It's not a new behavior.

projects/storefrontlib/cms-structure/services/cms-routes-impl.service.ts
```diff
@@ -4,24 +4,33 @@
...
+import { CmsGuardsService } from './cms-guards.service';
 
 // This service should be exposed in public API only after the refactor planned in https://github.com/SAP/spartacus/issues/7070
 @Injectable({ providedIn: 'root' })
 export class CmsRoutesImplService {
   constructor(
     private router: Router,
-    private cmsComponentsService: CmsComponentsService
+    private cmsComponentsService: CmsComponentsService,
+    private cmsGuardsService: CmsGuardsService
```
Matejk00, line 33: Isn't recommended to use `protected` instead of `private` for extensibility purposes?

Matejk00: Also, I am thinking, if shouldn't we document this change in our 6.0 migration doc.

Matejk00 (approved):

projects/storefrontlib/cms-structure/services/cms-routes-impl.service.ts
Platonn: `CmsRoutesImplService` is deliberately in private API of Spartacus. So no need to document breaking change in this case.

Platonn: Also in this exceptional case when the whole service is private, all params can be `private` it doesn't matter. Even more, it expresses deliberately the methods are not intended as extension points.

Platonn:
### Root cause of the bug 
When Angular's `Router` tries to inject a guard's instance it only performs lookup inside the root injector (for the guard as well as it's dependencies). See [Angular Router's source code](https://github.com/angular/angular/blob/61e69c87e6b577f1c5a71fb5c39c3bd6125c4b91/packages/router/src/operators/check_guards.ts#L65) - for injecting guards, the `EnvironmentInjector` is used, which in our case is simply the root injector.

In our code, when we tried to over-provide the `B2BUserService` only inside the child injector of a lazy loaded module, this over-provider was invisible to Angular's `Router`, because Angular's `Router` is not able to look up the child injector, but only the root injector.

It's is depcited in the following diagram:

image


### Solution
2 steps needed in conjunction:
1. Wrap `UserGuard` within a function that is able to lookup the child injector, and only then pass it to the `Route` object. Under the hood wrapper function it uses `CmsGuardsService` to inject the guard, because this service is capable of injecting the guards instances the child injectors of lazy-loaded modules.
2. Moreover, now we provide `UserGuard` in the child injector of a lazy-loaded module, instead of `providedIn: 'root'`. Thanks to that, the over-provider of `B2BUserService` (siting in the same child injector) is taken into account.

It's is depcited in the following diagram:

image

### Changes by the way
Similarly to `UserGuard`, also provided `AdminGuard` in the child injector of lazy-loaded module, instead of `providedIn: 'root'`.

related to https://jira.tools.sap/browse/CXSPA-2377

PR #17013, opened 2023-03-09
Replace GoogleFont by npm library Fontsource

FollowTheFlo:
This change is needed for compliance reason, to avoid loading Fonts form third-party server (Google in our case).
Now the fonts are self-hosted, installed with npm https://www.npmjs.com/package/@fontsource/open-sans (MIT license + licence for Open-Sans -> SIL Open Font License)
data added:
- the css font-face objects, pointing to woff2 file url, added in style.css 
- 4 woff2 files (lazy loaded, most of the time, only 3 needed)
See testing done on https://jira.tools.sap/browse/CXSPA-525
Bundle is slightly larger than before
Google Font: 39kb
FontSource: 4x16=64k
image


projects/storefrontstyles/scss/theme/sparta/_fonts.scss
```diff
@@ -1 +1,4 @@
-@import url('#{$font-url}');
+@import '@fontsource/open-sans/300.css';
```
FollowTheFlo, line 1: added 4 files of family 'open-sans' in weight: 300,400,600,700 to mirror the font that were previously downloaded on Google Fonts: https://fonts.googleapis.com/css?family=Open+Sans:300,400,600,700

plabadie (changes requested):
PR looks good, but the CI fails on unit tests.  It is likely that some unit test snapshots have to be updated because some of them contain the list of dependencies (which just got updated with a new lib)

projects/schematics/src/ng-add/__snapshots__/index_spec.ts.snap
```diff
@@ -56,6 +56,7 @@ exports[`Spartacus Schematics: ng-add should add spartacus properly with SSR 2`]
     "@angular/platform-server": "^15.2.0",
     "@angular/router": "^15.2.0",
     "@angular/service-worker": "^15.2.0",
+    "@fontsource/open-sans": "^4.5.14",
```
FollowTheFlo, line 59: generated with command:
npm --prefix projects/schematics run test --runInBand -- --u

plabadie (approved):

PR #17018, opened 2023-03-09
ASM Customer List Facade refactor

niehuayang:
JIRA: https://jira.tools.sap/browse/CXSPA-2409


feature-libs/asm/core/connectors/asm.adapter.ts
```diff
@@ -9,7 +9,7 @@ import { Observable } from 'rxjs';
 import {
   CustomerSearchOptions,
   CustomerSearchPage,
-} from '../models/asm.models';
+} from '../../root/model/asm.models';
```
ercultimate: This import is invalid.  It should be `@spartacus/asm/root`.  There are many other instances of this.  I'd recommend running a search for "../root/model" in the asm folder.

This file (`asm.adapter.ts`) is part of the `@spartacus/asm/core` module.  Relative imports will only work for files that are part of the same module.  The `asm.models.ts` file was moved to the `@spartacus/asm/root` module, so we need to use a fully qualified import.  The auto-import is not smart enough to recognize this and gave you the same bad import for all the files.

This issue would only appear once you tried to compile the Spartacus project into the Angular library using `npm build:libs`.

ercultimate (changes requested):
This is looking pretty good.  Just the imports to fix and then documenting the API changes in the migration document.

feature-libs/asm/core/connectors/asm.adapter.ts
niehuayang: Ah, I see, thanks very much for providing more details, and now I have corrected it.

feature-libs/asm/core/facade/asm.service.ts
```diff
@@ -53,31 +53,6 @@ export class AsmService {
...
-   * Returns the customer list customers search result loading status.
-   */
-  getCustomerListCustomersSearchResultsLoading(): Observable<boolean> {
-    return this.store.pipe(
-      select(AsmSelectors.getCustomerListCustomersSearchResultsLoading)
-    );
-  }
-
   /**
    * Returns the customer list customers search result error status.
```
niehuayang: Hi @ercultimate , It looks like the method should also move to AsmCustomerListFacade? Am I right?

feature-libs/asm/core/connectors/asm.adapter.ts
niehuayang: > This is looking pretty good. Just the imports to fix and then documenting the API changes in the migration document.

the API changes have been added in the migration document.

feature-libs/asm/core/facade/asm.service.ts
ercultimate: Yes, you are correct!  I think this method was added later and might have been left out of the ticket description.

ercultimate (changes requested):

feature-libs/asm/core/facade/asm.service.ts
niehuayang: Okay, I have changed it

ercultimate (dismissed):

niehuayang:
Hi @ercultimate , When I want to merge to develop,  the code conflicts appear, I have solved the conflicts, could you please approve it again?

kimhw0630 (approved):

PR #17038, opened 2023-03-14
fix: restart dialog shall not open when commerce feature toggle is disabled (CXSPA-2819)

Uli-Tiger:


feature-libs/product-configurator/rulebased/occ/variant/converters/occ-configurator-variant-normalizer.ts
```diff
@@ -44,7 +44,7 @@ export class OccConfiguratorVariantNormalizer
       pricingEnabled: source.pricingEnabled ?? true,
       hideBasePriceAndSelectedOptions: source.hideBasePriceAndSelectedOptions,
       immediateConflictResolution: source.immediateConflictResolution ?? false,
-      newConfiguration: source.newConfiguration ?? true,
+      newConfiguration: source.newConfiguration, // we need a trinary state true, false, undefined!
```
Larisa-Staroverova, line 47: Why not set `newConfiguration` to `false` if `undefined`?
`newConfiguration: source.newConfiguration ?? false`

feature-libs/product-configurator/rulebased/occ/variant/converters/occ-configurator-variant-normalizer.spec.ts
```diff
@@ -373,16 +373,22 @@ describe('OccConfiguratorVariantNormalizer', () => {
...
     });
 
-    it('should convert "newConfiguration" setting by default to true', () => {
+    it('should convert "newConfiguration" setting to false', () => {
+      configuration.newConfiguration = false;
+      const result = occConfiguratorVariantNormalizer.convert(configuration);
+      expect(result.newConfiguration).toBe(false);
+    });
+
+    it('should convert "newConfiguration" setting by default to undefined', () => {
```
Larisa-Staroverova, line 388: If you set: 
`newConfiguration: source.newConfiguration ?? false`
then
you don't need this test anymore

feature-libs/product-configurator/rulebased/core/state/reducers/configurator.reducer.ts
```diff
@@ -127,7 +127,9 @@ function handleActionUpdateConfigurationFinalizeSuccess(
   checkConflictSolverDialog(result);
   result.isCartEntryUpdateRequired = true;
   result.overview = undefined;
-  result.interactionState.newConfiguration = false;
+  if (state.interactionState.newConfiguration !== undefined) {
```
Larisa-Staroverova, line 130: If you set: 
`newConfiguration: source.newConfiguration ?? false`
then you could adjust this condition to
` if (state.interactionState.newConfiguration)`

Larisa-Staroverova (changes requested):
@Uli-Tiger, could you have a look at my comments?

feature-libs/product-configurator/rulebased/occ/variant/converters/occ-configurator-variant-normalizer.ts
Uli-Tiger: then you would get the restart-dialog even if you enter a default configuration see also configurator-fotm-component
```
   filter(
          (configuration) =>
            configuration.interactionState.newConfiguration === false
        )
```

feature-libs/product-configurator/rulebased/occ/variant/converters/occ-configurator-variant-normalizer.spec.ts
Uli-Tiger: see my previous comment

feature-libs/product-configurator/rulebased/core/state/reducers/configurator.reducer.ts
Uli-Tiger: you have to see this condition together with whats happening in the create success action.
in the create success action we copy the newConfiguration state into the interaction state. Here we reset the newConfiguration state in the interaction state. This is required to force the popup in case we re-enter the configuration in the same session, as the configuration is then fetched from the local ngrx state instead of from the back-end.

Larisa-Staroverova (approved):

PR #17050, opened 2023-03-15
fix intermittent b2b-unit-level-order e2e failure

FollowTheFlo:
Made the e2e compatible with Testisolation.
The failure test is now repeatable, it can be retried multiple times, which was not the case before.
Number of test about 'sorted list' has been decreased from 4 tp 2.
All the test of this spec are now compatible with TestIsolation.


projects/storefrontapp-e2e-cypress/cypress/e2e/b2b/regression/unit-level-order/b2b-unit-level-order-history.e2e.cy.ts
```diff
@@ -30,18 +29,13 @@ describe('B2B - Unit-Level Orders History', { testIsolation: false }, () => {
   let order0;
   let order1;
   let order2;
-  isolateTests();
   before(() => {
-    cy.window().then((win) => win.sessionStorage.clear());
```
FollowTheFlo, line 35: tesIsolation takes care of it

projects/storefrontapp-e2e-cypress/cypress/e2e/b2b/regression/unit-level-order/b2b-unit-level-order-history.e2e.cy.ts
```diff
@@ -168,22 +165,6 @@ describe('B2B - Unit-Level Orders History', { testIsolation: false }, () => {
       checkThatOrdersAreSorted(fieldBuyer);
     });
 
-    it("can select 'Buyer (Descending)' to see the list sorted descending by unit name", () => {
```
FollowTheFlo, line 171: remove 2 sorted tests out of 4 to keep it short, I don't think we need to test all of them.

projects/storefrontapp-e2e-cypress/cypress/e2e/b2b/regression/unit-level-order/b2b-unit-level-order-history.e2e.cy.ts
```diff
@@ -137,7 +134,7 @@ describe('B2B - Unit-Level Orders History', { testIsolation: false }, () => {
   });
 
   describe('User who wants to sort order history', () => {
-    before(() => {
+    beforeEach(() => {
```
FollowTheFlo, line 137: needed as testisolation reset the navigation, test are now repeatable, important when a retry is needed.

RadhepS (approved):

PR #17051, opened 2023-03-15
fix: Fix future stock making http calls as anonymous user

Zeyber:
Closes: https://jira.tools.sap/browse/CXSPA-2826


feature-libs/product/future-stock/core/services/future-stock.service.spec.ts
```diff
@@ -9,14 +14,20 @@ class MockFutureStockConnector implements Partial<FutureStockConnector> {
...
-const mockUserId = 'current';
+const mockUserId = OCC_USER_ID_CURRENT;
 
 class MockUserIdService implements Partial<UserIdService> {
   takeUserId(): Observable<string> {
     return of(mockUserId);
   }
 }
 
+class MockUserIdAnonymousService implements Partial<UserIdService> {
```
bgambocjaviniar, line 25: totally fine to me as it's a way of testing

----

I would have preferred making use of the createSpy().

What would you do to make things work instead of configuring the TestBed on demand would be the following:

1 - keep the TestBed module configuration in the before each as it was before

2 - make use of the spy for the `takeUserId`

```
class MockUserIdService implements Partial<UserIdService> {
  takeUserId = createSpy().and.returnValue(of(mockUserId));
}
```

3 - inject the service in question

```
userIdService = TestBed.inject(UserIdService)
```

4 - in the describe/context in question, you would change the return value as follows

```
userIdService.takeUserId = createSpy().and.returnValue(of(theMockUserInQuestion));
```

feature-libs/product/future-stock/core/services/future-stock.service.ts
```diff
@@ -16,14 +20,17 @@ import { ProductFutureStock } from '../model';
...
-  protected futureStockState$: Observable<ProductFutureStock> =
+  protected futureStockState$: Observable<ProductFutureStock | undefined> =
     this.routingService.getRouterState().pipe(
       withLatestFrom(this.userIdService.takeUserId()),
       switchMap(([{ state }, userId]) => {
-        return this.futureStockConnector.getFutureStock(
-          userId,
-          state.params[this.PRODUCT_KEY]
-        );
+        if (userId === OCC_USER_ID_CURRENT) {
```
bgambocjaviniar: let's change the condition to 

`userId !== OCC_USER_ID_ANONYMOUS`

reason being, it will not always be `current`. It may be a customer emulation user, which is `customerId` I believe.
Also, if customer changes the endpoint to not use current, but the actual userId to make the request, then they won't be able to see the future stocks too.

bgambocjaviniar (changes requested):
1 - testing comment is something you can change if you want, but don't have to

2 - condition in the actual service needs to be changed

feature-libs/product/future-stock/core/services/future-stock.service.spec.ts
Zeyber: I tried this method using spies but it would not take effect. I am guessing it is because of the usage of `takeUserId()` within the pipe that doesn't allow it to be "called" per se and therefore the spy will not return its value.

bgambocjaviniar (approved):

feature-libs/product/future-stock/core/services/future-stock.service.spec.ts
bgambocjaviniar: > 

future references as I did what I mentioned above on this recent merged commit https://github.com/SAP/spartacus/commit/f6b176dbe06d6677750e4cbd23c6462c3e023972#diff-a3a961dff01869c8cea9fc6650d0e7df092943420402eacee9eb816017f0b041R106

bgambocjaviniar: > I tried this method using spies but it would not take effect. I am guessing it is because of the usage of `takeUserId()` within the pipe that doesn't allow it to be "called" per se and therefore the spy will not return its value.

future references as I did what I mentioned above on this recent merged commit https://github.com/SAP/spartacus/commit/f6b176dbe06d6677750e4cbd23c6462c3e023972#diff-a3a961dff01869c8cea9fc6650d0e7df092943420402eacee9eb816017f0b041R106

PR #17056, opened 2023-03-15
feat: set default timeout 20s for outgoing requests in SSR

Platonn:
Set a default timeout for outgoing http calls in SSR to 20s.

Note: the timeout mechanism was introduced before in PR https://github.com/SAP/spartacus/pull/16658

closes https://jira.tools.sap/browse/CXSPA-2234


bgambocjaviniar (approved):

projects/core/src/http/http-timeout/default-http-timeout.config.ts
```diff
@@ -9,8 +9,7 @@ import { OccConfig } from '../../occ/config/occ-config';
 export const defaultBackendHttpTimeoutConfig: OccConfig = {
   backend: {
     timeout: {
-      // [CXSPA-2234] (breaking change): uncomment it only in the next major (v6.0?):
-      // server: 10_000,
+      server: 10_000,
```
bgambocjaviniar: shouldn't it be 20_000 for 20s

bgambocjaviniar (changes requested):
see comment

projects/core/src/http/http-timeout/default-http-timeout.config.ts
hackergil: I think so ;)

Platonn: 😅 Initially I thought 10s, but then decided to be more generous with 20s. However it seems I didn’t commit my change. 🤦 
Now pushed update 20_000.

bgambocjaviniar (approved):

PR #17060, opened 2023-03-15
chore: Update full version in styles lib 

bgambocjaviniar:


projects/storefrontstyles/scss/_versioning.scss
```diff
@@ -23,14 +23,14 @@
 // Any CSS added later than this version is omitted by default, unless the `$useLatestStyles`
 // or `$styleVersion` are enabled.
 $_fullVersion: (
-  major: 5,
+  major: 6,
```
plabadie, line 26: Minor should be 0

plabadie (approved):

PR #17061, opened 2023-03-15
fix intermittent cart.e2e failure

FollowTheFlo:
Make the test repeatable so 'retry' works as expected


projects/storefrontapp-e2e-cypress/cypress/e2e/regression/cart/cart.e2e.cy.ts
```diff
@@ -20,8 +20,7 @@ describe('Cart', () => {
     });
 
     context('Registered user', () => {
-      before(() => {
-        cy.window().then((win) => win.sessionStorage.clear());
```
FollowTheFlo, line 24: not  needed as testIsolation clears cookies/localStorage/sessionStorage before

projects/storefrontapp-e2e-cypress/cypress/e2e/regression/cart/cart.e2e.cy.ts
```diff
@@ -20,8 +20,7 @@ describe('Cart', () => {
...
+      beforeEach(() => {
```
FollowTheFlo, line 23: if test fails, the following attempt will execute beforeEach hook which make the test independant and repeatable. Previous behavior, attempt 2 and 3 were stuck on blank window, as testIsolation triggers navigation to blank page before each test.

bgambocjaviniar (approved):
got on a call, and said we can merge this 'hotfix' in, but if it will again, then we would need to refactor the test

PR #17062, opened 2023-03-16
Adjust overview menu styling calculation

Larisa-Staroverova:


feature-libs/product-configurator/rulebased/components/overview-menu/configurator-overview-menu.component.ts
```diff
@@ -16,6 +16,11 @@ import { ICON_TYPE } from '@spartacus/storefront';
 import { Configurator } from '../../core/model/configurator.model';
 import { ConfiguratorStorefrontUtilsService } from '../service/configurator-storefront-utils.service';
 
+export interface ConfiguratorOverviewStyle {
+  property: string;
+  value: string;
+}
```
Uli-Tiger: should't we move this to the models file, where all our interface are defined?

Uli-Tiger: anyways, in my opinion an named read-only inline tuple better represents whats really meant here. We just want to pass along 2 values (property and value) always together.
` styles: readonly [property: string, value: string][] `
https://www.w3schools.com/typescript/typescript_tuples.php

example usage
```
  protected changeStyling(
    querySelector: string,
    styles: readonly [property: string, value: string][]
  ): void {
    styles?.forEach((style) => {
      this.configuratorStorefrontUtilsService.changeStyling(
        querySelector,
        style[0],
        style[1]
      );
    });
  }
```

Uli-Tiger (changes requested):

feature-libs/product-configurator/rulebased/components/overview-menu/configurator-overview-menu.component.ts
Larisa-Staroverova: It is a good idea. 
Done!

Uli-Tiger (approved):

PR #17066, opened 2023-03-16
fix: click on the product search result list in quick order in Safari

kimhw0630:
Closes: https://jira.tools.sap/browse/CXSPA-2786

When a user clicks on a product, two events are triggered: `blur` and `click`. The `blur` event is fired before the `click` event, and if there is code that hides the element during the `blur` event, then the `click` event will not be triggered.
Changed the click event to the `mousedown` event, which fires before `blur`.


FollowTheFlo (approved):
looks good.

plabadie (approved):

PR #17069, opened 2023-03-17
fix: CXSPA-2587 Lack of state information when action performed

pawelfras:
Closes: https://jira.tools.sap/browse/CXSPA-2587


kimhw0630:
This might be optional but in accordion, you may need `aria-controls `and `aria-labelledby`.
So, in the 
button: aria-controls(value with content div id) and id attributes
On the content div: id and aria-labelledby(value with button id)

kimhw0630 (dismissed):

kimhw0630 (approved):

psmul (approved):

PR #17073, opened 2023-03-17
refactor(i18next-initializer): add dependency 'i18next-resources-to-backend' and introduce 'Applicable' multi-providers for choosing i18next backend plugin

Platonn:
To implement in a future release the feature of [loading of i18n JSON files from filesystem (CXSPA-2060)](https://jira.tools.sap/browse/CXSPA-2060), we'll need to do a few breaking changes.

However, according to _semver_, we shouldn' to introduce breaking changes in a minor release. So now in major release 6.0, we prepare for the future and introduce those breaking changes.

BREAKING CHANGES:
- add npm peer dependency `i18next-resources-to-backend` to `@spartacus/core`. 
- add a mechanism for `Applicable` multi-provided strategies for choosing i18next backend plugin and initializing it. However behavior doesn't change - we still use `i18next-http-backend` behind the scenes.

Note: eventually in the future we plan to provide a new `Applicable` strategy for initializing i18next backend, using the package `i18next-resources-to-backend`. Then customers will be able to configure which translations loading strategy they want: via raw http OR via (webpack) dynamic imports.

fixes https://jira.tools.sap/browse/CXSPA-2866


pawelfras:
There seem to be issues with building core package due to circular dependency:
image

pawelfras:
One test is failing due to a mismatch with the snapshot (probably has to be re-generated because of introducing the new library):
image

pawelfras (approved):

PR #17076, opened 2023-03-18
fix for b2b-saved-cart and cart import-export

kimhw0630:
test e2e for b2b-saved-cart


projects/storefrontapp-e2e-cypress/cypress/helpers/cart-import-export.ts
FollowTheFlo, line 1: i think .contains('view cart') is also essential as the label value needs to be checked as well. let's keep minimum changes, adding 'scrollIntoView()' should be enough.

FollowTheFlo (changes requested):

FollowTheFlo (approved):
looks good.

PR #17082, opened 2023-03-21
chore: Generate default Spartacus 6 migrations

bgambocjaviniar:


projects/schematics/src/shared/constants.ts
```diff
@@ -1135,6 +1135,8 @@ export const CSS_V4_DOCS_URL =
   'https://sap.github.io/spartacus-docs/css-changes-in-version-4/';
 export const CSS_V5_DOCS_URL =
   'https://sap.github.io/spartacus-docs/css-changes-in-version-5/';
+export const CSS_V6_DOCS_URL =
+  'https://sap.github.io/spartacus-docs/css-changes-in-version-6/';
```
plabadie: Are there still really GitHub docs for 6.0?

bgambocjaviniar: nope, I checked for 5 too, but it was there, so I did the same :P 

can remove though since 100% won't be there

plabadie (approved):

PR #17093, opened 2023-03-22
fix: use theme color for selection outline

sam-garland:
A red selection outline looks wrong in Santorini.
Use a theme colour.

Increase the outline width to compensate for colours which do not stand out as much.

Closes: https://jira.tools.sap/browse/CXSPA-2907

**======  Before  ======**

image

**======  After  ======**

image


sam-garland:
Targeted wrong branch.

PR #17123, opened 2023-03-24
fix: Append to previous review page hotfix + added important logger and expectations for getting into review page before placing an order

bgambocjaviniar:
closes [feature/CXSPA-2808](https://jira.tools.sap/browse/CXSPA-2808)


FollowTheFlo (approved):
looks good to me. As discussed, feel free to remove timestamp logs.

PR #17127, opened 2023-03-24
fix: bopis schematics

SVendittelli:
Correct config for pickup in store schematics.


scripts/install/functions.sh
cawmichae, line 1: we don't want line 135

cawmichae (changes requested):

cawmichae (dismissed):

bgambocjaviniar (changes requested):
tested as a customer would pull libraries (but I published it through verdaccio myself) 

Manually did the following
1. created a vanilla angular project
2. ng add @spartacus/schematics (I ticked the option to have pick up in store), but it still breaks due to our the schematics logic. **Not really a use case with published libs on the 'real' artifactory, but a problem when simulating using the libs taken directly from source.**

Screenshot 2023-03-24 at 3 15 09 PM

Screenshot 2023-03-24 at 3 10 25 PM


How to fix?
- align the version in the `feature-libs/pickup-in-store/package.json`  - set it to 5.0.0 as default as done in other feature-libs


In relation to the code review provided by your peer https://github.com/SAP/spartacus/pull/17127#discussion_r1147795446

- let's bring it back as I discussed with the team. I was for including it as all the libs are already included except pickup-in-store. However, for the near future, we'll decouple the libs in the installation script to install 'recipes'.

TLDR:
- bring back pick up in store in the functions.sh
- from source published libs to local registry, and manually creating the shellapp and bootstrap spartacus by default with pickup-in-store does not work 🔴 
- from source published libs to local registry, and manually creating the shellapp, installing default schematics without pickup-in-store, and then added pickup-in-store works 🟢 (see below scree
Screenshot 2023-03-24 at 3 18 06 PM
nshot)

projects/schematics/src/shared/lib-configs/pickup-in-store-schematics-config.ts
```diff
@@ -17,7 +17,7 @@ import {
 import { SchematicConfig } from '../utils/lib-utils';
```
bgambocjaviniar, line 17: for line 58 for `dependencyFeatures` 

I saw this in the history before https://github.com/SAP/spartacus/commit/20e072d40514f8b33f6f26eb0255f1b93a924a3e#diff-c4fd083d326d3f805913413a393689971fa098a77f6778ffe941bf801c510623L34-L39

Are all the listed features in the array for schematics necessary for this feature to function?

```
    CART_BASE_FEATURE_NAME,
    ORDER_FEATURE_NAME,
    STOREFINDER_FEATURE_NAME,
    USER_PROFILE_FEATURE_NAME,
```
Basically, are cart base, order, storefinder, user profile features necessary for this feature to work properly?

Thinking out loud, don't we need checkout for this feature to work properly (in a sense that you confirm for pick up with all the checkout steps till the confirmation order) --- I'm not too familiar with bopis, so I'm not really sure how the feature works without actually digging.

cawmichae: Hi Brian, BOPIS references resources in all these features.

bgambocjaviniar (approved):
LGTM

PR #17131, opened 2023-03-24
CXSPA-2946: Deprecated code pass

ercultimate:
implements https://jira.tools.sap/browse/CXSPA-2946


tools/build-lib/augmented-types/index.js
```diff
@@ -133,22 +133,22 @@ function propagateAugmentableTypes(libPath, logger) {
           continue;
         }
         const augTypesEnd =
-          publicApiFileSource.indexOf(DELIMITER_END) + DELIMITER_END.length + 1;
```
plabadie, line 136: Is removing the `+ 1`  intentional?

It changes the augTypesEnd value  and then we make changes like `.substr(augTypesEnd);` to `.substring(augTypesEnd);` which should have the same value to be equivalent.

projects/storefrontlib/cms-components/misc/icon/icon-loader.service.ts
```diff
@@ -21,11 +26,17 @@ import {
...
+
+  protected renderer: Renderer2;
+
   constructor(
     protected winRef: WindowRef,
     protected iconConfig: IconConfig,
-    protected sanitizer: DomSanitizer
-  ) {}
+    protected sanitizer: DomSanitizer,
+    rendererFactory: RendererFactory2
```
plabadie: To add a constructor arg we need to make sure this code goes in 6.0, this PR would then be a port of what was introduced in 6.0

tools/build-lib/augmented-types/index.js
ercultimate: It was intentional, however after further discussion we decided to preserve the original logic because it is unclear what the purpose of the `+ 1` is.

projects/storefrontlib/cms-components/misc/icon/icon-loader.service.ts
ercultimate: This will be removed if the code is not approved for 6.0 release.

hackergil (changes requested):
Let’s hold off on the breaking change for 6.0. Can you remove the breaking change so that we can merge this to develop?

projects/storefrontlib/cms-components/misc/icon/icon-loader.service.ts
ercultimate: Ticket was not approved for 6.0 so I have removed the breaking changes to this file.  

I am keeping the unit test changes since they are a non-breaking and reasonable alternative to checking the usage of `createElement()` and will still meet the goal of removing the deprecated usage.

ercultimate:
> Let’s hold off on the breaking change for 6.0. Can you remove the breaking change so that we can merge this to develop?

Breaking changes have been removed.

hackergil (approved):

plabadie (approved):

PR #17135, opened 2023-03-26
chore: fix ASM flaky

kimhw0630:
closes: https://jira.tools.sap/browse/CXSPA-1644

fixed the issue by selecting the 'cx-card' element more specifically on the 'Address Book' screen, as the 'Pick Up Store' feature also uses the same 'cx-card' element


FollowTheFlo (approved):
all good.

PR #17140, opened 2023-03-27
fix: Should not have a cart displayed after a cart clear from multiple re-orders

bgambocjaviniar:
closes https://jira.tools.sap/browse/CXSPA-2335


feature-libs/order/core/facade/reorder-order.service.ts
```diff
@@ -48,4 +52,24 @@ export class ReorderOrderService implements ReorderOrderFacade {
...
+      this.activeCartFacade.getActiveCartId(),
+    ]).pipe(
+      take(1),
+      map(([userId, cartId]) => {
+        if (!userId) {
+          throw new Error('Must be logged in to reorder');
+        }
+
+        if (cartId) {
+          this.multiCartFacade.deleteCart(cartId, userId);
```
WeizhengSap, line 68: Before reorder, the active cart will be removed? Can we borrow the idea of restore saved cart? (The saved cart becomes active cart, while the current active cart will be saved)

WeizhengSap: and if reorder is successful, then save the active cart.

WeizhengSap: maybe confirm with Bill, how to handle the active cart for reorder.

WeizhengSap (changes requested):
check my comments

WeizhengSap (approved):

feature-libs/order/core/facade/reorder-order.service.ts
bgambocjaviniar: As I mentioned in the DM, but i'll paste here for visibility.

1. Bill said to delete as this was the first phase of the feature
2. future improvement and second phase of the feature is to optionally allow users to the save the cart

All in all, we'll delete it for now, and future improvements may be done for this feature.

PR #17160, opened 2023-03-29
fix: CXSPA-2851 Unexpected error notification shows after customer login

pawelfras:
This pull request contains a fix for unexpected error notifications popping out after the user logs into his account. 
Screenshot 2023-03-29 at 13 26 53

Such an error occurred because of logic in `userConsentService.getConsent()`, only when the CDS feature is enabled due to its additional logic checking consent validity, but the error itself is rather connected with the disconnection between AuthService and UserIdService. The error occurred because method `authService.isUserLoggedIn()` returns true as under the hood it is checking whether the access token is stored in memory. Then logic that uses this method calls `userIdService.takeUserId()` to get the `userId` needed for HTTP Call. Assuming that user is logged in (based on the first condition) we should receive the `current` user. Because the API call is made as an anonymous user but with an access token (which is in place), API responds with a 403 Forbidden error that triggers an error notification.
image

Unfortunately `userId` and access token are changed separately and there is no connection between changing their state (due to the fact that we're using a third-party library for authentication). That said, `authService.isUserLoggedIn()` may return true even if userId is still `anonymous`.

To fix the issue with an invalid notification, apart from the `isUserLoggedIn()` method, the condition has been extended to check whether `userId === current`.

---
Steps to test the fix:

1. set `cds` to true in `environment.ts`
2. go to the login page
3. login 
4. observe that the error notification doesn't pop out.

### DIFF
--- a/projects/core/src/user/facade/user-consent.service.spec.ts
+++ b/projects/core/src/user/facade/user-consent.service.spec.ts
@@ -238,8 +238,20 @@ describe('UserConsentService', () => {
         });
       });
       describe('when the user is anonymous', () => {
-        it('should not call getConsents()', () => {
+        it('should not call getConsents() if isUserLoggedIn returns false', () => {
           spyOn(authService, 'isUserLoggedIn').and.returnValue(of(false));
+          spyOn(userIdService, 'getUserId').and.returnValue(
+            of(OCC_USER_ID_ANONYMOUS)
+          );
+          spyOn(service, 'getConsents').and.stub();
+
+          service.getConsent(mockTemplateId).subscribe().unsubscribe();
+
+          expect(service.getConsents).not.toHaveBeenCalled();
+        });
+
+        it('should not call getConsents() if isUserLoggedIn returns true', () => {
+          spyOn(authService, 'isUserLoggedIn').and.returnValue(of(true));
           spyOn(userIdService, 'getUserId').and.returnValue(
             of(OCC_USER_ID_ANONYMOUS)
           );
--- a/projects/core/src/user/facade/user-consent.service.spec.ts
+++ b/projects/core/src/user/facade/user-consent.service.spec.ts
@@ -4,7 +4,10 @@ import { Observable, of } from 'rxjs';
 import { AuthService } from '../../auth/user-auth/facade/auth.service';
 import { UserIdService } from '../../auth/user-auth/facade/user-id.service';
 import { Consent, ConsentTemplate } from '../../model/consent.model';
-import { OCC_USER_ID_CURRENT } from '../../occ/utils/occ-constants';
+import {
+  OCC_USER_ID_ANONYMOUS,
+  OCC_USER_ID_CURRENT,
+} from '../../occ/utils/occ-constants';
 import { PROCESS_FEATURE } from '../../process/store/process-state';
 import * as fromProcessReducers from '../../process/store/reducers';
 import { UserActions } from '../store/actions/index';
@@ -22,10 +25,14 @@ class MockUserIdService implements Partial<UserIdService> {
   takeUserId() {
     return of(OCC_USER_ID_CURRENT);
   }
+  public getUserId(): Observable<string> {
+    return of(OCC_USER_ID_CURRENT);
+  }
 }
 
 describe('UserConsentService', () => {
   let service: UserConsentService;
+  let userIdService: UserIdService;
   let authService: AuthService;
   let store: Store<StateWithUser>;
 
@@ -48,6 +55,7 @@ describe('UserConsentService', () => {
 
     store = TestBed.inject(Store);
     service = TestBed.inject(UserConsentService);
+    userIdService = TestBed.inject(UserIdService);
     authService = TestBed.inject(AuthService);
     spyOn(store, 'dispatch').and.callThrough();
   });
@@ -232,6 +240,9 @@ describe('UserConsentService', () => {
       describe('when the user is anonymous', () => {
         it('should not call getConsents()', () => {
           spyOn(authService, 'isUserLoggedIn').and.returnValue(of(false));
+          spyOn(userIdService, 'getUserId').and.returnValue(
+            of(OCC_USER_ID_ANONYMOUS)
+          );
           spyOn(service, 'getConsents').and.stub();
 
           service.getConsent(mockTemplateId).subscribe().unsubscribe();
--- a/projects/core/src/user/facade/user-consent.service.ts
+++ b/projects/core/src/user/facade/user-consent.service.ts
@@ -7,10 +7,18 @@
 import { Injectable } from '@angular/core';
 import { select, Store } from '@ngrx/store';
 import { iif, Observable } from 'rxjs';
-import { filter, map, switchMap, tap, withLatestFrom } from 'rxjs/operators';
+import {
+  distinctUntilChanged,
+  filter,
+  map,
+  switchMap,
+  tap,
+  withLatestFrom,
+} from 'rxjs/operators';
 import { AuthService } from '../../auth/user-auth/facade/auth.service';
 import { UserIdService } from '../../auth/user-auth/facade/user-id.service';
 import { Consent, ConsentTemplate } from '../../model/consent.model';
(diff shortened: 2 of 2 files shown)


projects/core/src/user/facade/user-consent.service.ts
```diff
@@ -119,7 +127,12 @@ export class UserConsentService {
    */
   getConsent(templateId: string): Observable<Consent | undefined> {
     return this.authService.isUserLoggedIn().pipe(
-      filter(Boolean),
+      distinctUntilChanged(),
```
Platonn, line 142: nitpick: Do we really need this `distinctUntilChanged`? I can see we already have it inside the implementation of `isUserLoggedIn`.

projects/core/src/user/facade/user-consent.service.ts
```diff
@@ -119,7 +127,12 @@ export class UserConsentService {
...
+      // to avoid redundant calls when userId is still ANONYMOUS
+      withLatestFrom(this.userIdService.getUserId()),
```
Platonn: Instead of `x$.pipe(withLatestFrom(y$))` in this case I suggest to use `combineLatest([x$,y$])` (where `x$` and `y$` are respectively `isUserLoggedIn()` and `getUserId()`).

I suggest it to be bullet-proof, regardless which of those 2 observables emit first. Depending on the customized version of the AuthService, we _never know_ which of those values (token and userId) will be updated as first. In the following examples the order is different:
1. In the original AuthService, the token is always updated as first (so do isUserLoggedIn!) and only later the `userId`. It's the case for both for _password flow_ and _implicit/code flow_. See snippets below (note: `oAuthLibWrapperService` updates the access_token behind the scenes)
   1. password flow:
   https://github.com/SAP/spartacus/blob/37cc60aa0ecd793a531df555162a0977ea3d3539/projects/core/src/auth/user-auth/facade/auth.service.ts#L100-L106
    2. implicit/code flow:
    https://github.com/SAP/spartacus/blob/37cc60aa0ecd793a531df555162a0977ea3d3539/projects/core/src/auth/user-auth/facade/auth.service.ts#L55-L64

2. CsAgentAuthService (which is used only when [ASM feature](https://help.sap.com/docs/SAP_COMMERCE_COMPOSABLE_STOREFRONT/eaef8c61b6d9477daf75bff9ac1b7eb4/8b02b85517af4804a6cc3743f54ead07.html) is enabled) first sets `userId` and only later the `token`. See snippet below:
https://github.com/SAP/spartacus/blob/37cc60aa0ecd793a531df555162a0977ea3d3539/feature-libs/asm/root/services/csagent-auth.service.ts#L72-L73

3. CdcAuthService (used only when [CDC feature](https://help.sap.com/docs/SAP_COMMERCE_COMPOSABLE_STOREFRONT/962112809f9a48f9b36aa05b208b3731/8cdca260bebf4425be0d50a42bbae643.html) is enabled) first sets `token` and then `userId`. See snippet below:
https://github.com/SAP/spartacus/blob/37cc60aa0ecd793a531df555162a0977ea3d3539/integration-libs/cdc/core/auth/facade/cdc-auth.service.ts#L109-L112

Side note: there might be also inconsistencies of the order of operations also for the logout process (e.g. first the `token` is unset vs. first the `userId` is unset), however I didn't check the code examples and... I don't think it's relevant to check it now).

projects/core/src/user/facade/user-consent.service.ts
```diff
@@ -119,7 +127,12 @@ export class UserConsentService {
...
+      filter(
+        ([loggedIn, userId]) => loggedIn && userId === OCC_USER_ID_CURRENT
```
Platonn, line 144: I think we're solving the problem near its symptom, but not in the roots.
Potentially, there might happen similar race-condition bugs in the future in other places. Such bugs might happen whenever some logic waiting for `isUserLoggedIn` to flip to `true` and immediately trigger a http call, without waiting for `getUserId()` to be set to `current`.
We can always patch such bugs in those places or we can try to fix it in the roots.

What do you think about fixing the implementation of the core `AuthService.isUserLoggedIn()` instead - to wait for the `getUserId()` to be set there? I mean:

```ts
isUserLoggedIn(): Observable<boolean> {
  // To ensure data consistency and avoid race-conditions, we only consider the user
  // as "logged in" when both the `access_token` observable emits a value
  // with a token and the `userId` observable emits a value with a non-anonymous user id.
  //
  // This is due to the fact that the observables with `access_token` and `userId`
  // emit values at slightly different timings during the process of login and logout.
  return combineLatest([
    this.authStorageService.getToken(),
  ]).pipe(
    map(([userToken, userId]) =>
      Boolean(userToken?.access_token && userId !== OCC_USER_ID_ANONYMOUS)
    ),
    distinctUntilChanged()
  );
}
```

Note: it's actually working the similar way in the `AsmAuthService`. See snippet below:
https://github.com/SAP/spartacus/blob/37cc60aa0ecd793a531df555162a0977ea3d3539/feature-libs/asm/root/services/asm-auth.service.ts#L133-L146

In the end of the day, I'd consult our solution with the ASM expert, Patrick.

projects/core/src/user/facade/user-consent.service.ts
pawelfras: Ahh, indeed - I didn't take into consideration the scenario when `userId` changed, but the value from the `isUserLogedIn` method is the same... thanks :)

projects/core/src/user/facade/user-consent.service.ts
pawelfras: Sure, I agree, but I was wondering whether we want to add now a change to such a crucial part of an application (actually parts because the ASM feature can be activated independently and then AsmAuthService is in use, so `isUserLoggedIn` in this service has to be extended as well)
Proposal for `AsmAuthService`:
```ts
  isUserLoggedIn(): Observable<boolean> {
    return combineLatest([
      this.authStorageService.getToken(),
      this.authStorageService.getTokenTarget(),
      this.userIdService.isEmulated(),
    ]).pipe(
      map(([token, tokenTarget, isEmulated, userId]) => {
        const isCurrentUser =
          tokenTarget === TokenTarget.User && userId === OCC_USER_ID_CURRENT;
        const isEmulatedUser =
          tokenTarget === TokenTarget.CSAgent && isEmulated;
        return (
          Boolean(token?.access_token) && (isCurrentUser || isEmulatedUser)
        );
      }),
      distinctUntilChanged()
    );
  }
```
Proposal for `AuthService`:
```ts
  isUserLoggedIn(): Observable<boolean> {
    return combineLatest([
      this.authStorageService.getToken(),
    ]).pipe(
      map(([userToken, userId]) => 
        Boolean(userToken?.access_token) && userId === OCC_USER_ID_CURRENT;
      ),
      distinctUntilChanged()
    );
  }
```

projects/core/src/user/facade/user-consent.service.ts
pawelfras: This code has been tested also against `AsmAuthService` where `distinctUntilChanged` is missing, that's why I decided to leave it here. But if we decide to modify `AsmAuthService` and `AuthService` then of course, I prefer to have it there 👌

projects/core/src/user/facade/user-consent.service.ts
Platonn: As discussed offline: Let's now fix the problem only in the symptom for 6.0. And in 6.x let's fix it in the roots.

projects/core/src/user/facade/user-consent.service.spec.ts
```diff
@@ -232,6 +240,9 @@ describe('UserConsentService', () => {
       describe('when the user is anonymous', () => {
         it('should not call getConsents()', () => {
           spyOn(authService, 'isUserLoggedIn').and.returnValue(of(false));
+          spyOn(userIdService, 'getUserId').and.returnValue(
+            of(OCC_USER_ID_ANONYMOUS)
+          );
```
Platonn, line 245: It would be good to also add an unit test against the bug, to avoid a possible regression in the future.

Platonn (approved):

PR #17176, opened 2023-03-29
fix: Change order consignment outlet position from REPLACE to AFTER

surya-prakash-singh1:
ix: Change order consignment outlet position from REPLACE to AFTER


JedidiahAwuku (dismissed):

package-lock.json
```diff
@@ -20647,7 +20647,7 @@
     },
     "projects/eslint": {
       "name": "@spartacus/eslint-plugin-eslint",
-      "version": "0.0.0-dev",
+      "version": "6.0.0-1",
```
WeizhengSap: why we need change this?

feature-libs/order/styles/components/order-details/_order-details-items.scss
```diff
@@ -35,12 +35,6 @@
     padding: 0 30px;
   }
 
-  .cx-review-header {
```
WeizhengSap, line 38: I think we need this. please double check

JedidiahAwuku: Hi Wei, it seems styling for the `cx-review-header` has been already made in the `_review.scss` file or is this deliberate? The UI looks the same when added and removed.

package-lock.json
JedidiahAwuku: Removed. thanks

WeizhengSap (approved):

PR #17203, opened 2023-04-03
Upgrade bootstrap library version to 4.6.2

giancorderoortiz:
Compliance with 12 month SDOL rule.
Version 4.6.2. is 8 months old.
Version 4.6.1. is 12+ months old.


bgambocjaviniar (changes requested):
Make sure to npm install in root to update the lock too. 

just to re-iterate, bootstrap in lock is already on 4.6.2, but the project's deps in the lock is still on ^4.6.1, which will update once you do the install properly

bgambocjaviniar (approved):

PR #17211, opened 2023-04-04
fix: CXSPA-2919 Issue with false consent state

pawelfras:
closes [CXSPA-2919](https://jira.tools.sap/browse/CXSPA-2919)

image

The following PR contains a solution to the incorrect user consent state in the "Consent management" view. The problem occurred when CDS was active due to excessive HTTP calls per user consent. The excess and uncontrolled amount of HTTP calls were caused by the `userConsentService.getConsent()` method calling the `getConsents(true)` method inside a `switchMap` to retrieve consent if it is not in the store. This method listened to every change in user consent in the state, causing the HTTP requests to retrieve consent to occur an uncontrolled number of times and with incorrect parameters (resulting in anonymous consent being saved in the user consent section) as long as there was at least one subscriber (and there are many due to the profile tag service provided by the feature-lib CDS).
To avoid such a scenario, the `getConsents()` method has been extended with operator `take` in a place where logic was responsible for calling a method that will load consents if needed.

---
Steps to test:

1. enable CDS
2. login
3. go to consent management and change some consent
4. logout and log in again as the same user
5. go to the consent management view and check whether the view contains changes applied in step 3.


projects/core/src/user/facade/user-consent.service.ts
```diff
@@ -143,7 +144,7 @@ export class UserConsentService {
       filter(
         ([loggedIn, userId]) => loggedIn && userId === OCC_USER_ID_CURRENT
       ),
-      switchMap(() => this.getConsents(true)),
+      switchMap(() => this.getConsents(true).pipe(take(1))),
```
WeizhengSap: looks good, but can we make it as:
```
switchMap(() => this.getConsents(true)),
take(1),
...
```

pawelfras: Sure! I like the idea to avoid nested `pipe` 👍

WeizhengSap (approved):

PR #17240, opened 2023-04-06
[backport 5.2] fix: CXSPA-2919 Issue with false consent state

pawelfras:


projects/storefrontapp/src/environments/environment.ts
```diff
@@ -25,6 +25,7 @@ export const environment: Environment = {
   occApiPrefix: '/occ/v2/',
   cds: buildProcess.env.CX_CDS ?? false,
   b2b: buildProcess.env.CX_B2B ?? false,
+  // b2b: true,
```
WeizhengSap: is it intended?

WeizhengSap (dismissed):

projects/storefrontapp/src/environments/environment.ts
pawelfras: I left it by mistake. Thanks!

WeizhengSap (approved):

PR #17246, opened 2023-04-11
Migrate workspace to Nx minimally (where applicable)

bgambocjaviniar:
Nx and ng cli tooling can co-exist, but most are changed to use nx under the hood.

Things where we need to add a packages using `ng add` had stayed the same, but may be changed in the future.
https://nx.dev/more-concepts/nx-and-angular#'ng-add'

P.S.
ran 'all' ci using the manual button, and it's green https://github.com/SAP/spartacus/actions/runs/4674657804

closes https://jira.tools.sap/browse/CXSPA-3045

`P.S make sure to install Nx globally to have the cli - npm install -g nx `


projects/storefrontlib/cms-components/myaccount/my-interests/my-interests.component.spec.ts
```diff
@@ -319,27 +318,6 @@ describe('MyInterestsComponent', () => {
     ).toEqual(2);
   });
 
-  it("should contain span element with class name 'cx-visually-hidden' that hides span element content on the UI", () => {
```
bgambocjaviniar, line 322: this is not even working + I believe it is redundant to even make it work. I don't see a point checking the DOM to see that the element has the class, which is statically there in the first place given the conditions in these assertions

bgambocjaviniar: same can be said for https://github.com/SAP/spartacus/pull/17246/files#diff-e4fbb09a4741caf2e068c444de92252a05ef757860c309656032b737085bbbf0R220, but for the reviewer, let me know if there's a problem!

feature-libs/cart/saved-cart/components/details/saved-cart-details-overview/saved-cart-details-overview.component.spec.ts
```diff
@@ -216,16 +215,10 @@ describe('SavedCartDetailsOverviewComponent', () => {
...
-        'button',
-        'cx-edit-cart',
-        undefined,
-        'aria-label',
-        'savedCartDetails.editSavedCart'
-      );
+        el.nativeElement.querySelector('.cx-edit-cart');
+
+      expect(editButton.attributes?.hasOwnProperty('aria-label')).toBe(true);
+      expect(editButton.ariaLabel).toEqual('savedCartDetails.editSavedCart');
```
plabadie, line 221: What is the impact of nx on this spec?

nx.json
```diff
@@ -0,0 +1,18 @@
+{
+  "npmScope": "spartacus",
+  "affected": {
+    "defaultBase": "develop"
```
plabadie: Let's remove the default base config to keep a low overhead with the upcoming branching strategy where the default will change over time.

plabadie (approved):

PR #17253, opened 2023-04-11
fix: Checkout button remains in Processing state when there is cart validation error

WeizhengSap:
ticket: cxspa-2987


feature-libs/cart/base/components/cart-proceed-to-checkout/cart-proceed-to-checkout.component.ts
plabadie, line 1: develop is v6.1

If this is intended for 7.0 it need to be merged elsewhere.
If this is intended for 6.1, perhaps change the deprecation comments.

I see there is also a backport,  same comment applies. Perhaps the comment should state the constructor is  deprecated since 5.2.

plabadie (approved):

PR #17259, opened 2023-04-12
feat: Decouple angular.json into multiple project.json

bgambocjaviniar:
closes https://jira.tools.sap/browse/CXSPA-3046

Splitting angular.json means that ng can't be used anymore since the file is now deleted.
Nx uses the project.json and compacts it into angular.json in memory to make things work as it should

ci run is 🟢 : https://github.com/SAP/spartacus/actions/runs/4722881059


tsconfig.json
```diff
@@ -19,56 +19,32 @@
...
-      "@spartacus/asm": [
-        "feature-libs/asm/public_api"
-      ],
-      "@spartacus/asm/occ": [
-        "feature-libs/asm/occ/public_api"
-      ],
-      "@spartacus/asm/root": [
-        "feature-libs/asm/root/public_api"
-      ],
+      "@spartacus/asm": ["feature-libs/asm/public_api"],
```
bgambocjaviniar, line 22: nx formatted this file to alphabetical order for this tsconfig

scripts/install/functions.sh
```diff
@@ -65,7 +65,6 @@ function prepare_install {
     ng config -g cli.packageManager npm
 
     mkdir -p ${INSTALLATION_DIR}
-    ng analytics off
```
bgambocjaviniar, line 68: to the reviewer:

line 343 of this file has ng build. Reason for that it is the 'customers' application and not our development one. Basically, we're converting source development to use Nx, but customers is business as usual.

plabadie: What is the purpose of removing ng analytics off?

plabadie (approved):

PR #17268, opened 2023-04-12
CXSPA-3072: Rxjs 7 compatible updates

ercultimate:


projects/core/src/auth/user-auth/services/auth-http-header.service.ts
```diff
@@ -273,6 +273,7 @@ export class AuthHttpHeaderService implements OnDestroy {
           // we want to refresh the access token only when it is old.
           // this is a guard for the case when there are multiple parallel http calls
           if (
+            token &&
```
plabadie: What is the purpose of adding `token &&` ?

projects/storefrontlib/cms-components/myaccount/my-coupons/my-coupons.component.ts
```diff
@@ -93,21 +93,20 @@ export class MyCouponsComponent implements OnInit, OnDestroy {
...
+          this.subscriptionFail(error);
+        })
+    );
+    this.subscriptions.add(
+      this.couponService
+        .getUnsubscribeCustomerCouponResultError()
+        .subscribe((error) => {
+          this.subscriptionFail(error);
+        })
+    );
```
plabadie, line 109: Should the observer object be used here?

feature-libs/cart/base/core/facade/active-cart.service.ts
```diff
@@ -378,54 +378,63 @@ export class ActiveCartService implements ActiveCartFacade, OnDestroy {
...
+        filter((cartState) => Boolean(cartState?.success || cartState?.error)),
+        // wait for active cart id to point to code/guid to avoid some work on temp cart entity
+        withLatestFrom(this.activeCartId$),
+        filter(
+          ([cartState, cartId]) => !this.isCartCreating(cartState, cartId)
+        ),
+        map(([cartState]) => cartState.value),
+        filter((cart): cart is Cart => !isEmpty(cart)),
+        take(1)
+      );
```
plabadie, line 436: It's a bit hard to see the diff.  Can you help me understand the change? It looks like there are 2 new `pipe().` some new `?.` and `cart is Cart`.  ( all of which I presume must resolve an rxjs 7 issue)

projects/core/src/auth/user-auth/services/auth-http-header.service.ts
ercultimate: discussion results: This results in a logical change, so we will defer this to investigation for 7.0 release.  For now, we will do a type casting.

feature-libs/cart/base/core/facade/active-cart.service.ts
ercultimate: Correct, the pipe operator only supports up to 10 arguments before the typing gets too expensive to infer and falls back to `Observable<any>`. To keep the type information available, I broke the pipe arguments down into multiple chained calls with less than 10 arguments.  This is functionally equivalent to all pipe operators being in a single call.

The optional chaining operators are unnecessary and will be removed.

The observable type did not match the return type due to still including `undefined` in the type union.  I adjusted the `filter` function's return type to be a type guard to allow the `filter` to narrow the stream emit type.

projects/storefrontlib/cms-components/myaccount/my-coupons/my-coupons.component.ts
ercultimate: No, only when using multiple callback functions is the listener object required.  When using a singular "next" callback function, we are not required to use a listener object.

plabadie (approved):

PR #17283, opened 2023-04-17
Handle PSP redirection and place order

FollowTheFlo:


integration-libs/opf/components/opf-checkout-payment-wrapper/opf-checkout-payment-wrapper.component.ts
```diff
@@ -56,10 +58,8 @@ export class OpfCheckoutPaymentWrapperComponent implements OnInit {
           configurationId: String(this.selectedPaymentId),
           cartId: this.activeCartId,
           // TODO: Move below as a part of a whole OPF configuration?
-          resultURL:
-            'http://localhost:4200/electronics-spa/en/USD/checkout/review-order',
-          cancelURL:
-            'http://localhost:4200/electronics-spa/en/USD/checkout/review-order',
+          resultURL: `http://localhost:4200/${this.config.opf?.successUrl}`,
```
FollowTheFlo: hard coded domain for demo purpose, i haven't investigate the best way to get client URL domain: routingService? window obj? config?

integration-libs/opf/components/opf-verify-payment/opf-verify-payment.component.html
FollowTheFlo, line 1: not CMS cpnt, only display blank page and Spinner, can be seen for few sec in slow connection. In any case user gets redirected.

projects/storefrontapp/src/app/spartacus/features/opf/opf-feature.module.ts
```diff
@@ -33,6 +34,8 @@ if (environment.b2b) {
         baseUrl:
           'https://opf-dev.api.commerce.stage.context.cloud.sap/commerce-cloud-adapter/storefront',
         commerceCloudPublicKey: 'ab4RhYGZ+w5B0SALMPOPlepWk/kmDQjTy2FU5hrQoFg=',
+        successUrl: 'redirect/success',
```
Matejk00: I would prefer to pass the absolute URL, not only a route. Maybe we can create a service and method which will return an absolute url based on routing service. For example: `getAbsoluteUrl('redirect/success')`. As an argument we can pass a route path.

Matejk00 (changes requested):

integration-libs/opf/components/opf-checkout-payment-wrapper/opf-checkout-payment-wrapper.component.ts
Matejk00: Right, so based on my comment here: https://github.com/SAP/spartacus/pull/17283#discussion_r1169763600
Might be better to pass here only a config param, instead of string concatenation.

Example:
```
    resultURL: this.config.opf?.successUrl
```

integration-libs/opf/components/opf-verify-payment/opf-verify-payment.component.ts
```diff
@@ -0,0 +1,91 @@
...
+          const paymentSessionId = this.getPaymentSessionId(list);
+          if (!paymentSessionId) return throwError('No paymentSessionId found');
+
+          return this.opfCheckoutService.verifyPayment(paymentSessionId, {
+            responseMap: [...list],
+          });
+        }),
+
+        switchMap((response: OpfVerifyPaymentResponse) => {
+          return response?.result === 'AUTHORIZED'
```
Matejk00: Can we create an `enum` in a model for 'AUTHORIZED', 'UNATHORIZED' and all other possible options? Then we can have IDE code suggestions available and reuse such enum in other places if needed.

Example:
```
export enum OpfPaymentVerficationResult {
  AUTHORIZED = 'AUTHORIZED',
  UNAUTHORIZED = 'UNAUTHORIZED',
  ...
}

```

integration-libs/opf/occ/adapters/occ-opf.adapter.ts
```diff
@@ -94,4 +97,39 @@ export class OccOpfAdapter implements OpfAdapter {
...
+
+  getVerifyPayment(
+    paymentSessionId: string,
+    payload: OpfVerifyPaymentPayload
+  ): Observable<OpfVerifyPaymentResponse> {
+    const headers = new HttpHeaders({
+      accept: 'application/json',
+      'Content-Type': 'application/json',
+      'Content-Language': 'en-us',
+      'sap-commerce-cloud-public-key':
```
Matejk00: For this key we can use a constant `OPF_CC_PUBLIC_KEY` from `opf/root/config/constants.ts`.

integration-libs/opf/occ/model/occ-opf-endpoints.model.ts
```diff
@@ -16,5 +16,6 @@ declare module '@spartacus/core' {
      * Endpoint to initiate payment session or call the PSP to initiate.
      */
     initiatePayment?: string | OccEndpoint;
+    getVerifyPayment?: string | OccEndpoint;
```
Matejk00: Sorry, maybe it's a nitpick but `getVerifyPayment` sounds not informative to me and get in the name suggests that we send a get request. What about change to `verifyPayment`?

Matejk00: Can you also add a comment above?

```
  /**
   * Endpoint to verify a response from PSP for Full Page Redirect 
   * and iFrame integration patterns.
   */
```

integration-libs/opf/components/opf-verify-payment/opf-verify-payment.component.ts
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  ngOnInit(): void {
+    this.subscription = this.route.queryParams
+      .pipe(
+        take(1),
+        switchMap((params) => {
+          if (!params) return throwError('No params');
+
+          const list: KeyValuePair[] = Object.entries(params).map((pair) => {
```
Matejk00: Can we have this part as a separate method, and ideally in a service?

integration-libs/opf/occ/adapters/occ-opf.adapter.ts
```diff
@@ -94,4 +97,39 @@ export class OccOpfAdapter implements OpfAdapter {
...
+        catchError((error) => throwError(error)),
+        backOff({
+          shouldRetry: isJaloError,
+        }),
+        this.converter.pipeable(OPF_VERIFY_PAYMENT_NORMALIZER)
+      );
+  }
+
+  protected getVerifyPaymentEndpoint(paymentSessionId: string): string {
+    return this.opfEndpointsService.buildUrl('getVerifyPayment', {
```
Matejk00, line 131: If we call ccv2 endpoints we can still use `occEndpointsService` from `@spartacus/core` to build url. Also if this one relates to the ccv2 endpoint would be good to extract this method into separate file. For example: `occ-order.adapter.ts`.

As an example please see: `opf/occ/adapters/occ-otp.adapter.ts`.
Similarly I created for OTP endpoint separate files like: `default-occ-otp-config.ts`, `occ-otp-endpoints.model.ts`. To not mixup endpoints destination and separate CC with CCv2 ones.

integration-libs/opf/components/opf-checkout-payment-wrapper/opf-checkout-payment-wrapper.component.ts
FollowTheFlo: I used a OpfUrlHandlerService shared between components that is picking the current domain.
having the absolute path already in config would involve fetching the current domain before any components load, I am not sure how to do it.

projects/storefrontapp/src/app/spartacus/features/opf/opf-feature.module.ts
FollowTheFlo: this config has been deleted. Now, only Routing config is used.

integration-libs/opf/components/opf-verify-payment/opf-verify-payment.component.ts
FollowTheFlo: done

integration-libs/opf/occ/model/occ-opf-endpoints.model.ts
FollowTheFlo: Thanks, renamed all 'actions context' with verifyPayment and files/data/cpnt with a 'state context' PaymentVerification, def more appropriate.

integration-libs/opf/components/opf-verify-payment/opf-verify-payment.component.ts
FollowTheFlo: created 'OpfUrlHandlerService' which has generic methods
convertParamsToKeyValuePairs
findFromKeyValuePairs
isIncludedInPath
getDomainUrl
This service is in Components folder, still need some work to move it to Services folder as it throws an error in runtime

projects/storefrontapp/src/app/spartacus/features/opf/opf-feature.module.ts
```diff
@@ -25,14 +26,28 @@ if (environment.b2b) {
   extensionProviders.push(provideConfig(defaultOPFCheckoutConfig));
 }
 
+const resultUrl = 'redirect/success';
+const cancelUrl = 'redirect/failure';
```
FollowTheFlo: It is essential that resultUrl/CancelUrl values be from the same source in Router and OpfConfig. Any mismatch would break the PaymentRedirection logic. Thus I also added the routing config (line 35).
The defaultOPFRoutingConfig config is also using resultUrl/CancelUrl values from same source (defaultOpfConfig)
Let me know if you see an alternative solution.

integration-libs/opf/root/config/opf-order-v2.config.ts
FollowTheFlo, line 1: will be deleted once PlaceOrderV2 will be implemented

integration-libs/opf/root/facade/opf-checkout.facade.ts
```diff
@@ -7,17 +7,26 @@
...
 import { ActiveConfiguration } from '../model/opf-checkout.model';
 
 @Injectable({
   providedIn: 'root',
   useFactory: () =>
     facadeFactory({
       facade: OpfCheckoutFacade,
-      feature: OPF_CORE_FEATURE,
-      methods: ['getActiveConfigurationsState', 'initiatePayment'],
+      feature: OPF_FEATURE,
```
FollowTheFlo, line 24: I changed OPF_CORE_FEATURE (initially set by patrik Smull) to OPF_FEATURE and registered OPF_FEATURE in opf-feature.module.
If I register OPF_CORE_FEATURE in opf-feature.module I get 'adapter not found' in runtime.

integration-libs/opf/occ/adapters/occ-opf.adapter.ts
FollowTheFlo: done

integration-libs/opf/components/opf-checkout-payment-wrapper/opf-checkout-payment-wrapper.component.ts
FollowTheFlo: as discussed, it is now handle by service using router.
```
 resultURL: this.opfUrlHandlerService.getAsoluteUrl(
              'paymentVerificationResult'
            ),
            cancelURL: this.opfUrlHandlerService.getAsoluteUrl(
              'paymentVerificationCancel'
```
            ),

integration-libs/opf/components/opf-url-handler.service.ts
```diff
@@ -0,0 +1,35 @@
...
+import { KeyValuePair } from '../root/model';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class OpfUrlHandlerService {
+  constructor(protected routingService: RoutingService) {}
+
+  convertParamsToKeyValuePairs(params: Params): KeyValuePair[] {
+    if (!params) return [];
```
Matejk00: Please add curly brackets to this if, otherwise we will not pass a SonarQube check: https://sonar.tools.sap/project/issues?resolved=false&severities=CRITICAL&inNewCodePeriod=true&pullRequest=17283&id=composable-storefront

integration-libs/opf/components/opf-verify-payment/opf-payment-verification.component.ts
```diff
@@ -0,0 +1,90 @@
...
+  ngOnInit(): void {
+    this.subscription = of(this.route.routeConfig?.data?.cxRoute as string)
+      .pipe(
+        switchMap((cxRoute) => {
+          return cxRoute === 'paymentVerificationResult'
+            ? this.route.queryParams
+            : throwError('CANCEL URL RETURNED BY PSP');
+        }),
+        switchMap((params) => {
+          if (!params) return throwError('No params');
```
Matejk00: Missing curly brackets.

integration-libs/opf/components/opf-verify-payment/opf-payment-verification.component.ts
```diff
@@ -0,0 +1,90 @@
...
+
+          const paramsList: KeyValuePair[] =
+            this.opfUrlHandlerService.convertParamsToKeyValuePairs(params);
+
+          const paymentSessionId =
+            this.opfUrlHandlerService.findFromKeyValuePairs(
+              OpfPaymenVerificationUrlInput.PAYMENT_SESSION_ID,
+              paramsList
+            );
+          if (!paymentSessionId) return throwError('No paymentSessionId found');
```
Matejk00: Missing curly brackets.

integration-libs/opf/components/opf-checkout-payment-wrapper/opf-checkout-payment-wrapper.component.ts
```diff
@@ -50,18 +54,22 @@ export class OpfCheckoutPaymentWrapperComponent implements OnInit {
...
-        },
-      })),
+      map(({ value: otpKey }) => {
+        return {
+          otpKey,
+          config: {
+            configurationId: String(this.selectedPaymentId),
+            cartId: this.activeCartId,
+            // TODO: Move below as a part of a whole OPF configuration?
+            resultURL: this.opfUrlHandlerService.getAsoluteUrl(
```
Matejk00: I think there is a typo 'getAsoluteUrl' -> 'getAbsoluteUrl'.

integration-libs/opf/root/config/default-opf-routing-config.ts
```diff
@@ -3,3 +3,18 @@
  *
  * SPDX-License-Identifier: Apache-2.0
  */
+
+import { RoutingConfig } from '@spartacus/core';
+
+export const defaultOPFRoutingConfig: RoutingConfig = {
```
Matejk00: Let's follow naming convention rule, so it should be `defaultOpfRoutingConfig`. I am aware that we have 'OPF' capitalized in the names in few places, but hopefully we will fix this later just for consistency.

integration-libs/opf/root/model/opf-payment-verification.model.ts
```diff
@@ -0,0 +1,26 @@
...
+ * SPDX-FileCopyrightText: 2023 SAP Spartacus team <spartacus-team@sap.com>
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+export interface OpfPaymentVerificationPayload {
+  responseMap: KeyValuePair[];
+}
+
+export interface KeyValuePair {
```
Matejk00: Just proposal, for me it also part of the OPF structure (but KeyValuePair pointing to the something more general). What do you think to name it like: `OpfResponseMapElement`?

integration-libs/opf/root/config/opf-order-v2.config.ts
```diff
@@ -0,0 +1,17 @@
+/*
+ * SPDX-FileCopyrightText: 2023 SAP Spartacus team <spartacus-team@sap.com>
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { OccConfig } from '@spartacus/core';
+
+export const occOrderV2Config: OccConfig = {
```
Matejk00: If you want to split this ticket into two parts, shouldn't this part be a port of placing order stuff?

integration-libs/opf/components/opf-url-handler.service.ts
```diff
@@ -0,0 +1,35 @@
+/*
```
Matejk00: My suggestion is to change a name of this service for `OpfVerifyPaymentService` the same for the filename (opf-verify-payment.service.ts) and move this into opf-verify-payment component folder to be along with other component files.

It's because we are not using methods provided in this service anywhere else, and hopefully we will not.

integration-libs/opf/components/opf-verify-payment/opf-payment-verification.component.ts
```diff
@@ -0,0 +1,90 @@
...
+
+          return this.opfCheckoutService.verifyPayment(paymentSessionId, {
+            responseMap: [...paramsList],
+          });
+        }),
+        switchMap((response: OpfPaymentVerificationResponse) => {
+          return response?.result === OpfPaymentVerificationResult.AUTHORIZED
+            ? this.orderFacade.placeOrder(true)
+            : throwError('UNAUTHORIZED payment from OPF Adapter');
```
Matejk00: One thing related to the throwing errors this way. You should be aware that is not optimal way for the case if we want to display specific error message to the customer.
We mosty use this approach in services to point out that some logic is an error.

However in terms of showing a message to the customer in UI the recommended way is using `GlobalMessageService`.

Example:
```
  protected displayError(): void {
      {
        key: 'opf.yourErrorTranslationKey',
      },
      GlobalMessageType.MSG_TYPE_ERROR
    );
  }
```

The main advantage of this solution is the possibility of passing a translation key. By doing this, customers will always receive an error message translated to the language they are currently using.

Matejk00 (changes requested):

integration-libs/opf/root/facade/opf-checkout.facade.ts
Matejk00: For this I think it should `OPF_FEATURE`. I don't see a purpose of `OPF_CORE_FEATURE` atm.

Matejk00 (approved):

PR #17298, opened 2023-04-18
CXSPA-2151: Move asm bind cart e2e tests to separate file

ercultimate:


projects/storefrontapp-e2e-cypress/cypress/e2e/regression/asm/asm.e2e-flaky.cy.ts
```diff
@@ -16,10 +15,6 @@ import { getSampleUser } from '../../../sample-data/checkout-flow';
 import { clearAllStorage } from '../../../support/utils/clear-all-storage';
 
 context('Assisted Service Module', () => {
-  before(() => {
-    clearAllStorage();
```
pawelfras, line 20: `query`: What's the reason behind removing this function call?
Sidenote: please double-check whether `clearAllStorage` is needed in `bind-cart.e2e-flaky.cy.ts`

ercultimate: I believe I saw a ticket for removing all of these calls as they are redundant with the new test isolation behavior in Cypress 12.  I'll verify this is the case.

pawelfras (dismissed):

plabadie (approved):

PR #17305, opened 2023-04-19
fix: use store finder configuration for map

kimhw0630:
Closes: https://jira.tools.sap/browse/CXSPA-3188

remove our own configuration for Google map key and radius.  share with storefinder configuration since our map component is using storefinder service.

### DIFF
--- a/feature-libs/asm/customer-360/components/sections/asm-customer-map/asm-customer-map.component.html
+++ b/feature-libs/asm/customer-360/components/sections/asm-customer-map/asm-customer-map.component.html
@@ -49,7 +49,7 @@
     </div>
     <div class="store-map-container">
       <iframe
-        *ngIf="!!(source.config$ | async)?.googleMapsApiKey"
+        *ngIf="storeFinderConfig?.googleMaps?.apiKey"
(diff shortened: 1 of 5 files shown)


Zeyber (changes requested):
ASM is its own library independent of the StoreFinder. It would be preferable to use a `config` with a map key for ASM (or 360) so that we do not need to import the store finder for ASM 360 (even if it is the same key). This also has the advantage of allowing us to use different map keys for different features (eg. a customer may wish to consolidate billing for ASM map usage separately to other map usages across spartacus).

kimhw0630:
> ASM is its own library independent of the StoreFinder. It would be preferable to use a `config` with a map key for ASM (or 360) so that we do not need to import the store finder for ASM 360 (even if it is the same key). This also has the advantage of allowing us to use different map keys for different features (eg. a customer may wish to consolidate billing for ASM map usage separately to other map usages across spartacus).

The Customer 360 feature uses the StoreFinderService, which means that in order to install Customer360, the StoreFinder feature must also be installed.  This is how the Customer360 map module is designed/architected..

Zeyber (approved):

PR #17306, opened 2023-04-20
docs: Update readme for creating a library

bgambocjaviniar:
closes https://jira.tools.sap/browse/CXSPA-3083


docs/libs/creating-lib.md
```diff
@@ -1,8 +1,8 @@
 # Creating a Spartacus library
 
-An easy way to create a new Spartacus library is to run: `ng g library <lib-name>` where the `lib-name` is the name of the new library.
+An easy way to create a new Spartacus library is to run: `nx g @schematics/angular:library <lib-name> --prefix=cx` where the `lib-name` is the name of the new library.
 
-The library will be generated in the `feature-libs` folder by default. You need to manually move the generated files to an appropriate directory (if necessary) and modify `angular.json` to reflect the new location.
+The library will be generated in the `root of the project` folder by default. You need to manually move the generated folder, which is under the `<lib-name>` you entered to the appropriate directory (if necessary). The entire generated folder should be moved either under `feature-libs` or `integration-libs`.
```
plabadie, line 5: About the logic and not the doc itself: could we implement a default config that is the same as we have at the moment with ng: the libs are generated in `./feature-libs` because that's where most libs need to go.  If for a reason it does not belong there we need to move it elsewhere, but those are exceptions, not the rule.

docs/libs/creating-lib.md
```diff
@@ -105,6 +106,28 @@ module.exports = function (config) {
...
+          "integration-libs/TODO:/**/*.ts",
+          "integration-libs/TODO:/**/*.html"
+        ]
+      }
+    }
+    ```
+
+  - add the tag(s)
+    - type:feature
+    - type:integration `IF AND ONLY IF` it is an integration lib.
```
plabadie, line 129: An example of tags in the json above would be helpful

plabadie (approved):

docs/libs/creating-lib.md
bgambocjaviniar: done

plabadie (approved):

PR #17325, opened 2023-04-24
CXINT-1912 [SPA 6.0] [CDC] Application throws error message after updating email in Profile Details

SherwinVarghese:
- Code fixes for the bug
- Added unit tests
- Added E2Es


anjana-bl (dismissed):
Tested the fix and approved

anjana-bl (approved):
approving again as base branch was changed

PR #17343, opened 2023-04-26
chore : update overview sections to use real data(CXSPA-3136)

kimhw0630:
Closes: https://jira.tools.sap/browse/CXSPA-3136

- Overview header section 
- Active cart section
- Saved cart section
- Interests section
- common components clean up

### DIFF
--- a/feature-libs/asm/customer-360/components/customer-360/avatar-image.pipe.ts
+++ b/feature-libs/asm/customer-360/components/customer-360/avatar-image.pipe.ts
@@ -1,3 +1,9 @@
+/*
+ * SPDX-FileCopyrightText: 2023 SAP Spartacus team <spartacus-team@sap.com>
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
 import { Pipe, PipeTransform } from '@angular/core';
 import { CustomerOverview } from '@spartacus/asm/customer-360/root';
 import { Image } from '@spartacus/core';
--- a/feature-libs/asm/customer-360/components/customer-360/avatar-image.pipe.ts
+++ b/feature-libs/asm/customer-360/components/customer-360/avatar-image.pipe.ts
@@ -8,7 +8,7 @@ import { Pipe, PipeTransform } from '@angular/core';
 import { CustomerOverview } from '@spartacus/asm/customer-360/root';
 import { Image } from '@spartacus/core';
 
-@Pipe({ name: 'avatarImage' })
+@Pipe({ name: 'cxAvatarImage' })
 export class AvatarImagePipe implements PipeTransform {
   transform(overview?: CustomerOverview): Image | undefined {
     return overview?.userAvatar?.url
--- a/feature-libs/asm/customer-360/components/customer-360/avatar-image.pipe.ts
+++ b/feature-libs/asm/customer-360/components/customer-360/avatar-image.pipe.ts
@@ -1,0 +1,16 @@
+import { Pipe, PipeTransform } from '@angular/core';
+import { CustomerOverview } from '@spartacus/asm/customer-360/root';
+import { Image } from '@spartacus/core';
+
+@Pipe({ name: 'avatarImage' })
+export class AvatarImagePipe implements PipeTransform {
+  transform(overview?: CustomerOverview): Image | undefined {
+    return overview?.userAvatar?.url
+      ? {
+          altText: overview.name,
+          url: overview.userAvatar.url,
+          format: overview.userAvatar.format,
+        }
+      : undefined;
+  }
+}
--- a/feature-libs/asm/customer-360/components/customer-360/avatar-label.pipe.ts
+++ b/feature-libs/asm/customer-360/components/customer-360/avatar-label.pipe.ts
@@ -1,0 +1,11 @@
+import { Pipe, PipeTransform } from '@angular/core';
+import { User } from '@spartacus/core';
+
+@Pipe({ name: 'avatarLabel' })
+export class AvatarLabelPipe implements PipeTransform {
+  transform(customer?: User): string {
+    customer = customer ?? {};
+    const { firstName = '', lastName = '' } = customer;
+    return `${firstName.charAt(0)}${lastName.charAt(0)}`;
+  }
+}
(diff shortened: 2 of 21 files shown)


feature-libs/asm/customer-360/components/customer-360/customer-360.component.html
```diff
@@ -18,63 +18,112 @@ <h3 id="asm-customer-360-title" class="modal-title">
...
-          </div>
+          <ng-container
+            *ngIf="customerOverview?.userAvatar?.url; else customerTextAvatar"
+          >
+            <div
+              class="cx-avatar-image"
+              *ngIf="customerOverview?.userAvatar?.url"
+            >
+              <cx-media
+                [container]="getAvatarImage(customerOverview)"
```
rmch91: Usage of methods inside of template is not good from performance point of view. Method will execute all the time when, change detection will run. Better to use pure pipe, which will be triggered only when the source will be changed. Maybe you can create pipe just for this component (if this functionality is used only here) and move logic from method to the pipe?

feature-libs/asm/customer-360/components/customer-360/customer-360.component.html
```diff
@@ -18,63 +18,112 @@ <h3 id="asm-customer-360-title" class="modal-title">
...
+              ></cx-media>
+            </div>
+          </ng-container>
+
+          <ng-template #customerTextAvatar>
+            <div class="cx-avatar" *ngIf="!customerOverview?.userAvatar?.url">
+              {{ getAvatarText() }}
```
rmch91: same as above

feature-libs/asm/customer-360/components/customer-360/customer-360.component.html
```diff
@@ -18,63 +18,112 @@ <h3 id="asm-customer-360-title" class="modal-title">
...
+            </div>
+          </ng-template>
+
           <div class="cx-customer-info ml-3">
             <div class="cx-customer-name">
               {{
                 'customer360.header.subTitle'
-                  | cxTranslate: { name: customer.firstName }
+                  | cxTranslate: { name: customerOverview.name }
```
rmch91: maybe customerOverview?.name ?

feature-libs/asm/customer-360/components/customer-360/customer-360.component.html
```diff
@@ -18,63 +18,112 @@ <h3 id="asm-customer-360-title" class="modal-title">
...
               }}
             </div>
             <div class="cx-customer-email">
-              {{ customer.uid }}
+              {{ customerOverview.email }}
```
rmch91: customerOverview?.email

feature-libs/asm/customer-360/components/customer-360/customer-360.component.html
```diff
@@ -18,63 +18,112 @@ <h3 id="asm-customer-360-title" class="modal-title">
...
             </div>
           </div>
         </div>
         <div class="header-account-details">
           <span
             class="header-account-details-active-cart"
-            *ngIf="activeCart$ | async as activeCart"
+            *ngIf="customerOverview.cartCode"
```
rmch91: customerOverview?.cartCode

feature-libs/asm/customer-360/components/customer-360/customer-360.component.html
```diff
@@ -18,63 +18,112 @@ <h3 id="asm-customer-360-title" class="modal-title">
...
           >
             <cx-icon class="account-icon" [type]="cartIcon"></cx-icon>
             {{
               'customer360.header.activeCartLabel'
-                | cxTranslate: { cartSize: activeCart.totalItems }
+                | cxTranslate: { cartSize: customerOverview.cartSize }
```
rmch91: There are places where you are using `customerOverview?.something` and somewhere you are not using question mark. So if we not sure that `customerOverview` always will exist, maybe we should add the question mark everywhere

feature-libs/asm/customer-360/root/model/customer-360.model.ts
```diff
@@ -130,43 +131,48 @@ export interface Customer360CartEntry {
...
 
 export interface Customer360Avatar {
   url?: string;
   format?: string;
 }
 
-export interface Customer360Overview {
-  type: Customer360Type.OVERVIEW;
+export interface CustomerOverview {
+  address: Address;
```
rmch91: should it be optional also?

rmch91 (changes requested):

feature-libs/asm/customer-360/components/customer-360/customer-360.component.html
kimhw0630: thanks created a pipe.

feature-libs/asm/customer-360/components/customer-360/avatar-image.pipe.ts
```diff
@@ -0,0 +1,22 @@
...
+ * SPDX-FileCopyrightText: 2023 SAP Spartacus team <spartacus-team@sap.com>
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { Pipe, PipeTransform } from '@angular/core';
+import { CustomerOverview } from '@spartacus/asm/customer-360/root';
+import { Image } from '@spartacus/core';
+
+@Pipe({ name: 'avatarImage' })
```
rmch91: Nitpick, but almost all of pipes names in the storefront starts from "cx".  Maybe for consistency we should add "cx" prefix also.

feature-libs/asm/customer-360/components/customer-360/avatar-label.pipe.ts
```diff
@@ -0,0 +1,17 @@
+/*
+ * SPDX-FileCopyrightText: 2023 SAP Spartacus team <spartacus-team@sap.com>
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { Pipe, PipeTransform } from '@angular/core';
+import { User } from '@spartacus/core';
+
+@Pipe({ name: 'avatarLabel' })
```
rmch91: Nitpick, but almost all of pipes names in the storefront starts from "cx".  Maybe for consistency we should add "cx" prefix also.

rmch91 (approved):

PR #17345, opened 2023-04-27
CXSPA-3113 As an assisted service agent, I want to open deep link for Spartacus containing customerID only so that the emulation session can be started automatically for the customer

scarai-sap:
See https://jira.tools.sap/browse/CXSPA-3113


feature-libs/asm/package.json
```diff
@@ -29,6 +29,7 @@
     "@angular/common": "^15.2.4",
     "@angular/core": "^15.2.4",
     "@angular/forms": "^15.2.4",
+    "@angular/router": "^15.2.4",
```
luoyifeng11: If considering adding a new dependency, please discuss the requirements with @Oezak, Levent for approval.

Wiki entry: https://wiki.one.int.sap/wiki/pages/viewpage.action?spaceKey=spar&title=%5BDRAFT%5D+Guideline%3A+Adding+3rd+party+dependencies

feature-libs/asm/components/asm-main-ui/asm-main-ui.component.ts
```diff
@@ -132,6 +162,30 @@ export class AsmMainUiComponent implements OnInit, OnDestroy {
           }
         })
     );
+    this.subscribeForDeeplink();
```
i53577, line 167: is this a breaking change?
https://sap.github.io/spartacus-docs/breaking-changes/
> Changing function behavior where the function returns different results. In other words, when changing function behavior, the function should return the same results for the same provided input as it did before the change.

scarai-sap: just call private routine, why breaking? Ah, will add featureLevel later if that's what you are saying

feature-libs/asm/components/asm-main-ui/asm-main-ui.component.ts
```diff
@@ -132,6 +162,30 @@ export class AsmMainUiComponent implements OnInit, OnDestroy {
...
+  }
+
+  private subscribeForDeeplink(): void {
+    if (this.router) {
```
i53577: According to the discussion with Jason and @luoyifeng11, if I understand correctly, customer has a flag representing the version he/she want to use. That means a customer could be able to upgrade his/her code to new version and keep the flag as old version. It will help he/she to fix some bug that being fixed in new versions and not introduce any new features. In this case, just using `if (this.router)` is not enough. We had better use `featureConfig` instead of this.
@luoyifeng11 Please help confirm if this is the case.  Thx.

luoyifeng11: I see in the PR description it says, 'Feature level will be added later to keep compatibility for old versions.' Does this mean that the featureLevel will be added later? @scarai-sap

scarai-sap: I just added when refactor the code.

feature-libs/asm/package.json
scarai-sap: Fixed with using existing component

feature-libs/asm/components/asm-main-ui/asm-main-ui.component.ts
scarai-sap: Added feature level to support it

ghost (commented):
How about add one scenario so that deep link could open B2B or B2C SPA site?

scarai-sap:
> How about add one scenario so that deep link could open B2B or B2C SPA site?

Thanks for your suggestion!

What's the benefit for that? Unless there has different behavior between them, we don't need separate them at all.

projects/storefrontapp-e2e-cypress/cypress/e2e/regression/asm/asm.e2e-flaky.cy.ts
```diff
@@ -23,6 +23,103 @@ context('Assisted Service Module', () => {
...
+
+      cy.log('--> Agent logging in with deeplink');
+      checkout.visitHomePage('asm=true&customerId=' + customer.email);
+      cy.get('cx-asm-main-ui').should('exist');
+      cy.get('cx-asm-main-ui').should('be.visible');
+
+      cy.log('--> Register user');
+      checkout.registerUser(false, customer);
+
+      asm.agentLogin('asagent', 'pw4all');
```
Zeyber: Lets use constants for the user and password as they recur many times.

feature-libs/asm/components/asm-main-ui/asm-main-ui.component.ts
```diff
@@ -132,6 +166,29 @@ export class AsmMainUiComponent implements OnInit, OnDestroy {
           }
         })
     );
+    this.subscribeForDeeplink();
+  }
+
+  private subscribeForDeeplink(): void {
```
Zeyber: In general, we prefer not to use `private` in our library because this makes the implementation invisible to users. Sometimes our lib users might like to customize a method. Methods we prefer users not to touch we use `protected` for as this still allows users to changes them if they really want to.

feature-libs/asm/components/asm-main-ui/asm-main-ui.component.ts
```diff
@@ -53,6 +57,8 @@ export class AsmMainUiComponent implements OnInit, OnDestroy {
   customer$: Observable<User | undefined>;
   isCollapsed$: Observable<boolean> | undefined;
   iconTypes = ICON_TYPE;
+  customerIdInURL: string;
+  emulated: boolean = false;
```
kimhw0630:
```suggestion
  emulated = false;
```

feature-libs/asm/components/asm-main-ui/asm-main-ui.component.ts
```diff
@@ -132,6 +166,29 @@ export class AsmMainUiComponent implements OnInit, OnDestroy {
           }
         })
     );
+    this.subscribeForDeeplink();
+  }
+
+  private subscribeForDeeplink(): void {
+    if ((this.featureConfig?.isLevel('6.1') ?? false) && this.location) {
```
Zeyber: Do we need the `?? false` condition? `undefined` is already a falsy value.

feature-libs/asm/components/asm-main-ui/asm-main-ui.component.ts
```diff
@@ -157,6 +214,13 @@ export class AsmMainUiComponent implements OnInit, OnDestroy {
     this.csAgentAuthService.authorizeCustomerSupportAgent(userId, password);
   }
 
+  private startSessionWithParameters(): void {
```
Zeyber: Same as above about `private` methods.

feature-libs/asm/components/asm-main-ui/asm-main-ui.component.ts
```diff
@@ -157,6 +214,13 @@ export class AsmMainUiComponent implements OnInit, OnDestroy {
...
+    if (this.customerIdInURL && !this.emulated) {
```
Zeyber: We will be accounting for many potential parameters in the very near future than just `customerId`. We should come up with a more robust solution to predicate this situation. For example:

```
getParametersFromUrl() {
  const ALLOWED_PARAMETERS = ['customerId', *future params*...]
  const queryString = this.location.path().split('?')[1];
  const params = ALLOWED_PARAMETERS.map(param => ({ [param]: new URLSearchParams(queryString).get(param)}))
  return params
}

...

startSessionWithParameters(this.getParametersFromUrl())
```

feature-libs/asm/components/asm-main-ui/asm-main-ui.component.ts
```diff
@@ -53,6 +57,8 @@ export class AsmMainUiComponent implements OnInit, OnDestroy {
   customer$: Observable<User | undefined>;
   isCollapsed$: Observable<boolean> | undefined;
   iconTypes = ICON_TYPE;
+  customerIdInURL: string;
```
Zeyber: Do we need to be setting global variables here? If so, we prefer to use Observables for variables to detect changes during code flow. It would be great if we can move the logic that defines these into a service. eg. `this.service.getIsEmulated(): Observable`.

Zeyber (changes requested):
Could we please add comments along the way to our new methods and some places where its not quite clear why we might have some logic or condition? (`auth-http-header.service.ts` is just one example of this done well). I can see that the ASM implementation over time has become a bit of a mess in its own right. Lets not contribute to that ;)

feature-libs/asm/components/asm-main-ui/asm-main-ui.component.ts
scarai-sap: good catch, will fix it.

feature-libs/asm/components/asm-main-ui/asm-main-ui.component.ts
scarai-sap: As the isEmulated is only used in this component, so maybe we can keep it in here, Anyway, I moved to amsComponentService

feature-libs/asm/components/asm-main-ui/asm-main-ui.component.ts
scarai-sap: fixed

feature-libs/asm/components/asm-main-ui/asm-main-ui.component.ts
scarai-sap: Fixed

feature-libs/asm/components/asm-main-ui/asm-main-ui.component.ts
scarai-sap: Fixed

feature-libs/asm/components/asm-main-ui/asm-main-ui.component.ts
scarai-sap: Did similar work in asmComponentService, not sure we need an allowedList.

projects/storefrontapp-e2e-cypress/cypress/e2e/regression/asm/asm.e2e-flaky.cy.ts
scarai-sap: Fixed

feature-libs/asm/components/asm-main-ui/asm-main-ui.component.ts
```diff
@@ -132,6 +164,58 @@ export class AsmMainUiComponent implements OnInit, OnDestroy {
...
+   */
+  protected subscribeForDeeplink(): void {
+    if (this.featureConfig?.isLevel('6.2')) {
+      if (this.asmEnableService?.isEmulateInURL()) {
+        //Always route to home page to avoid 404
+        this.routingService.go('/');
+      }
+      const parameters = {
+        customerId: this.asmComponentService.getSearchParameter('customerId'),
+        someId: this.asmComponentService.getSearchParameter('someId'),
```
Zeyber: `someId` was just an example. We can remove it from here.

Zeyber (approved):

PR #17346, opened 2023-04-27
Feature/cxspa 3256 Add clear opf ui state after order complete and logout

rmch91:


integration-libs/opf/core/events/opf-ui-events.listener.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  constructor(
+    protected eventService: EventService,
+    protected opfService: OpfService
+  ) {
+    this.onOpfUiClear();
+  }
+
+  /**
+   * Registers events for the cart base actions.
```
Matejk00: I think this comment does not refer to functionality.

Matejk00 (changes requested):

integration-libs/opf/core/events/opf-ui-events.listener.ts
rmch91: changed, thanks

Matejk00 (approved):

PR #17354, opened 2023-05-01
feat: Run unit tests based on impact analysis given by Nx

bgambocjaviniar:
closes https://jira.tools.sap/browse/CXSPA-3299


.github/workflows/ci.yml
```diff
@@ -41,37 +43,14 @@ jobs:
         run: |
           source ci-scripts/validate-e2e-execution.sh
           echo "::set-output name=SHOULD_RUN_E2E::$(echo $RUN_E2E)"
-  unit_tests_core:
+  unit_tests:
```
plabadie, line 44: One of the reason the tests were split into 2 jobs/files was to reduce the run time by providing parallellism.  Does the new way of doin things make it longer to run the unit test suite when running the whole suite?

.github/workflows/ci.yml
```diff
@@ -10,6 +10,8 @@ on:
 env:
   CYPRESS_KEY: ${{ secrets.CYPRESS_KEY }}
   NODE_VERSION: '16'
+  NX_BASE: origin/${{ github.event.pull_request.base.ref }}
+  NX_HEAD: origin/${{ github.event.pull_request.head.ref }}
```
plabadie: This flow is onlu triggered for the "push" event and not for PRs, should we remove those?

ci-scripts/unit-tests.sh
```diff
@@ -2,108 +2,33 @@
...
-
-npx nx test s4om --source-map --no-watch --code-coverage --browsers ChromeHeadless
-
-echo "Running schematics unit tests and code coverage for s4om library"
-
-npm --prefix integration-libs/s4om run test:schematics -- --coverage
-
-echo "Running unit tests and code coverage for storefinder library"
+EXCLUDE_JEST=storefrontapp,storefrontstyles,schematics,setup
+EXCLUDE_INTEGRATION_LIBS=cdc,cds,digital-payments,epd-visualization,s4om
```
plabadie, line 7: We could reduce maintenance by gathering the names of libs in the integration folder, but this could be a future improvement.

ci-scripts/unit-tests.sh
```diff
@@ -2,108 +2,33 @@
...
-echo "Running unit tests and code coverage for eslint-rules"
+echo "-----"
 
-npx nx test eslint-rules --source-map --no-watch --code-coverage --browsers ChromeHeadless
+function run_affected_unit_tests {
+    echo "Running JASMINE unit tests and code coverage for AFFECTED libraries"
+    npx nx affected --target=test --exclude="$EXCLUDE_JEST,$EXCLUDE_INTEGRATION_LIBS" -- --no-watch --source-map --code-coverage --browsers ChromeHeadless
+
+    echo "Running JEST (mostly schematics) unit tests and code coverage for AFFECTED libraries"
+    npx nx affected --target=test-schematics --exclude="$EXCLUDE_INTEGRATION_LIBS" -- --coverage --runInBand
```
plabadie: Overall, the target could perhaps have been called "test-jest".  I don't mean to be picky though, I wonder if this could improve the readability.  Thoughts? Your call in the end.

.github/workflows/ci.yml
bgambocjaviniar: true

ci-scripts/unit-tests.sh
bgambocjaviniar: that's fine with me

.github/workflows/ci.yml
bgambocjaviniar: yup because it will run all. To me it's a must to run all now since we're only running affected ones in pull_requests

plabadie (approved):

PR #17355, opened 2023-05-01
Handle unhappy paths

FollowTheFlo:
added error messages.
added logic in service rather than component, more comprehensive.

### DIFF
--- a/integration-libs/opf/root/components/opf-payment-verification/opf-payment-verification.service.ts
+++ b/integration-libs/opf/root/components/opf-payment-verification/opf-payment-verification.service.ts
@@ -13,7 +13,7 @@ import {
   RoutingService,
 } from '@spartacus/core';
 import { Observable, of, throwError } from 'rxjs';
-import { switchMap } from 'rxjs/operators';
+import { concatMap } from 'rxjs/operators';
 import { OpfCheckoutFacade, OpfOrderFacade } from '../../facade';
 import {
   OpfPaymentVerificationResponse,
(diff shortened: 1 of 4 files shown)


integration-libs/opf/root/components/opf-payment-verification/opf-payment-verification.service.ts
```diff
@@ -33,4 +56,82 @@ export class OpfPaymentVerificationService {
...
+        )
+      );
+  }
+
+  isPaymentSuccessful(response: OpfPaymentVerificationResponse) {
+    if (
+      response.result === OpfPaymentVerificationResult.AUTHORIZED ||
+      response.result === OpfPaymentVerificationResult.DELAYED
+    ) {
+      return of(true);
```
FollowTheFlo, line 121: I set this function return Observable<boolean> but technically, it never returns false. It is either of(true) or throwError.
Value 'true' is never read, we could also return EMPTY or of(}), it does not matter in my opinion.

integration-libs/opf/root/components/opf-payment-verification/opf-payment-verification.service.ts
```diff
@@ -30,7 +54,87 @@ export class OpfPaymentVerificationService {
...
+        ...this.defaultError,
+        message: 'opf.checkout.errors.cancelPayment',
+      });
+    } else {
+      return throwError(this.defaultError);
+    }
+  }
+
+  displayError(error: HttpErrorModel | undefined): void {
+    console.log('displayError', error);
```
RadhepS: remove this line

RadhepS (changes requested):

RadhepS (approved):

PR #17361, opened 2023-05-02
Fix for a circular dependency from CXSPA-3056

rmch91:


FollowTheFlo (approved):
good to me, was also tested last friday.

PR #17362, opened 2023-05-02
CXSPA-3073: RxJs 7

ercultimate:


feature-libs/cart/base/core/facade/multi-cart.service.ts
```diff
@@ -76,7 +76,7 @@ export class MultiCartService implements MultiCartFacade {
       // This flickering should only be avoided when switching from false to true
       // Start of loading should be showed instantly (no debounce)
       // Extra actions are only dispatched after some loading
-      debounce((isStable) => (isStable ? timer(0) : EMPTY)),
+      debounce((isStable) => (isStable ? timer(0) : of(undefined))),
```
WeizhengSap, line 79: why we change it?

WeizhengSap (approved):

feature-libs/cart/base/core/facade/multi-cart.service.ts
ercultimate: See commit [fix undocumented behavior change in debounce](https://github.com/SAP/spartacus/pull/17362/commits/c984c53f73267c017cee5ec5cb4670578ae1d659)

PR #17377, opened 2023-05-05
feat: Handle Stripe option

Matejk00:


integration-libs/opf/components/opf-checkout-payment-wrapper/opf-checkout-payment-wrapper.service.ts
```diff
@@ -0,0 +1,103 @@
...
+  ) {}
+
+  protected activeCartId: string;
+
+  protected renderPaymentMethodEvent$ =
+    new BehaviorSubject<OpfRenderPaymentMethodEvent>({
+      isLoading: false,
+    });
+
+  getRenderPaymentMethodEvent(): BehaviorSubject<OpfRenderPaymentMethodEvent> {
```
rmch91: IMO better to create public observable from subject with `this.renderPaymentMethodEvent$.asObservable()`.
Currently you have private subject but add method that make it public with all it's build in methods, for example right now it's possible to do `.next` to this subject from `OpfCheckoutPaymentWrapperComponent`. (`renderPaymentMethodEvent$.next()`). I think better to encapsulate this inside of service.

I'm usually doing this with creating subscription, something like 
`renderPaymentMethodEventSub = new BehaviorSubject` and after this declaring 
`renderPaymentMethodEvent$ = this.renderPaymentMethodEventSub.asObservable()`

But you could also return this `this.renderPaymentMethodEvent$.asObservable()` in `getRenderPaymentMethodEvent()` method.

integration-libs/opf/components/opf-checkout-payment-wrapper/opf-checkout-payment-wrapper.service.ts
```diff
@@ -0,0 +1,103 @@
...
+    return combineLatest([
+      this.userIdService.getUserId(),
+      this.activeCartService.getActiveCartId(),
+    ]).pipe(
+      switchMap(([userId, cartId]) => {
+        this.activeCartId = cartId;
+        return this.opfOtpService.generateOtpKey(userId, cartId);
+      }),
+      filter((response) => Boolean(response?.value)),
+      map(({ value: otpKey }) => {
```
rmch91: I think we could move part from `map` to separate method in the service, to be closer to single responsibility and make this method leaner. To have something like: `map(({ value: otpKey }) => this.someName(paymentOptionId))`

rmch91 (changes requested):

rmch91 (approved):

PR #17379, opened 2023-05-05
Error handling for init payment endpoint

rmch91:


integration-libs/opf/components/opf-checkout-payment-wrapper/opf-checkout-payment-wrapper.service.ts
```diff
@@ -107,4 +104,36 @@ export class OpfCheckoutPaymentWrapperService {
         });
     }
   }
+
+  protected onInitPaymentError(): Observable<boolean> {
```
Matejk00: In my opinion 'on' prefix might suggest that we are dealing with some event, however we are naming a method which reacts on such event. What do you tihnk about method name change to `handlePaymentInitiationError`?

integration-libs/opf/components/opf-checkout-payment-wrapper/opf-checkout-payment-wrapper.component.ts
```diff
@@ -8,51 +8,51 @@ import {
...
   }
 
-  ngOnDestroy() {
-    this.subscription.unsubscribe();
+  protected initiatePayment(): void {
+    this.service
+      .initiatePayment(this.selectedPaymentId)
+      .pipe(take(1))
+      .subscribe((paymentOptionConfig: PaymentSessionData | boolean) => {
+        if (typeof paymentOptionConfig !== 'boolean') {
```
Matejk00, line 52: Can we use `filter` in a pipe instead of checking this inside subscription? 

I think this should work:

```
protected initiatePayment(): void {
  this.service
    .initiatePayment(this.selectedPaymentId)
    .pipe(
      take(1),
      filter(paymentOptionConfig => !Boolean(paymentOptionConfig))
    )
    .subscribe((paymentOptionConfig: PaymentSessionData) => {
      this.service.renderPaymentGateway(paymentOptionConfig);
    });
}
```

integration-libs/opf/components/opf-checkout-payment-wrapper/opf-checkout-payment-wrapper.component.ts
```diff
@@ -8,51 +8,51 @@ import {
...
   templateUrl: './opf-checkout-payment-wrapper.component.html',
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
-export class OpfCheckoutPaymentWrapperComponent implements OnInit, OnDestroy {
-  protected subscription = new Subscription();
+export class OpfCheckoutPaymentWrapperComponent implements OnInit {
+  @Input() selectedPaymentId: number;
 
   renderPaymentMethodEvent$ = this.service.getRenderPaymentMethodEvent();
+  isInitPaymentFailed$ = this.service.getIsInitPaymentFailed();
```
Matejk00: Instead of creating additional behaviour subject in the service. What do you think about extending responsibility of `renderPaymentMethodEvent$`?

I guess just `isError` property has to be added into `OpfRenderPaymentMethodEvent` interface.

```
export interface OpfRenderPaymentMethodEvent {
  isLoading: boolean;
  isError: boolean;
  renderType?: OpfPaymentMethodType | null;
  data?: string | null;
}
```

Matejk00 (changes requested):

integration-libs/opf/components/opf-checkout-payment-wrapper/opf-checkout-payment-wrapper.component.ts
rmch91: "I would really like to do this in a filter operator, but it won't work that way. I was thinking about it yesterday, but I didn't come up with anything. The `BooleanConstructor` won't filter a boolean, it will just cast whatever you give it to a boolean, so it will return either true or false. In my case, with this typing: `paymentOptionConfig: PaymentSessionData | boolean`, the result can be an object, true, or false. If we use `!Boolean(paymentOptionConfig)`, we exclude the cases where `paymentOptionConfig === undefined` or `paymentOptionConfig === false`, but it's possible, from a typing perspective, that `paymentOptionConfig === true`, and in that case, that value will be emitted."

integration-libs/opf/components/opf-checkout-payment-wrapper/opf-checkout-payment-wrapper.service.ts
rmch91: Make sense, changed

integration-libs/opf/components/opf-checkout-payment-wrapper/opf-checkout-payment-wrapper.component.ts
rmch91: Changed

Matejk00 (approved):

PR #17380, opened 2023-05-08
feat: CXSPA-3278 As an assisted service agent, I want to open deep link for Spartacus containing inactiveCartId 

i333055:
https://jira.tools.sap/browse/CXSPA-3278


feature-libs/asm/components/asm-bind-cart/asm-bind-cart.component.html
```diff
@@ -32,6 +32,8 @@
     type="submit"
     [class.cx-asm-active]="valid$ | async"
     [class.cx-bind-loading]="loading$ | async"
+    *ngIf="inactiveCartId.length === 0"
```
scarai-sap: maybe move the ngif to the top of the element to make it easier to read?

feature-libs/asm/styles/components/_asm-save-cart-dialog.component.scss
```diff
@@ -0,0 +1,214 @@
+$sapFiori_Button_Standard_TextColor: #1d2d3e;
```
scarai-sap, line 1: May you tell us the source of the scss(copied from xxx and modified sth to let us review easier?

feature-libs/asm/components/services/asm-component.service.ts
```diff
@@ -39,6 +42,14 @@ export class AsmComponentService {
     this.isEmulatedByDeepLink$.next(emulated);
   }
 
+  setShowInactiveCartInfoAlert(display: boolean) {
+    this.showInactiveCartInfoAlert$.next(display);
+  }
+
+  showInactiveCartInfoAlert() {
```
scarai-sap: provide return value in func declaration

feature-libs/asm/components/asm-bind-cart/asm-bind-cart.component.ts
```diff
@@ -81,15 +94,24 @@ export class AsmBindCartComponent implements OnInit, OnDestroy {
...
   ngOnInit(): void {
+    this.subscribeForDeeplinkInactiveCart();
+
     this.subscription.add(
       this.activeCartFacade.getActiveCartId().subscribe((response) => {
         this.activeCartId = response ?? '';
 
-        this.cartId.setValue(this.activeCartId);
+        this.cartId.setValue(
+          this.inactiveCartId.length > 0
```
scarai-sap: just use  this.inactiveCartId || this.activeCartId

feature-libs/asm/components/asm-main-ui/asm-main-ui.component.html
```diff
@@ -70,6 +70,15 @@
         >
         </cx-message>
       </ng-container>
+      <ng-container *cxFeatureLevel="'6.2'">
+        <cx-message
+          *ngIf="showInactiveCartInfoAlert"
```
scarai-sap: using xxxx$ | async, then you don't need write code for subscribe

feature-libs/asm/components/asm-save-cart-dialog/asm-save-cart-dialog.component.html
```diff
@@ -0,0 +1,91 @@
...
+        {{ 'asm.saveCart.dialog.title' | cxTranslate }}
+      </h2>
+    </div>
+
+    <!-- Modal Body -->
+    <div class="cx-dialog-body modal-body">
+      <div class="message-container">
+        <cx-message
+          *ngIf="showDialogAlert && cartQty > 0"
+          [text]="'asm.saveCart.dialog.saveInfo' | cxTranslate"
```
scarai-sap: I guess we can set text and type by ng-if instead of two  message-container?

feature-libs/asm/components/asm-bind-cart/asm-bind-cart.component.ts
```diff
@@ -81,15 +94,24 @@ export class AsmBindCartComponent implements OnInit, OnDestroy {
...
 
   ngOnInit(): void {
+    this.subscribeForDeeplinkInactiveCart();
+
     this.subscription.add(
       this.activeCartFacade.getActiveCartId().subscribe((response) => {
         this.activeCartId = response ?? '';
 
-        this.cartId.setValue(this.activeCartId);
+        this.cartId.setValue(
```
scarai-sap: maybe in here always show cartID from URL.

feature-libs/asm/components/asm-bind-cart/asm-bind-cart.component.ts
```diff
@@ -70,8 +79,12 @@ export class AsmBindCartComponent implements OnInit, OnDestroy {
   );
 
   activeCartId = '';
+  inactiveCartId = '';
+  savedInactiveCart = false;
```
scarai-sap: not sure we really need global variable for them.  for boolean, rename savedInactiveCart to IsinactiveCartSaved?

feature-libs/asm/components/asm-bind-cart/asm-bind-cart.component.ts
```diff
@@ -147,8 +169,59 @@ export class AsmBindCartComponent implements OnInit, OnDestroy {
...
+        take(1),
+        map((state) => state.value as Cart),
+        concatMap((cart) =>
+          iif(() => Boolean(cart !== null), this.openASMSaveCartDialog(cart))
+        )
+      )
+      .subscribe();
+
+    this.savedCartFacade
+      .getSaveCartProcessSuccess()
```
scarai-sap: remove filter as it always raise success /error

feature-libs/asm/components/services/asm-component.service.ts
Zeyber: We should use typings to show the return values `(): Observable<boolean>`.

feature-libs/asm/components/asm-bind-cart/asm-bind-cart.component.ts
```diff
@@ -219,4 +306,70 @@ export class AsmBindCartComponent implements OnInit, OnDestroy {
...
+        cartId: inactiveCartId,
+        saveCartName: inactiveCartId,
+        saveCartDescription: '-',
+      });
+    });
+  }
+
+  protected goToSavedCartDetails(cartId: string): void {
+    this.routing.go({
+      cxRoute: 'savedCartsDetails',
```
scarai-sap, line 366: not sure it's a good idea adding routine here? Or do every routing in asm-main is better idea? In ASM main we are listening the dialog close info and navigate to other pages.  Maybe keep all routing in there is a better idea?

scarai-sap (commented):
Some comments added

feature-libs/asm/components/asm-bind-cart/asm-bind-cart.component.ts
i333055: for mockup, after navigate to saved cart detail we need set this save button disappear, so used this savedInactiveCart

feature-libs/asm/components/asm-save-cart-dialog/asm-save-cart-dialog.component.ts
```diff
@@ -0,0 +1,69 @@
...
+
+@Component({
+  selector: 'cx-asm-save-cart-dialog',
+  templateUrl: './asm-save-cart-dialog.component.html',
+})
+export class AsmSaveCartDialogComponent implements OnInit, OnDestroy {
+  BIND_CART_ACTION = SAVE_CART_DIALOG_ACTION;
+  showDialogAlert = true;
+  globalMessageType = GlobalMessageType;
+  private subscription = new Subscription();
```
Zeyber: In general, we prefer not to use `private` in our library because this makes the implementation invisible to users. Sometimes our lib users might like to customize a method. Methods we prefer users not to touch we use `protected` for as this still allows users to changes them if they really want to.

feature-libs/asm/components/asm-bind-cart/asm-bind-cart.component.ts
scarai-sap: OK it's used for html ng-if display

feature-libs/asm/components/asm-main-ui/asm-main-ui.component.html
```diff
@@ -70,6 +70,15 @@
         >
         </cx-message>
       </ng-container>
+      <ng-container *cxFeatureLevel="'6.2'">
+        <cx-message
```
Zeyber, line 74: The `AsmMainUiComponent` is rather bloated at this point making it difficult to interpret. The logic for the alert seems rather isolated though. Is it possible to move this alert into its own component with service instead and use that component here in the template (eg. `InactiveCartInfoAlertComponent` and `InactiveCartInfoAlertComponentService`)?

feature-libs/asm/components/asm-bind-cart/asm-bind-cart.component.ts
```diff
@@ -219,4 +306,70 @@ export class AsmBindCartComponent implements OnInit, OnDestroy {
...
+  protected isDeepLinkInactiveCart(): boolean {
+    const cartId = this.asmComponentService.getSearchParameter('cartId');
+    const cartType = this.asmComponentService.getSearchParameter('cartType');
+    if (cartType === 'inactive' && !this.isEmptyStr(cartId)) {
+      return true;
+    }
+    return false;
+  }
+
+  protected isEmptyStr(string?: string): boolean {
```
Zeyber: This sort of function is more of a utility function. You can check `utils.ts` files to see if this already exists. Otherwise, please add it.

Zeyber (changes requested):
Since the functionality should exist at this point, I would recommend writing the e2e tests so that we can quickly figure out where there are issues in test cases (if any) and if any code changes break anything. Units can wait until we finalize the components and services though.

Could we also please leave comments on methods where its less obvious what may be happening? This helps lib users as well as us. `auth-http-header.service.ts` is a good example of this being done :)

feature-libs/asm/components/asm-main-ui/asm-main-ui.component.html
i333055: Hi @Zeyber，there is a little logic to close alert in **AsmMainUiComponent**，seems no need to create a new component for InactiveCartInfoAlertComponent

```
  closeInactiveCartInfoAlert(): void {
    this.asmComponentService.setShowInactiveCartInfoAlert(false);
  }
```

feature-libs/asm/components/asm-bind-cart/asm-bind-cart.component.ts
i333055: thanks for your suggestion, seems `!cartId` is ok for me

feature-libs/asm/components/asm-bind-cart/asm-bind-cart.component.ts
i333055: fixed

feature-libs/asm/components/asm-bind-cart/asm-bind-cart.component.ts
i333055: fixed

feature-libs/asm/styles/components/_asm-save-cart-dialog.component.scss
i333055: copied from `_asm-bind-cart-dialog.component.scss`

feature-libs/asm/components/asm-save-cart-dialog/asm-save-cart-dialog.component.ts
i333055: fixed

feature-libs/asm/components/services/asm-component.service.ts
i333055: fixed

i333055: fixed

feature-libs/asm/components/asm-bind-cart/asm-bind-cart.component.html
i333055: fixed

feature-libs/asm/components/asm-main-ui/asm-main-ui.component.html
Zeyber: In simpler cases, I wouldn't mind additional logic in a component. But `AsmMainUiComponent` has reached 300 lines. It is also not a very straight forward component. Additional flags and trigger methods are being added on top of this component that are responsible for isolatable functions. It is becoming more difficult to figure out which variables may affect what function without deep diving (increasing cognitive complexity).

i333055: I understand your concern, but for this alert the logic has been extracted to asmComponentService and already exist component MessageComponent, I think there is no need to add new component InactiveCartInfoAlertComponent, maybe for other exist complex logic code should be refined.

feature-libs/asm/components/asm-bind-cart/asm-bind-cart.component.spec.ts
```diff
@@ -1,25 +1,50 @@
...
+    return Promise.resolve();
+  }
+  isCustomerSupportAgentLoggedIn(): Observable<boolean> {
+    return of(false);
+  }
+  getCustomerSupportAgentTokenLoading(): Observable<boolean> {
+    return of(false);
+  }
+  startCustomerEmulationSession(_customerId: string) {}
+}
```
scarai-sap, line 47:
```suggestion
}

```

feature-libs/asm/components/asm-bind-cart/asm-bind-cart.component.html
```diff
@@ -41,4 +43,25 @@
       [attr.aria-label]="'common.loading' | cxTranslate"
     ></cx-dot-spinner>
   </button>
+
+  <ng-container *cxFeatureLevel="'6.2'">
+    <button
+      id="asm-save-inactive-cart-btn"
+      #saveInactiveCart
+      *ngIf="inactiveCartId.length > 0 && !isInactiveCartSaved"
```
scarai-sap:
```suggestion
      *ngIf="inactiveCartId && !isInactiveCartSaved"
```

feature-libs/asm/components/asm-bind-cart/asm-bind-cart.component.html
```diff
@@ -41,4 +43,25 @@
...
+      class="cx-asm-bindCartToCustomer"
+      type="submit"
+      [class.cx-asm-active]="inactiveCartId.length > 0"
```
scarai-sap:
```suggestion
      [class.cx-asm-active]="!!inactiveCartId"
```

feature-libs/asm/components/asm-bind-cart/asm-bind-cart.component.ts
```diff
@@ -81,15 +94,22 @@ export class AsmBindCartComponent implements OnInit, OnDestroy {
     protected multiCartFacade: MultiCartFacade,
     protected asmBindCartFacade: AsmBindCartFacade,
     protected launchDialogService: LaunchDialogService,
-    protected savedCartFacade: SavedCartFacade
+    protected savedCartFacade: SavedCartFacade,
```
scarai-sap, line 124: constructor is changed!

feature-libs/asm/components/asm-bind-cart/asm-bind-cart.component.ts
```diff
@@ -147,8 +167,61 @@ export class AsmBindCartComponent implements OnInit, OnDestroy {
...
+
+    this.savedCartFacade
+      .getSaveCartProcessError()
+      .pipe(
+        filter((error) => error),
+        take(1),
+        tap(() => this.loading$.next(false))
+      )
+      .subscribe(() => {
+        this.isInactiveCartSaved = true;
```
scarai-sap: set to true even save failed,  we can't retry even the failure is caused by server/network issue?

feature-libs/asm/components/asm-bind-cart/asm-bind-cart.component.ts
```diff
@@ -219,4 +306,67 @@ export class AsmBindCartComponent implements OnInit, OnDestroy {
...
   }
+
+  protected subscribeForDeeplinkInactiveCart(): void {
+    if (this.featureConfig?.isLevel('6.2')) {
+      this.subscription.add(
+        this.asmComponentService
+          .isEmulatedByDeepLink()
+          .pipe(filter((emulated) => emulated && this.isDeepLinkInactiveCart()))
+          .subscribe(() => {
+            this.isInactiveCartSaved = false;
```
scarai-sap, line 309: why we set it in here again?

feature-libs/asm/components/asm-bind-cart/asm-bind-cart.component.ts
```diff
@@ -219,4 +306,67 @@ export class AsmBindCartComponent implements OnInit, OnDestroy {
...
+            this.asmComponentService.setShowInactiveCartInfoAlert(true);
+          })
+      );
+    }
+  }
+
+  protected isDeepLinkInactiveCart(): boolean {
+    const cartId = this.asmComponentService.getSearchParameter('cartId');
+    const cartType = this.asmComponentService.getSearchParameter('cartType');
+    if (cartType === 'inactive' && !!cartId) {
```
scarai-sap:
```suggestion
   return cartType === 'inactive' && !!cartId
```

feature-libs/asm/components/asm-bind-cart/asm-bind-cart.component.ts
```diff
@@ -219,4 +306,67 @@ export class AsmBindCartComponent implements OnInit, OnDestroy {
...
+      return true;
+    }
+    return false;
+  }
+
+  protected openASMSaveCartDialog(inactiveCart: Cart) {
```
scarai-sap: declare the return value

feature-libs/asm/components/asm-bind-cart/asm-bind-cart.component.ts
```diff
@@ -219,4 +306,67 @@ export class AsmBindCartComponent implements OnInit, OnDestroy {
...
+        inactiveCart
+      );
+      return this.launchDialogService.dialogClose.pipe(
+        filter((result) => Boolean(result)),
+        take(1)
+      ) as Observable<SAVE_CART_DIALOG_ACTION>;
+    }).pipe(
+      filter((dialogResult) => Boolean(dialogResult)),
+      concatMap((dialogResult) => {
+        return this.selectSaveAction(inactiveCart.code as string, dialogResult);
```
scarai-sap: why need as string?

feature-libs/asm/components/asm-main-ui/asm-main-ui.component.ts
```diff
@@ -30,8 +30,8 @@ import {
 } from '@spartacus/core';
 import {
   ICON_TYPE,
-  LAUNCH_CALLER,
   LaunchDialogService,
+  LAUNCH_CALLER,
```
scarai-sap, line 35: why move it to the end of list? the old sequence should be right

feature-libs/asm/components/services/asm-component.service.ts
```diff
@@ -39,6 +42,14 @@ export class AsmComponentService {
     this.isEmulatedByDeepLink$.next(emulated);
   }
 
+  setShowInactiveCartInfoAlert(display: boolean) {
+    this.showInactiveCartInfoAlert$.next(display);
+  }
+
+  showInactiveCartInfoAlert(): BehaviorSubject<boolean> {
```
scarai-sap: rename to isXxxx/shouldxxx if the return is a  boolean like

feature-libs/asm/components/services/asm-component.service.ts
```diff
@@ -18,6 +18,9 @@ import { BehaviorSubject, Observable } from 'rxjs';
 export class AsmComponentService {
   protected searchparam: URLSearchParams;
   isEmulatedByDeepLink$: BehaviorSubject<boolean> = new BehaviorSubject(false);
+  showInactiveCartInfoAlert$: BehaviorSubject<boolean> = new BehaviorSubject(
```
scarai-sap: How about change to shouldShowxxxx

scarai-sap (commented):
comment added.

There are some 'as string' not sure it's mandatory

scarai-sap:
when changing the cart number and click Save for later,  we still using the cart number in URL.

Please check with PO whether it's an issue

scarai-sap (commented):
if input cart in url is same as active cart, we should handle it;

projects/storefrontapp-e2e-cypress/cypress/e2e/vendor/asm/b2c/asm.e2e-flaky.cy.ts
```diff
@@ -271,4 +381,46 @@ context('Assisted Service Module', () => {
       });
     });
   }
+
+  function getInactiveCartId(customerEmail, customerPwd, isEmpty: boolean) {
```
Zeyber: Please export this function from a helper file and comment what it is doing.

projects/storefrontapp-e2e-cypress/cypress/e2e/vendor/asm/b2c/asm.e2e-flaky.cy.ts
```diff
@@ -271,4 +381,46 @@ context('Assisted Service Module', () => {
...
+    let token = null;
+    let inactiveCartId = null;
+    return new Promise((resolve, reject) => {
+      fetchingToken(customerEmail, customerPwd, false).then((res) => {
+        token = res.body.access_token;
+        createCart(token).then((response) => {
+          if (response.status === 201) {
+            inactiveCartId = response.body.code;
+            if (!isEmpty) {
+              addToCart(inactiveCartId, '1934793', '2', token).then(
```
Zeyber: Can we please move less obvious strings to constants and name them? This makes it easier to read and change later.

projects/storefrontapp-e2e-cypress/cypress/e2e/vendor/asm/b2c/asm.e2e-flaky.cy.ts
```diff
@@ -141,6 +143,114 @@ context('Assisted Service Module', () => {
...
+    });
+
+    it('should not save empty inative cart in deeplink after agent login (CXSPA-3278)', () => {
+      const customer = getSampleUser();
+      cy.log('--> Register new user');
+      cy.visit('/?asm=true');
+      checkout.registerUser(false, customer);
+
+      cy.visit('/?asm=true');
+      asm.agentLogin(agentToken.userName, agentToken.pwd);
```
Zeyber: This section appears to be repeated in tests. Perhaps we can move it to its own function?

projects/storefrontapp-e2e-cypress/cypress/e2e/vendor/asm/b2c/asm.e2e-flaky.cy.ts
```diff
@@ -141,6 +143,114 @@ context('Assisted Service Module', () => {
       });
     });
 
+    it('should save inative cart in deeplink after agent login (CXSPA-3278)', () => {
```
Zeyber: *inactive

projects/storefrontapp-e2e-cypress/cypress/e2e/vendor/asm/b2c/asm.e2e-flaky.cy.ts
```diff
@@ -141,6 +143,114 @@ context('Assisted Service Module', () => {
...
+      const customer = getSampleUser();
+      cy.log('--> Register new user');
+      cy.visit('/?asm=true');
+      checkout.registerUser(false, customer);
+
+      cy.visit('/?asm=true');
```
Zeyber: I don't think the second `visit` is necessary.

feature-libs/asm/components/asm-save-cart-dialog/asm-save-cart-dialog.component.spec.ts
```diff
@@ -0,0 +1,100 @@
...
+  it('should show meaasge based on `cx-message` selector', () => {
+    expect(el.query(By.css('cx-message'))).not.toBeNull();
+    component.closeDialogAlert();
+  });
+
+  it('should show meaasge based on `cx-message` selector', () => {
+    component.closeDialogAlert();
+    expect(component.showDialogAlert).toEqual(false);
+  });
+});
```
Zeyber, line 131: Missing case where save button should be disabled when no `cartQty`.

feature-libs/asm/components/services/asm-component.service.ts
Zeyber: Lets return this as an observable as it should only be modified by the service method.

feature-libs/asm/components/services/asm-component.service.ts
Zeyber: Lets protect this variable since it should only be accessed and modified by service methods.

feature-libs/asm/components/asm-save-cart-dialog/asm-save-cart-dialog.component.ts
```diff
@@ -0,0 +1,69 @@
...
+  ngOnInit(): void {
+    this.subscription.add(
+      this.launchDialogService.data$.subscribe((data: any) => {
+        this.cart = data;
+        this.calEntitiesQty();
+      })
+    );
+  }
+
+  calEntitiesQty(): void {
```
Zeyber: Maybe a more descriptive name that indicates how the qty is calculated and where it set?

feature-libs/asm/components/asm-save-cart-dialog/asm-save-cart-dialog.component.ts
```diff
@@ -0,0 +1,69 @@
...
+  selector: 'cx-asm-save-cart-dialog',
+  templateUrl: './asm-save-cart-dialog.component.html',
+})
+export class AsmSaveCartDialogComponent implements OnInit, OnDestroy {
+  BIND_CART_ACTION = SAVE_CART_DIALOG_ACTION;
+  showDialogAlert = true;
+  globalMessageType = GlobalMessageType;
+  protected subscription = new Subscription();
+  cart: Cart;
+  cartQty: number;
```
Zeyber, line 29: Lets use `BehaviourSubjects` for these since they are set dynamically. Then template values can be read asynchronously. This is safer when we cannot immediately infer a value.

projects/storefrontapp-e2e-cypress/cypress/e2e/vendor/asm/b2c/asm.e2e-flaky.cy.ts
Zeyber: Maybe we can add a few more products here and assert the values add up correctly in the save cart dialog?

feature-libs/asm/components/asm-save-cart-dialog/asm-save-cart-dialog.component.ts
```diff
@@ -0,0 +1,69 @@
...
+    block: true,
+    autofocus: true,
+    focusOnEscape: true,
+  };
+
+  constructor(protected launchDialogService: LaunchDialogService) {}
+
+  ngOnInit(): void {
+    this.subscription.add(
+      this.launchDialogService.data$.subscribe((data: any) => {
```
Zeyber: Do we need to stay subscribed? Can we `take(1)`?

feature-libs/asm/components/asm-save-cart-dialog/asm-save-cart-dialog.component.ts
```diff
@@ -0,0 +1,69 @@
...
+  SAVE = 'SAVE',
+}
+
+@Component({
+  selector: 'cx-asm-save-cart-dialog',
+  templateUrl: './asm-save-cart-dialog.component.html',
+})
+export class AsmSaveCartDialogComponent implements OnInit, OnDestroy {
+  BIND_CART_ACTION = SAVE_CART_DIALOG_ACTION;
+  showDialogAlert = true;
```
Zeyber: This dynamically changes the template with `*ngIf`. Can we please use a `BehaviorSubject` and check it with the `async` pipe?

feature-libs/asm/components/asm-main-ui/asm-main-ui.component.ts
```diff
@@ -56,6 +56,7 @@ export class AsmMainUiComponent implements OnInit, OnDestroy {
   isCollapsed$: Observable<boolean> | undefined;
   iconTypes = ICON_TYPE;
 
+  showInactiveCartInfoAlert$: Observable<boolean>;
```
Zeyber: Can we instantiate this immediately? 
`this.showInactiveCartInfoAlert$: Observable<boolean> =
        this.asmComponentService.showInactiveCartInfoAlert();`

feature-libs/asm/components/asm-bind-cart/asm-bind-cart.component.html
```diff
@@ -1,4 +1,4 @@
-<form (submit)="bindCartToCustomer()">
+<form>
```
Zeyber, line 1: Just to test, will pressing the "Enter" key on the input field of this form still submit the form?

feature-libs/asm/components/asm-bind-cart/asm-bind-cart.component.html
```diff
@@ -27,11 +27,13 @@
   </div>
   <button
     #bindToCart
+    *ngIf="inactiveCartId.length === 0"
```
Zeyber, line 30: I see in the new button `inactiveCartId.length > 0 && !isInactiveCartSaved` which isn't directly opposing this `inactiveCartId.length === 0` condition. Could it be possible to get into the scenario that neither button appears?

feature-libs/asm/components/asm-bind-cart/asm-bind-cart.component.html
Zeyber: The `ngIf` statement seems to indicate that this class will always be active as long as the button exists. Do we need the condition for the class?

feature-libs/asm/components/asm-bind-cart/asm-bind-cart.component.ts
```diff
@@ -10,17 +10,24 @@ import {
   ElementRef,
```
Zeyber, line 10: This is becoming a long component file. Could we please move some functions to a service to tidy this up a bit?

Zeyber (changes requested):

feature-libs/asm/components/asm-bind-cart/asm-bind-cart.component.html
i333055: yes, it still submit the form when pressing "Enter" key

feature-libs/asm/components/asm-bind-cart/asm-bind-cart.component.html
i333055: yes, according the mockup, after navigate to saved cart detail page, the button will not appear
image

feature-libs/asm/components/asm-bind-cart/asm-bind-cart.component.ts
i333055: fixed

feature-libs/asm/components/asm-bind-cart/asm-bind-cart.component.ts
i333055: for every time deeplink to save inactive cart, this property should be false

feature-libs/asm/components/asm-bind-cart/asm-bind-cart.component.ts
i333055: for the define Cart.code?: string | undefined

feature-libs/asm/components/asm-main-ui/asm-main-ui.component.ts
i333055: when I save file, it automatic change the sequence

feature-libs/asm/components/asm-main-ui/asm-main-ui.component.ts
i333055: that's ok, even if instantiate this, it will not be used before featureflag 6.2

feature-libs/asm/components/asm-bind-cart/asm-bind-cart.component.ts
i333055: fixed

feature-libs/asm/components/asm-save-cart-dialog/asm-save-cart-dialog.component.spec.ts
i333055: this case is tested in e2e test

feature-libs/asm/components/asm-bind-cart/asm-bind-cart.component.ts
niehuayang, line 124: Please be aware that this constructor change is considered a breaking change. see [break change wiki](https://wiki.one.int.sap/wiki/display/SAPCCWiki/Spartacus+Onboarding).(
Making any changes to the class constructor signature.)

feature-libs/asm/components/asm-save-cart-dialog/asm-save-cart-dialog.component.ts
niehuayang: Please avoid using any. It probably makes sense to use the type `Cart` here.

feature-libs/asm/components/asm-bind-cart/asm-bind-cart.component.ts
i333055: refined, move save logic to AsmSaveCartDialogComponent

feature-libs/asm/components/asm-bind-cart/asm-bind-cart.component.ts
i333055: disccussed with jason, if failed the button still exist

feature-libs/asm/components/asm-bind-cart/asm-bind-cart.component.ts
i333055: fixed

i333055:
> if input cart in url is same as active cart, we should handle it;

discussed with jason, if url changed the button will change to the bind cart button

projects/storefrontapp-e2e-cypress/cypress/support/utils/cart.ts
```diff
@@ -40,10 +40,34 @@ export function addToCart(
...
+/**
+ * @param accessToken access token
+ * @returns promise of inactive card id
+ */
+export function createInactiveCart(accessToken: string): Promise<string> {
+  let inactiveCartId = null;
+  return new Promise((resolve, reject) => {
+    createCart(accessToken).then((response) => {
+      if (response.status === 201) {
+        inactiveCartId = response.body.code;
```
scarai-sap: move inactiveCartId define in here

feature-libs/asm/components/asm-bind-cart/asm-bind-cart.component.ts
```diff
@@ -219,4 +295,73 @@ export class AsmBindCartComponent implements OnInit, OnDestroy {
...
+    }
+  }
+
+  protected isDeepLinkInactiveCart(): boolean {
+    const cartId = this.asmComponentService?.getSearchParameter('cartId');
+    const cartType = this.asmComponentService?.getSearchParameter('cartType');
+    return cartType === 'inactive' && !!cartId;
+  }
+
+  protected openASMSaveCartDialog(inactiveCart: Cart) {
```
scarai-sap:
```suggestion
  protected openASMSaveCartDialog(inactiveCart: Cart): void {
```

feature-libs/asm/components/asm-bind-cart/asm-bind-cart.component.ts
```diff
@@ -219,4 +295,73 @@ export class AsmBindCartComponent implements OnInit, OnDestroy {
...
+    this.launchDialogService.openDialogAndSubscribe(
+      LAUNCH_CALLER.ASM_SAVE_CART,
+      this.saveInactiveCartElemRef,
+      inactiveCart
+    );
+  }
+
+  protected afterCloseASMSaveCartDialog() {
```
scarai-sap:
```suggestion
  protected afterCloseASMSaveCartDialog(): void {
```

scarai-sap (commented):
small suggestions

scarai-sap:
there is css issue, when cart is valid inactive cart, the button looks it was disabled(but clickable) 

image

scarai-sap:
Another issue is, whether we need keep dialog till save succeed?  currently looks the dialog will disappear once we clicked save no matter what's the result will be?

i333055:
> Another issue is, whether we need keep dialog till save succeed? currently looks the dialog will disappear once we clicked save no matter what's the result will be?

yes

i333055:
> there is css issue, when cart is valid inactive cart, the button looks it was disabled(but clickable)
> 
> image

fixed

scarai-sap (approved):

feature-libs/asm/components/asm-bind-cart/asm-bind-cart.component.ts
```diff
@@ -70,26 +83,57 @@ export class AsmBindCartComponent implements OnInit, OnDestroy {
...
+    multiCartFacade: MultiCartFacade,
+    asmBindCartFacade: AsmBindCartFacade,
+    launchDialogService: LaunchDialogService,
+    savedCartFacade: SavedCartFacade,
+    // eslint-disable-next-line @typescript-eslint/unified-signatures
+    asmComponentService: AsmComponentService,
+    routing: RoutingService,
+    featureConfig: FeatureConfigService
+  );
+  constructor(
```
niehuayang, line 110: Could you please add a comment as follow ?
```
  /**
   * @deprecated since 7.0
   */
```

niehuayang (approved):
Looks good to me, I left a small comment for you to choose whether to fix or not

projects/storefrontapp-e2e-cypress/cypress/support/utils/cart.ts
```diff
@@ -40,10 +40,33 @@ export function addToCart(
       product: {
         code: productCode,
       },
-      qty: quantity,
+      quantity: quantity,
```
Zeyber, line 77: I think that changing this value will break other e2e tests that use this function.

projects/storefrontapp-e2e-cypress/cypress/e2e/vendor/asm/b2c/asm.e2e-flaky.cy.ts
```diff
@@ -248,6 +354,15 @@ context('Assisted Service Module', () => {
     });
   });
 
+  function emulateCustomerPrepare() {
```
Zeyber: Could we move this function to `... /helpers/asm` and export it?

projects/storefrontapp-e2e-cypress/cypress/e2e/vendor/asm/b2c/asm.e2e-flaky.cy.ts
```diff
@@ -271,4 +386,35 @@ context('Assisted Service Module', () => {
       });
     });
   }
+
+  function getInactiveCartIdAndAddProducts(
```
Zeyber, line 703: Could we move this function to `... /helpers/asm` and export it?

feature-libs/asm/components/asm-save-cart-dialog/asm-save-cart-dialog.component.ts
scarai-sap: they are fixed instead of dynamically, we can't change cart content in this scenario. So it should be OK

Zeyber (changes requested):

projects/storefrontapp-e2e-cypress/cypress/e2e/vendor/asm/b2c/asm.e2e-flaky.cy.ts
i333055: I don't think it needed, for the base common function `createInactiveCart` has extracted to helper, this is prepared for our test case, not so common

projects/storefrontapp-e2e-cypress/cypress/support/utils/cart.ts
i333055: ok，there is a bug for this method for the quantity not work，I will add a new method and comment it.

projects/storefrontapp-e2e-cypress/cypress/e2e/vendor/asm/b2c/asm.e2e-flaky.cy.ts
i333055: moved

Zeyber (approved):

PR #17383, opened 2023-05-08
CXSPA-3076: Cleanup deprecated `toPromise()`

ercultimate:


feature-libs/asm/root/services/asm-auth.service.ts
```diff
@@ -109,9 +109,8 @@ export class AsmAuthService extends AuthService {
    * To perform logout it is best to use `logout` method. Use this method with caution.
    */
   coreLogout(): Promise<any> {
-    return this.userIdService
-      .isEmulated()
-      .pipe(
+    return lastValueFrom(
```
plabadie, line 112: Could it be firstValueFrom() ?
Just asking out of curiosity. I presume that as long as the stream uses a take(1) it doesn't make a difference, of maybe the purpose of the take(1) was to play well with toPromise();

I guess the operator with the closest behaviour to `toPromise` is indeed `lastValueFrom`, but I also saw a pattern in the rest of the PR where we change the take(1) + toPromise() combo to firstValueFrom()

plabadie (approved):

feature-libs/asm/root/services/asm-auth.service.ts
ercultimate: In those other instances, I was using `firstValueFrom` to drop the need for the trailing `take()` in the operators.

I used last value here because I took a glance at `return from(super.coreLogout());` as possibly emitting multiple times.  Now that I look at it properly, it's guaranteed to emit once, then complete.  

Overall, the resulting observable is guaranteed to emit once and immediately complete, so there isn't a huge difference between either first or last value.  I'll update it if you would like.

On a side note (not sure if you were asking in this direction), the `take(1)` here shouldn't be removed because of the `switchMap`.  Observable `isEmulated()` could emit again while waiting for the `coreLogout()` promise to resolve, dropping the result and switching to the new observable.

PR #17408, opened 2023-05-11
feat: retry OPF placeOrder and paymentVerify calls when 500 statusCode

FollowTheFlo:
Cover CXSPA-3168 and CXSPA-3167
2 attempts on OPF Adapter Verify request when response is http statusCode 5xx
2 attempts on OCC PlaveOrderV2 request when response is http statusCode 5xx
image

### DIFF
--- a/integration-libs/opf/occ/utils/opf-occ-http-error-handlers.ts
+++ b/integration-libs/opf/occ/utils/opf-occ-http-error-handlers.ts
@@ -1,3 +1,9 @@
+/*
+ * SPDX-FileCopyrightText: 2023 SAP Spartacus team <spartacus-team@sap.com>
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
 import { HttpErrorModel } from '@spartacus/core';
 
 export function isHttp500Error(err: HttpErrorModel): boolean {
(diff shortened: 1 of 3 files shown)


integration-libs/opf/occ/utils/opf-occ-http-error-handlers.ts
```diff
@@ -0,0 +1,11 @@
+/*
+ * SPDX-FileCopyrightText: 2023 SAP Spartacus team <spartacus-team@sap.com>
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { HttpErrorModel } from '@spartacus/core';
+
+export function isHttp500Error(err: HttpErrorModel): boolean {
+  return err?.status && err.status >= 500 ? true : false;
```
rmch91: should be enough `return !!err?.status && err.status >= 500;` or `return Boolean(err?.status) && err.status >= 500;`, without ternary operator.

rmch91 (changes requested):

integration-libs/opf/occ/utils/opf-occ-http-error-handlers.ts
FollowTheFlo: Thanks! change done.

Matejk00 (approved):

PR #17413, opened 2023-05-12
add condition to display cx-card only when cardNumber exists

FollowTheFlo:


feature-libs/order/components/order-details/order-overview/order-overview.component.html
```diff
@@ -95,7 +95,11 @@
         </ng-container>
       </div>
 
-      <ng-container *ngIf="order.paymentInfo">
+      <ng-container
+        *ngIf="
+          order?.paymentInfo?.billingAddress || order?.paymentInfo?.cardNumber
+        "
+      >
```
FollowTheFlo, line 103: display 3rd column (cx-summary-card) only if at least one of the 2 cx-card will be displayed.

PR #17427, opened 2023-05-19
CXSPA-3478: fix display incomplete by account sorting in b2bCustomerList dialog

evy-yang01:
https://jira.tools.sap/browse/CXSPA-3478


feature-libs/asm/styles/components/_customer-list.component.scss
```diff
@@ -412,5 +412,11 @@ $sapFioriContent_LabelColor: #6a6d70;
         box-shadow: none;
       }
     }
+
+    .sort-list {
```
niehuayang: Create a new class seems unnecessary, just add it directly
eg:
```

.ng-select {
    .ng-select-container {
      border: 0;
      .ng-arrow-wrapper {
        .ng-arrow {
          border-color: $sapFioriLinkColor transparent transparent;
        }
      }
    }
     @include forVersion(6.1) {
      .ng-dropdown-panel {
        width: auto;
      }
    }
  }

```

feature-libs/asm/components/customer-list/customer-list.component.html
```diff
@@ -463,7 +463,7 @@ <h2 id="asm-customer-list-title" class="title modal-title">
       'asm.customerList.tableSort.sortBy' | cxTranslate
     }}</span>
     <cx-sorting
-      class="sort-selector"
+      class="sort-selector sort-list"
```
niehuayang: so, maybe remove it

niehuayang (approved):
Approved, and I would prefer to put this class together with 'ng-select', so it will look better

PR #17436, opened 2023-05-19
Split OPF lib in two features: Checkout & Payment

FollowTheFlo:
To support lazy loading, OPF is separated in 2 features:
- Checkout -> all logic/UI running in checkout steps
- Payment -> all the rest: PaymentVerification, maybe QuickBuy in future

Technically, changes are:
- 2 new folders:
--'Checkout' which contains all core/root/occ/adapters/cpnts/translation same as before without the PaymentVerification and OpfOrder part
--'Payment' which contains all core/root/occ/adapters/cpnts/translation same as before without the checkout part
-Schematics change:
--added 2 features 'OPF_PAYMENT_FEATURE' and 'OPF_CHECKOUT_FEATURE'. 
OPF_PAYMENT_FEATURE is lazy loaded and opfCheckout is loaded by CheckoutWrapper
- added paths in tsconfig.ts for the 2 features


Matejk00 (changes requested):
Regarding the naming convention, I would suggest splitting this library into two main parts:
- "base": This section can contain all the reusable components and services related to the overall functionality of the library.
- "checkout": This section can contain all the components and services specifically related to the checkout process.

This approach has previously been used in a similar "cart" feature.

Alternatively, you can consider organizing the library similar to the digital payments integration library. In this case, you can keep all the folders in the root directory, but move the components and services that are intended for use in the checkout process into a separate folder such as "src/checkout."

Example: integration-libs/digital-payments

Matejk00 (changes requested):
Additionally, based on the pipeline, there are still several checks that need to be fixed. For instance, configuration updates and linting.

PR #17439, opened 2023-05-22
Refactor opf-checkout-billing-address component

rmch91:
- Fixed bug with old address in subject by moving service to component providers from module. Now Service destroyed with the component and state of service properties are fresh related to component
- Removed delivery address from component and service, because basically component should not be interested in what it is displaying, delivery address or billing, thanks to this a little bit simplified logic inside of template and component.
- isSameAsDelivery flag moved to the service

### DIFF
--- a/integration-libs/opf/components/opf-checkout-billing-address-form/opf-checkout-billing-address-form.component.html
+++ b/integration-libs/opf/components/opf-checkout-billing-address-form/opf-checkout-billing-address-form.component.html
@@ -3,7 +3,7 @@
     *ngIf="{
       billingAddress: billingAddress$ | async,
       isSameAsDelivery: isSameAsDelivery$ | async
-    } as observables"
+    } as addressData"
   >
     <div class="form-group">
       <div class="form-check">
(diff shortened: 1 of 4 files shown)


integration-libs/opf/components/opf-checkout-billing-address-form/opf-checkout-billing-address-form.component.html
```diff
@@ -1,67 +1,61 @@
...
   <ng-container
-    *ngIf="
-      !isEditBillingAddress &&
-      isSameAsDelivery &&
-      (deliveryAddress$ | async) as deliveryAddress
-    "
+    *ngIf="{
+      billingAddress: billingAddress$ | async,
+      isSameAsDelivery: isSameAsDelivery$ | async
+    } as observables"
```
Matejk00: Naming this `as observables` does not sound good to me, because we don't have an access to observables but to their values.
What about a change to `data` or `addressData`?

Matejk00 (changes requested):

integration-libs/opf/components/opf-checkout-billing-address-form/opf-checkout-billing-address-form.component.html
rmch91: Thanks, changed

Matejk00 (approved):

PR #17441, opened 2023-05-23
feat: CXSPA-2251 Contextual logs in SSR

pawelfras:
closes CXSPA-2251


projects/core/src/i18n/i18next/i18next-plugins/i18next-logger-plugin.ts
```diff
@@ -0,0 +1,27 @@
...
+})
+export class I18nextLoggerPlugin {
+  logger = inject(LoggerService);
+
+  getPlugin(): LoggerModule {
+    return {
+      type: 'logger',
+      log: (args) => this.logger.log(...args),
+      warn: (args) => this.logger.warn(...args),
+      error: (args) => this.logger.error(...args),
```
Platonn: [MEDIUM]

Don't we need to deconstruct args with spread `...`?
```suggestion
      log: (...args) => this.logger.log(...args),
      warn: (...args) => this.logger.warn(...args),
      error: (...args) => this.logger.error(...args),
```

projects/core/src/i18n/i18next/i18next-initializer.ts
```diff
@@ -4,20 +4,23 @@
...
 import { I18nextBackendService } from './i18next-backend/i18next-backend.service';
 import { I18NEXT_INSTANCE } from './i18next-instance';
+import { I18nextLoggerPlugin } from './i18next-plugins/i18next-logger-plugin';
 
 /**
  * Initializes the i18next instance.
  */
 @Injectable({ providedIn: 'root' })
 export class I18nextInitializer implements OnDestroy {
+  protected loggerPlugin = inject(I18nextLoggerPlugin);
```
Platonn: [MEDIUM]

Instead of having an injectable plugin class (service) which returns a plugin via `getPlugin` method, what about having simply just a single injection token `I18NEXT_LOGGER_PLUGIN`:
```ts
/**
 * Logger for i18next. By default, it delegates all logs to the Spartacus LoggerService.
 */
export const I18NEXT_LOGGER_PLUGIN = new InjectionToken<LoggerModule>(
  'I18NEXT_LOGGER_PLUGIN',
  { providedIn: 'root', factory: () => {
    const logger = inject(LoggerService);
    return {
      type: 'logger',
      log: (...args) => logger.log(...args),
      warn: (...args) => logger.warn(...args),
      error: (...args) => logger.error(...args),
    
    }
  }
);
```

core-libs/setup/ssr/optimized-engine/ssr-optimization-options.ts
```diff
@@ -109,6 +110,12 @@ export interface SsrOptimizationOptions {
    * Enable detailed logs for troubleshooting problems
    */
   debug?: boolean;
+
+  /**
+   * Logger for troubleshooting problems
+   * related to the rendering process
+   */
```
Platonn: [MINOR]

I'm not sure we should mention about "troubleshooting" - it depends on how to logger is used.

What about:
```ts
/**
   * Config for improving logged messages with context and JSON structure.
   * 
   * It enhances the logs in SSR by adding context, including the request's details,
   * and structuring them as JSON. 
   * 
   * The `logger` property is optional and accepts two values:
   * - `true`:         Enables the default logger and enhances the logs.
   * - `ServerLogger`: Interprets the given `ServerLogger` as a custom logger
   * 
   * By default, the logger is disabled, meaning that logs in SSR are not enhanced.
   */
```

projects/core/src/config/config-validator/config-validator.ts
```diff
@@ -35,12 +36,15 @@ export function provideConfigValidator(
...
   config: Config,
-  configValidators: ConfigValidator[]
+  configValidators: ConfigValidator[],
+  logger?: LoggerService
 ) {
   for (const validate of configValidators) {
     const warning = validate(config);
     if (warning) {
-      console.warn(warning);
+      // remove before 7.0 release
```
Platonn: [MINOR]

```suggestion
      // remove console.log before 7.0 release and make `logger` required param
```

.eslintrc.json
```diff
@@ -93,7 +93,8 @@
               }
             ]
           }
-        ]
+        ],
+        "no-console": "error"
```
Platonn, line 97: [MEDIUM]

Is this rule configured for all projects? I guess it's not needed for all projects, for instance:
- schematics 
- tooling/scripts
- e2e

I believe having `no-console` rule enabled for all projects, would introduce a confusion for developers. They would need to use `/* eslint-disable no-console */` in places where it's not justified (in above listed types of projects). If we have this exception comment used _commonly_ in many places without explanation, then it might be confusing and look like a common/recommended practice to follow in new places.

PS. whenever we're bypassing `no-console` rule inside **libraries** or **storefrontapp,** it would be good to [explain in a code comment in each place](https://eslint.org/docs/latest/use/configure/rules#configuration-comment-descriptions), justifying why we do an exception in such a place

projects/core/src/logger/logger.service.ts
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { Injectable } from '@angular/core';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class LoggerService {
```
Platonn, line 18: [MINOR]

Let's add some JSDocs. What about:
```
By default it delegates logged messages to the native `console` object.

It's a good extension point for customizing the destination of logs
(e.g. to use a 3rd party logger library) or to enhance logs with more contextual data.
```

projects/core/src/state/utils/processes-loader/processes-loader.reducer.ts
```diff
@@ -40,6 +40,7 @@ export function processesLoaderReducer<T>(
         processesCountDiff &&
         state.processesCount + processesCountDiff < 0
       ) {
+        /* eslint-disable-next-line no-console */
         console.error(
```
Platonn, line 44: [MEDIUM]

This error can occur in runtime, so it would be good to also delegate this message to our common LoggerService. However IMHO it's not urgent for this PR (it can be postponed to another one - because of the low-probability of the error occuring and no low-hanging fruit solution).

FYI: It's an util console.error that indicates a software developer's mistake. 

I can understand we don't have access to an injetable logger service in reducer.

Quick idea of solution: we could move the logging logic from this reducer to a separate ngrx effect. Side note: I can see this reducer isn't a "pure function", as it should, because it not only outputs a new ngrx state, but also performs an I/O operation (logging - to which can go to console | file | http endpoint). So at the first sight an ngrx effect seems to be a more appropriate place for logging anyway. 

That said, IMHO it's not that critical for this PR, and this PR already brings a big added value for customers. So I don't mind if we postpone fixing this in another PR. Can we create a followup ticket for that? (that said, the fix shouldn't be that complex, the idea is straighforward, it just requires testing)

projects/core/src/util/normalize-http-error.ts
```diff
@@ -46,6 +46,7 @@ export function normalizeHttpError(
   }
 
   if (isDevMode()) {
+    /* eslint-disable-next-line no-console */
     console.error(
```
Platonn: [MEDIUM]

This error message can occur in runtime, so it should be delegated to our common LoggerService. Can we add a new `logger` optional param to `normalizeHttpError()`? (then we would need to also add a ticket to make this param required in a new major)

I'm aware `normalizeHttpError` is used 261 times in our repo 😕 
Hopefully a simple regex should do the job:
replace: `normalizeHttpError\((.*)\)`
with: `normalizeHttpError($1, logger)`
image

And then just add `protected logger = inject(LoggerService);` in all changed files.

projects/storefrontlib/cms-structure/page/page-layout/page-layout.service.ts
```diff
@@ -257,7 +260,7 @@ export class PageLayoutService implements OnDestroy {
     }
     if (page.template && !this.logSlots[page.template]) {
       // the info log is not printed in production
-      // eslint-disable-next-line no-console
+      /* eslint-disable-next-line no-console */
```
Platonn: [MEDIUM]

Let's delegate this runtime log to the common LoggerService as well. Because the original is `info()`, I can see 2 options:
a) change to `log()`
b) implement `info()` (and `debug()`) [as well](https://stackoverflow.com/a/25532848/11734692) in our logger layers and use `logger.info()` here

What do you think?

projects/schematics/.eslintrc.json
```diff
@@ -1,4 +1,12 @@
...
-  "ignorePatterns": ["**/*.d.ts"]
+  "ignorePatterns": ["**/*.d.ts"],
+  "overrides": [
+    {
+      "files": ["*.ts"],
+      "rules": {
+        "no-console": "off"
+      }
+    }
+  ]
```
Platonn, line 11: [QUESTION]

Why don't we need a similar config in:
- individual libraries' schematics
- e2e project folder
- tooling/scripts (e.g. `tools/schematics/testing.ts`)

?

projects/storefrontapp/server.ts
```diff
@@ -75,6 +76,7 @@ function run() {
   // Start up the Node server
   const server = app();
   server.listen(port, () => {
+    /* eslint-disable-next-line no-console */
```
Platonn: Let's add explanation, why we can use console.log in this case. What about?:
```suggestion
    /* eslint-disable-next-line no-console 
        --
        It's just an example application file. This message is not crucial
        to be logged using any special logger. Moreover, we don't have
        any special logger available in this context. */
```

projects/storefrontapp/src/main.ts
```diff
@@ -16,6 +16,7 @@ if (environment.production) {
 function bootstrap() {
   platformBrowserDynamic()
     .bootstrapModule(AppModule)
+    /* eslint-disable-next-line no-console */
```
Platonn: Let's add an explanation for the exception:
```suggestion
    /* eslint-disable-next-line no-console 
        --
        It's just an example application file. This message is not crucial
        to be logged using any special logger. Moreover, we don't have
        any special logger available in this context. */
```

core-libs/setup/ssr/error-handlers/ssr-error-handler.ts
```diff
@@ -0,0 +1,26 @@
...
+export const ssrErrorHandlerFactory = () => {
+  const isLoggerEnabled = inject(loggerEnabled);
+
+  return isLoggerEnabled ? new SsrErrorHandler() : new ErrorHandler();
+};
+
+@Injectable({
+  providedIn: 'root',
+})
+export class SsrErrorHandler implements ErrorHandler {
```
Platonn: [MEDIUM]

This class has nothing to do with SSR (but only to Logger which is a Core feature), can we move it to `@spartacus/core` ,e.g. to new folder `/error` ?
It can be e.g. `CxErrorHandler`

core-libs/setup/ssr/logger/loggers/default-server-logger.ts
```diff
@@ -0,0 +1,32 @@
...
+  protected createLogMessage(message: string, context: LogContext | undefined) {
+    return {
+      message,
+      context: {
+        timestamp: new Date().toISOString(),
+        ...context,
+      },
+    };
+  }
+}
```
Platonn: [MEDIUM]

I guess we're lacking here the JSON.stringify() and a return type `string` - comparing to `ExpressServerLogger`?

core-libs/setup/ssr/logger/loggers/express-server-logger.spec.ts
```diff
@@ -0,0 +1,85 @@
+import { ExpressServerLogger } from './express-server-logger';
+import { ServerLogger } from './server-logger';
+
+describe('DefaultExpressSsrLogger', () => {
```
Platonn: [MINOR]

ExpressServerLogger vs DefaultExpressSsrLogger?

core-libs/setup/ssr/logger/loggers/default-server-logger.ts
```diff
@@ -0,0 +1,32 @@
+/*
+ * SPDX-FileCopyrightText: 2023 SAP Spartacus team <spartacus-team@sap.com>
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { LogContext, ServerLogger } from './server-logger';
+
+export class DefaultServerLogger extends ServerLogger {
```
Platonn: [MEDIUM]

What about name `PrerenderingServerLogger`? As far as I understand this logger is used only for prerendering? And we already have `PrerenderingLoggerService`.
And also we have a pair for: `ExpressLoggerService` and `ExpressServerLogger`.

The word "Default" might also be a bit misleading in this context. Because, _in practice_ in most cases we will use the express

core-libs/setup/ssr/optimized-engine/optimized-ssr-engine.ts
```diff
@@ -405,29 +440,54 @@ export class OptimizedSsrEngine {
...
 
       renderCallback(err, html);
     });
   }
+
+  private initLogger(ssrOptions: SsrOptimizationOptions | undefined) {
+    if (typeof ssrOptions?.logger === 'boolean') {
+      return new ExpressServerLogger();
+    }
+    return ssrOptions?.logger ?? new ServerLogger();
```
Platonn: [MEDIUM]

I'm aware that the interface doesn't allow for `false`, but in the future we might add it (since next major). So let's handle `false` value simply here, to be future proof - when it's false, use the base ServerLogger.

What about:
```suggestion
    if (ssrOptions?.logger === true) {
      return new ExpressServerLogger();
    }
    return ssrOptions?.logger || new ServerLogger();
```
let's also add an unit test for false

core-libs/setup/ssr/logger/loggers/server-logger.ts
```diff
@@ -0,0 +1,34 @@
...
+    /* eslint-disable-next-line no-console */
+    console.error(message);
+  }
+}
+
+export const serverLoggerToken = new InjectionToken<ServerLogger>(
+  'ServerLogger'
+);
+
+export const loggerEnabled = new InjectionToken<boolean>('featureFlag.logger');
```
Platonn: [MEDIUM]

Do I understand correctly, that this flag controls only the behavior of the logger in Server (but not CSR)? To be precise, does this flag enable **adding context** to the logs in the server?

If yes, then `loggerEnabled` name might be a bit misleading.
What about `ENABLE_CONTEXTUAL_SERVER_LOGGER`?

PS I suggest to stick to the convention that injection tokens are in CONSTANT_CASE format.
PS2 Let's add JSDocs. (why it's needed. and that in next major it will be removed)

core-libs/setup/ssr/logger/loggers/express-server-logger.ts
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    return JSON.stringify(object);
+  }
+
+  protected mapRequest(request: Request): Record<string, any> {
+    return {
+      url: request.originalUrl,
+      render: request.res?.locals.cx.render,
+      headers: request.headers,
```
Platonn: TODO reminder: let's remove `headers` before merging it

core-libs/setup/ssr/logger/loggers/server-logger.ts
```diff
@@ -0,0 +1,34 @@
...
+  }
+  error(message: string, _context?: LogContext): void {
+    /* eslint-disable-next-line no-console */
+    console.error(message);
+  }
+}
+
+export const serverLoggerToken = new InjectionToken<ServerLogger>(
+  'ServerLogger'
+);
```
Platonn: [MEDIUM]

1. Let's stick to convention of CONSTANT_CASE for injection tokens.
2. What about renaming to `SERVER_LOGGER`?
```suggestion
export const SERVER_LOGGER = new InjectionToken<ServerLogger>(
  'SERVER_LOGGER'
);
```
3. Let's add JSDocs

core-libs/setup/ssr/logger/services/express-logger.service.ts
```diff
@@ -0,0 +1,42 @@
...
+import { Injectable, inject } from '@angular/core';
+import { REQUEST } from '@nguniversal/express-engine/tokens';
+import { LoggerService } from '@spartacus/core';
+import { formatWithOptions } from 'util';
+import { serverLoggerToken } from '../loggers/server-logger';
+
+@Injectable({ providedIn: 'root' })
+export class ExpressLoggerService implements LoggerService {
+  request = inject(REQUEST);
+  logger = inject(serverLoggerToken);
```
Platonn: [MEDIUM]

To avoid ambiguity and for consistency, I suggest to rename it to `serverLogger = ...`

core-libs/setup/ssr/logger/services/prerendering-logger.service.ts
```diff
@@ -0,0 +1,32 @@
...
+ */
+
+import { Injectable, inject } from '@angular/core';
+import { LoggerService, WindowRef } from '@spartacus/core';
+import { format } from 'util';
+import { serverLoggerToken } from '../loggers/server-logger';
+
+@Injectable({ providedIn: 'root' })
+export class PrerenderingLoggerService implements LoggerService {
+  ssrLogger = inject(serverLoggerToken);
```
Platonn: [MEDIUM]

To avoid ambiguity and for consistency, I suggest to rename it to `serverLogger = ...`, here too

core-libs/setup/ssr/optimized-engine/optimized-ssr-engine.ts
```diff
@@ -295,9 +329,9 @@ export class OptimizedSsrEngine {
     });
   }
 
-  protected log(message: string, debug = true): void {
-    if (!debug || this.ssrOptions?.debug) {
-      console.log(message);
+  protected log(message: string, logMetadata?: LogContext, debug = true): void {
```
Platonn: [MEDIUM]

Formally, we made a breaking change here - changed type of the 2nd argument. 
To fix it, we would need to add new param as the 3rd one.

Side note: I'm not fond of the current signature of this method, (especially with debug: boolean), so in the future we might refactor it anyway.

core-libs/setup/ssr/error-handlers/ssr-error-handler.ts
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+import { ErrorHandler, Injectable, inject } from '@angular/core';
+import { LoggerService } from '@spartacus/core';
+import { loggerEnabled } from '../logger';
+
+export const ssrErrorHandlerFactory = () => {
+  const isLoggerEnabled = inject(loggerEnabled);
+
+  return isLoggerEnabled ? new SsrErrorHandler() : new ErrorHandler();
```
Platonn: [QUESTION] 

Why do we do `new SsrErrorHandler` instead of `inject(SsrErrorhandler)`?

core-libs/setup/ssr/logger/services/prerendering-logger.service.ts
```diff
@@ -0,0 +1,32 @@
...
+      request: { url: this.windowRef.location.href },
+    });
+  }
+  warn(...args: Parameters<typeof console.warn>) {
+    this.ssrLogger.warn(format(...args), {
+      request: { url: this.windowRef.location.href },
+    });
+  }
+  error(...args: Parameters<typeof console.error>) {
+    this.ssrLogger.error(format(...args), {
```
Platonn: [QUESTION]

Why didn't we use `formatWithOptions( breakLength:Infinity` here?

projects/core/src/i18n/i18next/i18next-plugins/i18next-logger-plugin.ts
pawelfras: ~~Good point. In fact, it doesn't change the way how args are logged, but it's good to keep consistency.~~
Actually, I'll keep it as it is. When I'm deconstructing args in the arrow function, the log message contains an array instead of a simple string, e.g. `"message":"[ 'i18next::backendConnector: No backend was added via i18next.use. Will not load resources.' ]"`

projects/core/src/i18n/i18next/i18next-initializer.ts
pawelfras: Thanks! I really like this idea of decreasing complexity (and + looks modern 😄)

core-libs/setup/ssr/error-handlers/ssr-error-handler.ts
pawelfras: Thanks, good catch. It was made by mistake.

core-libs/setup/ssr/optimized-engine/optimized-ssr-engine.ts
pawelfras: Do we consider this a breaking change? correct me if I'm wrong, but AFAIK `OptimizedSsrEngine` is not a part of the public API, so I thought it doesn't change that much.

core-libs/setup/ssr/logger/loggers/express-server-logger.ts
pawelfras: And replace with info from `traceparent` 🔥 😎

projects/storefrontlib/cms-structure/page/page-layout/page-layout.service.ts
pawelfras: Ok, I like the idea of implementing `info()` and `debug()` methods.

projects/core/src/util/normalize-http-error.ts
pawelfras: > And then just add protected logger = inject(LoggerService); in all changed files.

Yeah, just add... 😄 
But I got the point. It's necessary.

projects/schematics/.eslintrc.json
pawelfras: Indeed, we should have such a flag (or even `.eslintrc.json` file if it's missing) for mentioned cases. BUT for some reasons, lack of this config is not causing any building or linting errors...

core-libs/setup/ssr/logger/loggers/server-logger.ts
```diff
@@ -0,0 +1,34 @@
...
+  }
+  warn(message: string, _context?: LogContext): void {
+    /* eslint-disable-next-line no-console */
+    console.warn(message);
+  }
+  error(message: string, _context?: LogContext): void {
+    /* eslint-disable-next-line no-console */
+    console.error(message);
+  }
+}
```
Platonn: [MAJOR]

1. It's strange that this logger ignores the second param. I'm afraid we've mixed 2 concepts in 1 class:
- we need a standardized interface for all types of server loggers -> I suggest to create **an interface** named `ServerLogger`
- we need temporarily (until the next major) a backward-compatible solution, which simulates the old behavior - just print string and no context -> I suggest to rename this class to `LegacyServerLogger` and mark it in JSDocs as @depreacated and plan for removing in the next major.

2. For the generic interface `ServerLogger`, the 2nd argument with context IMHO should be just `Record<string,any>`. The generic `ServerLogger` is just a logger without any notion of requests or options or anything else. 
However I agree, that for `ExpressServerLogger` or `PrerenderingServerLogger` we could create specialized interfaces for context params, e.g. `ExpressServerLoggerContext` (with required `request: Request` from express) and `PrerenderingServerLoggerContext` (with required `request: {url:string, host: string }`). And then `ExpressLoggerService` and `PrerenderingLoggerService` should also make sure they comply with those interfaces then passing data to a specific `ServerLogger`

What do you think about it?

core-libs/setup/ssr/logger/loggers/express-server-logger.ts
```diff
@@ -0,0 +1,49 @@
+/*
+ * SPDX-FileCopyrightText: 2023 SAP Spartacus team <spartacus-team@sap.com>
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { Request } from 'express';
+import { LogContext, ServerLogger } from './server-logger';
+
+export class ExpressServerLogger extends ServerLogger {
```
Platonn: [MINOR]
Let's add JSDocs (what's the purpose of this logger). e.g.:

```
Default logger used in SSR (ExpressJS). It outputs a JSON with properties "message" and "context",
which contains a "timestamp" and details of the "request" ("url", "uuid", "timeReceived")
```


`... and "openTracing")` 👈 in this or a different PR


PS. Analogically for PrerenderingServerLogger

core-libs/setup/ssr/logger/loggers/express-server-logger.ts
```diff
@@ -0,0 +1,49 @@
...
+        request: this.mapRequest(context.request),
+      });
+    }
+
+    return JSON.stringify(object);
+  }
+
+  protected mapRequest(request: Request): Record<string, any> {
+    return {
+      url: request.originalUrl,
```
Platonn, line 58: [MINOR, can be postponed]

Should we return only the URL path? Or also URL host (domain)? Or Both? IMHO both.

Why asking: because some customers might run a single NodeJS Spartacus instance to render multi-site (multiple basesites) hosted "virtually" in different domains. If both domains (e.g. `site-a.com` and `site-b.com`) have the same path URL (e.g. homepage '/' 😅), then those are 2 very different renderings, but they would appear with the same URL in kibana '/').

Note: we already have a function that we can use `getRequestOrigin(request)` - [source code](https://github.com/SAP/spartacus/blob/f3650ed2c6d05294f99012d8360ee7edbf4a0e8a/core-libs/setup/ssr/express-utils/express-request-origin.ts#L9-L26) - it returns the correct origin (domain), even if your NodeJS is behind a (chain of) reverse proxy.

```suggestion
      url: request.originalUrl,
      host: getRequestOrigin(request)
```

However I can see a downside of the above solution - it computes the origin from scratch for every logged message (instead of once per request).

Quick idea: pre-calculate the correct host once and then put it to `response.locals`. And then read it each time we log a message. 

We can postpone it to a future ticket. Would you mind creating an improvement ticket for it?

core-libs/setup/ssr/logger/loggers/server-logger.factory.ts
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { inject } from '@angular/core';
+import { DefaultServerLogger } from './default-server-logger';
+import { serverLoggerToken } from './server-logger';
+
+/**
+ * Factory provides logger that's used for pre-rendering purposes.
```
Platonn: [MINOR]

What about:

```suggestion
 * Returns a `ServerLogger` instance
 * 
 * If any specific `ServerLogger` implementation has been already provided
 * on the platform injector level (e.g. in ExpressJS app), we use it.
 * Otherwise we fallback to the `PrerenderigServerLogger`.
```

core-libs/setup/ssr/logger/services/server-logger-factory.ts
```diff
@@ -0,0 +1,17 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { inject } from '@angular/core';
+import { REQUEST } from '@nguniversal/express-engine/tokens';
+import { ExpressLoggerService } from './express-logger.service';
+import { PrerenderingLoggerService } from './prerendering-logger.service';
+
+export const serverLoggerServiceFactory = () => {
```
Platonn, line 12: [MINOR]

Let's add JSDocs. e.g.

```
Returns an applicable `LoggerService` implementation for the server environment,
different for SSR (ExpressJS) vs. for Prerendering.

File: core-libs/setup/ssr/optimized-engine/optimized-ssr-engine.spec.ts [hunk 31]
```diff
@@ -1147,4 +1188,68 @@ describe('OptimizedSsrEngine', () => {
...
+            "logger": "MockServerLogger",
+            "maxRenderTime": 300000,
+            "reuseCurrentRendering": true,
+            "timeout": 3000,
+          },
+        },
+      ]
+      `);
+    });
+    it('should use the default logger', () => {
```
Platonn: [MINOR]

```suggestion
    it('should use the legacy server logger, if logger option not specified', () => {
```

core-libs/setup/ssr/optimized-engine/optimized-ssr-engine.spec.ts
```diff
@@ -111,6 +124,17 @@ const getCurrentConcurrency = (
...
+  engineRunner: TestEngineRunner,
+  url: string
+): Request | undefined => {
+  const idx = engineRunner.requests.findIndex((request) => request[url]);
+  const request = engineRunner.requests[idx][url];
+  engineRunner.requests.splice(idx, 1);
+
+  return request;
+};
+
```
Platonn: [MEDIUM]

I'm not fond of adding more complexity to the `TestEngineRunner`, unless it's _really_ justified.

What do you think about:
- reverting changes to the `TestEngineRunner`
- change expects to simply check the `originalUrl` in the `request`? e.g,: 
`.toHaveBeenCalledWith(..., { request: expect.objectContaining({ originalUrl: <URL HERE> }) }`

core-libs/setup/ssr/optimized-engine/optimized-ssr-engine.ts
Platonn: Unfortunately it is in the public API . Personally, I wish it was in private API, as I think it's internals are worth refactoring due to various reasons.


image

core-libs/setup/ssr/providers/ssr-providers.ts
```diff
@@ -26,9 +33,21 @@ export function provideServer(options?: ServerOptions): StaticProvider[] {
       provide: SERVER_REQUEST_URL,
       useFactory: serverRequestUrlFactory(options),
     },
+    {
+      provide: ErrorHandler,
+      useFactory: ssrErrorHandlerFactory,
+    },
```
Platonn: [MEDIUM]

When we move CxErrorHandler to the `@spartacus/core` lib, let's move also this provider to the core ErrorModule

PS. `/error` vs. `/error-handling`?  `ErrorModule` vs `ErrorHandlingModule` ?
I fell I like the phrase "error handling" a bit more. Which do you prefer?

projects/core/src/i18n/translate.pipe.spec.ts
```diff
@@ -1,25 +1,36 @@
...
+    });
+    pipe = TestBed.inject(TranslatePipe);
+    service = TestBed.inject(TranslationService);
+    cd = TestBed.inject(ChangeDetectorRef);
   });
 
   describe('transform', () => {
     it('should return falsy when input is undefined', () => {
-      const result = pipe.transform(undefined);
+      const result = pipe.transform('');
```
Platonn: [QUESTION]

Are changes in this file _really_ related to this PR?

btw. why we changed `undefined` to `''`? The description says `should return falsy when input is undefined`

core-libs/setup/ssr/logger/loggers/express-server-logger.ts
Platonn: Leaving link just for reference, to split `traceparent`: https://www.w3.org/TR/trace-context/#traceparent-header

core-libs/setup/ssr/optimized-engine/optimized-ssr-engine.ts
```diff
@@ -233,11 +259,16 @@ export class OptimizedSsrEngine {
     options: any,
     callback: SsrCallbackFn
   ): void {
+    const render = { uuid: randomUUID(), timeStart: new Date().toISOString() };
+    options.req.res.locals = { cx: { render } };
```
Platonn: [MEDIUM]

As we discussed offline, those values (uuid and timeStart) are bound to a request, but not a render.
What about renaming `render` to `request`? and `timeStart` to `timeReceived`?

projects/schematics/.eslintrc.json
pawelfras: I've checked why such a config is not necessary for mentioned places like e2e project or tooling/scripts and the reason is they are not treated as an Nx project (doesn't contain project.json file), so while `"lint": "nx run-many --all --target=lint"` those projects are not taken into account.
When it comes to individual libraries' schematics - I believe we don't want to contain logs there (at least merged). It's still possible to use `console.log` for testing purposes or as a part of a snapshot.

projects/core/src/state/utils/processes-loader/processes-loader.reducer.ts
pawelfras: I agree, I'll prepare a followup ticket for this topic.

pawelfras: Link to the Jira ticket: https://jira.tools.sap/browse/CXSPA-3670

.eslintrc.json
pawelfras: https://github.com/SAP/spartacus/pull/17441#discussion_r1226240253

core-libs/setup/ssr/optimized-engine/optimized-ssr-engine.ts
pawelfras: Understood. I'll change the order of arguments 👌

projects/core/src/i18n/translate.pipe.spec.ts
pawelfras: No, this particular change is not related to this PR. I'll redo it.
Btw, I would change `undefined` to `''` and the test description too, because the definition of this function does not allow to provide `undefined`:
```
transform(
    input: Translatable | string,
    options: TranslatableParams = {}
  )
```

The rest of the changes are necessary because previously TranslatePipe wasn't tested as injectable what was causing the issue with `inject(LoggerService)`

core-libs/setup/ssr/logger/loggers/express-server-logger.ts
pawelfras: This feature won't be a part of this PR and will be implemented as a separate task: https://jira.tools.sap/browse/CXSPA-2067

core-libs/setup/ssr/providers/ssr-providers.ts
pawelfras: I prefer the second option as it provides more context, it's in line with the naming convention known from Angular.

core-libs/setup/ssr/logger/loggers/express-server-logger.ts
pawelfras: sure, the link is here: https://jira.tools.sap/browse/CXSPA-3679

core-libs/setup/ssr/optimized-engine/optimized-ssr-engine.spec.ts
pawelfras: Thank you for the advice. Change has been applied.

projects/core/src/i18n/i18next/i18next-plugins/i18next-logger-plugin.ts
Platonn: Just for reference:
we continued this discussion in a new PR: https://github.com/SAP/spartacus/pull/17512#discussion_r1227791790

Platonn:
closing in favor of https://github.com/SAP/spartacus/pull/17512

PR #17448, opened 2023-05-25
Update cdc-register-component.service.spec.ts

anjana-bl:


hackergil (approved):

integration-libs/cdc/user-profile/register/cdc-register-component.service.spec.ts
```diff
@@ -167,18 +167,23 @@ describe('CdcRegisterComponentService', () => {
...
+            titleCode: 'Mr.',
+            firstName: 'firstName',
+            lastName: 'lastName',
+            uid: 'uid',
+            password: 'password',
+          });
+        },
       });
       expect(cdcJsService.didLoad).toHaveBeenCalled();
       done();
```
Platonn, line 189: I believe this done() callback should be inside of the body of the subscription

In most cases, the Jasmine's `done` function should be put inside the body of the observable's `subscribe()` in the test, when this body contains any `expect()`.  Otherwise, Jasmine won't wait for the invocation of done. So any expect() checks placed in async operations near the done() might not be checked at all.

PR #17449, opened 2023-05-25
feat: handle active configurations empty array 

Matejk00:
Closes: CXSPA-3172

### DIFF
--- a/integration-libs/opf/components/opf-checkout-payments/opf-checkout-payments.component.ts
+++ b/integration-libs/opf/components/opf-checkout-payments/opf-checkout-payments.component.ts
@@ -38,10 +38,10 @@ export class OpfCheckoutPaymentsComponent implements OnInit, OnDestroy {
     .getActiveConfigurationsState()
     .pipe(
       tap((state: QueryState<ActiveConfiguration[] | undefined>) => {
-        if (state?.error) {
+        if (state.error) {
           this.displayError('loadActiveConfigurations');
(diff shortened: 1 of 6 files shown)


integration-libs/opf/components/opf-checkout-payments/opf-checkout-payments.component.ts
```diff
@@ -33,12 +37,13 @@ export class OpfCheckoutPaymentsComponent implements OnInit, OnDestroy {
...
-        if (state?.error) {
-          this.globalMessageService.add(
-            { key: 'opf.checkout.errors.loadActiveConfigurations' },
-            GlobalMessageType.MSG_TYPE_ERROR
-          );
+      tap((state: QueryState<ActiveConfiguration[] | undefined>) => {
+        if (state.error) {
+          this.displayError('loadActiveConfigurations');
+        } else {
+          if (!state.loading && !Boolean(state.data?.length)) {
```
rmch91: Why not to use else if instead of placing nesting if statement inside of else block?

rmch91 (changes requested):

integration-libs/opf/components/opf-checkout-payments/opf-checkout-payments.component.ts
Matejk00: done, thanks

rmch91 (approved):

PR #17459, opened 2023-05-29
feat: ASM Deep Links

Zeyber:
Closes: https://jira.tools.sap/browse/CXSPA-158


scarai-sap:
@Zeyber , Maybe the target should be 6.2?

Zeyber:
> @Zeyber , Maybe the target should be 6.2?

Yes. We need to switch this to target `develop-6.2.x`.

scarai-sap (approved):

PR #17471, opened 2023-05-31
chore: Release 6.1.0-1

hackergil:


package.json
```diff
@@ -210,7 +210,7 @@
     "ts-morph": "^9.1.0",
     "ts-node": "^10.6.0",
     "typescript": "^4.8.2",
-    "webpack": "^5.76.1",
+    "webpack": "~5.76.1",
```
plabadie: Is that intentional?

hackergil: It is to temporary fix an error with node 14 on CCv2

plabadie (approved):

plabadie (approved):

PR #17482, opened 2023-06-02
fix: Login page displayed for few seconds when payment canceled/failed

kpawelczak:
The other solution to handle this would be to just provide the default option for `renderingStrategyResolver`, which I think should work better than adding new options to the existing interface.

BREAKING CHANGE:

SsrOptimizationOptions has a new option.

Closes: https://jira.tools.sap/browse/CXSPA-3492


kpawelczak:
closing in favor of https://github.com/SAP/spartacus/pull/17801

PR #17485, opened 2023-06-05
styling: Style changes for mobile and desktop checkout UI

morganm58:
fixes CXSPA-3593

https://jira.tools.sap/browse/CXSPA-3593


kimhw0630:
The code seems fine to me, but when I test the UI, I notice some missing changes that deviate from the UX design.
(tested in Mac with Chrome (Version 113.0.5672.126)

need an underline beneath the 'ORDER SUMMARY' label.
image

CVV box is off
image

"Add New Address" button should be stacked together
image

background looks weird.. gray color??
image

Delivery days should be next line with the green color?
image

should be separated boxes with gray border and white background?
image

morganm58:
The only change that can be done is the ccv/cvc input label and the background color on mobile payment form. Everything else are breaking changes. So we can't make those changes until the next major release.

projects/assets/src/translations/en/payment.ts
```diff
@@ -15,7 +15,7 @@ export const payment = {
     },
     cardNumber: 'Card Number',
     expirationDate: 'Expiration Date',
-    securityCode: 'Security code (CVV)',
+    securityCode: 'CVV/CVC',
```
kimhw0630: Changing or removing translation keys or values is breaking change
see https://sap.github.io/spartacus-docs/breaking-changes/#avoiding-breaking-changes

kimhw0630 (dismissed):

kimhw0630:
I have noticed a minor issue in the mobile view where the boxes are not aligned properly. 
("Billing address" and "Payment" boxes should be aligned with "Ship to" box) 

image

kimhw0630 (dismissed):

PR #17486, opened 2023-06-06
fix: order confirmation page for guest user

kpawelczak:


kimhw0630:
I don't think this is the correct solution because it simply hides the feature that creates a user account for the guest user. Is there a way to obtain a guest email address?
CC: @Matejk00 , @FollowTheFlo

kpawelczak:
> I don't think this is the correct solution because it simply hides the feature that creates a user account for the guest user. Is there a way to obtain a guest email address? CC: @Matejk00 , @FollowTheFlo

whether we can obtain the email address from somewhere else or not, we always have to assume that the backend data can be incomplete. So we have to make sure that the application won't break in that case.

kpawelczak:
closing in favor of https://github.com/SAP/spartacus/pull/17735

PR #17492, opened 2023-06-07
refactor: Cleanup services for deep linking

Zeyber:


feature-libs/asm/components/asm-bind-cart/asm-bind-cart.component.ts
```diff
@@ -313,39 +314,24 @@ export class AsmBindCartComponent implements OnInit, OnDestroy {
             )
           )
           .subscribe(() => {
-            if (this.isDeepLinkInactiveCart()) {
-              this.displayBindCartBtn$.next(false);
+            const cartType =
+              this.asmComponentService?.getSearchParameter('cartType');
+            if (cartType === 'inactive' || cartType === 'active') {
+              this.displayBindCartBtn$.next(cartType === 'inactive');
```
i333055:
```suggestion
              this.displayBindCartBtn$.next(false);
```

feature-libs/asm/components/asm-bind-cart/asm-bind-cart.component.ts
```diff
@@ -313,39 +314,24 @@ export class AsmBindCartComponent implements OnInit, OnDestroy {
...
               this.displaySaveCartBtn$.next(true);
```
i333055, line 338:
```suggestion
              this.displaySaveCartBtn$.next(cartType === 'inactive');
```

feature-libs/asm/root/services/asm-deep-link.service.ts
```diff
@@ -0,0 +1,108 @@
...
+   * eg. customerId, customerId + ticketId return true.
+   */
+  validParameters(parameters: AsmDeepLinkParameters): boolean {
+    const keys = Object.keys(
+      Object.fromEntries(Object.entries(parameters).filter(([_, v]) => !!v))
+    );
+
+    const ALLOWED_PARAMS = [
+      ['customerId'],
+      ['customerId', 'cartType'],
```
scarai-sap: we don't support cartType without cartId,

We need support ticketid + orderId/ ticketId + cartId + cartType

feature-libs/asm/root/services/asm-deep-link.service.ts
```diff
@@ -0,0 +1,108 @@
...
+        });
+      }
+    }
+  }
+
+  /**
+   * Returns true if pairs of valid navigation params are in the url only.
+   * eg. customerId, customerId + ticketId return true.
+   */
+  validParameters(parameters: AsmDeepLinkParameters): boolean {
```
scarai-sap: why we need here, as in caller "handleNavigation" has enough check as if.. elif.. elif

It will block if other apps, consider price domain want to show mass prices with property with a flag "displayMassPrices", a url like below maybe composed, we should support navigate

cartId=123&displayMassPrices=true

feature-libs/asm/root/services/asm-deep-link.service.ts
Zeyber: What is the expected behaviour happen when `ticketid + orderId` or `ticketId + cartId + cartType ` are in the url? Isn't having both invalid?

feature-libs/asm/root/services/asm-deep-link.service.ts
Zeyber: Good point. We shouldn't have additional parameters preventing a navigation. However, the `getParamsInUrl()` method only checks for `AsmDeepLinkParameters` deep link params specifically and additional parameters will not be compared when using this method (ie. has no effect on navigation).

i333055 (approved):

kpawelczak (approved):

PR #17494, opened 2023-06-07
fix: sonar findings (CXSPA-3575)

Uli-Tiger:
* rename local var escape -> regexEscape to not override escape function
* ternary op has lower precendence than  'bigger than' op, hence grouping is required. Still outcome is same.


ChristophHi (dismissed):

giancorderoortiz (approved):

giancorderoortiz:
Approving & merging for historical reasons. 
Fix already in 6.2  https://github.com/SAP/spartacus/pull/17495

PR #17497, opened 2023-06-08
Feature/cxint 2019 - cdc b2c consent management

anjana-bl:

## General

- [ ] Epic has sufficient end to end (cypress) test coverage
- [ ] Schematics code deprecations have been provided and tested
- [ ] Epic CSS supports directionality (if required)
- [x] Epic renders successfully in SSR

## Audits/reviews

- [x] Architecture review of the epic
- [ ] Security and threat modeling audit/review from the Security team
- [ ] Accessibility audit/review from the UX team

## New Library

## Sample data

- [ ] Sample data has been added to the sampledata repo and merged to the next version to be released
- [ ] New sample data has been loaded and tested on a dev server
- [ ] New sample data has been loaded and tested in our CI server

## Documentation

- [x] Documentation input is provided for it to be published on the doc site.


feature-libs/user/profile/components/register/register-form.service.ts
vigneshkrish8691, line 1: I believe the form/service should be made consent-specific. Since you are dealing with consents. "register-consent.services.ts"

feature-libs/user/profile/components/register/register-form.service.ts
```diff
@@ -0,0 +1,25 @@
...
+
+import { Injectable } from '@angular/core';
+import { UntypedFormArray, UntypedFormBuilder } from '@angular/forms';
+import { ConsentTemplate } from '@spartacus/core';
+
+@Injectable()
+export class RegisterFormService {
+  constructor(protected fb: UntypedFormBuilder) {}
+
+  loadExtraRegistrationConsents(): {
```
vigneshkrish8691: Make the name of the method as 'loadAdditionalConsents'

feature-libs/user/profile/components/register/register.component.ts
```diff
@@ -70,14 +73,29 @@ export class RegisterComponent implements OnInit, OnDestroy {
     }
   );
 
+  registrationConsents: {
+    template: ConsentTemplate;
+    required: boolean;
+  }[];
+
+  get extraConsents(): UntypedFormArray {
```
vigneshkrish8691: Dont use "Extra"

feature-libs/user/profile/components/register/register-form.service.ts
anjana-bl: done

feature-libs/user/profile/components/register/register-form.service.ts
anjana-bl: done

projects/storefrontlib/cms-components/myaccount/consent-management/components/consent-form/consent-management-form.component.spec.ts
```diff
@@ -27,7 +27,7 @@ describe('ConsentManagementFormComponent', () => {
     component = fixture.componentInstance;
     el = fixture.debugElement;
 
-    fixture.detectChanges();
+    //fixture.detectChanges();
```
Platonn: Can we undo the commenting-out of the line?

feature-libs/user/profile/components/register/register.component.ts
anjana-bl: done

feature-libs/user/profile/components/register/register-form.service.ts
Platonn: Just asking: why do we even bother with the word "additional"? Can't be simply "loadConsents"? (and analogically `generateConsentsFormControl`)?

Platonn (commented):
just sending quick comments on the first spot.
I've made some notes aside, I need to think them through and then I'll probably add some other few comments.

integration-libs/cdc/root/consent-management/cdc-user-consent.adapter.ts
```diff
@@ -0,0 +1,86 @@
...
+      return super.giveConsent(
+        userId,
+        consentTemplateId,
+        consentTemplateVersion
+      );
+    } else {
+      return this.cdcUserConsentService
+        .updateCdcConsent(true, consentTemplateId)
+        .pipe(
+          catchError((error: any) => throwError(error)),
```
Platonn, line 50: nitpick: Is this line really needed?

(it applies to also other places with simply  "catchError->throwError")

projects/core/src/user/facade/user-consent.service.ts
```diff
@@ -236,12 +236,13 @@ export class UserConsentService {
...
    * @param consentCode for which to withdraw the consent
    */
-  withdrawConsent(consentCode: string): void {
+  withdrawConsent(consentCode: string, consentId?: string): void {
     this.userIdService.takeUserId().subscribe((userId) => {
       this.store.dispatch(
         new UserActions.WithdrawUserConsent({
           userId,
           consentCode,
+          consentId,
```
Platonn, line 245: [importance: high]

Is the concept of `consentId` relevant for CX Commerce? Or it's only relevant for CDC?

a) If it's common to CDC and CX Commerce, I'm fine with adding everywhere this new optional parameter. 

b) If only to CDC, it looks strange we leak this information in the core service (facade), ngrx action, connector and adapter.
If the `consentId` is relevant only to CDC (and not CX Commerce), what do you think about just adding a generic argument `params?: Record<string,any>` as an extensibility point (instead of concrete `consentId:string`) in the core clases/methods?
And in your custom CDC-flavoured classes (that extend from the original ones), we would pass those params along the layers of service, ngrx, connector and adapter?

integration-libs/cdc/root/consent-management/services/cdc-consent-management.service.ts
```diff
@@ -0,0 +1,49 @@
...
+})
+export class CdcConsentManagementService extends ConsentManagementService {
+  constructor(protected store: CdcConsentsLocalStorageService) {
+    super();
+  }
+  getRequiredConsents(templateList: ConsentTemplate[]): string[] {
+    let requiredConsents: string[] = [];
+    const cdcConsents = this.getCdcConsentIDs(true);
+    requiredConsents = super.getRequiredConsents(templateList);
+    requiredConsents.push(...cdcConsents);
```
Platonn: Let's not mutate the original array coming from the service (which I understand is the array taken from the static anonymousConsentsConfig -> in other words, let's not mutate accidentaly the static config).
```suggestion
    requiredConsents.push(...super.getRequiredConsents(templateList));
    requiredConsents.push(...cdcConsents);
```

projects/storefrontlib/cms-components/myaccount/consent-management/consent-management.service.ts
```diff
@@ -0,0 +1,18 @@
...
+ * SPDX-FileCopyrightText: 2023 SAP Spartacus team <spartacus-team@sap.com>
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { Injectable } from '@angular/core';
+import { AnonymousConsentsConfig, ConsentTemplate } from '@spartacus/core';
+
+@Injectable()
+export class ConsentManagementService {
```
Platonn: Just asking: This component is an util specific for the `ConsentManagementComponent`. Did you consider naming it `ConsentManagementComponentService`?

projects/storefrontlib/cms-components/myaccount/consent-management/consent-management.service.ts
```diff
@@ -0,0 +1,18 @@
...
+  constructor(protected anonymousConsentsConfig?: AnonymousConsentsConfig) {}
+  getRequiredConsents(_templateList: ConsentTemplate[]): string[] {
```
Platonn: We're not using the argument in the service in the core lib, which looks strange.
It it's inevitable, at least let's add a JSDoc comment (or ordinary code comment), explaining why we don't need to use the argument in the default implementation, but it's needed in CDC implementation

projects/storefrontlib/cms-components/myaccount/consent-management/components/consent-management.component.ts
```diff
@@ -95,11 +97,10 @@ export class ConsentManagementComponent implements OnInit, OnDestroy {
         }
       }),
       map(([templateList, anonymousTemplates]) => {
+        this.requiredConsents = this?.consentManagementService
```
Platonn: nitpick: `this` is defined

```suggestion
        this.requiredConsents = this.consentManagementService
```

projects/storefrontlib/cms-components/myaccount/consent-management/components/consent-form/consent-management-form.component.html
```diff
@@ -5,12 +5,15 @@
...
       [checked]="consentGiven"
+      [name]="consentTemplate.id"
       [disabled]="isRequired(consentTemplate?.id)"
     />
     <span class="form-check-label cx-be-bold">
-      {{ consentTemplate?.name }}
+      <div *ngIf="consentTemplate?.name">
+        {{ consentTemplate?.name }}
+        <br />
+      </div>
```
Platonn: introducing new `<div>` is a breaking change of the markup, we shouldn't do in _minor_ release. (for more, see [docs](https://sap.github.io/spartacus-docs/breaking-changes/#avoiding-breaking-changes))

Can we simply replace `<div>` with `<ng-container>`?

projects/storefrontlib/cms-components/myaccount/consent-management/components/consent-form/consent-management-form.component.html
```diff
@@ -5,12 +5,15 @@
       class="form-check-input"
       (change)="onConsentChange()"
       [checked]="consentGiven"
+      [name]="consentTemplate.id"
```
Platonn, line 8: Question: I'm not against adding a "name" attribute to this native `<input>`, I like to just better understand:

I'm just curious, is it crucial for this PR to add this "name" attribute here? In other words: if we don't add this attribute here, will anything break / not work? ` for your PR (or the original feature)?

projects/storefrontlib/cms-components/myaccount/consent-management/components/consent-management.component.ts
```diff
@@ -276,8 +281,8 @@ export class ConsentManagementComponent implements OnInit, OnDestroy {
         if (this.isRequiredConsent(template)) {
           return;
         }
+        consentsToGive.push(template);
       }
-      consentsToGive.push(template);
     });
```
Platonn: The closing parenthesis `}` of the `if` statement has been moved. I believe now the whole function `showAll` has a slightly different logic than original.  

Was it a typo? Or it was a deliberate change?
If deliberate, why it was an important change for you? Was it needed for CDC only or also it was fixing something in the core feature?

(it might be perceived as a behavioral breaking change. So we would need to justify it very well, if we want to release this breaking change in a minor version. That's why I need to understand it well)

feature-libs/user/profile/components/register/register-consents.service.ts
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { Injectable } from '@angular/core';
+import { UntypedFormArray, UntypedFormBuilder } from '@angular/forms';
+import { ConsentTemplate } from '@spartacus/core';
+
+@Injectable()
+export class RegisterConsentsService {
```
Platonn: The implementation of this core service looks quite strange and empty.
Let's add a JSDoc explaining that by default there are no consents during the registration. but this class is an extension point (e.g. for CDC) to provide some additional consents needed in the user registration form.

feature-libs/user/profile/components/register/register-consents.service.ts
```diff
@@ -0,0 +1,25 @@
...
+  constructor(protected fb: UntypedFormBuilder) {}
+
+  loadAdditionalConsents(): {
```
Platonn: Maybe it's only me.... The word `load` suggest that the operation will be asynchronous (so I'd expect Observable in the return type). Do you think that word `get` would be more appropriate?

PS. Did you consider simply `loadConsents` or `getConsents` (without the word "additional"). Similar question applies to another method in this class - did you consider removing the word "additional"? There are not "standard" consents in this class I guess (maybe I'm wrong?), so that's why I'm confused we had a word "additional" ;-)

feature-libs/user/profile/components/register/register-consents.service.ts
```diff
@@ -0,0 +1,25 @@
...
+    template: ConsentTemplate;
+    required: boolean;
+  }[] {
+    return [];
+  }
+
+  generateAdditionalConsentsFormControl(): UntypedFormArray {
+    return this.fb.array([]);
+  }
+}
```
Platonn: I'm just curious why did we create a brand new component's service, maybe there was a reason... Did you consider instead adding this same this logic to the existing `RegisterComponentService` (and extending it in existing `CdcRegisterComponentService`)?

integration-libs/cdc/root/consent-management/cdc-user-consent.adapter.ts
```diff
@@ -0,0 +1,86 @@
...
+    consentCode: string,
+    consentId?: string
+  ): Observable<{}> {
+    if (!this.cdcConsentsStorage.checkIfConsentExists(consentId ?? '')) {
+      return super.withdrawConsent(userId, consentCode);
+    } else {
+      return this.cdcUserConsentService
+        .updateCdcConsent(false, consentId ?? '')
+        .pipe(
+          catchError((error: any) => throwError(error)),
```
vigneshkrish8691, line 76: What is the error message thrown if CDC is unable to withdraw the consent?

integration-libs/cdc/root/consent-management/converters/cdc-user-preference.serializer.ts
```diff
@@ -0,0 +1,57 @@
...
+  constructor() {
+    // Intentional empty constructor
+  }
+
+  convert(source: ConsentTemplate, target?: any): any {
+    if (source) {
+      const preference = source.id?.concat('.isConsentGranted');
+      if (preference) {
+        if (source.currentConsent?.consentGivenDate) {
+          target = this.convertToCdcPreference(preference, true);
```
vigneshkrish8691: Both if and else condition calling the same method 'convertToCdcPreference' . you can set a boolean variable inside the if condition and call the method with boolean variable accordingly

integration-libs/cdc/root/consent-management/converters/cdc-user-preference.serializer.ts
```diff
@@ -0,0 +1,57 @@
...
+   * converts a dot separated string to deeply nested object
+   * @param path : dot separated string
+   * @param value : true if consent is given, false if consent is withdrawn
+   * @returns preference object compatible for cdc
+   * example:
+   * input path x.y.z.isConsentGranted
+   * input value: true
+   * output=  x:{y:{z:{isConsentGranted: true}}}
+   */
+  private convertToCdcPreference(path: string, value: any): any {
```
vigneshkrish8691, line 42: Can you add this method to converter.ts which can be reused when preference object have to be parsed?

projects/core/src/user/facade/user-consent.service.ts
anjana-bl: consentId is applicable for CX commerce as well (field id in interface ConsentTemplate  in projects/core/src/model/consent.model.ts)

integration-libs/cdc/root/consent-management/converters/cdc-user-preference.serializer.ts
anjana-bl: done

integration-libs/cdc/root/consent-management/cdc-user-consent.adapter.ts
anjana-bl: if there is any error from CDC SDK, it is re-thrown again. It will be caught in user-consents.effect.ts line 93 for further processing.

feature-libs/user/profile/components/register/register-consents.service.ts
anjana-bl: 1. 'load' is changed to 'get'
2. The reason why 'Additional Consents' was used instead of just 'Consents' because - we already have 'Anonymous Consents' and 'terms and conditions' - so just to make it even more clear that these are "Additional" ones, I used this name. is it fine ?

projects/core/src/user/facade/user-consent.service.ts
anjana-bl: did you mean 'relevant' as 'if it is being used in the functionality' or as 'if this property is present in core model' ?

feature-libs/user/profile/components/register/register-consents.service.ts
anjana-bl: I created a new service just to club all consent related methods to one place - i will merge it to already existing service if its a good practice

integration-libs/cdc/root/consent-management/converters/cdc-user-preference.serializer.ts
anjana-bl: i suppose that is not needed. we dont have to expose this method as the serializer is already exposed. we can re-use the serializer directly, instead of the method

feature-libs/user/profile/components/register/register-consents.service.ts
anjana-bl: done

projects/core/src/user/facade/user-consent.service.ts
Platonn: As far as I understand, consent ID is not used in the CX Commerce functionality, but it's present in the core model.
If that's true, I think it doesn't look that bad, if we add the ID parameter to the core services, because the ID is already present in the core model.

So let's resolve this comment. No need to fix anything.

feature-libs/user/profile/components/register/register-consents.service.ts
Platonn: No big need to move it. I was just curious about your reasoning.
I was thinking about moving it, because the `RegisterComponentService` is already a helper service for the component (e.g. loading the person's titles). So I wouldn't mind placing there also other methods needed for this component (unless there are plenty of them). At the same time, I don't insist on moving it. Up to you ;)

integration-libs/cdc/root/service/cdc-js.service.ts
```diff
@@ -308,7 +331,7 @@ export class CdcJsService implements OnDestroy {
    *
    * @param response
    */
-  protected handleLoginError(response: any) {
```
vigneshkrish8691: Why removed protected?This is  already exposed.

feature-libs/user/profile/components/register/register-consents.service.ts
Platonn: Thanks. Understood. No need to change

integration-libs/cdc/user-account/login-form/reconsent/cdc-reconsent.service.ts
```diff
@@ -0,0 +1,124 @@
...
+   * @param consentId - array of consent IDs
+   * @param userParams - data from login session
+   * @returns error message if any
+   */
+  saveReconsent(consentId: string[], userParams: any): string {
+    const errorMessage: string = '';
+    this.subscription.add(
+      this.cdcJsService.didLoad().subscribe((cdcLoaded) => {
+        if (cdcLoaded) {
+          consentId.forEach((consentTemplateId) => {
```
vigneshkrish8691: Send all consents as an array to the preference object.Instead of one by one which will cause many unnecessary network calls

projects/storefrontlib/cms-components/myaccount/consent-management/components/consent-management.component.ts
anjana-bl: All required consents will be disabled in the UI. When we do a 'select All' in my-account->consent management, ideally these disabled consents should not be picked for updated. This is done correctly in 'rejectAll' method, so I perceived this as a bug and fixed it here

feature-libs/user/profile/components/register/register-consents.service.ts
anjana-bl: okay .. any ways i moved it :)

integration-libs/cdc/user-profile/register/cdc-register-consents.service.ts
```diff
@@ -0,0 +1,71 @@
...
+  }[] {
+    const templates: ConsentTemplate[] = this.fetchCdcConsentsForRegistration();
+    const returnConsents: {
+      template: ConsentTemplate;
+      required: boolean;
+    }[] = [];
+    for (const template of templates) {
+      const returnConsent: any = {};
+      returnConsent['template'] = template;
+      returnConsent['required'] = true;
```
vigneshkrish8691: All consents are you setting required = true ?

projects/storefrontlib/cms-components/myaccount/consent-management/components/consent-form/consent-management-form.component.html
anjana-bl: I added it for the purpose of cypress test case. With this new development, both commerce and cdc consents will be shown in UI. In the new cypress test case, i would like to give and withdraw consent specific to cdc (instead of just picking the first consent (which is mostly a commerce consent) using get().first().check()). So i added name here and in cypress test case i pick a checkbox based on name

projects/storefrontlib/cms-components/myaccount/consent-management/components/consent-management.component.ts
anjana-bl: done

projects/storefrontlib/cms-components/myaccount/consent-management/components/consent-form/consent-management-form.component.html
anjana-bl: done

projects/storefrontlib/cms-components/myaccount/consent-management/consent-management.service.ts
anjana-bl: done

projects/storefrontlib/cms-components/myaccount/consent-management/consent-management.service.ts
anjana-bl: done

integration-libs/cdc/root/consent-management/services/cdc-consent-management.service.ts
anjana-bl: done

projects/core/src/user/facade/user-consent.service.ts
anjana-bl: okay

integration-libs/cdc/root/consent-management/cdc-user-consent.adapter.ts
anjana-bl: only if we do this, then the error will get caught in catchError of user-consents.effect.ts, hence doing it

integration-libs/cdc/user-profile/register/cdc-register-consents.service.ts
anjana-bl: yes, as mentioned by Tal in our group chat - during registration, all active cdc consents should be GIVEN mandatorily. hence making it to true

integration-libs/cdc/root/service/cdc-js.service.ts
anjana-bl: done

integration-libs/cdc/user-account/login-form/reconsent/cdc-reconsent.service.ts
```diff
@@ -0,0 +1,124 @@
...
+   * @param userId - user id
+   * @param password - password
+   */
+  reLogin(userId: string, password: string): void {
+    this.subscription.add(
+      this.cdcJsService.didLoad().subscribe((cdcLoaded) => {
+        if (cdcLoaded) {
+          // Logging in using CDC Gigya SDK
+          this.cdcJsService
+            .loginUserWithoutScreenSet(userId, password)
```
vigneshkrish8691: See if you can make use of the existing login functionality by passing the required param for the login.

integration-libs/cdc/user-account/login-form/reconsent/cdc-reconsent.service.ts
```diff
@@ -0,0 +1,124 @@
+/*
+ * SPDX-FileCopyrightText: 2023 SAP Spartacus team <spartacus-team@sap.com>
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { Injectable, OnDestroy } from '@angular/core';
```
vigneshkrish8691: Check if the whole service can be moved to one service where we have everything for the Consent mgmt feature.

integration-libs/cdc/root/consent-management/cdc-user-consent.adapter.ts
Platonn: Perhaps I'm missing something obvious... But to me at the first sight, it doesn't bring any value. We're not handling the error in any way but just re-throwing 🤔  it looks similar to:
```ts
try {
  getSomething()
} catch (error:any) {
  throw error;
}
```


I pretty much believe that even if you remove this line, the error will be propagated out to the `user-consents.effect.ts`.

It's not crucial, I don't want to block this PR with my doubt. But I really recommend to try removing it and checking if it still works as expected, when you have some free time. (If it turns our that this line is really needed, the benefit is I might learn something from it... 😉  )

projects/storefrontlib/cms-components/myaccount/consent-management/components/consent-management.component.ts
Platonn: Understood. Code-wise it seems to be a justified bugfix.

For the clarity of the git history: what do you think about reverting this change from this PR, and creating a new `fix: ...` PR as a followup of this PR (+ it would be good to add unit test there, because current tests didn't spot this bug)?

integration-libs/cdc/user-account/login-form/reconsent/cdc-reconsent.service.ts
anjana-bl: done

projects/storefrontlib/cms-components/myaccount/consent-management/components/consent-management.component.ts
anjana-bl: done

Platonn (dismissed):
changes in core libs LGTM

Platonn (approved):

vigneshkrish8691 (approved):

feature-libs/user/profile/components/register/register-component.service.ts
```diff
@@ -18,7 +20,8 @@ import { Observable } from 'rxjs';
 export class RegisterComponentService {
   constructor(
     protected userRegisterFacade: UserRegisterFacade,
-    protected globalMessageService: GlobalMessageService
+    protected globalMessageService: GlobalMessageService,
+    protected fb: UntypedFormBuilder
```
Platonn, line 24: New required constructor param is a breaking change, not allowed in minor. 

Let's make it optional:

```diff
- protected fb: UntypedFormBuilder
+ protected fb?: UntypedFormBuilder
```

feature-libs/user/profile/components/register/register-component.service.ts
```diff
@@ -46,4 +49,27 @@ export class RegisterComponentService {
...
+
+  /**
+   * Generate form control if any additional consents that needs to be provided during registration
+   * In core feature, no additional consents are necessary during registration.
+   * In integration scenarios, eg: cdc, this method will be overridden to return
+   * form controls for necessary cdc consents
+   */
+  generateAdditionalConsentsFormControl(): UntypedFormArray {
+    return this.fb.array([]);
+  }
```
Platonn, line 74: When we implement [this comment](https://github.com/SAP/spartacus/pull/17497#pullrequestreview-1485827527), we should gracefully fallback, when `this.fb` is undefined:


```diff
-  generateAdditionalConsentsFormControl(): UntypedFormArray {
-    return this.fb.array([]);
+  generateAdditionalConsentsFormControl(): UntypedFormArray | undefined {
+    return this.fb?.array([]) ?? undefined;
   }
```

feature-libs/user/profile/components/register/register.component.ts
```diff
@@ -60,6 +61,8 @@ export class RegisterComponent implements OnInit, OnDestroy {
         value: false,
         disabled: this.isConsentRequired(),
       }),
+      additionalConsents:
+        this.registerComponentService?.generateAdditionalConsentsFormControl(),
```
Platonn, line 65: When we implement [this comment](https://github.com/SAP/spartacus/pull/17497#pullrequestreview-1485827527),
we should gracefully fallback, when the returned value is undefined:

```diff
       additionalConsents:
-        this.registerComponentService?.generateAdditionalConsentsFormControl(),
+        this.registerComponentService.generateAdditionalConsentsFormControl?.() ??
+        this.fb.array([]),

feature-libs/user/profile/components/register/register-component.service.ts
anjana-bl: done

feature-libs/user/profile/components/register/register-component.service.ts
anjana-bl: done

feature-libs/user/profile/components/register/register.component.ts
anjana-bl: done

PR #17502, opened 2023-06-09
fix: delivery address resets when entering delivery-address route

kpawelczak:


feature-libs/checkout/base/components/checkout-delivery-address/checkout-delivery-address.component.ts
```diff
@@ -138,6 +139,9 @@ export class CheckoutDeliveryAddressComponent implements OnInit {
 
   addAddress(address: Address | undefined): void {
     if (!address) {
```
rmch91, line 141: Maybe it is better to get rid of the nested if statement at to leave this guard as it is, and add the new logic just before 
 ```
  if (!address && this.selectedAddress) {
      this.next();
    }

    if (!address) {
      return;
    }
```

rmch91 (approved):

PR #17503, opened 2023-06-09
fix: Ensure 'Create New Customer' button appearing on customer list page

i53577:
JIRA: https://jira.tools.sap/browse/CXSPA-3597

### DIFF
--- a/feature-libs/asm/styles/components/_customer-list.component.scss
+++ b/feature-libs/asm/styles/components/_customer-list.component.scss
@@ -155,6 +155,9 @@ $sapFioriContent_LabelColor: #6a6d70;
       padding-bottom: 0;
     }
     max-height: calc(100vh - 260px);
+    @include media-breakpoint-down(xs) {
+      max-height: calc(100vh - 470px);
+   }
     overflow-y: auto;
   }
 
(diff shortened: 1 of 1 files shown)


feature-libs/asm/styles/components/_customer-list.component.scss
```diff
@@ -155,6 +155,9 @@ $sapFioriContent_LabelColor: #6a6d70;
       padding-bottom: 0;
     }
     max-height: calc(100vh - 260px);
```
i53577, line 157: This height is too tall for mobile. With this height, the header including 'create new customer' button will be placed outside the visible range.

feature-libs/asm/styles/components/_customer-list.component.scss
```diff
@@ -155,6 +155,9 @@ $sapFioriContent_LabelColor: #6a6d70;
...
+    @include media-breakpoint-down(xs) {
```
i53577, line 158: the bug only exists in mobile and asm customer list uses `xs` for mobile, so we only need to adjust the height of customer list body for mobile to avoid too tall.

i53577, line 158: `@include media-breakpoint-down(xs)`
means the following class only works when the screen size is below the breakpoint of `xs`

scarai-sap (approved):
LGTM

PR #17508, opened 2023-06-12
feature: show validation message for empty required field if input is removed [CXSPA-3630]

Uli-Tiger:
Before we let the validation of empty required input fields was managed by the input control itself by just setting the required property to true. This had 2 draw backs.
* if an existing input was removed this was considered as validation error an not sent to the engine. After updating another attribute, the engine returned the previous value, hence the user input was lost.
* The corresponding 'missing required field' message is generated in the attribute footer, not by the validation, hence it was missing if an existing input was removed, as the attribute was still considered 'complete'.
With this PR only the 'ng-invalid' style class, which causes the red box, is set for empty required input fields, hence eliminating the two drawback mentioned before.

### DIFF
--- a/feature-libs/product-configurator/rulebased/components/attribute/types/input-field/configurator-attribute-input-field.component.html
+++ b/feature-libs/product-configurator/rulebased/components/attribute/types/input-field/configurator-attribute-input-field.component.html
@@ -1,30 +1,59 @@
 <div id="{{ createAttributeIdForConfigurator(attribute) }}" class="form-group">
-  <input
-    [formControl]="attributeInputForm"
-    class="form-control"
-    [class.ng-invalid]="
-      isRequired &&
-      attribute.incomplete &&
-      (!attributeInputForm.value || attributeInputForm.value.length === 0)
-    "
-    maxlength="{{ attribute.maxlength }}"
-    [attr.aria-label]="
-      attribute.userInput === undefined || attribute.userInput.length === 0
-        ? ('configurator.a11y.valueOfAttributeBlank'
-          | cxTranslate
-            : {
-                attribute: attribute.label
-              })
-        : ('configurator.a11y.valueOfAttributeFull'
-          | cxTranslate
-            : {
-                value: attribute.userInput,
-                attribute: attribute.label
-              })
-    "
-    [attr.aria-describedby]="createAttributeUiKey('label', attribute.name)"
-    [cxFocus]="{
-      key: createAttributeIdForConfigurator(attribute)
-    }"
-  />
+  <ng-container *cxFeatureLevel="'!6.2'">
+    <input
+      [formControl]="attributeInputForm"
+      [required]="isRequired"
+      class="form-control"
+      attr.required="{{ attribute.required }}"
+      maxlength="{{ attribute.maxlength }}"
+      [attr.aria-label]="
+        attribute.userInput === undefined || attribute.userInput.length === 0
+          ? ('configurator.a11y.valueOfAttributeBlank'
+            | cxTranslate
+              : {
+                  attribute: attribute.label
+                })
+          : ('configurator.a11y.valueOfAttributeFull'
+            | cxTranslate
+              : {
+                  value: attribute.userInput,
+                  attribute: attribute.label
+                })
+      "
+      [attr.aria-describedby]="createAttributeUiKey('label', attribute.name)"
+      [cxFocus]="{
+        key: createAttributeIdForConfigurator(attribute)
+      }"
+    />
+  </ng-container>
+  <ng-container *cxFeatureLevel="'6.2'">
+    <input
+      [formControl]="attributeInputForm"
+      class="form-control"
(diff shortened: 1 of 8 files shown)


feature-libs/product-configurator/rulebased/components/attribute/types/input-field/configurator-attribute-input-field.component.html
```diff
@@ -1,9 +1,12 @@
 <div id="{{ createAttributeIdForConfigurator(attribute) }}" class="form-group">
   <input
     [formControl]="attributeInputForm"
-    [required]="isRequired"
     class="form-control"
-    attr.required="{{ attribute.required }}"
+    [class.ng-invalid]="
```
ChristophHi: That is breaking.
Please also synch with @Larisa-Staroverova as I assume https://jira.tools.sap/browse/CXSPA-3155 also reworks the  message handling for required input fields?

feature-libs/product-configurator/rulebased/components/attribute/types/numeric-input-field/configurator-attribute-numeric-input-field.component.html
```diff
@@ -4,7 +4,10 @@
 <div id="{{ createAttributeIdForConfigurator(attribute) }}" class="form-group">
   <input
     [formControl]="attributeInputForm"
-    [required]="isRequired"
+    [class.ng-invalid]="
```
ChristophHi: Breaking

ChristophHi (changes requested):

ChristophHi (changes requested):
Numeric input component does not work, in the console I see
Can't bind to 'cxFeatureLevel' since it isn't a known property of 'ng-container' (used in the 'ConfiguratorAttributeNumericInputFieldComponent' component template).

Uli-Tiger:
> Numeric input component does not work, in the console I see Can't bind to 'cxFeatureLevel' since it isn't a known property of 'ng-container' (used in the 'ConfiguratorAttributeNumericInputFieldComponent' component template).

fixed

feature-libs/product-configurator/rulebased/components/attribute/types/input-field/configurator-attribute-input-field.component.spec.ts
```diff
@@ -277,4 +291,21 @@ describe('ConfigAttributeInputFieldComponent', () => {
...
+      component.attribute.userInput = 'abc';
+      expect(component.isUserInputEmpty).toBe(false);
+    });
+
+    it('should return true if the user input only contains blanks', () => {
+      component.attribute.userInput = '  ';
+      expect(component.isUserInputEmpty).toBe(true);
+    });
+
+    it('should return true if the ure is no user input', () => {
```
ChristophHi:
```suggestion
    it('should return true if there is no user input', () => {
```

feature-libs/product-configurator/rulebased/components/attribute/types/input-field/configurator-attribute-input-field.component.ts
```diff
@@ -98,6 +98,16 @@ export class ConfiguratorAttributeInputFieldComponent
     }
   }
 
+  /**
+   * Checks if the user input has a non blank value.
```
Larisa-Staroverova: Normally, we write something like this:
> Verifies whether the user input has a non-blank value

feature-libs/product-configurator/rulebased/components/attribute/types/input-field/configurator-attribute-input-field.component.ts
```diff
@@ -98,6 +98,16 @@ export class ConfiguratorAttributeInputFieldComponent
...
+   * @returns isEmpty?
```
Larisa-Staroverova: Normally, we write something like this:

> returns {boolean} - 'True' if a user input is not undefined, empty or contains a blank, otherwise 'false'.

feature-libs/product-configurator/rulebased/components/attribute/types/numeric-input-field/configurator-attribute-numeric-input-field.component.service.ts
```diff
@@ -348,7 +349,7 @@ export class ConfiguratorAttributeNumericInputFieldService {
     currentValue?: string
   ): ValidatorFn {
     return (control: AbstractControl): { [key: string]: any } | null => {
-      const input: string = control.value;
+      const input: string = control.value.trim();
```
ChristophHi:
```suggestion
      const input = control.value?.trim();
```
We later check if input is defined (don't know if control.value is defined in all cases)

ChristophHi (approved):

ChristophHi (changes requested):
See my comments

ChristophHi (dismissed):

ChristophHi (approved):

PR #17512, opened 2023-06-13
feat: CXSPA-2251 Contextual logs in SSR

pawelfras:
It's an end to the chaos of interlaced messages in Kibrana/OpenSearch, coming from multiple parallel server-side-renderings and NodeJS pods. Gone are the days of sifting through a disorganized mess with no additional context or structure in your messages. In the past, identifying the origin of a log/warn/error could be a blind guessing game, relying on approximate timestamps and frustrating trial and error. And no more unreadable error stacktraces splitted each line into a separate log entry in Kibana/OpenSearch.

With this new feature, your logs are now well structured, presented in JSON format with clear message and context fields including the request `url`, request `uuid` (randomly generated for each request) and request `timeReceived`. Kibana/OpenSearch intelligently interprets these fields and presents them in separate columns, enabling effortless filtering and searching. And that's not all! Multi-line error stacktraces are now elegantly consolidated into a single line log entries, streamlining your troubleshooting process. 

**Recommendation for customers: How to enable the feature**:

- enable `logger: true` in `SsrOptimizationOptions` in your `server.ts`
- import `ErrorHandlingModule.forRoot()` in your app (e.g. `app.module.ts`)

**Architectural details:**

- Introduced `LoggerService` to replace the global `console` object.
- Replaced almost all uses of `console` object in our libraries with the `LoggerService` wrapper.
- In the browser, `LoggerService` works just like `console` - under the hood it forwards its arguments to the global `console` object
- In SSR (ExpressJS), `LoggerService` is over-provided by `ExpressLoggerService`, which adds context of an Express `request` to the message. Instead of forwarding it to the console, it passes the log to a concrete logger, which can be injected from the ExpressJS app via the injection token `EXPRESS_SERVER_LOGGER`
- `OptimizedSsrEngine` provides a concrete logger via the injection token `EXPRESS_SERVER_LOGGER`
- For backward-compatibility, by default the `LegacyExpressServerLogger` is provided, which logs messages as before: via `console` and as a plain string without any additional context
- When customer explicitly passes the option `logger: true` of `SsrOptimizationOptions`, it instructs the `OptimizedSsrEngine` to provide the `DefaultExpressServerLogger` (which will be a *real* default since a new major release) that forwards logs as a **structured JSON** to the `console` and moreover **adds an useful context of the current Express request** to the JSON log.
- alternatively, customers can pass to the property `logger` of `SsrOptimizationOptions` any custom concrete logger class (matching the interface `ExpressServerLogger`). So they can enrich the context of the logger to match their needs (e.g. attach a value of some http header to each log). Moreover, the option to configure a custom `logger` class also allows to forward logs to some 3rd party logger library (e.g. Pino or Winston) instead of the `console` object.

**Customers have to import `ErrorHandlingModule.forRoot()` manually (until next major release):**

- For backward-comparibility, in this minor release by default all unhandled exceptions in your Angular SSR app are still logged as a plain string without any additional context. Why: Customers might have already provided a custom [ErrorHandler](https://angular.io/api/core/ErrorHandler) in their apps to handle errors. We would like to provide our custom Angular `Errorhandler` that would forward error messages to our `LoggerService`. But this could potentially over-provide custom `ErrorHandler` already provided by customers, which would be a breaking change. So we plan to provide our custom `ErrorHandler` only in a future major release.
- To enable a rich JSON context for error messages, customers need to import **manually** our new `ErrorHandlingModule.forRoot()` from `@spartacus/core` in the app (which under the hood provides our custom `ErrorHandler). This requires a deliberate manual import from customers, so it's not a breaking change.
- In a future major release, `ErrorHandlingModule.forRoot()` will be imported by default as part of our core module `BaseCoreModule`, so customers no longer will need to import it separately manually.

closes CXSPA-2251


pawelfras:
Previous PR (too many not related changes after changing target branch from `develop-6.1.x` to `develop-6.2.x`)
https://github.com/SAP/spartacus/pull/17441

Not resolved comments:
https://github.com/SAP/spartacus/pull/17441#discussion_r1221735659

https://github.com/SAP/spartacus/pull/17441#discussion_r1219171868

projects/core/src/i18n/i18next/i18next-plugins/i18next-logger-plugin.ts
```diff
@@ -0,0 +1,27 @@
...
+//TODO: add unit tests if approved
+export const I18NEXT_LOGGER_PLUGIN = new InjectionToken<LoggerModule>(
+  'I18NEXT_LOGGER_PLUGIN',
+  {
+    providedIn: 'root',
+    factory: () => {
+      const logger = inject(LoggerService);
+      return {
+        type: 'logger',
+        log: (args) => logger.log(...args),
```
pawelfras, line 28: Regarding https://github.com/SAP/spartacus/pull/17441#discussion_r1219171868:

when I used spread operator and then simply log to the console, as you suggested, 
```
    return {
        type: 'logger',
        log: (...args) => {
          console.log(...args);
          logger.log(...args);
        },
        warn: (args) => logger.warn(...args),
        error: (args) => logger.error(...args),
    };
```
it logged an array... 
image

maybe it's about how i18next is passing args to the plugin? Anyway spread operator on the function's args is messing up the output.

pawelfras: It looks like i18next is always passing one character which is an array (that can contain many values) - [source code](https://github.com/i18next/i18next/blob/5de7d4687bdf03dd5c5b4938d8bfb8db78dc4720/i18next.js#L34-L35)
Because of that, if we use the spread operator on args in the arrow function, we're saying that there may be more than one argument, that's why logger (or console.log) outputs an array when we use to spread on provided args again. See in example:
image

Platonn: As we discussed offline, i18next expects just a single agrument `args` which is an array of parameters to be passed to a concrete logger (which was suprising for us).

core-libs/setup/ssr/logger/loggers/express-server-logger.spec.ts
```diff
@@ -0,0 +1,111 @@
+import { Request } from 'express';
+import {
+  DefaultExpressServerLogger,
+  ExpressServerLoggerContext,
+} from './express-server-logger';
+
+describe('ExpressServerLogger', () => {
+  const logger = new DefaultExpressServerLogger();
+  jest.useFakeTimers().setSystemTime(new Date('2023-05-26'));
```
Platonn: [NITPICK]

I'm aware that our default logger is stateless and doesn't depend on third party state. That said, it's just a good practice to re-do a setup of the test for each run separately, to prevent a _potential_ cross-test interference:

(similar comment applies also to other spec files in this PR, where we do `new Xxx()`)

```suggestion
  let logger: DefaultExpressServerLogger;
  
  beforeEach(() => {
    logger = new DefaultExpressServerLogger();
    jest.useFakeTimers().setSystemTime(new Date('2023-05-26'));
  });
```

core-libs/setup/ssr/logger/loggers/express-server-logger.spec.ts
```diff
@@ -0,0 +1,111 @@
+import { Request } from 'express';
+import {
+  DefaultExpressServerLogger,
+  ExpressServerLoggerContext,
+} from './express-server-logger';
+
+describe('ExpressServerLogger', () => {
```
Platonn: [MINOR]

It's a test suite for `DefaultExpressServerLogger`. Let's adjust the description and the file name.
```suggestion
describe('DefaultExpressServerLogger', () => {
```

core-libs/setup/ssr/logger/loggers/express-server-logger.ts
```diff
@@ -0,0 +1,85 @@
...
+ */
+
+import { InjectionToken } from '@angular/core';
+import { Request } from 'express';
+import { SsrOptimizationOptions } from '../../optimized-engine';
+
+export interface ExpressServerLoggerContext {
+  request: Request;
+  options?: Partial<SsrOptimizationOptions>;
+}
```
Platonn: [MEDIUM]

IMHO only the `request: Request` is specific to the ExpressJS (and therefore to. `ExpressServerLoggerContext`)
I don't think we should couple it specifically with `options` and `SsrOptimizationOptions`.

Btw. I think this interface should: (1) require `request` and (2) allow for any context. What do you think?
```suggestion
export interface ExpressServerLoggerContext {
  request: Request;
  [_key: string]: any;
}
```

core-libs/setup/ssr/logger/loggers/express-server-logger.ts
```diff
@@ -0,0 +1,85 @@
...
+  warn(message: string, context: ExpressServerLoggerContext): void;
+  error(message: string, context: ExpressServerLoggerContext): void;
+  info(message: string, context: ExpressServerLoggerContext): void;
+  debug(message: string, context: ExpressServerLoggerContext): void;
+}
+
+/**
+ *
+ * Default logger used in SSR (ExpressJS) to enhance logs visible e.g. in monitoring tools e.g. Kibana.
+ * It logs It outputs a JSON with properties "message" and "context",
```
Platonn: [MINOR]

just typos
```suggestion
 * Default logger used in SSR (ExpressJS) to enhance logs visible e.g. in monitoring tools e.g. Kibana.
 * It outputs a JSON with properties "message" and "context",
```

core-libs/setup/ssr/logger/loggers/express-server-logger.ts
```diff
@@ -0,0 +1,85 @@
...
+
+    return JSON.stringify(object);
+  }
+
+  protected mapRequest(request: Request): Record<string, any> {
+    return {
+      url: request.originalUrl,
+      ...request.res?.locals.cx.request,
+      /* remove headers before release. */
+      headers: request.headers, //
```
Platonn: TODO: just a reminder to remove it before merge

core-libs/setup/ssr/logger/loggers/express-server-logger.ts
```diff
@@ -0,0 +1,85 @@
...
+  debug(message: string, context: ExpressServerLoggerContext): void;
+}
+
+/**
+ *
+ * Default logger used in SSR (ExpressJS) to enhance logs visible e.g. in monitoring tools e.g. Kibana.
+ * It logs It outputs a JSON with properties "message" and "context",
+ * which contains a "timestamp" and details of the "request" ("url", "uuid", "timeReceived")
+ */
+export class DefaultExpressServerLogger {
```
Platonn: [MINOR]

Let's move it to a separate file

core-libs/setup/ssr/logger/loggers/legacy-server-logger.spec.ts
```diff
@@ -0,0 +1,49 @@
+import { LegacyExpressServerLogger } from './legacy-server-logger';
+
+describe('LegacyServerLogger', () => {
```
Platonn: [MINOR]

Let's adjust the description and the spec file name
```suggestion
describe('LegacyExpressServerLogger', () => {
```

core-libs/setup/ssr/logger/loggers/legacy-server-logger.ts
```diff
@@ -0,0 +1,41 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import {
+  ExpressServerLogger,
+  ExpressServerLoggerContext,
+} from './express-server-logger';
+
+/**
+ * @deprecated since 6.2, will be removed in 7.0 as contextual logging will be enabled by default.
```
Platonn: [MINOR]

Let's not promise a 7.0 version, as the policy of releases might change (e.g. we might potentially limit the number of breaking changes in 7.0, and only upgrade Angular. nothing decided yet though)
```suggestion
 * @deprecated since 6.2, will be removed in a new major version, as contextual logging will be enabled by default.
```

core-libs/setup/ssr/logger/loggers/legacy-server-logger.ts
```diff
@@ -0,0 +1,41 @@
...
+ * Default implementation of ServerLogger that delegates log messages to the native `console` object without providing any context.
+ * It's used when contextual logging is disabled.
+ *
+ *
```
Platonn: typo & phrasing

```suggestion
 * Default implementation of `ExpressServerLogger` that just delegates log messages to the native `console` object without providing any context.
 * It's used when contextual logging is disabled.
```

core-libs/setup/ssr/logger/services/express-logger.service.ts
```diff
@@ -0,0 +1,55 @@
...
+ */
+
+import { Injectable, inject } from '@angular/core';
+import { REQUEST } from '@nguniversal/express-engine/tokens';
+import { LoggerService } from '@spartacus/core';
+import { formatWithOptions } from 'util';
+import { EXPRESS_SERVER_LOGGER } from '../loggers';
+
+@Injectable({ providedIn: 'root' })
+export class ExpressLoggerService implements LoggerService {
```
Platonn, line 14: [MINOR]

Let's add JSDocs.

Just for inspiration:
```
Custom `LoggerService` used in ExpressJS. 

It converts the input arguments to a final string message similar as the native `console`
does (using the native function `format` from `node:util`) and passes this message
to a concrete server logger, used in ExpressJS.

Besides the message, it also passes the current `request` of ExpressJS as an additional
context to the concrete server logger.

File: core-libs/setup/ssr/logger/services/express-logger.service.ts [hunk 12]
```diff
@@ -0,0 +1,55 @@
...
+  protected formatLogMessage(message?: any, ...optionalParams: any[]): string {
+    /**
+     * built-in util function 'formatWithOptions' was used to not break provided message.
+     * That helps to present logs in moonitoring tools like Kibana in a proper way.
+     */
+    return formatWithOptions(
+      { breakLength: Infinity },
+      message,
+      ...optionalParams
+    );
```
Platonn, line 53: [MINOR]

To be precise, `format` (or `formatWithOptions`) was used to convert multiple parameters to a string message.
I believe your comment refers to the specific option  `breakLength: Infinity`.

Let's move this comment over the specific parameter.

```suggestion
    return formatWithOptions(
      // Prevent automatically breaking a long string message into multiple lines.
      // Otherwise, multi-line logs would be treated on the server as separate log entries.
      { breakLength: Infinity },
      message,
      ...optionalParams
    );
```

PS. Side note: it's not Kibana's fault that shows multi-line logs as separate entries, but rather a default behavior of log agents (e.g. see docs of [FluentBit](https://docs.fluentbit.io/manual/administration/configuring-fluent-bit/multiline-parsing)) that interprets each line of `stdout` as a separate entry and forwards those separate entries to the logs store presented by Kibana

core-libs/setup/ssr/logger/services/prerendering-logger.service.ts
```diff
@@ -0,0 +1,34 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { Injectable } from '@angular/core';
+import { LoggerService } from '@spartacus/core';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class PrerenderingLoggerService extends LoggerService {
```
Platonn, line 13: [MINOR]

Let's add JSDocs

For inspiration:
```
Custom `LoggerService` used in pre-rendering in the server environment.

It simply forwards the arguments to the native `console` object.

File: core-libs/setup/ssr/optimized-engine/optimized-ssr-engine.spec.ts [hunk 14]
```diff
@@ -20,6 +21,12 @@ jest.mock('fs', () => ({
 }));
 const consoleLogSpy = jest.spyOn(console, 'log').mockImplementation();
 
+class MockServerLogger implements Partial<ExpressServerLogger> {
```
Platonn: [MINOR]

MockServerLogger vs MockExpressServerLogger ?

(similar comment applies to `server-logger-service-factory.spec.ts`)

core-libs/setup/ssr/optimized-engine/optimized-ssr-engine.spec.ts
```diff
@@ -795,12 +811,18 @@ describe('OptimizedSsrEngine', () => {
         });
 
         tick(100);
+        // const firstRequest = getRequest(engineRunner, requestUrl);
+        // const secondRequest = getRequest(engineRunner, requestUrl);
+
```
Platonn: [MINOR]

👀 
```suggestion
```

core-libs/setup/ssr/optimized-engine/optimized-ssr-engine.spec.ts
```diff
@@ -1147,4 +1176,54 @@ describe('OptimizedSsrEngine', () => {
       }));
     });
   });
+
+  describe('logger option', () => {
+    it('should use ExpressServerLogger if logger is true', () => {
```
Platonn, line 1179: [MINOR]

```suggestion
    it('should use DefaultExpressServerLogger if logger is true', () => {
```

core-libs/setup/ssr/optimized-engine/optimized-ssr-engine.ts
```diff
@@ -39,6 +47,7 @@ export type SsrCallbackFn = (
 export class OptimizedSsrEngine {
   protected currentConcurrency = 0;
   protected renderingCache = new RenderingCache(this.ssrOptions);
+  protected logger: ExpressServerLogger;
```
Platonn: [MEDIUM]

What about keeping it private?
```suggestion
  private logger: ExpressServerLogger;
```

core-libs/setup/ssr/optimized-engine/optimized-ssr-engine.ts
```diff
@@ -405,29 +454,62 @@ export class OptimizedSsrEngine {
...
+      this.log(`Rendering completed (${request?.originalUrl})`, true, {
+        request,
+      });
       this.currentConcurrency--;
 
       renderCallback(err, html);
     });
   }
+
+  //CXSPA-3680 - remove this method in 7.0
```
Platonn, line 498: [MINOR]

Should we drop this whole method or change it in 7.0?

```suggestion
  //CXSPA-3680 - in a new major version let's use `DefaultExpressServerLogger` by default. And drop `LegacyExpressServerLogger`
```

core-libs/setup/ssr/optimized-engine/optimized-ssr-engine.ts
```diff
@@ -80,14 +90,26 @@ export class OptimizedSsrEngine {
       if (typeof value === 'function') {
         return value.toString();
       }
+      if (this.isServerLogger(value)) {
+        return value.constructor.name;
+      }
```
Platonn, line 95: [MINOR]

What about simply:
```suggestion
      if (value?.constructor?.name) {
        return value.constructor.name;
      }
```

If we really have to use `isServerLogger` (or `isExpressServerLogger`), I think it shouldn't be a method of `OptimizedSsrEngine`. It's not its responsibility. I'd rather move this type guard to `express-server-logger.ts`

Anyway IMHO `if (value?.constructor?.name)` should suffice. What do you think?

core-libs/setup/ssr/optimized-engine/optimized-ssr-engine.ts
```diff
@@ -80,14 +90,26 @@ export class OptimizedSsrEngine {
...
+      : `[spartacus] SSR optimization engine initialized with the following options: ${stringifiedOptions}`;
+    this.log(message, true, {
+      options: {
+        ...this.ssrOptions,
+        logger:
+          this.ssrOptions.logger === true
+            ? this.ssrOptions.logger
+            : this.ssrOptions.logger?.constructor.name,
+      },
+    } as ExpressServerLoggerContext);
```
Platonn: [MINOR]

It's not obvious why we're adding `as ExpressServerLoggerContext` here. So IMHO it's worth adding a //code comment, that we're deliberately abusing here the interface of `ExpressServerLoggerContext` - it expects a `request` context, but the current logged message is printed at the start of the server and there is no `request` available yet.

core-libs/setup/ssr/optimized-engine/optimized-ssr-engine.ts
```diff
@@ -295,9 +335,16 @@ export class OptimizedSsrEngine {
     });
   }
 
-  protected log(message: string, debug = true): void {
-    if (!debug || this.ssrOptions?.debug) {
-      console.log(message);
+  protected log(
+    message: string,
+    debug = true,
+    logMetadata?: ExpressServerLoggerContext
```
Platonn, line 342: [MINOR]

let's add a TODO note (not to miss it in the future), because `... || ({} as ExpressServerLoggerContext)` in line 346 is just a temporary backward-compatible workaround

```suggestion
    //CXSPA-3680 - in a new major, let's make this argument required
    logMetadata?: ExpressServerLoggerContext
```

core-libs/setup/ssr/optimized-engine/optimized-ssr-engine.ts
```diff
@@ -80,14 +90,26 @@ export class OptimizedSsrEngine {
...
+    const message = this.ssrOptions.logger
+      ? `[spartacus] SSR optimization engine initialized`
+      : `[spartacus] SSR optimization engine initialized with the following options: ${stringifiedOptions}`;
+    this.log(message, true, {
+      options: {
+        ...this.ssrOptions,
+        logger:
+          this.ssrOptions.logger === true
+            ? this.ssrOptions.logger
+            : this.ssrOptions.logger?.constructor.name,
```
Platonn: [MEDIUM]

Let's make it fully safe against unexpected values

```suggestion
            : this.ssrOptions.logger?.constructor?.name,
```

core-libs/setup/ssr/optimized-engine/ssr-optimization-options.ts
```diff
@@ -109,6 +110,20 @@ export interface SsrOptimizationOptions {
...
+
+  /**
+   * Config for improving logged messages with context and JSON structure.
+   *
+   * It enhances the logs in SSR by adding context, including the request's details,
+   * and structuring them as JSON.
+   *
+   * The `logger` property is optional and accepts two values:
+   * - `true`:         Enables the default logger and enhances the logs.
+   * - `ServerLogger`: Interprets the given `ServerLogger` as a custom logger
```
Platonn: [MEDIUM]

typos
```suggestion
   * - `ExpressServerLogger`: Interprets the given `ExpressServerLogger` as a custom logger
```

projects/core/src/i18n/i18next/i18next-plugins/i18next-logger-plugin.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+    expect(plugin.type).toEqual('logger');
+  });
+
+  it('should log', () => {
+    const log = spyOn(logger, 'log');
+
+    plugin.log(['test']);
+
+    expect(log).toHaveBeenCalledWith('test');
+  });
```
Platonn, line 29: [MINOR]

Let's add some test with _multiple_ arguments

projects/core/src/i18n/translate.pipe.spec.ts
```diff
@@ -1,20 +1,31 @@
...
 import { EMPTY, of } from 'rxjs';
+import { MockTranslationService } from './testing/mock-translation.service';
 import { TranslatePipe } from './translate.pipe';
 import { TranslationService } from './translation.service';
-import createSpy = jasmine.createSpy;
+
+@Injectable()
+class MockChangeDetectorRef implements Partial<ChangeDetectorRef> {
+  markForCheck() {}
+}
```
Platonn, line 11: [MEDIUM]

Are changes in this file _really_ related to this PR?

projects/core/src/logger/logger.service.ts
```diff
@@ -0,0 +1,39 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { Injectable } from '@angular/core';
+
+/**
+ * By default it delegates logged messages to the native `console` object.
+ *
+ * It's a good extension point for customizing the destination of logs
+ *(e.g. to use a 3rd party logger library) or to enhance logs with more contextual data.
```
Platonn, line 13: [MINOR]

spacing
```suggestion
 * (e.g. to use a 3rd party logger library) or to enhance logs with more contextual data.
```

projects/storefrontapp/src/app/app.module.ts
```diff
@@ -49,6 +50,7 @@ if (!environment.production) {
     SpartacusModule,
     TestOutletModule, // custom usages of cxOutletRef only for e2e testing
     TestConfigModule.forRoot({ cookie: 'cxConfigE2E' }), // Injects config dynamically from e2e tests. Should be imported after other config modules.
+    ErrorHandlingModule.forRoot(), // custom error handling that delegates to LoggerService. Remove for 7.0 as it will be the part of BaseCoreModule (CXSPA-3680).
```
Platonn, line 53: [MINOR]

Let's it move before the `TestConfigModule` - its comment says _"Should be imported after other config modules."_

core-libs/setup/ssr/logger/index.ts
```diff
@@ -0,0 +1,8 @@
+/*
+ * SPDX-FileCopyrightText: 2023 SAP Spartacus team <spartacus-team@sap.com>
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+export * from './loggers';
+export * from './services';
```
Platonn, line 8: [MINOR]

TBH I'm not sure what's the difference, but we usually export from `/xxx/index`. I recall in the past it caused problems when building libraries with ng-packgr. Not sure if it's still relevant problem today
```suggestion
export * from './loggers/index';
export * from './services/index';
```

core-libs/setup/ssr/logger/loggers/express-server-logger.ts
```diff
@@ -0,0 +1,25 @@
...
+  log(message: string, context: ExpressServerLoggerContext): void;
+  warn(message: string, context: ExpressServerLoggerContext): void;
+  error(message: string, context: ExpressServerLoggerContext): void;
+  info(message: string, context: ExpressServerLoggerContext): void;
+  debug(message: string, context: ExpressServerLoggerContext): void;
+}
+
+export const EXPRESS_SERVER_LOGGER = new InjectionToken<ExpressServerLogger>(
+  'EXPRESS_SERVER_LOGGER'
+);
```
Platonn, line 25: [MINOR]

Let's add JSDocs

projects/core/src/i18n/translate.pipe.spec.ts
pawelfras: yes, this change is related to this PR. Previously TranslatePipe wasn't tested as an injectable and a new instance was created with the `new` operator. Such a way was causing an issue with the `inject` function used to inject `LoggerService`. When I provided `TestBed` to this test, I had to create a few mocks.

projects/core/src/state/utils/processes-loader/processes-loader.reducer.ts
```diff
@@ -40,6 +40,7 @@ export function processesLoaderReducer<T>(
         processesCountDiff &&
         state.processesCount + processesCountDiff < 0
       ) {
+        /* eslint-disable-next-line no-console */
```
Platonn, line 43: [MINOR]

This reducer is actually a place in which we'd like to use the LoggerService, but we just postponed it.

What about adding in a code comment a ticket number for this planned change?

Platonn:
[MINOR]

Let's remove a `console.log` from [this file](https://github.com/SAP/spartacus/blob/af826cfe37712abd727099ee0613fff179c8a895/projects/storefrontlib/cms-components/myaccount/address-book/address-form/suggested-addresses-dialog/suggested-addresses-dialog.component.ts#L64) as well - I believe it was not meant to run for customers.

Platonn:
[MINOR]

Let's enable the new logger in our storefrontapp's SSR (add this code to `server.ts`):

```ts
// server.ts
const ssrOptions: SsrOptimizationOptions = {
  /*...*/
  logger: true, // CXSPA-3680: remove this line in new major version, when the `logger` is enabled default
```

core-libs/setup/ssr/logger/loggers/default-express-server-logger.ts
```diff
@@ -0,0 +1,68 @@
...
+        ...context,
+      },
+    };
+    if (context.request) {
+      Object.assign(object.context, {
+        request: this.mapRequest(context.request),
+      });
+    }
+
+    return JSON.stringify(object);
```
Platonn, line 59: [MINOR]

Currently unit tests of this class are splitted per protected methods, but we don't test whether those protected methods are called by other methods.

Anyway, I think we should add some blackbox tests, .e.g:
I give such-and-such arguments to this class' method `log`, and I get such-and-such JSON string in return.
This would assert the shape of the output JSON - it has `message`, and a `context` with `timestamp` and `request` (with specific fields inside).

I'm especially missing a test for `timestamp`.

Platonn:
[MINOR]

Even in development mode, when I enabled `logger:true` in SsrOptimizationOptions, I could see JSONs in console that are hard to read:
image

In dev-mode it would be good to print this JSON in a pretty way:
image

ideally, the dev-mode should be recognized automatically (ideally: based on `isDevMode()` from Angular). But we need to check if it will work in the context of ExpressJS as well. At the first sight it looks risky to me.
Alternatively we could check `process.env.NODE_ENV === 'production'`. However I'm not sure if customers are setting `NODE_ENV` to `production` in their deployments (they should IMHO, but I'm not sure if it's the case by default, e.g. in CCV2 - to be checked too)

Platonn (dismissed):

Platonn:
LGTM! 🔥

Platonn (approved):

core-libs/setup/ssr/optimized-engine/optimized-ssr-engine.ts
pawelfras: My plan was to set `DefaultExpressServerLogger` as default in `defaultSsrOptimizationOptions`. Thanks to that it will be used unless we provide a custom logger. In that case, we don't need a method to initialize the logger.

core-libs/setup/ssr/optimized-engine/optimized-ssr-engine.ts
pawelfras: Since `replacer` in `JSON.stringify()` works recursively, if we put there such a condition and put an object as a value to be stringified, we will end with "Object" as an output from this method. (which is our case):
image

I'll move type guard to `express-server-logger.ts`

pawelfras: `logOpotions` method has been refactored and the type guard has been removed as it's no longer needed.

pawelfras: done in https://github.com/SAP/spartacus/pull/17538

projects/core/src/i18n/i18next/i18next-plugins/i18next-logger-plugin.spec.ts
pawelfras: done in https://github.com/SAP/spartacus/pull/17538

projects/core/src/logger/logger.service.ts
pawelfras: done in https://github.com/SAP/spartacus/pull/17538

projects/storefrontapp/src/app/app.module.ts
pawelfras: done in https://github.com/SAP/spartacus/pull/17538

core-libs/setup/ssr/logger/index.ts
pawelfras: done in https://github.com/SAP/spartacus/pull/17538

core-libs/setup/ssr/logger/loggers/express-server-logger.ts
pawelfras: done in https://github.com/SAP/spartacus/pull/17538

projects/core/src/state/utils/processes-loader/processes-loader.reducer.ts
pawelfras: done in https://github.com/SAP/spartacus/pull/17538

core-libs/setup/ssr/logger/loggers/default-express-server-logger.ts
pawelfras: done in https://github.com/SAP/spartacus/pull/17538

pawelfras:
> [MINOR]
> 
> Even in development mode, when I enabled `logger:true` in SsrOptimizationOptions, I could see JSONs in console that are hard to read: image
> 
> In dev-mode it would be good to print this JSON in a pretty way: image
> 
> ideally, the dev-mode should be recognized automatically (ideally: based on `isDevMode()` from Angular). But we need to check if it will work in the context of ExpressJS as well. At the first sight it looks risky to me. Alternatively we could check `process.env.NODE_ENV === 'production'`. However I'm not sure if customers are setting `NODE_ENV` to `production` in their deployments (they should IMHO, but I'm not sure if it's the case by default, e.g. in CCV2 - to be checked too)

done in https://github.com/SAP/spartacus/pull/17538

pawelfras:
> [MINOR]
> 
> Let's enable the new logger in our storefrontapp's SSR (add this code to `server.ts`):
> 
> ```ts
> // server.ts
> const ssrOptions: SsrOptimizationOptions = {
>   /*...*/
>   logger: true, // CXSPA-3680: remove this line in new major version, when the `logger` is enabled default
> ```

done in https://github.com/SAP/spartacus/pull/17538

core-libs/setup/ssr/logger/services/express-logger.service.ts
pawelfras: done in #17538

PR #17516, opened 2023-06-14
fix: enhance the error message when agent change customer's password

i53577:
JIRA: https://jira.tools.sap/browse/CXSPA-3399


feature-libs/user/profile/components/update-password/update-password-component.service.ts
```diff
@@ -93,7 +94,12 @@ export class UpdatePasswordComponentService {
     });
   }
 
-  protected onError(_error: Error): void {
+  protected onError(_error: HttpErrorModel): void {
```
scarai-sap: changed the signature of the routine?

feature-libs/user/profile/components/update-password/update-password-component.service.ts
```diff
@@ -93,7 +94,12 @@ export class UpdatePasswordComponentService {
...
+    if (_error.details?.[0].type === 'AccessDeniedError')
+    this.globalMessageService.add(
```
scarai-sap: code not aligned, and always use {} even we only have one line inside the if

evy-yang01: strange, it worked as expected. Indeed I missed the {}, I will add it. Nice observation

scarai-sap (dismissed):

scarai-sap (dismissed):

feature-libs/user/profile/components/update-password/update-password-component.service.ts
```diff
@@ -93,7 +94,16 @@ export class UpdatePasswordComponentService {
     });
   }
 
-  protected onError(_error: Error): void {
+  protected onError(_error: HttpErrorModel | Error): void {
```
kimhw0630: I spoke with Patrick Labadie (@plabadie ), and he mentioned that this change could potentially lead to compilation errors in the customer's custom code.  so we might need to wait for a major version

scarai-sap: if customer call onError with type Error, it should pass the compilation. Why it make potential compilation error?

PR #17530, opened 2023-06-16
fix: display confirm switch customer dialog when new customerId in the deeplink

i333055:
https://jira.tools.sap/browse/CXSPA-3380


feature-libs/asm/components/asm-main-ui/asm-main-ui.component.spec.ts
```diff
@@ -51,6 +51,10 @@ class MockUserAccountFacade implements Partial<UserAccountFacade> {
   get(): Observable<User> {
     return of({});
   }
+
+  getById(): Observable<User> {
+    return of({});
```
niehuayang: Please use the constant EMPTY instead of of( {} ).

feature-libs/asm/components/asm-main-ui/asm-main-ui.component.ts
```diff
@@ -218,6 +219,35 @@ export class AsmMainUiComponent implements OnInit, OnDestroy {
...
+            next: (switchCustomer) => {
+              this.launchDialogService.openDialogAndSubscribe(
+                LAUNCH_CALLER.ASM_SWITCH_CUSTOMER,
+                this.element,
+                { curCustomer: curCustomer, switchCustomer: switchCustomer }
+              );
+            },
+            error: (error: HttpErrorModel) => {
+              this.globalMessageService.add(
+                error.details?.[0].message ?? '',
```
niehuayang, line 240: Usually, we do not recommend using response data directly. Could you please add the translation mapping in `asm.ts`

i333055: this is keeping consistence with other scenes for global error message.

feature-libs/user/account/core/facade/user-account.service.ts
```diff
@@ -40,6 +40,13 @@ export class UserAccountService implements UserAccountFacade {
     protected query: QueryService
   ) {}
 
+  /**
+   * Returns the user according the userId
+   */
+  getById(userId: string): Observable<User | undefined> {
```
niehuayang, line 47: It is recommended to use a [Queries](https://wiki.one.int.sap/wiki/display/SAPCCWiki/Spartacus+Onboarding) here. Query has built-in caching for requests, so subsequent subscribers will not trigger additional requests.

i333055: for current customer cache reasonable，but for this case seems no need to cache，if emulated this customer，it will be cached.

i333055: no need to cache for on time to do reset and seems every time is reload

scarai-sap (approved):
LGTM

niehuayang (approved):

PR #17531, opened 2023-06-16
Fix build and introduce test command

ChristophHi:


ChristophHi:
Will work with draft instead

PR #17536, opened 2023-06-16
Replace logic to base module to make listeners work again

rmch91:
Closes: 
https://jira.tools.sap/browse/CXSPA-3717
https://jira.tools.sap/browse/CXSPA-3327
https://jira.tools.sap/browse/CXSPA-3722


integration-libs/opf/base/root/events/opf-event.listener.ts
```diff
@@ -0,0 +1,40 @@
...
+import { CreateCartEvent } from '@spartacus/cart/base/root';
+import { EventService, LoginEvent } from '@spartacus/core';
+import { OrderPlacedEvent } from '@spartacus/order/root';
+import { merge, Subscription } from 'rxjs';
+import { OpfService } from '../services';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class CheckoutOpfEventListener implements OnDestroy {
```
Matejk00: Should we still call this `CheckoutOpfEventListener` as it's not a part of the checkout folder?

integration-libs/opf/base/root/services/opf-payment-metadata-store.service.ts
```diff
@@ -0,0 +1,37 @@
...
+ */
+
+import { Injectable } from '@angular/core';
+import { BehaviorSubject, Observable } from 'rxjs';
+import { OpfPaymentMetadata, PaymentProcessingState } from '../model/opf.model';
+
+const initialState = {
+  termsAndConditionsChecked: false,
+  selectedPaymentOptionId: undefined,
+  paymentProcessingState: PaymentProcessingState.NOT_IN_PROGRESS,
```
Matejk00: Is this state only in progress or not? If so, maybe it's worth considering using a boolean value here. It would make things a bit simpler instead of creating an enum value.

integration-libs/opf/base/root/services/opf-payment-metadata-store.service.ts
```diff
@@ -0,0 +1,37 @@
...
+export class OpfPaymentMetadataStoreService {
+  opfPaymentMetadataState = new BehaviorSubject<OpfPaymentMetadata>(
+    initialState
+  );
+
+  getOpfMetadataState(): Observable<OpfPaymentMetadata> {
+    return this.opfPaymentMetadataState.asObservable();
+  }
+
+  opfMetadataUpdate(payload: Partial<OpfPaymentMetadata>): void {
```
Matejk00: I think the name "updateOpfMetadata" here would be more consistent with other methods.

Matejk00 (changes requested):
After pulling the changes, I noticed some empty folders in my IDE, such as integration-libs/opf/base/core/store. Would you mind removing all of them?

integration-libs/opf/base/root/events/opf-event.listener.ts
rmch91: Thanks, changed

integration-libs/opf/base/root/services/opf-payment-metadata-store.service.ts
rmch91: Was thinking maybe we will need some DECLINE in the future or something else, but you've right, for now it could be just boolean. Changed.

integration-libs/opf/base/root/services/opf-payment-metadata-store.service.ts
rmch91: changed

Matejk00 (approved):

PR #17538, opened 2023-06-16
feat: CXSPA-3742 - SSR logging - addresing minor comments

pawelfras:
This PR addresses minor comments pointed out during the review of [PR containing SSR contextual logging](https://github.com/SAP/spartacus/pull/17512).

In addition, it solves the invalid formatting of functions provided to optimization options. The logic responsible for mapping an object with optimization options to the object with primitives has been extracted to the function `getLoggableSsrOptimizationOptions` in a separate file `get-loggable-ssr-optimization-options.ts.`

closes CXSPA-3742


core-libs/setup/ssr/logger/loggers/express-server-logger.ts
```diff
@@ -20,6 +35,15 @@ export interface ExpressServerLogger {
...
 }
 
+/**
+ * Injection token for ExpressServerLogger used for log message in server side rendering.
+ * EXPRESS_SERVER_LOGGER is used to provide proper logger to LoggerService instance.
+ *
+ * Spartacus is providing three types of server loggers:
+ * - DefaultExpressServerLogger - default implementation used for logging contextual messages in SSR.
+ * - LegacyExpressServerLogger - used for logging if contextual logging is disabled
+ *
```
Platonn: [MEDIUM]

`three` vs `two`?

core-libs/setup/ssr/logger/loggers/default-express-server-logger.spec.ts
```diff
@@ -1,7 +1,22 @@
+import * as isDevModeFn from '@angular/core';
```
Platonn: [NITPICK]

`*` is not devModeFn, but the whole `@angular/core` library contents
```suggestion
import * as angularCore from '@angular/core';
```

core-libs/setup/ssr/logger/loggers/default-express-server-logger.ts
```diff
@@ -4,12 +4,31 @@
...
+ * @property traceFlags - traceFlags of traceparent header
+ *
+ * @see https://www.w3.org/TR/trace-context/#traceparent-header-field
+ */
+interface Traceparent {
+  version: string;
+  traceId: string;
+  spanId: string;
+  traceFlags: string;
+}
```
Platonn: [MEDIUM]

For separation of concerns, let's move to a separate folder, /w3c-trace-context:
- the interface
- the logic of extracting specific header (i.e. it's not a responsibility of a logger to know how specific header is named in w3c - it should delegate it to a specialized function)
- the logic of parsing (and validating) traceparent string into values

Btw. what renaming:
interface Traceparent -> interface W3cTraceContext
function mapTraceparent -> function parseTraceparent
?

core-libs/setup/ssr/logger/loggers/default-express-server-logger.spec.ts
```diff
@@ -1,7 +1,22 @@
...
+  res: {
+    locals: {
+      cx: {
+        request: { uuid: 'test', timeReceived: new Date('2023-05-26') },
+      },
+    },
+  },
+  headers: {
+    traceparent: '00-0af7651916cd43dd8448eb211c80319c-b9c7c989f97918e1-01',
+  },
```
Platonn: [MEDIUM]

Let's also add some test, for when the `traceparent` header is not present.

Why: Our code should run without errors, even when `traceparent` header is not present (e.g. when customer doesn't have Dynatrace / tracing agent installed). So let's cover it with unit test.

core-libs/setup/ssr/logger/loggers/default-express-server-logger.ts
```diff
@@ -56,13 +75,44 @@ export class DefaultExpressServerLogger implements ExpressServerLogger {
...
   protected mapRequest(request: Request): Record<string, any> {
-    return {
+    const mappedRequest = {
       url: request.originalUrl,
       ...request.res?.locals.cx.request,
     };
+
+    if (
+      request.headers?.traceparent &&
+      typeof request.headers?.traceparent === 'string'
```
Platonn: [NITPICK]

Do we _really_ need to check if a header's value has type `string`? I believe all HTTP header's values must be of type string.

core-libs/setup/ssr/logger/loggers/default-express-server-logger.ts
```diff
@@ -56,13 +75,44 @@ export class DefaultExpressServerLogger implements ExpressServerLogger {
...
+    ) {
+      Object.assign(mappedRequest, {
+        openTracing: this.mapTraceparent(request.headers.traceparent),
```
Platonn: [MEDIUM]

To be bullet proof (i.e. to be case-insensitive for the header name, `traceparent` vs `Traceparent`), instead of accessing `headers` object properties, let's use the function syntax `.header()` (or `.get()`). For instance `request.header('traceparent')`. For more see [Express docs](https://expressjs.com/en/api.html#req.get) and [SO thread](https://stackoverflow.com/a/60855626/11734692).

core-libs/setup/ssr/logger/loggers/default-express-server-logger.ts
```diff
@@ -4,12 +4,31 @@
...
+interface Traceparent {
+  version: string;
+  traceId: string;
+  spanId: string;
+  traceFlags: string;
+}
+
 /**
  *
  * Default logger used in SSR (ExpressJS) to enhance logs visible e.g. in monitoring tools e.g. Kibana.
```
Platonn: [MEDIUM]

Since we add a support for TraceContext, let's also mention it in the JSDoc. For inspiration:

```
 * [...]
 * which contains a "timestamp" and details of the "request" ("url", "uuid", "timeReceived")
 *
 * The output "context" JSON will contain also a property "traceContext"
 * with "traceId", "spanId", "version" and "traceFlags",
 * if only the given request has the special header "traceparent" (specifed in 
 * the "W3C TraceContext" document. See https://www.w3.org/TR/trace-context/#traceparent-header ).
```

core-libs/setup/ssr/logger/loggers/default-express-server-logger.ts
```diff
@@ -56,13 +75,44 @@ export class DefaultExpressServerLogger implements ExpressServerLogger {
...
+  /**
+   * Maps traceparent header to object with properties version, traceId, spanId and traceFlags.
+   *
+   * @param traceparent
+   * @returns Params of the traceparent header
+   *
+   * @see https://www.w3.org/TR/trace-context/#traceparent-header-field-values
+   */
+  protected mapTraceparent(traceparent: string): Traceparent {
+    const [version, traceId, spanId, traceFlags] = traceparent.split('-');
```
Platonn: [MAJOR]

To avoid runtime errors, when `traceparent` value is invalid, we should validate it first.
e.g. now a runtime error will happen, when splitted `traceparent` string has less than 4 elements, because we would try to destructure this array into 4 elements.

Note: (for customers that don't have OpenTracing configured) malicious end users might spoof the value of the `traceparent` header when they send a request. For more, see https://www.w3.org/TR/trace-context/#security-considerations

For example, we could change this function to:
```
function parseTraceparent(traceparent: string): W3cTraceContext | false {
  if (typeof traceparent !== 'string' || traceparent.length !== 55) {
    return false;
  }

  const version = traceparent.substr(0, 2);
  const traceId = traceparent.substr(3, 32);
  const parentId = traceparent.substr(36, 16);
  const traceFlags = traceparent.substr(53, 2);
  
  return { version, traceId, parentId, traceFlags };
}
```

core-libs/setup/ssr/logger/loggers/default-express-server-logger.ts
Platonn: [MEDIUM]

My bad. I've realized only now that we shouldn't mention the phrase `OpenTracing` in our code and JSdocs, but `TraceContext`. It's because TraceContext is the actual [W3C standard](https://www.w3.org/TR/trace-context). But OpenTracing is a popular open-source product that implements this standard (which was btw. renamed later to OpenTelemetry; and it's used e.g. by Dynatrace, see [dynatrace website](https://www.dynatrace.com/integrations/opentracing/)).

So I suggest to get rid of all mentions of OpenTracing in our JSDocs and properties, and replace them with more proper word: `TraceContext` or `W3cTraceContext`

core-libs/setup/ssr/optimized-engine/optimized-ssr-engine.ts
```diff
@@ -86,30 +87,20 @@ export class OptimizedSsrEngine {
...
-    const replacer = (_key: string, value: unknown): unknown => {
-      if (typeof value === 'function') {
-        return value.toString();
-      }
-      if (this.isServerLogger(value)) {
-        return value.constructor.name;
-      }
-      return value;
-    };
+    const options = ssrOptimizationOptionsResolver(this.ssrOptions);
```
Platonn: [MINOR]

The name `...Resolver` suggests something more significant than just making options more logger friendly. At the first sight I thought the `...Resolver` will transform the "insufficient" `ssrOptions` and return proper "resolved" options. But we actually need it only for the purpose of logging. What about renaming the function:
ssrOptimizationOptionsResolver -> getLoggableSsrOptimizationOptions
?

e.g.

```suggestion
    const loggableSsrOptions = getLoggableSsrOptimizationOptions(this.ssrOptions);
```

core-libs/setup/ssr/optimized-engine/optimized-ssr-engine.ts
```diff
@@ -86,30 +87,20 @@ export class OptimizedSsrEngine {
...
-        logger:
-          this.ssrOptions.logger === true
-            ? this.ssrOptions.logger
-            : this.ssrOptions.logger?.constructor?.name,
-      },
-    } as unknown as ExpressServerLoggerContext); //it expects ExpressServerLoggerContext, but the current logged message is printed a the start f the server and there is no request available yet.
+    if (this.ssrOptions.logger) {
+      this.log(`[spartacus] SSR optimization engine initialized`, true, {
+        options,
+      } as unknown as ExpressServerLoggerContext); //it expects ExpressServerLoggerContext, but the current logged message is printed a the start f the server and there is no request available yet.
```
Platonn: [MINOR]

typo
```suggestion
      } as unknown as ExpressServerLoggerContext); //it expects ExpressServerLoggerContext, but the current logged message is printed at the start of the server and there is no request available yet.
```

core-libs/setup/ssr/logger/loggers/default-express-server-logger.ts
```diff
@@ -56,13 +75,44 @@ export class DefaultExpressServerLogger implements ExpressServerLogger {
       });
     }
 
-    return JSON.stringify(object);
+    return isDevMode()
```
Platonn, line 60: [MAJOR]

_This comment requires from us doing more investigation, so I suggest let's revert this change and let's descope dev-mode improvement to a separate ticket&PR, not to delay this PR._

My gut feeling is that it's risky to use `isDevMode` outside of the context of an Angular app. So we should (1) test it well, and (2) really think twice if we don't abuse the public API of Angular.

So I'm curious how it works in practice. 
I wonder, when the actual setting of `dev mode` vs `prod mode` happens inside angular. I believe setting `prod mode` happens [here](https://github.com/SAP/spartacus/blob/9a85d093f68e74d5c9966dcde0bba44ae0d834ac/projects/storefrontapp/src/main.server.ts#L13-L15) in our `main.server.ts` in our storefrontapp. In our storefrontapp we guard this setting with a condition `environment.production` which is based on the `environment.ts` files replacable in build time (thanks to `angular.json` configuration). That said, I vaguely remember that in Angular v15 the core team no longer creates `environment` files for new apps, but uses [some different mechanism for setting production variables](https://dev.to/this-is-angular/angular-15-what-happened-to-environmentts-koh) - I didn't check it yet, so to be invesitaged. So I'm not sure how the setting of 'dev mode' vs 'prod mode' looks in modern Angular 15+ apps -> we also should check it.

I'd like us to test the following cases:
1. remove almost everything from `server.ts` and use `new DefaultEspressServerLogger()` in isolation - try out logging something with some context. Why: to be sure that `main.server.ts` doesn't influence it. Let's test it both with: prod build SSR and dev build SSR. What's the result - inline or multi-line?
2. bring back everything (including an import of AppServerModule to server.ts). Let's create manually `new DefaultEspressServerLogger()` in the top of the `server.ts` and log something very early with some context. Check prod build SSR and dev build SSR. is it multiline or single line? Then let's make request to this SSR to trigger an actual Angular rendering. What's the result inline or multi-line? Especially is the result in a different format than the early log we've put manually in the top of `server.ts`? Why to test it: I'm interested if there can be any "race condition" - e.g. if before the first render of any Angular app the `isDevMode()` would behave differently?

Note: I'm interested if any warnings thrown to console by Angular feamework about too early accessing `isDevMode` or that `dev mode` and then possibly changing this value in a `prod mode` later on, when the first rendering starts. Seeing [Angular source code of `isDevMode` from version 14.2.x](https://github.com/angular/angular/blob/61e69c87e6b577f1c5a71fb5c39c3bd6125c4b91/packages/core/src/util/is_dev_mode.ts#L30-L48), I'd expect that the value of `dev mode` cannot be changed after the first read of this value. However the [Angular source code of `isDevMode` since version 15.0.0](https://github.com/angular/angular/blob/b79ad226b2c6fc2dfaafe992facd7172fa0944c3/packages/core/src/util/is_dev_mode.ts#L20-L22) doesn't seem to have this check.
3. Let's figure out how modern ng15 applications work e.g. by generating aside a fresh ng15.0.0 angular app (since they don't have generated replacable `environment.ts` files). Then let's see how such modern applications setup a `prod mode` - do they do it explicitly as we do in our stoerfront app? Or since ng15 it's configured implicitly and internally by Anuglar CLI when building a project with "prod" or "dev" mode of the build? When we have an answer, it might be worth to repeat the experiment (2), but with storefrontap adapted to have more modern configuration (like in fresh ng15.0.0 app).

core-libs/setup/ssr/logger/loggers/default-express-server-logger.ts
Platonn: Then let's also:
- add some unit tests for handling invalid traceparent inputs
- add JSDoc for this function

core-libs/setup/ssr/logger/loggers/default-express-server-logger.ts
pawelfras: Unfortunately, we need to check whether it's a string. Headers are a type of `IncomingHttpHeaders` that extends `NodeJS.Dict<string | string[]>`. `set-cookie` can be an array: `'set-cookie'?: string[] | undefined;`

core-libs/setup/ssr/logger/loggers/default-express-server-logger.ts
pawelfras: Cool. That solves the problem mentioned above.

core-libs/setup/ssr/logger/loggers/default-express-server-logger.ts
pawelfras: As we discussed and tested, `isDevMode` can be safely used outside the Angular context.

core-libs/setup/ssr/logger/loggers/default-express-server-logger.ts
pawelfras: Will be addressed in PR for [CXSPA-2067](https://jira.tools.sap/browse/CXSPA-2067).

core-libs/setup/ssr/logger/loggers/default-express-server-logger.ts
pawelfras: Will be addressed in PR for [CXSPA-2067](https://jira.tools.sap/browse/CXSPA-2067).

core-libs/setup/ssr/logger/loggers/default-express-server-logger.ts
pawelfras: Will be addressed in PR for [CXSPA-2067](https://jira.tools.sap/browse/CXSPA-2067).

core-libs/setup/ssr/logger/loggers/default-express-server-logger.ts
pawelfras: Will be addressed in PR for [CXSPA-2067](https://jira.tools.sap/browse/CXSPA-2067).

core-libs/setup/ssr/logger/loggers/default-express-server-logger.ts
pawelfras: Will be addressed in PR for [CXSPA-2067](https://jira.tools.sap/browse/CXSPA-2067).

core-libs/setup/ssr/logger/loggers/default-express-server-logger.spec.ts
pawelfras: Will be addressed in PR for [CXSPA-2067](https://jira.tools.sap/browse/CXSPA-2067).

core-libs/setup/ssr/logger/loggers/default-express-server-logger.ts
pawelfras: Will be addressed in PR for [CXSPA-2067](https://jira.tools.sap/browse/CXSPA-2067).

core-libs/setup/ssr/logger/loggers/default-express-server-logger.ts
Platonn: For reference, I'm posting here results of the investigation of Angular's source code in version 15:

The reason that `isDevMode()` can be used safely anywhere in the code is that Angular (only since v15!) implicitly sets `ngDevMode` to false _in the build time_, when building app in prod mode (with optimizations enabled). In details:
- in `@angular/compiler-cli` the setting of `ngDevMode` to `false` is exported as a special config object `GLOBAL_DEFS_FOR_TERSER` (and as `GLOBAL_DEFS_FOR_TERSER_WITH_AOT`:
    https://github.com/angular/angular-cli/blob/8089c9388056b3caaf56f981848aca94f022da73/packages/angular_devkit/build_angular/src/tools/webpack/configs/common.ts#L296-L299
- Then `@angular-devkit/build-angular` picks this config. And when it's building an app with `optimizations: true` (in prod mode), it "hardcodes" those global constants values (`GLOBAL_DEFS_FOR_TERSER` with `ngDevMode` set to `false`), by passing those constants to special config key `"define"` of the options of `Terser` otimizer:
    https://github.com/angular/angular-cli/blob/8089c9388056b3caaf56f981848aca94f022da73/packages/angular_devkit/build_angular/src/tools/webpack/configs/common.ts#L296-L299

core-libs/setup/ssr/optimized-engine/get-loggable-ssr-optimization-options.spec.ts
```diff
@@ -0,0 +1,72 @@
...
+import { getLoggableSsrOptimizationOptions } from './get-loggable-ssr-optimization-options';
+import { SsrOptimizationOptions } from './ssr-optimization-options';
+
+class MockLogger extends DefaultExpressServerLogger {
+  constructor() {
+    super();
+  }
+}
+
+describe('SsrOptimizationOptionsResolver', () => {
```
Platonn: [NITPICK]

```suggestion
describe('getLoggableSsrOptimizationOptions', () => {
```

core-libs/setup/ssr/optimized-engine/get-loggable-ssr-optimization-options.ts
```diff
@@ -0,0 +1,29 @@
+/*
+ * SPDX-FileCopyrightText: 2023 SAP Spartacus team <spartacus-team@sap.com>
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { SsrOptimizationOptions } from './ssr-optimization-options';
+
+/**
+ * Helper functions that maps optimization options to primitive values.
```
Platonn: [NITPICK]

```suggestion
 * Helper function that maps optimization options to primitive values.
```

core-libs/setup/ssr/optimized-engine/get-loggable-ssr-optimization-options.ts
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * @param value optimization options that should be logged
+ * @returns options containing only primitive values that are easier to read by developers and monitoring tools
+ */
+export const getLoggableSsrOptimizationOptions = (
+  value: SsrOptimizationOptions
+) => {
+  const newValue: Record<string, any> = { ...value };
+  Object.keys(value).forEach((key) => {
+    if (typeof newValue[key] === 'object') {
```
Platonn: [MINOR]

To be future proof (to log properly plain objects with nested config properties), let's print object's constructor only of objects that are not just plain objects.
```suggestion
    // check for class instance, but not a plain object
    if (typeof newValue[key] === 'object' && plainObject.constructor !== Object) {
```

projects/storefrontlib/layout/main/storefront.component.ts
```diff
@@ -61,7 +61,11 @@ export class StorefrontComponent implements OnInit, OnDestroy {
     private routingService: RoutingService,
     protected elementRef: ElementRef<HTMLElement>,
     protected keyboardFocusService: KeyboardFocusService
-  ) {}
+  ) {
+    setTimeout(() => {
+      throw new Error('test');
+    }, 2000);
+  }
```
Platonn: [MEDIUM]

Let's remove it before merging
```suggestion
  ) {}
```

core-libs/setup/ssr/optimized-engine/get-loggable-ssr-optimization-options.ts
```diff
@@ -18,7 +18,10 @@ export const getLoggableSsrOptimizationOptions = (
 ) => {
   const newValue: Record<string, any> = { ...value };
   Object.keys(value).forEach((key) => {
-    if (typeof newValue[key] === 'object') {
+    if (
+      typeof newValue[key] === 'object' &&
+      newValue[key].constructor !== Object
```
Platonn: [NITPICK]

IMHO our intention (that we want to check for a class instance but not a plain object) is not that obvious for a first-come reader of this code.

For readability, can we either add a code comment or preferrably extract this condition to a separate local function in this file?
```ts
/**
 * Returns whether the given value is a class instance (not a plain object).
 */
isClassInstance(value: any): boolean {
  return typeof value === 'object' && value?.constructor !== Object;
}
```

Platonn (approved):

PR #17562, opened 2023-06-22
Feature/cxspa 3639 initiation threshold

ChristophHi:


feature-libs/quote/components/quote-actions-by-role/quote-actions-by-role.component.spec.ts
```diff
@@ -203,6 +211,88 @@ describe('QuoteActionsByRoleComponent', () => {
...
+      const allowedActions = [
+        { type: QuoteActionType.SUBMIT, isPrimary: true },
+      ];
+      const newMockQuoteWithSubmitAction: QueryState<Quote> = {
+        error: false,
+        loading: false,
+        data: {
+          ...mockQuote,
+          allowedActions: allowedActions,
+        },
```
Uli-Tiger: this 'newMockQuoteWithSubmitAction' (or similar) is declared in several tests. I would suggest to move it to the 'before' callback with default state and just modify the parts relevant for the test within the test. Such as setting the total value or actions.
declare it like below, so you do not have to deal with data possibly being undefined strict mode checks
```
let newMockQuoteWithAction: {
    loading: boolean;
    error: false | Error;
    data: Quote;
  };
```
and then in before
```
 newMockQuoteWithAction = {
      error: false,
      loading: false,
      data: {
        ...mockQuote,
      },
    };
```
so in the test it can be written as
```
      newMockQuoteWithAction.data.allowedActions = [
        { type: QuoteActionType.SUBMIT, isPrimary: true },
      ];
```

Uli-Tiger (changes requested):

Uli-Tiger (approved):

PR #17567, opened 2023-06-23
fix: better error info when can't edit password

scarai-sap:
Let's fix it in 6.3
https://jira.tools.sap/browse/CXSPA-3399


evy-yang01 (dismissed):

i53577 (dismissed):

i333055 (dismissed):

feature-libs/user/profile/components/update-password/update-password-component.service.ts
```diff
@@ -93,7 +94,16 @@ export class UpdatePasswordComponentService {
     });
   }
 
-  protected onError(_error: Error): void {
+  protected onError(_error: HttpErrorModel | Error): void {
+    if (
+      _error instanceof HttpErrorModel &&
+      _error.details?.[0].type === 'AccessDeniedError'
```
Zeyber, line 100: This is fine for the scope of the issue. But can we get other errors from the backend? Could we lay the groundwork to handle in case of other error types (eg. switch statement)?

Zeyber (changes requested):

feature-libs/user/profile/components/update-password/update-password-component.service.ts
evy-yang01: @Zeyber , thanks for reminding. For other error scenario, it has been handled by the core http-interceptors. Only  special scenario of agent modify customer's password is the missing one. And for other negative scenarios, I have tested them manually, it works well.

pawelfras: A notation with the underscore `_error` suggests this parameter is unused. Since you added some logic, the underscore should be removed.

Apart from that, changing the parameter type may be considered a breaking change which we want to avoid in minor releases. But the type has been changed to the union type that contains the current type, so IMO it won't cause any issues because the method behaves as before in the case of usage before your changes.
Nevertheless, please double-check whether such a change may cause some issues if e.g. customers have some custom implementation of this method in their application (it is possible because this method has a `protected` access modifier)

feature-libs/user/profile/components/update-password/update-password-component.service.ts
```diff
@@ -93,7 +94,16 @@ export class UpdatePasswordComponentService {
...
+    ) {
+      this.globalMessageService.add(
```
Zeyber, line 102: I might also ask that we can add a check for this case in a test so that we can block the pipeline in case someone sees this and removes it to avoid regression.

evy-yang01:
@Zeyber Please review the e2e test case, thanks

projects/storefrontapp-e2e-cypress/cypress/e2e/regression/my-account/update-password.e2e.cy.ts
```diff
@@ -87,5 +88,77 @@ describe('My Account - Update Password', () => {
...
+      });
+
+      after(() => {
+        signOutUser();
+      });
+    });
   });
 });
+
+export function startCustomerEmulation(customer, b2b = false): void {
```
niehuayang: Why do you not use the exist function `asm.tartCustomerEmulation()`?

niehuayang (changes requested):

projects/storefrontapp-e2e-cypress/cypress/e2e/regression/my-account/update-password.e2e.cy.ts
evy-yang01: since existing startcustomeremulation use different customer model which is not consistent with current standardUser        model

niehuayang: I see, How about defining a const to meet the exsit customer model. e.g:
```
const customer = {
    fullName: standardUser.registrationData.firstName  + ' ' +standardUser.registrationData.lastName,
    email: standardUser.registrationData.email,
  };
```

feature-libs/user/profile/components/update-password/update-password-component.service.ts
evy-yang01: @pawelfras, If customer have some custom implementation with override current file method, the ui issue will still be there, this is the current impact that it might be.

projects/storefrontapp-e2e-cypress/cypress/e2e/regression/my-account/update-password.e2e.cy.ts
evy-yang01: Agree

i53577:
LGTM

feature-libs/user/profile/components/update-password/update-password-component.service.ts
evy-yang01: @pawelfras  Hi Pawelfras, I have added the feature flag for the update password function, please review.

niehuayang (dismissed):

feature-libs/user/profile/components/update-password/update-password-component.service.ts
```diff
@@ -69,10 +92,17 @@ export class UpdatePasswordComponentService {
     const oldPassword = this.form.get('oldPassword')?.value;
     const newPassword = this.form.get('newPassword')?.value;
 
-    this.userPasswordService.update(oldPassword, newPassword).subscribe({
-      next: () => this.onSuccess(),
-      error: (error: Error) => this.onError(error),
-    });
+    if (this.featureConfigService?.isLevel('6.3')) {
```
Zeyber: Why don't we have this feature flag in the `onError` method rather than making a new `onNewError` method?

Zeyber (changes requested):

feature-libs/user/profile/components/update-password/update-password-component.service.ts
evy-yang01: @Zeyber ,Since it is api change, we can not use feature flag on api level.

feature-libs/user/profile/components/update-password/update-password-component.service.spec.ts
```diff
@@ -166,14 +169,41 @@ describe('UpdatePasswordComponentService', () => {
...
-      it('should not update the password', () => {
+      it('should not update the password when old password is empty', () => {
         newPassword.setValue('testpassword123');
         service.updatePassword();
         expect(userPasswordFacade.update).not.toHaveBeenCalled();
         expect(globalMessageService.add).not.toHaveBeenCalled();
         expect(routingService.go).not.toHaveBeenCalled();
         expect(authService.coreLogout).not.toHaveBeenCalled();
       });
+      it('should get access denied error message text with agent modifying customer password when feature level is above 6.3', () => {
```
niehuayang: I think adding a version description is unnecessary, could you please delete it? `when feature level is above 6.3 `

feature-libs/user/profile/components/update-password/update-password-component.service.spec.ts
```diff
@@ -166,14 +169,41 @@ describe('UpdatePasswordComponentService', () => {
...
+        oldPassword.setValue('Old123!');
+        newPassword.setValue('New123!');
+        newPasswordConfirm.setValue('New123!');
+        service.updatePassword();
+        expect(globalMessageService.add).toHaveBeenCalledWith(
+          { key: 'updatePasswordForm.accessDeniedError' },
+          GlobalMessageType.MSG_TYPE_ERROR
+        );
+      });
+      it('should not get access denied error message text with agent modifying customer password when feature level is below 6.3', () => {
```
niehuayang: I think adding a version description is unnecessary, could you please delete it? `when feature level is above 6.3 `

feature-libs/user/profile/components/update-password/update-password-component.service.ts
pawelfras: @evy-yang01 @Zeyber After rethinking the problem, union type you proposed won't cause breaking changes because this union contained previous type (and method behaves as previous if error type === `Error`). The feature flag is not needed. Sorry for the confusion.

Zeyber (approved):

PR #17576, opened 2023-06-26
fix: Need to adapt to the width of mobile webpage on CustomerList

i53577:
JIRA: https://jira.tools.sap/browse/CXSPA-3596


feature-libs/asm/components/customer-list/customer-list.component.html
```diff
@@ -294,7 +294,7 @@ <h2 id="asm-customer-list-title" class="title modal-title">
         <!-- Modal Foot -->
         <div
           class="cx-dialog-sub-header modal-header cx-dialog-foot"
-          [class.tablet-mobile]="(breakpoint$ | async) !== BREAKPOINT.md"
```
i53577, line 297: `tablet-mobile` is useless.
use `asm-mobile-pagination` instead to make sure the align direction is correct.

feature-libs/asm/components/default-asm-pagination.config.ts
```diff
@@ -8,7 +8,7 @@ import { PaginationConfig } from '@spartacus/storefront';
 
 export const defaultAsmPaginationConfig: PaginationConfig = {
   pagination: {
-    rangeCount: 4,
+    rangeCount: 2,
```
i53577, line 11: reduce the length of the pagination component.

feature-libs/asm/styles/components/_customer-emulation.component.scss
```diff
@@ -39,6 +39,7 @@
     }
     .cx-asm-uid {
       color: #556b82;
+      word-break:break-all;
```
i53577: break the line when `uid` is too long.

feature-libs/asm/styles/components/_customer-emulation.component.scss
```diff
@@ -39,6 +39,7 @@
     }
     .cx-asm-uid {
       color: #556b82;
+      word-break: break-all;
```
i53577, line 42: break the line when `uid` is too long

i333055 (approved):

PR #17584, opened 2023-06-27
Add hosted-fields payment logic

FollowTheFlo:
[CXSPA-3719](https://jira.tools.sap/browse/CXSPA-3719)
Focus on Pattern#1
Note:
OTP facade has been moved to opf/base (thus remove from opf/checkout)
'submit' facade has been added to opf/base
globalFunctionsService to register/unregister GF exposed in facade on opf/base

Upscale refs:

https://github.tools.sap/Upscale/web-app/blob/develop/src/app/components/checkout/open-payment-checkout/open-payment-checkout.component.ts
https://github.tools.sap/Upscale/web-app/blob/d3e50b3e2550dc531b339dd8d810e140eb684edd/src/app/core/global-payment-function/global-payment-function.service.ts

image


integration-libs/opf/base/core/connectors/opf-payment.adapter.ts
```diff
@@ -19,4 +21,10 @@ export abstract class OpfPaymentAdapter {
     paymentSessionId: string,
     payload: OpfPaymentVerificationPayload
   ): Observable<OpfPaymentVerificationResponse>;
+
```
Matejk00, line 24: Can we add comment to this method, just for consistency?

integration-libs/opf/base/core/facade/opf-payment.service.ts
```diff
@@ -24,18 +53,130 @@ export class OpfPaymentService implements OpfPaymentFacade {
...
+      returnPath,
+    } = payload.submitInput;
+    const currentDate = new Date();
+    const timeZoneOffset = currentDate.getTimezoneOffset();
+    const submitRequest: SubmitRequest = {
+      paymentMethod,
+      cartId,
+      additionalData,
+      channel: 'BROWSER',
+      browserInfo: {
```
Matejk00: Maybe let's delegate this mapping to some separate OPF helper method?

integration-libs/opf/base/core/facade/opf-payment.service.ts
```diff
@@ -24,18 +53,130 @@ export class OpfPaymentService implements OpfPaymentFacade {
...
+      take(1),
+      concatMap(({ value: otpKey }) =>
+        this.opfPaymentConnector.submitPayment(
+          submitRequest,
+          otpKey,
+          paymentSessionId
+        )
+      ),
+      concatMap((response: SubmitResponse) => {
+        if (
```
Matejk00: This response handling also can be a part of the separate method. :) What do you think? I just don't want to overloading the responsibility of `submitPaymentCommand` method.

integration-libs/opf/base/core/facade/opf-payment.service.ts
```diff
@@ -24,18 +53,130 @@ export class OpfPaymentService implements OpfPaymentFacade {
...
+    protected winRef: WindowRef,
+    protected opfOtpFacade: OpfOtpFacade,
+    protected activeCartFacade: ActiveCartFacade,
+    protected userIdService: UserIdService,
+    protected routingService: RoutingService,
+    protected opfOrderFacade: OpfOrderFacade,
+    protected globalMessageService: GlobalMessageService
   ) {}
 
+  defaultError: PaymentError = {
```
Matejk00: I'm thinking about separate file for this declaration, we can call it `payment-error.model.ts` and put this as an export constant. Then here what we need to do is just import such variable if needed.

integration-libs/opf/base/core/facade/opf-payment.service.ts
```diff
@@ -45,4 +186,56 @@ export class OpfPaymentService implements OpfPaymentFacade {
...
+  protected displayError(error: PaymentError | undefined): void {
+    this.globalMessageService.add(
+      {
+        key: error?.message ? error.message : this.defaultError.message,
+      },
+      GlobalMessageType.MSG_TYPE_ERROR
+    );
+  }
+
+  protected handleBadRequestError(errorType: string): string {
```
Matejk00: There are several methods related to error handling in this file. Perhaps it would be worthwhile to create a service that consolidates all these methods, allowing us to reuse them in various contexts. However, I'm unsure if we should implement this in the current pull request, as it might necessitate a broader perspective on how to address it across the entire OPF

integration-libs/opf/base/core/facade/opf-payment.service.ts
```diff
@@ -45,4 +186,56 @@ export class OpfPaymentService implements OpfPaymentFacade {
...
+    let message = this.defaultError.message;
+    switch (errorType) {
+      case 'EXPIRED':
+        message = 'opf.payment.errors.cardExpired';
+        break;
+      case 'INSUFFICENT_FUNDS':
```
Matejk00: Does it make sense here to create an enum with such values in the model file? Just for sake potential re-use in different places. It's always better also from IDE perspective to see possible options.

integration-libs/opf/checkout/components/opf-checkout-payment-wrapper/opf-checkout-payment-wrapper.component.ts
```diff
@@ -53,7 +62,28 @@ export class OpfCheckoutPaymentWrapperComponent implements OnInit, OnDestroy {
...
+          this.registerHostedFieldsGlobalFunctions(paymentSessionData);
+        },
+      })
     );
   }
+
+  protected registerHostedFieldsGlobalFunctions(
+    paymentSessionData: PaymentSessionData | Error
+  ) {
+    if (
```
Matejk00: I would prefer to create some method like `isHostedField(args)` and use it in this statement.

package.json
```diff
@@ -84,7 +84,7 @@
     "test:libs": "concurrently \"nx test core --code-coverage\" \"nx test storefrontlib --code-coverage\" \"nx test cart --code-coverage\" \"nx test organization --code-coverage\" \"nx test storefinder --code-coverage\" \"nx test smartedit --code-coverage\" \"nx test asm --code-coverage\" \"nx test qualtrics --code-coverage\" \"nx test product --code-coverage\" \"nx test product-configurator --code-coverage\" \"nx test customer-ticketing --code-coverage\" \"nx test cdc --code-coverage\" \"nx test setup --code-coverage\" \"nx test checkout --code-coverage\" \"nx test order --code-coverage\" \"nx test digital-payments --code-coverage\" \"nx test epd-visualization --code-coverage\" \"nx test pickup-in-store --code-coverage\"",
     "test:storefront:lib": "nx test storefrontlib --source-map --code-coverage",
     "test:opf:lib": "nx test opf --source-map --code-coverage",
-    "dev:ssr": "env-cmd --no-override -e dev,b2c,$SPA_ENV cross-env NODE_TLS_REJECT_UNAUTHORIZED=0 nx run storefrontapp:serve-ssr",
+    "dev:ssr": "env-cmd --no-override -e opf,b2c,$SPA_ENV cross-env NODE_TLS_REJECT_UNAUTHORIZED=0 nx run storefrontapp:serve-ssr",
```
Matejk00: I think this is leftover ;) but we can create additional script for the OPF SSR.

Matejk00 (changes requested):

integration-libs/opf/base/core/facade/opf-payment.service.ts
FollowTheFlo: good idea

integration-libs/opf/base/core/facade/opf-payment.service.ts
FollowTheFlo: yes,I though about it as it is stuffing the file. It will in root folder so VerificationPage can use it too.

package.json
FollowTheFlo: adding dev:opf:ssr cmd

integration-libs/opf/base/core/facade/opf-payment.service.ts
FollowTheFlo: creating a opf-payment-error-handler.service.ts in opf/Base core/services folder, that will be use for the 2 other sub-patterns. I think no need to be in root as it is used within opf/Base facades

integration-libs/opf/base/core/facade/opf-payment.service.ts
FollowTheFlo: done, btw payment errors topic is still in discussion with Richard/bamboo. 
It looks like we might only need to handle level1 errors (success, rejected and cancel). Those above authorization errors belong to level2, better to keep them for now. We'll remove them when confirmed by Richard/Bamboo.

Matejk00 (approved):

PR #17588, opened 2023-06-28
Enable configurator read-only view for quote display

Larisa-Staroverova:


feature-libs/product-configurator/common/components/configure-cart-entry/configure-cart-entry.component.ts
```diff
@@ -31,14 +31,18 @@ export class ConfigureCartEntryComponent {
...
-    return this.cartEntry.orderCode !== undefined
-      ? CommonConfigurator.OwnerType.ORDER_ENTRY
-      : CommonConfigurator.OwnerType.CART_ENTRY;
+    if (this.cartEntry.quoteCode !== undefined) {
+      return CommonConfigurator.OwnerType.QUOTE_ENTRY;
+    } else {
+      return this.cartEntry.orderCode !== undefined
+        ? CommonConfigurator.OwnerType.ORDER_ENTRY
+        : CommonConfigurator.OwnerType.CART_ENTRY;
+    }
```
Uli-Tiger, line 45: so if there is a quoteCode and orderCode, we still assume QUOTE_ENTRY. Is this desired and if yes a test for this case would be nice

feature-libs/product-configurator/common/components/configure-cart-entry/configure-cart-entry.component.ts
```diff
@@ -53,14 +57,21 @@ export class ConfigureCartEntryComponent {
...
       : entryNumber.toString();
   }
 
+  protected getCode(): string | undefined {
+    if (this.cartEntry.quoteCode) {
+      return this.cartEntry.quoteCode;
+    } else if (this.cartEntry.orderCode) {
+      return this.cartEntry.orderCode;
+    }
+    return undefined;
```
Uli-Tiger: same as for OWNER_TYPE: is it desired that we return the quoteCode if both  quote  and order code are defined? If yes a test for this case would be good.

feature-libs/product-configurator/common/components/configure-cart-entry/configure-cart-entry.component.ts
Uli-Tiger: so if there is a quoteCode and orderCode, we still assume QUOTE_ENTRY. Is this desired and if yes a test for this case would be nice

feature-libs/quote/tsconfig.lib.json
Uli-Tiger, line 1: changes are intended?

Uli-Tiger (changes requested):
some remarks based on a first look - need to check in more detail tomorrow

Uli-Tiger:
@ChristophHi 
  FAIL  src/shared/utils/graph-utils_spec.ts
  ● Graph utils › library dependency graph › should have generated the correct order

feature-libs/quote/tsconfig.lib.json
ChristophHi: @Uli-Tiger Yes those changes are intended. Somehow a dependency to product-configurator was introduced before (done by the teams that dealt with that before) which does not make sense. Removing it clears a number of other dependencies as well)

feature-libs/product-configurator/common/components/configure-cart-entry/configure-cart-entry.component.ts
ChristophHi: Agree will improve that

ChristophHi:
> @ChristophHi FAIL src/shared/utils/graph-utils_spec.ts ● Graph utils › library dependency graph › should have generated the correct order

Probably also because I resolved the dependencies between quote and product-configurator

feature-libs/product-configurator/common/components/configure-cart-entry/configure-cart-entry.component.ts
```diff
@@ -31,14 +33,24 @@ export class ConfigureCartEntryComponent {
...
    * @returns - an owner type
    */
   getOwnerType(): CommonConfigurator.OwnerType {
-    return this.cartEntry.orderCode !== undefined
-      ? CommonConfigurator.OwnerType.ORDER_ENTRY
-      : CommonConfigurator.OwnerType.CART_ENTRY;
+    if (this.cartEntry.quoteCode || this.cartEntry.orderCode) {
+      if (!this.cartEntry.quoteCode) {
+        return CommonConfigurator.OwnerType.ORDER_ENTRY;
+      }
```
steinsebastian, line 44: Why is the if-statement negative?
Why not using:
      if (this.cartEntry.quoteCode) {
        return CommonConfigurator.OwnerType.QUOTE_ENTRY;
      }

Uli-Tiger (approved):

feature-libs/product-configurator/common/components/configure-cart-entry/configure-cart-entry.component.ts
ChristophHi: @steinsebastian Please see the line 41 above, and also consider that we throw an error if both codes are provided (which is an illegal state)

PR #17595, opened 2023-06-30
fix: 'this' is undefined in ProductEffects.productLoadEffect [port to 6.3.x]

Platonn:
(port of https://github.com/SAP/spartacus/pull/17594 to 6.3.x)

Previous behavior: When /products endpoint returned a http error, the code broke in [this line](https://github.com/SAP/spartacus/blob/ed1e1a78c488b1e1214491ffa736612287f8cf70/projects/core/src/product/store/effects/product.effect.ts#L77), complaining that this is undefined.

Fix: Preserve the context of this which was lost in [this line](https://github.com/SAP/spartacus/blob/ed1e1a78c488b1e1214491ffa736612287f8cf70/projects/core/src/product/store/effects/product.effect.ts#L52)

The problem was revealed only after we implemented [CXSPA-2251](https://jira.tools.sap/browse/CXSPA-2251) where we referenced this by adding this.logger to the method ProductEffects.productLoadEffect

fixes https://jira.tools.sap/browse/CXSPA-3902


kpawelczak (approved):

querplex-gw:
When will this fix be part of 6.3.x or 6.4.x?

PR #17596, opened 2023-06-30
CXSPA-3856 force refresh for quote details

ChristophHi:


feature-libs/quote/components/details/overview/quote-details-overview.component.spec.ts
```diff
@@ -143,4 +145,36 @@ describe('QuoteDetailsOverviewComponent', () => {
       expect(result).toEqual(expected);
     });
   });
+
+  describe('getTotalPrice', () => {
+    it('should return the total price formatted value in case it is available', () => {
+      expect(component.getTotalPrice(mockQuote)).toBe(totalPriceFormattedValue);
+    });
```
Stofftiger, line 148:
```suggestion
    }); 
    /* since a  '0' value is treated as if its an undefined value the totalPrice is returned instead of the  previousEstimatedTotal. 
    
    previousEstimatedTotal: {
   currencyIso: 'USD',
    formattedValue: '$0.00',
    value: 0,
   } 
   */
```

PR #17597, opened 2023-06-30
feat: CXSPA-2067 SSR Logging - Trace Context

pawelfras:
Introducing the Trace Context extension, a significant advancement in server-side rendering (SSR) logging. This feature elevates contextual logging by seamlessly incorporating Trace Context parameters into SSR logs.

For detailed information on Trace Context, please refer to the official documentation at https://www.w3.org/TR/trace-context/#overview.

With this pull request, we are now bridging the power of Trace Context to your SSR logs. Our utilities seamlessly map headers to trace context parameters, enabling you to effortlessly expand the context of your SSR logs. This enhancement opens up new possibilities, allowing you to effortlessly connect logs from popular monitoring tools like Kibana to distributed trace presentation tools like Dynatrace.

The following is an example of a log message decorated with the trace context:
```
{
  "message": "Rendering completed (/electronics-spa/en/USD/)",
  "context": {
    "timestamp": "2023-09-13T12:14:21.377Z",
    "request": {
      "url": "/electronics-spa/en/USD/",
      "uuid": "a076a5ba-7889-4c19-840e-395e89fde4b5",
      "timeReceived": "2023-09-13T12:14:13.450Z",
      "traceContext": {
        "version": "00",
        "traceId": "0af7651916cd43dd8448eb211c80319c",
        "parentId": "b7ad6b7169203331",
        "traceFlags": "01"
      }
    }
  }
}
```

closes [CXSPA-2067](https://jira.tools.sap/browse/CXSPA-2067)


core-libs/setup/ssr/optimized-engine/optimized-ssr-engine.spec.ts
```diff
@@ -1175,9 +1176,120 @@ describe('OptimizedSsrEngine', () => {
...
+    const headers: Record<string, string> = {
+      traceparent: '00-d745f6735b44e81c0ae5410cb1fc8a0c-1b527c3828976b39-01',
+    };
+    const request = {
+      originalUrl: 'test',
+      headers,
+      get: (header: string): string | string[] | null | undefined => {
+        return headers[header];
+      },
+    } as unknown as Request;
```
Platonn: Let's move `headers` and `request` to `beforeEach` block.

Why: unit tests shouldn't share a mutable object. If the function under test by any chance mutates this shared object (e.g. due to a bug in the function's implementation), then it might cause unwanted interference in between the separate tests, in worst case causing a false positive result for some of them.

core-libs/setup/ssr/optimized-engine/optimized-ssr-engine.ts
```diff
@@ -485,6 +484,32 @@ export class OptimizedSsrEngine {
     });
   }
 
+  /**
+   * Returns the request context object, which is used to enrich the logs.
+   * It contains the request's UUID, timestamp and trace context (if available).
```
```suggestion
   * It contains the random request's UUID, time of receiving the request and the W3C TraceContext (if available).
```

core-libs/setup/ssr/optimized-engine/optimized-ssr-engine.ts
```diff
@@ -491,21 +489,49 @@ export class OptimizedSsrEngine {
    * the "W3C TraceContext" document. See https://www.w3.org/TR/trace-context/#traceparent-header
    * for more details.
    * @param request - the request object
-   * @returns the context of the request
+   * @returns the context of the request and error if occurred during parsing traceparent header
    * @private
    */
-  private getRequestContext(request: Request): ExpressServerLoggerContext {
+  private getRequestContext(
```
Platonn: OptimizedSsrEngine class already has a lot of responsibilities and low cohesion. IMHO it would be nice to extract out to other class the logic of getting request's context and parsing traceparent, and handling errors related to it.

If I understood correct during our offline conversation, Levent proposed extracting it to configurable so-called... <details><summary>transformers</summary>

image</details>?

What do you think?

core-libs/setup/ssr/optimized-engine/optimized-ssr-engine.ts
```diff
@@ -491,21 +489,49 @@ export class OptimizedSsrEngine {
...
+   * @private
+   */
+  private updateWithRequestContext(options: any) {
+    const [requestContext, error] = this.getRequestContext(options.req);
+    options.req.res.locals = {
+      cx: { request: requestContext },
+    };
+    if (error) {
+      this.logger.error(error.message, { request: options.req });
+    }
```
Platonn: Did you consider passing `request` instead of all `options`?
I can see we update onlye the Response's `locals` property, but not `options`.

Calling `updateWithRequestContext(options)` sounds to me like we're updating those options, which is ambigious and not precise. I have hard time proposing another name for the method. But I think it suficess to pass as an argument `request` instead of options.

core-libs/setup/ssr/optimized-engine/optimized-ssr-engine.ts
```diff
@@ -491,21 +489,49 @@ export class OptimizedSsrEngine {
...
+      }
+    } catch (e) {
+      error =
+        e instanceof Error
+          ? e
+          : new Error('Unexpected error during parsing traceparent header');
     }
 
-    return requestContext;
+    return [requestContext, error];
```
Platonn: Did you consider rethrowing the error from this function instead of returning a touple type?
IMHO having custom a touple type here looks quite strange, while try/catch would suffice.
What do you think?

core-libs/setup/ssr/logger/loggers/w3c-trace-context/parse-traceparent.ts
```diff
@@ -0,0 +1,45 @@
...
+ * @returns Params of the traceparent header
+ *
+ * @see https://www.w3.org/TR/trace-context/#traceparent-header-field-values
+ */
+export function parseTraceparent(
+  traceparent: string | undefined | null
+): W3cTraceContext | null {
+  if (typeof traceparent !== 'string') {
+    return null;
+  }
```
Platonn: For consistency, did you consider throwing an error e.g. `InvalidTraceparentTypeError` instead of returning null?

IMHO it looks more robust to have return type of `parseTransparent` only `W3cTraceContext` (without  `... | null`).

What do you think?

core-libs/setup/ssr/optimized-engine/optimized-ssr-engine.ts
```diff
@@ -485,6 +482,56 @@ export class OptimizedSsrEngine {
...
+   * @param options - the options of the request
+   * @private
+   */
+  private updateWithRequestContext(options: any) {
+    const [requestContext, error] = this.getRequestContext(options.req);
+    options.req.res.locals = {
+      cx: { request: requestContext },
+    };
+    if (error) {
+      this.logger.error(error.message, { request: options.req });
```
Platonn: Do you think we should output the whole native `Request` object of ExpressJS in this case (which containes a lot of [properties and methods](https://expressjs.com/en/api.html#req))? Did you consider any other alternatives? e.g. to output only the usual minimal information (without `traceparent`).

core-libs/setup/ssr/optimized-engine/optimized-ssr-engine.ts
pawelfras: As I mentioned on Slack, I am not in favour of keeping this logic in OptimizedSsrEngine, however, I added it here after our conversation and the suggestion that parsing should happen once per request because `traceparent` does not change within the same request. However, I think that introducing even more complexity into SsrOptimizedEngine is worse than parsing this header before every log (or maybe there is another way to do this).

PS. Separating the logic into a separate class and the concept of "transformers" was presented in this PoC: https://github.com/SAP/spartacus/pull/17788
Let's discuss whether we want to go that way before I start working on the rest of the comments.

core-libs/setup/ssr/optimized-engine/optimized-ssr-engine.ts
pawelfras: I passed the whole request, but under the hood, the logger is using only values from `locals` and `request.originalUrl`. The decision we made was to pass the whole request to the logger as customers may want to log more values from this object.

core-libs/setup/ssr/logger/loggers/w3c-trace-context/parse-traceparent.ts
pawelfras: What about the case where `traceparent` is `null` or  `undefined`? I don't want to throw an error if the value is missing. Because an input type can be `string | null | undefined`, the `if statement` mentioned above seems to be a type-safe condition that checks whether `traceparent` is defined or not.

core-libs/setup/ssr/optimized-engine/optimized-ssr-engine.ts
pawelfras: In addition, I'm not sure what you meant by the usual minimal information, but I believe it's worth logging the context or request that has a faulty `traceparent` header to help localize the issue.

Platonn: I'm sorry, my bad. I somehow read by mistake that we're doing `console.error(....)` and print the whole Request object to stdout. But now you helped me realize it's `logger.error(....)` - which performs a proper transformation (resulting in minimal but not full information about the Request in stdout).

core-libs/setup/ssr/optimized-engine/request-context.ts
```diff
@@ -0,0 +1,56 @@
...
+  export interface Request {
+    res: {
+      locals: {
+        cx: {
+          request: ExpressServerLoggerContext;
+        };
+      };
+    };
+  }
+}
```
Platonn: nice!

core-libs/setup/ssr/optimized-engine/request-context.ts
```diff
@@ -0,0 +1,56 @@
...
+    const error =
+      e instanceof Error
+        ? e
+        : new Error('Unexpected error during parsing traceparent header');
+
+    logger.error(error.message, {
+      request: {
+        ...request,
+        res: { locals: { cx: { request: requestContext } } },
+      } as Request,
```
Platonn: 1. At the first sight it looks strange to me that we have this logic here:
```ts
res: { locals: { cx: { request: requestContext } } },
```
and similar logic in the OptimizedSsrEngine:
```ts
options.req.res.locals = {
  cx: { request: getRequestContext(options.req, this.logger) },
};
```

Shouldn't this implementation detail be defined in just one place?

2. we're replacing the whole `res` object. Shouldn't we spread also the nested levels?
```ts
request: {
  ...request,
res: { 
  ...request.res,
  locals: { 
    ...request.res.locals,
    cx: { 
      ...request.res.locals.cx,  // please note that other features might also contribute to this area "request.res.locals.cx"
      request: requestContext
    }
  }
}
```

core-libs/setup/ssr/optimized-engine/request-context.ts
```diff
@@ -9,44 +9,73 @@ import { Request } from 'express';
...
+const setRequestContext = (
+  request: Request,
+  context: ExpressServerLoggerContext
+) => {
+  request.res.locals = {
+    cx: {
+      request: context,
+    },
+  };
+};
```
Platonn: Let's not overwrite the whole `locals` object. Other features might also contributes to it. Same holds for the `cx` object - other features might contribute to it.

Let's do safe spreading on each level.

core-libs/setup/ssr/optimized-engine/request-context.ts
```diff
@@ -9,44 +9,73 @@ import { Request } from 'express';
...
   const requestContext: ExpressServerLoggerContext = {
     uuid: randomUUID(),
     timeReceived: new Date().toISOString(),
   };
+  setRequestContext(request, requestContext);
+};
 
+/**
+ * Parses the `traceparent` header and adds the trace context to the request context.
+ * In case of an error, the error is logged wit the initial request context.
```
Platonn: typo `with`

core-libs/setup/ssr/optimized-engine/optimized-ssr-engine.ts
```diff
@@ -485,6 +481,56 @@ export class OptimizedSsrEngine {
...
+  //  */
+  // private updateWithRequestContext(options: any) {
+  //   const [requestContext, error] = this.getRequestContext(options.req);
+  //   options.req.res.locals = {
+  //     cx: { request: requestContext },
+  //   };
+  //   if (error) {
+  //     this.logger.error(error.message, { request: options.req });
+  //   }
+  // }
```
Platonn: let's remove the commented out code

core-libs/setup/ssr/optimized-engine/request-context.ts
```diff
@@ -0,0 +1,91 @@
...
+/**
+ * Prepares and updates a request with the context object, which is used to enrich the logs.
+ * It contains the random request's UUID, time of receiving the context and the W3C Trace Context (if available).
+ * The trace context is parsed from the `traceparent` header, which is specified in
+ * the "W3C TraceContext" document. See https://www.w3.org/TR/trace-context/#traceparent-header
+ * for more details.
+ * @param request - the request object
+ * @param logger - the ExpressServerLogger object. It is used to log the error if occurred during parsing traceparent header
+ * @returns the context of the request and error if occurred during parsing traceparent header
+ */
```
Platonn: Nice docs!

core-libs/setup/ssr/logger/loggers/express-server-logger.ts
```diff
@@ -6,13 +6,15 @@
...
 import { Request } from 'express';
+import { W3cTraceContext } from './w3c-trace-context/w3c-trace-context.model';
 
 /**
  * ExpressServerLoggerContext is used for log message in server side rendering.
  * It contains optional request object and additional properties that can be used in log message.
  */
 export interface ExpressServerLoggerContext {
   request?: Request;
+  traceContext?: W3cTraceContext;
```
Platonn: Is this addition really needed? Don't we pass only the express `Request` as a param to the ExpressServerLogger? e.g. `logger.log('some message', { request })`?

pawelfras: No, we're passing the whole `ExpressServerLoggerContext` object so I thought it would be nice to extend this interface with `traceContext` as it will be added by default if the header exists.
```
log(message: string, context: ExpressServerLoggerContext): void {
    /* eslint-disable-next-line no-console */
    console.log(this.stringifyWithContext(message, context));
  }
```

core-libs/setup/ssr/optimized-engine/request-context.ts
pawelfras: Unfortunately, I had to change this part a bit and augment only the `Locals` interface as augmenting such a high level interface which was `Request` was causing issues during the build due to the fact that it was forcing the different shape of the `res` field.

core-libs/setup/ssr/optimized-engine/request-context.ts
pawelfras: Unfortunately, I had to change this part a bit and augment only the Locals interface as augmenting such a high level interface which was Request was causing issues during the build due to the fact that it was forcing the different shape of the res field.

core-libs/setup/ssr/optimized-engine/request-context.ts
```diff
@@ -9,19 +9,23 @@ import { Request } from 'express';
 import { ExpressServerLogger, ExpressServerLoggerContext } from '../logger';
 import { parseTraceparent } from '../logger/loggers/w3c-trace-context/parse-traceparent';
 
-const getRequestContext = (request: Request) => {
-  return request.res.locals.cx.request;
+const getRequestContext = (request: Request): ExpressServerLoggerContext => {
+  return request.res?.locals.cx.request;
```
Platonn: let's add `?.` on more levels.
```ts
return request.res?.locals?.cx?.request;
```

Why: theoretically it may happen that `cx` object is not defined YET. Then accessing `cx.request` may throw a runtime error

pawelfras: Good point, thanks!

Platonn (dismissed):

core-libs/setup/ssr/optimized-engine/request-context.ts
```diff
@@ -0,0 +1,111 @@
...
+        ...request.res.locals.cx,
+        request: context,
+      },
+    };
+  }
+};
+
+/**
+ * Adds the initial request context to the request object.
+ * @param request - the request object
```
Platonn:
```suggestion
 * Creates the initial request context to the request object.
 * @param request - the request object
 * @returns object with a randomly generated UUID and the current time
```

core-libs/setup/ssr/optimized-engine/request-context.ts
```diff
@@ -0,0 +1,111 @@
...
+  const requestContext: RequestContext = {
+    uuid: randomUUID(),
+    timeReceived: new Date().toISOString(),
+  };
+  return requestContext;
+};
+
+/**
+ * Parses the `traceparent` header and adds the trace context to the request context.
+ * In case of an error, the error is logged with the initial request context.
```
Platonn:
```suggestion
 * Parses the `traceparent` header and returns an object with the W3C TraceContext.
 * In case when the `traceparent` header is absent or invalid, `undefined` value is returned.
 * Only when the `traceparent` header is invalid, additionally an error is logged with the context of the original request and the original raw value of the `traceparent` header.
```

Platonn (dismissed):

Platonn (approved):

PR #17605, opened 2023-07-05
CXINT- 2038 : segment reference

anjana-bl:

## General

- [x] Epic has sufficient end to end (cypress) test coverage
- [x] Schematics code deprecations have been provided and tested
- [x] Epic CSS supports directionality (if required)
- [ ] Epic renders successfully in SSR

## Audits/reviews

- [ ] Architecture review of the epic
- [ ] Security and threat modeling audit/review from the Security team
- [x] Accessibility audit/review from the UX team

## New Library

## Sample data

- [x] Sample data has been added to the sampledata repo and merged to the next version to be released
- [x] New sample data has been loaded and tested on a dev server
- [x] New sample data has been loaded and tested in our CI server

## Documentation

- [x] Documentation input is provided for it to be published on the doc site.


.env-cmdrc
```diff
@@ -34,6 +34,10 @@
     "CX_B2B": "true",
     "CX_CPQ": "true"
   },
+  "segment-refs": {
+    "CX_BASE_URL": "https://api.cg79x9wuu9-eccommerc1-p7-public.model-t.myhybris.cloud",
```
vigneshkrish8691, line 38: Is it required to checkin our ccv2 api url?

anjana-bl: this file is not shipped to customers & for internal purpose only. Since Personalization is configured in P7 system, if we run SPA_ENV-segment-refs npm start - it will automatically take P7 instead of the "https://40.76.109.9:9002" to test this feature

vigneshkrish8691 (dismissed):

integration-libs/segment-refs/schematics/add-segment-refs/__snapshots__/index_spec.ts.snap
```diff
@@ -0,0 +1,29 @@
...
+"import { NgModule } from '@angular/core';
+import { SegmentRefsModule } from "@spartacus/segment-refs";
+
+@NgModule({
+  declarations: [],
+  imports: [
+    SegmentRefsModule
+  ]
+})
+export class SegmentRefsFeatureModule { }
```
Platonn: For consistency with other features, let's import statically the RootModule in the FeatureModule, e.g. in this case `SegmentRefsRootModule`.

`SegmentRefsModule` would be used only for lazy loading, however we have nothing to lazy load now, so let's remove `SegmentRefsModule` (usage and source code file) and let's leave only `SegmentRefsRootModule` imported here

integration-libs/segment-refs/segment-refs.module.ts
```diff
@@ -0,0 +1,12 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { NgModule } from '@angular/core';
+import { SegmentRefsRootModule } from './root/segment-refs-root.module';
+@NgModule({
+  imports: [SegmentRefsRootModule],
+})
+export class SegmentRefsModule {}
```
Platonn: As noted in another comment, we can remove this module.

projects/schematics/src/shared/lib-configs/integration-libs/segment-refs-schematics-config.ts
```diff
@@ -0,0 +1,44 @@
...
+    mainScope: SPARTACUS_SEGMENT_REFS,
+  },
+  folderName: SEGMENT_REFS_FOLDER_NAME,
+  moduleName: SEGMENT_REFS_MODULE_NAME,
+  featureModule: {
+    name: SEGMENT_REFS_MODULE,
+    importPath: SPARTACUS_SEGMENT_REFS,
+  },
+  rootModule: {
+    importPath: SPARTACUS_SEGMENT_REFS, //as SPARTACUS_SEGMENT_REFS_ROOT is same as SPARTACUS_SEGMENT_REFS, no extra module/features are present
```
Platonn: Let's reflect the changes from above comment in the schematics configuration as well

integration-libs/segment-refs/root/feature-name.ts
```diff
@@ -0,0 +1,7 @@
+/*
+ * SPDX-FileCopyrightText: 2023 SAP Spartacus team <spartacus-team@sap.com>
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+export const SEGMENT_FEATURE = 'Segment-Refs';
```
Platonn: typo? `SEGMENT_FEATURE` vs `SEGMENT_REFS_FEATURE`?

projects/storefrontapp/src/app/spartacus/features/segment-refs/segment-refs-feature.module.ts
```diff
@@ -0,0 +1,13 @@
...
+ * SPDX-FileCopyrightText: 2023 SAP Spartacus team <spartacus-team@sap.com>
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { NgModule } from '@angular/core';
+import { SegmentRefsModule } from '@spartacus/segment-refs';
+
+@NgModule({
+  imports: [SegmentRefsModule],
```
Platonn: According to above comments, let's use `SegmentRefsRootModule` here

projects/schematics/src/shared/lib-configs/integration-libs/segment-refs-schematics-config.ts
```diff
@@ -0,0 +1,44 @@
...
+    name: SEGMENT_REFS_MODULE,
+    content: `${SEGMENT_REFS_MODULE}`,
+  },
+  dependencyFeatures: [TRACKING_PERSONALIZATION_FEATURE_NAME],
```
Platonn, line 39: Can you help me understand, how this feature depends on `TRACKING_PERSONALIZATION_FEATURE_NAME`?
I couldn't find any imports  from `@spartacus/tracking` library in the source code of your new feature. Is there any implicit interference in between segment-refs and tracking/personalization?

integration-libs/segment-refs/root/feature-name.ts
anjana-bl: typo - will change it

projects/schematics/src/shared/lib-configs/integration-libs/segment-refs-schematics-config.ts
Platonn: please add a code comment explaining why we have to add this dependency in schematics configuration (e.g. what would be consequences when personalization feature is not installed)

also IMHO it's worth adding some JSDoc for OccSegmentRefsInterceptor, if it anyhow depends on personalization feature/ personalization interceptor - and explain how they interfere with each other

integration-libs/segment-refs/root/http-interceptors/occ-segment-refs.interceptor.ts
```diff
@@ -0,0 +1,70 @@
...
+      );
+    }
+    if (this.winRef.isBrowser()) {
+      if (!this.config.segmentRefs?.httpHeaderName && isDevMode()) {
+        this.logger.warn(`There is no httpHeaderName configured in Segment`);
+      }
+      this.requestHeader =
+        this.config.segmentRefs?.httpHeaderName?.toLowerCase();
+    }
+  }
```
Platonn: For easier extensibility for customers, we should never store any logic in constructor body - because it's not extendable/replacable.
Let's move the whole body to a protected method `initialize()` and then call it inside the constructor

integration-libs/segment-refs/root/http-interceptors/occ-segment-refs.interceptor.ts
```diff
@@ -0,0 +1,70 @@
...
+  protected logger = inject(LoggerService);
+
+  constructor(
+    protected config: SegmentRefsConfig,
+    protected occEndpoints: OccEndpointsService,
+    protected winRef: WindowRef
+  ) {
+    const url = this.winRef.location.href ?? '';
+    const queryParams = new URLSearchParams(url.substring(url.indexOf('?')));
+    this.segmentRefs = queryParams.get('segmentrefs');
```
Platonn: For easier possible extensibility, let's move this query param name to a `protected readonly` property

integration-libs/segment-refs/root/http-interceptors/occ-segment-refs.interceptor.ts
```diff
@@ -0,0 +1,70 @@
...
+  HttpInterceptor,
+  HttpRequest,
+} from '@angular/common/http';
+import { Injectable, inject, isDevMode } from '@angular/core';
+import { Observable } from 'rxjs';
+import { LoggerService, OccEndpointsService, WindowRef } from '@spartacus/core';
+import { SegmentRefsConfig } from '../config/segment-refs-config';
+
+@Injectable({ providedIn: 'root' })
+export class OccSegmentRefsInterceptor implements HttpInterceptor {
```
Platonn, line 19: The behavior of this interceptor and the motivation for such a behavior is not trivial. Let's add JSDocs explaining why we need such a behavior in various user journey types:
- when user opens website via link from email with segmentRef 
- then during navigations inside SPA, the query param is lost
- but the query param is remembered in local storage

That said, I have a few questions: 
- when should we forget the remembered segment ref? Shouldn't we clear the segment ref after some time? Or on on some events, e.g. on user logout? (suposing many users of the same machine can use the same storefront)
- how is it compliant with e.g. GDPR in terms of tracking consents? Shouldn't we first check consents before passing the segmentRef value in a header to further occ calls?

A question from SSR and SEO area:
- should we server-side render (and cache in CDN and index in Google) urls with segmentRef query param? Example end-user scenario: I got an email with link to a promo offer with segmentRef query param. I post this link on social media, so my friends can benefit from the special offer too. Then the URL is picked by search engine crawler and indexed e.g. in Google. | My question is: should we render the "personalized" version of that page (e.g. with special price) according to segmentRef in the query param? I guess so. But I'm just double checking if I understand correctly.

integration-libs/segment-refs/schematics/add-segment-refs/__snapshots__/index_spec.ts.snap
anjana-bl: done

integration-libs/segment-refs/segment-refs.module.ts
anjana-bl: done

projects/schematics/src/shared/lib-configs/integration-libs/segment-refs-schematics-config.ts
anjana-bl: done

projects/storefrontapp/src/app/spartacus/features/segment-refs/segment-refs-feature.module.ts
anjana-bl: done

projects/schematics/src/shared/lib-configs/integration-libs/segment-refs-schematics-config.ts
anjana-bl: Backend requires segment-ref along with personalization. From spartacus perspective, they require storefront to work with Personalization alone or Personalization+SegmentRefs. Hence added this dependency. I have added comments in this file explaining the same.

integration-libs/segment-refs/root/http-interceptors/occ-segment-refs.interceptor.ts
anjana-bl: done

integration-libs/segment-refs/root/http-interceptors/occ-segment-refs.interceptor.ts
```diff
@@ -0,0 +1,70 @@
...
+      this.segmentRefs = this.winRef.localStorage?.getItem(
+        this.SEGMENT_REFS_KEY
+      );
+    }
+    if (this.winRef.isBrowser()) {
+      if (!this.config.segmentRefs?.httpHeaderName && isDevMode()) {
+        this.logger.warn(`There is no httpHeaderName configured in Segment`);
+      }
+      this.requestHeader =
+        this.config.segmentRefs?.httpHeaderName?.toLowerCase();
```
Platonn: When `this.config.segmentRefs?.httpHeaderName` is undefined, then .`?.toLowerCase` is also undefined and NOT a function, so you cannot call it with `()`. To avoid runtime error `toLowerCase is not a function`, you need to also add `.?` before function call `()` 
```suggestion
        this.config.segmentRefs?.httpHeaderName?.toLowerCase?.();
```

integration-libs/segment-refs/root/http-interceptors/occ-segment-refs.interceptor.ts
Platonn: Now I've read the implementation of this interceptor carefully, and as far as I can understand it, it will take any effect only in the browser, but not in SSR. 
Why:
- the name of the `this.requestHeader` is set only in the browser:
  ```ts
    if (this.winRef.isBrowser()) {
      /* ... */
      this.requestHeader =
        this.config.segmentRefs?.httpHeaderName?.toLowerCase();
    }
  ```
- if `this.requestHeader` is not defined, then the method `intercept()` doesn't append any header to the request:
  ```ts
  intercept(
    request: HttpRequest<any>,
    next: HttpHandler
    if (
      this.requestHeader && // <------------------------------ this condition
      this.segmentRefs &&
      request.url.includes(this.occEndpoints.getBaseUrl())
    ) {
      request = request.clone({
        setHeaders: {
          [this.requestHeader]: this.segmentRefs,
        },
      });
    }
    return next.handle(request);
  }
  ```
  
So I understand that if in SSR the query param is given, e.g. user requests `mystore.com/?segmentrefs=123`, then the content of the server-side rendered page will not take this segmentrefs into account, because in SSR the outgoing calls made to OCC won't have contain the segmentRefs-specific http header.

This was not obvious for me at the first sight when reading this code. You might consider adding a check `isBrowser()` to the body of the `intercept()` method - then it will be more clear that this interceptor is disabled in SSR. However I don't insist on this change, I give just a suggestion.

integration-libs/segment-refs/root/http-interceptors/occ-segment-refs.interceptor.ts
anjana-bl: done

integration-libs/segment-refs/root/http-interceptors/occ-segment-refs.interceptor.ts
anjana-bl: done

integration-libs/segment-refs/root/http-interceptors/occ-segment-refs.interceptor.ts
anjana-bl: done

projects/schematics/src/shared/lib-configs/integration-libs/segment-refs-schematics-config.ts
```diff
@@ -26,19 +26,22 @@ export const SEGMENT_REFS_SCHEMATICS_CONFIG: SchematicConfig = {
...
+  If Personalization library is not installed, Personalization & Segment-refs won't apply */
   dependencyFeatures: [TRACKING_PERSONALIZATION_FEATURE_NAME],
   importAfter: [
     {
       markerModuleName: PERSONALIZATION_MODULE,
-      featureModuleName: SEGMENT_REFS_MODULE,
+      featureModuleName: SEGMENT_REFS_ROOT_MODULE,
     },
   ],
 };
```
Platonn: Do we _really_ need this config `importAfter` (which creates lazy loadable wrapper module for personalization feature)?
I guess we don't want to create a lazy-loadable wrapper module for personalization feature and we don't want to  append SegmentRefsRootModule after the original PersonalizationModule within this lazy-loadable wrapper module. What do you think?

anjana-bl: done

Platonn (approved):

PR #17609, opened 2023-07-05
feat: Captcha

i53577:
Original PR: https://github.com/SAP/spartacus/pull/16464


feature-libs/user/profile/components/register/register.component.html
```diff
@@ -211,6 +211,10 @@
               </label>
             </div>
           </div>
+          <cx-captcha (confirmed)="captchaConfirmed()"></cx-captcha>
```
i53577: The function `captchaConfirmed()` is used to set `registerForm.captcha`. It will help to handle `cx-form-errors`.

feature-libs/user/profile/occ/adapters/occ-user-profile.adapter.ts
```diff
@@ -191,4 +199,20 @@ export class OccUserProfileAdapter implements UserProfileAdapter {
...
+  protected appendCaptchaToken(currentHeaders: HttpHeaders): HttpHeaders {
+    if (this.injector && this.captchaConfig?.captchaProvider) {
+      const provider = this.injector.get<CaptchaProvider>(
+        this.captchaConfig.captchaProvider
+      );
+      const isCaptchaEnabled = provider.getCaptchaConfig().subscribe((config) => {
+        return config.enabled;
+      });
+
+      if (provider?.getToken() && isCaptchaEnabled) {
```
i53577: only append token when captcha is enabled.

projects/storefrontapp-e2e-cypress/cypress/e2e/regression/user_access/accessibility/captcha-tabbing-order.cy.ts
```diff
@@ -0,0 +1,33 @@
...
+  });
+
+  context('Captcha', () => {
+    context('Register page', () => {
+      it('should allow to navigate to captcha with tab key', () => {
+        cy.intercept('GET', /\.*\/basesites\?fields=.*/, (req) => {
+          req.continue((res) => {
+            res?.body?.baseSites?.forEach((baseSite) => {
+              baseSite.captchaConfig = {
+                enabled: true,
```
i53577: set captcha enabled as `true` for testing

projects/storefrontapp-e2e-cypress/cypress/e2e/regression/user_access/recaptchaV2.core-e2e.cy.ts
```diff
@@ -0,0 +1,45 @@
...
+      cy.window().then((win) => win.sessionStorage.clear());
+    });
+
+    it('should register and redirect to login page (CXSPA-805)', () => {
+      cy.visit('/');
+      cy.intercept('GET', /\.*\/basesites\?fields=.*/, (req) => {
+        req.continue((res) => {
+          res?.body?.baseSites?.forEach((baseSite) => {
+            baseSite.captchaConfig = {
+              enabled: true,
```
i53577: set captcha `enabled` as true for testing

projects/storefrontapp-e2e-cypress/cypress/e2e/regression/user_access/recaptchaV2.core-e2e.cy.ts
```diff
@@ -0,0 +1,45 @@
...
+        clickHamburger();
+      });
+      cy.findByText(/Sign in \/ Register/i).click();
+
+      cy.intercept('GET', /\.*\/bframe\?hl=.*/).as('getIFrame');
+      cy.get('cx-login-register').findByText('Register').click();
+      cy.wait('@getIFrame').its('response.statusCode').should('eq', 200);
+      cy.get('cx-captcha').find('iframe').should('exist');
+      registerWithCaptcha(user);
+      verifyGlobalMessageAfterRegistration();
```
i53577: Checking if a customer can be registered successfully with captcha enabled.

projects/storefrontapp-e2e-cypress/cypress/helpers/auth-forms.ts
```diff
@@ -63,6 +63,22 @@ export function register(
...
+  hiddenConsent?
+) {
+  fillRegistrationForm(user, giveRegistrationConsent, hiddenConsent);
+  cy.get('iframe')
+    .first()
+    .then((recaptchaIframe) => {
+      const body = recaptchaIframe.contents();
+      cy.wrap(body).find('#recaptcha-anchor').click();
+    });
+  cy.wait(3000);
```
Zeyber: It is antipattern to wait for an arbiatrary amount of time. We should rather wait for something that indicates something is completed (eg. an http request/an element existing).

projects/core/src/model/misc.model.ts
```diff
@@ -128,4 +128,10 @@ export interface BaseSite {
   baseStore?: BaseStore;
   requiresAuthentication?: boolean;
   isolated?: boolean;
+  captchaConfig?: CaptchaConfig;
+}
+
+export interface CaptchaConfig {
+  enabled: boolean;
```
Zeyber: Does it make sense `enabled` optional and assume truthy or falsy if it does not exist?

projects/storefrontapp-e2e-cypress/cypress.config.ci.ts
```diff
@@ -29,6 +29,7 @@ export default defineConfig({
     OCC_PREFIX: '/occ/v2',
     OCC_PREFIX_USER_ENDPOINT: 'users',
     OCC_PREFIX_ORDER_ENDPOINT: 'orders',
+    RECAPTCHA_PUBLIC_KEY: '6LeIxAcTAAAAAJcZVRqyHh71UMIEGNQ_MXjiZKhI',
```
Zeyber: For security reasons, I would use a different key exclusive stored as a secret to the CI's domain(s). This is so that not anybody with `localhost` can use our key.

projects/storefrontapp-e2e-cypress/cypress.config.ts
```diff
@@ -23,6 +23,7 @@ export default defineConfig({
     OCC_PREFIX: '/occ/v2',
     OCC_PREFIX_USER_ENDPOINT: 'users',
     OCC_PREFIX_ORDER_ENDPOINT: 'orders',
+    RECAPTCHA_PUBLIC_KEY: '6LeIxAcTAAAAAJcZVRqyHh71UMIEGNQ_MXjiZKhI',
```
Zeyber: Is having this key in code approved by the security team? I'm not sure another easy way to go about it in local testing but I'd like to cross off any concern they may have.

feature-libs/user/profile/components/register/register.component.ts
```diff
@@ -65,6 +65,7 @@ export class RegisterComponent implements OnInit, OnDestroy {
         this.registerComponentService.generateAdditionalConsentsFormControl?.() ??
         this.fb.array([]),
       termsandconditions: [false, Validators.requiredTrue],
+      captcha: [false, Validators.requiredTrue],
```
Zeyber: This is a breaking change. No user will be able to register anymore without enabling captcha. So we need to make this validator opt-in (eg. with feature flags or checking whether some captcha service has been included) until the next major version `7.0`.

projects/storefrontapp-e2e-cypress/cypress/e2e/regression/user_access/recaptchaV2.core-e2e.cy.ts
```diff
@@ -0,0 +1,45 @@
+/*
```
Zeyber: It could be nice to stress test our captcha feature in e2e tests so that we can make sure captcha will not fail if we do things like (for example, but does not necessarily need to be these scenarios):
- Register multiple users
- Navigate to and from registration and try register
- Register incorrectly before correctly
In essence, we want to make sure that we are not reinitializing our script or doing anything funny if we were to trigger our captcha service multiple times.

projects/storefrontlib/shared/components/captcha/captcha.component.ts
```diff
@@ -0,0 +1,91 @@
...
+export class CaptchaComponent implements AfterViewInit, OnDestroy {
+  // Emits true if user confirms captcha
+  @Output() confirmed = new EventEmitter<boolean>();
+
+  @ViewChild('captcha', { static: false }) captchaRef: ElementRef;
+
+  protected subscription = new Subscription();
+
+  constructor(
+    protected config: GoogleRecaptchaApiConfig,
```
Zeyber: We want the captcha service to be agnostic of which captcha we are using. It should not be tied to Google's Recaptcha service. 

Please see our `TagManagementService` feature for a great example of how this can be done. In that example, the TmsService itself is agnostic but can be used for `Adobe` or `Google`'s Tag Managers interchangeably as we would like to do here.

Zeyber (changes requested):

projects/core/src/model/misc.model.ts
i53577: According to the logic of backend commerce, it is either true or false, and it will not be null.

projects/storefrontlib/shared/components/captcha/captcha.component.ts
i53577: IMHO, it doesn't make much sense to do so. SPA needs to call backend to complete the captcha feature. And the backend commerce has indicated that it uses Google's Recaptcha service. Without google recaptcha, the register process can not be completed successfully. See TAM Diagram in the following link: https://wiki.one.int.sap/wiki/display/SAPCCWiki/%5BTM%5D+Spartacus+Feature+Registration+Captcha

feature-libs/user/profile/components/register/register.component.ts
i53577: Thx for your info. I will do it after confirming the release version with PO.

projects/storefrontapp-e2e-cypress/cypress.config.ts
i53577: Just confirm with security guy, the answer is not. But we do need it to enable captcha for e2e test. 
Do you have any other ideas?  Can we put the publicKey in pipeline env, and get the value from it?

projects/storefrontapp-e2e-cypress/cypress/e2e/regression/user_access/recaptchaV2.core-e2e.cy.ts
i53577: > register multiple users

why do captcha need this? The original test for register component only register one user. Is it for testing trigger captcha service multiple times?
> Navigate to and from registration and try register

this scenario has been covered. 
> Register incorrectly before correctly

Added.  Please see: https://github.com/SAP/spartacus/pull/17609/files#diff-f84f08fb6911c235a18ff25f1b7d7c2b933ab4b358ad40f741e85e151b4ec47eR75

projects/storefrontapp-e2e-cypress/cypress/helpers/auth-forms.ts
i53577: I used to use `cy.waitUtil()` for this condition. But seems SPA's cypress version doesn't support it. I will find another way to do this.

i53577: Fixed. Please check the latest code.

projects/storefrontapp-e2e-cypress/cypress.config.ts
Zeyber: Yes, it should be possible to add a secret key in the ci workflow (our github actions) using github secrets. I believe you will need to speak to an admin to do this.

projects/storefrontlib/shared/components/captcha/captcha.component.ts
Zeyber: While our goal has been to reach parity with the old accelerator, this is only a minimum scope. Backend has been designed with this in mind but also with the idea to expand functionality in the future. Google reCaptcha v2 is only the minimum but clients are likely to want to use captchas by different providers (even google recaptcha v3 is technically different).

We want captchas to have the possibility to be interchangable for our users and we need to develop with this in mind (Spartacus is all about extensibility afterall). We only want to use this Google captcha as a default for Spartacus, but also leave the possibility for it to be swapped our for a different configuration. Therefore, I strongly believe we should decouple `captcha` from `google-captcha-v2` as much as possible.

Does it make sense?

projects/storefrontapp-e2e-cypress/cypress/e2e/regression/user_access/recaptchaV2.core-e2e.cy.ts
Zeyber: `Is it for testing trigger captcha service multiple times?`

Yes, exactly. I have personally had some funny behaviour from google captcha when validation is retriggered. We want to ensure this doesn't happen for us.

Zeyber (changes requested):
One more e2e test that could be good to make sure Captcha is working correctly (covering all our basic):
- Is it possible to test that Captcha will block registration if the public key is incorrect?

PR #17610, opened 2023-07-05
fix: include PDF and Video modules to schematics for Spartacus installations

RadhepS:
closes: https://jira.tools.sap/browse/CXSPA-3888


FollowTheFlo (approved):
good to me

PR #17624, opened 2023-07-11
feat: handle http errors in ssr

kpawelczak:


projects/core/src/error-handling/http-error-handler/http-error-handler.interceptor.ts
```diff
@@ -0,0 +1,38 @@
...
+  HttpErrorResponse,
+  HttpEvent,
+  HttpHandler,
+  HttpInterceptor,
+  HttpRequest,
+} from '@angular/common/http';
+import { Observable } from 'rxjs';
+import { tap } from 'rxjs/operators';
+
+@Injectable()
```
Platonn, line 25: Let's add a bit of JSDocs. Just for inspiration:
```
This interceptor forwards all HTTP errors (e.g. 5xx or 4xx status response from backend)
to Angular's `ErrorHandler`.

Thanks to this, in SSR, any HTTP error from backend can potentially mark the Server-Side Rendering
as errord and therefore allow for sending an appropriate error response to a final client of SSR.
```

projects/core/src/error-handling/http-error-handler/http-error-handler.interceptor.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+    });
+    spyOn(errorHandler, 'handleError');
+
+    next.handle = () => throwError(error);
+
+    interceptor.intercept(request, next).subscribe({
+      error: (err) => {
+        expect(err).toEqual(error);
+        expect(errorHandler.handleError).toHaveBeenCalledWith(error);
+      },
```
Platonn, line 58: Let's include the Jasmine `done` callback inside the body of the `subscribe()` method. Otherwise, there is a possibility that the callback function may not be executed at all. This can happen if the observable doesn't emit any values due to a bug in the code-under-test or a mistake in the mocking logic.

(the same remark applies to subscription in line 69-71).

projects/core/src/error-handling/http-error-handler/http-error-handler.interceptor.ts
kpawelczak: thanks, I've added jsdocs

projects/core/src/error-handling/http-error-handler/http-error-handler.interceptor.spec.ts
kpawelczak: fixed

Platonn (approved):

PR #17631, opened 2023-07-12
doc(installation schematics config): update docs for 'dependencyFeatures' and 'importAfter'

Platonn:


projects/schematics/README.md
```diff
@@ -93,12 +93,11 @@ The objects has to conform to the `SchematicsConfig` interface:
...
+- `importAfter` - related to lazy loadable wrapper modules, and specifies after which module (aka "marker" module) should the given module be imported. E.g. the `CheckoutB2BModule` should be imported after the base checkout's `CheckoutModule` withing the lazy loadable `CheckoutWrapperModule`. It's needed for extension feature modules, when they overwrite some cms components or services of a core feature that is lazy-loaded.
```
znikola:
```suggestion
- `importAfter` - related to lazy loadable wrapper modules, and specifies after which module (aka "marker" module) should the given module be imported. E.g. the `CheckoutB2BModule` should be imported after the base checkout's `CheckoutModule` within the lazy loadable `CheckoutWrapperModule`. This is needed for extension feature modules, when they overwrite some cms components or services of a core feature that is lazy-loaded.
```

projects/schematics/README.md
```diff
@@ -93,12 +93,11 @@ The objects has to conform to the `SchematicsConfig` interface:
...
-- `importAfter` - related to wrapper modules, and specifies after which module (aka "marker" module) should the given module be imported. E.g. the `CheckoutB2BModule` should be imported after the base checkout's `CheckoutModule`.
+- `dependencyFeatures` - should configure the runtime features on which your library depends on (e.g. `Order` feature depends on `User` feature). By specifying this dependency you ensure 2 things: 1) when the user chooses in the spartacus installation schematics to install the feature `Checkout` , then the feature `Order` will be installed automatically even when the user doesn't explicitly choose to install `Order`. 2) Moreover, during the actual process of inserting code via schematics into customer's app codebase, we ensure that installation schematics of `Order` will run first, and`Checkout` only later. The aforementioned 2 points apply also for any transitive dependencies, e.g. when `Checkout` depends on `Order` which depends on `User` - in this case `User` will be installed first, then `Order` and finally `Checkout`, even when the user only explicitly chose to install `Checkout`.
```
znikola:
```suggestion
- `dependencyFeatures` - should configure the runtime features on which your library depends (e.g. `Order` feature depends on `User` feature). By specifying this dependency you ensure 2 things: 1) when the user chooses in the Spartacus installation schematics to install the feature `Checkout`, then the feature `Order` will be installed automatically even when the user doesn't explicitly choose to install `Order`. 2) Moreover, during the actual process of inserting code via schematics into the customer's codebase, we ensure that installation schematics of `Order` will run first, and`Checkout` only later. The aforementioned 2 points apply also for any transitive dependencies, e.g. when `Checkout` depends on `Order` which depends on `User` - in this case `User` will be installed first, then `Order` and finally `Checkout`, even when the user only explicitly chose to install `Checkout`.
```

PR #17632, opened 2023-07-12
fix 'blank address' for order-review cpnt

FollowTheFlo:
CXSPA-3330
context: OPF adapter server cannot provide billing addr and paymentInfo immediately, it needs few minutes. During that time, Confirmation Page displays PaymentInfo with blank date. We expect to not display the PaymentInfo in this scenario. BillingAddress already works this way (not displayed if not present).
more UI logic: When both are missing, we also need to remove the entire 3rd column
code change:
split order-review component with component service, 
- this service only contains the method we are modifying (getPaymentInfoCardContent) and flag 'paymentIntegration' that is use to control UI -> display PaymentIntegrationTemplate if paymentIntegration flag is true.
- no need to modify unit tests as component logic stays the same 
image


integration-libs/opf/base/root/components/opf-order-overview/opf-order-overview-component.service.ts
FollowTheFlo, line 1: this file is in root folder as we need to specify the 'overwriting of standard service' in opf-base-root.module.ts
```
OpfOrderOverviewComponentService,
    {
      provide: OrderOverviewComponentService,
      useExisting: OpfOrderOverviewComponentService,
    },
```
afaik, Base root module can only import files from root folder.

integration-libs/opf/base/root/components/opf-order-overview/opf-order-overview-component.service.ts
```diff
@@ -0,0 +1,51 @@
...
+  getPaymentInfoCardContent(payment: PaymentDetails): Observable<Card> {
+    return combineLatest([
+      this.translation.translate('paymentForm.payment'),
+      this.translation.translate('paymentCard.expires', {
+        month: Boolean(payment) ? payment.expiryMonth : '',
+        year: Boolean(payment) ? payment.expiryYear : '',
+      }),
+    ]).pipe(
+      filter(() => Boolean(payment)),
+      filter(() => this.isPaymentInfoCardFull(payment)),
```
FollowTheFlo, line 31: the only additional logic: don't display the payment card if any missing info.

feature-libs/order/components/order-details/order-overview/order-overview.component.html
```diff
@@ -130,4 +137,26 @@
       <cx-order-detail-billing></cx-order-detail-billing>
     </div>
   </ng-container>
+
+  <ng-template #paymentIntegrationTemplate>
+    <ng-container
```
FollowTheFlo, line 142: 3rd column not displayed if both (PaymentInfo and BillingAddr) have empty value

PR #17636, opened 2023-07-13
CXSPA-3836 first e2e test for configurator integration

ChristophHi:


projects/storefrontapp-e2e-cypress/cypress/e2e/regression/quote/quote-configurator.e2e-flaky.cy.ts
```diff
@@ -0,0 +1,65 @@
...
+      common.goToPDPage(POWERTOOLS, testProductConfigurableWithIssues);
+      quote.setQtyOnPD('1');
+      common.clickOnAddToCartBtnOnPD();
+      common.clickOnViewCartBtnOnPD();
+      quote.clickOnRequestQuoteInCart();
+
+      //we are still in cart, for now just check that
+      //TODO check for messages once https://jira.tools.sap/browse/CXSPA-4079 is done
+      cy.get('cx-cart-details').should('exist');
+      quote.clearCart();
```
Uli-Tiger: this fails when i execute the tests (and hence all subsequent tests, as the cart contains product with issues, hence no quote can be created)
image
Additionally a more robust approach would be to move this to a beforeEach hook, that clears the cart if required.

projects/storefrontapp-e2e-cypress/cypress/e2e/regression/quote/quote-configurator.e2e-flaky.cy.ts
```diff
@@ -0,0 +1,65 @@
...
+    it('should result in a quote in draft state', () => {
+      quote.requestQuote(POWERTOOLS, testProductConfigurableTextfield, '1');
+      quote.checkQuoteIsDraft();
+    });
+
+    it('should show navigation link to configurator on quote details and allow to navigate to textfield form', () => {
+      quote.requestQuote(POWERTOOLS, testProductConfigurableTextfield, '1');
+      configurationCart.clickOnDisplayConfigurationLink(0);
+      cy.get('cx-configurator-textfield-form').should('be.visible');
+    });
```
Uli-Tiger: Those both test cases could be merged into one. They test the same process, just assert different aspects. As each cypress test has some overhead (login, create quote, etc) we should aim for more process like tests with several steps and asserts, rather than this unit test like approach where we only assert one aspect per test case.

projects/storefrontapp-e2e-cypress/cypress/e2e/regression/quote/quote-configurator.e2e-flaky.cy.ts
```diff
@@ -0,0 +1,65 @@
...
+      //TODO check for messages once https://jira.tools.sap/browse/CXSPA-4079 is done
+      cy.get('cx-cart-details').should('exist');
+      quote.clearCart();
+    });
+
+    it('should show navigation link to configurator on quote details page and allow to navigate to OV page', () => {
+      quote.requestQuote(POWERTOOLS, testProductConfigurable, '1');
+      configurationCart.clickOnDisplayConfigurationLink(0);
+      cy.get('cx-configurator-overview-sidebar').should('be.visible');
+      configuratorOverview.clickContinueToCartBtnOnOPAndExpectQuote();
```
Uli-Tiger: This test case could be merged with the first one of this group They test the same process, just assert different aspects. As each cypress test has some overhead (login, create quote, etc) we should aim for more process like tests with several steps and asserts, rather than this unit test like approach where we only assert one aspect per test case.

projects/storefrontapp-e2e-cypress/cypress/e2e/regression/quote/quote-configurator.e2e-flaky.cy.ts
```diff
@@ -0,0 +1,65 @@
...
+      quote.requestQuote(POWERTOOLS, testProductConfigurable, '1');
+      quote.checkQuoteIsDraft();
+    });
+
+    it('should not allow to request quote if configuration has issues', () => {
+      common.goToPDPage(POWERTOOLS, testProductConfigurableWithIssues);
+      quote.setQtyOnPD('1');
+      common.clickOnAddToCartBtnOnPD();
+      common.clickOnViewCartBtnOnPD();
+      quote.clickOnRequestQuoteInCart();
```
Uli-Tiger: could be a method in the quote helper similar to requestQuote

Uli-Tiger (changes requested):

projects/storefrontapp-e2e-cypress/cypress/e2e/regression/quote/quote-configurator.e2e-flaky.cy.ts
ChristophHi: I replaced that with a better, simpler method from the cart e2e tools (but did not do that in beforeEach, as I struggled to get that stable. I didn't want to build the tools for that in the cart area, seemed to be over-engineered).

projects/storefrontapp-e2e-cypress/cypress/e2e/regression/quote/quote-configurator.e2e-flaky.cy.ts
ChristophHi: Agree, changed that

projects/storefrontapp-e2e-cypress/cypress/e2e/regression/quote/quote-configurator.e2e-flaky.cy.ts
Uli-Tiger: @ChristophHi still not working for me, because it seems that the @delete_cart_item alias is registered for electronics store route, while the test runs in powertools.
image
image
How are you running the tests (maybe i missed something?)
i just run `npm run cy:open` and selected your spec?

Uli-Tiger: ok found it myself - when starting cypres with `npm run cy:open:b2b` it works fine

Uli-Tiger (approved):

PR #17638, opened 2023-07-14
test: Remove the asm.agentSignOut() in all ASM E2E Tests

luoyifeng11:
[CXSPA-3849](https://jira.tools.sap/browse/CXSPA-3849)

### DIFF
--- a/projects/storefrontapp-e2e-cypress/cypress/e2e/regression/asm/create-customer.e2e-flaky.cy.ts
+++ b/projects/storefrontapp-e2e-cypress/cypress/e2e/regression/asm/create-customer.e2e-flaky.cy.ts
@@ -16,6 +16,7 @@ context('Assisted Service Module', () => {
   });
 
   describe('Create Customer', () => {
+    const user = getSampleUser();
     it('should be able to create a new customer by agent click button (CXSPA-1594)', () => {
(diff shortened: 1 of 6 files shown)


projects/storefrontapp-e2e-cypress/cypress/e2e/regression/asm/create-customer.e2e-flaky.cy.ts
```diff
@@ -120,7 +118,9 @@ context('Assisted Service Module', () => {
           'Saved items'
         );
       });
+    });
 
+    it('vertify login to Spartacus (CXSPA-1594)', () => {
```
niehuayang, line 123: For this scenario, I don't think that split into two cases, otherwise, it's not a complete scenario. You can see the [scenario](https://jira.tools.sap/browse/CXSPA-1594) form Linda's comment:
```
Cypress scenario: 
Asagent access SPA ASM then click top-right link "create customer"
Filling customer info then save
Customer is created within success message and emulated
Go to backoffice and set password for customer (Can API to cover?)
Back to SPA then new customer should be login in
```

niehuayang (changes requested):

niehuayang (approved):

PR #17641, opened 2023-07-14
add error message on PaymentType CLASS_MISMATCH_ERROR

FollowTheFlo:
context: Create a user on b2c, then purchase aproduct on b2b, paymentMethod step will load infinitly without an error displayed.
This code change add a global message only in this specifuc situation (when CLASS_MISMATCH_ERROR is returned from occ endpoint) and hide the spinner
-> added globalMessage at component level
-> message 'You are not authorized to perform this action. Please contact your administrator if you think this is a mistake.'

About breaking change:
constructor-> added optional GlobalMessageService in cpnt, html template is modified only when GlobalMessageService is implemented.


feature-libs/checkout/b2b/components/checkout-payment-type/checkout-payment-type.component.ts
```diff
@@ -108,7 +134,8 @@ export class CheckoutPaymentTypeComponent {
   constructor(
     protected checkoutPaymentTypeFacade: CheckoutPaymentTypeFacade,
     protected checkoutStepService: CheckoutStepService,
-    protected activatedRoute: ActivatedRoute
+    protected activatedRoute: ActivatedRoute,
+    protected globalMessageService: GlobalMessageService
```
kimhw0630: This is breaking change ... as far as I know, you need to handle it.

kimhw0630: here is example how to handle (launchDialogService is a newly added property in the constructor)
image

feature-libs/checkout/b2b/components/checkout-payment-type/checkout-payment-type.component.html
```diff
@@ -3,79 +3,83 @@ <h2 class="cx-checkout-title d-none d-lg-block d-xl-block">
...
-    </div>
-
-    <div class="row">
-      <div class="col-md-12 col-lg-6">
-        <label class="cx-payment-type-container">
-          <span class="label-content">{{
-            'paymentTypes.title' | cxTranslate
-          }}</span>
-          <div class="form-check" *ngFor="let type of paymentTypes">
+  <ng-container *ngIf="!paymentTypesError">
```
FollowTheFlo, line 6: This condition is the only change made in template: don't display form if an error occurs when fetching PaymentTypes

feature-libs/checkout/b2b/components/checkout-payment-type/checkout-payment-type.component.ts
FollowTheFlo: Thanks Hak, i'll check it out LaunchDialog for an error message

kimhw0630 (approved):

PR #17657, opened 2023-07-18
feat: created effect for handling ngrx errors

kpawelczak:
jira ticket: https://jira.tools.sap/browse/CXSPA-4178
We've added a new feature called the "Effects Error Handler" to better manage errors in NgRx, excluding HTTP errors.
In our first attempt, we tried to catch errors by looking at Actions that extended EntityFailAction or EntityScopedFailActions.. However, this didn't cover all the failed Actions used in Spartacus. As a solution, we've decided to create a new interface called ErrorAction and simplify all the fail actions. 

### **Testing:**

For quick testing refer to this branch https://github.com/SAP/spartacus/tree/feat/CXSPA-3779-effects-action-error-testing
find HttpErrorHandlerInterceptor to change blocked requests or modify the error message. 
To test specific action you need to know the request it uses and then add it to the blocked requests array. In the intercept function, you can modify the error in any way it can be a HttpError or a regular one.

Steps:
1. Checkout to `feat/CXSPA-3779-effects-action-error-testing`
or find and change HttpErrorHandlerInterceptor with a following code:
```typescript
@Injectable()
export class HttpErrorHandlerInterceptor implements HttpInterceptor {
  private blockedRequests: string[] = [
    'product',
  ];

  constructor(protected errorHandler: ErrorHandler) {}

  intercept(
    request: HttpRequest<any>,
    next: HttpHandler
    return next.handle(request).pipe(
      tap(
        () => {
          if (this.isRequestBlocked(request.url)) {
            const error = new Error('NgRx test error')
            throw error;
          }
        }
      )
    );
  }

  protected handleError(error: HttpErrorResponse): void {
    this.errorHandler.handleError(error);
  }

  private isRequestBlocked(url: string): boolean {
    return this.blockedRequests.some((blockedUrl) => url.includes(blockedUrl));
  }
}
``` 
2. To test a specific action you have to add the request URL that it uses and add it to the blocked `blockedRequests` requests array.
3. If needed change the error message
4. Start the application by using `npm start`
5. Go to the home page and open up the devtools
6. Notice the error in the console


projects/core/src/error-handling/effects-error-handler/effects-error-handler.service.ts
```diff
@@ -0,0 +1,42 @@
...
+    if (isNotHttpError) {
+      this.errorHandler.handleError(error);
+    }
+  }
+
+  filterActions(_action: Action): boolean {
+    return false;
+  }
+
+  protected getError(action: any): any {
```
Platonn: Could you add JSDoc explaining _why_ we need this, please?

projects/core/src/error-handling/effects-error-handler/effects-error-handler.service.ts
```diff
@@ -0,0 +1,42 @@
...
+    if (action.error) {
+      return action.error;
+    } else if (action.payload?.error) {
+      return action.payload.error;
+    } else if (action.payload) {
+      return action.payload;
+    } else {
+      return `Action error: ${action?.type}`;
+    }
```
Platonn: nitpick: did you consider using `??`

```
return action.error 
      ?? action.payload?.error 
      ?? action.payload 
      ?? `Action error: ${action?.type}`;
```

projects/core/src/base-core.module.ts
```diff
@@ -41,6 +42,7 @@ import { HttpErrorHandlerModule } from './error-handling';
 
     /* This module should be imported by default starting from version 7.0 (CXSPA-3680)*/
     //ErrorHandlingModule.forRoot(),
+    EffectsErrorHandlerModule.forRoot(),
```
Platonn, line 45: Just an idea. no need to change.
IMHO it would be good not to "pollute" this BaseCoreModule with fine-grained modules, but rather group them. To be discussed.

projects/core/src/error-handling/effects-error-handler/effects-error-handler.service.ts
```diff
@@ -0,0 +1,42 @@
...
+@Injectable()
+export class EffectsErrorHandlerService {
+  constructor(protected errorHandler: ErrorHandler) {}
+
+  handleError(action: any): void {
+    const error = this.getError(action);
+    const isNotHttpError =
+      !(error instanceof HttpErrorModel) &&
+      !(error instanceof HttpErrorResponse);
+
```
Platonn, line 25: Let's explain in a code comment, why we don't propagate http error response (and http error model) to ErrorHandler. i.e. becasue it's already done in the interceptor (tell which one)

projects/core/src/error-handling/effects-error-handler/effects-error-handler.service.ts
```diff
@@ -0,0 +1,42 @@
...
+    if (isNotHttpError) {
+      this.errorHandler.handleError(error);
+    }
+  }
+
+  filterActions(_action: Action): boolean {
+    return false;
+  }
```
Platonn: What about moving the default logic here, as well? 
Now this default method looks a bit emtpy...

projects/core/src/base-core.module.ts
kpawelczak: I think it's a good idea to revisit the base core module once we're done with the error-handling feature. This way, we can get a better understanding of the whole context and also explore the possibility of other modules that could be added in the meantime.

projects/core/src/error-handling/effects-error-handler/effects-error-handler.service.ts
kpawelczak: Added jsdocs

projects/core/src/error-handling/effects-error-handler/effects-error-handler.service.ts
kpawelczak: Yeah looks better

projects/core/src/error-handling/effects-error-handler/effects-error-handler.service.ts
kpawelczak: added comment

projects/core/src/error-handling/effects-error-handler/effects-error-handler.service.ts
kpawelczak: It does look empty, but also I don't think it's a bad thing. Anyways, I've added the default error action types and left a comment.

feature-libs/asm/core/store/actions/customer.action.ts
```diff
@@ -32,9 +32,12 @@ export class CustomerSearch extends StateUtils.LoaderLoadAction {
...
 }
 
-export class CustomerSearchFail extends StateUtils.LoaderFailAction {
+export class CustomerSearchFail
+  extends StateUtils.LoaderFailAction
+  implements ErrorAction
+{
   readonly type = CUSTOMER_SEARCH_FAIL;
-  constructor(public payload: any) {
+  constructor(public error: any) {
```
pawelfras: `query:` Cannot we use `ErrorActionType` instead of `any`? Correct me if I'm wrong, but in the end, we're expecting this type due to the `ErrorAction` implementation.

feature-libs/asm/core/store/actions/customer.action.ts
```diff
@@ -32,9 +32,12 @@ export class CustomerSearch extends StateUtils.LoaderLoadAction {
   }
 }
 
-export class CustomerSearchFail extends StateUtils.LoaderFailAction {
+export class CustomerSearchFail
+  extends StateUtils.LoaderFailAction
+  implements ErrorAction
```
pawelfras: As we discussed, implement `ErrorAction` in classes that are part of utils, e.g. `LoaderFailAction` (and refactor all places that use utils where `ErrorAction` is implemented directly)

projects/core/src/cms/store/effects/navigation-entry-item.effect.ts
```diff
@@ -54,7 +54,9 @@ export class NavigationEntryItemEffects {
                     of(
                       new CmsActions.LoadCmsNavigationItemsFail(
                         data.nodeId,
-                        normalizeHttpError(error, this.logger)
+                        normalizeHttpError(error, this.logger) ?? {
+                          message: error,
+                        }
```
pawelfras: `query:` Maybe it's worth wrapping this logic with another function?

projects/core/src/error-handling/effects-error-handler/effects-error-handler.service.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  /** Here we want to filter which error actions should be handled.
+   * By default, we check if action implements interface ErrorAction  */
+  filterActions(action: Action): boolean {
+    const isErrorAction = (_action: Action): _action is ErrorAction => {
+      return 'error' in _action;
+    };
+    return !!isErrorAction(action);
+  }
```
pawelfras:
```suggestion
  filterActions(action: Action): action is ErrorAction {
    return 'error' in action;
  }
```

projects/core/src/error-handling/effects-error-handler/effects-error-handler.service.ts
pawelfras, line 1: `major`: Unit tests for handler service are missing.

projects/core/src/error-handling/effects-error-handler/cx-error-handler.effect.ts
pawelfras, line 1: `major`: Unit tests for effect are missing.

feature-libs/asm/core/store/actions/customer.action.ts
kpawelczak: This situation is a bit tricky. We're aiming for the right data type, but there's a complication. In each Effect, errors get normalized using `normalizeHttpError`. This sometimes leads to the return value being `undefined`. To tackle this, I suggest making a new method like `getError` that always gives us some type of error. The main question is whether we should adjust `normalizeHttpError` to return the provided `error` instead of `undefined`.

projects/core/src/cms/store/effects/navigation-entry-item.effect.ts
kpawelczak: Yes, that would be a good idea but still, I don't know if we won't change normlizeHttpError to not return undefined but error.

feature-libs/asm/core/store/actions/customer.action.ts
kpawelczak: I've replaced the `any` after all. It seemed like a right thing to do.

feature-libs/asm/core/store/actions/customer.action.ts
kpawelczak: fixed

projects/core/src/cms/store/effects/navigation-entry-item.effect.ts
kpawelczak: I've created a new function `tryNormalizeHttpError`. In any case should be almost easy to remove.

projects/core/src/error-handling/effects-error-handler/effects-error-handler.service.ts
kpawelczak: replaced

projects/core/src/error-handling/effects-error-handler/effects-error-handler.service.ts
kpawelczak: done

projects/core/src/error-handling/effects-error-handler/cx-error-handler.effect.ts
kpawelczak: done

projects/core/src/util/normalize-http-error.ts
```diff
@@ -57,3 +57,10 @@ export function normalizeHttpError(
 
   return undefined;
 }
+
+export function tryNormalizeHttpError(
```
pawelfras: `query`: what do you think about changing the name to `handleHttpError`?
Regardless of the function name, I would move this function to a separate file.

projects/core/src/state/utils/entity-loader/entity-loader.action.ts
```diff
@@ -26,9 +27,11 @@ export interface EntityLoaderAction extends Action {
   readonly meta?: EntityLoaderMeta;
 }
 
+export type EntityLoadId = string | string[] | null;
```
pawelfras: Is this change in the scope of the task? If not, let's do it later in a separate PR.

projects/core/src/error-handling/effects-error-handler/effects-error-handler.service.ts
```diff
@@ -0,0 +1,36 @@
...
+    if (isNotHttpError) {
+      this.errorHandler.handleError(error);
+    }
+  }
+
+  /** Here we want to filter which error actions should be handled.
+   * By default, we check if action implements interface ErrorAction  */
+  filterActions(action: Action): boolean {
+    return 'error' in action;
+  }
```
pawelfras:
```suggestion
  filterActions(action: Action): action is ErrorAction {
    return 'error' in action;
  }
```

then in `cx-error-handler.effect.ts` thanks to type guard, you'll not need to use type casting
```
error$: Observable<ErrorAction> = createEffect(
    () =>
      this.actions$.pipe(
        filter(this.effectErrorHandler.filterActions),
        tap(this.effectErrorHandler.handleError)
      ),
    { dispatch: false }
  );
 ```

File: projects/core/src/error-handling/effects-error-handler/cx-error-handler.effect.ts [hunk 13]
```diff
@@ -0,0 +1,35 @@
...
+        filter((action: Action) =>
+          this.effectErrorHandler.filterActions(action)
+        ),
+        map((action: Action) => action as ErrorAction),
+        tap((action: ErrorAction) => {
+          this.effectErrorHandler.handleError(action);
+        })
+      ),
+    { dispatch: false }
+  );
```
pawelfras:
```suggestion
 error$: Observable<ErrorAction> = createEffect(
    () =>
      this.actions$.pipe(
        filter(this.effectErrorHandler.filterActions),
        tap((action) => this.effectErrorHandler.handleError(action))
      ),
    { dispatch: false }
  );
```

after change proposed for `effects-error-handler.service.ts`

projects/core/src/state/utils/entity-loader/entity-loader.action.ts
kpawelczak: Since I touched this file, Sonar Cloud threw me an error, hence the change.

pawelfras: In that case, I propose to change the name to `EntityId`

pawelfras: As we discussed, there are more changes related to this union type, so let's move this to a separate ticket.

pawelfras (approved):

PR #17662, opened 2023-07-19
CXSPA-4049 Add tests for payment verification process

rmch91:
Closes https://jira.tools.sap/browse/CXSPA-4049

### DIFF
--- a/integration-libs/opf/base/root/services/opf-payment-metadata-store.service.spec.ts
+++ b/integration-libs/opf/base/root/services/opf-payment-metadata-store.service.spec.ts
@@ -14,6 +14,12 @@ const initialState = {
   isPaymentInProgress: false,
 };
 
+const state: OpfPaymentMetadata = {
+  isPaymentInProgress: true,
+  selectedPaymentOptionId: 111,
+  termsAndConditionsChecked: true,
+};
+
 describe('OpfPaymentMetadataStoreService', () => {
   let service: OpfPaymentMetadataStoreService;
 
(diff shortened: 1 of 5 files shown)


integration-libs/opf/base/root/services/opf-state-persistence.service.spec.ts
```diff
@@ -1 +1,119 @@
...
+    );
+
+    service['getOpfState']().subscribe((state) => {
+      expect(state).toEqual({ metadata: mockOpfMetadata });
+      done();
+    });
+  });
+
+  it('should update OpfPaymentMetadataStoreService when onRead is called', () => {
+    const mockOpfMetadata: OpfPaymentMetadata = {
```
Matejk00: Can we declare this const more at the top of the test scenarios? So it's no need to re-declare this property each time.

integration-libs/opf/base/root/services/opf-payment-metadata-store.service.spec.ts
```diff
@@ -0,0 +1,86 @@
...
+
+    service.getOpfMetadataState().subscribe((state) => {
+      expect(state).toEqual(state);
+      done();
+    });
+  });
+
+  it('should update opfPaymentMetadataState with the given payload', () => {
+    const state: OpfPaymentMetadata = {
+      isPaymentInProgress: false,
```
Matejk00, line 54: The same case here.

Matejk00 (changes requested):

integration-libs/opf/base/root/services/opf-payment-metadata-store.service.spec.ts
rmch91: done

integration-libs/opf/base/root/services/opf-state-persistence.service.spec.ts
rmch91: Changed

Matejk00 (approved):

PR #17665, opened 2023-07-20
1. Spyke for using cx-card in  quote details tiles

Larisa-Staroverova:


feature-libs/quote/components/details/overview/quote-details-overview.component.ts
```diff
@@ -5,24 +5,181 @@
...
+    const saveCardEventPropertyNames =
+      Object.getOwnPropertyNames(saveCardEvent);
+
+    metaDataPropertyNames?.forEach((name) => {
+      const index = saveCardEventPropertyNames.findIndex((element) =>
+        element.includes(name + '_')
+      );
+      if (index >= 0) {
+        const propertyName = saveCardEventPropertyNames[index];
+        metaData[name] = saveCardEvent[propertyName];
```
ChristophHi, line 69: Please fix strict mode issue
```
Element implicitly has an 'any' type because expression of type 'string' can't be used to index type 'QuoteMetadata'.
  No index signature with a parameter of type 'string' was found on type 'QuoteMetadata'.
```

feature-libs/quote/components/details/overview/quote-details-overview.component.html
```diff
@@ -5,70 +5,47 @@ <h3 class="quote-details-header">
...
-            getCardContent(
-              quoteDetails.updatedTime | cxDate,
-              'quote.details.lastUpdated'
-            ) | async
-          "
-        ></cx-card>
+        <div class="cx-content">
+          <cx-card
+            [content]="
+              getQuoteInformation(quoteDetails.name, quoteDetails.description)
```
ChristophHi, line 18: Please use async pipe before creating card component (because otherwise card component will be initialized with null)

feature-libs/quote/components/details/overview/quote-details-overview.component.html
```diff
@@ -5,70 +5,47 @@ <h3 class="quote-details-header">
...
-        <cx-card
-          class="cx-card-description"
-          [content]="
-            getCardContent(
-              quoteDetails.description,
-              'quote.details.description'
+            getEstimatedAndDate(
+              quoteDetails,
+              quoteDetails.creationTime | cxDate
             ) | async
```
ChristophHi, line 39: See above

feature-libs/quote/components/details/overview/quote-details-overview.component.html
```diff
@@ -5,70 +5,47 @@ <h3 class="quote-details-header">
...
-        ></cx-card>
-        <cx-card
-          [content]="
-            getCardContent(
-              quoteDetails.expirationTime | cxDate,
-              'quote.details.expiryDate'
+            getUpdate(
+              quoteDetails.updatedTime | cxDate,
+              quoteDetails.expirationTime | cxDate
             ) | async
```
ChristophHi, line 49: See above

projects/storefrontlib/shared/components/card/card.component.ts
```diff
@@ -5,8 +5,14 @@
  */
 
 import { Component, EventEmitter, Input, OnInit, Output } from '@angular/core';
+import { UntypedFormControl, UntypedFormGroup } from '@angular/forms';
 import { ICON_TYPE } from '../../../cms-components/misc/icon/icon.model';
 
+export interface SaveCardEvent {
+  saveMode: boolean;
```
ChristophHi, line 12: Please use a different name (yes editMode is already taken it seems...)

projects/storefrontlib/shared/components/card/card.component.ts
```diff
@@ -79,11 +96,22 @@ export class CardComponent implements OnInit {
...
   cancelEdit(): void {
     this.editMode = false;
     this.cancelCard.emit(5);
   }
 
+  cancelSave(): void {
+    this.saveMode = false;
+    const saveCardEvent: SaveCardEvent = { saveMode: this.saveMode };
+    this.saveCard.emit(saveCardEvent);
+    this.cancelCard.emit(5);
```
ChristophHi, line 112: Why 5?

projects/storefrontlib/shared/components/card/card.component.html
```diff
@@ -18,7 +18,11 @@
     {{ content.header }}
   </div>
   <!-- Card Body -->
-  <div class="card-body cx-card-body" [class.cx-card-delete]="editMode">
+  <div
+    class="card-body cx-card-body"
+    [class.cx-card-delete]="editMode"
+    [class.cx-card-save]="saveMode"
```
ChristophHi, line 24: Isn't that breaking?

projects/storefrontlib/shared/components/card/card.component.html
```diff
@@ -51,14 +55,58 @@
             ></cx-truncate-text-popover>
           </ng-template>
         </div>
-        <div class="cx-card-paragraph" *ngFor="let item of content.paragraphs">
-          <div class="cx-card-paragraph-title">{{ item.title }}</div>
-          <div *ngFor="let text of item.text">
-            <div class="cx-card-paragraph-text">
-              {{ text }}
+        <form [formGroup]="saveForm">
```
ChristophHi, line 58: breaking?

ChristophHi (changes requested):

feature-libs/quote/components/details/overview/quote-details-overview.component.ts
```diff
@@ -5,24 +5,181 @@
...
+    const metaData = this.defineQuoteMetaData(saveCardEvent);
+
+    this.quoteFacade
+      .editQuote(quote.code, metaData)
+      .pipe(take(1))
+      .subscribe(
+        // success
+        () => {
+          this.eventService.dispatch({}, QuoteDetailsReloadQueryEvent);
+        }
```
Uli-Tiger, line 86: should we display an error  on the UI, if editing fails?

feature-libs/quote/components/details/overview/quote-details-overview.component.ts
```diff
@@ -5,24 +5,181 @@
...
+    return this.saveMode;
+  }
+
+  isEditable(quote: Quote): boolean {
+    return (
+      (quote.state &&
+        (quote.state === QuoteState.BUYER_DRAFT ||
+          quote.state === QuoteState.SELLER_DRAFT)) ||
+      quote.isEditable
+    );
```
Uli-Tiger, line 100:
```suggestion
    return (
      quote.state === QuoteState.BUYER_DRAFT ||
      quote.state === QuoteState.SELLER_DRAFT ||
      quote.isEditable
    );
```
1) i think this can be a bit simplified, as we can compare to undefined savefly - see my suggestion
2) do we need the additional check for the status at all?  I believe a quote in status QuoteState.BUYER_DRAFT or QuoteState.SELLER_DRAFT will be marked as isEditable already in the action normalizer. 

so in the end you might just `return quote.isEditable;`

feature-libs/quote/components/details/overview/quote-details-overview.component.ts
```diff
@@ -5,24 +5,181 @@
...
+      this.translationService.translate('quote.details.name'),
+      this.translationService.translate('quote.details.description'),
+    ]).pipe(
+      map(([infoTitle, nameTitle, descriptionTitle]) => {
+        return {
+          title: infoTitle,
+          paragraphs: [
+            {
+              title: nameTitle,
+              text: [name ?? '-'],
```
Uli-Tiger, line 119: may be we can introduce a cost for '-', something like` const NO_DATA = '-';`

projects/storefrontlib/shared/components/card/card.component.html
```diff
@@ -51,14 +55,58 @@
...
+          <div
+            class="cx-card-paragraph"
+            *ngFor="let item of content.paragraphs; trackBy: trackByFn"
+          >
+            <div class="cx-card-paragraph-title">{{ item.title }}</div>
+            <div
+              *ngFor="let text of item.text; let i = index; trackBy: trackByFn"
```
Uli-Tiger, line 65:
```suggestion
              *ngFor="let text of item.text; let i = index; trackBy: trackByIndex"
```
lets rename the track by function to trackByIndex

projects/storefrontlib/shared/components/card/card.component.html
```diff
@@ -51,14 +55,58 @@
...
         </div>
-        <div class="cx-card-paragraph" *ngFor="let item of content.paragraphs">
-          <div class="cx-card-paragraph-title">{{ item.title }}</div>
-          <div *ngFor="let text of item.text">
-            <div class="cx-card-paragraph-text">
-              {{ text }}
+        <form [formGroup]="saveForm">
+          <div
+            class="cx-card-paragraph"
+            *ngFor="let item of content.paragraphs; trackBy: trackByFn"
```
Uli-Tiger, line 61:
```suggestion
            *ngFor="let item of content.paragraphs; trackBy: trackByIndex"
```

projects/storefrontlib/shared/components/card/card.component.ts
Uli-Tiger, line 112: what is meant by cancelSave and why does it emit both savedCard and cancelCard event?
i would expect that you can either cancel or save your changes, but not both.

projects/storefrontlib/shared/components/card/card.component.ts
```diff
@@ -101,6 +129,37 @@ export class CardComponent implements OnInit {
...
+    };
+
+    Object.keys(this.saveForm.controls).forEach((control) => {
+      if (this.saveForm.get(control)?.touched) {
+        const value = this.saveForm.get(control)?.value;
+        if (value) {
+          saveCardEvent[control] = value;
+        }
+        this.saveForm.get(control)?.markAsUntouched();
+        this.saveForm.get(control)?.markAsPristine();
```
Uli-Tiger, line 145:
```suggestion
this.saveForm.get(control)?.reset(value)
```
not 100% sure about this, but maybe you can give it a try.
or even better call the rest on the form group just once, it should reset all controls

https://stackoverflow.com/questions/51449859/how-can-i-reset-a-reactive-form-control-to-originate-state

projects/storefrontlib/shared/components/card/card.component.ts
```diff
@@ -22,7 +28,12 @@ export interface Card {
   title?: string;
   textBold?: string;
   text?: Array<string>;
-  paragraphs?: Array<{ title?: string; text?: Array<string> }>;
+  paragraphs?: Array<{
+    title?: string;
```
Uli-Tiger, line 32: is there any need to make the title and text optional? - if not you could make them mandatory - that would ease the coding below.

projects/storefrontlib/shared/components/card/card.component.ts
```diff
@@ -101,6 +129,37 @@ export class CardComponent implements OnInit {
...
+    });
+
+    this.saveCard.emit(saveCardEvent);
+  }
+
+  setFormControlName(name: string | undefined, index: number): string {
+    if (name) {
+      return name.toLocaleLowerCase().replace(/\s/g, '_') + '_' + index;
+    }
+    return '';
```
Uli-Tiger, line 156: see my previous comment regarding title being optional. another reason to make it mandatory - if you would have several paragraphs with undefined name aka title they would be linked to the same form control, which does not make any sense.

projects/storefrontlib/shared/components/card/card.component.ts
```diff
@@ -101,6 +129,37 @@ export class CardComponent implements OnInit {
...
+  }
+
+  trackByFn(index: any) {
```
Uli-Tiger, line 159:
```suggestion
  trackByIndex(index: any) {
```

projects/storefrontlib/shared/components/card/card.component.ts
```diff
@@ -101,6 +129,37 @@ export class CardComponent implements OnInit {
...
+    const saveCardEvent: SaveCardEvent = {
+      saveMode: this.saveMode,
+    };
+
+    Object.keys(this.saveForm.controls).forEach((control) => {
+      if (this.saveForm.get(control)?.touched) {
+        const value = this.saveForm.get(control)?.value;
+        if (value) {
+          saveCardEvent[control] = value;
+        }
```
Uli-Tiger, line 143: so an property will only be present in the event, if the value was changed. So any caller need to analysis the event if an property is existing. Why not just emitting the actual value for each control no matter whether it was changed or not?
why not have an object per control with attributes: changed and value. So the caller can rely on that the event will contain an object for each control and still has the data whether something was changed or not

Uli-Tiger (changes requested):

projects/storefrontlib/shared/components/card/card.component.ts
```diff
@@ -22,7 +28,12 @@ export interface Card {
   title?: string;
   textBold?: string;
   text?: Array<string>;
-  paragraphs?: Array<{ title?: string; text?: Array<string> }>;
+  paragraphs?: Array<{
+    title?: string;
+    text?: Array<string>;
```
Uli-Tiger, line 33: why is the text in Array?

projects/storefrontlib/shared/components/card/card.component.ts
```diff
@@ -5,8 +5,14 @@
  */
 
 import { Component, EventEmitter, Input, OnInit, Output } from '@angular/core';
+import { UntypedFormControl, UntypedFormGroup } from '@angular/forms';
 import { ICON_TYPE } from '../../../cms-components/misc/icon/icon.model';
 
+export interface SaveCardEvent {
+  saveMode: boolean;
+  [key: string]: string | any;
```
Uli-Tiger, line 13: why not use an array of objects similar to the input
`paragraphs: Array<{title:string; text:string, hasChanged:boolean}>`

feature-libs/quote/components/details/overview/quote-details-overview.component.ts
```diff
@@ -5,24 +5,181 @@
...
+      const index = saveCardEventPropertyNames.findIndex((element) =>
+        element.includes(name + '_')
+      );
+      if (index >= 0) {
+        const propertyName = saveCardEventPropertyNames[index];
+        metaData[name] = saveCardEvent[propertyName];
+      }
+    });
+    return metaData;
+  }
```
Uli-Tiger, line 73: As already mentioned here: https://github.com/SAP/spartacus/pull/17665/files#r1282305889 we should try to define and fill the saveCardEvent in a way, that we can easily extract/build the quote meta data.

PR #17666, opened 2023-07-20
test: SSR E2E Test Scenarios

Zeyber:
Closes: https://jira.tools.sap/browse/CXSPA-3780


ssr-e2e/proxy-test.spec.js
```diff
@@ -0,0 +1,176 @@
...
+}
+
+function getLogMessages() {
+  const data = fs.readFileSync(SSR_LOG_PATH);
+  console.log(data, data.toString()); // TODO: Temp read log in ci
+  const messages = data
+    .toString()
+    .split('\n')
+    .filter((text) => text.indexOf('"message":') > -1)
+    .map((text) => text.split('":"')[1].split('",')[0]);
```
Platonn: for readability and maintainability:
let's do JSON.parse(text)

(that will work only when JSON is single-line, not prettified -> so app has to be built prod mode)
you may leave a code comment about it, as well

.github/workflows/ci-continuous-integration.yml
```diff
@@ -46,6 +46,40 @@ jobs:
...
+        run: npm run build:libs && npm run build && npm run build:ssr
+      - name: Start SSR Server
+        run: npm run serve:ssr > ssr-e2e/ssr.log &
+      - name: Expose dirs    
+        run: |
+          ls ./ &&
+          ls ../ &&
+          ls ./ssr-e2e
+      - name: Run SSR tests
+        run: npm run test:ssr
```
Platonn: As discussed offline:
Now the long-running SSR server instance is shared among many test specs.
This is prone to specs interfering with each other via the shared SSR server instance, which may lead to unexpected false-positives and false-negatives, depending on the behavior of SSR server (which is a subject of implementation, not tests).

If possible, it would be good to isolate specs, by starting and killing SSR server for each spec separately - so no server instance is shared among 2 specs. I'd be happy to learn if we can do it without increasing significantly the total time of running all the specs.

If the total execution time of e2e is big, we might consider to NOT run them on every PR, but only on `develop` branch (like other test suites which are not crucial or time-costly).

ssr-e2e/proxy-test.spec.js
```diff
@@ -0,0 +1,176 @@
...
+  host: 'localhost',
+  port: 4000,
+};
+
+describe('SSR E2E', () => {
+  let server;
+
+  beforeAll(async () => {
+    // Hacky way to make sure server starts
+    await new Promise((res) => setTimeout(res, 30000));
```
Platonn: As we discussed offline, it would be good to avoid big magic numbers for waiting until something happens. Instead it would be good to react to a specific event that we're interested in (e.g. the server actually started).

ssr-e2e/proxy-test.spec.js
```diff
@@ -0,0 +1,176 @@
...
+    });
+    const response = await sendRequest('/');
+    expect(response.statusCode).toEqual(200);
+
+    assertMessages([
+      'Rendering started (/)',
+      'Request is waiting for the SSR rendering to complete (/)',
+      'Rendering completed (/)',
+      'Render from cache (/)',
+    ]);
```
Platonn: As discussed offline:
The current implementation works great for proof of concept - we're checking altogether what's in the log file.
1. For maintenance reasons, it would be good to allow for asserting whether the logs just _contain_ some message. Why? In the future we the app might log new things unrelated to this test spec. And it would make this spec fail.
2. Moreover, the current solution forces us to include in the asertion the messages coming from previous specs. It would be good to assert only what's related to the current spec

ssr-e2e/proxy-test.spec.js
Zeyber: I have attempted to utilize `JSON.parse` but I have had difficulty formatting the log file cleanly. You can see from a typical log file that not all lines are formatted in a JSON-like way which means we need a bunch of logic to reformat the log anyhow.
Typical log file:
image

Zeyber: Kris's idea to break by line and JSON parse when starts with `{`.

.github/workflows/ci-merge-checks.yml
```diff
@@ -150,8 +150,34 @@ jobs:
...
+        with:
+          node-version: ${{ env.NODE_VERSION }}
+      - name: Cache node_modules
+        id: cache-node-modules
+        uses: actions/cache@v2
+        with:
+          path: | 
+            node_modules
+            projects/storefrontapp-e2e-cypress/node_modules
+            ~/.cache/Cypress
```
Platonn: [QUESTION]
Is caching cypress stuff really needed? I believe we don't use cypress in the SSR tests?

Note: similar question for the file [ci.yml](https://github.com/SAP/spartacus/pull/17666/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03f)

projects/ssr-tests/jest.config.js
```diff
@@ -0,0 +1,35 @@
+const { pathsToModuleNameMapper } = require('ts-jest');
+const { compilerOptions } = require('./tsconfig.json');
+const { defaultTransformerOptions } = require('jest-preset-angular/presets');
+
+/** @type {import('ts-jest/dist/types').JestConfigWithTsJest} */
```
Platonn, line 5: Nice

projects/ssr-tests/package.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "ssr-tests",
+  "description": "Spartacus SSR Tests",
+  "keywords": [
+    "spartacus",
+    "ssr",
+    "tests"
+  ],
+  "author": "SAP, Spartacus team",
+  "license": "MIT",
```
Platonn:
```suggestion
  "license": "Apache-2.0",
```

projects/ssr-tests/src/log.utils.ts
```diff
@@ -0,0 +1,74 @@
...
+
+/**
+ * Reads log and returns messages as string array.
+ */
+export function getLogMessages(): string[] {
+  const data = fs.readFileSync(SSR_LOG_PATH).toString();
+  return data
+    .toString()
+    .split('\n')
+    .filter((text: string) => text.charAt(0) === '{')
```
Platonn: It's not obvious why we filter lines starting with `{`, so IMHO it would be good to leave a code comment:

```suggestion
    // We're interested only in JSON logs from Spartacus SSR app.
    // We ignore plain text logs coming from other sources, like `Node Express server listening on http://localhost:4200`
    .filter((text: string) => text.charAt(0) === '{')
```

projects/ssr-tests/src/log.utils.ts
```diff
@@ -0,0 +1,74 @@
...
+export function assertMessages(expected: string[]): void {
+  const messages = getLogMessages();
+  for (const message of expected) {
+    expect(messages).toContain(message);
+  }
+}
+
+/**
+ * Check log every interval to see if log contains text.
+ * Keeps waiting until log contains test or test times out.
```
Platonn: Typo?
```suggestion
 * Keeps waiting until log contains text or test times out.
```

projects/ssr-tests/src/ssr-testing_spec.ts
Platonn, line 1: [NITPICK]
`_spec.ts` vs `.spec.ts` ?

projects/ssr-tests/src/ssr-testing_spec.ts
```diff
@@ -0,0 +1,85 @@
...
+      `Rendering started (${REQUEST_PATH})`,
+      `Request is waiting for the SSR rendering to complete (${REQUEST_PATH})`,
+    ]);
+  });
+
+  // Wait for SSR server to complete rendering.
+  it('should complete rendering within a minute', async () => {
+    await Log.waitUntilLogContainsText(`Rendering completed (${REQUEST_PATH})`);
+    Log.assertMessages([`Rendering completed (${REQUEST_PATH})`]);
+  });
```
Platonn: [MINOR]
Do I understand correctly, that we have a problem that in SSR we have a short `timeout` option (3s ?) after which a fallback to CSR happens, but the rendering of the homepage in practice takes more than 3s, right? 

It's too bad that this configuration issue is somehow reflected in our test specs, because it's not related to what we want to test.

Did you consider alternative approaches? E.g. increasing the `timeout` in SSR options? I hope a custom timeout value could be passed via Node env variable (not sure whether during the build time or when running ssr - `build:ssr` vs `run:ssr`):
https://github.com/SAP/spartacus/blob/2e6a1a41d6bb1886b52188d438446d55ed84f534/projects/storefrontapp/server.ts#L25-L28

projects/ssr-tests/src/ssr-testing_spec.ts
```diff
@@ -0,0 +1,85 @@
...
+const BACKEND_BASE_URL: string = process.env.CX_BASE_URL || '';
+
+describe('SSR E2E', () => {
+  let proxy: any;
+  const REQUEST_PATH = '/electronics-spa/en/USD/';
+
+  beforeAll(async () => {
+    Log.clearSsrLogFile();
+    await Ssr.startSsrServer();
+  });
```
Platonn: [MEDIUM]
I believe, my intention was to start SSR server separately for _each_ spec (not have one shared SSR instance for all specs - because it's prone to specs interfering with each other, e.g. because some leftover side-effect behaviours after the previous request, that might affect the next request's handling).

projects/ssr-tests/src/ssr-testing_spec.ts
```diff
@@ -0,0 +1,85 @@
...
+    });
+    proxy.on('proxyRes', function (proxyRes: any) {
+      proxyRes.statusCode = 500;
+    });
+    const response: any = await ProxyServer.sendRequest('/');
+    expect(response.statusCode).toEqual(500);
+  });
+
+  // Note: Currently, the ssr server still responds with 200 quickly despite the proxy delay
+  xit('should receive 500 error response with timed-out request', async () => {
```
Platonn: [MEDIUM]
What about leaving `//TODO` comments before each `xit()`, so SonarQube won't let us accidentally merge the Epic branch "SSR error handling" to `develop` until we write those tests. Does it make sense?

projects/ssr-tests/src/ssr.utils.ts
```diff
@@ -0,0 +1,36 @@
...
+ * Start an ssr server instance at the given port (default 4000).
+ * The server will output a log file at the test project root named "ssr.log".
+ * Funtion finishes once the server is initialized.
+ */
+export async function startSsrServer(port = 4000) {
+  childProcess.exec(
+    `NODE_TLS_REJECT_UNAUTHORIZED=0 PORT=${port} npm run serve:ssr --prefix ../../> ssr.log`
+  );
+  await Log.waitUntilLogContainsText(`Node Express server listening on `);
+}
```
Platonn, line 24: Very nice. I like how it's encapsulated in one util `await startSsrServer()`

projects/ssr-tests/src/ssr-testing_spec.ts
Platonn, line 1: I love how you made those specs so easy to read and ergonomic to write! Those test objects `Log`, `Ssr`, `ProxyServer` having functions with super clear semantics and a handy async-await API - they look very nice!

projects/ssr-tests/src/ssr-testing_spec.ts
Zeyber: I removed this test as its not really checking anything important. It was to make sure the server was doing and completing rending so that we could check a cache response is returned in a following test. But I think we can move this logic to the next test which is scoped in this [ticket](https://jira.tools.sap/browse/CXSPA-4742).

Platonn (dismissed):

PR #17669, opened 2023-07-24
introduce guard to change routing targets if quote cart

ChristophHi:
is active


ChristophHi:
Instead https://github.com/SAP/spartacus/pull/17721 is merged

PR #17670, opened 2023-07-24
feat: item level commnets for quotes [CXSPA-3676]

Uli-Tiger:


Uli-Tiger:
@ChristophHi: Regarding the one minor [sonar issue](https://sonar.tools.sap/project/issues?pullRequest=17670&resolved=false&types=CODE_SMELL&inNewCodePeriod=true&id=composable-storefront). Fixing this as suggested would require a change in the `tsconfig` file. one would have to add lib `DOM.Iterable`. hence i suggest to make as false positive/wont fix.

feature-libs/quote/components/details/comment/quote-details-comment.component.ts
```diff
@@ -70,6 +77,22 @@ export class QuoteDetailsCommentComponent {
       );
   }
 
+  onItemClicked(event: { item: Item }) {
+    this.quoteDetailsCartService.setQuoteEntriesExpanded(true);
+    this.quoteDetailsCartService
+      .getQuoteEntriesExpanded()
+      .pipe(take(1), delay(0)) // delay this task until the UI has expanded
+      .subscribe(() => {
+        const aTags = this.document.getElementsByTagName('a');
```
ChristophHi:
```suggestion
       const aTags = Array.from(this.document.getElementsByTagName('a'));
```
@Uli-Tiger 
Then you can avoid the sonar issue

feature-libs/quote/components/details/comment/quote-details-comment.component.ts
```diff
@@ -70,6 +77,22 @@ export class QuoteDetailsCommentComponent {
...
+        for (let ii = 0; ii < aTags.length; ii++) {
```
ChristophHi:
```suggestion
           for (let aTag of aTags) {
```

ChristophHi (changes requested):

feature-libs/quote/components/details/cart/quote-details-cart.component.ts
```diff
@@ -17,8 +18,15 @@ import { Observable } from 'rxjs';
...
   readonly cartOutlets = CartOutlets;
+  showCart$ = this.quoteDetailsCartService.getQuoteEntriesExpanded();
 
-  constructor(protected quoteFacade: QuoteFacade) {}
+  constructor(
+    protected quoteFacade: QuoteFacade,
+    protected quoteDetailsCartService: QuoteDetailsCartService
+  ) {}
+
+  onToggleShowCart(showCart: boolean) {
```
Stofftiger: maybe rename function into "onToggleShowOrHideCart"

also the showCart could be renamed isCartShown

feature-libs/quote/core/facade/quote.service.spec.ts
```diff
@@ -266,6 +267,33 @@ describe('QuoteService', () => {
       });
   });
 
+  it('should call addQuoteComment command when called with empty string a entry number', () => {
```
Stofftiger: shouldn't this be "with empty string instead of an entry number" ?

feature-libs/quote/core/facade/quote.service.spec.ts
```diff
@@ -266,6 +267,33 @@ describe('QuoteService', () => {
...
+    service
+      .addQuoteComment(mockQuote.code, mockComment, '')
```
Stofftiger, line 272: since the entrynumber is optional, is the '' parameter even needed?

Uli-Tiger: the previous test  is testing the case without the optional parameter. Here we want to explicitly test the behavior for an empty string.

feature-libs/quote/components/details/cart/quote-details-cart.service.ts
```diff
@@ -0,0 +1,26 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { Injectable } from '@angular/core';
+import { Observable, ReplaySubject } from 'rxjs';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class QuoteDetailsCartService {
```
ChristophHi: Please rename to QuoteDetailsCartComponentService in file quote-details-cart.component.service.ts.
That's a kind of best practise for services that serve components

feature-libs/quote/components/details/cart/quote-details-cart.component.html
```diff
@@ -1,19 +1,25 @@
-<div class="cart-toggle" (click)="showCart = !showCart">
-  <cx-icon
-    aria-hidden="false"
-    [type]="showCart ? iconTypes.CARET_UP : iconTypes.CARET_DOWN"
-  >
-  </cx-icon>
-  <span class="cart-toggle-text">{{ 'quote.commons.cart' | cxTranslate }}</span>
-</div>
+<ng-container *ngIf="{ showCart: showCart$ | async } as state">
```
ChristophHi, line 1: Why not simply
```suggestion
<ng-container *ngIf="(showCart$ | async ) as showCart">
```

feature-libs/quote/components/details/cart/quote-details-cart.component.html
```diff
@@ -1,19 +1,25 @@
...
+  <div class="cart-toggle" (click)="onToggleShowOrHideCart(state.showCart)">
```
ChristophHi:
```suggestion
  <div class="cart-toggle" (click)="onToggleShowOrHideCart(showCart)">
```

feature-libs/quote/components/details/cart/quote-details-cart.component.html
```diff
@@ -1,19 +1,25 @@
...
+    <cx-icon
+      aria-hidden="false"
+      [type]="state.showCart ? iconTypes.CARET_UP : iconTypes.CARET_DOWN"
```
ChristophHi, line 5:
```suggestion
      [type]="showCart ? iconTypes.CARET_UP : iconTypes.CARET_DOWN"
```

feature-libs/quote/components/details/cart/quote-details-cart.component.html
```diff
@@ -1,19 +1,25 @@
...
 
-<ng-container *ngIf="showCart && (quoteDetails$ | async) as quoteDetails">
-  <ng-template
-    [cxOutlet]="cartOutlets.CART_ITEM_LIST"
-    [cxOutletContext]="{
-      items: quoteDetails.entries,
-      readonly: true
-    }"
+  <ng-container
+    *ngIf="state.showCart && (quoteDetails$ | async) as quoteDetails"
```
ChristophHi, line 14:
```suggestion
    *ngIf="showCart && (quoteDetails$ | async) as quoteDetails"
```

feature-libs/quote/components/details/comment/quote-details-comment.component.html
```diff
@@ -14,6 +14,7 @@
       [messageEvents$]="messageEvents$"
       [messagingConfigs]="messagingConfigs"
       (send)="onSend($event, quoteDetails.code)"
```
ChristophHi, line 16: strict issue

feature-libs/quote/components/details/comment/quote-details-comment.component.html
```diff
@@ -14,6 +14,7 @@
...
+      (itemClicked)="onItemClicked($event)"
```
ChristophHi, line 17: strict issue

feature-libs/quote/components/details/comment/quote-details-comment.component.ts
```diff
@@ -79,27 +102,77 @@ export class QuoteDetailsCommentComponent {
         map((quote) => quote.isEditable)
       ),
       dateFormat: 'MMMM d, yyyy h:mm aa',
+      defaultItemId: ALL_PRODUCTS_ID,
+      itemList$: this.prepareItemList(),
     };
   }
 
+  private prepareItemList(): Observable<Item[]> | undefined {
```
ChristophHi: make protected please

feature-libs/quote/components/details/comment/quote-details-comment.component.ts
```diff
@@ -79,27 +102,77 @@ export class QuoteDetailsCommentComponent {
...
+          if (item) {
+            itemList.push(item);
+          }
+        });
+        return itemList;
+      })
+    );
+  }
+
+  private convertToItem(entry: OrderEntry | undefined): Item | undefined {
```
ChristophHi:
```suggestion
  protected convertToItem(entry: OrderEntry | undefined): Item | undefined {
```

feature-libs/quote/components/details/comment/quote-details-comment.component.ts
```diff
@@ -79,27 +102,77 @@ export class QuoteDetailsCommentComponent {
...
+    if (entry?.entryNumber !== undefined) {
```
ChristophHi: Do not return undefined. We always expect an entryNumber, so throw an error if it is not provided

ChristophHi (changes requested):

feature-libs/quote/components/details/cart/quote-details-cart.component.html
Uli-Tiger: In case `showCart=false` the container would not be rendered, hence even the toggle, which is inside would not be rendered, which is not desired.
On the other hand in case of `{showCart: false}` the container is still rendered.

feature-libs/quote/components/details/comment/quote-details-comment.component.html
Uli-Tiger: what exactly? - code is a mandatory property of quote

feature-libs/quote/components/details/comment/quote-details-comment.component.html
Uli-Tiger: what exactly?

ChristophHi: Argument of type '{ item?: Item | undefined; }' is not assignable to parameter of type '{ item: Item; }'.
  Types of property 'item' are incompatible.
    Type 'Item | undefined' is not assignable to type 'Item'.
      Type 'undefined' is not assignable to type 'Item'

feature-libs/quote/components/details/comment/quote-details-comment.component.html
ChristophHi: Argument of type '{ files: File | undefined; message: string; itemId?: string | undefined; }' is not assignable to parameter of type '{ message: string; itemId: string; }'.
  Types of property 'itemId' are incompatible.
    Type 'string | undefined' is not assignable to type 'string'.
      Type 'undefined' is not assignable to type 'string'

feature-libs/quote/components/details/cart/quote-details-cart.component.html
ChristophHi: @Uli-Tiger But then still I doubt that 
<ng-container *ngIf="{ showCart: showCart$ | async } as state">
exactly does what you intend. Because state will be instantly defined, right? (With showCart = null as long as the async pipe has not completed)

Uli-Tiger: Later we use showCart in boolean expressions, hence  `null` will be treated as `false`. So the cart is not rendered unless the async pipe completes, which is the desired behavior.
see also: https://stackoverflow.com/questions/58799165/angular-how-to-pass-boolean-async-variable-in-template-function

ChristophHi: My point: The ngIf in line #1 is misleading, right? Because this will instantly be defined.

ChristophHi: Just did a bit of searching: Yes, this seems to be the standard way, no other directive available. Even if it's a bit weird. So: ok from my side

ChristophHi (approved):

Issue #5969, opened 2020-01-02
NavigationNodes(cx-category-navigation) with CMS restrictions(e.g. loggedInUser) not shown until a Browser refresh


gitshamster:
## Environment Details

- Spartacus:  1.3.0 (but should be the same in 1.4.0-RC.0
- Browser: Firefox 64.0.2 , Chrome 79.0.3945.88
- OS: Windows 10 1909
- Device: Dell Precision 7530
- Other environment details: Angular 8.2.9

## Steps to Reproduce
1. Configure one of the existing Navigation Links with a user restriction ('Logged in User') in BackOffice or SmartEdit. e.g. CamerasFlashesCategoryLink
2. The link is not shown for anonymous user in Spartacus. 
3. Login with registered user
4. The link is still not visible.
![inititial-screen_with_user_restrictions_noLink](https://user-images.githubusercontent.com/29032953/71680431-5ac7c600-2d8a-11ea-81d2-baa93dfa797e.png)

## Observed Results
 Even though I could see the Link value in the JSON responses, the same is not rendered in the Category Header of SPA.

The intended Navigation Link is rendered only if I refresh the browser window

## Expected Results
After login. The user should be able to see the Navigation Link (for example - CamerasFlashesCategoryLink )
While there are related bugs on this topic - #4603 and #3769 

It would be great if we could have some hint to a temporary solution while waiting for a fix in a future release

## Additional Information - Things tried out so far: 
1. Overritten the category-navigation-component and printed time-stamp to see if the component is actually updated. The time-stamp is printed correctly, but yet the ''CamerasFlashesCategoryLink'' is still not shown in the Category Header.

2 . I have tried with calling cmsService.refreshPage() and cmsService.refreshComponent('CamerasFlashesCategoryLink') but that too did not succeed

Sample code
`this.auth.isUserLoggedIn().pipe(
      distinctUntilChanged(),
      filter(( isLoggedIn: boolean ) => {
          return isLoggedIn === true;
      }),
      map((event: boolean) => {
        console.log('Login Changes in SwitchMap..');
        this.categoryNavigationService.setLoadPage(true);
        return event;
      })
    ).subscribe(
      (loggedIn: boolean) => {
        console.log('Login Changes..');
        this.reloadPage = loggedIn;
        //this.cmsService.refreshComponent('NavigationBar');
        //this.cmsService.refreshComponent('CategoryNavigationComponent');
[...]

PR #17673, opened 2023-07-24
fix(navigation): prevent updating ngrx state when no components received from backend

Platonn:
### When the bug happens

The edge-case bug happens when OCC returns inconsistent responses among `/pages` API vs `/components` API. In particular, when OCC `/pages` API returns at least one `CmsLink` component (in the cms navigation) that the OCC `/components` API has no idea about (so `/components` API returns empty response when asked for details of this CmsLink component id), then Spartacus `NavigationService` goes into infinite loop trying to load details of this `CmsLink` component over and over again.

When this edge-case could happen in reality: when the cached APIs `/pages` and `/components` had cache different invalidation times. The following is a detailed example:
- Customer is caching all OCC APIs behind a proxy cache
- their CMS business user removed some `CmsLink` from the `CmsNavigation`
- a little later a race condition happened: the cache of `/pages` API was NOT YET invalidated, but the cache of `/components` API was ALREADY invalidated. This means, that the stale `/pages` API was still returning ID of the removed `CmsLink` component as part of CMS page structure. But the up-to-date `/components` API asked in a followup call for details of this component, didn't return any data, because this component no longer existed (was removed).
- Then the SSR was rendering some page and it called the OCC `/pages` API, and received in (stale cached) response the ID of this removed CmsLink (among also other navigation component IDs as usual)
Then during the same render SSR called the endpoint `/components` asking for the details of this `CmsLink`, but the (up-to-date) API didn't return any details of this component.
- Then Spartacus went into an infinite loop, trying to load the details of this missing `CmsLink` component via the `/components` API over and over again.

Note: such an infinite loop could happen regardless whether Spartacus runs in the browser (CSR) or in NodeJS (SSR).

### Why Spartacus needs to load missing components?

One could ask: Why Spartacus bothers to load missing components from the `/components` API? Answer: there is a valid business usecase for it - components that are hidden for anonymous users (due to a CMS restriction mechanism), but visible for logged in users are considered "missing" when the user logs in, and their details need to be loaded after the user logs in. For more, see the old Spartacus github bug ticket:  [NavigationNodes(cx-category-navigation) with CMS restrictions(e.g. loggedInUser) not shown until a Browser refresh #5969](https://github.com/SAP/spartacus/issues/5969)., which introduced the [logic of loading missing cms components' details](https://github.com/SAP/spartacus/commit/1d1ebd328f3a627cbee758c3139fa244ab8e86ec#diff-379fc51d89107b1307dc16acb153893ef0de860ee04fe422a6f5549a155b5bbeR52-R67).

### Why this PR prevents infinite loop?

Short answer: now we don't update ngrx state, when receiving empty response from `/components` OCC API. This breaks the circuit: load missing cms components -> received empty response -> _update ngrx state_ -> load missing cms components -> ...  

Explanation: From Spartacus point of view, when `/pages` API returns IDs of components that are in the cms structure, then Spartacus tries to load details of all those components via `/components` API. Currently, when `/components` API returns an empty response for the given ID, the ngrx state object reference is updated despite that this empty response didn't actually changed the ngrx state - it's only the state's object reference that changed for no good reason. Such a change of object's reference triggers an emission of RxJs observable with this state. This leads to emiting a new value from the `$data` observable in the stream returned by `NavigationService.getNavigationNode()`. Inside the logic of this stream, Spartacus recognizes that some components's details are still missing, and therefore tries to load them by calling `CmsService.loadNavigationItems(navigation.id, missingItems)` - see the source code of the branches [release/4.3.1](https://github.com/SAP/spartacus/blob/release/4.3.x/projects/storefrontlib/cms-components/navigation/navigation/navigation.service.ts#L64-L69) and [develop-6.3.x](https://github.com/SAP/spartacus/blob/develop-6.3.x/projects/storefrontlib/cms-components/navigation/navigation/navigation.service.ts#L75-L80).

In order to break the infinite loop, it suffices that we fix the Spartacus ngrx `navigation-item-entry.reducer.ts` (see its source code, which is the same in both branches: [release/4.3.x](https://github.com/SAP/spartacus/blob/release/4.3.x/projects/core/src/cms/store/reducers/navigation-entry-item.reducer.ts#L6-L35) and [develop-6.3.x)](https://github.com/SAP/spartacus/blob/develop-6.3.x/projects/core/src/cms/store/reducers/navigation-entry-item.reducer.ts#L12-L41). The fixed version should check for the `length` of the `components` payload - if it's empty, then no state change should happen.

fixes https://jira.tools.sap/browse/CXSPA-4157


Platonn:
### QA steps

We need to simulate that OCC `/pages` API is returning ID of some CMS component, but the `/components` API should return empty response when being asked for this component ID. In the case of customer it can happen due to different timings of invalidation of the OCC cache in between the endpoints `/components` (earlier cache invalidation) vs. `/pages` (late cache invalidation).
In our case for local development it suffices to implementing a custom Angular `HTTP_INTERCEPTOR` which never returns an information about some specific CMS component. The following is an example code of such interceptor implemented e.g. in the file `app.module.ts`: https://gist.github.com/Platonn/e70928786c8267dd40173b0d65608e08
Then open the storefront and observe in the ChromeDevTools Network tab the infinite loop of the http calls to `/components` endpoint.

projects/core/src/cms/store/reducers/navigation-entry-item.reducer.ts
```diff
@@ -24,9 +24,7 @@ export function reducer(
               [`${component.uid}_AbstractCMSComponent`]: component,
             };
           },
-          {
-            ...{},
-          }
```
Platonn, line 29: this code cleanup is done by the way. it has nothing to do with the main purpose (bugfix) of this PR

projects/core/src/cms/store/reducers/navigation-entry-item.reducer.ts
```diff
@@ -15,7 +15,7 @@ export function reducer(
 ): NodeItem | undefined {
   switch (action.type) {
     case CmsActions.LOAD_CMS_NAVIGATION_ITEMS_SUCCESS: {
-      if (action.payload.components) {
+      if (action.payload.components.length) {
```
Zeyber, line 18: Should this have `?.length`  so that there is no error in the case `components` is not in the payload?

projects/storefrontapp-e2e-cypress/cypress/e2e/regression/cms/cms-navigation.core-e2e.cy.ts
```diff
@@ -0,0 +1,47 @@
...
+    cy.wait('@getComponents');
+    cy.get('cx-category-navigation cx-generic-link').should('exist');
+
+    // `/cms/components` endpoint is called for each different Navigation components (e.g. header and footer)
+    const EXPECTED_MAX_COUNT_LOADING_CMS_COMPONENTS = 3;
+
+    // Our heuristic to assert that there is no infinite loop of calling `/cms/components` endpoint
+    // is to check whether this endpoint was called not too many times during 1st second after the page is rendered.
+    cy.wait(1_000);
+    cy.get('@getComponents.all', { timeout: 0 }).should(
```
Zeyber: I am noticing that this test will fail at this assertion even though the page is no longer looping requests following the QA steps with the fix (tested on `develop` and `fix` branches). Should we increase the expected count/circuit breaker limit in case there are pages that load make than 3 component requests?

Zeyber (changes requested):

projects/core/src/cms/store/reducers/navigation-entry-item.reducer.ts
Platonn: comopnents is a required property

projects/storefrontapp-e2e-cypress/cypress/e2e/regression/cms/cms-navigation.core-e2e.cy.ts
```diff
@@ -0,0 +1,47 @@
...
+        return res;
+      });
+    }).as('getComponents');
+
+    cy.visit('/');
+    cy.wait('@getComponents');
+    cy.get('cx-category-navigation cx-generic-link').should('exist');
+
+    // `/cms/components` endpoint is called for each different Navigation components (e.g. header and footer)
+    const EXPECTED_MAX_COUNT_LOADING_CMS_COMPONENTS = 3;
```
Platonn:
```suggestion
    const EXPECTED_MAX_COUNT_LOADING_CMS_COMPONENTS = 4;
```

projects/storefrontapp-e2e-cypress/cypress/e2e/regression/cms/cms-navigation.core-e2e.cy.ts
Platonn: Pasting here my explanation from Slack, just for reference:

> 
> I had problem to choose the right expected amount of calls to /components
> 
> 1. it's called once per every navigation - we have 2 : top nav and footer nav
> 2. it's called once per a set of missing components in those navigations - according to my mocked response I deliberately remove 1 component from the response: it likely happens both for the call for top nav and call for bottom nav
> 
> therefore there are 4 requests
> 
> I don't know why I wrote 3
> 
> alternatively I could change my test to deliberately remove 1 component of a specific uid, e.g. CanonBrandCategoryLink
> 
> but then the tests would depend hardly on sample data
> 
> currently it depends not that hard on sample data - however it depends on the fact that the CMS structure returned from /pages endpoint cotains only 2 navigations
> 
> if there where 3 navigation menus (top, footer, and something else), then there would be 6 calls to /components in total (3 normal ones, and additional 3 with attempt to load missing components)
> 
> I was scratching my head what would be a good expected number
> 
> I could go for:
> - wait 5-10 seconds
> - and the number of calls should be less than 20 (with infinite loop it is more than 100 by that time)
> 
> bud decided to go:
> - wait 1 second (because it's not that much... we can afford it)
> - and the number of calls should be exact  as expected  (with infinite loop, during this 1st second there can happen only a few calls to /components)

Zeyber (approved):

PR #17681, opened 2023-07-25
fix: mini cart stuck with value 0

FollowTheFlo:
add totalUnitCount param in cart related endpoints
It fixes 'item count' in Mini cart, saved-cart-list, saved-cart-details


kimhw0630:
The code looks fine to me..
e2e was failing due to ASM(B2C) and Account checkout (B2C)
I tried to test in S7 but it seems the server has a problem so I couldn't test it.( getting 400 when trying to add a product to the cart.. tried both above code change and without change)

kimhw0630 (dismissed):

PR #17687, opened 2023-07-27
Feature/cxspa 4110 quote cart

ChristophHi:


feature-libs/quote/occ/converters/occ-quote-action-normalizer.ts
```diff
@@ -52,13 +59,35 @@ export class OccQuoteActionNormalizer implements Converter<OccQuote, Quote> {
...
     const order = this.quoteConfig.quote?.actions?.actionsOrderByState?.[state];
-
-    return order
-      ? list
-          .filter((item) => order.includes(item))
-          .sort((a, b) => order.indexOf(a) - order.indexOf(b))
-      : list;
+    if (order) {
+      //deep copy order list
+      const clonedActionList = order.map((actionType) => actionType);
```
pawelfras: 'minor': consider using the new global method `structuredClone` that has been introduced for deep copying.
For more, see: https://developer.mozilla.org/en-US/docs/Web/API/structuredClone

feature-libs/quote/quote-aware/component/quote-aware.component.ts
```diff
@@ -0,0 +1,12 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { Component } from '@angular/core';
+
+@Component({
+  templateUrl: './quote-aware.component.html',
+})
+export class QuoteAwareComponent {}
```
pawelfras, line 12: `query:` Can you remind me what was the idea behind this component and related logic?

feature-libs/quote/root/guards/quote-cart.service.ts
```diff
@@ -0,0 +1,37 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { Injectable } from '@angular/core';
+import { Observable, ReplaySubject } from 'rxjs';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class QuoteCartService {
```
pawelfras, line 15: `minor`: I propose creating a ticket about removing this service once the backend is ready and adding comments with the ticket number to all related places. It may be useful in terms of tracking and refactoring.

feature-libs/quote/root/guards/quote-cart.service.ts
```diff
@@ -0,0 +1,37 @@
...
+  private quoteId: Observable<string> = new ReplaySubject<string>(1);
+  private quoteCartActive: Observable<boolean> = new ReplaySubject<boolean>(1);
```
pawelfras: `minor`: Instead of casting this field each time you want to use `ReplaySubject` methods, I propose reversed approach - keep the `ReplaySubject` type and update the stream each time you need, but when you want some method returns an observable that can't be modified, use `asObservable`:

```ts
private quoteCartActive = new ReplaySubject<boolean>(1);

 [...]

 public setQuoteId(quoteId: string): void {
    this.quoteId.next(quoteId);
  }

  public getQuoteId(): Observable<string> {
    return this.quoteId.asObservable();
  }
```

ChristophHi:
Instead I have merged https://github.com/SAP/spartacus/pull/17721 which is based on this one

PR #17701, opened 2023-07-28
fix: product image normalizer now properly checks for gallery index

znikola:
closes https://jira.tools.sap/browse/CXSPA-4181

PR #17706, opened 2023-07-28
fix: Product image normalizer now properly checks for gallery index - 6.3.x

bgambocjaviniar:
taken from https://github.com/SAP/spartacus/pull/17701 

and closes closes https://jira.tools.sap/browse/CXSPA-4181


bgambocjaviniar:
nvm, the referenced ticket was pointing to 6.3.x instead of only 6.4.x 

made a mistake.

PR #17726, opened 2023-08-03
fix: modify install script to now correctly install spartacus versions for features

RadhepS:
closes: https://jira.tools.sap/browse/CXSPA-3568


FollowTheFlo (approved):
good to me. tested succesfully.

PR #17731, opened 2023-08-03
feat: quote item list shall be editable when quote is switched to edit mode [CXSPA-4231]

Uli-Tiger:
When quote is in edit mode, one can update the item quantity as well as remove any item from the quote cart.

### DIFF
--- a/feature-libs/cart/base/components/cart-shared/cart-item-list/cart-item-list.component.ts
+++ b/feature-libs/cart/base/components/cart-shared/cart-item-list/cart-item-list.component.ts
@@ -61,6 +61,7 @@ export class CartItemListComponent implements OnInit, OnDestroy {
   @Input() cartId: string;
 
   protected _items: OrderEntry[] = [];
+  protected _foreReRender: boolean = false;
   form: UntypedFormGroup = new UntypedFormGroup({});
 
   @Input('items')
@@ -106,6 +107,7 @@ export class CartItemListComponent implements OnInit, OnDestroy {
 
   protected getInputsFromContext(): Subscription | undefined {
     return this.outlet?.context$.subscribe((context) => {
+      console.log(context);
       if (context.readonly !== undefined) {
         this.readonly = context.readonly;
       }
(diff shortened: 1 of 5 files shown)


feature-libs/quote/components/details/cart/quote-details-cart.component.ts
```diff
@@ -4,28 +4,46 @@
...
 
+  ngOnInit(): void {
+    // if anything in the cart changes, reload the quote
+    this.subscription = this.eventService.get(CartEvent).subscribe(() => {
+      this.eventService.dispatch({}, QuoteDetailsReloadQueryEvent);
+    });
+  }
+
+  ngOnDestroy(): void {
+    this.subscription?.unsubscribe();
```
ChristophHi:
```suggestion
    this.subscription.unsubscribe();
```

feature-libs/cart/base/components/cart-shared/cart-item-list/cart-item-list.component.ts
```diff
@@ -61,6 +61,7 @@ export class CartItemListComponent implements OnInit, OnDestroy {
   @Input() cartId: string;
 
   protected _items: OrderEntry[] = [];
+  protected _foreReRender: boolean = false;
```
ChristophHi:
```suggestion
  protected _forceReRender: boolean = false;
```

feature-libs/cart/base/components/cart-shared/cart-item-list/cart-item-list.component.ts
```diff
@@ -119,6 +120,7 @@ export class CartItemListComponent implements OnInit, OnDestroy {
         this.cartId = context.cartId;
       }
       if (context.items !== undefined) {
+        this._foreReRender = true;
```
ChristophHi: That means we re-render _all_ entries whenever they are provided.
Shouldn't we set forceReRender only if readOnly changes?

ChristophHi (changes requested):

feature-libs/cart/base/components/cart-shared/cart-item-list/cart-item-list.component.ts
Uli-Tiger: i think you are right - it's sufficient to do if the read-only flag changes

ChristophHi (approved):

PR #17734, opened 2023-08-04
feature/seller-edit

ChristophHi:


feature-libs/quote/components/details/cart/summary/quote-details-cart-summary.component.spec.ts
```diff
@@ -0,0 +1,117 @@
...
+import { BehaviorSubject, EMPTY, Observable, of } from 'rxjs';
+
+import { QuoteDetailsCartSummaryComponent } from './quote-details-cart-summary.component';
+import createSpy = jasmine.createSpy;
+import { LAUNCH_CALLER, LaunchDialogService } from '@spartacus/storefront';
+import { ElementRef, ViewContainerRef } from '@angular/core';
+import { createEmptyQuote } from '../../../../core/testing/quote-test-utils';
+
+const mockCartId = '1234';
+const mockCode = '3333';
```
Uli-Tiger: just a little remark to naming - these are constants representing test data, but not mocks, hence i  suggest to not prefix the const with 'mock'.

feature-libs/quote/components/details/cart/summary/quote-details-cart-summary.component.spec.ts
```diff
@@ -0,0 +1,117 @@
...
+const threshold = 20;
+const totalPrice: Price = { value: threshold + 1 };
+
+const mockQuote: Quote = {
```
Uli-Tiger:
```suggestion
const quote: Quote = {
```

feature-libs/quote/assets/translations/en/quote.i18n.ts
```diff
@@ -47,6 +47,13 @@ export const quote = {
       creationSuccess: 'Quote #{{ code }} created successfully',
       cart: 'Cart',
     },
+    seller: {
+      apply: 'Apply',
+      placeholder: '0',
+      discount: 'Absolute Discount',
+      expiryDate: 'Expiry Date',
+      validation: 'Enter a valid absolute discount',
```
Larisa-Staroverova: Please, rename, for instance, to **discountValidationText** or ???

feature-libs/quote/components/details/cart/summary/quote-details-cart-summary.component.spec.ts
```diff
@@ -0,0 +1,117 @@
...
+    fixture = TestBed.createComponent(QuoteDetailsCartSummaryComponent);
+    component = fixture.componentInstance;
+    facade = TestBed.inject(QuoteFacade);
+    mockQuoteDetails$.next(mockQuote);
+  });
+
+  it('should create component', () => {
+    expect(component).toBeDefined();
+    expect(facade).toBeDefined();
+  });
```
Larisa-Staroverova, line 116: Please, add component tests that test the structure of the component.

feature-libs/quote/components/quote-seller-edit/quote-seller-edit.component.service.ts
```diff
@@ -0,0 +1,268 @@
...
+    return this.getLocalizationElements().pipe(
+      map((localizationElements) => {
+        input = input.replace(localizationElements.currencySymbol, '');
+        return this.parseInput(input, localizationElements.locale);
+      })
+    );
+  }
+
+  /**
+   * Gets formatter according to current locale and currency
```
Larisa-Staroverova: I would suggest to use **Retrieves** instead of **Gets**

feature-libs/quote/components/quote-seller-edit/quote-seller-edit.component.service.ts
```diff
@@ -0,0 +1,268 @@
...
+   * @returns Observable of formatters
+   */
+  getFormatter(): Observable<Intl.NumberFormat> {
+    return this.getLocalizationElements().pipe(
+      map((localizationElements) => localizationElements.formatter)
+    );
+  }
+  /**
+   * Gets localization elements according to current locale and currency
```
Larisa-Staroverova: I would suggest to use **Retrieves** instead of **Gets**

feature-libs/quote/components/quote-seller-edit/quote-seller-edit.component.service.ts
```diff
@@ -0,0 +1,268 @@
...
+          locale,
+          formatter,
+          symbol?.value
+        );
+      })
+    );
+  }
+
+  /**
+   * Gets number format validator according to inputs
```
Larisa-Staroverova: I would suggest to use **Retrieves** instead of **Gets**

feature-libs/quote/components/quote-seller-edit/quote-seller-edit.component.service.ts
```diff
@@ -0,0 +1,268 @@
...
+   * Removes time portion from timestamp
+   * @param timestamp Timestamp as string
+   * @returns Date portion of timestamp
+   */
+  removeTimeFromDate(timestamp?: string): string | undefined {
+    return timestamp?.toString().substring(0, 10);
+  }
+
+  /**
+   * Check if quote state belongs to seller and can be edited
```
Larisa-Staroverova: I would suggest to use **Verifies** instead of **Check**

feature-libs/quote/components/quote-seller-edit/quote-seller-edit.component.service.ts
```diff
@@ -0,0 +1,268 @@
...
+   * @param quoteState
+   * @returns Is it for seller?
+   */
+  isEditableForSeller(quote: Quote): boolean {
```
Larisa-Staroverova: Please, rename to **isEditable**. 
I think it is clear from the service name that it will be verified  if it is editable for a seller

feature-libs/quote/components/quote-seller-edit/quote-seller-edit.component.service.ts
```diff
@@ -0,0 +1,268 @@
...
+    const quoteState = quote.state;
+    return (
+      (quoteState === QuoteState.SELLER_REQUEST ||
+        quoteState === QuoteState.SELLER_DRAFT) &&
+      quote.isEditable
+    );
+  }
+
+  /**
+   * Get maximum number of decimal places. This supports validation, but it is not sufficient to do a complete validation,
```
Larisa-Staroverova: I would suggest to use **Retrieves** instead of **Gets**

feature-libs/quote/components/quote-seller-edit/quote-seller-edit.component.ts
```diff
@@ -0,0 +1,138 @@
...
+import { filter, take } from 'rxjs/operators';
+import { QuoteSellerEditComponentService } from './quote-seller-edit.component.service';
+import { ICON_TYPE } from '@spartacus/storefront';
+
+@Component({
+  selector: 'cx-quote-seller-edit',
+  templateUrl: './quote-seller-edit.component.html',
+})
+export class QuoteSellerEditComponent implements OnInit, OnDestroy {
+  quoteDetailsForSeller$: Observable<Quote> = this.quoteFacade
```
Larisa-Staroverova, line 32: Please, rename to **quoteDetails$**
It is clear from the component name that these quote details are for the seller.

feature-libs/quote/components/quote-seller-edit/quote-seller-edit.component.ts
```diff
@@ -0,0 +1,138 @@
...
+        })
+    );
+  }
+
+  ngOnDestroy(): void {
+    this.subscription.unsubscribe();
+  }
+  /**
+   * Do we display validation message?
+   * @returns True in case discount control has errors
```
Larisa-Staroverova, line 92: How about
> Should the validation message be displayed / shown?
More generic

feature-libs/quote/components/quote-seller-edit/quote-seller-edit.component.ts
```diff
@@ -0,0 +1,138 @@
...
+   */
+  mustDisplayValidationMessage(): boolean {
+    return !this.form.controls.discount.valid;
+  }
+
+  /**
+   * On applying discount, we send its value to the facade layer
+   * @param quoteCode Quote code
```
Larisa-Staroverova, line 100: How about
> On applying discount, the value will be sent to the facade layer

feature-libs/quote/components/quote-seller-edit/quote-seller-edit.component.ts
```diff
@@ -0,0 +1,138 @@
...
+          };
+          this.quoteFacade.addDiscount(quoteCode, discount);
+          this.form.controls.discount.setValue(formatter.format(parsedValue));
+        });
+    }
+  }
+
+  /**
+   * On setting a date, we send its value to the facade layer
+   * @param quoteCode Quote code
```
Larisa-Staroverova, line 124: How about
>On setting a date, the value will be sent to the facade layer

Uli-Tiger (commented):
instead of the complex determination of the total number of places, which took me some time to understand, wouldn't it be possible to simply check that `totalValue - discountValue >= 0`. 
So that we allow any number of places, but have an additional custom validation that the discount Value is not greater than the undiscounted value?

feature-libs/quote/components/quote-seller-edit/quote-seller-edit.component.html
```diff
@@ -0,0 +1,49 @@
...
+          <input
+            required="true"
+            type="text"
+            class="form-control"
+            name="discount"
+            placeholder="{{ 'quote.seller.placeholder' | cxTranslate }}"
+            formControlName="discount"
+          />
+          <button
+            class="btn btn-block btn-secondary cx-apply"
```
Larisa-Staroverova: It is not necessary to add additional class as **cx-apply**
You could style to referring to the existing class like
`.btn-secondary {
    max-width: 6.5rem;
    height: 48px;
  }`

feature-libs/quote/components/quote-seller-edit/quote-seller-edit.component.ts
```diff
@@ -0,0 +1,137 @@
...
+    validityDate: new UntypedFormControl(),
+  });
+
+  iconType = ICON_TYPE;
+
+  protected subscription: Subscription = new Subscription();
+
+  constructor(
+    protected quoteFacade: QuoteFacade,
+    protected quoteSellerEditComponentService: QuoteSellerEditComponentService
```
Larisa-Staroverova, line 51: How about:
> protected quoteSellerEditService: QuoteSellerEditComponentService

feature-libs/quote/components/quote-seller-edit/quote-seller-edit.component.service.spec.ts
```diff
@@ -0,0 +1,241 @@
...
+        )
+      ).toBe(true);
+    });
+  });
+
+  describe('getMaximumNumberOfTotalPlaces', () => {
+    it('should compile number of total places from total if no quote discount is present, taking 2 decimal points into account ', () => {
+      expect(service.getMaximumNumberOfTotalPlaces(quote)).toBe(
+        Math.log10(TOTAL_PRICE) + 3
+      );
```
Uli-Tiger: Does calculating the expected value not just (partly) repeat the code under test? why not expect a  just '6'
```suggestion
      expect(service.getMaximumNumberOfTotalPlaces(quote)).toBe(6); //1000.00 => 6
```

feature-libs/quote/components/quote-seller-edit/quote-seller-edit.component.service.spec.ts
```diff
@@ -0,0 +1,241 @@
...
+    });
+
+    it('should compile number of total places from absolute discount if that exceeds total ', () => {
+      quote.quoteDiscounts = { value: DISCOUNT_RATE };
+      expect(service.getMaximumNumberOfTotalPlaces(quote)).toBe(
+        Math.log10(DISCOUNT_RATE) + 3
+      );
```
Uli-Tiger:
```suggestion
      expect(service.getMaximumNumberOfTotalPlaces(quote)).toBe(7)); //10000.00 => 7
```

Larisa-Staroverova (changes requested):

feature-libs/quote/components/quote-seller-edit/quote-seller-edit.component.service.spec.ts
```diff
@@ -0,0 +1,241 @@
...
+
+    it('should round properly according to log10 for different maximum totals', () => {
+      const quote999: Quote = { ...quote, totalPrice: { value: 999 } };
+      const quote100: Quote = { ...quote, totalPrice: { value: 100 } };
+      expect(service.getMaximumNumberOfTotalPlaces(quote999)).toBe(
+        service.getMaximumNumberOfTotalPlaces(quote100)
+      );
+      expect(service.getMaximumNumberOfTotalPlaces(quote)).toBe(
+        service.getMaximumNumberOfTotalPlaces(quote999) + 1
+      );
```
Uli-Tiger: those both test seem to assume that Math.log10 is used to count the number of places which is an implementation detail.  in the end they just check
2000 => 6
3000 => 6
999 => 5
100 => 5

See also my comment regarding the usage Math.floor for always rounding down. I believe these tests are obsolete if use Math.floor.

feature-libs/quote/components/quote-seller-edit/quote-seller-edit.component.service.ts
```diff
@@ -0,0 +1,268 @@
...
+   * @param quote Quote
+   * @returns Maximum number of places, including 2 decimal places
+   */
+  getMaximumNumberOfTotalPlaces(quote: Quote): number {
+    const numberOfDecimalPlaces = 2;
+    const maximum = Math.max(
+      quote.totalPrice.value ?? 1,
+      quote.quoteDiscounts?.value ?? 1
+    );
+    return Math.round(Math.log10(maximum) - 0.5) + 1 + numberOfDecimalPlaces;
```
Uli-Tiger: if you want to always round down, you may use Math.floor
```suggestion
    return Math.floor(Math.log10(maximum)) + 1 + numberOfDecimalPlaces;
```

feature-libs/quote/assets/translations/en/quote.i18n.ts
ChristophHi: Agree to discountValidationText

feature-libs/quote/components/quote-seller-edit/quote-seller-edit.component.service.ts
ChristophHi: Agree. Makes sense in the context of this component/service

feature-libs/quote/components/details/cart/summary/quote-details-cart-summary.component.spec.ts
ChristophHi: Agree to this naming suggestion and the previous one

feature-libs/quote/components/details/cart/summary/quote-details-cart-summary.component.spec.ts
ChristophHi: As I said this is only a skeleton to get the test coverage green. No actual component and component test, that's not part of this PR

feature-libs/quote/components/quote-seller-edit/quote-seller-edit.component.service.spec.ts
ChristophHi: Don't fully agree. Here I am just stating what I expect, having 6 would be ok but I would lose the reference to 'TOTAL_PRICE' (comment needed, but then if one changes the constant value one might not update the comment)

feature-libs/quote/components/quote-seller-edit/quote-seller-edit.component.service.spec.ts
ChristophHi: See above comment

feature-libs/quote/components/quote-seller-edit/quote-seller-edit.component.ts
ChristophHi: Don't fully agree. We often have quoteDetails$ with the exact same meaning in different components. So here I wanted to make clear that it's a different thing (it does not necessarily emit!)

feature-libs/quote/components/quote-seller-edit/quote-seller-edit.component.ts
ChristophHi: ok, agree, probably better

feature-libs/quote/components/quote-seller-edit/quote-seller-edit.component.ts
ChristophHi: no, I would rather stick to the convention to name the members like the underlying class

feature-libs/quote/components/quote-seller-edit/quote-seller-edit.component.service.ts
ChristophHi: Agree, good point

feature-libs/quote/components/quote-seller-edit/quote-seller-edit.component.service.spec.ts
ChristophHi: Partially agree: I wasn't aware of Math.floor and wanted to test my rounding => too many tests if one believes that Math.floor does what it's supposed to do.
But let's still keep the second test, otherwise the rounding is not covered at all

ChristophHi: Renamed the test to clarify what it's supposed to do (not referring to the implementation details)

ChristophHi:
@Uli-Tiger 

> instead of the complex determination of the total number of places, which took me some time to understand, wouldn't it be possible to simply check that totalValue - discountValue >= 0.
> So that we allow any number of places, but have an additional custom validation that the discount Value is not greater than the undiscounted value?

There are 2 issues here: (1) We don't know the undiscounted value (2) the validator has not that much context. It cannot easily check on the quote state but focuses on one control.

feature-libs/quote/components/quote-seller-edit/quote-seller-edit.component.service.spec.ts
Uli-Tiger: @ChristophHi: than maybe 
```suggestion
      expect(service.getMaximumNumberOfTotalPlaces(quote)).toBe(TOTAL_PRICE.toFixed(0).length + 2);
```
the `+ 2 ` also corresponds to the test description of `..., taking 2 decimal points into account`.

Larisa-Staroverova (approved):

feature-libs/quote/components/quote-seller-edit/quote-seller-edit.component.service.spec.ts
ChristophHi: Yes, agree

ChristophHi: Will also change in the other occurence

Uli-Tiger (approved):

PR #17738, opened 2023-08-08
fix: when asm login, it changed the logged in user to anonymous

scarai-sap:
See https://jira.tools.sap/browse/CXSPA-3974

- use userAccountFacade to get customerId as UserProfileFacade is not work.
- Call get customer id before get new token(agent token


feature-libs/asm/root/services/csagent-auth.service.spec.ts
```diff
@@ -4,12 +4,12 @@ import {
   AuthActions,
   AuthService,
   AuthToken,
+  FeatureConfigService,
   OAuthLibWrapperService,
   OCC_USER_ID_ANONYMOUS,
   OCC_USER_ID_CURRENT,
   UserIdService,
 } from '@spartacus/core';
-import { UserProfileFacade } from '@spartacus/user/profile/root';
```
scarai-sap, line 12: UserProfileFacade doesn't work

feature-libs/asm/root/services/csagent-auth.service.ts
```diff
@@ -48,6 +75,15 @@ export class CsAgentAuthService {
     password: string
   ): Promise<void> {
     let userToken: AuthToken | undefined;
+    // Start emulation for currently logged in user
```
scarai-sap, line 79: use userAccountFacade to get customerId as UserProfileFacade is not work.

Call get customer id before get new token(agent token

scarai-sap (commented):
add description

i53577 (approved):
LGTM

PR #17739, opened 2023-08-08
Verify endpoint failed unit tests

rmch91:
@FollowTheFlo Hello Florent. I did one change in OccOpfPaymentAdapter. Add` normalizeHttpError` - ` catchError((error) => throwError(normalizeHttpError(error))) `. This was only one place where we not use this, there was only `throwError(error)`. I'm not sure that backoff for jalo error worked properly. But i'm not sure if it was intended that `normalizeHttpError` was not used in this case or not. Could you please make a review for me?


kimhw0630 (approved):

kimhw0630:
coverage looks good 
image
Some of the code is not covered in occ-opf.adapter.ts, but it is not within the scope of this ticket.

PR #17744, opened 2023-08-08
feat: add hostedFields pattern#2

FollowTheFlo:
CXSPA-3879
1- from Verification Page (the existing one), add new logic to extract 'afterRedirectScriptFlag' param from URL.
2- call after-redirections-scripts GET request
3- load dynamic script from its response
4- create global function 'submitCompleteRedirect' and 'getRedirectParams' that will be called by dynamic script from step3

image


integration-libs/opf/base/root/services/opf-resource-loader.service.ts
FollowTheFlo, line 1: This service is a duplicate of 
spartacus/integration-libs/opf/checkout/core/services/opf-resource-loader.service.ts
It will be handled later in [CXSPA-4714](https://jira.tools.sap/browse/CXSPA-4714). I attempted to put this service in a facade but hit some prob.

integration-libs/opf/base/core/facade/opf-global-functions.service.ts
```diff
@@ -197,4 +221,58 @@ export class OpfGlobalFunctionsService implements OpfGlobalFunctionsFacade {
       });
     };
   }
+
+  protected registerSubmitCompleteRedirect(
```
rmch91, line 239: Could you consider splitting this method for a few smaller methods? It seems to be too long, complex and hard to read.

integration-libs/opf/base/root/components/opf-payment-verification/opf-payment-verification.component.ts
```diff
@@ -30,19 +34,45 @@ export class OpfPaymentVerificationComponent implements OnInit, OnDestroy {
...
-            responseMap
-          );
-        }),
-        concatMap(() => {
-          return this.paymentService.placeOrder();
-        })
+            paramsMap: paramsMap,
+            afterRedirectScriptFlag,
+          }) => {
+            if (afterRedirectScriptFlag === 'true') {
```
rmch91: I think it also will be good to move to separate method, which will return proper method from payment service based on condition

FollowTheFlo: thx!divided in more comprehensive methods:
runPaymentPattern
  runHostedFieldsPattern
  runHostedPagePattern

integration-libs/opf/base/core/facade/opf-global-functions.service.ts
FollowTheFlo: done, created runSubmitComplete method that get called twice.

integration-libs/opf/base/occ/adapters/occ-opf.adapter.ts
```diff
@@ -43,17 +45,25 @@ export class OccOpfPaymentAdapter implements OpfPaymentAdapter {
     protected config: OpfConfig
   ) {}
 
-  header: { [name: string]: string } = {
```
FollowTheFlo, line 46: The header situation is part of CXSPA-4688 that will be merged soon. You can ignore. You won't be able to test it without it.

rmch91 (approved):

PR #17746, opened 2023-08-09
Enable text truncation for the card paragraph

Larisa-Staroverova:


Uli-Tiger:
i would suggest to align the more links with the edit pencil, like below
image
for this i added this css rule to the button
```
{
    margin-inline-end: -25px;
    float: right;
    margin-block-start: 5px;
    font-size: smaller;
}
```

@Uli-Tiger, I will check your suggestion with UI designer and come back with her feedback.
By the way your suggestion is nice as long as there are characters before the link:
image
It is common to locate `more` link directly after the truncated text and not at the end of the link that's I would not recommend to change the position of the link by css. 
It is also possible to calculate the distance to the end of the line and fill it with the characters but do we really want it!?!

feature-libs/quote/components/details/overview/quote-details-overview.component.html
```diff
@@ -17,6 +17,8 @@ <h3 class="cx-status">
             getQuoteInformation(quoteDetails.name, quoteDetails.description)
               | async
           "
+          [truncateText]="true"
+          [charactersLimit]="30"
```
Uli-Tiger: can we make those configurable?
per default we maybe want to use 60 chars (2 lines on smartphone, 3 lines in desktop, assuming we use the styling with the more link right aligned) 30 is really short.

Uli-Tiger: or even better would be to remove the fixed character limit and truncate after one or two lines, so the styling adapts to the device. see also https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-line-clamp

Uli-Tiger (changes requested):

feature-libs/quote/components/details/overview/quote-details-overview.component.html
Larisa-Staroverova: @Uli-Tiger, your suggestion make sense on the mobile widgets but not on the desktop. Let me share the screenshot:
image
As you could see in the example with 60 characters there is too much empty space in the last 2 card tiles.
The UI designers said us to avoid this.

Larisa-Staroverova: It would be possible to configure the number of characters for the different devices, but I don't see any benefit in doing so.
Then the question remains why truncation is used when all the text could be displayed, but all the text explodes the size of the card tile.

feature-libs/quote/components/config/quote-ui.config.ts
```diff
@@ -14,6 +14,7 @@ import {
 
 export interface QuoteUIConfigFragment {
   maxCharsForComments?: number;
+  maxCharsForCardTile?: number;
```
Uli-Tiger:
```suggestion
  truncateCardTileContentAfterNumChars?: number;
```
maxCharsForComments is really a limit, here we just configure the truncation, so we should name it differently.

Uli-Tiger (changes requested):

feature-libs/quote/components/config/quote-ui.config.ts
Larisa-Staroverova: Okay! I see what you mean. I will adjust right now

PR #17760, opened 2023-08-15
Fix styling issues (UX feedback)

Larisa-Staroverova:
See [CXSPA-4345](https://jira.tools.sap/browse/CXSPA-4345)

### DIFF
--- a/feature-libs/quote/components/details/cart/quote-details-cart.component.html
+++ b/feature-libs/quote/components/details/cart/quote-details-cart.component.html
@@ -1,13 +1,11 @@
 <ng-container *ngIf="{ showCart: showCart$ | async } as state">
-  <div class="cart-toggle" (click)="onToggleShowOrHideCart(!!state.showCart)">
+  <div class="cx-toggle" (click)="onToggleShowOrHideCart(!!state.showCart)">
     <cx-icon
       aria-hidden="false"
       [type]="state.showCart ? iconTypes.CARET_UP : iconTypes.CARET_DOWN"
     >
     </cx-icon>
-    <span class="cart-toggle-text">{{
-      'quote.commons.cart' | cxTranslate
-    }}</span>
+    <span class="cx-text">{{ 'quote.commons.cart' | cxTranslate }}</span>
   </div>
 
   <ng-container
--- a/feature-libs/quote/components/details/comment/quote-details-comment.component.html
+++ b/feature-libs/quote/components/details/comment/quote-details-comment.component.html
@@ -1,14 +1,12 @@
 <ng-container *ngIf="quoteDetails$ | async as quoteDetails">
   <ng-container *ngIf="showComments(quoteDetails)">
-    <div class="cx-comment-toggle" (click)="expandComments = !expandComments">
+    <div class="cx-toggle" (click)="expandComments = !expandComments">
       <cx-icon
         aria-hidden="false"
         [type]="expandComments ? iconTypes.CARET_UP : iconTypes.CARET_DOWN"
       >
       </cx-icon>
-      <span class="cx-comment-text">{{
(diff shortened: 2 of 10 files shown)


feature-libs/quote/components/details/cart/quote-details-cart.component.html
```diff
@@ -1,13 +1,11 @@
...
+  <div class="cx-toggle" (click)="onToggleShowOrHideCart(!!state.showCart)">
     <cx-icon
       aria-hidden="false"
       [type]="state.showCart ? iconTypes.CARET_UP : iconTypes.CARET_DOWN"
     >
     </cx-icon>
-    <span class="cart-toggle-text">{{
-      'quote.commons.cart' | cxTranslate
-    }}</span>
+    <span class="cx-text">{{ 'quote.commons.cart' | cxTranslate }}</span>
```
Uli-Tiger, line 8:
```suggestion
    <span class="cx-toggle-text">{{ 'quote.commons.cart' | cxTranslate }}</span>
```
i think cx-text is to common, what about cx-toggle-text?

feature-libs/quote/components/details/comment/quote-details-comment.component.html
```diff
@@ -1,14 +1,12 @@
...
+    <div class="cx-toggle" (click)="expandComments = !expandComments">
       <cx-icon
         aria-hidden="false"
         [type]="expandComments ? iconTypes.CARET_UP : iconTypes.CARET_DOWN"
       >
       </cx-icon>
-      <span class="cx-comment-text">{{
-        'quote.comments.title' | cxTranslate
-      }}</span>
+      <span class="cx-text">{{ 'quote.comments.title' | cxTranslate }}</span>
```
Uli-Tiger, line 9:
```suggestion
      <span class="cx-toggle-text">{{ 'quote.comments.title' | cxTranslate }}</span>
```
i think cx-text is to common, what about cx-toggle-text?

Uli-Tiger (changes requested):

projects/storefrontstyles/scss/components/misc/chat-messaging/_messaging.scss
```diff
@@ -119,6 +119,7 @@
         .cx-message-input {
           display: flex;
           gap: 1rem;
+          padding-block-start: 0.5rem;
```
Uli-Tiger, line 122: This would not only affect the usage of the messaging component within quotes, but everywhere. Is this desired?

Let's add the additional space as `margin-block-end` to `cx-message-item-selection`, which is the product DDLB that was newly added by us. So it does not influence existing usage without product DDLB.

feature-libs/quote/components/details/comment/quote-details-comment.component.html
Larisa-Staroverova: I don't agree because `cx-text` is framed by `div` with `cx-toggle` class. 
Please, see  `_quote-toggle.scss` mixin.
The main SPA guideline is to keep it simple and consider within one component.
It is unnecessary to concatenate the classes. In this case the class name will be unnecessary long but we don't win anything through this concatenation only complexity.

projects/storefrontstyles/scss/components/misc/chat-messaging/_messaging.scss
Larisa-Staroverova: I am not sure whether I understand what you mean.
122 line simple adds some padding between the message input field and `Add New Message` and `All Products` drop-down list. That is all.
It is not required to add some margin between the message input field and the characters left info text.
Please, see the task description: [CXSPA-4345](https://jira.tools.sap/browse/CXSPA-4345)

PR #17765, opened 2023-08-17
fix: agent starts customer emulation after creating new customer, a CONFIRMATION message popup

i53577:
ticket: https://jira.tools.sap/browse/CXSPA-4089

### DIFF
--- a/feature-libs/asm/components/asm-main-ui/asm-main-ui.component.ts
+++ b/feature-libs/asm/components/asm-main-ui/asm-main-ui.component.ts
@@ -302,6 +302,7 @@ export class AsmMainUiComponent implements OnInit, OnDestroy {
     { customerId }: { customerId?: string },
     parameters?: AsmDeepLinkParameters
   ): void {
+    this.showCreateCustomerSuccessfullyAlert = false;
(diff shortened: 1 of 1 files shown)


feature-libs/asm/components/asm-main-ui/asm-main-ui.component.ts
```diff
@@ -302,6 +302,7 @@ export class AsmMainUiComponent implements OnInit, OnDestroy {
     { customerId }: { customerId?: string },
     parameters?: AsmDeepLinkParameters
   ): void {
+    this.showCreateCustomerSuccessfullyAlert = false;
```
i53577: avoid having `CreateCustomerSuccessfullyAlert` not closed for the previous created customer

scarai-sap: maybe move to line 309 is better choice as other set/reset flag are in there(just avoid corner case the selected user  has no customer id. like  women@hybris.com)?

niehuayang (approved):
LGTM

scarai-sap (approved):

PR #17766, opened 2023-08-17
fix: CXINT-2321 and CXINT-2322 (#17766)

SherwinVarghese:
- Fixes https://jira.tools.sap/browse/CXINT-2321 and https://jira.tools.sap/browse/CXINT-2322
- Added date validation logic to handle textual inputs
- Only valid future dates are sent to the OCC APIs
- Fixed issue with clear button defaulting to the previous date 

### E2ES pasing
Screenshot 2023-08-17 at 5 19 27 PM

### Unit tests passing

```
=============================== Coverage summary ===============================
Statements   : 92.7% ( 89/96 )
Branches     : 85.71% ( 36/42 )
Functions    : 89.47% ( 34/38 )
Lines        : 92.47% ( 86/93 )
================================================================================
```


feature-libs/requested-delivery-date/root/components/shared/date-validation.service.ts
```diff
@@ -21,8 +21,37 @@ export class DateValidationService {
...
   }
+
+  /**
+   * Compares 2 Date strings in the format 'dd-mm-yyy'
+   * @param date1
+   * @param date2
+   * @returns -1 if date2 is greater, 0 if both the dates are equal, 1 if date1 is greater, -2 for invalid inputs
+   */
+  compareDateStrings(date1: string, date2: string): number {
+    if (date1.length === 0 || date2.length === 0) {
```
Matejk00: I think this function can be simplified, as right now there is a lot of if's.

```
compareDateStrings(date1: string, date2: string): number {
  if (!date1 || !date2) {
    return -2;
  }
  
  const d1 = this.getDateFromDateString(date1);
  const d2 = this.getDateFromDateString(date2);
  
  return d1 === d2 ? 0 : d1 < d2 ? -1 : 1;
}
```

Also could you let me know, why as a result for any specific case we need to return a number (like -2  for example) instead of some `enum` value with the status?

feature-libs/requested-delivery-date/root/components/delivery-mode-date-picker/delivery-mode-date-picker.component.ts
```diff
@@ -48,6 +48,7 @@ export class DeliveryModeDatePickerComponent implements OnInit, OnDestroy {
     requestDeliveryDate: new FormControl(),
   });
   protected isDatePickerReadOnly: boolean = true;
+  protected readonly KEY_DEBOUNCE_TIME = 500;
```
Matejk00: Are we using this property somewhere?

Matejk00 (changes requested):

feature-libs/requested-delivery-date/root/components/delivery-mode-date-picker/delivery-mode-date-picker.component.ts
SherwinVarghese: It was added while testing the debounce changes, will remove it.
thanks

feature-libs/requested-delivery-date/root/components/shared/date-validation.service.ts
SherwinVarghese: returning a number would make it easier to compare by checking `compareDateStrings(date1, date2) < 0`

Matejk00: Still I am not sure if I get a concept here. You are calling this `compareDateStrings` once in a component's statement:

```
 this.dateValidationService.compareDateStrings(
        requestedDate,
        this.earliestRetrievalAt || ''
      ) < 0
```

by `< 0` you expect date2 to be greater or -2 for invalid inputs. I guess we could change a logic of your function and return `true` or `false` depending on an arguments passed to the function. There is not need to do this check in a component in my opinion.

Matejk00 (changes requested):
https://github.com/SAP/spartacus/pull/17766#discussion_r1297180546

feature-libs/requested-delivery-date/root/components/shared/date-validation.service.ts
SherwinVarghese: The logic in the lines https://github.com/SAP/spartacus/blob/4f88f530773a5c4f5dffa17405487774767ac573/feature-libs/requested-delivery-date/root/components/delivery-mode-date-picker/delivery-mode-date-picker.component.ts#L114-L117 
check for invalid date inputs, so the `compareDateStrings` method will not be passed an invalid input.

That being said, we return `-2` as a safety in case this method gets reused in the future.

We can definitely write a wrapper method that returns a `boolean` like - 

```
isDateGreaterOrEqual(source: string, target: string): boolean {
    return this.compareDateStrings(source, target) >= 0;
  }
```

What do you suggest @Matejk00 ?

Matejk00: If we are using `compareDateStrings` only in one place and it's not used anywhere else, we could transform this method to be `isDateGreaterOrEqual`. No need to wrap stuff in my opinion, just change a name and a logic of the existing method.

What do you think @SherwinVarghese?

SherwinVarghese: okay @Matejk00, sounds good.

Matejk00 (approved):

PR #17788, opened 2023-08-23
feat: CXSPA-4370 PoC with alternative way of providing Trace Context (and other transformations in the future)

pawelfras:
This PR contains a PoC of alternative ways to extend loggers. At the moment, we want to extend logs with Trace Context values, but for possible future extensions, it would be good to provide a mechanism to add new transformers for each extension and combine them instead of modifying the default logger for this purpose.


core-libs/setup/ssr/logger/transformers/default-transformers.ts
```diff
@@ -0,0 +1,13 @@
...
+ * SPDX-FileCopyrightText: 2023 SAP Spartacus team <spartacus-team@sap.com>
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { DefaultTraceparentTransformer } from './traceparent/default-traceparent-transformer';
+
+export interface TransformerConfig {}
+
+export const defaultTransformersConfig: TransformerConfig = {
```
Platonn, line 11: word `transformer` is too general in the context of public api of `@spartacus/setup/ssr`.
What about `LogTransformer`? 

LogTransformerConfig
defaultLogTransformerConfig
TraceparentLogTransformer
DefaultTraceparentLogTransformer
etc.

core-libs/setup/ssr/logger/transformers/express-log-transformer.ts
```diff
@@ -0,0 +1,15 @@
...
+ */
+
+import { ExpressServerLogger, ExpressServerLoggerContext } from '../loggers';
+
+export interface ExpressLogTransformer {
+  transform(
+    message: string,
+    context: ExpressServerLoggerContext,
+    logger?: ExpressServerLogger
+  ): [string, ExpressServerLoggerContext];
```
Platonn, line 14: instead of tuple return type, what about an object with semantic properties:
{ message: string, context: ExpressServerLoggerContext }

This would help mentally interpreting the result of this function and it would be self-explanatory

core-libs/setup/ssr/logger/transformers/utils/transformers-utils.ts
```diff
@@ -0,0 +1,43 @@
...
+): ExpressLogTransformer[] {
+  return Object.keys(transformersConfig)
+    .map((key) => transformersConfig[key as keyof TransformerConfig])
+    .filter(isTransformer);
+}
+
+export function isTransformer(
+  transformer: ExpressLogTransformer | undefined
+): transformer is ExpressLogTransformer {
+  return transformer !== undefined;
```
Platonn, line 42: Didn't you mean that `transformer is ExpressLogTransformer` only when:
`return typeof transformer?.transform === 'function'` ?

core-libs/setup/ssr/logger/transformers/utils/transformers-utils.ts
```diff
@@ -0,0 +1,43 @@
...
+  };
+  if (transformers) {
+    transformers.remove?.forEach((transformerToBeRemoved) => {
+      delete parsedTransformers[transformerToBeRemoved];
+    });
+  }
+  return parseTransformers(parsedTransformers);
+}
+
+export function parseTransformers(
```
Platonn, line 31: nitpick: we don't "parse" those transformers, I believe 🤔 naming issue IMHO, but don't understand the author's intention, so hard to propose alternative

core-libs/setup/ssr/logger/transformers/utils/transformers-utils.ts
```diff
@@ -0,0 +1,43 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { TransformerConfig } from '../default-transformers';
+import { ExpressLogTransformer } from '../express-log-transformer';
+
+export interface ExpressLogTransformerConfig {
+  replace?: TransformerConfig;
+  remove?: [keyof TransformerConfig];
```
Platonn, line 12: Instead of introducing a custom convention with a configurable `replace/remove` did you consider one of those 2 alternatives:
a) transformers being simply an array
b) transformers being an key-object map. we could use `deepMerge() from @spartacus/core` to reset/overwrite default object's keys.

core-libs/setup/ssr/optimized-engine/optimized-ssr-engine.ts
```diff
@@ -525,20 +509,26 @@ export class OptimizedSsrEngine {
...
+        ssrOptions?.logger === true
+          ? new DefaultExpressServerLogger()
+          : ssrOptions.logger;
+      logger.setTransformers(logTransformers);
+      return logger;
     }
-    return ssrOptions?.logger || new LegacyExpressServerLogger();
+    return new LegacyExpressServerLogger();
   }
 }
```
Platonn, line 534: > `//CXSPA-3680 - remove this method in 7.0`

Now this method seems to be needed. So we should move and change this code comment, to describe more precisely what should be done as part of CXSPA-3680, instead of removing the whole method.

core-libs/setup/ssr/optimized-engine/ssr-optimization-options.ts
```diff
@@ -124,6 +125,14 @@ export interface SsrOptimizationOptions {
...
    */
   logger?: true | ExpressServerLogger; //CXSPA-3680 - allow only providing ExpressServerLogger implementations
+
+  /**
+   * Config for transforming logs.
+   * It allows to add, remove or replace log transformers.
+   * The default transformers are: TraceparentTransformer
+   * @see ExpressLogTransformer
+   */
+  logTransformers?: ExpressLogTransformerConfig;
```
Platonn, line 135: I suggest to use `deepMerge() from @spartacus/core` for merging custom config into the default config. And the `logTransformers` should have type simply simply `TransformerConfig` (`LogTransformerConfig`)

core-libs/setup/ssr/logger/transformers/utils/transformers-utils.ts
pawelfras: Yeah, naming is the hardest part of the programming 😅 I didn't find a better name and didn't want to spend any more time on this PoC. We can go through this together and discuss it.

core-libs/setup/ssr/optimized-engine/optimized-ssr-engine.ts
```diff
@@ -492,30 +493,13 @@ export class OptimizedSsrEngine {
...
-      }
-    } catch (e) {
-      error =
-        e instanceof Error
-          ? e
-          : new Error('Unexpected error during parsing traceparent header');
-    }
-
-    return [requestContext, error];
+    return requestContext;
```
Platonn: Now this method doesn't return request's context, but only generates a random uuid and timestamp - which is request-agnositc. Unless we find a better name for this method, maybe we should move it's body to inside the `updateWithRequestContext()` method?

core-libs/setup/ssr/logger/transformers/utils/transformers-utils.ts
pawelfras: Good point! I had in mind that we can have only two states - `ExpressLogTransformer` and `undefined` and It seems I focused only on how to filter `undefined` values... Of course, your proposal is more robust 💪

Platonn:
Shall we close this PR?

PR #17799, opened 2023-08-29
fix: smart edit infinite loader fix

kpawelczak:
ticket: https://jira.tools.sap/browse/CXSPA-3584

This function - `processCmsPage() ` is crucial for setting up the things SmartEdit needs. To make sure the function doesn't run unintentionally its execution is placed in a lazy-loaded module. In Spartacus, we use guards like `ProtectedRoutesGuard` and `AuthGuard`, which effectively block the loading of this particular module. As a result, the `processCmsPage() `function remains dormant and doesn't execute.


feature-libs/smartedit/core/services/smart-edit.service.ts
```diff
@@ -32,8 +31,7 @@ export class SmartEditService {
     protected baseSiteService: BaseSiteService,
     protected zone: NgZone,
     protected winRef: WindowRef,
-    protected rendererFactory: RendererFactory2,
-    protected config: SmartEditConfig
```
pawelfras: `should`: Let's not remove  `SmartEditConfig`.

Reasons:
1. it's not related to the bugfix
2. Removing the constructor's property is a breaking change.

feature-libs/smartedit/root/services/smart-edit-init.service.ts
```diff
@@ -0,0 +1,29 @@
...
+    protected smartEditService: SmartEditService,
+    protected smartEditLauncherService: SmartEditLauncherService
+  ) {}
+
+  init(): void {
+    if (this.smartEditLauncherService.isLaunchedInSmartEdit()) {
+      this.smartEditService.processCmsPage();
+    }
+  }
+}
```
pawelfras:
```suggestion
export const smartEditServiceInitializerFactory = (smartEditService: SmartEditService, smartEditLauncherService: SmartEditLauncherService) => () => {
 if(smartEditLauncherService.isLaunchedInSmartEdit()) {
   smartEditService.processCmsPage();
 }
}
```

feature-libs/smartedit/root/smart-edit-root.module.ts
```diff
@@ -40,6 +42,12 @@ export function smartEditFactory(
         },
       },
     }),
+    {
+      provide: MODULE_INITIALIZER,
+      useFactory: SmartEditInitService.factory,
+      deps: [SmartEditInitService],
+      multi: true,
+    },
```
pawelfras:
```suggestion
    {
      provide: MODULE_INITIALIZER,
      useFactory: smartEditServiceInitializerFactory,
      deps: [SmartEditService, SmartEditLauncherService],
      multi: true,
    },
```

pawelfras (commented):
Apart from added comments, from the architectural point of view (and because of possible circular dependencies), I have doubts about whether the `core` and `root` elements should be used with each other inside the library. Let's discuss it with one of our architects.

feature-libs/smartedit/root/services/smart-edit-init.service.ts
kpawelczak: I am not fond of this change as it reduces code extensibility and also negatively impacts the testing process.

pawelfras: The question is whether such a code is going to be extended. I believe this factory function shouldn't be part of the public API as it may cause many issues if used in the wrong way. The other thing is that it's an uncommon way of creating a factory and I don't know whether I like how the service is used - `SmartEditInitService` has to be passed to `deps` in providers to use the `factory` function that is part of `SmartEditInitService` as well... 🤔

pawelfras: In addition, the convention that you proposed is new to the project, which may lead to ambiguity when someone will have to add similar logic in the future.

projects/storefrontlib/cms-structure/guards/cms-page-guard.service.ts
```diff
@@ -71,11 +69,6 @@ export class CmsPageGuardService {
     state: RouterStateSnapshot
   ): Observable<boolean | UrlTree> {
     return this.cmsService.getPageComponentTypes(pageContext).pipe(
-      map((componentTypes) =>
-        pageContext.id === SMART_EDIT_CONTEXT
-          ? [SMART_EDIT_DUMMY_COMPONENT_TYPE, ...componentTypes]
-          : componentTypes
-      ),
```
Platonn, line 78: Leaving a note just for reference (no need to change the code):
Theoretically, removing the `const SMART_EDIT_DUMMY_COMPONENT_TYPE` from the public API and removing this behavior of loading `SmartEditModule` when the cms component type `SMART_EDIT_DUMMY_COMPONENT_TYPE` is present - those are breaking changes.

However, we need to observe that:
- the probability of customers using `SMART_EDIT_DUMMY_COMPONENT_TYPE` is very low
- the probability they dependend on the behavior that `SMART_EDIT_DUMMY_COMPONENT_TYPE` triggered the lazy-loading of `SmartEditModule` is very low
- by introducing `this.featureModules?.resolveFeature(SMART_EDIT_FEATURE).subscribe();` to the `SmartEditLauncherService` we preserve the original behavior - when the route is `/cx-preview`, the `SmartEditModule` is being lazy loaded. But now the logic is much simpler without unnecessary workarounds and depending on fake CMS component types and CmsGuard.

All that said, I agree to turn a blind eye to this slight breaking change.

Platonn (approved):

kpawelczak:
closing in favor of https://github.com/SAP/spartacus/pull/17877

PR #17801, opened 2023-08-29
feat: Default rendering strategy resolver

kpawelczak:
ticket: https://jira.tools.sap/browse/CXSPA-3492

### DIFF
--- a/core-libs/setup/ssr/optimized-engine/rendering-strategy-resolver-options.ts
+++ b/core-libs/setup/ssr/optimized-engine/rendering-strategy-resolver-options.ts
@@ -1,0 +1,15 @@
+/*
+ * SPDX-FileCopyrightText: 2023 SAP Spartacus team <spartacus-team@sap.com>
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+export interface RenderingStrategyResolverOptions {
+  excludedUrls?: string[];
+  excludedParams?: string[];
+}
+
+export const defaultRenderingBlocked: RenderingStrategyResolverOptions = {
+  excludedUrls: ['checkout', 'my-account'],
+  excludedParams: ['asm'],
+};
--- a/core-libs/setup/ssr/optimized-engine/rendering-strategy-resolver-options.ts
+++ b/core-libs/setup/ssr/optimized-engine/rendering-strategy-resolver-options.ts
@@ -3,13 +3,13 @@
  *
  * SPDX-License-Identifier: Apache-2.0
  */
-
 export interface RenderingStrategyResolverOptions {
   excludedUrls?: string[];
   excludedParams?: string[];
 }
 
-export const defaultRenderingBlocked: RenderingStrategyResolverOptions = {
-  excludedUrls: ['checkout', 'my-account'],
-  excludedParams: ['asm'],
-};
+export const defaultRenderingStrategyResolverOptions: RenderingStrategyResolverOptions =
+  {
+    excludedUrls: ['checkout', 'my-account'],
+    excludedParams: ['asm'],
+  };
--- a/core-libs/setup/ssr/optimized-engine/rendering-strategy-resolver-options.ts
+++ b/core-libs/setup/ssr/optimized-engine/rendering-strategy-resolver-options.ts
@@ -3,6 +3,7 @@
  *
  * SPDX-License-Identifier: Apache-2.0
  */
+
 export interface RenderingStrategyResolverOptions {
   excludedUrls?: string[];
   excludedParams?: string[];
--- a/core-libs/setup/ssr/optimized-engine/rendering-strategy-resolver.ts
+++ b/core-libs/setup/ssr/optimized-engine/rendering-strategy-resolver.ts
@@ -17,7 +17,7 @@ const hasExcludedParams = (request: Request): boolean => {
 };
 
 const hasExcludedUrl = (request: Request): boolean => {
-  const excludedUrls = ['/checkout'];
+  const excludedUrls = ['/checkout', '/my-account'];
 
   return request.url
     ? excludedUrls.some((url: string) => request.url.search(url) > -1)
--- a/core-libs/setup/ssr/optimized-engine/rendering-strategy-resolver.ts
+++ b/core-libs/setup/ssr/optimized-engine/rendering-strategy-resolver.ts
@@ -1,25 +1,31 @@
-import { Request } from "express";
-import { RenderingStrategy } from "./ssr-optimization-options";
+import { Request } from 'express';
+import { RenderingStrategy } from './ssr-optimization-options';
 
-const hasBlacklistedParams = (request: Request): boolean => {
-  const blacklistedParams = ['gclid', 'asm'];
-  const params = Object.getOwnPropertyNames(request.query);
+const hasExcludedParams = (request: Request): boolean => {
+  const excludedParams = ['gclid', 'asm'];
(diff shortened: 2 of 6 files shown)


core-libs/setup/ssr/optimized-engine/ssr-optimization-options.ts
```diff
@@ -132,13 +133,26 @@ export enum RenderingStrategy {
   ALWAYS_SSR = 1,
 }
 
+export interface RenderingBlocked {
+  urls?: string[];
+  params?: string[];
+}
```
pawelfras: `query:` What would you say about changing the interface name to `RenderingStrategyResolverOptions`?
```suggestion
export interface RenderingStrategyResolverOptions {
  excludedUrls?: string[];
  excludedParams?: string[];
}
```

Such names seem to be self-explanatory and in addition, the proposed interface name is a noun which I believe is a good practice.

Btw. I think this interface should be moved to a different place... Doesn't it cause circular dependency?

core-libs/setup/ssr/optimized-engine/rendering-strategy-resolver.ts
```diff
@@ -0,0 +1,53 @@
...
+  renderingBlocked: RenderingBlocked
+) => {
+  return request.url && renderingBlocked?.urls
+    ? renderingBlocked.urls.some((url) => request.url.search(url) > -1)
+    : false;
+};
+
+const shouldFallbackToCsr = (
+  request: Request,
+  renderingBlocked: RenderingBlocked
```
pawelfras: `could:` consider destructuring the object in this place and pass to `hasExcludedUrl` and `hasExcludedParams` only related values instead of the whole object.

core-libs/setup/ssr/optimized-engine/rendering-strategy-resolver.ts
```diff
@@ -0,0 +1,53 @@
...
+) => {
+  const params = request.query ? Object.getOwnPropertyNames(request.query) : [];
+
+  if (!renderingBlocked?.params) {
+    return false;
+  }
+
+  return renderingBlocked.params.some((excludedParam) =>
+    params.some((param) => excludedParam === param)
```
pawelfras: `could:` consider to use e.g. `indexOf` to to break the loop ASAP instead of going through the whole list each time.

core-libs/setup/ssr/optimized-engine/ssr-optimization-options.ts
```diff
@@ -132,13 +133,26 @@ export enum RenderingStrategy {
...
+};
+
 export const defaultSsrOptimizationOptions: SsrOptimizationOptions = {
   concurrency: 10,
   timeout: 3_000,
   forcedSsrTimeout: 60_000,
   maxRenderTime: 300_000,
   reuseCurrentRendering: true,
   debug: false,
+  renderingStrategyResolver: defaultRenderingStrategyResolver(
```
pawelfras, line 147: Cool! This is exactly what I had in mind 🔥

core-libs/setup/ssr/optimized-engine/rendering-strategy-resolver.ts
pawelfras, line 1: I think the logic contained in this file deserves its own unit tests.

core-libs/setup/ssr/optimized-engine/rendering-strategy-resolver.ts
kpawelczak: I believe `some` does break the loop.

core-libs/setup/ssr/optimized-engine/rendering-strategy-resolver.ts
kpawelczak: created

core-libs/setup/ssr/optimized-engine/rendering-strategy-resolver.ts
kpawelczak: fixed

core-libs/setup/ssr/optimized-engine/ssr-optimization-options.ts
kpawelczak: changed

core-libs/setup/ssr/optimized-engine/rendering-strategy-resolver.ts
pawelfras: Yes, sorry! You're right. Sorry for the confusion 👍

core-libs/setup/ssr/optimized-engine/rendering-strategy-resolver-options.ts
```diff
@@ -0,0 +1,15 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+export interface RenderingStrategyResolverOptions {
+  excludedUrls?: string[];
+  excludedParams?: string[];
+}
+
+export const defaultRenderingBlocked: RenderingStrategyResolverOptions = {
```
pawelfras:
```suggestion
export const defaultRenderingStrategyResolverOptions: RenderingStrategyResolverOptions = {
```
What do you think?

pawelfras (dismissed):

core-libs/setup/ssr/optimized-engine/rendering-strategy-resolver-options.ts
```diff
@@ -0,0 +1,16 @@
...
+export interface RenderingStrategyResolverOptions {
+  excludedUrls?: string[];
+  excludedParams?: string[];
+}
+
+export const defaultRenderingStrategyResolverOptions: RenderingStrategyResolverOptions =
+  {
+    excludedUrls: ['checkout', 'my-account'],
+    excludedParams: ['asm'],
+  };
```
pawelfras, line 16: `should:` as we discussed, this should be a part of the public API.

core-libs/setup/ssr/optimized-engine/rendering-strategy-resolver.ts
```diff
@@ -0,0 +1,53 @@
...
+  request: Request,
+  { excludedParams, excludedUrls }: RenderingStrategyResolverOptions
+) => {
+  return (
+    hasExcludedParams(request, excludedParams) ||
+    hasExcludedUrl(request, excludedUrls)
+  );
+};
+
+export const defaultRenderingStrategyResolver =
```
pawelfras, line 58: `could:` consider to add the jsdoc

pawelfras: The description of the `return` doesn't look good in the doc popup:
image

Apart from that, I wondering whether it's necessary to add `excludedUrls` and `excludedParams` to the description (or at least shouldn't be marked `@param` as they are not params directly). The same thought about `request` - it's not a direct param of the resolver.

pawelfras (approved):

core-libs/setup/ssr/optimized-engine/rendering-strategy-resolver.ts
kpawelczak: as discussed I've left options params but removed the request

core-libs/setup/ssr/optimized-engine/rendering-strategy-resolver-options.ts
kpawelczak: added it to the public api

PR #17809, opened 2023-08-31
feat: Customer360 main branch(CXSPA-700)

kimhw0630:
Closes: https://jira.tools.sap/browse/CXSPA-700


feature-libs/asm/customer-360/components/sections/asm-customer-section/asm-customer-section.component.html
```diff
@@ -0,0 +1,7 @@
+<ng-container
+  [cxComponentWrapper]="{
+    flexType: component,
+    typeCode: component,
+    uid: ''
+  }"
```
Platonn: [MAJOR QUESTION]
Can you help me understand, why we use ComponentWrapperDirective for
rendering and passing data to specific Customer 360 section components?
Why we load those components as CMS components? 

1. What are the benefits of using CMS components for this purpose? 

2. What downsides of this approach that you can see? Does using CMS components for this purpose cause a UI delay, because of a need to call for OCC CMS for this component's data? Btw. what are the data that we actually need from CMS for those components? I could see that the layout and implementation of those components is driven only by the static Spartacus config. I have problem to understand what an added value it gives us to leverage the mechanism of CMS components here.

Did you consider any other alternative approaches that doesn't use CMS components?

PS. answers to those questions affect our architecture and public API, so IMHO should be answered before releasing this feature. Otherwise making changes in the future may cause big breaking changes.

feature-libs/asm/customer-360/components/customer-360-components.module.ts
```diff
@@ -0,0 +1,119 @@
...
+        },
+        AsmCustomer360CouponComponent: {
+          component: AsmCustomerCouponComponent,
+        },
+        AsmCustomer360PromotionComponent: {
+          component: AsmCustomerPromotionComponent,
+        },
+        AsmCustomer360CustomerCouponComponent: {
+          component: AsmCustomerCustomerCouponComponent,
+        },
```
Platonn: Prerequisite question: [do we _really_ need to employ the CMS components mechanism for the purpose of Customer 360](https://github.com/SAP/spartacus/pull/17809/files#r1324461081)?

If we really need to use this mechanism, only then the following question applies:

[MEDIUM QUESTION]
I can see that CMS components types' names don't match the Angular implementation components names (and their selectors). Was it a deliberate choice to keep them different? Would you see a value in unifying the names of CMS component types, class names and their html selectors (`<cx-...>`)

It will be impossible to unify them after the release without a breaking change.

projects/schematics/src/shared/lib-configs/asm-schematics-config.ts
```diff
@@ -53,3 +58,43 @@ export const ASM_SCHEMATICS_CONFIG: SchematicConfig = {
...
+  i18n: {
+    resources: CUSTOMER_360_TRANSLATIONS,
+    chunks: CUSTOMER_360_TRANSLATION_CHUNKS_CONFIG,
+    importPath: SPARTACUS_CUSTOMER_360_ASSETS,
+  },
+  styles: {
+    scssFileName: ASM_SCSS_FILE_NAME,
+    importStyle: SPARTACUS_ASM,
+  },
+  dependencyFeatures: [ASM_FEATURE_NAME, STOREFINDER_FEATURE_NAME],
```
Platonn: [MEDIUM QUESTION]
Do you think we should add here also `ORDER_FEATURE_NAME`?

PS. can be answered/fixed in between Feature Freeze and Code Freeze

feature-libs/asm/customer-360/components/sections/asm-customer-map/asm-customer-map.component.ts
```diff
@@ -0,0 +1,151 @@
...
+import { concatMap, mapTo, take, tap } from 'rxjs/operators';
+
+import { Customer360SectionContext } from '../customer-360-section-context.model';
+
+@Component({
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  selector: 'cx-asm-customer-map',
+  templateUrl: './asm-customer-map.component.html',
+})
+export class AsmCustomerMapComponent implements OnDestroy, OnInit {
```
Platonn: [MINOR QUESTION]
Are we reusing any UI/services from the Storefinder library, e.g. to display list of stores, plug in the google maps, etc? Or we're "reinventing/copying" a lot of UI and logic to ASM library? Preferrably whe should reuse the business logic and _perhaps_ some reusable UI parts. What do you think about it?

PS. IMHO this question doesn't block the release of this feature. But it would be good to get an answer.

feature-libs/asm/components/customer-emulation/customer-emulation.component.html
```diff
@@ -4,6 +4,17 @@
     <label class="cx-asm-uid">{{ customer?.uid }}</label>
   </div>
   <cx-asm-bind-cart></cx-asm-bind-cart>
+  <!-- TODO(CXSPA-3090): Remove feature flag in 7.0 -->
```
Platonn: [MEDIUM REMARK]
let's remove this html comment. As far as I understand, otherwise it may land in customer's production applications.

feature-libs/asm/components/customer-emulation/customer-emulation.component.html
```diff
@@ -4,6 +4,17 @@
...
+  <ng-container *cxFeatureLevel="'6.3'">
```
Platonn: [MAJOR REMARK - breaking change]
We're targeting `6.5`, so let's update it to 6.5. Otherwise, our customer that configured their `featureLevel` `'6.3'` and they to bump `@spartacus/asm` package to v6.5 from v6.3 (or v6.4), then they will immediately get this feature enabled.
But the expected behavior is that they should get this feature enabled only when they configure their `featureLevel` to be `'6.5'`.

ADDITIONAL ACTION NEEDED: 
please review other parts of this PR and look for other leftovers of incorrect feature levels. (we should use `6.5`)

feature-libs/asm/components/services/asm-component.service.ts
```diff
@@ -46,6 +51,7 @@ export class AsmComponentService {
     protected winRef: WindowRef,
     // TODO(CXSPA-3090): Remove optional flag in 7.0 where service is used
     @Optional() protected asmEnablerService?: AsmEnablerService,
+    @Optional() protected routingService?: RoutingService,
     @Optional() protected asmDeepLinkService?: AsmDeepLinkService
```
Platonn: [MAJOR REMARK - breaking change]
We should add new constructor parameters only as the _last_ parameters, being optional.
In the case above the new parameter is optional, but it's added in the place of the `asmDeepLinkService`, which is a breaking change.

Why it's breaking change:
A compiling error will happen for a customer that used to extend this service and calling it's `super` constructor, and then they upgrade Spartacus to the v6.5:
```ts
export class MyAsmComponentService extends AsmComponentService {

  constructor(
    protected authService: AuthService,
    protected csAgentAuthService: CsAgentAuthService,
    protected winRef: WindowRef,
    protected asmEnablerService?: AsmEnablerService,
    protected asmDeepLinkService?: AsmDeepLinkService
  ){
     super(
       authService,
       csAgentAuthService,
       winRef,
       asmEnablerService,
       asmDeepLinkService // <---- error, the new version of this constructor expects `routingService` in this place!
  }
}

ADDITIONAL ACTION NEEDED:
Please review also other constructors in this PR and avoid similar breaking change in those files.

File: feature-libs/asm/package.json [hunk 8]
```diff
@@ -29,12 +29,15 @@
...
     "@angular/forms": "^15.2.4",
+    "@angular/platform-browser": "^15.2.4",
     "@ng-select/ng-select": "^10.0.3",
     "@ngrx/effects": "^15.3.0",
     "@ngrx/store": "^15.3.0",
     "@spartacus/cart": "6.3.0-1",
     "@spartacus/core": "6.3.0-1",
+    "@spartacus/order": "6.3.0-1",
     "@spartacus/schematics": "6.3.0-1",
+    "@spartacus/storefinder": "6.3.0-1",
```
Platonn: [MAJOR REMARK - breaking change]
Adding new required peerDependencies of a package `@spartacus/asm` is a breaking change. It will cause compilation errors for customers who used to use `@spartacus/asm` library, but didn't have installed e.g. `@spartacus/storefinder`.

When they upgrade to 6.5, they'll get compilation errors because the npm package `@spartacus/storefinder` is not installed. The fix for them would be to install this library manually (`npm install @spartacus/storefinder`).

IMHO this _unavoidable_ breaking change means to me that that this feature should be released only in a Major version of Spartacus.

Alternatively we might want to accept it (with PO's agreement) that we turn a blind eye on that breaking change, as long as it's well communicated to customers.

projects/storefrontstyles/scss/_versioning.scss
```diff
@@ -39,7 +39,7 @@ $styleVersion: $_majorVersion !default;
 // For Spartacus development or demo's, the usage of latest styles is recommended.
 // To avoid frequent changes of the $minorVersion in the application, we offer a
 // flag that can be used to keep up with the latest version at any time.
-$useLatestStyles: false !default;
+$useLatestStyles: true !default;
```
Platonn: [MAJOR REMARK - breaking change]
IMHO we shouldn't change this flag to true. It opts-in all customers for using the latest styles, regardless the _feature level_ they configured in their SCSS!
But then please make sure to bump the SCSS feature level used _only in our storefrontapp_ (in our repo) to 6.5

feature-libs/asm/customer-360/components/sections/components.ts
```diff
@@ -0,0 +1,14 @@
...
+ */
+
+export { AsmCustomerActiveCartComponent } from './asm-customer-active-cart/asm-customer-active-cart.component';
+export { AsmCustomerActivityComponent } from './asm-customer-activity/asm-customer-activity.component';
+export { AsmCustomerMapComponent } from './asm-customer-map/asm-customer-map.component';
+export { AsmCustomerProductInterestsComponent } from './asm-customer-product-interests/asm-customer-product-interests.component';
+export { AsmCustomerProductReviewsComponent } from './asm-customer-product-reviews/asm-customer-product-reviews.component';
+export { AsmCustomerProfileComponent } from './asm-customer-profile/asm-customer-profile.component';
+export { AsmCustomerSavedCartComponent } from './asm-customer-saved-cart/asm-customer-saved-cart.component';
+export { AsmCustomerSupportTicketsComponent } from './asm-customer-support-tickets/asm-customer-support-tickets.component';
```
Platonn: [MEDIUM REMARK] can be fixed in a patch release
1. all those components aren't in our public API now, because this file is not imported in `public_api.ts` (directly or transitively). So let's import this file in `public_api.ts`. Otherwise customers will have hard time importing those classes and customizing them.
2. Let's stick to the convetion of naming such "barrel" files to `index.ts` (which simply re-export API items higher in the folders structure). Let's change name of this file from `components.ts` to `index.ts`

ADDITIONAL ACTION NEEDED:
Please review the code in this PR and check if there other missing re-exports to the public API of our components/services etc.

feature-libs/asm/customer-360/components/asm-customer-promotion-listing/asm-customer-promotion-listing.component.ts
```diff
@@ -0,0 +1,38 @@
...
+  @Input() showAlertForApplyAction: boolean | null;
+  @Input() showRemoveButton: boolean;
+  @Input() showApplyButton: boolean;
+  @Input() isCustomerCoupon: boolean;
+  @Output() apply = new EventEmitter<PromotionListEntry>();
+  @Output() remove = new EventEmitter<PromotionListEntry>();
+  @Output() removeAlert = new EventEmitter();
+  @Output() removeAlertForApplyAction = new EventEmitter();
+  globalMessageType = GlobalMessageType;
+}
```
Platonn: [NITPICK - no need to change]
At the first sight such an API doesn't look ergonomic, each customization requires adding yet a new @Input. Did you consider using [ngTemplateOutlet](https://indepth.dev/posts/1405/ngtemplateoutlet) for customizations of this component?
No need to change, I'm just sharing knowledge about ngTemplateOutlet.

feature-libs/asm/customer-360/components/customer-360/customer-360.component.ts
```diff
@@ -0,0 +1,315 @@
...
+    this.subscription.unsubscribe();
+  }
+
+  get errorAlert$(): Observable<boolean> {
+    return this.showErrorAlert$.asObservable();
+  }
+
+  get errorTab$(): Observable<boolean> {
+    return this.showErrorTab$.asObservable();
+  }
```
Platonn: [MINOR REMARK] can be fixed in a patch release
Creating fresh observables in functions that are called in html template is suboptimal. In other words, let's not wrap it inside `get` function, but simply let's have a property:

```ts
errorAlert$ = this.showErrorAlert$.asObservable();
...
```

Why it was suboptimal (in terms of performance): on each change detection cycle of this component, the HTML template is evaluated and the functions of the getters used in the HTML tempalte are called.
Then a fresh observable is created (becasue calling $something.asObservable() - which creates a fresh observable). And then `async` pipe in template unsubscribes from the previous observable and subscribes to the new observable returned from this getter.

ADDITIONAL ACTION NEEDED:
Please review this PR for other similar places of creating a fresh observable within a function that is called in HTML template.

feature-libs/asm/customer-360/components/customer-360/customer-360.component.ts
```diff
@@ -0,0 +1,315 @@
...
+      }
+    } else if (this.isForwardsNavigation(event)) {
+      selectedIndex++;
+      if (selectedIndex > maxTab) {
+        selectedIndex = 0;
+      }
+    } else if (event.code === 'Home') {
+      selectedIndex = 0;
+    } else if (event.code === 'End') {
+      selectedIndex = maxTab;
```
Platonn: [NITPICK] can be fixed anytime
Let's not use hardcoded values in component TS, but rather some shared constants or Enum.

feature-libs/asm/customer-360/components/sections/asm-customer-profile/asm-customer-profile.component.ts
```diff
@@ -0,0 +1,92 @@
...
+        return card;
+      })
+    );
+  }
+
+  getCardIcon(code: string): string {
+    let ccIcon: string;
+    if (code === 'visa') {
+      ccIcon = this.iconTypes.VISA;
+    } else if (code === 'master' || code === 'mastercard_eurocard') {
```
Platonn: [NITPICK] can be fixed anytime
Let's not use hardcoded values in component TS, but rather some shared constants or Enum.

feature-libs/storefinder/core/facade/store-finder.service.ts
```diff
@@ -56,6 +56,7 @@ export class StoreFinderService implements OnDestroy {
 
   /**
    * Returns observable for store's entities
+   * TODO: The return value of thism ethod signature is wrong, should be StoreFinderSearchPage.
```
Platonn: [MINOR REMARK] 
1. let's create a ticket for it instead of putting a raw TODO
2. I believe SonarQube checks in CI won't let you merge with new `TODO` comments included, so I guess you'll need to remove it or rephrase not to be `TODO` 😅

Platonn:
My intention wasn't doing deep code review, but instead I've scanned the PR, looking for architectural questions and concerns. By the way, I found a few places I had to comment on, e.g. breaking changes introduced or minor/medium comments affecting the public API.

IMHO one question (about understanding why we've used CMS components) is very important from architectural point of view - to understand the pros and cons of this solution, to eventually understand if such an approach was really justified.  
What's why I'd like to wait with merging this PR until I understand this decision and the consequences.

Moreover, I've heard that each sub-PR for this Epic breanch was individually reviewed. And I've heard this PR was co-created by various teams and handed over from one team to another over the months. 
That's why IMHO it would good if someone reviews it holistically soon, before we give the final green light to merge.

projects/schematics/src/shared/lib-configs/asm-schematics-config.ts
kimhw0630: Thank you. I will discuss this with Radhep.

feature-libs/asm/customer-360/components/sections/asm-customer-map/asm-customer-map.component.ts
kimhw0630: We only utilize services from the Storefinder library. Displaying a map simply involves embedding an iframe where we provide a map URL. The design for displaying stores differs from ours, as our design features a list view, whereas Storefinder employs a tile view

feature-libs/asm/components/customer-emulation/customer-emulation.component.html
kimhw0630: got it

feature-libs/asm/components/customer-emulation/customer-emulation.component.html
kimhw0630: yes, this should be 6.5.

feature-libs/asm/components/services/asm-component.service.ts
kimhw0630: Thank you. It seems that this occurred after we merged the 6.5.x branch into our own. we will move routingService to the end.

projects/schematics/src/shared/lib-configs/asm-schematics-config.ts
kimhw0630: I have discussed it with Radhep, and we should include 'ORDER_FEATURE_NAME' in the list.  Thanks

feature-libs/asm/package.json
Platonn: UPDATE:
We've done an ultimate test and confirm there were no npm warnings during npm update
nor compilation errors after the update, when building the app. 

**The test was made this way:**
1. create a fresh Angular 15 app - `npx @angular/cli@15 new test-app`
1. install Spartacus 6.5 with ASM (e.g. using RBSC as npm repository) - `ng add @spartacus/schematics)`, BUT without Storefinder(!) - uncheck the `Store Finder` feature checkbox in the installation schematics prompt
1. verify in `package.json` of the fresh app there is no `@spartacus/storefinder` mentioned
1. verify in `node_modules` of the fresh app there is no `node_modules/@spartacus/storefinder` folder
1. then build the libraries on our local epic branch and publish them to local verdaccio npm repo (for more see, [docs](https://github.com/SAP/spartacus/blob/0dd42d4ab0121265024ee092d7cb3ef8640bf475/projects/schematics/README.md#L16) and our [script ](https://github.com/SAP/spartacus/blob/0dd42d4ab0121265024ee092d7cb3ef8640bf475/tools/schematics/testing.ts#L215) for publishing to local verdaccio repository) 
1. then in the fresh app bump the version of spartacus:
   1. change `~6.5.0-1` and `^6.5.0-1` versions of `@partacus/...` packages to _exaclty_ `6.5.0-1` (with no tilde or carret!)
   3. run `npm update`
1. Verify there are no warnings during the `npm update`
2. Verify the `package.json` of the fresh app _now_ contains `@spartacus/storefinder` entry (it was installed implicitly)
3. (Refresh the files tree view in your IDE and) verify the folder `node_modules/@spartacus/storefinder` _now_ exists

feature-libs/asm/components/services/asm-component.service.ts
kimhw0630: I used `inject`
` protected routingService = inject(RoutingService);`

feature-libs/asm/customer-360/components/sections/asm-customer-section/asm-customer-section.component.html
Platonn: As discussed offline, CMS components were not necessary. They are removed in separate PR https://github.com/SAP/spartacus/pull/17874

feature-libs/asm/customer-360/components/customer-360-components.module.ts
Platonn: All names unified in PR https://github.com/SAP/spartacus/pull/17878
Thanks Hakwoo!

feature-libs/asm/package.json
Platonn: For reference:
**Reason why there were no npm warnings:**
During the `npm update`, NPM discoveres that the existing primary dependency `@spartacus/asm` has a new peerDependency (`@spartacus/storefinder`) and installs it automatically behind the scenes. That said, NPM doesn't put `@spartacus/storefinder` explicitly in `package.json` as a primary dependency, but you can find notion about `@spartacus/storefinder` only in the updated `package-lock.json` and also in the newly-added folder `node_modules/@spartacus/storefront`.

**Reason why there were no compilation errors:**
Since the `@spartacus/storefront` library is installed impliclty by `npm update`, then there is no potential reason for compilation error due to lack of `@spartacus/storefront`

Platonn (dismissed):

Platonn (approved):

PR #17813, opened 2023-09-05
CXSPA-4538 Allow empty control value and display placeholder

ChristophHi:


feature-libs/quote/components/quote-seller-edit/quote-seller-edit.component.ts
```diff
@@ -70,11 +72,12 @@ export class QuoteSellerEditComponent implements OnInit, OnDestroy {
               )
             );
           this.form.controls.discount.addValidators([numberFormatValidator]);
-          this.form.controls.discount.setValue(
-            localizationElements.formatter.format(
-              quote.quoteDiscounts?.value ?? 0
-            )
-          );
+          const discountValue = quote.quoteDiscounts?.value ?? 0;
```
Uli-Tiger:
```suggestion
          const discountValue = quote.quoteDiscounts?.value;
```
Is this default still required? - tests are still green if i delete it.

Uli-Tiger (changes requested):

feature-libs/quote/components/quote-seller-edit/quote-seller-edit.component.ts
ChristophHi: Ah yes you are right. Forgot to delete that after my changes

Uli-Tiger (approved):

PR #17825, opened 2023-09-13
feat: disable SSR in smartedit

kpawelczak:


pawelfras (approved):

pawelfras:
Regarding the default resolver itself, I'm wondering whether we should add official information about another resolve, how to use it and for which params/urls we're disabling SSR by default. Let's discuss with Kris

PR #17827, opened 2023-09-13
CXSPA-4705 Take quote/cart link from OCC

ChristophHi:


feature-libs/quote/root/guards/quote-cart.service.spec.ts
```diff
@@ -1,48 +1,77 @@
...
 
-  describe('setQuoteId', () => {
-    it('should trigger emission of new quote id', (done) => {
-      quoteCartService.setQuoteId(QUOTE_CODE);
+  describe('setCheckoutActive', () => {
+    it('should trigger emission of new state', (done) => {
+      quoteCartService.setCheckoutAllowed(true);
+      quoteCartService.isCheckoutAllowed().subscribe((isAllowed) => {
+        expect(isAllowed).toBe(true);
+        done();
```
Uli-Tiger: done is not required here - as all the mocks used emit immediately using of() operator, etc.
so the test runs through sequentially without any real asynchronicity.
Same is  true for the following tests.

Uli-Tiger:
seems like the PR breaks the edit quote case.
1) login as gi sun.
2) open an existing draft quote
3) make sure your cart is not link to a quote (e.g click on new cart in quote details) and open the draft again
4) click on edit
issue: the mini cart quantity updates from 0 to the quantity of the quote, so it seems that the cart is indeed linked with the quote, **but the quote remains in read-only mode.**

Uli-Tiger (changes requested):
see also my remark regarding the edit mode

ChristophHi:
> seems like the PR breaks the edit quote case.
> 
> 1. login as gi sun.
> 2. open an existing draft quote
> 3. make sure your cart is not link to a quote (e.g click on new cart in quote details) and open the draft again
> 4. click on edit
>    issue: the mini cart quantity updates from 0 to the quantity of the quote, so it seems that the cart is indeed linked with the quote, **but the quote remains in read-only mode.**

Cannot reproduce it. Please verify you are using the correct commerce backend release

Uli-Tiger (approved):

PR #17856, opened 2023-09-20
CXSPA-4783 Add dialog for checkout action and adapt guard

ChristophHi:


Uli-Tiger (changes requested):
Accept an Vendor Quote and order it.
Afterwards go to catalog and add an item to cart
**issue: Cart Not Found error**
try again - now it works

-- tried it again -- now it worked

Uli-Tiger (approved):

PR #17865, opened 2023-09-21
feat: CXSPA-3781 SSR - Multi-provided error interceptors

pawelfras:
This pull request introduces methodologies for integrating multiple error interceptors that manage errors within the Server-Side Rendering (SSR) framework. This architectural augmentation preserves backward compatibility, mitigating any potential disruptions for end-users upon the incorporation of new error interceptors into the system.

With the introduction of this enhancement, it becomes easier for users to include new error interceptors, giving them the flexibility to determine the order in which these interceptors are applied within the system. This priority setting allows users to control how these interceptors operate and influence the workflow of the system.

The order is:
 1. High priority
 2. Normal or no priority
 3. Low priority
 
 Preserves the original order within a group of interceptors with the same priority.

closes [CXSPA-3781](https://jira.tools.sap/browse/CXSPA-3781)


projects/core/src/error-handling/error-interceptors/error-interceptor.ts
```diff
@@ -0,0 +1,55 @@
...
+ * The order is:
+ * 1. High priority - the latest added with the high priority is first
+ * 2. No priority - the order is not changed
+ * 3. Low priority - the latest added with the low priority is last
+ *
+ * @public
+ */
+export enum ErrorInterceptorPriority {
+  HIGH = -1,
+  LOW = 1,
```
Platonn: [MEDIUM]
At the first sight, it looks strange, that HIGH has negative value and LOW has positive value. What about swaping it?
btw. for completeness we could also explicitly include here the default

```suggestion
export enum ErrorInterceptorPriority {
  HIGH = 1,
  DEFAULT = 0,
  LOW = -1,
```

projects/core/src/error-handling/error-interceptors/error-interceptor.ts
```diff
@@ -0,0 +1,55 @@
...
+ * Error interceptors can be provided with the proper priority. The higher the priority, the earlier the interceptor will be called.
+ *
+ * @member priority - Priority of the error interceptor. The higher the priority, the earlier the interceptor will be called. Parameter is optional.
+ * @method interceptError - Intercepts the error and passes it to the next interceptor in the chain if provided.
+ *
+ * @public
+ */
+export interface ErrorInterceptor {
+  priority?: ErrorInterceptorPriority;
+  interceptError(error: unknown, next?: ChainedErrorInterceptorFn): void;
```
Platonn: [MEDIUM] 
It's already an `Error`Interceptor class. For breviety, what about dropping word `Error` from the method name?

projects/core/src/error-handling/error-interceptors/error-interceptor.ts
```diff
@@ -0,0 +1,55 @@
...
+}
+
+/**
+ * Injection token for error interceptors responsible for intercepting errors handled by the CxErrorHandler.
```
Platonn: [NITPICK]
I have problem to understand this sentence. What about simply:
```suggestion
 * Injection token for Error Interceptors.
```

projects/core/src/error-handling/error-interceptors/error-interceptor.ts
```diff
@@ -0,0 +1,55 @@
...
+ * SPDX-FileCopyrightText: 2023 SAP Spartacus team <spartacus-team@sap.com>
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { InjectionToken } from '@angular/core';
+
+/**
+ * ChainedErrorInterceptorFn is a function that is passed to the error interceptor.
+ * It is responsible for passing the error to the next interceptor in the chain if provided.
```
Platonn: [MINOR]
What about: 
```suggestion
 * It is responsible for handling an error in some way. If the error is not resolved by this interceptor, it should pass the original error (or a new error) to the next interceptor in the chain.
```

projects/core/src/error-handling/cx-error-handler.ts
```diff
@@ -4,14 +4,56 @@
...
+  ErrorInterceptor,
+} from './error-interceptors/error-interceptor';
+import {
+  handleInterceptors,
+  sortErrorInterceptors,
+  tailChain,
+} from './utils/error-interceptor-utils';
 
+/**
+ * The CxErrorHandler is the default ErrorHandler for Spartacus in SSR mode.
```
Platonn: [MEDIUM]
It's always a default ErrorHandler in Spartacus.
```suggestion
 * The CxErrorHandler is the default ErrorHandler for Spartacus.
```

projects/core/src/error-handling/cx-error-handler.ts
```diff
@@ -4,14 +4,56 @@
...
+    if (this.featureConfig.isLevel('6.6')) {
+      // Similar to Angular's interceptors, error interceptors are organized from right to left,
+      // ensuring that the ultimate execution order is from left to right.
+      // In other words, if the interceptors array contains `[a, b, c]`,
+      // our goal is to create a chain that can be envisioned as c(b(a(end))),
+      // constructed by progressively adding elements from the innermost to the outermost.
+      const chain = this.errorInterceptors.reduceRight(
+        (next, interceptor) => handleInterceptors(next, interceptor),
+        tailChain
+      );
```
Platonn: [MEDIUM] (micro) performance improvement:
I believe it should suffice to calculate the chained function only once, at the construction of `CxErrorHandler`, instead of doing it for every appearing Error separately. But I might be missing something important here... What do you think?

projects/core/src/error-handling/error-interceptors/logger-error.interceptor.ts
```diff
@@ -0,0 +1,27 @@
...
+
+import { Injectable, inject } from '@angular/core';
+import { LoggerService } from '../../logger';
+import {
+  ChainedErrorInterceptorFn,
+  ErrorInterceptor,
+  ErrorInterceptorPriority,
+} from './error-interceptor';
+
+Injectable({
```
Platonn: [MEDIUM]
Missing `@`. Does it compile?
```suggestion
@Injectable({
```

projects/core/src/error-handling/error-interceptors/logger-error.interceptor.ts
```diff
@@ -0,0 +1,27 @@
...
+});
+export class LoggerErrorInterceptor implements ErrorInterceptor {
+  priority = ErrorInterceptorPriority.HIGH;
+  logger = inject(LoggerService);
+  interceptError(error: Error, next?: ChainedErrorInterceptorFn): void {
+    this.logger.error(error);
+    if (next) {
+      next(error);
+    }
+  }
```
Platonn: [MEDIUM]
IMHO the `next` param should be mandatory: `next?: ChainedErrorInterceptorFn`

It's not responsibility of the error interceptor to _check_ whether the next interceptor exists. The `next` parameter of the method should be mandatory and always present IMHO.
For this to happen, please pass noop function `()=>{}` as the `initial value` parameter to the `reduceRight()`

projects/core/src/error-handling/utils/error-interceptor-utils.ts
```diff
@@ -0,0 +1,71 @@
...
+export const sortErrorInterceptors = (
+  errorInterceptors: ErrorInterceptor[]
+): ErrorInterceptor[] => {
+  const highPriorityErrorInterceptors = new Array<ErrorInterceptor>();
+  const lowPriorityErrorInterceptors = new Array<ErrorInterceptor>();
+  const noPriorityErrorInterceptors = new Array<ErrorInterceptor>();
+
+  errorInterceptors.forEach((handler) => {
+    if (handler.priority === ErrorInterceptorPriority.HIGH) {
+      highPriorityErrorInterceptors.unshift(handler);
```
Platonn: [MAJOR]
Why do we change the original order among the HIGH priority interceptors?
Example:
The original order A,B:
```
{ name: 'A', priority: HIGH }
{ name: 'B', priority: HIGH}
```
will result in: `B, A`

--- 

My intention was to only move groups in between each other, but not change order within the groups. But maybe you have some arguments for changing the order _just_ for the group of HIGHs - then I'm happy to learn them.

---

PS. I've elaborated more (with implementation idea) in another comment https://github.com/SAP/spartacus/pull/17865#discussion_r1333223082

projects/core/src/error-handling/utils/error-interceptor-utils.ts
```diff
@@ -0,0 +1,71 @@
...
+ */
+
+import {
+  ChainedErrorInterceptorFn,
+  ErrorInterceptor,
+  ErrorInterceptorPriority,
+} from '../error-interceptors/error-interceptor';
+
+/**
+ * Sorts error interceptors based on priority.
```
Platonn: [MINOR]
Let's change this doc, only when we agree on the strategy of storting
```suggestion
 * Sorts error interceptors based on priority. Preserves the original order within a group of interceptors with the same priority.
```

projects/core/src/error-handling/utils/error-interceptor-utils.ts
```diff
@@ -0,0 +1,71 @@
...
+ * @returns sorted error interceptors
+ *
+ * @internal
+ */
+export const sortErrorInterceptors = (
+  errorInterceptors: ErrorInterceptor[]
+): ErrorInterceptor[] => {
+  const highPriorityErrorInterceptors = new Array<ErrorInterceptor>();
+  const lowPriorityErrorInterceptors = new Array<ErrorInterceptor>();
+  const noPriorityErrorInterceptors = new Array<ErrorInterceptor>();
```
Platonn: [MINOR]
Now this mechanism is fixed to only 3 possible priority values, for no good reason. At the same time, I don't expect anyone using plenty of various interceptor priority values. That said, I'd like our mechanism to be effortlessly open for customers using many custom various priority numbers. Generic solution could use a hashmap like in the example below:
```ts
const interceptorsGroupedByPriority = new Map<ErrorInterceptorPiority,ErrorInterceptors[]>()


// 1. put interceptors to proper groups (preserving the original order within each group):
errorInterceptors.forEach((interceptor => addToProperGroup(interceptor))

function addToProperGroup(interceptor) {
  //  i. create group if empty
  // ii. push to the group
}

// 2. get the list of all different various unique priority values (in case when customer used also values like 3 or -3)

// 3. sort the list of different priorities (e.g. so it becomes `[-3, -1, 0, 1, 3]`
// why? Because Object.keys may potentially return the Map's keys in a random order.

// 4. concat the whole groups next to each other, in the order of their priorites
return
```

Expected example, where H=High, L=LOW, D=Default; numbers are just additional identifiers):
```
input:  [H1, D1, L1, H2, D2, L2]
output: [H1, H2, D1, D2, L1, L2]
```
Please note in the result, the group of `H`s is before the group of `D`s, which is before the group of `L`s. And that the original order within each group is preserved  - `1` is before `2` inside each group

projects/core/src/error-handling/utils/error-interceptor-utils.ts
```diff
@@ -0,0 +1,71 @@
...
+// maybe this is a good place to forward an error to the ExpressJS?
+/**
+ * The last function in the chain. Does nothing.
+ * @param _error - error
+ *
+ * @internal
+ */
+export const tailChain: ChainedErrorInterceptorFn = (_error: Error) => {
+  // do nothing
+};
```
Platonn: [MEDIUM]
I believe the `priority` mechanism already allows for putting stuff at the end (expecially when we make it open for new custom priority values like `VERY_LOW = -10`

projects/core/src/error-handling/utils/error-interceptor-utils.spec.ts
```diff
@@ -0,0 +1,103 @@
+import {
+  ErrorInterceptor,
+  ErrorInterceptorPriority,
+} from '../error-interceptors/error-interceptor';
+import {
+  handleInterceptors,
+  sortErrorInterceptors,
+} from './error-interceptor-utils';
+
+describe('error interceptor utils', () => {
```
Platonn, line 10: unit tests may need to be adjusted, depending on the agreed strategy of storting

projects/core/src/error-handling/cx-error-handler.ts
pawelfras: For some reason, the Angular team is doing a similar calculation in the `handle` inside the `HttpInterceptorHandler` method each time it's invoked and I'm wondering whether it's related to the fact that there may be scenarios that customers provide `HTTP_INTERCEPTOR` after `HttpInterceptorHandler` is constructed. Thanks to calculating the chain in the `handle` method, it's sure all interceptors are taken into account. I would love to discuss this.

projects/core/src/error-handling/error-interceptors/logger-error.interceptor.ts
pawelfras: yes, without any errors! that's strange...

projects/core/src/error-handling/utils/error-interceptor-utils.ts
pawelfras: In that case, I misunderstood the reason behind it. During the discussion we had we talked about two options - introduce a `priority` parameter or let users know that they need to provide their interceptors before ours (so they will be invoked before ours). I wanted to achieve the same behaviour as is in option 2, but using the `priority` param for better readability. I believe customers may want to do something with an error before logging in. Let's discuss this.

projects/core/src/error-handling/utils/error-interceptor-utils.ts
pawelfras: We didn't consider such a scenario during discussions so I didn't implement the logic in like this. Actually, such a way of sorting may resolve the problem described in this comment: https://github.com/SAP/spartacus/pull/17865#discussion_r1333989010

projects/core/src/error-handling/utils/error-interceptor-utils.ts
pawelfras: Ok, so let's keep this function empty.

projects/core/src/error-handling/utils/error-interceptor-utils.ts
pawelfras: Actually, such a way of sorting may resolve the problem described in this comment: https://github.com/SAP/spartacus/pull/17865#discussion_r1333989010

projects/core/src/error-handling/utils/error-interceptor-utils.ts
pawelfras: As we discussed, the sorting function has been modified.

projects/core/src/error-handling/cx-error-handler.ts
pawelfras: As we discussed, `reduceRight` stays inside `handleError` because sorting is done once at the construction of `CxErrorHandler`

projects/core/src/error-handling/utils/error-interceptor-utils.ts
pawelfras: The sorting has been changed to address comments from [this](https://github.com/SAP/spartacus/pull/17865#discussion_r1333196192) thread

projects/core/src/error-handling/cx-error-handler.ts
```diff
@@ -27,33 +26,26 @@ import {
  */
 @Injectable()
 export class CxErrorHandler implements ErrorHandler {
-  //TODO: Verify whether feature flag in this case is justified.
-  // if yes - Add Jira ticket to remove in next major release or after 6 months
-  private featureConfig = inject(FeatureConfigService);
-  logger = inject(LoggerService);
+  /**
+   * @deprecated Since 6.6
```
Platonn: [MINOR]

```suggestion
* @deprecated Since 6.6 - `LoggerService` is unused here. Instead it's now used in `LoggerErrorInterceptor`
```

projects/core/src/error-handling/cx-error-handler.ts
```diff
@@ -27,33 +26,26 @@ import {
...
+   */
+  protected logger = inject(LoggerService);
+  protected injector = inject(Injector);
 
-  injector = inject(Injector);
-
-  errorInterceptors: ErrorInterceptor[] = sortErrorInterceptors(
+  protected errorInterceptors: ErrorInterceptor[] = sortErrorInterceptors(
     this.injector.get(ERROR_INTERCEPTORS, [])
   );
```
Platonn: [MEDIUM]
Why using `Injector`? Can't we simply `inject(ERROR_INTERCEPTORS, [])`

projects/core/src/error-handling/cx-error-handler.spec.ts
```diff
@@ -70,55 +61,6 @@ describe('CxErrorHandler', () => {
     ).and.callThrough();
   });
 
-  describe('error interceptors feature flag', () => {
```
Platonn: [MEDIUM]
Can we have some unit test that checks whether we chain the interceptors in the correct way?
1. e.g. provide 3 interceptors. Then pass an error to the chain. Then assert that the error was handled by A,B,C in the correct order?
2. e.g. provide 3 interceptors: A,B,C. Mock B so it doesn't pass an error to the _next_ interceptor. Then pass an error to the chain. Assert that the error was handled only by A and B, but not C.

projects/core/src/error-handling/cx-error-handler.ts
```diff
@@ -27,33 +26,26 @@ import {
...
-      this.logger.error(error);
-    }
+    // Similar to Angular's interceptors, error interceptors are organized from right to left,
+    // ensuring that the ultimate execution order is from left to right.
+    // In other words, if the interceptors array contains `[a, b, c]`,
+    // our goal is to create a chain that can be envisioned as c(b(a(end))),
+    // constructed by progressively adding elements from the innermost to the outermost.
+    const chain = this.errorInterceptors.reduceRight(
+      (next, interceptor) => handleInterceptors(next, interceptor),
+      tailChain
```
Platonn: [NITPICK]
`tailChain` vs `noop` vs `()=>{}` ?

projects/core/src/error-handling/utils/error-interceptor-utils.ts
```diff
@@ -69,3 +58,45 @@ export const handleInterceptors = (
...
+  interceptorsGroupedByPriority: Map<
+    ErrorInterceptorPriority,
+    ErrorInterceptor[]
+  >
+) => {
+  // sort interceptors by priority, from highest to lowest
+  return Array.from(interceptorsGroupedByPriority)
+    .sort((a, b) => b[0] - a[0])
+    .map(([, interceptors]) => interceptors)
+    .flat();
```
Platonn: [MINOR]
I'm having hard time understanding this code. Although it's written very shortly and I'm not sure if readable (maybe it's only for me...). Did you consider breaking it into few steps with named variables?

Questions that I ask myself reading this:
- what is actually being sorted?
- why do we have to pass to `sort()` a custom comparator
- why in the comparator we take the 0th elements of the `a` and `b`. What are `a` and `b` by the way?
- why the sorted array contains elements that are tuples (that are destructured to `[, interceptors]`

(Note: flat() is intuitive and justified at the first sight - as I expect we want to concatenate many arrays).

---
Feel free to ignore the code snippet below. I'm sharing it only for inspiration - a _boring_ code that I can imagine:
```ts
// get the list of various unique priorites
const priorites = Object.keys(interceptorsGroupedByPriority)

// sort the list of priorites
const sortedPriorites = priorities.sort() // this should return array like [-1, 0, 1]. Why did we sort keys? Because Object.keys may potentially return the Map's keys in a random order.

// concat the groups in the order of priorites
const sortedInteceptors = priorites
  .map(priority => interceptorsGroupedByPriority.get(priority))
  .flat();

return sortedInteceptors;
```

projects/core/src/error-handling/error-interceptors/logger-error.interceptor.ts
Platonn: If it didn't throw any errors, was this class even used in runtime? Did you check if our app is still logging stuff to the console in case of errors?

projects/core/src/error-handling/error-interceptors/logger-error.interceptor.ts
Platonn: I can see this comment is not addressed yet. I believe we should:
1. make `next` required
2. remove the condition `if(next)`, but simply call `next(error)`

projects/core/src/error-handling/cx-error-handler.spec.ts
pawelfras: AD 1. test `should call all error interceptors` does exactly what you described. I've changed its desc to `should call all error interceptors in the correct order`
 
AD 2. I'll add such a test case.

projects/core/src/error-handling/cx-error-handler.ts
pawelfras: Even if the function does nothing, I like its name and description is related to the error interceptor chain. I'd like to keep it as it is.

projects/core/src/error-handling/cx-error-handler.spec.ts
Platonn: My bad. I missed that. Thanks!

projects/core/src/error-handling/error-interceptors/error-interceptor.service.spec.ts
```diff
@@ -0,0 +1,154 @@
...
+
+      errorInterceptorService = TestBed.inject(ErrorInterceptorService);
+    });
+
+    it('should break the chain if any interceptor does not pass an error to the next interceptor', () => {
+      errorInterceptorService.interceptorsChain(new Error('test error'));
+      expect(interceptorsOrder).toEqual([
+        'C - high priority',
+        'A - normal priority',
+        'No Pass',
```
Platonn, line 150: Nice

Gandalf: You shall not pass!

projects/core/src/error-handling/error-interceptors/error-interceptor.service.ts
```diff
@@ -0,0 +1,121 @@
...
+})
+export class ErrorInterceptorService {
+  protected errorInterceptors: ErrorInterceptor[] =
+    inject(ERROR_INTERCEPTORS, { optional: true }) ?? [];
+
+  protected sortedErrorInterceptors: ErrorInterceptor[] =
+    this.sortErrorInterceptors(this.errorInterceptors);
+
+  get interceptorsChain(): ChainedErrorInterceptorFn {
+    console.log('this.sortedErrorInterceptors', this.sortedErrorInterceptors);
```
Platonn: 👀 
```suggestion
```

projects/core/src/error-handling/error-interceptors/error-interceptor.service.ts
```diff
@@ -0,0 +1,121 @@
...
+   *
+   */
+  protected handleInterceptors(
+    next: ChainedErrorInterceptorFn,
+    interceptor: ErrorInterceptor
+  ): ChainedErrorInterceptorFn {
+    return (error: unknown) => {
+      interceptor.intercept(error, next);
+    };
+  }
```
Platonn: [MINOR]
`chain` vs. `handle`?

```suggestion
  protected chainInterceptors(
    partialChain: ChainedErrorInterceptorFn,
    interceptor: ErrorInterceptor
  ): ChainedErrorInterceptorFn {
    return (error: unknown) => {
      interceptor.intercept(error, partialChain);
    };
  }
```

projects/core/src/error-handling/error-interceptors/error-interceptor.service.ts
```diff
@@ -0,0 +1,121 @@
...
+
+  get interceptorsChain(): ChainedErrorInterceptorFn {
+    console.log('this.sortedErrorInterceptors', this.sortedErrorInterceptors);
+    // Similar to Angular's interceptors, error interceptors are organized from right to left,
+    // ensuring that the ultimate execution order is from left to right.
+    // In other words, if the interceptors array contains `[a, b, c]`,
+    // our goal is to create a chain that can be envisioned as c(b(a(end))),
+    // constructed by progressively adding elements from the innermost to the outermost.
+    return this.sortedErrorInterceptors.reduceRight<ChainedErrorInterceptorFn>(
+      (next, interceptor) => this.handleInterceptors(next, interceptor),
```
Platonn: [MINOR]
`chain` vs. `handle`?

```suggestion
      (partialChain, interceptor) => this.chainInterceptors(partialChain, interceptor),
```

Platonn (dismissed):

Platonn (approved):

Platonn (approved):

PR #17873, opened 2023-09-25
feat: Multi-config checkout integration

Matejk00:
Closes: CXSPA-4704


feature-libs/checkout/base/components/services/checkout-config.service.ts
```diff
@@ -20,7 +21,30 @@ export class CheckoutConfigService {
   private defaultDeliveryMode: Array<DeliveryModePreferences | string> =
     this.checkoutConfig.checkout?.defaultDeliveryMode || [];
 
-  constructor(private checkoutConfig: CheckoutConfig) {}
+  protected checkoutFlow =
+    this.checkoutFlowOrchestratorService?.getCheckoutFlow();
+
+  // TODO(NO_TICKET): make checkoutFlowOrchestratorService a required dependency
```
rmch91: Why NO_TICKET?

feature-libs/checkout/base/components/services/checkout-config.service.ts
```diff
@@ -20,7 +21,30 @@ export class CheckoutConfigService {
...
+  constructor(
+    checkoutConfig: CheckoutConfig,
+    // eslint-disable-next-line @typescript-eslint/unified-signatures
+    checkoutFlowOrchestratorService: CheckoutFlowOrchestratorService
+  );
+  /**
+   * @deprecated since 6.6
```
rmch91: What do you think about usage of inject() instead of adding dependency to constructor to avoid breaking changes? You can take a look on Kris last message on tribe for the reference.

feature-libs/checkout/base/components/services/checkout-step.service.ts
```diff
@@ -49,10 +50,28 @@ export class CheckoutStepService {
...
 
+  // TODO(NO_TICKET): make checkoutFlowOrchestratorService a required dependency
+  constructor(
+    routingService: RoutingService,
+    checkoutConfig: CheckoutConfig,
+    routingConfigService: RoutingConfigService,
+    // eslint-disable-next-line @typescript-eslint/unified-signatures
+    checkoutFlowOrchestratorService: CheckoutFlowOrchestratorService
+  );
+  /**
```
rmch91: Same comment as above

rmch91 (changes requested):

feature-libs/checkout/base/components/services/checkout-config.service.ts
Matejk00: Good idea, done.

feature-libs/checkout/base/components/services/checkout-step.service.ts
Matejk00: Changed.

feature-libs/checkout/base/components/services/checkout-config.service.ts
Matejk00: If we use `inject` instead there is no need to create a ticket for the future removal.

rmch91 (approved):

PR #17875, opened 2023-09-26
feat: add throwPaymentError global function

FollowTheFlo:
goal: add a global function that can be called by third party script to popup a simple window with message & confirm labels.
image

To test:
npm run start:opf
In Checkout, select AdyenHostedField
Open the console (F12) Paste:
```
window.Opf.payments.throwPaymentError({
    "messageString": "This is an error message",
    "confirmString": "OK",
})
```
press enter


integration-libs/opf/checkout/styles/_index.scss
FollowTheFlo, line 1: moved from 'opf/checkout' to 'opf' level, so includes both 'opf/checkout' and 'opf-base' selectors.

integration-libs/opf/base/components/opf-error-modal/opf-error-modal.component.ts
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  constructor(
+    protected launchDialogService: LaunchDialogService,
+    protected el: ElementRef,
+    protected cd: ChangeDetectorRef,
+    protected opfErrorModalService: OpfErrorModalService
+  ) {
+    // Mechanism needed to trigger the cpnt life cycle hooks.
+    timer(1).subscribe({
```
FollowTheFlo, line 50: This observable which immediately completes and forces Change detector to render, it is used to trigger the cpnt hooks (ngOnInit..).
When this cpnt is called from global function (eg: from console), hooks are not triggered for reason I am not fully sure, I belive setTimeout run behind the hood of timer(1) thus trigger event from zone.js, that would explain why it does not work if using EMPTY.subscribe(...)
It is the best solution I found, I tried to play with ngZone or some other ChangeDetector config...

integration-libs/opf/base/components/opf-error-modal/opf-error-modal.service.ts
```diff
@@ -0,0 +1,69 @@
...
+      })
+    );
+  }
+
+  protected getLabelTranslation(
+    defaultKey: string,
+    labelString?: string,
+    labelKey?: string,
+    labelReplacements?: any
+  ) {
```
FollowTheFlo, line 48: below is business logic with fallbacks
 labelString -> labelKey with Replacements -> labelKey -> default labelKey
 tested in unit tests

rmch91 (approved):

PR #17885, opened 2023-09-28
feat: Update SPA(ccv2) to adapt to CMS OCC API changes

kpawelczak:
ticket: https://jira.tools.sap/browse/CXSPA-4601


projects/core/src/occ/adapters/cms/legacy-occ-cms-component.adapter.ts
```diff
@@ -44,7 +44,7 @@ export class LegacyOccCmsComponentAdapter extends OccCmsComponentAdapter {
 
     return this.http
       .post<Occ.ComponentList>(
-        this.getComponentsEndpoint(requestParams, fields),
+        this.getComponentsEndpoint(requestParams, fields, ''),
```
Matejk00: We are passing empty string as a third argument (userId). Shouldn't be here any default value like: 'anonymous' or 'current'? Or maybe we can make this argument as an optional?

Matejk00 (changes requested):

projects/core/src/occ/adapters/cms/legacy-occ-cms-component.adapter.ts
kpawelczak: Tough question. This is the legacy adapter. The old backend will not be compatible with the new endpoint. If I add the default value, the current adapter will add userId params. Making this value optional would be incorrect in my opinion because with the new API we need it. Now I don't know the use case of this legacy adapter so it's hard for me to say what to do with it. With this change the users have to be prepared to use old endpoint.

kpawelczak: removed it and made userId optional

projects/core/src/occ/adapters/cms/occ-cms-page.adapter.ts
```diff
@@ -44,16 +49,34 @@ export class OccCmsPageAdapter implements CmsPageAdapter {
...
-
-    return this.http
-      .get(endpoint, { headers: this.headers })
-      .pipe(this.converter.pipeable(CMS_PAGE_NORMALIZER));
+          return this.http.get(endpoint, { headers: this.headers });
+        }),
+        this.converter.pipeable(CMS_PAGE_NORMALIZER)
+      );
+    } else {
+      const endpoint = !pageContext.type
```
Matejk00: Hopefully, we can refactor this section of code to remove it from the else statement. If the condition `this.featureConfigService.isEnabled('newCmsEndpoint')` evaluates to true, the function will return a value without proceeding further. Otherwise, if the condition is false, the if block will be skipped, and a different value will be returned.

There is no need to nesting a code if it's not necessary.

projects/core/src/occ/adapters/cms/occ-cms-component.adapter.ts
```diff
@@ -53,31 +69,78 @@ export class OccCmsComponentAdapter implements CmsComponentAdapter {
...
+      const attributes = userId
+        ? {
+            urlParams: { userId },
+            queryParams,
+          }
+        : { queryParams };
+
+      return this.occEndpoints.buildUrl('components', attributes);
+    } else {
+      return this.occEndpoints.buildUrl('components', {
```
Matejk00: Also we can move this part out of else statement.

projects/core/src/occ/adapters/cms/occ-cms-component.adapter.ts
```diff
@@ -53,31 +69,78 @@ export class OccCmsComponentAdapter implements CmsComponentAdapter {
...
+      const attributes = userId
+        ? {
+            urlParams: { id, userId },
+            queryParams,
+          }
+        : { urlParams: { id }, queryParams };
+
+      return this.occEndpoints.buildUrl('component', attributes);
+    } else {
+      return this.occEndpoints.buildUrl('component', {
```
Matejk00: The same here I believe.

projects/core/src/occ/adapters/cms/occ-cms-component.adapter.ts
```diff
@@ -32,11 +36,23 @@ export class OccCmsComponentAdapter implements CmsComponentAdapter {
     id: string,
     pageContext: PageContext
   ): Observable<T> {
-    return this.http
-      .get<T>(this.getComponentEndPoint(id, pageContext), {
+    if (this.featureConfigService.isEnabled('newCmsEndpoint')) {
```
Matejk00: Shouldn't we first check if `this.featureConfigService` is not `undefined`?
I was rather expecting:
```
this.featureConfigService?.isEnabled(...)
```

projects/core/src/occ/adapters/cms/occ-cms-page.adapter.ts
```diff
@@ -44,16 +49,34 @@ export class OccCmsPageAdapter implements CmsPageAdapter {
    */
   load(pageContext: PageContext): Observable<CmsStructureModel> {
     const params = this.getPagesRequestParams(pageContext);
+    if (this.featureConfigService.isEnabled('newCmsEndpoint')) {
```
Matejk00: What do you think about assigning the feature name 'newCmsEndpoint' to a constant export value and relocating it to 'core/src/model'. This way, we can reuse the variable throughout the codebase instead of defining the string each time.

Also I am uncertain if the name 'newCmsEndpoint' is fitting here. Word 'new' is from the time-being perspective risky to use (right now it's new but we don't know the future). Maybe we can switch to:

Example:
```
export const USER_CMS_ENDPOINTS = 'userCmsEndpoints';
```

projects/core/src/occ/adapters/cms/occ-cms-component.adapter.ts
```diff
@@ -53,31 +69,78 @@ export class OccCmsComponentAdapter implements CmsComponentAdapter {
...
+            {
+              headers: this.headers,
+            }
+          );
+        }),
         map((componentList) => componentList.component ?? []),
         this.converter.pipeableMany(CMS_COMPONENT_NORMALIZER)
       );
+    } else {
+      return this.http
```
Matejk00: Also we can avoid nesting here.

projects/core/src/cms/cms.module.ts
```diff
@@ -5,11 +5,29 @@
...
+} from './config/default-cms-config';
 import { CmsService } from './facade/cms.service';
 import { PageMetaModule } from './page/page-meta.module';
 import { CmsStoreModule } from './store/cms-store.module';
+import { provideDefaultConfig, provideDefaultConfigFactory } from '../config';
+import { isFeatureEnabled } from '../features-config';
+
+const config = {
+  features: {
+    newCmsEndpoint: false,
```
Matejk00: I guess we don't need to define this config here, as by default it will be false, or undefined. And customers can enabled it from the app module perspective.

projects/core/src/cms/cms.module.ts
```diff
@@ -18,7 +36,11 @@ export class CmsModule {
   static forRoot(): ModuleWithProviders<CmsModule> {
     return {
       ngModule: CmsModule,
-      providers: [CmsService, provideDefaultConfig(defaultCmsModuleConfig)],
+      providers: [
+        CmsService,
+        provideDefaultConfig(config), // This overrides default config in AppModule - to be removed
```
Matejk00: So should it be removed now?

Matejk00 (changes requested):

projects/core/src/occ/adapters/cms/occ-cms-component.adapter.ts
kpawelczak: it's injected at the top, it will not be undefined

projects/core/src/occ/adapters/cms/occ-cms-page.adapter.ts
kpawelczak: sure, fixed

projects/core/src/occ/adapters/cms/occ-cms-component.adapter.ts
kpawelczak: fixed

projects/core/src/occ/adapters/cms/occ-cms-component.adapter.ts
kpawelczak: fixed

projects/core/src/occ/adapters/cms/occ-cms-component.adapter.ts
kpawelczak: fixed

projects/core/src/cms/cms.module.ts
kpawelczak: I've refactored the whole thing

projects/core/src/cms/cms.module.ts
kpawelczak: removed, went with different approach

projects/core/src/occ/adapters/cms/occ-cms-page.adapter.ts
kpawelczak: good point

projects/core/src/cms/cms.module.ts
```diff
@@ -18,7 +40,10 @@ export class CmsModule {
   static forRoot(): ModuleWithProviders<CmsModule> {
     return {
       ngModule: CmsModule,
-      providers: [CmsService, provideDefaultConfig(defaultCmsModuleConfig)],
+      providers: [
+        CmsService,
+        provideDefaultConfigFactory(getDefaultCmsConfig, [ConfigChunk]),
```
Matejk00, line 45: I think we can skip couple of if's and move a logic directly here. What do you think?

```
@NgModule({
  imports: [CmsStoreModule, PageMetaModule.forRoot()],
})
export class CmsModule {
  static forRoot(): ModuleWithProviders<CmsModule> {
    return {
      ngModule: CmsModule,
      providers: [
        CmsService,
        provideDefaultConfigFactory(
          (configChunks) => {
            configChunks.find((config: Config) => {
              if (Boolean(config.features?.[USER_CMS_ENDPOINTS])) {
                return defaultUserCmsModuleConfig;
              }
            });

            return defaultCmsModuleConfig;
          },
          [ConfigChunk]
        ),
      ],
    };
  }
}
```

projects/core/src/occ/adapters/cms/occ-cms-component.adapter.ts
```diff
@@ -32,11 +40,23 @@ export class OccCmsComponentAdapter implements CmsComponentAdapter {
...
+            this.getComponentEndPoint(id, pageContext, userId),
+            {
+              headers: this.headers,
+            }
+          );
+        }),
+        this.converter.pipeable<any, T>(CMS_COMPONENT_NORMALIZER)
+      );
+    } else {
+      return this.http.get<T>(this.getComponentEndPoint(id, pageContext), {
```
Matejk00: Let's move this return also outside else statement. Just for consistency.

Matejk00 (changes requested):

projects/core/src/occ/adapters/cms/occ-cms-component.adapter.ts
kpawelczak: missed it, fixed now

projects/core/src/cms/cms.module.ts
kpawelczak: this won't work and most of the time function names provide a brief explanation of what it does.

projects/core/src/occ/adapters/cms/occ-cms-component.adapter.ts
```diff
@@ -32,11 +40,23 @@ export class OccCmsComponentAdapter implements CmsComponentAdapter {
...
+            this.getComponentEndPoint(id, pageContext, userId),
+            {
+              headers: this.headers,
+            }
+          );
+        }),
+        this.converter.pipeable<any, T>(CMS_COMPONENT_NORMALIZER)
+      );
+    }
+    return this.http.get<T>(this.getComponentEndPoint(id, pageContext), {
```
Matejk00: Originally it looks like:

```
    return this.http
      .get<T>(this.getComponentEndPoint(id, pageContext), {
        headers: this.headers,
      })
      .pipe(this.converter.pipeable<any, T>(CMS_COMPONENT_NORMALIZER));
```

the pipe part is missing. I guess we can't delete an existing code.

Matejk00 (changes requested):

Matejk00 (approved):

PR #17886, opened 2023-09-28
Throw QuoteDetailsReloadQueryEvent before fetching quote details

Larisa-Staroverova:


feature-libs/quote/core/facade/quote.service.ts
```diff
@@ -479,7 +479,10 @@ export class QuoteService implements QuoteFacade {
       map(([isLoading, state]) => ({
         ...state,
         loading: state.loading || isLoading,
-      }))
+      })),
+      tap(() => {
```
ChristophHi, line 483: Please don't trigger refresh, this will lead to an endless loop...We somehow need to do this only on configuration update

ChristophHi (changes requested):

PR #17891, opened 2023-09-29
fix: OPF unit test coverage opf-checkout-billing-address-component-service

kpawelczak:


integration-libs/opf/checkout/components/opf-checkout-billing-address-form/opf-checkout-billing-address-form.service.ts
```diff
@@ -164,14 +164,14 @@ export class OpfCheckoutBillingAddressFormService {
     this.isSameAsDeliverySub.next(value);
   }
 
-  protected getDeliveryAddress(): Observable<Address | undefined> {
+  getDeliveryAddress(): Observable<Address | undefined> {
```
rmch91: Is the access modifiers were changed only for using this in spec? If yes, i don't think it's a good idea.
I think will be better to access this methods with `service['getDeliveryAddress']()` or to use `service as any` in spy if needed

integration-libs/opf/checkout/components/opf-checkout-billing-address-form/opf-checkout-billing-address-form.service.ts
```diff
@@ -164,14 +164,14 @@ export class OpfCheckoutBillingAddressFormService {
...
     return this.checkoutDeliveryAddressFacade.getDeliveryAddressState().pipe(
       filter((state) => !state.loading),
       map((state) => state.data)
     );
   }
 
-  protected getPaymentAddress(): Observable<Address | undefined> {
+  getPaymentAddress(): Observable<Address | undefined> {
```
rmch91: as above

rmch91 (changes requested):

integration-libs/opf/checkout/components/opf-checkout-billing-address-form/opf-checkout-billing-address-form.service.ts
kpawelczak: fixed

integration-libs/opf/checkout/components/opf-checkout-billing-address-form/opf-checkout-billing-address-form.service.ts
kpawelczak: fixed

rmch91 (approved):

PR #17893, opened 2023-09-29
fix: component with restriction is not displayed after changed in SmartEdit

FollowTheFlo:
CXSPA-3495
Add 2 parameters in query url when loading ProductList page.


feature-libs/smartedit/root/http-interceptors/cms-ticket.interceptor.ts
```diff
@@ -11,28 +11,65 @@ import {
...
 import { Observable } from 'rxjs';
+import { switchMap, take } from 'rxjs/operators';
 import { SmartEditLauncherService } from '../services/smart-edit-launcher.service';
 
 @Injectable({ providedIn: 'root' })
 export class CmsTicketInterceptor implements HttpInterceptor {
-  constructor(private service: SmartEditLauncherService) {}
+  constructor(
+    private service: SmartEditLauncherService,
+    protected routingService: RoutingService
```
rmch91: You can consider the usage of inject() to avoid breaking changes

FollowTheFlo: will do

rmch91 (approved):

PR #17894, opened 2023-09-30
fix: ApplicationRef not stable when asm-session-timer is displayed

otzelpov:
Every time the component `cx-asm-session-timer` is rendered, the app (angular-stable) becomes unstable (false). This is caused by the macro task `setInterval()` and can be fixed by running it outside of angular.

For example, QA can benefit from this if they use angular-stable to automate tests.


XC3S:
Thank you very much! I've run into the same problem.

scarai-sap:
wow, awesome changes!

Thanks for that. Does that means we can't use setInterval inside angular code(especially if the interval may change UI?)

otzelpov:
hi @scarai-sap, 

yes, and no. 

You should not use micro or macro tasks like setInterval during the initialization phase of the app or the bootstrapped components. This would result in "angular-stable" being false, because ngZone would be "busy" working on that task. 

But you can run it outside the ngZone via runOutsideAngular().

[See official docs](https://angular.io/errors/NG0506)

scarai-sap (approved):

PR #17895, opened 2023-09-30
CXINT:2317 - My Account->Order History & Details

anjana-bl:


feature-libs/order/components/order-details/order-details-extended/download-invoices/download-order-invoices-dialog.component.ts
```diff
@@ -0,0 +1,74 @@
...
+  LaunchDialogService,
+} from '@spartacus/storefront';
+import { Subscription } from 'rxjs';
+
+@Component({
+  selector: 'cx-download-order-invoices-dialog',
+  templateUrl: './download-order-invoices-dialog.component.html',
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class DownloadOrderInvoicesDialogComponent
```
Platonn: [MEDIUM]
Let's stick to the convention of synchronizing the component's: selector, class name and filename (and component module name).
Currently we have:
```
DownloadOrderInvoicesDialogComponent
cx-download-invoices-dialog
download-order-invoices-dialog.component.ts
```

Note: please review all the components in this PR and make sure the names are consistent

feature-libs/order/assets/translations/en/index.ts
```diff
@@ -4,8 +4,9 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { order } from './order.i18n';
+import { order, orderEnhancedUI } from './order.i18n';
 
 export const en = {
   order,
+  orderEnhancedUI,
```
Platonn, line 11: [MAJOR]
Consistent qualifier suffix for the new classes/translations/services/etc. - to be decided.
example Ideas:
OrderHistory**Customer360**Module
OrderHistory**V2**Module
OrderHistory**Extended**Module
OrderHistory**EnhancedUi**Module
OrderHistory**ExtendedView**Module

Currently we're the most fond of **Extended**. Let's wait for answer from stakeholders about the idea of **Customer360** - whether it's relevant at all.

feature-libs/order/README.md
```diff
@@ -2,4 +2,8 @@
 
 Spartacus Order feature can be added to the existing Spartacus application by running `ng add @spartacus/order`. For more information about Spartacus schematics, visit the [official Spartacus schematics documentation page](https://sap.github.io/spartacus-docs/schematics/).
 
+To see enhanced UI in Order History and Order Details page, 
+1. Add `{ provide: MYACCOUNT_ORDER_ENHANCED_UI, useValue: true },` to order-feature.module.ts file `providers` array.
+2. Add `pdf-invoices` feature to the storefront, if not installed already.
```
Platonn, line 7: [MAJOR]
Does it mean, that for the New-MyAccount feature, the "PDF invoices" feature library will becomes a mandatory library to be installed? 

If yes, then also our installation documentation for New-MyAccount (and it's schematics) should reflect that.

Question: What will happen for a customer that doesn't have "PDF invoices" library installed?

feature-libs/order/components/order-details/order-details-extended/consignment-tracking/consignment-tracking-link.component.ts
```diff
@@ -0,0 +1,77 @@
...
+export class ConsignmentTrackingLinkComponent implements OnInit, OnDestroy {
+  consignment: Consignment;
+  protected subscription = new Subscription();
+  protected orderConsignmentsService = inject(OrderConsignmentsService);
+  protected launchDialogService = inject(LaunchDialogService);
+  protected vcr = inject(ViewContainerRef);
+  protected cd = inject(ChangeDetectorRef);
+  protected outlet = inject(OutletContextData);
+
+  @ViewChild('element') element: ElementRef;
```
Platonn: [QUESTION]
I have a problem to find the declaration of `#element` identifier in the component's template. Could you help me understand how it works, please? Perhaps I'm missing something obvious.

feature-libs/order/components/order-details/order-details-extended/consignment-tracking/consignment-tracking-link.component.ts
```diff
@@ -0,0 +1,77 @@
...
+  templateUrl: './consignment-tracking-link.component.html',
+  host: { class: 'cx-list-header col-12' },
+})
+export class ConsignmentTrackingLinkComponent implements OnInit, OnDestroy {
+  consignment: Consignment;
+  protected subscription = new Subscription();
+  protected orderConsignmentsService = inject(OrderConsignmentsService);
+  protected launchDialogService = inject(LaunchDialogService);
+  protected vcr = inject(ViewContainerRef);
+  protected cd = inject(ChangeDetectorRef);
```
Platonn: [NITPICK]
Let's stick to convention of naming `cdr`, instead of `cd`

feature-libs/order/components/order-details/order-details-extended/consignment-tracking/consignment-tracking-link.component.ts
```diff
@@ -0,0 +1,77 @@
...
+} from '@spartacus/storefront';
+import { of, Subscription } from 'rxjs';
+import { take } from 'rxjs/operators';
+import { Consignment } from '@spartacus/order/root';
+import { OrderConsignmentsService } from '../../order-consignments.service';
+
+@Component({
+  selector: 'cx-consignment-tracking-link',
+  templateUrl: './consignment-tracking-link.component.html',
+  host: { class: 'cx-list-header col-12' },
```
Platonn: [QUESTION]
Did you consider putting `.col-12` explicitly on the host element, in the parent component's template? What pros/cons of alternative solutions do you see?

Why I ask: Now this component cannot be re-used outside the Bootstrap grid. So it would be good if we have other good arguments to justify the current solution.

PS. I have similar concern for the `.cx-list-header` class, but maybe arguments for justifying it are a bit different than `.col-12`. I'm happy to hear your opinion and arguments.

feature-libs/order/components/order-details/order-details-extended/download-invoices/download-order-invoices-dialog.component.ts
```diff
@@ -0,0 +1,74 @@
...
+  ngOnInit(): void {
+    this.subscription.add(
+      this.launchDialogService.data$.subscribe((data: Order) => {
+        this.order = data;
+      })
+    );
+  }
+
+  ngAfterViewChecked() {
+    this.cdr.detectChanges();
```
Platonn, line 47: [MEDIUM]
Calling `detectChanges()` in `ngAfterViewChecked()` at the first sight it looks like a workaround for problems with change detection, that might have roots somewhere else. I'd be happy to understand those problems and hopefully try to fix them in their roots

feature-libs/order/components/order-details/order-details-extended/download-invoices/download-order-invoices-dialog.component.ts
```diff
@@ -0,0 +1,74 @@
...
+  readonly OrderOutlets = OrderOutlets;
+  invoiceCount: number | undefined = undefined;
+  iconTypes = ICON_TYPE;
+  focusConfig: FocusConfig = {
+    trap: true,
+    block: true,
+    autofocus: true,
+    focusOnEscape: true,
+  };
+  order: Order;
```
Platonn: [MEDIUM]
I have problem to find usages of the `order` property in this component's TS and html. Is this property _really_ needed?

.env-cmdrc
```diff
@@ -59,5 +59,9 @@
   "pdf-invoices": {
     "CX_BASE_URL": "https://api.cg79x9wuu9-eccommerc1-s5-public.model-t.myhybris.cloud",
     "CX_PDF_INVOICES": "true"
+  },
+  "myaccount-order-enhanced-ui":{
+    "CX_BASE_URL": "https://api.cg79x9wuu9-eccommerc1-s9-public.model-t.myhybris.cloud",
```
Platonn: [MEDIUM]
I believe that before merging to `develop` branch, we should first merge and deploy the required sample data / new backend version to standard Dev and CI servers used. And then we won't need this new entry in the env-cmdrc. Would you agree?

feature-libs/order/root/tokens/context.ts
```diff
@@ -13,3 +13,8 @@ export const OrderDetailsOrderEntriesContextToken = new InjectionToken(
 export const OrderConfirmationOrderEntriesContextToken = new InjectionToken(
   'OrderConfirmationOrderEntriesContext'
 );
+
+export const MYACCOUNT_ORDER_ENHANCED_UI = new InjectionToken<boolean>(
```
Platonn: [MEDIUM]
Of course the final naming of the suffix qualifier (e.g. `Extended` vs `EnhancedUi` vs .....) is to be decided. But regardless of the decision, IMHO the name of the feature should go after the full qualifier:
`MYACCOUNT_ORDER_ENHANCED_UI` vs `MYACCOUNT_ENHANCED_UI_ORDER`.
Thanks to that we will have a common prefix and only the ending will change among other features, e.g.:
```
MYACCOUNT_ENHANCED_UI_ORDER
MYACCOUNT_ENHANCED_UI_CHANGE_PASSWORD
MYACCOUNT_ENHANCED_UI_USER_PROFILE
...
```

feature-libs/order/components/order-details/order-details-extended/download-invoices/download-order-invoices-dialog.component.html
```diff
@@ -0,0 +1,32 @@
+<div
+  [cxFocus]="focusConfig"
+  (esc)="close('Escape clicked')"
+  class="cx-modal-container"
+>
+  <div class="cx-modal-content">
+    <ng-container>
```
Platonn: [NITPICK]
Do we _really_ need this `<ng-container>` ?

feature-libs/order/components/order-details/order-details.module.ts
```diff
@@ -114,6 +160,19 @@ const moduleComponents = [
...
+      inject(MYACCOUNT_ORDER_ENHANCED_UI) ? enhancedUICmsMapping : {}
+    ),
+    {
+      provide: MODULE_INITIALIZER,
+      useFactory: registerOrderOutletFactory,
+      deps: [
+        MYACCOUNT_ORDER_ENHANCED_UI,
+        OutletService,
+        ComponentFactoryResolver,
+      ],
```
Platonn: [MEDIUM]
Declaring `deps` array is no longer recommended, because we can use `inject()` function inside the factory function instead. Thanks to that, later on adding/removing dependencies of the factory won't require breaking changes.
```suggestion
```
and then change the factory function
```diff
- function registerOrderOutletFactory(
-   token: boolean,
-   outletService: OutletService,
-   componentFactoryResolver: ComponentFactoryResolver
- ): () => void {

+ function registerOrderOutletFactory(): () => void {
+   const token = inject(MYACCOUNT_ORDER_ENHANCED_UI);
+   const outletService = inject(OutletService);
+   const componentFactoryResolver = inject(ComponentFactoryResolver);
+ 
  return () => {
    /*...*/
  };
}
```
Note: `inject()` function should be invoked at the top level of the factory function's body. In particular it shouldn't be invoked in a nested function - becasue then it won't work, according to Angular's constraints. For more, see the official docs for [Angular inject function](https://angular.io/api/core/inject)

---

Note 2: Similar remark applies to other factory functions in this PR - to get rid of the `deps` param. E.g. in. `feature-libs/order/occ/order-occ.module.ts`

feature-libs/order/components/order-details/order-details.module.ts
```diff
@@ -35,9 +44,43 @@ import { OrderDetailItemsComponent } from './order-detail-items/order-detail-ite
...
+      outletService.add(config.id, template, config.position);
+    }
+  };
+}
+
+const enhancedUICmsMapping: CmsConfig = {
+  cmsComponents: {
+    AccountOrderDetailsActionsComponent: {
+      component: OrderDetailsActionsExtendedComponent,
+      guards: [AuthGuard],
```
Platonn, line 78: [QUESTION]
Is declaring guards here _really_ needed? I guess the cms-guards declaration is "inherited" from the original component's cms mapping ("inherited" - in terms of deep-merging all the config chunks together).

Note: same question about (re-)declaring guards, applies also to other files in this PR

feature-libs/order/components/order-details/order-consignments.service.ts
```diff
@@ -0,0 +1,86 @@
...
+@Injectable({
+  providedIn: 'root',
+})
+export class OrderConsignmentsService {
+  consignmentStatus: string[] = [
+    'SHIPPED',
+    'IN_TRANSIT',
+    'DELIVERY_COMPLETED',
+    'DELIVERY_REJECTED',
+    'DELIVERING',
```
Platonn: [MINOR]
What about moving those values to an enum?

feature-libs/order/components/order-history/order-history-extended/consignment-entries/consignment-entries.component.ts
```diff
@@ -0,0 +1,37 @@
...
+  templateUrl: './consignment-entries.component.html',
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class ConsignmentEntriesComponent {
+  @Input()
+  consignments?: Consignment[];
+  @Input()
+  orderCode?: string;
+
+  consignmentEntriesLength(consignment: Consignment): number {
```
Platonn: [MEDIUM]
For consistency, let's stick to a convention of starting names of methods with verbs, e.g. `getConsignmentEntriesLength()`

Note: same remark applies to other places in this PR. Please review all the methods names in this PR.

feature-libs/order/components/order-history/order-history-extended/consignment-entries/consignment-entries.component.ts
```diff
@@ -0,0 +1,37 @@
...
+    if (consignment.entries) {
+      return consignment.entries.length;
+    }
+    return 0;
+  }
+
+  consignmentNumber(code?: string) {
```
Platonn: [MEIDUM]
Let's add an explicit return type to the method.

feature-libs/order/components/order-history/order-history-extended/consolidated-information/order-consolidated-information.component.ts
```diff
@@ -0,0 +1,105 @@
...
+@Component({
+  selector: 'cx-order-consolidated-information',
+  templateUrl: './order-consolidated-information.component.html',
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class OrderConsolidatedInformationComponent {
+  protected orderConsignmentsService = inject(OrderConsignmentsService);
+  @Input()
+  order?: OrderHistory;
+  private IMAGE_COUNT = 4; //showing fixed no.of images, without using carousel
```
Platonn: [QUESTION]
Why this property is `private`? Did you consider making it `protected`? What pros/cons of both solutions do you see?

feature-libs/order/components/order-history/order-history-extended/consolidated-information/order-consolidated-information.component.ts
```diff
@@ -0,0 +1,105 @@
...
+
+  orderEntriesCount(orderEntries: OrderEntry[] | undefined): number {
+    if (orderEntries) {
+      return orderEntries.length;
+    }
+    return 0;
+  }
+
+  isStatusCritical(status: string): boolean {
+    const criticalStatus = ['cancelled', 'error', 'Error', 'rejected'];
```
Platonn: [MINOR]
Did you consider moving those hardcoded values to an enum

feature-libs/order/components/order-history/order-history-extended/order-history-extended.component.html
```diff
@@ -0,0 +1,108 @@
...
+
+      <!-- PAGINATION -->
+      <div
+        *ngIf="orderHistory.pagination.totalPages > 1"
+        class="cx-order-history-pagination"
+        [attr.aria-label]="
+          'orderHistoryEnhancedUI.orderListPagination' | cxTranslate
+        "
+      >
+        <!-- !!!!!!!re-check this above div -->
```
Platonn: [MEDIUM]
Is this a sign of some uncompleted work or testing?

feature-libs/order/occ/adapters/occ-order-history-extended.adapter.ts
```diff
@@ -0,0 +1,162 @@
...
+
+@Injectable()
+export class OccOrderHistoryExtendedAdapter extends OccOrderHistoryAdapter {
+  /**
+   * fills the tracking information for each consignment in the input order details
+   * @param userId user id
+   * @param order order details with consignments
+   * @returns order details along with tracking information for each of it's consignments
+   */
+  fillConsignmentTrackingDetails(
```
Platonn: [MEDIUM]
Can we make this method `protected`? I can see it isn't used anywhere outside this class.

feature-libs/order/occ/adapters/occ-order-history-extended.adapter.ts
```diff
@@ -0,0 +1,162 @@
...
+    userId: string,
+    order: Order
+  ): Observable<Order> {
+    const requests = order.consignments?.map((consignment: Consignment) => {
+      if (consignment.code && consignment.trackingID) {
+        return this.getConsignmentTracking(
+          order?.code ?? '',
```
Platonn: [MINOR]
I'm not fond of tricking type safety by fallbacking to a dummy value (e.g. `''`) just for the code to compile, but won't work correctly in runtime. Instead I suggest to first check: `if(!order?.code) { throw new Error(<short meaningful error message here)> }`

feature-libs/order/components/order-details/order-details-extended/consignment-tracking/consignment-tracking-link.component.ts
```diff
@@ -0,0 +1,77 @@
...
+  @ViewChild('element') element: ElementRef;
+  consignmentStatus: string[] = this.orderConsignmentsService.consignmentStatus;
+  ngOnInit(): void {
+    this.subscription.add(
+      this.outlet?.context$.subscribe((context: { item?: Consignment }) => {
+        if (context.item !== undefined) {
+          this.consignment = context.item;
+        }
+        this.cd.markForCheck();
+      })
```
Platonn: [MEDIUM]
I don't recommend mixing observable programming with imperative programming (e.g. subscribing to observables and copying values to the component's static properties), unless _really_ justified.

It's prone to issues with synchronization of data and the change detection issues (and often requires calling `cd.markForCheck()` manually). Moreover it requires manually creating and destroying a subscription.

Instead, I'd suggest just using the observables as far as possible (e.g. pass observable to in the template and unwrap the value there with `async` pipe)

In this example, I suggest to simply create an observable that derives from the `outlet.context$` observable via `map()` operator:
```ts
this.consignment$ = this.outlet?.context$.pipe(
  map(context => context.item)
);
```
then unwrap this observable in the html template:
```html
<span *ngIf="consignment$ | async as consignment">
   ...
   <button (click)="openTrackingDialog(consignment)> <!-- NOTE: passing object to method ! -->
```
then add parameter to the TS `openTrackingDialog()` method:
```ts
  openTrackingDialog(consignment: WhateverTypeAppropriate_Perhaps_Consignment) {
    ...
  }
```

--- 

Note: the analogical remark applies also to other files in this PR. Please review other subscriptions in components in this PR, and check if we could simply create a derived observable instead of copying values from the source observable to a static variable.

feature-libs/order/components/order-details/order-details-extended/consignment-tracking/consignment-tracking-link.component.html
```diff
@@ -0,0 +1,28 @@
+<div class="cx-list-status">
+  <span>
+    <span *ngIf="consignment">
+      {{ 'orderDetails.deliveryStatus_' + consignment.status | cxTranslate }}
+    </span>
+    <span
+      *ngIf="consignmentStatus.includes(consignment.status ?? '')"
```
Platonn, line 7: [MINOR]
Instead of having logic in the HTML template, what about exposing a method for it (that is testable with unit tests), e.g. `isConsignmentStatusValid()` - it's just an example name, not sure if appropriate in this case. Maybe we could even move this method to the `OrderConsignmentService` helper service, instead of exposing the `consignmentStatus` public property

feature-libs/order/components/order-details/order-consignments.service.ts
```diff
@@ -0,0 +1,86 @@
...
+  ];
```
Platonn: [MEDIUM]
Do we really have to expose this raw array as a public field of the service? If we do it as is, then any consumer of it can potentially:
1. replace this field value with any custom array
2. mutate the array (push new values or remove some)

I believe the array should be made `readonly` and wrapped with `Object.freeze()` so it's not mutable
```suggestion
  readonly consignmentStatuses: ReadonlyArray<string> = Object.freeze([
    'SHIPPED',
    'IN_TRANSIT',
    'DELIVERY_COMPLETED',
    'DELIVERY_REJECTED',
    'DELIVERING',
  ]);
```
Explanation:
- `readonly property` modifier prevents changing the property value to another array
- `Object.freeze()` prevents mutating the array later in runtime - it will throw exception in such a case. for more see [Docs of Object.freeze on arrays](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/freeze?retiredLocale=pl#freezing_arrays)

All that said, maybe we can avoid exposing this array as `public` field at all, e.g. by exposing an util method instead, like `isStatusValid()` as proposed in the [another comment in this PR.](https://github.com/SAP/spartacus/pull/17895/files#r1350051164)

---

Side note: by the way, suggesting plural form `Statuses`, becasue it's an array of statuses, not a single status.

feature-libs/order/root/model/order.model.ts
```diff
@@ -79,6 +80,8 @@ export interface Consignment {
   status?: string;
   statusDate?: Date;
   trackingID?: string;
+  //myaccount-enhanced-ui related
+  consignmentTracking?: ConsignmentTracking;
```
Platonn, line 84: [MAJOR]
As discussed offline, let's create a separate View Model for adding `consignmentTracking`.

Platonn: [MAJOR]
As discussed offline, let's create a separate View Model for adding `consignmentTracking`.

feature-libs/order/occ/order-occ.module.ts
```diff
@@ -79,6 +85,30 @@ import { defaultOccOrderConfig } from './config/default-occ-order-config';
...
+        } else {
+          return new OccOrderHistoryAdapter(
+            httpClient,
+            occEndpointsService,
+            converterService
+          );
+        }
+      },
+      deps: [HttpClient, OccEndpointsService, ConverterService],
+    },
```
Platonn: [MAJOR]
It's not recommended to instantiate classes manually with `new`, but instead we should rely on the Angular Dependency injection, as much as possible.

Did you consider simply injecting the original instances conditionally?
```suggestion
    OccOrderHistoryAdapter,
    OccOrderHistoryExtendedAdapter,
    {
      provide: OrderHistoryAdapter,
      useFactory: () => {
        const enhancedUI = inject(MYACCOUNT_ORDER_ENHANCED_UI);
        return enhancedUI
          ? inject(OccOrderHistoryExtendedAdapter)
          : inject(OccOrderHistoryAdapter);
      },
    },
```

anjana-bl: done

feature-libs/order/components/order-details/order-details-extended/download-invoices/download-order-invoices-dialog.component.ts
anjana-bl: done

feature-libs/order/components/order-details/order-details-extended/consignment-tracking/consignment-tracking-link.component.ts
anjana-bl: Created a dummy property to just invoke launchDialogService.openDialog. But since it is an optional parameter, I have now replaced this with undefined.

feature-libs/order/components/order-details/order-details-extended/consignment-tracking/consignment-tracking-link.component.ts
anjana-bl: removed this

feature-libs/order/components/order-details/order-details-extended/consignment-tracking/consignment-tracking-link.component.ts
anjana-bl: this is an implementation of outlet. so i suppose we cannot put it in parent component's template... how about using  @HostBinding('className') ?

feature-libs/order/components/order-details/order-details-extended/download-invoices/download-order-invoices-dialog.component.ts
anjana-bl: not needed. have removed it.

feature-libs/order/components/order-details/order-details-extended/download-invoices/download-order-invoices-dialog.component.html
anjana-bl: nope, removed it

feature-libs/order/components/order-details/order-details.module.ts
anjana-bl: done

feature-libs/order/components/order-history/order-history-extended/consignment-entries/consignment-entries.component.ts
anjana-bl: removing this method, as its not used

feature-libs/order/components/order-history/order-history-extended/consignment-entries/consignment-entries.component.ts
anjana-bl: done

feature-libs/order/components/order-details/order-details-extended/consignment-tracking/consignment-tracking-link.component.html
anjana-bl: done

feature-libs/order/components/order-history/order-history-extended/order-history-extended.component.html
anjana-bl: its some comment written during testing, have removed it now

feature-libs/order/occ/adapters/occ-order-history-extended.adapter.ts
anjana-bl: done

feature-libs/order/components/order-details/order-details-extended/consignment-tracking/consignment-tracking-link.component.ts
anjana-bl: done

feature-libs/order/components/order-details/order-details-extended/consignment-tracking/consignment-tracking-link.component.html
anjana-bl: done

feature-libs/order/components/order-details/order-consignments.service.ts
anjana-bl: done

feature-libs/order/README.md
anjana-bl: Yes, PDF Invoices will become a mandatory library. But if its not installed, it will not break the UI. When the user clicks on 'Download Invoices' button in Order Details screen, it will show spinner and no data will load there. If user wants to see data inside that dialog, PDF Invoices should be installed. Since this new UI is controlled based on token when user can set only after installing the storefront, i have not altered the schematics to download PDF Invoices as well. I thought we can document that- if user wants to use the new UI for orders, then add PDF Invoices feature to your storefront.

feature-libs/order/components/order-details/order-details-extended/download-invoices/download-order-invoices-dialog.component.ts
anjana-bl: without this 'this.cdr.detectChanges();',     
if (this.invoiceComponent.pagination !== undefined) {
      this.invoiceCount = this.invoiceComponent.pagination.totalResults;
    }
is not getting detected.And shows just spinner and no message

feature-libs/order/components/order-history/order-history-extended/consolidated-information/order-consolidated-information.component.ts
anjana-bl: done

feature-libs/order/root/model/order.model.ts
anjana-bl: done

feature-libs/order/README.md
Platonn: Yes. let's document it. 
If/When new MyAccount will become the default one, our default installation schematics might need to install PDF library. Docs will help to remind us in the feature that we need to do it. At least there will be a written track about it.

feature-libs/order/components/order-details/order-details-extended/consignment-tracking/consignment-tracking-link.component.ts
Platonn: Ah, now understood. Because it's outlet component, there is no way to define css class in the parent component.

Because it's not obvious on the first sight, would you mind please leaveing a code comment near the `host: {....}  // ...` configuration, explaining why we have to set those css classes here (becasue it's an outlet component)?

feature-libs/order/README.md
```diff
@@ -2,7 +2,7 @@
 
 Spartacus Order feature can be added to the existing Spartacus application by running `ng add @spartacus/order`. For more information about Spartacus schematics, visit the [official Spartacus schematics documentation page](https://sap.github.io/spartacus-docs/schematics/).
 
-To see enhanced UI in Order History and Order Details page, 
+To see enhanced UI in Order History and Order Details page in B2C storefront,
```
Platonn, line 5: [QUESTION]
Won't our New MyAccount feature work in B2B storefronts? If yes, what are constraints that make it impossible?

feature-libs/order/components/order-details/order-details-extended/consignment-tracking/consignment-tracking-link.component.ts
```diff
@@ -4,74 +4,57 @@
...
     const modalInstanceData = {
       tracking$: of(tracking),
-      shipDate: this.consignment.statusDate,
+      shipDate: consignment.statusDate,
     };
 
     const dialog = this.launchDialogService.openDialog(
       LAUNCH_CALLER.CONSIGNMENT_TRACKING,
-      this.element,
-      this.vcr,
```
Platonn: [QUESTION]
Why don't we pass the element and VCR anymore? Any disadvantaves of passing them?
Are there not any advantages of passing them? e.g. keeping the dialog as a child in the change-detection tree?

feature-libs/order/core/facade/order-history-extended-view.service.ts
```diff
@@ -0,0 +1,260 @@
...
+            });
+          } else {
+            return of(orderHistory);
+          }
+        }
+      )
+    );
+  }
+
+  private getOrderDetailsValue(code: string): Observable<Order> {
```
Platonn: [MINOR]
Why lots of methods in this service are `private` but not `protected`? For extensibility for customers it's better to have `protected` but maybe you have some arguments of keeping them private?

feature-libs/order/components/order-details/order-details-extended/consignment-tracking/consignment-tracking-link.component.ts
```diff
@@ -5,56 +5,28 @@
...
-  }
-  isStatusValid(status: string): boolean {
-    return this.orderConsignmentsService.isStatusValid(status);
+    this.outlet?.context$.subscribe((context: ConsignmentOutletContextData) => {
+      this.consignment = context.item;
+    });
+    this.outlet?.context$.subscribe((context: ConsignmentOutletContextData) => {
+      this.orderCode = context?.order?.code ?? '';
+    });
+    super.ngOnInit();
```
Platonn, line 30: [MEDIUM]
Why we got back to copying observable values to static properties?

If it's only to be able to inherit from the class `ConsignmentTrackingComponent`... is inheriting from this class bringing really enough value to justify the mixed approach (observables and static properties)? I guess we're not reusing from the parent component that much.

feature-libs/order/components/order-history/order-history-extended/order-history-extended.component.ts
```diff
@@ -4,11 +4,36 @@
...
+    .getOrderHistoryList(this.ITEMS_PER_PAGE)
+    .pipe(
+      tap((orders: OrderHistoryListView | undefined) => {
+        this.isLoaded$.next(true);
+        if (orders?.pagination?.sort) {
+          this.sortType = orders.pagination.sort;
+        }
+        this.hasPONumber =
+          orders?.orders?.[0]?.purchaseOrderNumber !== undefined;
+      })
```
Platonn: [MINOR]
This part is copy-pasted from the super component. Did you consider exposing it as an reusable protected method that could be used here?

feature-libs/order/core/facade/order-history-extended-view.service.ts
```diff
@@ -0,0 +1,260 @@
...
+  getConsignmentTrackingById,
+  getConsignmentTrackingByIdState,
+} from '../store/selectors/order-group.selectors';
+import { OrderHistoryService } from './order-history.service';
+import { OrderReturnRequestService } from './order-return-request.service';
+
+@Injectable()
+export class OrderHistoryExtendedViewService {
+  protected orderReturnRequestService = inject(OrderReturnRequestService);
+  protected orderHistoryConnector = inject(OrderHistoryConnector);
```
Platonn: [MINOR]
I have problem to find usages of this property. Anyway, I believe we agreed to use ngrx (not connector) in service.

feature-libs/order/core/facade/order-history-extended-view.service.ts
```diff
@@ -0,0 +1,260 @@
...
+  protected store = inject(Store);
+  protected userIdService = inject(UserIdService);
+  protected orderHitsoryService = inject(OrderHistoryService);
```
Platonn: [MEDIUM]
Typo. hitsory -> history

feature-libs/order/core/facade/order-history-extended-view.service.ts
```diff
@@ -0,0 +1,260 @@
...
+
+  clearOrderList(): void {
+    this.orderHitsoryService.clearOrderList();
+  }
+  getOrderDetailsWithTracking(
```
Platonn, line 46: [MINOR]
For clarity of purpose, this method can be made `protected` as it's not used in any other class.

feature-libs/order/core/facade/order-history-extended-view.service.ts
```diff
@@ -0,0 +1,260 @@
...
+        (
+          responses: [
+            OrderHistoryListView | undefined,
+            ReturnRequestList | undefined
+          ]
+        ) => {
+          const returnRequests = responses?.[1]?.returnRequests;
+          const orderHistory = responses?.[0];
+          if (returnRequests && orderHistory?.orders) {
+            console.log(returnRequests);
```
Platonn: [MINOR]
Let's remove debugging console.logs from the PR.

Please review the whole PR and remove other `console` usages, if present.

feature-libs/order/core/facade/order-history-extended-view.service.ts
```diff
@@ -0,0 +1,260 @@
...
+        if (requests !== undefined) {
+          return requests[0];
+        } else {
+          return {};
+        }
+      })
+    );
+  }
+
+  getOrderHistoryList(
```
Platonn, line 133: [MEDIUM]
Method names in this service sometimes say `order` and sometimes `orderHistory`. Let's make them consistent. e.g:
`getOrderHistoryList` - history
`getOrderListWithDetails` - no history

I don't mind dropping the word `history` or keeping it. Just let's make it consistent

feature-libs/order/core/facade/order-history-extended-view.service.ts
```diff
@@ -0,0 +1,266 @@
...
+          new OrderActions.LoadConsignmentTrackingByID({
+            orderCode,
+            consignmentCode,
+            userId,
+          })
+        ),
+    });
+  }
+
+  getConsignmentTracking(
```
Platonn, line 247: [MEDIUM]
Let's make this method `protected`. it's not used anywhere else

feature-libs/order/core/facade/order-return-request.service.ts
```diff
@@ -61,7 +61,7 @@ export class OrderReturnRequestService implements OrderReturnRequestFacade {
    * Gets order return request list
    */
   getOrderReturnRequestList(
-    pageSize: number
+    pageSize?: number
```
Platonn, line 64: [MEDIUM]
When we talked offline in Slack about this change, I assumed we'll provide some fallback value/behaviour in case when the argument is not present.

But here we're just passing `undefined` down the methods chain. Shouldn't we provide some default value on some level? Maybe I'm missing some important detail... please help me understand

feature-libs/order/core/store/actions/consignment-tracking-by-id.action.ts
```diff
@@ -0,0 +1,66 @@
...
+  constructor(
+    public payload: {
+      orderCode: string;
+      consignmentCode: string;
+      userId: string;
+    }
+  ) {
+    super(
+      CONSIGNMENT_TRACKING_BY_ID_ENTITIES,
+      `${payload.orderCode},${payload.consignmentCode}`
```
Platonn: [NITPICK]
Can we extract this "contatenation logic" to a separate function? I can see it's repeated in other places, but the shared function could be used there instead.

feature-libs/order/core/store/effects/order-by-id.effect.ts
```diff
@@ -0,0 +1,44 @@
...
+import { Actions, createEffect, ofType } from '@ngrx/effects';
+import { normalizeHttpError } from '@spartacus/core';
+import { Order } from '@spartacus/order/root';
+import { Observable, of } from 'rxjs';
+import { catchError, concatMap, map } from 'rxjs/operators';
+import { OrderHistoryConnector } from '../../connectors/order-history.connector';
+import { OrderActions } from '../actions/index';
+
+@Injectable()
+export class OrderByIDEffect {
```
Platonn: [MINOR]
For consistency, let's use camelCase for `Id` (instead of `ID`).

This remark applies also to other places in this PR

feature-libs/order/README.md
anjana-bl: it works for B2B as well but business wants to advise customers to use it for B2C alone, and from March, we can inform B2B to use as well

feature-libs/order/components/order-details/order-details-extended/consignment-tracking/consignment-tracking-link.component.ts
anjana-bl: since we are extending already existing component , it uses static variable, The parent component holds the logic to show dialog.

feature-libs/order/core/store/effects/order-by-id.effect.ts
anjana-bl: done

feature-libs/order/core/store/actions/consignment-tracking-by-id.action.ts
anjana-bl: done

feature-libs/order/core/facade/order-history-extended-view.service.ts
anjana-bl: done

feature-libs/order/core/facade/order-history-extended-view.service.ts
anjana-bl: done

feature-libs/order/core/facade/order-history-extended-view.service.ts
anjana-bl: shall we leave this method public? as this is like getter method for the new ngrx entity?

feature-libs/order/core/facade/order-history-extended-view.service.ts
anjana-bl: done

feature-libs/order/core/facade/order-history-extended-view.service.ts
anjana-bl: done

feature-libs/order/components/order-history/order-history-extended/order-history-extended.component.ts
anjana-bl: done

feature-libs/order/core/facade/order-history-extended-view.service.ts
anjana-bl: done

feature-libs/order/core/facade/order-return-request.service.ts
anjana-bl: I have removed this so that the whole return request list of current user is retrieved, and the orders of the current page will be looped to fetch return requests of that current order. Discussed this approach with @vigneshkrish8691  . Since there was no other API available to get return requests based on order code, proceeding with this approach

feature-libs/order/core/facade/order-history-extended-view.service.ts
anjana-bl: shall we leave this method public? as this is like getter method for the new ngrx entity?

feature-libs/order/components/order-history/order-history-extended/consolidated-information/order-consolidated-information.component.ts
anjana-bl: done

feature-libs/order/root/tokens/context.ts
anjana-bl: done

feature-libs/order/components/order-details/order-details-extended/download-invoices/download-order-invoices-dialog.component.ts
Platonn: I'd love to investigate it further and find the root causes of the problems with change detection and fix it in the roots. But due to tight schedule of release, we might have to postpone it.

feature-libs/order/assets/translations/en/index.ts
Platonn: new proposition,using an unique code name:
OrderHistory*EnhancedMyAccount*Module

.env-cmdrc
anjana-bl: Hi Kris, actually this new development needs to be tested with backend CDP commerce integration extension (from business perspective). So I believe its better to have a separate commerce system and not use standard Dev and CI servers.

feature-libs/order/occ/adapters/occ-order-history-extended.adapter.ts
anjana-bl: the 'code' for any order is the key to identify that order uniquely. There is no way that can be empty functionally. hence I added ?? to just satisfy the checks, and did not throw any errors messages

feature-libs/order/components/order-details/order-details.module.ts
anjana-bl: yes, guards are inherited. will do the change

feature-libs/order/assets/translations/en/index.ts
anjana-bl: will change all names with prefix 'MyAccountV2' as discussed

anjana-bl:
THIS DEVELOPMENT IS CONTINUED IN PR https://github.com/SAP/spartacus/pull/17976

Platonn:
closing in favor of https://github.com/SAP/spartacus/pull/17976

PR #17898, opened 2023-10-02
fix: Prevent B2B module installation when adding CDC to a B2C storefront

anjana-bl:
Issue Description: When user tries to add CDC feature to their existing B2C feature, it downloads B2B- Organization Administration module as well.


projects/schematics/src/shared/lib-configs/integration-libs/cdc-schematics-config.ts
```diff
@@ -119,6 +120,54 @@ export const CDC_SCHEMATICS_CONFIG: SchematicConfig = {
...
+    {
+      markerModuleName: USER_ACCOUNT_MODULE,
+      featureModuleName: CDC_USER_ACCOUNT_MODULE,
+    },
+    {
+      markerModuleName: USER_PROFILE_MODULE,
+      featureModuleName: CDC_USER_PROFILE_MODULE,
+    },
+  ],
+};
```
Platonn: [QUESTION]
Is that true that everything that is related to B2C is also included in B2B? In other words, I is the CDC B2B a superset of features in CDC B2C?

I yes, then copying the whole config to a separate variable (while a lot of config parts are shared) is prone to human errors in the future - we might forget to add new CDC features to both CDC and CDC-B2B Configs in the future, but add it just in B2C or just in B2B.

What do you think about defining a shared cofing both for B2B and B2C and reuse it in both configs? 
OR just reusing parts of the B2C config in the B2B config? Thanks to that, any addition to B2C will be also reflected in B2B. And if you want to add something only for B2B, then we can add it just to B2B config.
What do you think?

integration-libs/cdc/schematics/add-cdc/index_spec.ts
```diff
@@ -121,19 +112,115 @@ describe('Spartacus CDC schematics: ng-add', () => {
...
+            continue;
+          }
+          // TODO: after 4.0: use this test, as we'll have synced versions between lib's and root package.json
+          // const expectedVersion = (peerDependencies as Record<
+          //   string,
+          //   string
+          // >)[toAdd];
+          const expectedDependency = dependencies[toAdd];
+          expect(expectedDependency).toBeTruthy();
+          // expect(expectedDependency).toEqual(expectedVersion);
```
Platonn: [MINOR]
I'm aware your bug fix is urgent and I don't want to block this PR because of this copy-pasted (outdated) TODO.

But do you think that afterwards (when it's merged) you could take a look at it and try to fulfill the recommendation of this TODO and the remove the TODO comment (in a separate PR), so we clean up the code but not pollute it with TODOs even more? I hope it's doable, as the comment says "after 4.0" and we're now "6.x" 😉 

Btw. SonarQube in CI likely won't let you merge this PR, becasue it contains //TODO comment.
I suggest to create a ticket for it and put `// CXSPA-xxxx:`  (with the number of the ticket you create) instead of `//TODO: ` to workaround the SonarQube.

projects/schematics/src/shared/lib-configs/integration-libs/cdc-schematics-config.ts
anjana-bl: done

integration-libs/cdc/schematics/add-cdc/index_spec.ts
anjana-bl: done - have created [CXSPA-4872](https://jira.tools.sap/browse/CXSPA-4872)

Platonn (approved):

PR #17900, opened 2023-10-02
feat: ASM C360 Address comments

kimhw0630:
Closes: https://jira.tools.sap/browse/CXSPA-4774

addressed comments and some clean up.


feature-libs/asm/customer-360/components/customer-360/customer-360.component.html
```diff
@@ -162,24 +162,7 @@ <h3 id="asm-customer-360-title" class="modal-title">
...
-            (click)="selectTab(i)"
-            [class.active]="activeTab === i"
-            (keydown)="switchTab($event, i)"
-            *ngIf="tabHeader.i18nNameKey !== 'customer360.promotionsTab'"
-          >
-            {{ tabHeader.i18nNameKey | cxTranslate }}
-          </button>
-        </ng-container>
-      </div>
-      <div class="cx-tab-headers" role="tablist" *cxFeatureLevel="'6.5'">
```
Platonn, line 182: [QUESTION]

Can you help me understand why in this PR we're removing the "branch" of code wrapped within `*cxFeatureLevel`?

feature-libs/asm/customer-360/components/sections/customer-360-map/customer-360-map.component.html
```diff
@@ -84,7 +84,7 @@
           class="store-openings-date"
           *ngFor="let opening of specialDayOpeningList"
         >
-          <span class="store-openings-day bold">{{ opening.weekDay }}</span
+          <span class="store-openings-day bold">{{ opening.name }}</span
```
Platonn, line 87: [QUESTION]

Is this really a deliberate change in this PR?
(Why asking: it seems not to be related to my review comments, I believe. I don't mind adding other fixes in this PR as well. Just wanted to understand and double-check it's a deliberate fix/change)

feature-libs/storefinder/core/facade/store-finder.service.ts
```diff
@@ -56,7 +56,7 @@ export class StoreFinderService implements OnDestroy {
 
   /**
    * Returns observable for store's entities
-   * TODO: The return value of thism ethod signature is wrong, should be StoreFinderSearchPage.
+   * Notes: The return value of this method signature is wrong, should be StoreFinderSearchPage.
```
Platonn: [MINOR]

What do you think about replacing `Notes:` with `CXSPA-xxxx:` (with the number of the relevant ticket you create for this improvement to be done in the future.

If we don't create ticket and schedule it accordingly, the wrong type will likely stay here forever.

feature-libs/asm/customer-360/components/sections/customer-360-map/customer-360-map.component.html
kimhw0630: It is a deliberate change that I discovered while testing and reviewing the code.  thanks

feature-libs/asm/customer-360/components/customer-360/customer-360.component.html
kimhw0630: Since Customer360 was never released, we don't need a feature flag.  It was there because one of the tab section(Promotions) was planned for release later but we are releasing it in 6.6.  thanks

feature-libs/storefinder/core/facade/store-finder.service.ts
kimhw0630: replaced with a ticket number (CXSPA-4871). thanks

Platonn (approved):

PR #17903, opened 2023-10-03
test: SSR E2E Test Scenarios (#17666)

Zeyber:


projects/ssr-tests/src/proxy.utils.ts
```diff
@@ -0,0 +1,85 @@
...
+    });
+  });
+}
+
+/**
+ * Send an http GET request to a given url.
+ */
+export async function sendRequest(path: string) {
+  return new Promise((resolve, reject) => {
+    const req = http.get({ ...REQUEST_OPTIONS, path }, (res: any) => {
```
Platonn, line 68: Let's allow for injecting extra Spartacus config chunk, by passing it via a special cookie. 
We already have a mechanism for this (implemented in Spartacus `TestConfigModule`) and it's already used in our Cypress E2E tests. Let's just use it here as well.
The cookie name is `cxConfigE2E` and the Spartacus `TestConfigModule` reads it and injects
the config passed  via the cookie into Spartacus global config.

Beloww a snippet of code to pass a the cxConfig as an extra cookie:
NOTE: btw. I've changed the signature of the function's argument from `path:string` to object `{path:string, cxConfig?: Config}`

```suggestion
import { Config } from '@spartacus/core';
/* ... */

export async function sendRequest({
  path,
  cxConfig
}: {
  path: string,
  cxConfig?: Config
}) {
  return new Promise((resolve, reject) => {
    const req = http.get({ 
      ...REQUEST_OPTIONS, 
      path,
      headers: {
        Cookie: buildCxConfigCookie(cxConfig)
    }, (res: any) => {
```

and let's add a function:

```ts
/**
  * Builds a cookie string with key 'cxConfigE2E' and the given `cxConfig`
  * object as a value that is stringified and URI-encoded.
  *
  * Note: `TestConfigModule` of Spartacus will read this cookie
  * and inject the passed config chunk into Spartacus global config.
  */
function buildCxConfigE2ECookie(cxConfig?: object): string {
  if(!cxConfig) {
    return {};
  }
  const cookieKey = 'cxConfigE2E';
  const cookieValue = encodeURIComponent(JSON.stringify(cxConfig));
  return `${cookieKey}=${cookieValue}`
}
```

PR #17904, opened 2023-10-03
chore: CXSPA-4841 Fix unit tests and dependencies

pawelfras:
This PR contains dependency updates and fixes for issues in unit tests and issues captured by unit tests caused by upgrading RxJS to version 7.

closes [CXSPA-4841](https://jira.tools.sap/browse/CXSPA-4841)


integration-libs/cdc/user-profile/register/cdc-register-component.service.spec.ts
```diff
@@ -38,6 +38,9 @@ const userRegisterFormData: UserSignUp = {
   preferences: {},
 };
 
+// Disable unhandled error logging
+config.onUnhandledError = () => {};
```
Platonn: [MEDIUM]
I'm believe that setting this flag in the _global context_ might affect also behavior of other test suites that run after this test suite. I believe changing the global rxjs settings in `beforeEach/All` and reseting to original value in `afterEach/All` would be more bullet-proof and not affect other test suites which run in the same shared global JS enviroment.

Anyway, I'm wondering whether we need to set this config at all, or we can adjust our tests to behave better? Did you consider any alternative solutions, that don't require setting this global rxjs config?

feature-libs/pdf-invoices/karma.conf.js
```diff
@@ -4,9 +4,8 @@
 module.exports = function (config) {
   config.set({
     basePath: '',
-    frameworks: ['parallel', 'jasmine', '@angular-devkit/build-angular'],
```
Platonn, line 7: [QUESTION]

Why is removing `karma-parallel` related to the upgrade of rxjs?

btw. I can understand that for small libraries that have little unit tests, the parallelization of the unit tests execution might not bring much value. `pdf-invoices` might be one of such libraries. But I'm wondering why you made this change in _this_ rxjs-upgrad PR ?

feature-libs/pdf-invoices/karma.conf.js
```diff
@@ -4,9 +4,8 @@
...
+    frameworks: ['jasmine', '@angular-devkit/build-angular'],
     plugins: [
-      require('karma-parallel'),
       require('karma-jasmine'),
```
Platonn, line 9: [QUESTION]
Why is removing `karma-parallel` related to the upgrade of rxjs?

btw. I can understand that for small libraries that have little unit tests, the parallelization of the unit tests execution might not bring much value. `pdf-invoices` might be one of such libraries. But I'm wondering why you made this change in _this_ rxjs-upgrad PR ?

Note: analogical question to other files with removed karma-parallel

feature-libs/product-configurator/package.json
```diff
@@ -33,13 +33,13 @@
...
-    "@spartacus/schematics": "5.0.0",
-    "@spartacus/storefront": "5.0.0",
-    "@spartacus/styles": "5.0.0",
+    "@spartacus/cart": "6.5.0-1",
+    "@spartacus/checkout": "6.5.0-1",
+    "@spartacus/core": "6.5.0-1",
+    "@spartacus/order": "6.5.0-1",
+    "@spartacus/schematics": "6.5.0-1",
+    "@spartacus/storefront": "6.5.0-1",
+    "@spartacus/styles": "6.5.0-1",
```
Platonn: [QUESTION]
Is this change _really_ related to this PR?
(I don't mind it, as it's just a partial PR to an epic branch. But just wanted to understand and dbl-check it)

feature-libs/pdf-invoices/karma.conf.js
pawelfras: I had the same question in my mind when I started reviewing already closed PRs that were related to the RxJS upgrade. It wasn't my decision to remove it and I'm following the trend ;) I asked @ercultimate for the reasons behind such a change and it turned out that `karma-parallel` doesn't work with the new version of `jasmine-marbles`. Eric sent me a [link](https://github.com/joeljeske/karma-parallel/issues/64) to the GitHub issue that contains his PR with the fix, but this library seems to be not maintained anymore (maybe it's another good reason to not use it). That said, I heard someone from the team suggested removing the problematic library as it was used in very few test suites.

feature-libs/product-configurator/package.json
pawelfras: No, this change is not related to this ticket, but as you mentioned - because this is a partial PR to the epic branch, I decided to make it here.

integration-libs/cdc/user-profile/register/cdc-register-component.service.spec.ts
pawelfras: While I agree, that the config change can be moved to `beforeEach/afterEach` for better safety, I'm wondering what value-added brings adjusting tests to behave better.
I followed the trend from the previous PRs containing RxJS upgrade changes and there are more usages of `config.onUnhandledError` in our code, I suggest addressing this (if needed) in a separate ticket.

Platonn: **Value:**
Consistency of tests results. Test suites should not interfere with each other implicitly. E.g. if I run some other test suite in isolation (locally, using `--include` flag, for debugging purposes) and that suite implicitly dependend on the fact that `config.onUnhandledError` was set globally (but in other test suite), I might get unexpected errors in my run of the individual test suite. But I wouldn't those errors if I run all tests all together.

If it's a common pattern to set globally `config.onUnhandledError` in the whole epic branch, it sounds reasonable to address it in a separate sub-PR.

pawelfras: We can also add a comment that such config is used only for already existing tests and we shouldn't use it for new tests.

feature-libs/pdf-invoices/karma.conf.js
Platonn: Then, shall we remove `karma-parallel` from all the libraries, or just those that use `jasmine-marbles`?
Just to better understand the consequences: What's the impact of removing `karma-parallel` from all libraries on the overall time waiting for CI to pass?

Note: as it's a common pattern in the epic branch, it can be addressed in a separate PR, if neeed.

projects/core/src/util/command-query/command.service.ts
```diff
@@ -53,6 +54,16 @@ export class CommandService implements OnDestroy {
     const commands$ = new Subject<PARAMS>();
     const results$ = new Subject<ReplaySubject<RESULT>>();
 
+    // We have to provide specify handlers after RxJS version 7.3.0
```
Platonn: [NITPICK]
```suggestion
    // We have to specify selectively the handlers after RxJS version 7.3.0.
    // Otherwise, the `unsubscribe` handler would be forwarded implicitly
    // to `notifier$` when calling `tap(notifier$)`.
    // But we don't want to forward `unsubscribe`, in particular.
```

Platonn (approved):

integration-libs/cdc/user-profile/register/cdc-register-component.service.spec.ts
pawelfras: To be sure we're on the same page - I agree it's better to move such a config to `beforeEach/afterEach` or similar to not change the global scope and I see an added value for this. What I'm wondering about is what added value brings to adjusting already existing unit tests to avoid setting such a config. Let's explore this in a separate task https://jira.tools.sap/browse/CXSPA-4870

Platonn: Yup. there are 2 levels of changes:
1) SHOULD: move the global rxjs config setting (and resetting) to `before/AfterEach`. lets do it in a separate PR
2) COULD: investigate if we can get rid of setting (and resetting) this config at all - maybe by adapting a little bit the tests bodies. We can to it but don't have to. A _potential_ added value _might be_ reducing the complexity of tests by getting rid of the config setting. But the real return of investment depends on how much more/less complex the **test bodies** might have to become then...

feature-libs/pdf-invoices/karma.conf.js
pawelfras: Here's the separate ticket for this: https://jira.tools.sap/browse/CXSPA-4869

Platonn (approved):

feature-libs/checkout/base/core/facade/checkout-delivery-address.service.spec.ts
```diff
@@ -55,15 +55,23 @@ class MockCheckoutQueryFacade implements Partial<CheckoutQueryFacade> {
...
+  beforeAll(() => {
+    // configure rxjs to not crash node instance with thrown errors
+    // TODO: CXSPA-4870 verify if can be avoided
+    config.onUnhandledError = () => {};
+  });
+
+  afterAll(() => {
+    // reset rxjs configuration
+    config.onUnhandledError = null;
+  });
```
Platonn: Shouldn't we reset to the original function (which I guess throws error?)
```suggestion
  let originalRxjsOnUnhandledError;
  
  beforeAll(() => {
    // configure rxjs to not crash node instance with thrown errors
    // TODO: CXSPA-4870 verify if can be avoided
    originalRxjsOnUnhandledError = config.onUnhandledError;
    config.onUnhandledError = () => {};
  });

  afterAll(() => {
    // reset rxjs configuration
    config.onUnhandledError = originalRxjsOnUnhandledError;
  });
```

Platonn (approved):

PR #17915, opened 2023-10-05
test: Test ssr server returns 500 status code when api request returns 500 code

Zeyber:
Closes: https://jira.tools.sap/browse/CXSPA-4744

### DIFF
--- a/.github/workflows/ci-continuous-integration.yml
+++ b/.github/workflows/ci-continuous-integration.yml
@@ -46,6 +46,33 @@ jobs:
       - name: Run unit tests
         run: |
           ci-scripts/unit-tests.sh          
+  build-libs:
+    # TODO: For Testing SSR Epic only, remove job before merge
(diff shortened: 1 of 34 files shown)


Zeyber:
The corresponding ssr error is not currently handled correctly in our app so we need to test this by checking that the spec will pass if we force our app to return a 500 code.

QA Steps: 
- Add the following to `app.module.ts`:

app.module.ts
```
const appInitializer = () => {
  throw new Error('error');
};
```

In the @NgModule
```
providers: [
  {
    provide: APP_INITIALIZER,
    useFactory: appInitializer,
    multi: true,
  }
]
```

- Run `npm run build && npm run build:ssr:local-http` to rebuild the ssr app.
- Run `npm run test:ssr` to run the tests.
- Test spec `should receive 500 error response with request` should now pass.

Platonn (approved):

PR #17933, opened 2023-10-06
feat: show confirm dialog when editing a quote (CXSPA-4900)

Uli-Tiger:


feature-libs/quote/assets/translations/en/quote.i18n.ts
```diff
@@ -154,6 +154,13 @@ export const quote = {
           successMessage: 'Quote rejected',
         },
       },
+      all: {
```
Larisa-Staroverova, line 157: What about **forAll**?

Larisa-Staroverova (changes requested):
@Uli-Tiger, 
the message is misleading. There are no items in the cart. 
image

Larisa-Staroverova (approved):

PR #17934, opened 2023-10-06
feat: rename components (classes, html selectors, filenames) and modules

kimhw0630:
Closes: https://jira.tools.sap/browse/CXSPA-4916


feature-libs/asm/customer-360/components/config/asm-customer-360-config.ts
```diff
@@ -12,10 +12,10 @@ import { Config } from '@spartacus/core';
   providedIn: 'root',
   useExisting: Config,
 })
-export abstract class Customer360Config {
+export abstract class AsmCustomer360Config {
   customer360?: AsmCustomer360TabsConfig;
```
Platonn: [MEDIUM]
Let's rename also the global Spartacus config key to `asmCustomer360`

Platonn (approved):

PR #17946, opened 2023-10-10
a11y: improve aria labels for, button input and DDLB (CXSPA-4922)

Uli-Tiger:
@steinsebastian tried to to a minimal solution, as this a reuse component and it should work in all contexts

- DDLB: added an aria-label to describe it
- button/input linked to existing label with aria-labelledby


Uli-Tiger:
duplicate of https://github.com/SAP/spartacus/pull/17950

PR #17948, opened 2023-10-10
Feature/cxspa 4931 installation schematics

ChristophHi:


Uli-Tiger (approved):
lgtm
but was not bale to actually test this, as the test script seems only to work in mac os

PR #17950, opened 2023-10-10
a11y: improve aria labels for, button input and DDLB (CXSPA-4922)

Uli-Tiger:
@steinsebastian tried to to a minimal solution, as this a reuse component and it should work in all contexts
- DDLB: added an aria-label to describe it
- button/input linked to existing label with aria-labelledby


projects/storefrontlib/shared/components/chat-messaging/messaging/messaging.component.html
```diff
@@ -1,5 +1,5 @@
 <ng-container *ngIf="messageEvents$ | async as messageEvents">
-  <div class="container">
+  <div class="container" role="list">
```
steinsebastian: The attribute role="list" has no effect on this div as the message-card divs are lower in the hierarchy. The attribute role="list" should be moved to the div with id="cx-messages".

projects/storefrontlib/shared/components/chat-messaging/messaging/messaging.component.html
```diff
@@ -1,5 +1,5 @@
...
     <div class="cx-avatar-line"></div>
     <div
       class="cx-messages"
```
steinsebastian, line 6: The role="list" attribute should go to this div.

steinsebastian (changes requested):

steinsebastian (approved):

PR #17959, opened 2023-10-11
A11Y: Quote modal needs enhancements for screen readers

Larisa-Staroverova:
Closes [CXSPA-4923](https://jira.tools.sap/browse/CXSPA-4923)


feature-libs/quote/components/actions/confirm-dialog/quote-actions-confirm-dialog.component.html
```diff
@@ -83,5 +83,16 @@
         </button>
       </div>
     </div>
+
+    <!-- For screen reader purposes (not visual)-->
+    <div
+      class="cx-visually-hidden"
+      aria-atomic="true"
+      aria-live="assertive"
+      role="alert"
```
steinsebastian: I would remove role=alert and set aria-live to "polite" like in the add-to-cart dialog (see added-to-cart-dialog.component.html). So that both modals are similar in terms of accessibility.

feature-libs/quote/components/actions/confirm-dialog/quote-actions-confirm-dialog.component.html
```diff
@@ -83,5 +83,16 @@
...
+      aria-relevant="additions"
+    >
+      {{ confirmationContext.a11y.title | cxTranslate }}
```
steinsebastian: I would provide the detailed text to the screenreader user. I.e. for cancel-modal " Are you sure you want to cancel this quote request?"
Furthermore if a confirm text and a warning message is displayed. Both should be included.

steinsebastian (changes requested):

feature-libs/quote/components/testing/common-quote-test-utils.service.ts
```diff
@@ -192,4 +192,77 @@ export class CommonQuoteTestUtilsService {
       caret.click();
     }
   }
+
+  protected static collectFountElements(
```
steinsebastian: Please correct typo: fount vs. found

steinsebastian (changes requested):

feature-libs/quote/components/testing/common-quote-test-utils.service.ts
Larisa-Staroverova: Good point! Thanks a lot!
I renamed the method to `collectElementsWithClassName`

steinsebastian (approved):

PR #17960, opened 2023-10-11
Follow up of PR #17895 

anjana-bl:


anjana-bl:
temporary PR, hence closing

PR #17966, opened 2023-10-12
test: Test for SSR 404 status code

Zeyber:
Closes: https://jira.tools.sap/browse/CXSPA-4743


Platonn (approved):
Works well. 

For reference: I tested it by adding the following custom `APP_INITIALIZER` to the `app.module`, so the app really returns 404, when the URL ends with `/not-existing-page`. 

```ts
    {
      provide: APP_INITIALIZER,
      multi: true,
      useFactory: () => {
        const location = inject(Location);
        const response = inject(RESPONSE);

        return () => {
          const url = location.path();
          if (url.endsWith('/not-existing-page')) {
            response.status(404);
          }
        };
      },
    },
```

PR #17967, opened 2023-10-12
test: SSR tests for timeouts

Zeyber:
Closes: https://jira.tools.sap/browse/CXSPA-4745

### DIFF
--- a/projects/ssr-tests/src/ssr-testing.spec.ts
+++ b/projects/ssr-tests/src/ssr-testing.spec.ts
@@ -33,7 +33,7 @@ describe('SSR E2E', () => {
     ]);
   });
 
-  it('should receive cached response with next request', async () => {
+  it('should receive cached response with next request if previous request timed out', async () => {
     await Ssr.startSsrServer();
     proxy = await ProxyServer.startProxyServer({
       target: BACKEND_BASE_URL,
--- a/projects/ssr-tests/src/ssr-testing.spec.ts
+++ b/projects/ssr-tests/src/ssr-testing.spec.ts
@@ -47,17 +47,20 @@ describe('SSR E2E', () => {
     Log.assertMessages([`Render from cache (${REQUEST_PATH})`]);
   });
 
-  // TODO: Test incomplete
-  xit('should receive 404 response when page is not existing', () => {});
+  it('should receive 404 response when page is not existing', async () => {
+    proxy = await ProxyServer.startProxyServer({
+      target: BACKEND_BASE_URL,
+    });
+    const response: any = await ProxyServer.sendRequest(
+      REQUEST_PATH + '/not-existing-page'
+    );
+    expect(response.statusCode).toEqual(404);
+  });
 
-  // TODO: Test incomplete
-  xit('should receive 500 error response with request', async () => {
+  it('should receive 500 error response when a backend API returned server error', async () => {
     proxy = await ProxyServer.startProxyServer({
       target: BACKEND_BASE_URL,
       throwStatus: 500,
-    });
-    proxy.on('proxyRes', function (proxyRes: any) {
-      proxyRes.statusCode = 500;
     });
     const response: any = await ProxyServer.sendRequest('/');
     expect(response.statusCode).toEqual(500);
--- a/projects/ssr-tests/src/ssr-testing.spec.ts
+++ b/projects/ssr-tests/src/ssr-testing.spec.ts
@@ -33,15 +33,18 @@ describe('SSR E2E', () => {
     ]);
   });
 
-  // TODO: Test incomplete
-  xit('should receive cached response with next request', async () => {
+  it('should receive cached response with next request', async () => {
+    await Ssr.startSsrServer();
     proxy = await ProxyServer.startProxyServer({
       target: BACKEND_BASE_URL,
     });
     const response: any = await ProxyServer.sendRequest(REQUEST_PATH);
     expect(response.statusCode).toEqual(200);
+    Log.waitUntilLogContainsText(`Rendering completed (${REQUEST_PATH})`);
+    const response2: any = await ProxyServer.sendRequest(REQUEST_PATH);
+    expect(response2.statusCode).toEqual(200);
 
-    Log.assertMessages(['Render from cache (/)']);
+    Log.assertMessages([`Render from cache (${REQUEST_PATH})`]);
   });
 
   // TODO: Test incomplete
@@ -69,7 +72,7 @@ describe('SSR E2E', () => {
     const response: any = await ProxyServer.sendRequest(REQUEST_PATH);
 
     // Waits a time for server to timeout
-    // await new Promise((res) => setTimeout(res, 15000));
+    await new Promise((res) => setTimeout(res, 15000));
 
     expect(response.statusCode).toEqual(500);
     Log.assertMessages(['timeout']);
--- a/projects/ssr-tests/src/ssr-testing.spec.ts
+++ b/projects/ssr-tests/src/ssr-testing.spec.ts
@@ -10,7 +10,6 @@ describe('SSR E2E', () => {
 
   beforeEach(async () => {
     Log.clearSsrLogFile();
-    await Ssr.startSsrServer();
   });
 
   afterEach(async () => {
@@ -19,6 +18,7 @@ describe('SSR E2E', () => {
   });
 
   it('should receive success response with request', async () => {
+    await Ssr.startSsrServer();
     proxy = await ProxyServer.startProxyServer({
       target: BACKEND_BASE_URL,
     });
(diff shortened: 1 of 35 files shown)


projects/ssr-tests/src/ssr-testing.spec.ts
```diff
@@ -60,21 +63,18 @@ describe('SSR E2E', () => {
     expect(response.statusCode).toEqual(500);
   });
 
-  // TODO: Currently, the ssr server still responds with 200 quickly despite the proxy delay
-  // TODO: Test incomplete
-  xit('should receive 500 error response with timed-out request', async () => {
+  it('should receive 500 error response with timed-out request', async () => {
```
Platonn, line 66:
```suggestion
  it('should receive 500 error response with timed-out backend request', async () => {
```

projects/ssr-tests/src/ssr-testing.spec.ts
```diff
@@ -33,15 +33,18 @@ describe('SSR E2E', () => {
     ]);
   });
 
-  // TODO: Test incomplete
-  xit('should receive cached response with next request', async () => {
+  it('should receive cached response with next request', async () => {
```
Platonn: [QUESTION]
Would you agree with such a rephrasing of the description?:

```suggestion
  it('should receive cached response with next request if previous request timed out', async () => {
```

projects/ssr-tests/src/ssr-testing.spec.ts
```diff
@@ -33,15 +33,18 @@ describe('SSR E2E', () => {
...
       target: BACKEND_BASE_URL,
     });
     const response: any = await ProxyServer.sendRequest(REQUEST_PATH);
     expect(response.statusCode).toEqual(200);
+    Log.waitUntilLogContainsText(`Rendering completed (${REQUEST_PATH})`);
 
-    Log.assertMessages(['Render from cache (/)']);
+    const response2: any = await ProxyServer.sendRequest(REQUEST_PATH);
+    expect(response2.statusCode).toEqual(200);
+    Log.assertMessages([`Render from cache (${REQUEST_PATH})`]);
```
Platonn: [QUESTION]
How do we ensure that the previous request _really_ timed out, so the next request _for sure_ will get response from the cache?

projects/ssr-tests/src/ssr.utils.ts
```diff
@@ -16,9 +16,11 @@ import * as Log from './log.utils';
  * The server will output a log file at the test project root named "ssr.log".
  * Funtion finishes once the server is initialized.
  */
-export async function startSsrServer(port = 4000) {
+export async function startSsrServer(port = 4000, args?: string) {
   childProcess.exec(
-    `NODE_TLS_REJECT_UNAUTHORIZED=0 PORT=${port} npm run serve:ssr --prefix ../../> ssr.log`
+    `NODE_TLS_REJECT_UNAUTHORIZED=0 PORT=${port} ${
+      args ? args + ' ' : ''
+    }npm run serve:ssr:dev --prefix ../../> ssr.log`
```
Platonn: [NITPICK] no need to change

What about simply assigning a default value for the parameter:
```diff
- export async function startSsrServer(port = 4000, args?: string) {
+ export async function startSsrServer(port = 4000, args?: string = '') {
```

and then:
```suggestion
    `NODE_TLS_REJECT_UNAUTHORIZED=0 PORT=${port} ${args} npm run serve:ssr:dev --prefix ../../> ssr.log`
```

projects/ssr-tests/src/ssr-testing.spec.ts
```diff
@@ -60,21 +63,18 @@ describe('SSR E2E', () => {
...
-    Log.assertMessages(['timeout']);
+    const response: any = await ProxyServer.sendRequest(REQUEST_PATH);
 
+    await Log.waitUntilLogContainsText(
+      `SSR rendering exceeded timeout ${TIMEOUT}, fallbacking to CSR for ${REQUEST_PATH}`
+    );
     expect(response.statusCode).toEqual(500);
-  }, 20000);
+  }, 10000);
 });
```
Platonn: [MEDIUM]
I believe this test doesn't match with the description.
e.g. we don't check anywhere that the reponse from SSR server is 500.
Moreover there is no chance to get 500, becasue - according to the current test configuration - we'll always get the _SSR request timeout_ sooner that being able to get a 500 error response.

I believe we want instead to:
- _increase_ the tolerated _request timeout_ to very-very long (e.g. `timeout: 40_000`)
- _increase_ the proxy delay over backend API requests to match the Spartacus default backend timeout config (e.g. `delay: 20_000`)
- (according to default Spartacus config and the implemented http timeout interceptor, the backend requests taking more than 20s are killed and turned into an error. the error should be eventually converted to setting 500 status on the SSR response)
- so we'd like to make a request, then wait ~20-25s and expect that the app will return a response with 500 status

It's too bad that Spartacus has such a big default time limit for the outgoing requests to backend API - 20s. Without changing the app's code (e.g. it's default config), we have wait that long in our test body....
Alternatively, we might want to shorten this time limit, by setting a custom value only in our storefrontapp's `app.module`, e.g. `provideConfig({ backend: { timeout: { server: 10_000 } } })`. Then it would suffice to wait ~10-15s in our tests.

What do you think about it?

projects/ssr-tests/src/ssr-testing.spec.ts
Zeyber: I thought that the SSR engine retains the request in cache regardless of timeout so that it could serve the cached page to future requests for that url (ie. render page first time, serve cached in future). Is there another purpose for this feature? I may have understood the mechanism.

projects/ssr-tests/src/ssr-testing.spec.ts
Zeyber: I suppose it depends on [this question](https://github.com/SAP/spartacus/pull/17967/files#r1365017454). I was thinking to rename it to `should receive cached response with next request if ssr engine completes rendering`.

projects/ssr-tests/src/ssr-testing.spec.ts
Platonn: Short answer:
In general Spartacus by default doesn't cache rendered results for future use (unless the `cache: true` SsrOptimizationOption is set, which is not recommended - CDN caching should be used instead).

_However_ Spartacus by default caches the result in one exceptional case, and only for a _single_ usage in the future - when the original request timed out but the render is still pending in the background. Then we don't waste this eventually rendered result - we put it into a cache for a single use. Then we return it immediately to the very next incoming request. And we delete it from the cache. See the [spec](https://github.com/SAP/spartacus/blob/b59e27b5011fd79096731e94c1f58cc4a34e8ec3/core-libs/setup/ssr/optimized-engine/optimized-ssr-engine.spec.ts#L182-L190).

---
Long answer:
Here is a step-by-step explanation of how SAP Spartacus SSR (Server-Side Rendering) works in the case of fallback to CSR (Client-Side Rendering) when the request times out due to the rendering taking too long. (I wrote it in the past somewhere, now I'm just pasting it here for knowledge sharing):

1. When an incoming request comes to the NodeJS server, the server starts counting the time for the request timeout. This timeout is typically set to a specific duration, after which the server will return a response to the client.

2. The rendering process starts, which involves generating the HTML content for the requested page. However, if the rendering process takes longer than the request timeout duration, the server will not be able to wait for the rendering to complete before returning a response.

3. In the meantime, as the request timeout time passes, the server decides to return a CSR (Client-Side Rendering) response to the request. This means that instead of waiting for the server-side rendering to complete, the server sends back a response that includes the necessary JavaScript files and instructions for the client to perform the rendering on the browser side.

4. It's important to note that even though the server has returned a CSR response, the rendering process is still pending in the background. This is because in Angular SSR, there is no way to abort a pending rendering.

5. Eventually, the rendering process completes, and the server logs a message indicating that the rendering has been completed. However, since the response has already been sent to the client, this completion message is logged to the server's console and does not affect the response that has already been returned.

Angular continues the rendering process in the background even though the request has already been answered. In Spartacus SSR we've built additional features that take advantage of this behavior.

One such feature is the "reuseCurrentRendering" functionality. If another request comes in for the same URL while the rendering is still pending, and the rendering completes within the new request's timeout window, then this new request "wins" the response of the completed rendering. This means that the client will receive the fully rendered page as a response.

If there are multiple pending requests for the same URL at the moment when the rendering completes, all of them will "win" the response of the completed rendering.

If no request was awaiting the rendering completion for a specific URL, then the result HTML is stored in the in-memory cache, waiting for the next request. The very next request for the same URL will get an immediate response taken from the cache, and the cached HTML will be cleared. However, if you have configured the cache to be kept longer (using the `cache:true` configuration), the cache will be retained for a longer period, although this is not recommended.

I hope this explanation helps! Let me know if you have any further questions.

Zeyber: Thanks for the thorough explanation! I wonder if there is a need to test the rendering caching again in these e2es if it is already tested in the link you provided?

Zeyber:
Moved rendering part of this PR to https://github.com/SAP/spartacus/pull/18148

projects/ssr-tests/src/ssr-testing.spec.ts
Zeyber: These are the default options I can see in our app. Noticing max render time is 300 seconds and forced ssr timeout is 60 seconds, wouldn't we need to wait much longer than the aforementioned wait times in your suggestion? 

```
  concurrency: 10,
  timeout: 3_000,
  forcedSsrTimeout: 60_000,
  maxRenderTime: 300_000,
  reuseCurrentRendering: true,
  debug: false,
  renderingStrategyResolver: defaultRenderingStrategyResolver(
    defaultRenderingStrategyResolverOptions
  )
};
```

Zeyber: Ok I have just found the http-timeout-intercept with the 20 second delay. Thinking...

Zeyber: I have modified the test to reflect the suggestions. How does it look to you?

PR #17976, opened 2023-10-13
MyAccount-1: CXINT: 2317 -> Order History & Details

anjana-bl:
This is a Follow-up of https://github.com/SAP/spartacus/pull/17895 

The following is a checklist for new epics or features acceptance for Spartacus, based on [our Definition of Done document](https://sap.github.io/spartacus-docs/definition-of-done/)

## General

- [x] Epic has been tested on a production-like server using Spartacus installation script
- [x] Epic has been tested in all supported browsers and devices (Check list on DoD page)
- [ ] Epic has sufficient end to end (cypress) test coverage
- [ ] Schematics code deprecations have been provided and tested
- [x] Epic complies with [Security best practices](https://sap.github.io/spartacus-docs/security-best-practices/)
- [x] Epic complies with [Accessibility best practices](https://sap.github.io/spartacus-docs/a11y-best-practices/)
- [ ] Epic CSS supports directionality (if required)
- [x] Epic code scans have been conducted [Compliance scans for branches](https://wiki.one.int.sap/wiki/display/spar/Compliance+scans+for+branches)
- [ ] Epic renders successfully in SSR

## Audits/reviews

- [x] Architecture review of the epic
- [x] Security and threat modeling audit/review from the Security team
- [x] Accessibility audit/review from the UX team

## New Library - NO NEW LIBRARY IS INTRODUCED IN THIS PR

If your epic will be introduced or introduces a new library:

- [x] Library provides schematics to install the library
- [x] Library modules can be lazy loaded appropriately
- [x] Library has been tested for a customer using Spartacus installation script

## Sample data - NO EXTRA SAMPLE DATA NEEDED FOR THIS PR

If the new feature requires new or updated sample data for a specific Commerce backend version:

- [x] Sample data has been added to the sampledata repo and merged to the next version to be released
- [x] New sample data has been loaded and tested on a dev server
- [x] New sample data has been loaded and tested in our CI server

## Documentation

- [ ] Documentation input is provided for it to be published on the doc site.

PR #17978, opened 2023-10-13
PR for landing page

anjana-bl:


projects/storefrontstyles/scss/components/content/navigation/_myaccount-v2-navigation.scss
```diff
@@ -0,0 +1,49 @@
...
+        margin-bottom: 20px;
+      }
+    }
+  }
+
+  .fas {
+    font-family: sans-serif, 'Font Awesome 5 Free';
+    font-size: var(--cx-font-medium);
+    display: inline-block;
+  }
```
Platonn: [QUESTION]
Could you help me understand how it all works now that the icons are prepended before the link name inside each CMS navigation link? e.g. what "CSS classes" (or other properties like `title`?) are now required in sample data of the CMS link components for the mechanism to work? 

Besides elaborating in the PR comment, I believe such a context would be helpful to add in a // code comment and in the official docs of the feature in SAP help page - for cusomers and for ourselves (for long-term maintenance reasons)

feature-libs/user/account/assets/translations/en/user-account.ts
```diff
@@ -25,4 +25,7 @@ export const userAccount = {
     userGreeting: 'Hi, {{name}}',
     signInRegister: 'Sign In / Register',
   },
+  MyAccountV2User: {
```
Platonn: [MEDIUM]
For consistency reasons, let's keep i18n translation keys camelCase.

Note: an analogical remark applies to the `MyAccountV2Order` main key

feature-libs/user/account/components/myaccount-v2-user/myaccount-v2-user.component.ts
```diff
@@ -0,0 +1,14 @@
...
+ * SPDX-FileCopyrightText: 2023 SAP Spartacus team <spartacus-team@sap.com>
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { Component } from '@angular/core';
+import { LoginComponent } from '../login';
+
+@Component({
+  selector: 'cx-myaccount-v2-user',
```
Platonn: [MEDIUM]
For consistency reasons, let's keep "my account" as 2 separate words, which comes down to `my-account` (instead of `myaccount`) when using using kebab-case.


Note: an analogical remark applies also to other places in this PR using kebab-case for "my account", e.g.  component selectors, css classes, file names, etc

anjana-bl: hi @Platonn  - as of now, we have kept 'myaccount' as one work across both the PRs, thinking my-account looks elongated... should we change it?

feature-libs/order/assets/translations/en/order.i18n.ts
```diff
@@ -240,4 +240,16 @@ export const MyAccountV2Order = {
...
+    items: '{{value}} Items',
+    heading: 'Orders And Returns',
+    orderNumber: 'Order Number ({{value}})',
+    purchasedOn: 'Purchased On: {{value}}',
+    orderedItems: 'Ordered Items: {{value}}',
+    totalPrice: 'Total Price: {{value}}',
+    orderDetails: 'Order Details',
+    returnOrder: 'Return Order',
+    showMore: 'Show More',
+  },
```
Platonn: [MEDIUM]
I have problem to understand why we're binding the very-general-sounding i18n namespace `myAccountView.xxx` to the (lazy-loadable) translation chunk `myAccountV2Order.json` (which is meant to contain only order-feature-related in the redesigned my account pages stuff).

Did you consider to name this main translation key to be bound to _order_, e.g. `myAccountViewOrder` instead?

feature-libs/customer-ticketing/assets/translations/en/customer-ticketing.ts
```diff
@@ -52,4 +52,10 @@ export const customerTicketing = {
     closeRequest: 'Close Request',
     ticketNotFound: 'Ticket not found.',
   },
+  myAccountV2CustomerTicketing: {
+    heading: 'Customer Service',
+    showMore: 'Show More',
+    changedOn: 'Changed On: {{value}}',
+    ticketId: 'ID: {{value}}',
+  },
```
Platonn: [QUESTION]
By placing `myAccountV2CustomerTicketing` inside the old `customerTicketing` namespace, the new labels for the `myAccountV2CustomerTicketing` will be (lazy-)loaded together with all the other customer-ticketing (non-v2!) labels. My question is: was it a deliberate choice or just accidental? 

Why I'm asking: I can see that for the _order_ feature we're loading `v2` translations separaterly from `non-v2`.

projects/storefrontstyles/scss/components/content/navigation/_myaccount-v2-navigation.scss
AshreeJay: as per this impex https://github.tools.sap/cx-commerce/spartacussampledata/blob/release/6.x/resources/spartacussampledata/import/contentCatalogs/electronicsContentCatalog/myaccount-navigation.impex icon information is passed in the styleclasses directly . updated in code comment.will be updating the docs accordingly.

feature-libs/order/assets/translations/en/order.i18n.ts
Platonn, line 1: [MINOR]
I suggest to move translations that are lazy-loaded separately (`export const myAccountV2Order = { ... }`) to a separate file, so to have 2 files:
`order.ts`
`my-account-v2-order-ts`

Otherwise (if we would keep everything in the same TS file), it might be confusing at the first sight that maybe all those translations are (lazy-)loaded in the same one JSON chunk.

feature-libs/user/account/components/myaccount-v2-user/myaccount-v2-user.component.ts
Platonn: For consistency reasons, yes please.
But feel free to defer it as a last task, if it works better for you (in terms of potential `git` conflicts etc)

feature-libs/user/account/assets/translations/en/user-account.ts
AshreeJay: done

projects/storefrontstyles/scss/components/content/navigation/_myaccount-v2-navigation.scss
Platonn: From the impex I can understand that we set CSS classes for the CMS components e.g. `fas fa-shopping-cart`.
Would you mind pasting here a screenshot of the nested HTML structure (from the Chrome html inspection dev tools) containing the navigation html?

Why I ask: I don't have an access to your server to run it locally and check it myself. Maybe I'm missing something, I'm just confused that the final HTML might not look like how I would imagine it.

As far as I remember the Font Awesome classes are usually be applied to some "dummy" elements like `<i>`, e.g. `<i class="fas fa-shoping-cart" aria-hidden="true"></i>` which live besides the string descriptions.
 So I imagine html of our navigation link should look like:
```html
<!-- <a> contains <i> with font-awesome class , BUT the string description lives aside -->
<a href="..."><i class="fas fa-shoping-cart" aria-hidden="true"></i> Order Information</a>
```

But what I _only guess_ based on the impex and the internals of `navigation-ui.component`, I guess the html of our navigation link looks like:
```html
<!-- <a> has a font-awesome class  and inside it contains the string description  -->
<a href="..." class="fas fa-shoping-cart">Order Information</a>
```

AshreeJay: yes, you are right..html looks like this 
image.

Please help me with the approach expected here.

AshreeJay: whether adding `<cx-icon>` in the navigation-ui.html based on the icon information in the styleClasses attribute will be ok?

Platonn: Thanks for pasting the UI with the HTML. It seems that it looks OK so I no longer have any concerns about it. i can see that Font Awesome simply adds a `::before` block with an icon before the string link description. Thank you for helping me understand.

So no need to change IMHO. Works nice.

feature-libs/user/account/components/myaccount-v2-user/myaccount-v2-user.component.ts
AshreeJay: done

feature-libs/order/assets/translations/en/order.i18n.ts
anjana-bl: yes done

feature-libs/customer-ticketing/assets/translations/en/customer-ticketing.ts
anjana-bl: done

feature-libs/order/assets/translations/en/order.i18n.ts
anjana-bl: done

projects/storefrontlib/cms-components/myaccount/my-account-v2-navigation/my-account-v2-navigation.component.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+import createSpy = jasmine.createSpy;
+import { Component, Input } from '@angular/core';
+
+@Component({
+  selector: 'cx-navigation-ui',
+  template: '',
+})
+class MockNavigationUIComponent {
+  // @Input()
+  // dropdownMode = 'list';
```
Platonn: [MINOR]
Unused testing code can be removed

AshreeJay: done

projects/storefrontstyles/scss/components/content/navigation/_my-account-v2-navigation.scss
```diff
@@ -0,0 +1,49 @@
...
+    > nav > ul {
+      > li {
+        a {
+          border: 1px solid var(--cx-color-light);
+        }
+        margin-bottom: 20px;
+      }
+    }
+  }
+  //to apply style to the icon from the stleClasses
```
Platonn: [MINOR]
I suggest to add a bit of more context why we don't use the standard approach with `<cx-icon>` and how the usage of the `cssClasses` options look in practice: 
```suggestion
  // The links in the my-account-v2-navigation are meant to be decorated with an icon.
  // Instead of leveraging Spartacus `<cx-icon>` component to display icons,
  // we use here a pure CSS classes approach. The classes are defined in CMS
  // via the `styleClasses` option. The final HTML looks like in the following example:
  // <a href="..." class="fas fa-shoping-cart">Order Information</a>
```

projects/schematics/src/shared/constants.ts
```diff
@@ -866,6 +866,8 @@ export const LOGIN_MODULE = 'LoginModule';
 export const LOGIN_FORM_MODULE = 'LoginFormModule';
 export const LOGIN_REGISTER_COMPONENT = 'LoginRegisterComponent';
 export const LOGIN_REGISTER_MODULE = 'LoginRegisterModule';
+export const MYACCOUNT_V2_USER_MODULE = 'MyAccountV2UserModule';
+
```
Platonn: [NITPICK]
I have problem to find this const being used anywhere in the codebase

projects/schematics/src/add-spartacus/spartacus-features.ts
```diff
@@ -147,6 +148,7 @@ function configureSpartacusModules(
                 'PaymentMethodsModule',
                 'NotificationPreferenceModule',
                 'MyInterestsModule',
+                'MyAccountV2NavigationModule',
```
Platonn: [MEDIUM]
As discussed offline, let's import in the `spartacus-features.module` (in our repo and in customers' repos via schematics) the new parent module named `MyAccountV2Module` which would encapsulate any shared my-account-v2 stuff (that doesn't belong to any feature library), e.g. it would import inside the `MyAccountV2NavigationModule`.

I'm aware that now `MyAccountV2Module` would import only `MyAccountV2NavigationModule` inside, so the introducing the parent module might seen redundant. However, the benefit of it is that the parent module (`MyAccountV2Module`) allows us in the future to add underneeth more modules, if we need to add  in the upcoming iterations of development of the My Account V2, without the need to update the contents of  `spartacus-features.module` of customers.

projects/storefrontapp/src/app/spartacus/spartacus-features.module.ts
```diff
@@ -167,6 +168,8 @@ if (environment.requestedDeliveryDate) {
     PaymentMethodsModule,
     NotificationPreferenceModule,
     MyInterestsModule,
+    MyAccountV2NavigationModule,
+    MyAccountV2NavigationModule,
```
Platonn: [NITPICK]
typo: Unnecessary 2x module import

(however it's not a public API so not a big deal)

feature-libs/order/styles/components/my-account-v2/_my-account-v2-order.scss
```diff
@@ -0,0 +1,106 @@
+%cx-my-account-v2-orders {
+  margin: 1rem 0rem;
+  border: 1px solid var(--cx-color-medium);
+  .cx-myaccount-view-header {
```
Platonn: [MEIDUM]
As we agreed in the [comment under PR 1](https://github.com/SAP/spartacus/pull/17976#discussion_r1360511058): for consistency reasons, let's keep "my account" as 2 separate in our codebase (especially in the new code which is part of our public API). `myaccount` -> `my-account`

Note: analogical remark applies to many other places in this PR.

feature-libs/order/assets/translations/translations.ts
```diff
@@ -20,5 +20,9 @@ export const orderTranslationChunksConfig: TranslationChunksConfig = {
     'returnRequest',
     'reorder',
   ],
-  MyAccountV2Order: ['myAccountV2OrderHistory', 'myAccountV2OrderDetails'],
+  MyAccountV2Order: [
```
Platonn: [MEDIUM]
For consistency reasons, let's keep translation keys and chunk names camelCase
```suggestion
  myAccountV2Order: [
```

projects/storefrontstyles/scss/components/content/navigation/_my-account-v2-navigation.scss
AshreeJay: done

feature-libs/order/assets/translations/en/order.i18n.ts
Platonn: I have problem to find the code change in the latest version of this PR. I can still see `myAccountView` instead of `myAccountViewOrder`.

projects/storefrontapp/src/app/spartacus/spartacus-features.module.ts
AshreeJay: done

feature-libs/order/assets/translations/en/my-account-v2-order.i18n.ts
```diff
@@ -0,0 +1,57 @@
...
+    estimateDeliveryLabel: 'Estimated Delivery Date',
+  },
+  myAccountV2OrderDetails: {
+    returnItems: 'Return Items',
+    cancelItems: 'Cancel Items',
+    downloadInvoices: 'Download Invoices',
+    viewAllOrders: 'View All Orders',
+    noInvoices: 'Invoices are not generated yet. Please come back later.',
+  },
+  myAccountView: {
```
Platonn: [MEDIUM]
It seems that somehow [the previous comment](https://github.com/SAP/spartacus/pull/17978#discussion_r1361895374) suggesting to use here `myAccountViewOrder` instead of `myAccountView` wasn't covered. Maybe the latest changes were not pushed?

projects/schematics/src/shared/constants.ts
AshreeJay: removed

feature-libs/order/styles/components/my-account-v2/_my-account-v2-order.scss
anjana-bl: done

feature-libs/order/assets/translations/translations.ts
anjana-bl: done

feature-libs/order/assets/translations/en/my-account-v2-order.i18n.ts
anjana-bl: done

projects/schematics/src/add-spartacus/spartacus-features.ts
anjana-bl: done

feature-libs/order/assets/translations/en/my-account-v2-order.i18n.ts
```diff
@@ -0,0 +1,57 @@
...
+    estimateDeliveryLabel: 'Estimated Delivery Date',
+  },
+  myAccountV2OrderDetails: {
+    returnItems: 'Return Items',
+    cancelItems: 'Cancel Items',
+    downloadInvoices: 'Download Invoices',
+    viewAllOrders: 'View All Orders',
+    noInvoices: 'Invoices are not generated yet. Please come back later.',
+  },
+  myAccountViewOrder: {
```
Platonn: [QUESTION]
Only now I noticed that this translation key namespace doesn't have `myAccountV2` in it's prefix.

Other namespaces have it:
`myAccountV2ConsignmentTracking`
`myAccountV2OrderHistory`
`myAccountV2OrderDetails`

What do you think about having here `myAccountV2ViewOrder` for consistency? 
Thanks to this consistency, other devs (or even teams) taking this PR as a representative example to copy and follow, they will have no confusion about "when to add the prefix and when not". The answer should be: for everything that is related to the new my account, let's add the prefix "my account v2"

anjana-bl: done, changed the name to myAccountV2Orders

Platonn (approved):

PR #17990, opened 2023-10-19
feat: add terms and conditions message alert

Matejk00:
Closes: CXSPA-3171

### DIFF
--- a/integration-libs/opf/checkout/components/opf-checkout-payments/opf-checkout-payments.component.html
+++ b/integration-libs/opf/checkout/components/opf-checkout-payments/opf-checkout-payments.component.html
@@ -1,7 +1,6 @@
 <h3>
   {{ 'opf.checkout.paymentOption' | cxTranslate }}
 </h3>
-
 <ng-container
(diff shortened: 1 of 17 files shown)


integration-libs/opf/checkout/components/opf-checkout-payments/opf-checkout-payments.component.html
```diff
@@ -1,13 +1,17 @@
 <h3>
   {{ 'opf.checkout.paymentOption' | cxTranslate }}
 </h3>
+
```
FollowTheFlo: extra blank line, looks like linting is not complaining, change it if you wish

FollowTheFlo (approved):
good to me

integration-libs/opf/checkout/components/opf-checkout-payments/opf-checkout-payments.component.html
Matejk00: I removed as it seems unnecessary

PR #17996, opened 2023-10-20
fix: add new cart to tab chain

Uli-Tiger:


steinsebastian (approved):

feature-libs/quote/components/actions/link/quote-actions-link.component.html
```diff
@@ -2,7 +2,7 @@
   <section>
     <ul>
       <li>
-        <a (click)="goToNewCart()" class="link cx-action-link">
+        <a tabindex="0" (click)="goToNewCart()" class="link cx-action-link">
```
Larisa-Staroverova, line 5: There is `add-to-saved-cart.component` that behaves and looks the same way as this one, namely there are 2 links that enables the navigation accordingly.
Why not to add `cxAutoFocus` to both links and ` tabindex="0"` to the 2. one?
I would even go further and suggest to refactor this component so that its structure looks the same as `add-to-saved-cart.component`. The styling should be adjusted accordingly.

Larisa-Staroverova (commented):
Please, have a look to my comment.

feature-libs/quote/components/actions/link/quote-actions-link.component.html
Uli-Tiger: @Larisa-Staroverova: why shall we add `cxAutoFocus `directive - we only want both links to be in the tab chain, we do not want to change the focus behavior.
Regarding the second point - structures are quite similar. `add-to-savedcart` uses `divs `instead of `<ul>` and `<li>` tags, however the later has the better a11y support. So i don't see the benefit of changing that.

Larisa-Staroverova: @Uli-Tiger, good point!
@steinsebastian, what do you think where the focus should be set when one navigates to the quote overview, for instance from the quote list?

Uli-Tiger: ok i'll merge now and lets take the focus discussion offline.

PR #18009, opened 2023-10-24
fix: Load product list from transfer state in SSR

Zeyber:
Closes: https://jira.tools.sap/browse/CXSPA-2968


projects/core/src/product/facade/product-search.service.ts
```diff
@@ -31,6 +34,15 @@ export class ProductSearchService {
   }
 
   getResults(): Observable<ProductSearchPage> {
+    const results = this.getResultsFromTransferState();
+
+    if (results) {
+      return merge(
+        this.store.pipe(select(ProductSelectors.getSearchResults)),
```
kpawelczak: I wonder won't this trigger a request?

Zeyber: I don't think that selectors do like actions.

kpawelczak: Right, my mistake. How do you stop the request then, I believe the `search` function is triggered in ngOnInit in one of the components. I may be missing something here.

Zeyber: In short, in the product list component, it will call the `search()` function in this product search service only if the results are not populated in the store. The results are got with this `getResults()` method. 

This fix will return the transfer state products if they exist along with the products in the store. So search won't get called again if it is already done on the server-side.

projects/storefrontlib/cms-components/product/product-list/container/product-list-component.service.ts
```diff
@@ -84,10 +85,26 @@ export class ProductListComponentService {
...
+    this.productSearchService
+      .getResults()
+      .pipe(take(1))
+      .subscribe((results) => {
+        const loadedQuery = results?.currentQuery?.query?.value;
+        if (loadedQuery !== criteria.query) {
+          this.search(criteria);
+        }
+      });
+  }
```
Platonn: 👍 
Nice idea. The solution feels to me like very close to the roots of the issue, not fixing the symptoms. We check if we really need to reload the data, instead of checking `TransferState` issues (which is only a derivative of the root problem). I like this solution doesn't leak information about using binding TransferState to ngrx state (which is responsibility of an other layer of the application).

projects/storefrontlib/cms-components/product/product-list/container/product-list-component.service.ts
```diff
@@ -84,10 +85,26 @@ export class ProductListComponentService {
...
 
+  /**
+   * Search only if the previous search query does NOT match the new one.
+   * This prevents repeating product search calls for queries that already have loaded data.
+   */
+  protected searchIfQueryHasChanged(criteria: SearchCriteria) {
+    this.productSearchService
+      .getResults()
+      .pipe(take(1))
+      .subscribe((results) => {
```
Platonn, line 112: [QUESTION]
Can it happen that in ngrx will have no "old" state? If yes, then the search will never happen?

projects/storefrontlib/cms-components/product/product-list/container/product-list-component.service.ts
```diff
@@ -84,10 +85,26 @@ export class ProductListComponentService {
...
-        this.search(criteria);
+        this.searchIfQueryHasChanged(criteria);
       })
     );
 
+  /**
+   * Search only if the previous search query does NOT match the new one.
+   * This prevents repeating product search calls for queries that already have loaded data.
+   */
+  protected searchIfQueryHasChanged(criteria: SearchCriteria) {
```
Platonn: [NITPICK] (no need to change, just sharing an idea for a more reactive-approach implementation, which unfortunately cannot be applied because it would be a formal breaking change)

To avoid of doing an imperative `subscribe()` and naming a method with `if` (`IfSomethingSomething`), I'd consider using more stream/reactive approach (e.g. using `withLatestFrom(this.currentSearchResultsQuery$)` and `filter(""newQuery !== oldQuery"")`, and keeping `tap()` scoped only to executing the side effect `tap(()=>this.search()`):

```diff  
-      tap((state: ActivatedRouterStateSnapshot) => {
-        const criteria = this.getCriteriaFromRoute(
-          state.params,
-          state.queryParams
-        );
-        this.searchIfQueryHasChanged(criteria);
-      })
+      map((state: ActivatedRouterStateSnapshot) =>
+       this.getCriteriaFromRoute(
+          state.params,
+          state.queryParams
+        )
+      ),
+      withLatestFrom(this.currentSearchResultsQuery$),
+      filter(([newCriteria, currentQuery]) => newCriteria.query !== currentQuery)
+      tap(newCriteria => search(criteria))
```

and add:

```ts
protected currentSearchResultsQuery$ = productSearchService
  .getResults()
  .pipe(
    map(results => results?.currentQuery?.query?.value)
  );
```

---

All that said, it would change the return type of the `protected searchByRouting$` from `Observable<ActivatedRouterStateSnapshot>` to `Observable<[SearchCriteria, string]>`, which is a breaking change. Of course we're not using this observable anywhere to get it's value but only to trigger the side-effect of calling `search()`, but by the way we're changing the return type of the `protected` observable which formally is a breaking change. 

So anyway... my idea above to use more reactive approach cannot be applied here without a breaking change. But let it be just a hint for the future for seeking reactive programming instead of imperative one inside streams, if possible.

projects/storefrontlib/cms-components/product/product-list/container/product-list-component.service.ts
Zeyber: I have not managed to find a situation where a state will not be returned. In case results are not loaded yet, `getResults()` will immediately return `{}` on subscription.

Zeyber:
QA Steps:
- Run ssr server (eg. `npm run dev:ssr`).
- Navigate to any PLP page.
- Keep refreshing the page and check network tab for `products/search?` request.
  - Request should not occur whenever ssr manages to render the page before serving to client. You should see the `Rendering completed` message in the server logs whenever this happens.
    - ie. Check the network logs that `products/search?` request does not always occur (about every second page load).
  - Check that the product list does not flicker (caused by not needed search request).

For extra QA, checking nothing weird is happening when:
- Navigate away and back to AND to other PLP pages.
- Typing stuff in the search box.

projects/storefrontlib/cms-components/product/product-list/container/product-list-component.service.ts
```diff
@@ -84,10 +85,26 @@ export class ProductListComponentService {
...
+    this.productSearchService
+      .getResults()
+      .pipe(take(1))
+      .subscribe((results) => {
+        const loadedQuery = results?.currentQuery?.query?.value;
+        if (loadedQuery !== criteria.query) {
+          this.search(criteria);
```
Platonn: [MAJOR]
I did QA now and I can see the pagination is not working anymore.
I believe that we should check not only for `results.currentQuery` but also for `results.pagination`and compare it with our `SearchCriteria` (pageSize,currentPage,sort)

Platonn: I realized only now, that even our fixed implementation covering also `pageSize,currentPage,sort` might be a breaking change - it won't take into account any other custom URL params that customers might have created and used in their customized search mechanism.

That said, I think we should cover our bugfix with a feature flag.

Zeyber: These issues should be fixed now. I had to add quite few different methods to handle each case unfortunately because the handling for each situation is non-uniform (as you will be able to tell by the functions I added).

Matejk00 (approved):

PR #18010, opened 2023-10-24
refactor: add 3 global functions domains

FollowTheFlo:
Domains are needed to separate global functions available for various  pages, domains split into:
window.Opf.payments.checkout
for global functions running on Checkout pages
window.Opf.payments.redirect
for global functions running on Verification page
window.Opf.payments.global
for global functions running for QuickBuy, CTA script, it would includes at least PDP, cart pages.

aggreed with Bamboo team on those domains, see:
https://sap-cx.slack.com/archives/C04JDEA9C90/p1695219327477429?thread_ts=1694785620.784269&cid=C04JDEA9C90
This logic is used for Upscale.
https://wiki.one.int.sap/wiki/pages/viewpage.action?pageId=2592032286

### DIFF
--- a/integration-libs/opf/base/core/facade/opf-global-functions.service.ts
+++ b/integration-libs/opf/base/core/facade/opf-global-functions.service.ts
@@ -13,6 +13,7 @@ import {
 import { WindowRef } from '@spartacus/core';
 import {
   ErrorDialogOptions,
+  GlobalFunctionsDomain,
   GlobalFunctionsInput,
   GlobalOpfPaymentMethods,
   KeyValuePair,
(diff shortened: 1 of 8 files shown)


integration-libs/opf/base/core/facade/opf-global-functions.service.ts
```diff
@@ -40,46 +38,44 @@ export class OpfGlobalFunctionsService implements OpfGlobalFunctionsFacade {
...
         break;
       default:
         break;
     }
-
-    this._isGlobalServiceInit = true;
   }
 
-  removeGlobalFunctions(): void {
-    if (!this._isGlobalServiceInit) {
```
FollowTheFlo, line 66: this flag was letting the global functions available even when payment methods 'not using global functions'(eg:iframe,page redirect) were selected. more secure to remove it.

rmch91 (approved):

PR #18022, opened 2023-10-26
Upgrade to Angular 16

pawelfras:
Migration:
- [x] - Nx migrate to 16.10.0
- [x] - Nx updated "@angular/" packages to 16.2.8, "@angular-devkit/" to 16.2.5 and "@angular-eslint/" to 16.2.0
- [x] - manual bump "@angular/" to 16.2.10
- [x] - manual bump "@angular-devkit/" to 16.2.7
- [x] - manual bump "ngx-infinite-scroll" to v16.0.0
- [x] - manual bump "@ng-select/ng-select" to v11.1.1
- [x] - manual bump "@angular-eslint/schematics" to v16.2.0
- [x] - manual bump "@ngrx/" to v16.0.1
- [x] - bump versions in librariers
- [x] - automatic migration - changes from "@nrwl" to "@nx"
- [x] - automatic migration - removed deprecated "CanActivate" interface
- [x] fixed type errors after TypeScript upgrade
- [x] went through [Angular update guide](https://update.angular.io/?l=3&v=15.0-16.0)

Build issues:
- [x] - remove 'entryComponents' from modules: See: https://github.com/angular/angular/blob/229331e11b6c214f74c9801aa52eb9cd47d7fc76/CHANGELOG.md?plain=1#L732
- [x] - handle new type of 'routerEvent' property from Scroll class. See: https://github.com/angular/angular/blob/b9a4941a327954c5c6b06a995de8f959ee0d63f6/packages/router/src/events.ts#L586
- [x] - schematics - removed 'defaultProject' that's no longer  part of angular.json
- [x] - schematics - addressed changes in SchematicTestRunner API
- [x] - ssr - "renderModuleFactor" has been removed, use "renderModule" instead

Linting issues:
- [x] - added "plugin:@angular-eslint/recommended" to .eslintrc.json
- [x] - added to project rules removed from "@angular-eslint". See: https://github.com/angular-eslint/angular-eslint/issues/1353#issuecomment-1529924573
- [x] - added "tsconfig.base.json" required by "@nx/eslint-plugin"
- [x] - all errors related to "@angular-eslint/no-empty-lifecycle-method" marked as warnings
- [x] - updated "validate-lint.sh" with a new path to nx

Unit tests issues:

- [x] - set `clearContext: true` in all karma config files to avoid `Some of your tests did a full page reload!` error
- [x] - fixed issues caused by changes in Router API - `events: Observable<RouterEvent>` -> `events: Observable<Event>`
- [x] - fixed issues with `jest.spyOn` when used on properties that are not configurable - caused by `zone.js` update (see: https://github.com/angular/angular/issues/50756#issuecomment-1613944076)
- [x] - fixed `Error: NG04002: Cannot match any routes.` that causes an infinite loop during running tests. See: https://github.com/angular/angular/blob/main/CHANGELOG.md#router-17
- [x] - changed `waitForAsync` causing issues to `async` in `unit-level-order-overview.component.spec.ts`
- [x] - fix async test in quick-order.service.spec.ts causing errors thrown in `afterAll`
- [x] - TS issue - `observable.toPromise()` -> `firstValueFrom(observable)` in tests for schematics
- [x] - updated schematics snapshot
- [x] - removed `globalSetup: 'jest-preset-angular/global-setup'` as it is no longer needed and to avoid `ngcc` warning


feature-libs/cart/quick-order/core/services/quick-order.service.spec.ts
```diff
@@ -325,24 +325,26 @@ describe('QuickOrderService', () => {
...
+
+    service
+      .getSoftDeletedEntries()
+      .pipe(take(1))
       .subscribe((result) => {
         expect(result).toEqual({});
       });
-    done();
-  });
+  }));
```
Platonn: [MINOR]
I can see `expect` in a `.subscribe()` callback (two times). How are we sure those `expects` will be called, if the implementation is possibly broken in the future so the observable doesn't emit at all?

I believe we should use either:
- call `done()` callback in the subscription callback (however now we have 2 subscribe callbacks. And we shouldn't call `done()` more than once!)
- if the observable is meant to emit synchronously, we can simply copy the `result` to a variable declared higher above (`subscribe((result) => variable = result)`. And then **after the subscription** call the `expect(variable).toEqual(xxx)`

---

Note: I can see that even before your changes, this test wasn't caring about observable possible not-emitting (i.e. done callback was called outside the subscription - in the wrong place).

docs/libs/creating-lib.md
```diff
@@ -292,15 +292,15 @@ Use the following template:
 }
 ```
 
-- `tsconfig.spec.json` - add `"target": "es2020", "module": "es2020"` in `"compilerOptions"`:
+- `tsconfig.spec.json` - add `"target": "es2022", "module": "es2022"` in `"compilerOptions"`:
```
Platonn, line 295: [QUESTION] (low priority)
(not related to this file, but to es2020 in general):

Should we also change this place to es2022? https://github.com/SAP/spartacus/blob/26c6c14939bb2512952fb95a9826558e05d7700c/tools/config/manage-dependencies.ts#L165

pawelfras: good catch, thanks!

feature-libs/cart/quick-order/core/services/quick-order.service.spec.ts
```diff
@@ -325,24 +325,20 @@ describe('QuickOrderService', () => {
...
-  });
+    service.getSoftDeletedEntries().subscribe((val) => {
+      result = val;
+    });
+    expect(result).toEqual({ mockCode1: mockEntry1 });
+
+    tick(7000);
+
+    expect(result).toEqual({});
+  }));
```
Platonn, line 341: [NITPICK]
For memory safety, let's unsubscribe, when we no longer need the subscription.

Platonn (dismissed):
✅ LGTM

Note: while building schematics lib, I had an intermittent problem that the build output JS file was malformed (in place where the file should end with `//# sourceMapUrl=<url here>` there was additionally a part of the file copied once again in the end of the file). This made the file unable to execute correctly later on, when I tried to install Spartacus in a fresh app, resulting in an error:
```
ng add @spartacus/schematics@6.6.0-1 --ssr --baseUrl="https://40.76.109.9:9002/"
[...]
Unexpected token '}'
```

Below the screenshot with the content of the malformed JS file, `schematics-config-mappings.js`, the output of the build:
image

Note: a similar intermittent issue could be observed once on the machine of @pawelfras 

All that said, I wasn't able to reproduce this build issue anytime again. I've tried many times: cloning the fresh Spartacus repo, building all libs, publishing via verdaccio and installing in a fresh app. Nor Paweł was able to reproduce it again. I don't know the root causes.

Anyway, before releases we do a check whether we can install the built spartacus in a fresh Angular app. So if such a problem happens again, we will catch it before release. The remedy is to JUST TO REBUILD SCHEMATICS PROJECT. For now, we don't know the root cause of the problem, which seems to be intermittent. Also, I have no evidence whether this problem is related to the Angular 16 ugprade or not, becasue I couldn't reproduce it again on my machine either on this branch or on the old `develop` branch.

Platonn (approved):

PR #18023, opened 2023-10-26
fix: VO does not read the top section of the asm 360 dialog (CXSPA-3678)

kimhw0630:
Closes: https://jira.tools.sap/browse/CXSPA-3678


feature-libs/asm/customer-360/components/asm-customer-360/asm-customer-360.component.ts
```diff
@@ -143,10 +140,6 @@ export class AsmCustomer360Component
     this.setTabData();
   }
 
-  ngAfterViewInit(): void {
-    this.updateTabIndex(this.activeTab);
```
kpawelczak, line 147: Is this not needed anymore?

kimhw0630: we don't need this anymore because we want to focus on the close button when the dialog opens.

kpawelczak (approved):

PR #18025, opened 2023-10-27
Remove deprecated hasError

kpawelczak:


projects/core/src/util/command-query/command.service.ts
```diff
@@ -54,6 +54,7 @@ export class CommandService implements OnDestroy {
     const commands$ = new Subject<PARAMS>();
     const results$ = new Subject<ReplaySubject<RESULT>>();
 
+    let notifierHasError = false;
```
pawelfras, line 57: A few comments:
- did you verify as we discussed whether there is a way to avoid introducing new variables and handle this in a reactive way?
- it seems to be a shared state for each notifier related to the command. It may cause a false result as there is no place where `notifierHasError` is set to the initial value. Maybe there is a way to have a separate state for each notifier? 🤔
- if we have to stick to the imperative way of handling the error state, such a logic will be on our side which means it requires more unit tests that cover all scenarios, e.g.  one related to the note above, where are multiplier notifiers and one has an error

PR #18030, opened 2023-10-29
CXSPA-5174

anjana-bl:
1. fixing alignment issue in Landing Page
2. Altering font size in landing page
3. renaming injection token MY_ACCOUNT_V2_ORDER to USE_MY_ACCOUNT_V2_ORDER

### DIFF
--- a/projects/storefrontstyles/scss/layout/page-templates/_myaccount-v2-page.scss
+++ b/projects/storefrontstyles/scss/layout/page-templates/_myaccount-v2-page.scss
@@ -1,33 +1,19 @@
 %MyAccountViewPageTemplate {
   max-width: var(--cx-page-width-max);
   margin: auto;
-  display: flex;
   cx-page-slot {
     cx-banner {
-      height: 120px;
-      margin-top: 40px;
-      margin-bottom: -500px;
-      border: 1px solid var(--cx-color-light);
+      margin: 2rem 0rem 0.5rem 0rem;
+      border: 1px solid var(--cx-color-medium);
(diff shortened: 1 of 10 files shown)


Platonn (approved):

projects/storefrontstyles/scss/layout/page-templates/_myaccount-v2-page.scss
```diff
@@ -1,33 +1,19 @@
...
+    float: left;
     max-width: 25%;
-    @include media-breakpoint-down(md) {
-      width: 100%;
-      max-width: none;
-    }
   }
 
   .RightContentSlot {
+    float: left;
```
Platonn, line 16: [NITPICK] 
Do we _really_ have to use `float` instead of `flex` for styling `%MyAccountViewPageTemplate`? I believe for consistency with the rest of the project, it would be good to stick to `flex` styling.

I can see `float` is used also in other files in my-account-v2 area. Perhaps it would be good to fix it before 6.6 release:
image
(yes, I can see `float` is used also in ASM, which ideally should have used `flex`)

Platonn (commented):
1. ✅ Approving changes to `MY_ACCOUNT_V2_ORDER` -> `USE_MY_ACCOUNT_V2_ORDER`
2. Nitpick: IMHO the CSS changes should be done in a different PR, for clarity of git history
3. I don't have access to your env, so please do the QA of the styling changes in with the help of sb in your team (including cross-browsers & screen width testing)

PR #18033, opened 2023-10-30
fix Blob is not defined In SSR mode

kpawelczak:
ticket: https://jira.tools.sap/browse/CXSPA-4780


feature-libs/organization/account-summary/root/http-interceptors/blob-error.interceptor.ts
```diff
@@ -13,14 +13,18 @@ import {
...
 import { catchError, switchMap } from 'rxjs/operators';
 
 @Injectable({
   providedIn: 'root',
 })
 export class BlobErrorInterceptor implements HttpInterceptor {
-  constructor(private fileReaderService: FileReaderService) {}
+  constructor(
+    protected fileReaderService: FileReaderService,
+    protected windowRef: WindowRef
```
pawelfras: Let's use the `inject` function instead of the new parameter in the constructor to reduce the amount of breaking changes and to be in line with the agreement we introduced in the Spartacus project.

pawelfras (commented):
Apart from this one comment, let's adjust unit tests to verify the new condition 👍

feature-libs/organization/account-summary/root/http-interceptors/blob-error.interceptor.ts
kpawelczak: This wasn't part of public API, I don't think we need to use inject

pawelfras: But it's going to be a part of public API and it's hard to predict whether we'll remove this injection in the future or not. That said, actually I propose, completely removing the `constructor` and using the `inject` function in both cases :)

feature-libs/organization/account-summary/root/http-interceptors/blob-error.interceptor.ts
```diff
@@ -11,16 +11,18 @@ import {
...
 import { Observable, throwError } from 'rxjs';
 import { catchError, switchMap } from 'rxjs/operators';
 
 @Injectable({
   providedIn: 'root',
 })
 export class BlobErrorInterceptor implements HttpInterceptor {
-  constructor(private fileReaderService: FileReaderService) {}
+  fileReaderService = inject(FileReaderService);
+  windowRef = inject(WindowRef);
```
pawelfras: let's add a `protected` access modifier as we shouldn't give a chance to use them outsize 'intercept' logic

feature-libs/organization/account-summary/root/http-interceptors/blob-error.interceptor.spec.ts
```diff
@@ -34,9 +36,12 @@ describe('BlobErrorInterceptor', () => {
 
     httpMock = TestBed.inject(HttpTestingController);
     http = TestBed.inject(HttpClient);
+    windowRef = TestBed.inject(WindowRef);
   });
 
-  it(`Should extract json from errors wrapped in blob`, async () => {
+  it(`Should extract json from errors wrapped in blob`, (done) => {
```
pawelfras, line 54: what do you think about adding a test case verifying the behaviour where we're running the interceptor's logic out of the browser platform?

feature-libs/organization/account-summary/root/http-interceptors/blob-error.interceptor.ts
kpawelczak: sure, I think we should also add `readonly`

feature-libs/organization/account-summary/root/http-interceptors/blob-error.interceptor.spec.ts
kpawelczak: I had this but the test was almost the same, so I scratched it

pawelfras (dismissed):

kimhw0630 (approved):

PR #18039, opened 2023-10-30
CXSPA-4925: Set focus on create/edit quote

steinsebastian:


feature-libs/quote/core/facade/quote.service.ts
```diff
@@ -418,6 +422,20 @@ export class QuoteService implements QuoteFacade {
       }
     );
 
+  protected setFocusForCreateOrEditAction(action: QuoteActionType) {
+    if (
+      (action === QuoteActionType.EDIT || action === QuoteActionType.CREATE) &&
+      this.windowRef.isBrowser()
+    ) {
+      const storefrontElement = this.windowRef.document.querySelector(
```
Larisa-Staroverova: @steinsebastian, could you please define a separate method that fetches a HTML element? Something like this:

>  /**
>    * Get HTML element based on querySelector when running in browser
>    *
>    * @param querySelector - querySelector
>    * @returns selected HTML element
>    */
>   getElement(querySelector: string): HTMLElement | undefined {
>     if (this.windowRef.isBrowser()) {
>       return this.windowRef.document.querySelector(
>         querySelector
>       ) as HTMLElement;
>     }
>   }

Maybe we have to define a new class that takes care of fetching HTML relevant data (see `ConfiguratorStorefrontUtilsService` in `product-configurato`r lib)

Larisa-Staroverova (changes requested):

feature-libs/quote/core/facade/quote.service.ts
steinsebastian: Done

Larisa-Staroverova (approved):

PR #18041, opened 2023-10-30
fix: mobile view issue in customer360(CXSPA-5171)

kimhw0630:
Closes: https://jira.tools.sap/browse/CXSPA-5171


FollowTheFlo (approved):
good to me.

PR #18044, opened 2023-10-31
fix: Error message is not prompt when removing customer coupon (CXSPA-5148)

kimhw0630:
Closes: https://jira.tools.sap/browse/CXSPA-5148

added new action to reset process state


i53577 (approved):
LGTM and thx for your help

kpawelczak:
@kimhw0630 Consider changing only the `getDisclaimCustomerCouponResultError` function. It should work and you won't have to add additional reset action and change test cases.

```typescript
  getDisclaimCustomerCouponResultError(): Observable<boolean> {
    return combineLatest([
      /**
       * getProcessErrorFactor state might not change and won't emit error state,
       * to fix that we have to combine loading state with error.
       */
      (<Store<StateWithProcess<void>>>this.store).pipe(
        select(getProcessLoadingFactory(DISCLAIM_CUSTOMER_COUPON_PROCESS_ID))
      ),
      (<Store<StateWithProcess<void>>>this.store).pipe(
        select(getProcessErrorFactory(DISCLAIM_CUSTOMER_COUPON_PROCESS_ID))
      ),
    ]).pipe(map(([, error]) => error));
  }
```

kpawelczak (approved):

PR #18045, opened 2023-10-31
CXINT-2613

anjana-bl:
1. 'undefined undefined' is shown as name in 'Ship To' address if first name and last name are undefined.

### DIFF
--- a/feature-libs/order/root/utils/order-card-utils.ts
+++ b/feature-libs/order/root/utils/order-card-utils.ts
@@ -33,10 +33,18 @@ export function deliveryAddressCard(
   }
 
   const numbers = getAddressNumbers(deliveryAddress, textPhone, textMobile);
+  let fullName;
+  if (deliveryAddress.firstName && deliveryAddress.lastName) {
+    fullName = deliveryAddress.firstName + ' ' + deliveryAddress.lastName;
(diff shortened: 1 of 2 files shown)


feature-libs/order/root/utils/order-card-utils.ts
```diff
@@ -33,10 +33,18 @@ export function deliveryAddressCard(
...
+  } else if (deliveryAddress.firstName) {
+    fullName = deliveryAddress.firstName;
+  } else if (deliveryAddress.lastName) {
+    fullName = deliveryAddress.lastName;
+  }
 
   return {
     title: textTitle,
-    textBold: deliveryAddress.firstName + ' ' + deliveryAddress.lastName,
+    textBold: fullName,
```
Platonn, line 47: [MEDIUM]
1. What if both first and last name are `undefined`? Then `fullName` is `undefined` and it will be presented in the UI I guess? What should we present instead?

2. Let's add unit tests describing the expected behaviors

anjana-bl: 1. if fullName=undefined, nothing will be shown in UI. with old code, name will be 'undefined undefined' (string datatype)

Platonn (dismissed):

Platonn (approved):

PR #18047, opened 2023-10-31
A11Y: Improvements for quote list

Larisa-Staroverova:


feature-libs/quote/components/list/quote-list.component.html
```diff
@@ -1,111 +1,119 @@
...
+              *ngFor="let quote of quotesState.data.quotes"
+              [routerLink]="
+                {
+                  cxRoute: 'quoteDetails',
+                  params: { quoteId: quote.code }
+                } | cxUrl
+              "
+              (keydown.enter)="row.click()"
+              title="{{ quote.name }}"
+              [attr.aria-label]="'quote.list.clickableRow' | cxTranslate"
```
steinsebastian: Adding the aria-label to the row leads to only reading this label when the row is selected. The content of the row is not read anymore.
The aria-label should be moved to the cx-icon tag in the last column.

feature-libs/quote/components/list/quote-list.component.html
```diff
@@ -1,111 +1,119 @@
...
+                }}</span>
+              </td>
+              <td role="cell" class="cx-updated-date">
+                <div class="cx-mobile-header">
+                  {{ 'quote.list.updated' | cxTranslate }}
+                </div>
+                {{ quote?.updatedTime | cxDate: dateFormat }}
+              </td>
+              <td role="cell" class="cx-nav-caret">
+                <cx-icon [type]="iconTypes.CARET_RIGHT"></cx-icon>
```
steinsebastian: Here the aria-label for the information about clicking the row should be added. Then we would get all the information of the row plus the hint for clicking.
Screenshot 2023-11-02 at 11 13 54

steinsebastian (changes requested):

steinsebastian (approved):

PR #18049, opened 2023-10-31
Spartacus OPF schematics update

Matejk00:
Closes: CXSPA-4396


FollowTheFlo (changes requested):
Unit tests is failing on
nx run opf:test-jest --coverage --runInBand
replicable locally.
I tried to regenerate snapshots (npm test -- -u) but does not fix the problem. I believe
`spartacus/integration-libs/opf/schematics/add-opf/__snapshots__/index_spec.ts.snap`
needs to be updated manually to reflect latest change.

FollowTheFlo (approved):

PR #18059, opened 2023-11-02
Feature/cxspa 5177 abstract order context

ChristophHi:


feature-libs/product-configurator/common/components/configure-cart-entry/configure-cart-entry.component.ts
```diff
@@ -11,31 +11,46 @@ import {
...
-    .pipe(map((cart) => cart.code));
-
   protected static readonly ERROR_MESSAGE_ENTRY_INCONSISTENT =
     "We don't expect order and quote code defined at the same time";
   @Input() cartEntry: OrderEntry;
   @Input() readOnly: boolean;
   @Input() msgBanner: boolean;
   @Input() disabled: boolean;
 
+  abstractOrderContext = inject(AbstractOrderContext, { optional: true });
```
Platonn, line 43: 👍 nice you made it `optional:true`!

feature-libs/quote/components/items/quote-items.component.ts
```diff
@@ -5,50 +5,84 @@
...
+          this.abstractOrderContextSource.type$.next(AbstractOrderType.QUOTE);
+          return zip(of(quote), of(true));
+        } else if (!quote.isEditable) {
+          this.abstractOrderContextSource.id$.next(quote.cartId);
+          this.abstractOrderContextSource.type$.next(
+            AbstractOrderType.SAVED_CART
+          );
+          return combineLatest([
+            this.multiCartFacade.getCart(quote.cartId),
+            of(true),
```
Platonn, line 84: nitpick: for readability instead of `true` I suggest passing an object with semantic key, e.g. `{readOnly: true}`. Otherwise the reader has to deduce the meaning of this `true` only from the code in the following operators - where this boolean is named _readonly and readonly

Note: same nitpick applies for other places in this file where we have `of(true)` or `of(false)`

feature-libs/quote/components/items/quote-items.component.ts
```diff
@@ -5,50 +5,84 @@
...
+          this.abstractOrderContextSource.id$.next(quote.code);
+          this.abstractOrderContextSource.type$.next(AbstractOrderType.QUOTE);
+          return zip(of(quote), of(true));
+        } else if (!quote.isEditable) {
+          this.abstractOrderContextSource.id$.next(quote.cartId);
+          this.abstractOrderContextSource.type$.next(
+            AbstractOrderType.SAVED_CART
+          );
+          return combineLatest([
+            this.multiCartFacade.getCart(quote.cartId),
```
Platonn, line 83: question: don't you want to wait until the cart is stable (I can see you've tried (re)loading it in the code above)? I mean `this.multiCartFacade.isStable(quote.cartId)`? Just asking from top of my head, not sure if it matters in your case.

Perhaps similar question applies to line 77 with `getActive()` but not sure... from it's source I can see that any emitted value is already filtered by "isStable" internally - https://github.com/SAP/spartacus/blob/7b6dc7c3cef6dd617859ce1190fce654995620e9/feature-libs/cart/base/core/facade/active-cart.service.ts#L107

feature-libs/quote/components/items/quote-items.component.ts
ChristophHi: Yes, agree

feature-libs/quote/components/items/quote-items.component.ts
ChristophHi: Yes you are right, multiCartFacade.getCart(quote.cartId) just returns from the store. So it would be good to only render the page when the cart is stable.

For cartFacade.getActive(): The internal check that you mention does not block the emission of unstable carts. Still the quote component receives only one emission of the cart, probably because other components are already requesting it and the activeCartService uses shareReply. 
In any case I will add the check for stable here as well

PR #18066, opened 2023-11-03
fix: order confirmation shipping component not being displayed

FollowTheFlo:


feature-libs/order/components/order-confirmation/order-confirmation-shipping/order-confirmation-shipping.component.spec.ts
```diff
@@ -104,6 +104,39 @@ describe('OrderConfirmationShippingComponent', () => {
...
+          deliveryAddress: { id: 'testAddress' },
+          deliveryMode: { code: 'testCode' },
+        })
+      );
+    }
+    function configureTestingModule(): TestBed {
+      return TestBed.configureTestingModule({
+        imports: [I18nTestingModule, PromotionsModule],
+        declarations: [OrderConfirmationShippingComponent],
+        providers: [{ provide: OrderFacade, useClass: MockOrderFacade }],
```
FollowTheFlo, line 148: I couldn't find a way to override the existing MockOrderFacade to just change 1 property.
So I recreated it in this 'describe' scope.
I also tried to create an additional spy on getOrderDetails method, but no success.

kimhw0630 (approved):

PR #18076, opened 2023-11-06
fix: component with restriction not displayed after being changed in SmartEdit

FollowTheFlo:
tested successfully with s7.
eg url:
https://api.cg79x9wuu9-eccommerc1-s7-public.model-t.myhybris.cloud/occ/v2/electronics-spa/cms/pages/productList?lang=en&curr=USD&cmsTicketId=6602839949356992486e7464-519e-4d3a-a567-1d9f8cedb81d&categoryCode=576

- we use Page API ('pages/productList')

- use `categoryCode` query param


kimhw0630 (approved):
code look good to me

PR #18084, opened 2023-11-08
CXSPA-5269: Quote: Improvements for JAWS screenreader on Windows

steinsebastian:


feature-libs/quote/components/list/quote-list.component.ts
```diff
@@ -61,4 +66,47 @@ export class QuoteListComponent {
     const lastPart = stateAsString.substring(indexSeparator + 1);
     return 'quote-' + lastPart.toLowerCase();
   }
+
+  /**
+   * Retrieves an accessibility text for a row in the quote list.
+   *
+   * @param quote - quote
+   */
```
Larisa-Staroverova, line 75: Please, add a return value, for instance `@returns an a11y text for a row in the quote list`

feature-libs/quote/components/list/quote-list.component.ts
```diff
@@ -5,17 +5,22 @@
...
 
 @Component({
   selector: 'cx-quote-list',
   templateUrl: './quote-list.component.html',
   changeDetection: ChangeDetectionStrategy.OnPush,
+  providers: [CxDatePipe],
 })
 export class QuoteListComponent {
   protected quoteListComponentService = inject(QuoteListComponentService);
+  protected translation = inject(TranslationService);
```
Larisa-Staroverova: Please, rename to `translationService` because the injected service or facade should have the same name as a corresponding instance.

Larisa-Staroverova (changes requested):
@steinsebastian, could you have a look at my comments?

feature-libs/quote/components/list/quote-list.component.ts
steinsebastian: done

Larisa-Staroverova (approved):

PR #18089, opened 2023-11-09
fix: B2B user form update on standalone basesite

Matejk00:
Closes: CXSPA-5149


feature-libs/organization/administration/components/user/form/user-form.service.ts
```diff
@@ -4,21 +4,23 @@
...
-import { B2BUser, B2BUserRole } from '@spartacus/core';
+import { B2BUser, B2BUserRole, FeatureConfigService } from '@spartacus/core';
 import { CustomFormValidators } from '@spartacus/storefront';
 import { FormService } from '../../shared/form/form.service';
 
 @Injectable({
   providedIn: 'root',
 })
 export class UserFormService extends FormService<B2BUser> {
+  protected readonly featureConfigService = inject(FeatureConfigService);
```
Zeyber: I know that we can add `inject(Service, { optional: true })` as an option to our services. Do you know if we need to do this to make services optional?

Zeyber (changes requested):

feature-libs/organization/administration/components/user/form/user-form.service.ts
Matejk00: I belive we can use an optional option.

Zeyber (approved):

PR #18095, opened 2023-11-13
CXSPA-5282: Ensure agent name and password to be fully typed in e2e test

i53577:
Jira: https://jira.tools.sap/browse/CXSPA-3263


i53577:
The test was run several times on pipeline without failure

niehuayang (approved):

PR #18111, opened 2023-11-15
fix:cxspa-5366 ASM session timer is not continued when laptop screen is locked

scarai-sap:
ATT


feature-libs/asm/components/asm-session-timer/asm-session-timer.component.ts
```diff
@@ -20,6 +20,8 @@ export class AsmSessionTimerComponent implements OnInit, OnDestroy {
   protected interval: any;
   protected maxStartDelayInSeconds = 60000;
   timeLeft: number;
+  expiredTime: number;
+  timeout: number;
```
niehuayang: It seems that this variable `timeout` is redundant, we can directly use `timeleft`, can we remove it?

scarai-sap: timeout is a fixed value only read from configuration in onInit, timeleft is a countdown value, also bound to html

scarai-sap: hmm..  I guess we can read it again to avoid store it

scarai-sap: updated code

feature-libs/asm/components/asm-session-timer/asm-session-timer.component.ts
```diff
@@ -30,11 +31,11 @@ export class AsmSessionTimerComponent implements OnInit, OnDestroy {
   ) {}
 
   ngOnInit(): void {
-    this.timeLeft = this.getTimerStartDelayInSeconds();
+    this.initTimer();
```
i53577, line 34: Maybe a stupid question, when agent reopen his/her laptop, will the page be refreshed? Will the ngOnInit() be called to init timer again?

scarai-sap: yes, if it refresh the page, the timer will be reset, similar behavior also could be observed in acc

scarai-sap: In old code,     this.timeLeft = this.getTimerStartDelayInSeconds(); also refresh it back to 10 minutes.

Unless we can take timeout in backend, there is no way to avoid this scenario

i53577: If refresh page will reset timer, then it should be ok.

i53577:
LGTM

i53577 (approved):

i53577:
if reset timer will also reset page, it should be ok, LFTM

PR #18115, opened 2023-11-16
fix: adjust property and fix exception thrown

FollowTheFlo:
cta request model has changed:

- 'orderId' is now called'cartId'.

- enforced logic: when cartId property has a value, paymentsAccuntIds property must be undefined.

Tested succefully with steps:
login bruce.lee@pronto-hw.com/Password.123
go in order history pgae -> select order 00308007
you should see a 'goCardless' iframe displayed

### DIFF
--- a/integration-libs/opf/base/root/services/opf-resource-loader.service.ts
+++ b/integration-libs/opf/base/root/services/opf-resource-loader.service.ts
@@ -133,7 +133,7 @@ export class OpfResourceLoaderService extends ScriptLoader {
     if (html) {
       const element = new DOMParser().parseFromString(html, 'text/html');
       const script = element.getElementsByTagName('script');
-      if (!script?.length || !script[0]?.innerText) {
+      if (!script?.[0]?.innerText) {
         return;
       }
       Function(script[0].innerText)();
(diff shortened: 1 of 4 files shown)


integration-libs/opf/base/root/services/opf-resource-loader.service.ts
```diff
@@ -133,6 +133,9 @@ export class OpfResourceLoaderService extends ScriptLoader {
     if (html) {
       const element = new DOMParser().parseFromString(html, 'text/html');
       const script = element.getElementsByTagName('script');
+      if (!script?.length || !script[0]?.innerText) {
```
FollowTheFlo: Before, an exception was thrown when html did not have script tag which lead html snippet to not be displayed.
Html without 'script' tag is valid use case.

integration-libs/opf/base/root/services/opf-resource-loader.service.ts
```diff
@@ -133,6 +133,9 @@ export class OpfResourceLoaderService extends ScriptLoader {
     if (html) {
       const element = new DOMParser().parseFromString(html, 'text/html');
       const script = element.getElementsByTagName('script');
+      if (!script?.[0]?.innerText) {
```
FollowTheFlo, line 136: Before, an exception was thrown when html did not have script tag which lead html snippet to not be displayed.
Html without 'script' tag is valid use case.

Matejk00 (approved):

PR #18117, opened 2023-11-16
Update i18next libraries

kpawelczak:


projects/core/src/i18n/i18next/i18next-translation.service.ts
```diff
@@ -48,7 +48,7 @@ export class I18nextTranslationService implements TranslationService {
           return;
         }
         if (this.i18next.exists(namespacedKey, options)) {
-          subscriber.next(this.i18next.t(namespacedKey, options));
+          subscriber.next(this.i18next.t(namespacedKey, options) as string);
```
Platonn: [QUESTION]
Why do we have to cast it to string now?

projects/core/src/i18n/i18next/i18next-backend/i18next-http-backend-client.spec.ts
```diff
@@ -5,7 +5,7 @@ import {
   TestRequest,
 } from '@angular/common/http/testing';
 import { TestBed } from '@angular/core/testing';
-import { RequestCallback } from 'i18next-http-backend';
+import { RequestCallback } from 'i18next-http-backend/cjs';
```
Platonn: [QUESTION]
Why can't we import types from a primary entry point anymore, but have to use /cjs secondary entry point?

projects/core/src/i18n/i18next/i18next-translation.service.ts
kpawelczak: had type error, they must have tinkered with the types, still looking at the changelog..

projects/core/src/i18n/i18next/i18next-backend/i18next-http-backend-client.spec.ts
kpawelczak: RequestCallBack type is not exported by the lib... the ts version used commonjs and they just recently changed it, that's why it was possible before

projects/core/src/i18n/i18next/i18next-translation.service.ts
Platonn: I've experimented with the type of the `options` param. When it's `any`, then the return type of the `t()` is some funky type which is not matching `string`.

If only we fix (or workaround) the type of the `options` param, the return type of the `t()` function matches `string`.

So I'd vote for working around the types closer to the root issue - which would mean in this case casting `options` to type `TOptions` from `i18next`

```diff
- import { i18n } from 'i18next';
+ import { i18n, TOptions } from 'i18next';

- ... this.i18next.t(namespacedKey, options) as string
+ ... this.i18next.t(namespacedKey, options as TOptions)
```

What do you think about it?

--- 
Note: this comment applies to all places in this file where we've casted `t()` to `as string`

package.json
```diff
@@ -125,8 +125,8 @@
     "comment-json": "^4.2.3",
     "express": "^4.18.2",
     "hamburgers": "^1.2.1",
-    "i18next": "^21.10.0",
-    "i18next-http-backend": "^1.4.5",
+    "i18next": "^23.7.6",
+    "i18next-http-backend": "^2.4.1",
     "i18next-resources-to-backend": "^1.1.2",
```
Platonn: Let's bump also `i18next-resources-to-backend`.
I can see nearly no changes in the changelog (i.e. commit history of their library) in between 1.1 and 1.2(latest), but let's bump it just to help satisfy the compliance rules of using possible latest versions.

See https://github.com/i18next/i18next-resources-to-backend/commits/main/

projects/core/src/i18n/i18next/i18next-backend/i18next-http-backend-client.spec.ts
Platonn: It looks strange at the first sight that we use some typings from `/cjs` secondary entry point of a library, while our code is compiled from and targeted to ESM.

Alternatively, I've tried importing from their `/esm` secondary entry point, but it seems that there the type `RequestCallback` is not exported. Very strange.

I I'm having hard time understanding why the usage of `RequestCallback` is discouraged in ESM context by library authors. I only guess that's an overlook from their side....

What about:
1. remove usage of explicit type `RequestCallback` at all and also revert using `/cjs` secondary entry point. Instead rely on implicit types inferred from `HttpBackendOptions["request"]` (our custom type: `export type I18nextHttpBackendClient = HttpBackendOptions['request'];`). Note: `HttpBackendOptions` is exported in `i18next-http-backend` :) See proposed change I made on my local:
    image
    Please note the types are implicitly inferred, when I hover over them with a mouse (thanks to the fact that the whole fuction is declared as of type `I18nextHttpBackendClient`: 
    image
    I motivate my proposal this way: it _looks just strange_ that in our ESM code we have to import some types from a `/cjs` secondary entry point of a library, while IMHO it's not justified at all.

2. Additionally to the point (1) I suggest to [create a bug ticket in their repo](https://github.com/i18next/i18next-http-backend/issues/new?assignees=&labels=&projects=&template=bug.md), asking why certain types are not exported in the main entry point, e.g. `RequestCallback`, but only in the secondary entry point `/cjs`. At the first sight it looks like an overlook on their side. Then IMHO it would be nice to help the authors of the library realize they have a potential flaw to fix

kpawelczak: I don't see it being exported in CJS either. I assume because it's a different module system, it works somehow. 
I tried your change and it breaks the building process. Typescript tries to import the type anyway:
```typescript
export declare const I18NEXT_HTTP_BACKEND_CLIENT: InjectionToken<
  | ((
      options: import('i18next-http-backend').HttpBackendOptions,
      url: string,
      payload: string | {},
      callback: import('i18next-http-backend').RequestCallback
    ) => void)
  | undefined
>;
```

Platonn: Thanks for checking it! Ok, let's go with your solution then (`/cjs`), if it's the only one working.

And in the meantime I suggest to create a ticket in **their repo** to sort this situation out, explaining the problem, workarounds and asking for elaboration from them - what is the solution they would recommend or whether they think they should export more types in their public API. 

Maybe they overlooked exporting those types? Or maybe we're missing something important and we'll learn from their answer.
Do you think you could create a ticket in their repo?

kpawelczak: sure I can create a ticket, but I doubt they fix that.

Platonn (dismissed):

Platonn (approved):

PR #18118, opened 2023-11-16
fix: Remove error from confirmation shipping component, and error from normalize error util

rmch91:
Partially closes https://jira.tools.sap/browse/CXSPA-5225. Fixed errors from normalizeHttpError by providing LoggerService


feature-libs/order/core/store/effects/order-details.effect.spec.ts
```diff
@@ -40,6 +41,14 @@ class MockUserIdService implements Partial<UserIdService> {
   }
 }
 
+class MockLoggerService {
```
FollowTheFlo: as discussed, this mock class could be imported to avoid repetition. 
suggestion to put it in  cypress/helpers folder, new file logger.ts

rmch91: Changed. But I created file with the mock class just near the logger service. I think its better than to put it in cypress, because it is related to e2e tests, and this mock is needed only on unit tests level. Also it is required in different libs, so seems fine for me that it is in core

FollowTheFlo (approved):
good to me

PR #18126, opened 2023-11-17
fix: OPF missing Card Type (CXSPA-5221)

kimhw0630:
Closing: https://jira.tools.sap/browse/CXSPA-5221


FollowTheFlo (approved):
good to me

PR #18147, opened 2023-11-23
Add a11y regions

Larisa-Staroverova:


feature-libs/quote/assets/translations/en/quote.i18n.ts
```diff
@@ -141,6 +141,7 @@ export const quote = {
       },
     },
     comments: {
+      regionTitle: 'Comment box',
```
steinsebastian: According to wireframe: shouldn't it be "Contact/Comment Box"?

steinsebastian (changes requested):

steinsebastian (approved):

PR #18156, opened 2023-11-24
QuickBuy: ApplePay and GooglePay

FollowTheFlo:
Apple Pay + GooglePay


integration-libs/opf/base/components/opf-quick-buy/apple-pay/apple-pay.component.ts
```diff
@@ -0,0 +1,71 @@
...
+      .pipe(
+        switchMap(([product, _]) =>
+          this.applePayService.start(
+            product as Product,
+            this.itemCounterService.getCounter(),
+            this.activeConfiguration.acquirerCountryCode
+          )
+        )
+      )
+      .subscribe(() => {});
```
Matejk00: Should not to be sufficent to use `.subscribe()` as we have an empty block?

integration-libs/opf/base/components/opf-quick-buy/apple-pay/apple-pay.component.ts
```diff
@@ -0,0 +1,71 @@
...
+  protected cartHandlerService = inject(OpfCartHandlerService);
+  protected applePaySession = inject(ApplePaySessionFactory);
+
+  protected sub: Subscription;
+  isApplePaySupported$: Observable<boolean>;
+
+  ngOnInit(): void {
+    const merchantId =
+      this.activeConfiguration?.digitalWalletQuickBuy?.find(
+        (dw) => dw.merchantId === OpfProviderType.APPLE_PAY
```
Matejk00: What about:
```(digitalWallet: DigitalWalletQuickBuy) => digitalWallet.merchantId === OpfProviderType.APPLE_PAY```
It's a nitpick, but I prefer to use full names for that kind of properties. I think it makes code more readable.

File: integration-libs/opf/base/components/opf-quick-buy/apple-pay/apple-pay.service.ts [hunk 3]
```diff
@@ -0,0 +1,354 @@
...
+  };
+
+  protected initLocalCart(product: Product, quantity: number): LocalCart {
+    return {
+      quantity,
+      product,
+      addressIds: [],
+      isPdp: true,
+      total: {
+        amount: '' + (product.price?.value as number) * quantity,
```
Matejk00: Why do we need adding `''`? It is because we need conversion to the string?

integration-libs/opf/base/components/opf-quick-buy/apple-pay/apple-pay.service.ts
```diff
@@ -0,0 +1,354 @@
...
+        });
+        result.newShippingMethods = newShippingMethods;
+        return result;
+      }),
+      switchMap(() => {
+        return this.cartHandlerService.getCurrentCartTotalPrice();
+      }),
+      switchMap((price) => {
+        if (price) {
+          this.localCart.total.amount = '' + price;
```
Matejk00: I think is more semantic to use `toString()` method, instead of adding an empty string.

integration-libs/opf/base/components/opf-quick-buy/apple-pay/apple-pay.service.ts
```diff
@@ -0,0 +1,354 @@
...
+      })
+    );
+  }
+
+  protected convertAppleToOpfAddress(
+    addr: ApplePayJS.ApplePayPaymentContact,
+    partial = false
+  ): Address {
+    return {
+      firstName: partial ? 'xxxx' : addr?.givenName,
```
Matejk00: For google pay I created some const for such placeholder data:
```
    const ADDRESS_FIELD_PLACEHOLDER = '[FIELD_NOT_SET]';
```

then I can re-use it each time is needed. Based on that I can also recognise any address at the end if contains such placeholder. If so, it can be easily removed.

integration-libs/opf/base/components/opf-quick-buy/apple-pay/apple-pay.service.ts
FollowTheFlo: yes, using toString() would do the same

projects/storefrontapp/src/.well-known/apple-developer-merchantid-domain-association.txt
```diff
@@ -0,0 +1,74 @@
+MIIQdQYJKoZIhvcNAQcCoIIQZjCCEGICAQExCzAJBgUrDgMCGgUAMIGDBgkqhkiG9w0BBwGgdgR0
```
Matejk00: Not sure how the apple pay works, but this seems to be some hashed value. Is it the file that we should keep in our repo, or in other words is it any kind of sensitive data?

FollowTheFlo: yep, mistake. this file must be added by the admin.

PR #18167, opened 2023-11-27
Feature/cxspa 5177 aoc directive

ChristophHi:


feature-libs/quote/components/items/quote-items.component.ts
```diff
@@ -42,29 +49,66 @@ export class QuoteItemsComponent {
...
-        });
-      }
-    }),
-    switchMap(([quote, _userId]) => {
-      return quote.cartId !== undefined
-        ? this.multiCartFacade.getCart(quote.cartId)
-        : NEVER;
-    })
-  );
+  quoteItemsData$: Observable<QuoteItemsData> = this.userIdService
```
Platonn: [MINOR] Extensibility

It's quite a long stream. I only guess that some customers might want to customize parts of it.
Perhaps it's worth moving it (or parts of it) to separate function(s) in a component service?

feature-libs/cart/base/components/abstract-order/abstract-order.directive.ts
```diff
@@ -0,0 +1,46 @@
...
+import { AbstractOrderContext } from './abstract-order-context.model';
+
+@Directive({
+  selector: '[cxAbstractOrder]',
+  providers: [
+    AbstractOrderContextSource,
+    { provide: AbstractOrderContext, useExisting: AbstractOrderContextSource },
+  ],
+})
+export class AbstractOrderDirective implements OnChanges {
```
Platonn: [QUESTION]
`[cxAbstractOrder]=...` vs `[cxAbstractOrderContext]=...` ? I mean what about adding word `Context` at the end of directive name? Just to match the token `AbstractOrderContext` name that it provides, so it perhaps can be more obviously associated (the token with the directive), by eye, by mind, but also by a simple search-find in VSCode?

consequences renamings:
AbstractOrderContextDirective
cxAbstractOrderContext
AbstractOrderContextModule

feature-libs/cart/base/components/abstract-order/abstract-order.directive.ts
Platonn, line 1: ✅ To my knowledge /cart/base/components is a good place for it, even though other feature libraries might want to import the directive from here, and even though this entry point is meant mailny for being lazy loaded in the CART_BASE_FEATURE.

Now it's in the cart/base/components which is part of the lazy loaded JS. If we eagerly import this module from somewhere else, to my understanding Webpack will create a shared JS chunk with this directive and share it among various lazy loaded features that depend on this directive. 
Alternatively, we could have put AbstractOrderContext (and ...Source) and ...Directive in the /root entry point and eagerly load it. But IMHO it's not justified to eagerly load and bundle this code in the main.js shipped to anonymous user visiting homepage/PDP/PLP. It would only add up to worsening the kilobytes footprint and performance (web vitals).
So I think /cart/base/components is fine. Unless you have other opinion or doubts...?

feature-libs/cart/base/components/abstract-order/abstract-order.directive.ts
ChristophHi: Yes, agree, better to use more 'uniform' names here

feature-libs/quote/components/items/quote-items.component.ts
ChristophHi: Yes, agree. Moved it to the component service and also the non-intuitive parts into protected methods in that service

Larisa-Staroverova:
@ChristophHi,
because of the wrong import in **feature-libs/cart/base/components/abstract-order-context/abstract-order-context.model.ts**:
> import { AbstractOrderType } from '../../root/models/cart.model';
should be changed to 
> import { AbstractOrderType } from '@spartacus/cart/base/root';

`npm run build:libs` ends up with the following issue:
image

feature-libs/cart/base/components/abstract-order-context/abstract-order-context.directive.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+
+  it('should propagate abstract order ID to inner component', () => {
+    expect(fixture.nativeElement.innerHTML).toContain(abstractOrderId);
+  });
+
+  it('should propagate abstract order type to inner component', () => {
+    expect(fixture.nativeElement.innerHTML).toContain(AbstractOrderType.CART);
+  });
+
+  it('should should emit changes only if actual value has changed for id', () => {
```
Larisa-Staroverova: Please, remove one `should`

feature-libs/cart/base/components/abstract-order-context/abstract-order-context.directive.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+    tap(() => (emissionCounterId = emissionCounterId + 1))
+  );
+  myType$ = this.abstractOrderContext?.type$.pipe(
+    tap(() => (emissionCounterType = emissionCounterType + 1))
+  );
+}
+
+describe('AbstractOrderContextDirective', () => {
+  let fixture: ComponentFixture<TestOuterComponent>;
+  let testOuterComponent: TestOuterComponent;
```
Larisa-Staroverova: I would suggest to rename `testOuterComponent` to `component`

feature-libs/product-configurator/common/components/configure-cart-entry/configure-cart-entry.component.ts
```diff
@@ -46,106 +57,97 @@ export class ConfigureCartEntryComponent {
...
-      return CommonConfigurator.OwnerType.SAVED_CART_ENTRY;
-    } else {
-      return CommonConfigurator.OwnerType.CART_ENTRY;
-    }
-  }
-
-  protected getEntityKeyInternal(readOnlyCartIsActive: boolean): string {
+  retrieveEntityKey(abstractOrderData: AbstractOrderKey): string {
+    const orderType = abstractOrderData.type;
+    const ownerDocumentIdNeeded: boolean =
```
Larisa-Staroverova: what about to rename `ownerDocumentIdNeeded` to `ownerIdRequired`?

feature-libs/product-configurator/common/components/configure-cart-entry/configure-cart-entry.component.ts
```diff
@@ -46,106 +57,97 @@ export class ConfigureCartEntryComponent {
...
-    if (this.cartEntry.orderCode) {
-      code = this.cartEntry.orderCode;
-    } else if (this.readOnly && this.cartEntry.quoteCode) {
-      code = this.cartEntry.quoteCode;
-    } else if (this.readOnly && !readOnlyCartIsActive) {
-      code = this.cartEntry.savedCartCode;
-    }
-    return code
-      ? this.commonConfigUtilsService.getComposedOwnerId(code, entryNumber)
+    return ownerDocumentIdNeeded
```
Larisa-Staroverova: If `ownerDocumentIdNeeded` is renamed to `ownerIdRequired`, then it will be clearer when getConfiguratorOwnerId() method is called in case `ownerIdRequired is **true**

Larisa-Staroverova: If `ownerDocumentIdNeeded` is renamed to `ownerIdRequired`, then it will be clearer when getConfiguratorOwnerId() method is called in case `ownerIdRequired is **true**

feature-libs/product-configurator/common/components/configure-cart-entry/configure-cart-entry.component.ts
```diff
@@ -46,106 +57,97 @@ export class ConfigureCartEntryComponent {
...
-        return this.cartEntry.orderCode;
-      } else if (this.cartEntry.quoteCode) {
-        return this.cartEntry.quoteCode;
-      } else {
-        return this.cartEntry.savedCartCode;
-      }
-    } else {
-      return undefined;
+  protected getConfiguratorOwnerId(
+    abstractOrderData: AbstractOrderKey,
```
Larisa-Staroverova, line 138: What about to rename  `abstractOrderData` to  `abstractOrderKey`?

feature-libs/quote/components/items/quote-items.component.service.spec.ts
```diff
@@ -1,14 +1,124 @@
...
+  isEditable: false,
+  entries: [{ entryNumber: 1 }],
+};
+
+const quoteEditable: Quote = {
+  ...quote,
+  isEditable: true,
+};
+
+const quoteWoCartId: Quote = {
```
Larisa-Staroverova, line 43: `Wo` stands for?

feature-libs/quote/components/items/quote-items.component.service.ts
```diff
@@ -4,24 +4,136 @@
...
+      of({ readOnly: true }),
+      of(AbstractOrderType.SAVED_CART),
+    ]);
+  }
+
+  protected prepareQuote(quote: Quote): QuoteOrCartWithEditState {
+    return zip(of(quote), of({ readOnly: true }), of(AbstractOrderType.QUOTE));
+  }
+
+  protected loadQuoteCartIfNeeded(quote: Quote, userId: string) {
```
Larisa-Staroverova, line 130: I would suggest to rename the method to `loadQuoteCart()`.
According to my point of view it is not required to append  `IfNeed` at the end of the method.

Larisa-Staroverova (changes requested):
@ChristophHi, could you, please, have a look at my comments?

feature-libs/cart/base/components/abstract-order-context/abstract-order-context.directive.spec.ts
ChristophHi: testComponent, ok?

feature-libs/product-configurator/common/components/configure-cart-entry/configure-cart-entry.component.ts
ChristophHi: Good point, but I rather would go for 'documentIdRequired' in order to not confuse it withe the configuration owner (specific for this component that deals with the configuration)

feature-libs/quote/components/items/quote-items.component.service.spec.ts
ChristophHi: 'without'

feature-libs/quote/components/items/quote-items.component.service.ts
ChristophHi: But the method first checks if it's needed to load the cart, why should that _not_ be reflected in its name?

Larisa-Staroverova: The comment could be added if the method name is not self-explanatory

Larisa-Staroverova (approved):

PR #18176, opened 2023-11-28
Save active cart when starting to edit a quote

steinsebastian:


feature-libs/quote/core/facade/quote.service.spec.ts
```diff
@@ -160,6 +165,7 @@ describe('QuoteService', () => {
   let cartUtilsService: CartUtilsService;
   let globalMessageService: GlobalMessageService;
   let quoteStorefrontUtilsService: QuoteStorefrontUtilsService;
+  let savedCartService: SavedCartFacade;
```
Larisa-Staroverova: Please, rename to `savedcartFacade`

feature-libs/quote/core/facade/quote.service.ts
```diff
@@ -75,6 +76,7 @@ export class QuoteService implements QuoteFacade {
   protected cartUtilsService = inject(CartUtilsService);
   protected globalMessageService = inject(GlobalMessageService);
   protected quoteStorefrontUtilsService = inject(QuoteStorefrontUtilsService);
+  protected savedCartService = inject(SavedCartFacade);
```
Larisa-Staroverova: Please, rename to `savedCartFacade`

Larisa-Staroverova (changes requested):

feature-libs/quote/core/facade/quote.service.spec.ts
steinsebastian: Done

feature-libs/quote/core/facade/quote.service.ts
steinsebastian: Done

Larisa-Staroverova (approved):

PR #18197, opened 2023-12-04
feat: CXSPA-5517 Upgrade to Angular 17 - SSR adjustment

pawelfras:
- [x] - added ngExpressEngine with unit tests
- [x] - added REQUEST and RESPONSE injection tokens
- [x] - removed 'jest-preset-angular/setup-jest' from 'setup-jest.ts' for 'setup' library to prevent NG0400 error in express engine tests. See: https://github.com/angular/angular/blob/d4b423690210872b5c32a322a6090beda30b05a3/packages/core/src/application_ref.ts#L353
- [x] - added 'TestBed.initTestEnvironment' to test files where needed
- [x] - bumped packages in 'setup' library

closes [CXSPA-5517](https://jira.tools.sap/browse/CXSPA-5517)

### DIFF
--- a/core-libs/setup/ssr/engine-decorator/cx-express-engine-decorator.ts
+++ b/core-libs/setup/ssr/engine-decorator/cx-express-engine-decorator.ts
@@ -4,7 +4,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { NgSetupOptions } from '@nguniversal/express-engine';
+import { CommonEngineOptions, CommonEngineRenderOptions } from '@angular/ssr';
 import {
   OptimizedSsrEngine,
   SsrCallbackFn,
@@ -15,43 +15,45 @@ import {
 } from '../optimized-engine/ssr-optimization-options';
 import { getServerRequestProviders } from '../providers/ssr-providers';
 
-export type NgExpressEngineInstance = (
+export type CxExpressEngineInstance = (
   filePath: string,
   options: object,
   callback: SsrCallbackFn
 ) => void;
 
-export type NgExpressEngine = (
-  setupOptions: Readonly<NgSetupOptions>
-) => NgExpressEngineInstance;
+export type CxExpressEngine = (
+  setupOptions: Readonly<CommonEngineRenderOptions & CommonEngineOptions>
+) => CxExpressEngineInstance;
 
 /**
  * The wrapper over the standard ngExpressEngine, that provides tokens for Spartacus
- * @param ngExpressEngine
+ * @param cxExpressEngine
  */
-export class NgExpressEngineDecorator {
+export class CxExpressEngineDecorator {
   /**
    * Returns the higher order ngExpressEngine with provided tokens for Spartacus
    *
-   * @param ngExpressEngine
+   * @param cxExpressEngine
    */
   static get(
-    ngExpressEngine: NgExpressEngine,
+    ngExpressEngine: CxExpressEngine,
     optimizationOptions?: SsrOptimizationOptions | null
-  ): NgExpressEngine {
+  ): CxExpressEngine {
     return decorateExpressEngine(ngExpressEngine, optimizationOptions);
   }
 }
 
 export function decorateExpressEngine(
-  ngExpressEngine: NgExpressEngine,
+  cxExpressEngine: CxExpressEngine,
   optimizationOptions:
     | SsrOptimizationOptions
     | null
     | undefined = defaultSsrOptimizationOptions
-): NgExpressEngine {
-  return function (setupOptions: NgSetupOptions) {
-    const engineInstance = ngExpressEngine({
+): CxExpressEngine {
+  return function (
+    setupOptions: CommonEngineRenderOptions & CommonEngineOptions
+  ) {
+    const engineInstance = cxExpressEngine({
       ...setupOptions,
       providers: [
         // add spartacus related providers
--- a/core-libs/setup/ssr/engine/cx-express-engine.spec.ts
+++ b/core-libs/setup/ssr/engine/cx-express-engine.spec.ts
@@ -1,0 +1,187 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+// eslint-disable-next-line import/no-unassigned-import
+import '@angular/compiler';
+
+import { Component, Inject, InjectionToken, NgModule } from '@angular/core';
+import { BrowserModule } from '@angular/platform-browser';
+import { ServerModule } from '@angular/platform-server';
+import { REQUEST, RESPONSE } from '../public_api';
+import { cxExpressEngine } from './cx-express-engine';
+//@ts-ignore
+
+/**
+ * @license
+ * The MIT License
+ * Copyright (c) 2010-2023 Google LLC. http://angular.io/license
+ *
+ * Inspired by tests for ngExpressEngine.
+ *
+ * See:
+ * - https://github.com/angular/universal/blob/e798d256de5e4377b704e63d993dc56ea35df97d/modules/express-engine/spec/index.spec.ts
+ * - https://github.com/angular/universal/blob/e798d256de5e4377b704e63d993dc56ea35df97d/modules/express-engine/spec/mock.server.module.ts
+ *
+ */
+@Component({ selector: 'cx-mock', template: 'some template' })
+export class MockComponent {}
+
+@NgModule({
+  imports: [BrowserModule, ServerModule],
+  declarations: [MockComponent],
+  bootstrap: [MockComponent],
+})
+export class MockServerModule {}
+
+@Component({ selector: 'cx-request', template: `url:{{ _req.url }}` })
+export class RequestComponent {
+  constructor(@Inject(REQUEST) public readonly _req: any) {}
+}
+
+@NgModule({
+  imports: [BrowserModule, ServerModule],
+  declarations: [RequestComponent],
+  bootstrap: [RequestComponent],
+})
+export class RequestServerModule {}
+
+@Component({
+  selector: 'cx-response',
+  template: `statusCode:{{ _res.statusCode }}`,
+})
+export class ResponseComponent {
+  constructor(@Inject(RESPONSE) public readonly _res: any) {}
+}
+
+@NgModule({
+  imports: [BrowserModule, ServerModule],
+  declarations: [ResponseComponent],
+  bootstrap: [ResponseComponent],
+})
+export class ResponseServerModule {}
+
+export const SOME_TOKEN = new InjectionToken<string>('SOME_TOKEN');
+
+@Component({
+  selector: 'cx-token',
+  template: `message:{{ _someToken.message }}`,
+})
+export class TokenComponent {
+  constructor(@Inject(SOME_TOKEN) public readonly _someToken: any) {}
+}
+
+@NgModule({
+  imports: [BrowserModule, ServerModule],
+  declarations: [TokenComponent],
+  bootstrap: [TokenComponent],
+})
+export class TokenServerModule {}
+
+describe('test runner', () => {
+  beforeAll(() => {
+    jest.spyOn(console, 'log').mockImplementation();
+  });
+
+  it('should render a basic template', (done) => {
+    cxExpressEngine({ bootstrap: MockServerModule })(
+      null as any as string,
+      {
+        req: { get: () => 'localhost' } as any,
+        document: '<cx-mock></cx-mock>',
+      },
+      (err, html) => {
+        if (err) {
+          throw err;
+        }
+        expect(html).toContain('some template');
+        done();
+      }
+    );
+  });
+
+  it('Should throw when no module is passed', () => {
+    cxExpressEngine({ bootstrap: null as any })(
+      null as any as string,
+      {
+        req: {} as any,
+        bootstrap: null as any,
+        document: '<cx-mock></cx-mock>',
+      },
+      (_err, _html) => {
+        expect(_err).toBeTruthy();
+      }
+    );
+  });
+
+  it('should be able to inject REQUEST token', (done) => {
+    cxExpressEngine({ bootstrap: RequestServerModule })(
+      null as any as string,
+      {
+        req: {
+          get: () => 'localhost',
+          url: 'http://localhost:4200',
+        } as any,
+        document: '<cx-request></cx-request>',
+      },
+      (err, html) => {
+        if (err) {
+          throw err;
+        }
+        expect(html).toContain('url:http://localhost:4200');
+        done();
+      }
+    );
+  });
+
+  it('should be able to inject RESPONSE token', (done) => {
+    const someStatusCode = 400;
+    cxExpressEngine({ bootstrap: ResponseServerModule })(
+      null as any as string,
+      {
+        req: {
+          get: () => 'localhost',
+          res: {
+            statusCode: someStatusCode,
+          },
+        } as any,
+        document: '<cx-response></cx-response>',
+      },
+      (err, html) => {
+        if (err) {
+          throw err;
+        }
+        expect(html).toContain(`statusCode:${someStatusCode}`);
+        done();
+      }
+    );
+  });
+
+  it('should be able to inject some token', (done) => {
+    const someValue = { message: 'value' + new Date() };
+    cxExpressEngine({
+      bootstrap: TokenServerModule,
+      providers: [{ provide: SOME_TOKEN, useValue: someValue }],
+    })(
+      null as any as string,
+      {
+        req: {
+          get: () => 'localhost',
+          url: 'http://localhost:4200',
+        } as any,
+        document: '<cx-token></cx-token>',
+      },
+      (err, html) => {
+        if (err) {
+          throw err;
+        }
+        expect(html).toContain(someValue.message);
+        done();
+      }
+    );
+  });
+});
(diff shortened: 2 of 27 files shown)


projects/storefrontapp/server.ts
```diff
@@ -29,7 +29,7 @@ const ssrOptions: SsrOptimizationOptions = {
   logger: true,
 };
 
-const ngExpressEngine = NgExpressEngineDecorator.get(engine, ssrOptions);
+const cxExpressEngine = CxExpressEngineDecorator.get(engine, ssrOptions);
```
Platonn: [QUESTION]
Could you help me understand the reason behind changing the types to `cxExpressEngine`? This seems to require more migration effort since we have to change both the import path and the item name, whereas we could change only import paths.

projects/storefrontapp/server.ts
```diff
@@ -5,11 +5,11 @@
  */
 
 import { APP_BASE_HREF } from '@angular/common';
-import { ngExpressEngine as engine } from '@nguniversal/express-engine';
 import {
-  NgExpressEngineDecorator,
+  CxExpressEngineDecorator,
   SsrOptimizationOptions,
   defaultSsrOptimizationOptions,
+  cxExpressEngine as engine,
```
Platonn: [MAJOR]
I guess we longer need the `ngExpressEngine` to be passed to some other function that decorates it, because now we control and create the `ngExpressEngine`. The decorator was meant to wrap the original Angular's `ngExpressEngine` that we don't create.

That said, removing it would be yet another migration to provide.

Example, but let's not do it now, only if we have more time:
```ts
import { createNgExpressEngine } from '@spartacus/setup/ssr;

const ngExpressEngine = createNgExpressEngine(ssrOptions};
```

projects/storefrontapp/server.ts
```diff
@@ -44,7 +44,7 @@ export function app() {
   // Our Universal express-engine (found @ https://github.com/angular/universal/tree/master/modules/express-engine)
   server.engine(
     'html',
-    ngExpressEngine({
+    cxExpressEngine({
       bootstrap: AppServerModule,
     })
```
Platonn: [QUESTION]
Do you know by any chance, if our `cxExpressEngine` is compatible both with NgModule's app and 'standalone' app? 
In this example I can see`cxExpressEngine` is accepting `AppServerModule`. But from a fresh ng17 standalone app what is imported from the `main.server.ts` file is only  the `bootstrap()` function. See:

---

image
And then `bootstrap` is passed to the `CommonEngine.render()` function:

---

image

Any ideas, how (if?) we support it in our ngExpressEngine?

core-libs/setup/ssr/engine/cx-express-engine.spec.ts
```diff
@@ -0,0 +1,187 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
```
Platonn: [NITPICK] Do we really have to put license header twice in this file

core-libs/setup/ssr/engine/cx-express-engine.ts
```diff
@@ -0,0 +1,91 @@
...
+export type CxSetupOptions = Readonly<
+  CommonEngineRenderOptions & CommonEngineOptions
+>;
+
+/*
+ * Copyright (c) 2010-2023 Google LLC. http://angular.io/license
+ * SPDX-FileCopyrightText: 2023 SAP Spartacus team <spartacus-team@sap.com>
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
```
Platonn: [MAJOR]
IMHO we should start the file with the correct SAP's Apache license text, and if there's a specific section for Angular, it should be clearly marked and attributed to Google later on. 
Now this file starts with a SAP-looking header which contains Google company in it.

Note: the same remark applies to also other places in this PR

core-libs/setup/ssr/logger/services/express-logger.service.spec.ts
```diff
@@ -19,6 +23,14 @@ describe('ExpressLoggerService', () => {
   let logger: ExpressServerLogger;
   let loggerService: ExpressLoggerService;
 
+  beforeAll(() => {
+    TestBed.initTestEnvironment(
+      BrowserDynamicTestingModule,
+      platformBrowserDynamicTesting(),
+      {}
+    );
+  });
```
Platonn: [QUESTION]
Could you help me understand the reason for adding `BrowserDynamicTestingModule` and `platformBrowserDynamicTesting()` in the unit test setup for SSR?

I have probvlem to undestand why we include browser-related utils, while the code-under-test is meant only to run in Node (SSR). Also the tests deliberately run in Node (jest).

core-libs/setup/ssr/tokens/express.tokens.ts
```diff
@@ -0,0 +1,15 @@
...
+
+import { InjectionToken } from '@angular/core';
+import { Request, Response } from 'express';
+
+export const REQUEST: InjectionToken<Request> = new InjectionToken<Request>(
+  'REQUEST'
+);
+export const RESPONSE: InjectionToken<Response> = new InjectionToken<Response>(
+  'RESPONSE'
+);
```
Platonn: [NITPICK]
Let's add some JSDocs for those tokens

projects/storefrontapp/server.ts
Platonn: Looking at [the new Angular interface types](https://github.com/angular/angular-cli/blob/8914ef0a01e1d7673fdc5c25ec2bb0ed43cd3b3c/packages/angular/ssr/src/common-engine.ts#L23-L34), I can deduce that the `bootstrap` config property can accept both:
- `Type<any>` - which looks like an NgModule class reference OR
- `() => Promise<ApplicationRef>` - function that promises to return ApplicationRef - which is exactly the type of the new 'standalone' `bootstrap()` function exported from `main.server.ts` in fresh ng17 apps:
    image

core-libs/setup/ssr/engine-decorator/cx-express-engine-decorator.ts
```diff
@@ -15,43 +15,45 @@ import {
...
   filePath: string,
   options: object,
   callback: SsrCallbackFn
 ) => void;
 
-export type NgExpressEngine = (
-  setupOptions: Readonly<NgSetupOptions>
-) => NgExpressEngineInstance;
+export type CxExpressEngine = (
+  setupOptions: Readonly<CommonEngineRenderOptions & CommonEngineOptions>
```
Platonn: [QUESTION]
Why do we mix `CommonEngineRenderOptions` with `CommonEngineOptions`, being our new setup options to initialize our new `NgExpressEngine`?

`CommonEngineRenderOptions` and  `CommonEngineOptions` are semantically different from each other.
- `CommonEngineOptions` are used to _create and initialize_ `CommonEngine` (with `new CommonEngine(CommonEngineOptions)`
- `CommonEngineRenderOptions` are used _only when rendering_ with the method `render()` of `CommonEngine`

Are you sure we want to blend those 2 interfaces together? (they might have some potential conflicts of properties types in the future)

core-libs/setup/ssr/engine/cx-express-engine.ts
```diff
@@ -0,0 +1,91 @@
...
+}
+
+export function cxExpressEngine(
+  setupOptions: Readonly<CommonEngineRenderOptions & CommonEngineOptions>
+) {
+  const engine = new CommonEngine({
+    bootstrap: setupOptions.bootstrap,
+    providers: setupOptions.providers,
+    enablePerformanceProfiler: setupOptions.enablePerformanceProfiler,
+  });
```
Platonn: [MINOR]
Do you think it would it be more beneficial to use a spread operator to pass all props? I believe it would be more future-proof in case Angular adds or removes something?

projects/storefrontapp/server.ts
pawelfras: I thought it may be more efficient to rename it `cxExpressEngine` since it's no longer part of Angular code and we used to add `cx` prefix for some of our creations. I don't mind to revert this change.

pawelfras: Related issue: https://github.com/SAP/spartacus/issues/18203

projects/storefrontapp/server.ts
pawelfras: related issue: https://github.com/SAP/spartacus/issues/18203

projects/storefrontapp/server.ts
pawelfras: I've tested it and it should work. If new `CommonEngine` and related interfaces support both `bootstrap()` and `NgModule`, so do we.

core-libs/setup/ssr/engine/cx-express-engine.spec.ts
pawelfras: That's the mistake. We should put our license header here.

core-libs/setup/ssr/engine/cx-express-engine.ts
pawelfras: I wasn't sure how to highlight which part was inspired by Angular's code. I'll add the license annotation for each part separately.

core-libs/setup/ssr/logger/services/express-logger.service.spec.ts
pawelfras: Such a config is provided by `jest-preset-angular/setup-jest` [(source code)](https://github.com/thymikee/jest-preset-angular/blob/85896348ca208f069a4e34c9b36f1b4b8b50ac7e/setup-jest.js) which we're using in our app for Jest. My assumption is each Jest test that uses Angular's  API e.g. DI needs such a configuration, otherwise, tests throw errors.

Before my changes, `jest-preset-angular/setup-jest` was imported in the `setup-jest.ts` file and it was set up before all tests in the `ssr/setup` library. Because the `TestBed` config mentioned above creates a new platform, it caused the issues with tests for `ngExpressEngine`:
`A platform with a different configuration has been created. Please destroy it first`

As the platform was already there (due to the mentioned config). To prevent such an error, I removed `TestBed.initTestEnvironment` from the global config for the whole library and added it to the specific tests where it was needed.

We can investigate in the future if there's another way to configure tests for the `ssr/setup` library, but for now, I haven't found another option and I believe the most important is to have working tests.

core-libs/setup/ssr/tokens/express.tokens.ts
pawelfras: Agree

core-libs/setup/ssr/engine-decorator/cx-express-engine-decorator.ts
pawelfras: In `ngExpressEngine` we need params from both interfaces during the instantiation, but indeed, mixing all of them might be undesirable. We can do it similarly to how `@nguniversa` did it:
```ts
export type CxSetupOptions = 
  Pick<CommonEngineRenderOptions, 'providers' | 'publicPath' | 'inlineCriticalCss'> 
& CommonEngineOptions;
```

core-libs/setup/ssr/engine/cx-express-engine.ts
pawelfras: I'm not sure whether we should do it. `setupOptions` contains parameters that are not used for `CommonEngine` instantiation.

projects/storefrontapp/server.ts
Platonn: Yes, let's revert it, please. The only reason we want to add our won NgExpressEngine i backward compatibility with the rest of our tools (e.g. OptimizedSsrEngine). Let's not broaden the amount of breaking changes to be migrated, unless really justified

core-libs/setup/ssr/logger/services/express-logger.service.spec.ts
Platonn: Agreed, not a priority, Let's create a ticket for later.

I have a gut feeling, the root cause of a problem is somewhere else and adding `platformBrowserDynamicTesting()` is just a dirty workaround, which looks strange at the first sight. Maybe I'm missing something, but I feel that browser-utils shouldn't be needed in testing a NodeJS program.

---
You said:
> To prevent such an error, I removed TestBed.initTestEnvironment from the global config for the whole library and added it to the specific tests where it was needed.

In this PR I have problem to find any removal of `TestBed.initTestEnvironment` 🤔 Did you do it in a different PR or meant to do it in this one?

pawelfras: I wouldn't call it a dirty workaround (or it's a workaround but we have had it for a while) - I checked which part of `jest-preset-angular/setup-jest` is causing the issue (by commenting code of this package in node_modules) and `TestBed.initTestEnvironment`  and this is something we need (and we have it in other places).

Related issue: https://github.com/SAP/spartacus/issues/18205

core-libs/setup/ssr/engine/cx-express-engine.spec.ts
```diff
@@ -0,0 +1,187 @@
...
+
+// eslint-disable-next-line import/no-unassigned-import
+import '@angular/compiler';
```
Platonn: [QUESTION]
Why do we need to import it here?

pawelfras: It's something I took from the `@nguniversal` and it's needed due to render logic which runs under the hood of CommonEngine. Without such an import, the following error occurs:
```
The injectable '_PlatformLocation' needs to be compiled using the JIT compiler, but '@angular/compiler' is not available.
```

core-libs/setup/ssr/logger/services/express-logger.service.spec.ts
pawelfras: > You said:
> 
> > To prevent such an error, I removed TestBed.initTestEnvironment from the global config for the whole library and added it to the specific tests where it was needed.
> 
> In this PR I have a problem finding any removal of `TestBed.initTestEnvironment` 🤔 Did you do it in a different PR or meant to do it in this one?

You won't find the removal of `TestBed.initTestEnvironment` but you can find the removal of `jest-preset-angular/setup-jest` from `setup-jest.ts` - `TestBed` init is done under the hood of this module.

Platonn: What about using `ServerTestingModule` and `platformServerTesting` instead in all spec.ts files in `setup/ssr` library? IMHO they seem to be more appropriate in the context of server/NodeJS program e.g.

```ts
import {
  platformServerTesting,
  ServerTestingModule,
} from '@angular/platform-server/testing';

/* ... */

  beforeAll(() => {
    TestBed.initTestEnvironment(
      ServerTestingModule,
      platformServerTesting(),
      {}
    );
  });
```

---

Note: The internal providers are different in between `platformBrowserDynamicTesting()` vs `platformServerTesting()`. For example `BrowserDomAdapter` vs `DominoAdapter` is used under the hood, depending on which testing platform you use:

### platform browser testing
image

https://github.com/angular/angular/blob/1e3bcfecbab260fb66a1ac50e77192bb98cc0f3e/packages/platform-browser/src/browser.ts#L159-L160

https://github.com/angular/angular/blob/1e3bcfecbab260fb66a1ac50e77192bb98cc0f3e/packages/platform-browser-dynamic/testing/src/testing.ts#L22-L24

### platform server testing
image

https://github.com/angular/angular/blob/1e3bcfecbab260fb66a1ac50e77192bb98cc0f3e/packages/platform-server/testing/src/server.ts#L20-L21

https://github.com/angular/angular/blob/1e3bcfecbab260fb66a1ac50e77192bb98cc0f3e/packages/platform-server/src/server.ts#L26

pawelfras: Nice finding! Additionally, `platformServerTesting` doesn't cause platform-related errors in unit tests for `ng-express-engine` so config initialization can be kept in `setup-jest.ts` for the whole library. What a simplification 👌

core-libs/setup/ssr/engine/ng-express-engine.ts
```diff
@@ -0,0 +1,122 @@
...
+ * See:
+ * - https://github.com/angular/universal/blob/e798d256de5e4377b704e63d993dc56ea35df97d/modules/express-engine/src/main.ts
+ */
+export interface RenderOptions extends CommonEngineRenderOptions {
+  req: Request;
+  res?: Response;
+}
+
+/**
+ * @license
```
Platonn, line 75: Before the @license, let's add to this JSDoc information for us and for customers - why we need it. That it was present in Angular for long time and it got removed only in version Angular 17. But we neeed it in Spartacus, so we copied their code to our repo for backward compatibility to avoid bigger refactors in Spartacus

core-libs/setup/ssr/tokens/express.tokens.ts
```diff
@@ -0,0 +1,26 @@
...
+
+/**
+ * Represents an injection token for the response object in server-side rendering.
+ * This token is used to provide the response object to components and services.
+ * It's a replacement for the `RESPONSE` token that has been removed from Angular's source code.
+ */
+export const RESPONSE: InjectionToken<Response> = new InjectionToken<Response>(
+  'RESPONSE'
+);
```
Platonn:
```suggestion
/**
 * Represents an injection token for the Request object of ExpressJS in server-side rendering.
 * This token is used to provide the request object to components and services.
 * It's a replacement for the `REQUEST` token from `@nguniversal/express-engine` that was removed only in Angular 17. Now this token is provided by Spartacus.
 */
export const REQUEST: InjectionToken<Request> = new InjectionToken<Request>(
  'REQUEST'
);

/**
 * Represents an injection token for the Response object of ExpressJS in server-side rendering.
 * This token is used to provide the response object to components and services.
 * It's a replacement for the `RESPONSE` token from `@nguniversal/express-engine` that was removed only in Angular 17. Now this token is provided by Spartacus.
 */
export const RESPONSE: InjectionToken<Response> = new InjectionToken<Response>(
  'RESPONSE'
);
```

Platonn (approved):

PR #18211, opened 2023-12-08
Display section with action buttons when there are any buttons

Larisa-Staroverova:
Closes [CXSPA-5593](https://jira.tools.sap/browse/CXSPA-5593)


feature-libs/quote/components/summary/actions/quote-summary-actions.component.html
```diff
@@ -1,17 +1,19 @@
...
-        *ngFor="let action of quoteDetails.allowedActions"
-        class="btn"
-        [ngClass]="getButtonStyle(quoteDetails.allowedActions, action)"
-        [disabled]="mustDisableAction(action.type, quoteDetails)"
-        (click)="onClick(action.type, quoteDetails, cartDetails)"
-      >
-        {{ 'quote.summary.actions.' + action.type | cxTranslate }}
-      </button>
-    </section>
+    <ng-container *ngIf="quoteDetails.allowedActions?.length > 0">
```
ChristophHi: If that's not too much effort: Could we add a test that checks that <section> is not available in case there are no buttons?

ChristophHi (changes requested):

ChristophHi (approved):

PR #18221, opened 2023-12-08
feat: add Loader Indicator global functions

FollowTheFlo:
Added 2 global functions
startLoadIndicator
stopLoadIndicator
each wrapped in a registration method.
to test:
- go on checkout page, select a method with 'Hosted-filed pattern'
- in console,
to launch the functions
window.Opf.payments.checkout.startLoadIndicator()
window.Opf.payments.checkout.stopLoadIndicator()

### DIFF
--- a/integration-libs/opf/base/core/facade/opf-global-functions.service.ts
+++ b/integration-libs/opf/base/core/facade/opf-global-functions.service.ts
@@ -9,6 +9,7 @@ import {
   Injectable,
   NgZone,
   ViewContainerRef,
+  inject,
 } from '@angular/core';
 import { WindowRef } from '@spartacus/core';
 import {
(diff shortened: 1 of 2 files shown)


integration-libs/opf/base/core/facade/opf-global-functions.service.ts
```diff
@@ -30,12 +31,15 @@ import { finalize, take } from 'rxjs/operators';
...
 @Injectable()
 export class OpfGlobalFunctionsService implements OpfGlobalFunctionsFacade {
-  constructor(
-    protected winRef: WindowRef,
-    private ngZone: NgZone,
-    protected opfPaymentFacade: OpfPaymentFacade,
-    protected launchDialogService: LaunchDialogService
-  ) {}
+  protected winRef = inject(WindowRef);
+  protected ngZone = inject(NgZone);
```
FollowTheFlo, line 35: no reason to have ngZone scoped as 'private', so changed to 'protected'.

rmch91 (approved):

PR #18196, opened 2023-12-04
feat: Enable non-destructive hydration

Zeyber:
Closes: https://jira.tools.sap/browse/CXSPA-5524

PR #18226, opened 2023-12-11
feat: Enable non-destructive hydration skipping suppressing errors

Zeyber:
Closes: https://jira.tools.sap/browse/CXSPA-5524

Note: This PR uses research from the investigation done in #18196.


Zeyber:
There are two known issues with this PR as it stands that need to be fixed.

Major issue:
- [ ] My account dropdown menu does not appear when logged in.

Minor issue:
- [ ] Store finder service error logged in console when loading stores list page.

PR #18228, opened 2023-12-11
feat: CXSPA-5521 Upgrade to Angular 17 - schematics for SSR

pawelfras:
- [x] - adjusted schematics for ssr
- [x] - adjusted server.ts
- [x] - adjusted unit tests for 'add-ssr' and 'ng-add'
- [x] - removed a reference to "@nguniversal/express-engine" from 'constants.ts'

To run Spartacus with SSR in a fresh Angular v17 app:

- deploy Spartacus libraries locally using Verdaccio
- npx @angular/cli new my-app --standalone=false --ssr=false
- npx @angular/cli add @spartacus/schematics@latest --ssr --baseUrl="https://40.76.109.9:9002"
- in the fresh app, remove `prerender: true` from `angular.json` (temporary issue)
- run `npm run build`
- run `npm run serve:ssr:my-app`

As for now, Angular app with Spartacus are not working in development mode due to issues with new default Angular builder. As a workaround:

- run `npm watch` to run build in watch mode 
- modify `serve:ssr:my-app` script in `package.json` to:
   ```json
   "serve:ssr:my-app": "node --watch dist/my-app/server/server.mjs"
   ```
- run modified "serve:ssr:my-app" that now will watch on changes in the `dist` folder

Closes [CXSPA-5521](https://jira.tools.sap/browse/CXSPA-5521)

### DIFF
--- a/projects/schematics/src/add-ssr/__snapshots__/index_spec.ts.snap
+++ b/projects/schematics/src/add-ssr/__snapshots__/index_spec.ts.snap
@@ -60,7 +60,7 @@ const ngExpressEngine = NgExpressEngineDecorator.get(engine);
 // The Express app is exported so that it can be used by serverless Functions.
 export function app() {
   const server = express();
-  const distFolder = join(process.cwd(), 'dist/ng-17-test-ssr-module/browser');
+  const distFolder = join(process.cwd(), 'dist/schematics-test/browser');
   const indexHtml = existsSync(join(distFolder, 'index.original.html'))
     ? 'index.original.html'
     : 'index';
--- a/projects/schematics/src/add-ssr/__snapshots__/index_spec.ts.snap
+++ b/projects/schematics/src/add-ssr/__snapshots__/index_spec.ts.snap
@@ -1,6 +1,6 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`add-ssr app.server.module.ts should be updated 1`] = `
+exports[`add-ssr app.module.server.ts should be updated 1`] = `
 "import { NgModule } from '@angular/core';
 import { ServerModule } from '@angular/platform-server';
 
@@ -43,23 +43,24 @@ exports[`add-ssr index.html should contain occ-backend-base-url attribute in met
 `;
 
 exports[`add-ssr server.ts should be configured properly 1`] = `
-"import 'zone.js/dist/zone-node';
+"import { APP_BASE_HREF } from '@angular/common';
+import {
+  NgExpressEngineDecorator,
+  ngExpressEngine as engine,
+} from '@spartacus/setup/ssr';
 
-import { ngExpressEngine as engine } from '@nguniversal/express-engine';
-import { NgExpressEngineDecorator } from '@spartacus/setup/ssr';
-import * as express from 'express';
+import express from 'express';
 import { join } from 'path';
 
-import { AppServerModule } from './src/main.server';
-import { APP_BASE_HREF } from '@angular/common';
-import { existsSync } from 'fs';
+import { existsSync } from 'node:fs';
+import AppServerModule from './src/main.server';
 
 const ngExpressEngine = NgExpressEngineDecorator.get(engine);
 
 // The Express app is exported so that it can be used by serverless Functions.
 export function app() {
   const server = express();
-  const distFolder = join(process.cwd(), 'dist/schematics-test/browser');
+  const distFolder = join(process.cwd(), 'dist/ng-17-test-ssr-module/browser');
   const indexHtml = existsSync(join(distFolder, 'index.original.html'))
     ? 'index.original.html'
     : 'index';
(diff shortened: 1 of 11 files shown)


projects/schematics/src/add-ssr/__snapshots__/index_spec.ts.snap
```diff
@@ -1,6 +1,6 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`add-ssr app.server.module.ts should be updated 1`] = `
+exports[`add-ssr app.module.server.ts should be updated 1`] = `
```
Platonn, line 3: I'm surprised Angular team made this change. 😂

projects/schematics/src/add-ssr/index.ts
```diff
@@ -154,7 +154,7 @@ export function addSSR(options: SpartacusOptions): Rule {
 
     return chain([
       addPackageJsonDependencies(prepareDependencies(), packageJson),
-      externalSchematic(NGUNIVERSAL_EXPRESS_ENGINE, 'ng-add', {
+      externalSchematic(ANGULAR_SSR, 'ng-add', {
```
Platonn, line 157: Not actionable. Just sharing:
Based on the [source code of the `@angular/ssr ng-add` schematics](https://github.com/angular/angular-cli/blob/8914ef0a01e1d7673fdc5c25ec2bb0ed43cd3b3c/packages/schematics/angular/ssr/index.ts#L272-L288) when the main builder is not the application builder, then this `@angular/ssr ng-add` schematic should create the old-fashioned `server` and `serve-ssr` builders.

However, not sure when/if this condition will change, so it'd not take it as a starting point for any fancy mechanisms on our side.

Crazy idea on my mind:
- revert just the main builder to the webpack builder
- only then run the `@angular/ssr ng-add` schematics - and it will do the job of creating all of those other builders

But as said, above, I don't consider this idea seriously long-term

projects/storefrontapp/server.ts
```diff
@@ -16,7 +16,7 @@ import { Express } from 'express';
 import { existsSync } from 'fs';
 import { join } from 'path';
 import 'zone.js/node';
-import { AppServerModule } from './src/main.server';
+import AppServerModule from './src/main.server';
```
Platonn, line 19: 👍  
I like that our local /storefrontapp is as close as possible to a fresh app.

Platonn (commented):
code LGTM. yet need to test it

projects/schematics/src/add-ssr/files/server.__typescriptExt__
```diff
@@ -1,20 +1,21 @@
...
+import {
+  NgExpressEngineDecorator,
+  ngExpressEngine as engine,
+} from '@spartacus/setup/ssr';
 
-import { ngExpressEngine as engine } from '@nguniversal/express-engine';
-import { NgExpressEngineDecorator } from '@spartacus/setup/ssr';
-import * as express from 'express';
+import express from 'express';
 import { join } from 'path';
```
Platonn, line 8: [NITPICK] Can be done in a separate PR/ticket as "further improvements" later:

For consistency, this `server.ts` template and `server.ts` in our storefrontapp, ideally, should be (nearly) the same as fresh angular's created `server.ts`.

So let's use:
```ts
import { fileURLToPath } from 'node:url';
import { dirname, join, resolve } from 'node:path';
```
and later on:
```ts
  const serverDistFolder = dirname(fileURLToPath(import.meta.url));
  const browserDistFolder = resolve(serverDistFolder, '../browser');
  const indexHtml = join(serverDistFolder, 'index.server.html');


...
```

projects/schematics/src/add-ssr/files/server.__typescriptExt__
```diff
@@ -1,20 +1,21 @@
...
+import { existsSync } from 'node:fs';
+import AppServerModule from './src/main.server';
 
 const ngExpressEngine = NgExpressEngineDecorator.get(engine);
 
 // The Express app is exported so that it can be used by serverless Functions.
 export function app() {
   const server = express();
-  const distFolder = join(process.cwd(), '<%= browserDistDirectory %>');
+  const distFolder = join(process.cwd(), 'dist/ng-17-test-ssr-module/browser');
```
Platonn: [MEDIUM]
Now it's hardcoded. It didn't work, when my app name was different than "ng-17-test-ssr-module". 🚨 The error thrown while running `serve:ssr:<app-name>`:
`Error: Failed to lookup view "index" in views directory "<base directory>/test-ng17-pr18228/dist/ng-17-test-ssr-module/browser"`

Let's bring back the template variable. `<%  .... %>`

```suggestion
  const distFolder = join(process.cwd(), '<%= browserDistDirectory %>');
```

projects/schematics/src/add-ssr/files/server.__typescriptExt__
pawelfras: To use `import.meta.url` in our `server.ts` we'll have to change the value of the `module` property in `tsconfig.json` to at least `es2020` and thick change causes issues when building libraries. Let's do it later.

projects/schematics/src/add-ssr/files/server.__typescriptExt__
pawelfras: Yes, good catch! Initially, I added this variable to the snapshot, but then I made some changes, copy-pasted code from different places and eventually forgot to revert this change.

Platonn (approved):

Platonn (approved):

PR #18234, opened 2023-12-11
Upgrade to Angular 17 - init docs

Platonn:
init migration and installation docs


docs/migration/7_0/7_0-migration.md
pawelfras, line 1: Maybe let's change the name of the folder and file to contain `2211` as we know that at least this number is going to be a part of the new release version.

It concerns both the migration and installation file.

pawelfras (approved):

PR #18235, opened 2023-12-12
feat: CXSPA-5606 Upgrade to Angular 17 - adjusting unit tests for schematics

pawelfras:
- added flag `standalone: false` to configuration for angular's schematics used in tests
- adjusted tests for file utils to fit new AppComponent shape
- updated snapshots

Closes [CXSPA-5606](https://jira.tools.sap/browse/CXSPA-5606)

### DIFF
--- a/projects/schematics/src/shared/utils/__snapshots__/workspace-utils_spec.ts.snap
+++ b/projects/schematics/src/shared/utils/__snapshots__/workspace-utils_spec.ts.snap
@@ -4,15 +4,12 @@ exports[`Workspace utils getProject should return project 1`] = `
 {
   "architect": {
     "build": {
-      "builder": "@angular-devkit/build-angular:browser",
+      "builder": "@angular-devkit/build-angular:application",
       "configurations": {
         "development": {
-          "buildOptimizer": false,
           "extractLicenses": false,
-          "namedChunks": true,
           "optimization": false,
           "sourceMap": true,
-          "vendorChunk": true,
         },
         "production": {
           "budgets": [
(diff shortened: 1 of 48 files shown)


projects/schematics/src/shared/utils/__snapshots__/workspace-utils_spec.ts.snap
```diff
@@ -4,15 +4,12 @@ exports[`Workspace utils getProject should return project 1`] = `
 {
   "architect": {
     "build": {
-      "builder": "@angular-devkit/build-angular:browser",
+      "builder": "@angular-devkit/build-angular:application",
```
Platonn, line 7: [QUESTION]
I'm surprised in this snapshot (and others in this PR) there is no `prerender: true` option and `ssr: { some config here }`. But we know and checked that ng17 creates new apps with this flag enabled in the `build` section of the `application` builder.

Why we don't have it in our snapshots?

Platonn: Ah, I probably get now... it's because our tests run Angular schematics with no `--ssr` flag.

Would you perceive it as a gap in our test cases coverage? Shall we create at least one test for our installation schematics that asses the shape of the outcome app, when our schematics are run with `--ssr` flag?

e.g. I don't see anywhere in our epic branch a notion of a phrase `"prerender": true` - thus I deduce we nowhere test the snapshot of angular.json in fresh app created with our Spartacus installation schematics.

Note: we didn't introduce regression in this PR. What about _creating a separate ticket_ for improving the test coverage - to include `--ssr` in testing of our installation schematics?

Platonn (approved):

projects/schematics/src/shared/utils/__snapshots__/workspace-utils_spec.ts.snap
pawelfras: We have separate tests for Spartacus with SSR and they have been adjusted in a previous PR: https://github.com/SAP/spartacus/pull/18228/files

pawelfras: But, indeed - there are no tests verifying the shape of `angular.json` with SSR.

PR #18239, opened 2023-12-12
Upgrade to Angular 17 - update v 15/16 notions to v 17

Platonn:
Notes:
- main `README.md`: I need to talk to Matt. IMHO it would be better not to store our compatibility matrix with Angular there, but in our official docs instead. And here only link to it.

- [ ] TODO before merge:
  - build libs and publish to verdaccio
  - check if runs successfully: `scripts/install/run.sh install_npm`

closes [CXSPA-5528](https://jira.tools.sap/browse/CXSPA-5528)

### DIFF
--- a/docs/self-publishing-spartacus-libraries.md
+++ b/docs/self-publishing-spartacus-libraries.md
@@ -112,39 +112,39 @@ This procedure can be used to create a fresh application.
 
 ## Upgrading an Existing Spartacus App
 
-Before upgrading your Spartacus app to version 6.0, you first need to make sure your Angular libraries are up to date. Spartacus 6.0 requires Angular 15.
+Before upgrading your Spartacus app to newer version, you first need to make sure your Angular libraries are up to date in relation to Spartacus version. For example, Spartacus 6.0 requires Angular 15.
 
 ### Upgrading Your Angular Libraries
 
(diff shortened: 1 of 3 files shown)


pawelfras (commented):
File `functions.sh` has to be adjusted due to `standalone: true` as a default option when creating a new Angular app in v17:
https://github.com/SAP/spartacus/blob/821fcb68ca31524c0da7f685e140b9cc77e71375/scripts/install/functions.sh#L91

```bash
function create_shell_app {
    ( cd ${INSTALLATION_DIR} && ng new ${1} --style scss --no-routing --standalone=false)
}
```

pawelfras:
> File `functions.sh` has to be adjusted due to `standalone: true` as a default option when creating a new Angular app in v17:
> 
> https://github.com/SAP/spartacus/blob/821fcb68ca31524c0da7f685e140b9cc77e71375/scripts/install/functions.sh#L91
> 
> ```shell
> function create_shell_app {
>     ( cd ${INSTALLATION_DIR} && ng new ${1} --style scss --no-routing --standalone=false)
> }
> ```

Just for reference - as we discussed offline, flag `--ssr=false` is also required 👍

Platonn:
duplicate of #18254

PR #18254, opened 2023-12-14
Upgrade to Angular 17 - update v 15/16 notions to v 17

Platonn:
Notes:
- main `README.md`: I need to talk to Matt. IMHO it would be better not to store our compatibility matrix with Angular there, but in our official docs instead. And here only link to it.

QA Steps - test our installation script in the following scenarios:
  - [x] with `ANGULAR_CLI_VERSION='^17.0.5'`
    - [x] `run.sh install_npm` , using libs built locally manually and published in verdaccio
      - [x] then test if the created apps `/csr` and `/ssr` work correctly
    - [x] `run.sh install` (install from sources) with `BRANCH='feature/CXSPA-5528'` (and `SPARTACUS_VERSION='6.6.0-1'` - because this is the currently hardcoded version in our repo)
      - [x] then test if the created apps `/csr` and `/ssr` work correctly
  - [x] with `ANGULAR_CLI_VERSION='^15.0.0'` - check backward compatibility of the script (if it's still able to install older versions of Spartacus)
    - [x] `run.sh install_npm` (and `SPARTACUS_VERSION='latest` - to use latest version from RBSC)
      - [x] then test if the created apps `/csr` and `/ssr` work correctly

Note: in this branch there is a missing configuration for running installation script for publishing the library `@spartacus/pdf-invoices` to verdaccio. It's [already fixed](https://github.com/SAP/spartacus/commit/47dfc22ba1a31fe08709280532eab05140c4ebf2) in `develop-6.8.x` branch, which should be back-synced with our epic branch soon. So for the testing purposes I've fixed it only locally while running script, but didn't commit to this PR.

closes [CXSPA-5528](https://jira.tools.sap/browse/CXSPA-5528)

### DIFF
--- a/scripts/install/functions.sh
+++ b/scripts/install/functions.sh
@@ -88,7 +88,7 @@ function update_projects_versions {
 }
 
 function create_shell_app {
-    ( cd ${INSTALLATION_DIR} && ng new ${1} --style scss --no-routing)
+    ( cd ${INSTALLATION_DIR} && ng new ${1} --style scss --no-routing --standalone=false)
 }
 
 function add_b2b {
--- a/scripts/install/functions.sh
+++ b/scripts/install/functions.sh
@@ -50,7 +50,7 @@ function prepare_install {
     cmd_clean
 
     printh "Installing installation script prerequisites"
-    
+
     VERDACCIO_PID=`lsof -nP -i4TCP:4873 | grep LISTEN | tr -s ' ' | cut -d ' ' -f 2`
     if [[ -n ${VERDACCIO_PID} ]]; then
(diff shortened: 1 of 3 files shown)


scripts/install/functions.sh
```diff
@@ -39,7 +41,7 @@ function cmd_clean {
     printh "Cleaning old spartacus installation workspace"
 
     delete_dir ${BASE_DIR}
-    delete_dir storage
+    delete_dir ${VERDACCIO_STORAGE_DIR}
```
Platonn, line 44: This change was done by the way to improve readability of the code (to know what means the "magic string" `storage`)

PR #18268, opened 2023-12-18
fix: (CXSPA-1027) - Improve a11y of star rating component

Pio-Bar:
While not interactive the star rating should no longer present itself as such. It also provides proper text alternatives.


projects/storefrontlib/shared/components/star-rating/star-rating.component.html
```diff
@@ -1,11 +1,29 @@
...
-  [type]="icon"
-  class="star"
-  (mouseover)="setRate(i)"
-  (click)="saveRate(i)"
-  (keydown.space)="saveRate(i)"
-  [attr.tabindex]="disabled ? null : 0"
-  role="button"
-  attr.aria-label="{{ 'productReview.addRate' | cxTranslate: { count: i } }}"
-></cx-icon>
+<div
```
Zeyber: This situation would be considered a breaking change because we are changing the DOM structure:
```
// from
<cx-icon />

// to
<div>
  <cx-icon />
</div>
```

In order to resolve this issue while making the feature opt-in for users, we can make use of the `cxFeatureLevelDirective` like in this example:

```
<cx-icon *cxFeatureLevel="'!6.8'"> // or nested in <ng-container>

<div *cxFeatureLevel="'6.8'">
  <cx-icon />
</div>
```

The example above would preserve the original DOM without the fix for any user that has `featureLevel` in their spartacus config set to any version below `6.8` and use the fixed DOM once they set their feature level to that version. This is how we handle making breaking changes before a major release so that we can incrementally make changes in minors.

This does mean that we will need to clean up our changes in the next major (eg. next major would be `7.0`). To handle this, we create "deprecation" tickets with instructions of where to find the code and how to remove the issues. We may also need to fix e2e and unit tests accordingly. 

It is also helpful to add `TODO: @deprecated - remove such and such (CXSPA-xxxx)` comments to our corresponding clean up tickets so that it is easier to catch issues and clean up tech-debt.

Our docs have information on feature flags usage here: https://sap.github.io/spartacus-docs/configuring-feature-flags/

Zeyber (changes requested):

Zeyber (dismissed):

developpeurweb (dismissed):
Tested with JAWS. Element gets narrated.

Zeyber (approved):

PR #18277, opened 2023-12-18
fix: (CXSPA-1047) - stop focus indicator from expanding

Pio-Bar:
https://jira.tools.sap/browse/CXSPA-1047

Fixed stretched focus indicators on: 
- `Save cart for later` button on cart page.
- product name on PLP.

I haven't found more instances of this problem.


projects/storefrontlib/cms-components/product/product-list/product-list-item/product-list-item.component.html
```diff
@@ -19,12 +19,12 @@
     </a>
   </div>
   <div class="col-12 col-md-8">
-    <a
-      [routerLink]="{ cxRoute: 'product', params: product } | cxUrl"
-      class="cx-product-name"
-    >
-      <h2 [innerHtml]="product.nameHtml"></h2>
-    </a>
+    <h2 class="cx-product-name">
```
Zeyber: This will also be a breaking DOM change similar to this one:
https://github.com/SAP/spartacus/pull/18268#discussion_r1432623453

feature-libs/cart/saved-cart/styles/_add-to-saved-cart.scss
```diff
@@ -8,14 +8,13 @@ cx-add-to-saved-cart {
 
     .cx-action-link {
       &:nth-child(n + 2) {
-        margin-inline-end: 0;
+        position: relative;
```
Zeyber: Changes to styling are also considered breaking changes. But fortunately we have a way to handle these too with similar feature flags to DOM breaking changes:
https://sap.github.io/spartacus-docs/css-architecture/#style-versioning

We should leave clean up instructions in a similar way to the discussion here:
https://github.com/SAP/spartacus/pull/18268#discussion_r1432623453

@developpeurweb Do you have any other comments in this regard?

Zeyber (changes requested):

projects/storefrontlib/cms-components/product/product-list/product-list-item/product-list-item.component.html
```diff
@@ -19,12 +19,24 @@
...
-      [routerLink]="{ cxRoute: 'product', params: product } | cxUrl"
-      class="cx-product-name"
-    >
-      <h2 [innerHtml]="product.nameHtml"></h2>
-    </a>
+    <ng-container *cxFeatureLevel="'6.8'">
+      <h2 class="cx-product-name">
+        <a
+          [routerLink]="{ cxRoute: 'product', params: product } | cxUrl"
+          [innerHtml]="product.nameHtml"
```
Zeyber, line 26: Do you think it will be possible here to somehow render the html inside the h2 tag without using the `innerHtml` directive? 

The reason for this is to take advantage of the breaking change to make this component work with Angular 17's new feature "non-destructive" hydration. In which, we need to avoid manipulating the DOM to take advantage of this feature, and this is one of the places it doesn't work.

https://angular.io/guide/hydration#direct-dom-manipulation

Zeyber (changes requested):

projects/storefrontlib/cms-components/product/product-list/product-list-item/product-list-item.component.html
Pio-Bar: I've tested it quite a bit and it probably is not possible to keep the hydration benefits and render that HTML. Any direct DOM changes will cause a mismatch and I am not sure there is a way around it.

Zeyber: No problem. It can be quite a challenge so thanks for taking a look at it. Supporting hydration is not currently a requirement (who know's if it will become one).

Zeyber (dismissed):

feature-libs/cart/saved-cart/styles/_add-to-saved-cart.scss
developpeurweb: @Zeyber no further comments, all good.

developpeurweb (dismissed):
LGTM!

Pio-Bar:
New feature flags have been implemented

Zeyber (approved):

PR #18297, opened 2023-12-20
feat: (CXSPA-981) - Add keyboard navigation to 'navigation-ui'

Pio-Bar:
https://jira.tools.sap/browse/CXSPA-981

Because of the recursive approach to navigation, the keyboard controls have proven to be quite tricky to implement. If you find a more elegant approach I'd be happy to know.


projects/storefrontlib/cms-components/navigation/navigation/navigation-ui.component.ts
```diff
@@ -64,12 +64,19 @@ export class NavigationUIComponent implements OnInit, OnDestroy {
   private openNodes: HTMLElement[] = [];
   private subscriptions = new Subscription();
   private resize = new EventEmitter();
+  private arrowControls: Subject<KeyboardEvent> = new Subject();
```
Zeyber: Lets use the `protected` modifier here and in general across the spartacus project. The reason is that `protected` shows that it is not intended for direct manipulation by lib users. But they can if they really need to. 
If we make these `private`, then it is not possible for them to do so without making a copy of the spartacus library so that they can modify it (its really annoying to do that).

projects/storefrontlib/cms-components/navigation/navigation/navigation-ui.component.ts
```diff
@@ -173,6 +180,29 @@ export class NavigationUIComponent implements OnInit, OnDestroy {
     event.stopPropagation();
   }
 
+  onSpace(event: UIEvent) {
```
Zeyber: We should include unit tests here to test the functionality of these new methods. Possibly also e2e tests.
This is somewhat of a complex core component that does a lot of things and we really need to ensure nothing funny happens.

Zeyber (changes requested):

projects/storefrontlib/cms-components/navigation/navigation/navigation-ui.component.ts
Pio-Bar: done ;)

projects/storefrontlib/cms-components/navigation/navigation/navigation-ui.component.ts
Pio-Bar: I'm not super familiar with Jasmine thus not sure how to approach testing such a function. I have provided an e2e test though. :)

Zeyber: I took a look at the e2e test and its spot on! Nice work

In regards to unit tests, we have some coverage checks to make sure that code is tested too. This usually blocks us from merging code that doesn't have enough testing (something like 90% of our code is required) which is why we need to keep on top of them. 

This is important if we are adding new methods to make sure they still behave exactly as we expect. E2Es do of course do a lot of this for us. But lib users will sometimes just use one or two methods from a component and expect them to behave in a certain way. That's why we need to have unit tests with the `'should do this' ... 'should call that' ... 'should NOT do this'` type scenarios too.

As in how to do this, there are plenty of examples in our `.spec.ts` files and it helps to find similar component tests to base off when possible. In our `KeyboardFocus` feature, we have plenty of keyboard manipulation going on to support a11y such of focus directives and tab traps. You can take a look at those spec files for what I think will be the most relevant for these cases (eg. the `TabFocusDirective` has tests for arrow key presses).

Zeyber (changes requested):

projects/storefrontlib/cms-components/navigation/navigation/navigation-ui.component.ts
Pio-Bar: I gave it a go and proposed some unit tests for the keyboard controls. Let me know if this is the right direction, thanks!

Pio-Bar: Since getting access to JAWS i've tested it a bit and made some improvements too. ;)

projects/storefrontapp-e2e-cypress/cypress/e2e/accessibility/aux-key.e2e.cy.ts
```diff
@@ -12,7 +12,7 @@ context('Auxiliary Keys', () => {
       loadPageWithComponenents('/');
     });
 
-    it('should open and close menu with space key', () => {
+    it('should open menu with space key', () => {
```
Zeyber: Why are we removing the ability to close the menu with the space key?

projects/storefrontapp-e2e-cypress/cypress/e2e/accessibility/keyboard-navigation.e2e.cy.ts
```diff
@@ -0,0 +1,42 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+describe('Kayboard navigation', () => {
+  context('Navigation UI component', () => {
+    beforeEach(() => {
+      cy.visit('/');
+      cy.get('cx-navigation-ui [aria-haspopup="true"]').as('firstDropdown');
+    });
+    it('opens a node after pressing space', () => {
```
Zeyber: Depending on if there is a reason not to, we can also assert that the space button closes the menu.

projects/storefrontapp-e2e-cypress/cypress/e2e/accessibility/keyboard-navigation.e2e.cy.ts
```diff
@@ -0,0 +1,42 @@
...
+    it('navigates node children with upArrow key', () => {
+      cy.get('@firstDropdown').first().focus().type(' ');
+      cy.contains(' Fujifilm ').focus().type('{upArrow}');
+      cy.contains(' Toshiba ').should('have.focus').type('{upArrow}');
+      cy.contains(' Samsung ').should('have.focus').type('{upArrow}');
+      cy.contains(' Kodak ').should('have.focus').type('{upArrow}');
+      cy.contains(' Sony ').should('have.focus').type('{upArrow}');
+      cy.contains(' Canon ').should('have.focus').type('{upArrow}');
+      cy.contains(' Canon ').should('have.focus');
+    });
```
Zeyber, line 42: Should we also assert that pressing enter will navigate to the focused link's page? I'm thinking just to make sure the `a` tag is the element focused.

projects/storefrontlib/cms-components/navigation/navigation/navigation-ui.component.html
```diff
@@ -45,14 +47,15 @@
...
         </cx-generic-link>
         <button
-          [attr.tabindex]="depth < 1 ? 0 : -1"
           [attr.aria-haspopup]="true"
           [attr.aria-expanded]="false"
           [attr.aria-label]="node.title"
           (click)="toggleOpen($any($event))"
           (mouseenter)="onMouseEnter($event)"
-          (keydown.space)="toggleOpen($any($event))"
+          (keydown.space)="onSpace($any($event))"
```
Zeyber: Breaking change here and with the `tabindex`.

In regards to the `onSpace` method, it is also an option the `featureLevel` condition in the `onSpace` method itself (ie. `if (isLevel(x)) ... else toggleOpen`). But it might be easier to have a template condition since the tabindex is also being changed. Up to you.

projects/storefrontlib/cms-components/navigation/navigation/navigation-ui.component.html
```diff
@@ -45,14 +47,15 @@
...
           (keydown.esc)="back()"
+          (keydown.arrowDown)="focusOnNode($any($event))"
```
Zeyber: Just curious, if we have `arrowControls` in the class, do we need the `arrowDown` input in the template too?

projects/storefrontlib/cms-components/navigation/navigation/navigation-ui.component.ts
```diff
@@ -173,6 +184,45 @@ export class NavigationUIComponent implements OnInit, OnDestroy {
     event.stopPropagation();
   }
 
+  onSpace(event: UIEvent): void {
```
Zeyber, line 217: It could be nice to have JSDoc comments to explain what these new methods do.

projects/storefrontlib/cms-components/navigation/navigation/navigation-ui.component.ts
```diff
@@ -117,12 +129,6 @@ export class NavigationUIComponent implements OnInit, OnDestroy {
       typeof navNode.url === 'string' &&
       this.winRef.nativeWindow?.location.href.includes(navNode.url)
     ) {
-      this.elemRef.nativeElement
```
Zeyber, line 120: Any changes to existing methods are considered breaking changes so we should feature flag any functionality addition/modification/removal. Even if its a small change :/

Zeyber (changes requested):

projects/storefrontapp-e2e-cypress/cypress/e2e/accessibility/aux-key.e2e.cy.ts
Pio-Bar: Initially i removed it since a node can possibly be many children deep. It is now restored and the menu should close if you press space on a node with no children.

projects/storefrontapp-e2e-cypress/cypress/e2e/accessibility/keyboard-navigation.e2e.cy.ts
Pio-Bar: I've added the tests to see if the menu closes.

projects/storefrontapp-e2e-cypress/cypress/e2e/accessibility/keyboard-navigation.e2e.cy.ts
Pio-Bar: The problem here is Cyprees, (suprisingly) is unable to press enter after focusing an element. It is a long known issue (https://github.com/cypress-io/cypress/issues/8267). I think we can safely assume pressing enter will work since we target anchor tags and do not mess with default behaviour here.

projects/storefrontlib/cms-components/navigation/navigation/navigation-ui.component.html
Pio-Bar: I've added feature flags and a ticket to remove them.
https://jira.tools.sap/browse/CXSPA-5919

projects/storefrontlib/cms-components/navigation/navigation/navigation-ui.component.html
Pio-Bar: This helps with UX since you can enter the next node by just pressing down, especially on mobile.

projects/storefrontlib/cms-components/navigation/navigation/navigation-ui.component.ts
Pio-Bar: Comments added :)

projects/storefrontlib/cms-components/navigation/navigation/navigation-ui.component.ts
Pio-Bar: Added feature flags + task
https://jira.tools.sap/browse/CXSPA-5919

projects/storefrontlib/cms-components/navigation/navigation/navigation-ui.component.ts
```diff
@@ -117,13 +130,18 @@ export class NavigationUIComponent implements OnInit, OnDestroy {
...
-        .querySelectorAll('li.is-open:not(.back), li.is-opened')
-        .forEach((el: any) => {
-          this.renderer.removeClass(el, 'is-open');
-          this.renderer.removeClass(el, 'is-opened');
-        });
-      this.reinitializeMenu();
+      // TODO: (CXSPA-5919) Remove feature flag next major release
+      if (this.featureConfigService.isLevel('6.8')) {
```
Zeyber: It looks like earlier versions still need to also call this `reinitializeMenu()` function. So I think we just need to wrap the `this.elementRef.nativeElement...` DOM manipulation with `!6.8`.

projects/storefrontlib/cms-components/navigation/navigation/navigation-ui.component.ts
```diff
@@ -64,18 +69,26 @@ export class NavigationUIComponent implements OnInit, OnDestroy {
...
+  }
+
   constructor(
     private router: Router,
     private renderer: Renderer2,
     private elemRef: ElementRef,
     protected hamburgerMenuService: HamburgerMenuService,
-    protected winRef: WindowRef
+    protected winRef: WindowRef,
+    protected featureConfigService: FeatureConfigService
```
Zeyber: We need to make this an optional service because it is possible for services not to be provided when users extend this library. Users will need to make sure it is included to enable breaking change features/fixes.

eg `Optional() protected featureConfigService?: FeatureConfigService`

projects/storefrontlib/cms-components/navigation/navigation/navigation-ui.component.ts
```diff
@@ -186,7 +263,6 @@ export class NavigationUIComponent implements OnInit, OnDestroy {
 
   clear(): void {
     this.openNodes = [];
-    this.updateClasses();
```
Zeyber, line 189: Small breaking change

projects/storefrontlib/shared/test/mock-feature-level-directive.ts
```diff
@@ -16,6 +16,8 @@ export class MockFeatureLevelDirective {
   ) {}
 
   @Input() set cxFeatureLevel(_feature: string | number) {
-    this.viewContainer.createEmbeddedView(this.templateRef);
+    if (!_feature.toString().includes('!')) {
```
Zeyber: This is a really interesting condition you've added. Perhaps you could leave a comment explaining what it is for? It is a widely used directive so we need to be careful about changing anything here.

Zeyber (changes requested):

projects/storefrontlib/cms-components/navigation/navigation/navigation-ui.component.ts
Pio-Bar: I've included this invocation in the feature flag.

projects/storefrontlib/cms-components/navigation/navigation/navigation-ui.component.ts
Pio-Bar: done

projects/storefrontlib/cms-components/navigation/navigation/navigation-ui.component.ts
Pio-Bar: Good catch ;) Feature flag added

projects/storefrontlib/shared/test/mock-feature-level-directive.ts
Pio-Bar: Added comment for clarity

projects/storefrontlib/cms-components/navigation/navigation/navigation-ui.component.spec.ts
```diff
@@ -31,6 +36,13 @@ class MockHamburgerMenuService {
   toggle(_forceCollapse?: boolean): void {}
 }
 
+// TODO: (CXSPA-5919) Remove mock next major release
+class MockFeatureConfigService {
+  isLevel() {
+    return true;
```
Zeyber, line 42: Ok, I see why you have added the condition in the `MockFeatureLevelDirective`. I believe you don't need to import this directive.

Instead, you can continue to use the `MockFeatureConfigService` and return `false` for `isLevel()` for tests by default. But with the new tests you have added, you can add a `spy` for the `isLevel` function to make it return `true` for those new test cases. As I recall, it was only new cases that failed.

Zeyber: You can change the return value in the test by adding the line `spyOn(featureConfigService, 'isLevel').and.returnValue(true);`.

Zeyber (changes requested):

projects/storefrontlib/cms-components/navigation/navigation/navigation-ui.component.spec.ts
Pio-Bar: > As I recall, it was only new cases that failed.

It's actually the existing tests that are failing, the new ones are fine. It doesn't feel right to modify them just to accommodate some temporary code. I think leaving this condition is more elegant as this will future proof our tests and save some headaches. 

I can't imagine a scenario when additionally rendering DOM marked as deprecated will be the expected outcome. That's why the tests count twice as many nodes here. I'm not trying to argue this, I'm just curious about the reasoning here ;)

Zeyber (dismissed):
I am still concerned with the implementation of the MockFeatureDirective since it does not actually take the feature level into consideration when testing (ie. it only renders the new DOM and never the old even if the feature level is set to `false` when we use this directive. But I see this is something that we have never picked up on before so it is a good catch.

We will be changing the way the feature mechanism works in the next major so I want to approve this but readdress the issue based on however the new mechanism might work (if it is still relevant) once we have it.

developpeurweb (dismissed):

projects/schematics/src/shared/schematics-config-mappings.ts
```diff
@@ -5,11 +5,11 @@
  */
 
 import { SchematicsException } from '@angular-devkit/schematics';
```
Zeyber: Lets undo these changes not to mix up our commit.

Zeyber (dismissed):

projects/schematics/src/shared/schematics-config-mappings.ts
Pio-Bar: Reverted

Zeyber (approved):

PR #18313, opened 2023-12-21
fix: (CXSPA-965) - Refocus on input after quick order with keyboard

Pio-Bar:
https://jira.tools.sap/browse/CXSPA-965

After the item has been ordered the focus should go back to the input field.


feature-libs/cart/quick-order/components/quick-order/form/quick-order-form.component.ts
```diff
@@ -99,6 +103,7 @@ export class QuickOrderFormComponent implements OnInit, OnDestroy {
     }
 
     this.quickOrderService.addProduct(product);
+    this.quickOrderInput.nativeElement.focus();
```
Zeyber, line 106: Lets include a check in the quick order form e2e tests to assert the input gets focused after adding a product to prevent regression.

Zeyber (changes requested):

Zeyber (dismissed):

developpeurweb (dismissed):
Works well on my end.

Zeyber (approved):

PR #18319, opened 2023-12-27
chore: Google pay unit tests

rmch91:


integration-libs/opf/base/components/opf-quick-buy/google-pay/google-pay.service.spec.ts
```diff
@@ -4,4 +4,693 @@
...
+              }
+            );
+            expect(submitPaymentArgs.cartId).toBe(mockCartId);
+            expect(submitPaymentArgs.cartId).toBe(mockCartId);
+            expect(submitPaymentArgs.encryptedToken).toBe(encodedMockToken);
+            expect(submitPaymentArgs.paymentMethod).toBe(
+              PaymentMethod.GOOGLE_PAY
+            );
+            expect(mockCartHandlerService.getCurrentCartId).toHaveBeenCalled();
+            expect(result).toEqual({ transactionState: 'SUCCESS' });
```
rmch91, line 589: @FollowTheFlo Hello, Florent. Want to consult this. Seems like result always will be `transactionState: 'SUCCESS'`.

In the service, in line 253 the result has type 
```
{
    submitPayment: boolean;
}
```

In line 254 we have assignment `const isSuccess = result`
so isSuccess will also have type 
```
{
    submitPayment: boolean;
}
```

Thats mean that in this check `isSuccess ? 'SUCCESS' : 'ERROR'` because `isSuccess` is object, it will return 'SUCCESS' every time.

Seems like here `const isSuccess = result` should be `const isSuccess = result.submitPayment`. 

Is it make sense for you? I'm not sure because I haven't worked on the implementation, so I need your consultation. Thank you

Matejk00: I think that for now we need to mock always the "SUCCESS" status as we are working on the testing environment, and covered only a happy path scenario. So in this case such check looks ok.

integration-libs/opf/base/components/opf-quick-buy/google-pay/google-pay.service.spec.ts
```diff
@@ -4,4 +4,693 @@
...
+      );
+    });
+  });
+
+  describe('initClient', () => {
+    it('should initialize the Google Payment client with configurations', () => {
+      const activeConfiguration = {};
+      service.initClient(activeConfiguration);
+
+      const client = (service as any).googlePaymentClient;
```
Matejk00: Do we need to cast a serivce as `any` each time?

integration-libs/opf/base/components/opf-quick-buy/google-pay/google-pay.component.spec.ts
```diff
@@ -3,4 +3,86 @@
...
+    component.activeConfiguration = {};
+  });
+
+  it('should create', () => {
+    expect(component).toBeTruthy();
+  });
+
+  it('should initialize Google Pay client on init', async () => {
+    fixture.detectChanges();
+    await fixture.whenStable();
```
Matejk00: We are duplicating these two lines in several blocks. What do you think about creating a function with that and just call a function?

Matejk00 (changes requested):

integration-libs/opf/base/components/opf-quick-buy/google-pay/google-pay.component.spec.ts
rmch91: Changed

integration-libs/opf/base/components/opf-quick-buy/google-pay/google-pay.service.spec.ts
rmch91: the thing is that it is protected property of the service, changed it to bracket notation instead of casting to any

Matejk00 (approved):

PR #18341, opened 2024-01-03
chore: Unit tests for quick buy component

rmch91:


integration-libs/opf/base/components/opf-quick-buy/opf-quick-buy.service.spec.ts
```diff
@@ -4,4 +4,225 @@
...
+        expect(result).toBeTruthy();
+      });
+    });
+
+    it('should return true if the user is a guest and guest checkout is supported', () => {
+      baseSiteServiceMock.get.and.returnValue(
+        of({ baseStore: { paymentProvider: 'providerWithGuestSupport' } })
+      );
+
+      checkoutConfigMock.checkout.and.returnValue({
```
FollowTheFlo: small thing: I think it can be removed as defined on line 195. Moreover, following 'it' block does not have it.

rmch91: Right, that was a leftover. Thanks for catching it. Removed from code.

FollowTheFlo (approved):
good to me, coverage all good.

PR #17839, opened 2023-09-18
feat: CXSPA-4442 New components for email and personal info

scarai-sap:
ATT. Below is how it works now

Update new video with new css on 18:12 21 Sept. Please check it again

https://github.com/SAP/spartacus/assets/38693320/0aeccc2b-e62c-4855-a3bb-4b45ea484102

PR #18348, opened 2024-01-09
Feat: my account v2 (holidays)

i53577:


integration-libs/cdc/user-profile/update-email/cdc-update-email.module.ts
```diff
@@ -48,6 +66,9 @@ import { CDCUpdateEmailComponentService } from './cdc-update-email-component.ser
         },
       },
     }),
+    provideDefaultConfigFactory(() =>
+      inject(USE_MY_ACCOUNT_V2_EMAIL) ? myAccountV2EmailWithCDCMapping : {}
```
i53577: use `myAccountV2EmailWithCDCMapping` when USE_MY_ACCOUNT_V2_EMAIL is true

integration-libs/cdc/user-profile/update-email/cdc-update-email.module.ts
```diff
@@ -6,23 +6,41 @@
...
+  UpdateEmailComponentService,
+} from '@spartacus/user/profile/components';
 import { CDCUpdateEmailComponentService } from './cdc-update-email-component.service';
 
+const myAccountV2EmailWithCDCMapping: CmsConfig = {
+  cmsComponents: {
+    UpdateEmailComponent: {
+      providers: [
+        {
+          provide: MyAccountV2EmailComponentService,
```
i53577: when `USE_MY_ACCOUNT_V2_EMAIL` is true, provide `MyAccountV2EmailComponentService` instead of `UpdateEmailComponentService`

integration-libs/cdc/user-profile/update-password/cdc-update-password-component.service.ts
```diff
@@ -57,8 +68,37 @@ export class CDCUpdatePasswordComponentService extends UpdatePasswordComponentSe
       });
   }
 
+  protected onSuccess(): void {
+    let successMessage;
+    if (this.enableMyAccountV2) {
```
i53577: logic for `myAccountV2`

i53577:
The changes in cdc files are to fix the bug [CXSPA-5853](https://jira.tools.sap/browse/CXSPA-5853). It is caused by cdc doesn’t support my-account-v2. Because of timeline(FF in Feb.14th), I helped to fixed it in this PR.

feature-libs/user/_index.scss
```diff
@@ -7,7 +7,9 @@
 $skipComponentStyles: () !default;
 $selectors: cx-address-book, cx-address-form, cx-suggested-addresses-dialog,
   cx-login, cx-login-form, cx-register, cx-reset-password, cx-close-account,
-  cx-close-account-modal !default;
+  cx-close-account-modal, cx-my-new-account-v2-profile,
+  cx-my-new-account-v2-email, cx-my-account-v2-password,
+  cx-my-account-v2-combined-profile-email !default;
```
Platonn: [MEDIUM]
For consistency, let's keep the prefixes `cx-my-account-v2-...` and not inject in the middle other words like `new`. This will help later quickly search and identify all my-account-v2 components.

Note: Please check also other files, where selectors' prefixes need to be unified

feature-libs/user/profile/components/my-account-v2/new-password/my-account-v2-password-component.service.ts
```diff
@@ -0,0 +1,108 @@
...
+  HttpErrorModel,
+  RoutingService,
+} from '@spartacus/core';
+import { CustomFormValidators } from '@spartacus/storefront';
+import { UserPasswordFacade } from '@spartacus/user/profile/root';
+import { BehaviorSubject } from 'rxjs';
+import { tap } from 'rxjs/operators';
+
+@Injectable()
+export class MyAccountV2PasswordComponentService {
```
Platonn: [MAJOR]
(this comment is related to all component-services in the `@spartacus/user` library, but I'm just placing it here)

I can see we copy-pasted the whole implementation of all original component-services to new my-account-v2's component-services. And I can see the my-account-v2 copies don't differ essentially from their originals, so I'm wondering if copy-pasting was justified.

Did you consider alternative solutions? If yes, what pros/cons of them do you see? 
E.g. did you consider creating new my-account-v2 components, but reusing there the existing component-services (with adding tiny non-breaking improvements there)?

I can see tiny differences between original component-services and the new ones:
1. update-profile-component.service - the new one has a new propery `updateSecceed$`
2. update-email-component.service - the new one has  a new propery `updateSecceed$` AND new form control `oldEmail`
3. update-password-component.service - the new one has an added EmailValidator (which was missing in the original one, which I guess was a bug) AND global messages have a different translation key `myAccountV2PasswordForm` vs `updatePasswordForm` (but the same value: `"passwordUpdateSuccess": "Password updated with success"` and `"accessDeniedError": "Access is denied"`)

I'm happy to hear your thoughts, because I might be missing something important.

projects/storefrontlib/cms-components/myaccount/my-account-v2/index.ts
```diff
@@ -6,3 +6,6 @@
 
 export * from './my-account-v2-navigation/index';
 export * from './my-account-v2.module';
+export * from './my-account-v2-consent-management/index';
+export * from '../my-account-v2-notification-preference/index';
```
Platonn: [MAJOR]
(this comment is related to all components in the `@spartacus/storefront` lib, but I'm placing it just here)

Please correct me if I'm wrong: I can see that the my-account-v2 components in storefrontlib are just copy-pasted from the original components, without any essential differences.
Did you consider other alternatives, e.g. reusing old CMS components as-is (perhaps with minimal additions covered with a feature flag)?

I can see:
- consent-management-form.component - the v2 version only added a property `consentApprovedTime` in a non-breaking way
- consent-management.component - v2 has identical TS and HTML DOM structure to v1
- notification-preference.component - v2 has identical TS and HTML DOM structure to v1
Again, please correct me if I'm wrong (I might be missing something!).

Note: components from the `storefront` lib are eagerly loaded (due to historical reasons), which means they are bundled in the `main.js` - which is loaded even for an anonymous user visiting the homepage. We shouldn't enlarge the `main.js` bundle, unless _really_ justified. Adding copies of existing components in this library seems to be contributing considerably to the eagerly loaded `main.js` bundle size, which eventually impacts Core Web Vitals and SEO.
If it was a lazy-loadable JS chunk, we could be less strict about it.

feature-libs/user/profile/components/my-account-v2/email/my-account-v2-email.component.html
```diff
@@ -0,0 +1,135 @@
+<cx-spinner
+  class="overlay"
+  *ngIf="(isUpdating$ | async) || !(user$ | async); $"
+>
+</cx-spinner>
+<div>
+  <div class="">
+    <div class="row">
+      <div class="col-sm-6">
+        <label class="email-enhancedUI-headertext"
```
Platonn: [NITPICK]
Do you think we can drop the part `-enhancedUI-` part in the CSS classes in this PR? I believe that prefixes`cx-my-acocunt-v2` in selectors are enough scoping the CSS rules.

Note: I can see in many other components in this PR there is no `-enhancedUI-` part in the CSS classe.

feature-libs/user/profile/components/my-account-v2/email/my-account-v2-email-component.service.ts
```diff
@@ -0,0 +1,113 @@
...
+    protected routingService: RoutingService,
+    protected globalMessageService: GlobalMessageService,
+    protected authService: AuthService,
+    protected authRedirectService: AuthRedirectService
+  ) {}
+
+  protected busy$ = new BehaviorSubject(false);
+
+  updateSucceed$ = new Subject();
+  // ToolHeaderUtilitySeparator.
```
Platonn: [NITPICK]
Unused comment can be removed

feature-libs/user/profile/components/my-account-v2/index.ts
```diff
@@ -0,0 +1,11 @@
...
+ * SPDX-FileCopyrightText: 2023 SAP Spartacus team <spartacus-team@sap.com>
+ * SPDX-FileCopyrightText: 2024 SAP Spartacus team <spartacus-team@sap.com>
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+export * from './email/index';
+export * from './user-profile/index';
+export * from './new-password/index';
+export * from './index';
```
Platonn: [QUESTION]
Is this line really needed? I seems to be re-exporting everything from _itself_?

projects/assets/src/translations/en/my-account-v2-consent.i18n.ts
Platonn, line 1: [MEDIUM]
Could you change all `.i18n.ts` files in this PR to `.json` files, please?

It's due to a recent change we made in the whole repo (but I'm afraid we didn't announce it enough yet, I apologize). We started using JSONs for translation resources to conform the requirements of the SAP Translations Team, who might contribute to our repo in the future with new JSON files but translated to other langauges than English. They simply prefer consuming JSON files from us rather than TS files.

Take as an example the ASM or any other library:
- https://github.com/SAP/spartacus/blob/f2f32933945b4daaeb4fc19377e841f37e83496d/feature-libs/asm/assets/translations/en/asm.json
- https://github.com/SAP/spartacus/blob/f2f32933945b4daaeb4fc19377e841f37e83496d/feature-libs/asm/assets/translations/en/index.ts
- https://github.com/SAP/spartacus/blob/f2f32933945b4daaeb4fc19377e841f37e83496d/feature-libs/asm/assets/translations/translations.ts

feature-libs/user/profile/components/my-account-v2/email/my-account-v2-email.component.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  user$ = this.userProfile
+    .get()
+    .pipe(filter((user): user is User => Boolean(user)));
+  globalMessageType = GlobalMessageType;
+
+  constructor(
+    protected emailComponentservice: MyAccountV2EmailComponentService,
+    protected userProfile: UserProfileFacade
+  ) {}
```
Platonn: [MEDIUM]
In new classes (and for new dependencies in existing classes), let's not use `constructor()` for dependency injection, but the `inject()` function of Angular - and let's place it's usage the top of the class' body. For more, see our [coding guidelines](https://wiki.one.int.sap/wiki/display/spar/Coding+Guidelines#CodingGuidelines-Useinject()functioninsteadofconstructorforDependencyInjection).

feature-libs/user/profile/components/my-account-v2/combined-profile-email/my-account-v2-combined-profile-email.component.html
```diff
@@ -0,0 +1,15 @@
+<div>
+  <hr />
+  <div class="row">
+    <div class="col combined-profile-container">
```
Platonn: [MEDIUM]
New Spartacus features, should not use Bootstrap, but custom styles based on flexbox.
cc @developpeurweb 

Note: This comment applies also to other HTML files in this PR.

feature-libs/user/profile/components/my-account-v2/combined-profile-email/my-account-v2-combined-profile-email.component.html
```diff
@@ -0,0 +1,15 @@
+<div>
```
Platonn: [MEDIUM]
Unnecessary DOM elements should be avoided.

Motivation: especially in mobile devices we should avoid having too big DOM elements complexity. The more DOM elements, the more challenging it is to render a page for a browser, especially on low-end mobile devices.
cc @developpeurweb 

Note: Please review also other places in this PR to avoid unnecessary DOM elements

feature-libs/user/profile/components/my-account-v2/email/my-account-v2-email.component.ts
```diff
@@ -0,0 +1,61 @@
...
+import { GlobalMessageType, User } from '@spartacus/core';
+
+import { UserProfileFacade } from '@spartacus/user/profile/root';
+import { filter } from 'rxjs/operators';
+
+@Component({
+  selector: 'cx-my-new-account-v2-email',
+  templateUrl: './my-account-v2-email.component.html',
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  host: {},
```
Platonn: [NITPICK]
Unused `host` declaration can be removed.

Note: it applies also to `my-account-v2-profile.component.ts`

feature-libs/user/profile/styles/my-account-v2/_my-account-v2-combine-profile.scss
Platonn, line 1: [NITPICK]
`combine` vs `combined` - typo in filename? Let's sync filename with the css class name

feature-libs/user/profile/components/update-email/token/context.ts
Platonn, line 1: [MINOR]
The filename `token/context.ts` IMHO is not very semantic and doesn't describe the feature flag it contains.
What about:
- rename the file to `use-my-account-v2-email.ts`
- move it to the `./my-account-v2` folder, as it's related only to my-acocunt-v2 topic

Note: similar remark applies also to other analogical files `token/context.its` in this file

feature-libs/user/profile/components/my-account-v2/combined-profile-email/my-account-v2-combined-profile-email.module.ts
```diff
@@ -0,0 +1,74 @@
...
+            MyAccountV2ProfileComponentService,
+          ],
+        },
+      },
+    }),
+  ],
+  declarations: [MyAccountV2CombinedProfileEmailComponent],
+  exports: [MyAccountV2CombinedProfileEmailComponent],
+})
+export class NewCombinedProfileModule {}
```
Platonn: [MEDIUM]
This module's name doesn't follow the naming convention of prefixing with MyAccountV2, and is not related to the component that it contains.
What about renaming to `MyAccountV2CombinedProfileEmailModule`?

feature-libs/user/profile/styles/my-account-v2/_my-account-v2-email.scss
```diff
@@ -0,0 +1,83 @@
+%cx-my-new-account-v2-email {
+  .email-enhancedUI-headertext {
+    font-size: $h3-font-size;
+    font-weight: var(--cx-font-weight-semi);
+    color: $text;
+    padding-bottom: 1.5rem;
+  }
+
+  .email-enhancedUI-editButton {
+    float: right;
```
Platonn: [MEDIUM]
Let's not use float, but flexbox in our CSS styles.
cc @developpeurweb 

Note: similar remark applies also to other SCSS files in this PR

integration-libs/cdc/user-profile/update-email/cdc-update-email-component.service.ts
```diff
@@ -13,11 +14,17 @@ import {
...
+  USE_MY_ACCOUNT_V2_EMAIL,
+  UpdateEmailComponentService,
+} from '@spartacus/user/profile/components';
 import { UserEmailFacade } from '@spartacus/user/profile/root';
+import { Subject } from 'rxjs';
 
 @Injectable()
 export class CDCUpdateEmailComponentService extends UpdateEmailComponentService {
+  updateSucceed$ = new Subject();
+  enableMyAccountV2 = inject(USE_MY_ACCOUNT_V2_EMAIL);
```
Platonn: [MAJOR]
In other {MAJOR} comment I asked if we could possibly not create MyAccountV2-flavored component-services, but just make small changes in the original ones. Then the consequence would be that we would likely not need v2-flags conditions in the CDC-flavored component-service.

Of course it's still open for discussion, whether we will leave separate MyAccountV2-flavored component-services. But let me leave this comment here, just as a reminder to followup on the topic CDC component-services after we make the decision about the MyAccountV2 component-services.

projects/storefrontapp-e2e-cypress/cypress/helpers/accessibility/tabbing-order.config.ts
```diff
@@ -390,6 +390,9 @@ export const tabbingOrderConfig: TabbingOrderConfig = {
   notificationPreference: [
     { value: 'Email', type: TabbingOrderTypes.CHECKBOX_WITH_LABEL },
   ],
+  notificationPreferenceV2: [
```
Platonn: [MINOR]
For consistency, let's prefix all code related to my-account-v2 with the prefix `my-account-v2` (in appropriate casing), but let's not add a `V2` prefix just to anything, because it might be confusing.
So I suggest to rename it to `myAccountV2NotificationPreference`.

Note: analogical suggestion applies to also other places in e2e tests in this PR.
Note 2:  It's not a part of our public API, so it's a [MINOR] comment.

feature-libs/user/profile/components/my-account-v2/email/my-account-v2-email.component.ts
```diff
@@ -0,0 +1,61 @@
...
+  isEditing: boolean;
+  showingAlert: boolean;
+
+  user$ = this.userProfile
+    .get()
+    .pipe(filter((user): user is User => Boolean(user)));
+  globalMessageType = GlobalMessageType;
+
+  constructor(
+    protected emailComponentservice: MyAccountV2EmailComponentService,
```
Platonn: [MEDIUM]
typo. let's rename `emailComponentservice` to `emailComponenService`

feature-libs/user/_index.scss
i53577: renamed

feature-libs/user/profile/components/my-account-v2/new-password/my-account-v2-password-component.service.ts
i53577: Please check the latest code. @scarai-sap did some refactors for this.

feature-libs/user/profile/components/my-account-v2/email/my-account-v2-email-component.service.ts
i53577: removed it

feature-libs/user/profile/components/my-account-v2/index.ts
i53577: removed it

projects/assets/src/translations/en/my-account-v2-consent.i18n.ts
i53577: done

feature-libs/user/profile/components/my-account-v2/email/my-account-v2-email.component.ts
i53577: done

feature-libs/user/profile/components/my-account-v2/combined-profile-email/my-account-v2-combined-profile-email.component.html
i53577: removed this file. it's not used.

feature-libs/user/profile/components/my-account-v2/combined-profile-email/my-account-v2-combined-profile-email.component.html
i53577: removed

feature-libs/user/profile/components/my-account-v2/email/my-account-v2-email.component.ts
i53577: removed

feature-libs/user/profile/styles/my-account-v2/_my-account-v2-combine-profile.scss
i53577: removed this file. it's not used.

feature-libs/user/profile/components/update-email/token/context.ts
i53577: renamed de moved

integration-libs/cdc/user-profile/update-email/cdc-update-email-component.service.ts
i53577: When start server with cdc and my-account-v2 together, CDCUpdatexxxxComponentService will overwrite some functions in V1XxxxService. But my-account-v2 also need to add some logic in those functions. In this condition, I'm afraid we can’t avoid touching cdc files.

feature-libs/user/profile/components/my-account-v2/combined-profile-email/my-account-v2-combined-profile-email.module.ts
i53577: removed this module. it's not used

projects/storefrontapp-e2e-cypress/cypress/helpers/accessibility/tabbing-order.config.ts
i53577: renamed

feature-libs/user/profile/components/my-account-v2/email/my-account-v2-email.component.ts
i53577: fixed

feature-libs/user/profile/components/my-account-v2/email/my-account-v2-email.component.html
i53577: sure. renamed.

projects/storefrontlib/cms-components/myaccount/my-account-v2/index.ts
i53577: Hi @Platonn

- consent-management-form.component - yes, v2 only add a property consentApprovedTime, but it changed almost all the DOM elements and changed almost all the styles
- consent-management.component - v2 and v1 have different  HTML DOM structure , v2 changed some DOM elements and styles
- notification-preference.component - v2 and v1 have different HTML DOM structure, v2 added a header and changed the styles.

Do you want to combine v1 html and v2 html, then use `ngif` to consider which should be rendered? Or could we just keep it as it is?

feature-libs/user/profile/components/update-email/update-email-component.service.ts
```diff
@@ -34,12 +35,15 @@ export class UpdateEmailComponentService {
...
 
+  updateSucceed$ = new Subject();
+
   isUpdating$ = this.busy$.pipe(
     tap((state) => (state === true ? this.form.disable() : this.form.enable()))
   );
 
   form: UntypedFormGroup = new UntypedFormGroup(
     {
+      oldEmail: new UntypedFormControl(''),
```
Platonn: [QUESTION]
1. Perhaps I'm missing something obvious. From UX perspective, could you help me understand the purpose of asking an user to type an old email into separate field, when they just want to set a new one? 
2. I have problem to find a place in the code, where `newEmail` would be actually used and sent to the backend. I can see in the [lines 67-73](https://github.com/SAP/spartacus/pull/18348/files#diff-e04ca982ffc07e608054f16f121c89b564ca8f2751209619be2ff70d0e3d01cdR67-R73) that we pass to the `UserEmailFacade.update()` only the `password` and the `newEmail`. Was it deliberate? This makes me even more confused about our addition of the form field `oldEmail` in the v2 component.

feature-libs/user/profile/components/update-email/update-email.module.ts
```diff
@@ -16,18 +16,45 @@ import {
...
+          useClass: UpdateEmailComponentService,
+          deps: [
+            UserEmailFacade,
+            RoutingService,
+            GlobalMessageService,
+            AuthService,
+            AuthRedirectService,
+          ],
+        },
+      ],
```
Platonn: [MINOR]
Do we really need to re-define `providers` in the cms mapping of v2? Since they are now identical to the `providers` in the cms mapping v1, I believe this can be dropped.

Note: analogical observation applies to also other NgModules in this PR (where parts of cms mappings identical in v1 and v2 don't need to be repeated in v2).
Note: similar observation applies also to property `guards` - it is also "inherited" from the cms mapping v1, and don't need to be re-declared in v2 if it's identical to v2

feature-libs/user/profile/components/update-password/update-password-component.service.ts
```diff
@@ -42,10 +42,7 @@ export class UpdatePasswordComponentService {
   form: UntypedFormGroup = new UntypedFormGroup(
     {
       oldPassword: new UntypedFormControl('', Validators.required),
-      newPassword: new UntypedFormControl('', [
-        Validators.required,
-        CustomFormValidators.passwordValidator,
```
Platonn, line 47: [MEDIUM]
Was it a deliberate change? Why don't we check the validity of the new email anymore?

feature-libs/user/profile/components/update-password/update-password-component.service.ts
```diff
@@ -78,7 +75,7 @@ export class UpdatePasswordComponentService {
 
   protected onSuccess(): void {
     this.globalMessageService.add(
-      { key: 'updatePasswordForm.passwordUpdateSuccess' },
+      { key: 'myAccountV2PasswordForm.passwordUpdateSuccess' },
```
Platonn: [MEDIUM]
Because this componet-service is shared in between component v1 and v2, the i18n key used should be conditional, based on the feature flag.

PS. if we're going to inject a feature flag to the class, please:
- let's use `inject()` function 
- let's make the property of the feature flag `private`, not to pollute the Public API surface, and to be able to seamlessly remove this property in the future without a breaking change

Note: similar remark applies to other places in component-services in this PR

integration-libs/cdc/user-profile/update-email/cdc-update-email-component.service.ts
Platonn: [MEDIUM]
When injecting v2 feature flags, please let's make them `private` properties, not to pollute the Public API surface of the component, and so we're able to remove this property in the future without a breaking change.

Note: similar remark applies to also other places in this PR, where we inject a feature flag and assign to a class' property.

integration-libs/cdc/user-profile/update-email/cdc-update-email-component.service.ts
```diff
@@ -33,6 +40,9 @@ export class CDCUpdateEmailComponentService extends UpdateEmailComponentService
       authService,
       authRedirectService
     );
+    if (this.enableMyAccountV2) {
+      this.form.addControl('oldEmail', new UntypedFormControl(''));
+    }
```
Platonn: [MEDIUM]
Since we've already introduced property `updateSucceed$` and form control `oldEmail` to the parent class `UpdateEmailComponentService`, I believe we don't need to re-declare those things in the CDC-flavored child class? 
Perhaps I'm missing something obvious. My understand is that now all changes in `CDCUpdateEmailComponentService` could be reverted. Would you agree?

Note: analogically, my understanding is that all the changes in all the CDC-flavored component-services can be reverted, because the added logic there is already inherited from their relevant parent classes.

projects/storefrontlib/cms-components/myaccount/my-account-v2/index.ts
Platonn: Thank you for the explanation. I missed the fact that that HTML/CSS are different. I can however that TS are identical.

I'm still concerned about adding new kilobytes of code to the `main.js` bundle, that will be loaded for every user on the first page's visit.

As far as I understand, the components of Consent Management and Notification Preferences are displayed actually only for logged-in users and only when they enter the `/my-account/...` route. If yes, then fundamentally they don't belong to the `@spartacus/storefront` library but more to `@spartacus/user` library - which is lazy loadable.

Up until now we didn't split those components to the user library.
But since we're copy-pasting those components anyway, what do you think about placing those v2 components to the `@spartacus/user` library? (e.g. `@spartacus/user/profile or `@spartacus/user/account` - I'm not sure which one is more appropriate). Of course, then the CMS sample data should be different: we won't map the Angular implementations of v2 components anymore to the old CMS component types. But for v2, a brand new, different strings CMS component types should be used. And the old cms component types should be removed from sample data in such a case - but only if our customer wants to use v2.

On one hand it would be great to place new components in the more appropriate place - in a lazy laodable library instead of the storefront lib.
On the other hand, complicates the setup of v2 versus v1. It would require that on routes /notification-preferences and /consent-management the CMS sample data (the CMS component types) should be served differently for a storefront that wants to use MyAccount v1 versus a storefront that wants to use MyAccount v2.

Because of the requirement to change conditionally sample data, I perceive this whole improvement cumbersome and not a low-hanging fruit.

Therefore let me propose a different solution for minimizing the footprint of additional KB in the `main.js` bundle:

### Inheriting TS of components v1 in components v2
I can see that TS of components v2 are identical as TS of components v1. This copied source code contributes to the additional footprint of KB in the `main.js` bundle, which is probably not necessary. I have an idea that we could simply extend the v1 class by the v2 class, and only define a custom HTML. For example:

```ts
@Component({
  selector: 'cx-my-account-v2-consent-management',
  templateUrl: './my-account-v2-consent-management.component.html',
})
export class MyAccountV2ConsentManagementComponent extends ConsentManagementComponent {}
```

Could you please try if it works for you?

PS. in case of `consentApprovedTime` (inside `MyAccountV2ConsentManagementFormComponent`) we can just add this logic seamlessly to the TS of component v1, so it's inherited in v2. IMHO it's a fair tradeoff - it's less KB footprint than duplicating the whole TS of class v1 to the class v2.

Note: this applies to all components in the `@spartacus/storefront` library, because they all have (almost) identical TS of components v1 to components v2

feature-libs/user/profile/components/update-email/my-account-v2-email.component.html
```diff
@@ -0,0 +1,128 @@
+<cx-spinner
+  class="overlay"
+  *ngIf="(isUpdating$ | async) || !(user$ | async); $"
+>
+</cx-spinner>
+<div>
+  <div class="">
+    <div class="row">
+      <div class="col-sm-10">
```
Platonn: [MEDIUM]
New Spartacus features, should not use Bootstrap, but custom styles based on flexbox.
cc @developpeurweb

Note: This comment applies also to other HTML files in this PR.
Note: it's a continuation of another [comment](https://github.com/SAP/spartacus/pull/18348#discussion_r1473311971), which was resolved due to the other file being removed. However we still have other HTML files in this PR which use Bootstrap styles.

feature-libs/user/profile/components/update-email/update-email-component.service.ts
scarai-sap: Q1.It's just show old email(read only) similar as show customerId when update profile, The end user don't need type it again. I have a video recording pasted in https://github.com/SAP/spartacus/pull/17839 and UX/PO are happy with it.

Q2,   newEmail and password is enough, as the token present who is the people try to update it's email.

Hope it's clear

feature-libs/user/profile/components/update-password/update-password-component.service.ts
scarai-sap: Only check password validity. It's known issue as the backend also check password validity. We'll not check them in front end.

In the past we only fix it in V2. Now it's applied to shared libraries

integration-libs/cdc/user-profile/update-email/cdc-update-email-component.service.ts
scarai-sap: Good points, reverted

integration-libs/cdc/user-profile/update-email/cdc-update-email-component.service.ts
scarai-sap: allo reverted

scarai-sap: reverted

feature-libs/user/profile/components/update-email/my-account-v2-email.component.html
scarai-sap: Fixed. Please check.

BTW: where is the guideline for it.

feature-libs/user/profile/components/update-email/update-email-component.service.ts
scarai-sap: A pic is more helpful for answer Q1. The old value is only for display, designed by UX.


image

feature-libs/user/profile/components/update-password/update-password-component.service.ts
scarai-sap: Fixed.

feature-libs/user/profile/components/update-email/update-email.module.ts
scarai-sap: Fixed

projects/storefrontlib/cms-components/myaccount/my-account-v2-notification-preference/index.ts
Platonn, line 1: [MINOR]
I can see we have now 2 sibling folders on this level:
```
- my-account-v2-notification-preference
- my-account-v2
  - my-account-v2-consent-management
  - my-account-v2-navigation
 ```
 
For consistency, what about moving folder also `my-account-v2-notification-preference` into `/my-account-v2`?

File: feature-libs/user/profile/components/update-email/update-email-component.service.ts (hunk 19 above)
Comment (Platonn) [reply feature-libs/user/profile/components/update-email/update-email-component.service.ts]: [MEDIUM]
If the `oldEmail` is not used for the purpose of the form (i.e. it's not submitted to the backend, nor validated), but only for informing the user, IMHO it's a bit misleading that it's a part of the _form model_ preapared in the component-service.

Instead, I'd suggest to remove it entirely from the `this.form` model in component-service (and therefore not bind it in the html temlate via `[ngModel]`), but in the HTML rather create a standalone dummy input for it with a simple binding for the attribute `value`:
```diff
   <input
-    required="true"
     type="text"
     class="form-control"
-    name="oldEmail"
-    formControlName="oldEmail"
-    [ngModel]="(user$ | async)?.uid"
+    [value]="(user$ | async)?.uid"
     readonly
   />
```

What do you think about it?

projects/storefrontlib/cms-components/myaccount/my-account-v2-notification-preference/index.ts
scarai-sap: Moved to new place as suggested

Platonn (dismissed):
LGTM.

Please let's just wait with merging for a feedback from @developpeurweb , focused on a11y and styles.

projects/storefrontstyles/scss/components/myaccount/_new-consent-management.scss
```diff
@@ -5,17 +5,17 @@
   }
 
   .header {
-    font-weight: $line-height-base * 375;
+    font-weight: var(--cx-font-weight-semi);
     font-size: $font-size-base * 1.375;
-    font-family: 'Open Sans';
-    margin-bottom: $font-size-base * 1.875;
+    font-family: $font-family-sans-serif;
```
developpeurweb: * Please use our `—cx-color` variables for colours and `@include type()` mixin for predefined fonts. We are no longer using Bootstrap variables.
* Please make sure all your CSS code follows the guidelines provided on this wiki page https://wiki.one.int.sap/wiki/display/spar/UI+Styling+and+Accessibility+Onboarding 
* Please reach out to Spartacus Theme Owner if any doubts and to follow up with code review: @morganm58

feature-libs/user/profile/styles/_new-profile.scss
```diff
@@ -45,4 +38,20 @@
...
+  }
+
+  .btn-group {
+    display: flex;
+    justify-content: space-between;
+    width: 40%;
+  }
+
+  .ng-select .ng-select-container {
+    background-color: $background ;
```
developpeurweb: * Please use our `—cx-color` variables for colours and `@include type()` mixin for predefined fonts. We are no longer using Bootstrap variables.
* Please make sure all your CSS code follows the guidelines provided on this wiki page https://wiki.one.int.sap/wiki/display/spar/UI+Styling+and+Accessibility+Onboarding 
* Please reach out to Spartacus Theme Owner if any doubts and to follow up with code review: @morganm58

projects/storefrontstyles/scss/components/myaccount/_cx-my-account-v2-notification-preference.scss
```diff
@@ -0,0 +1,82 @@
...
+    flex-wrap: wrap;
+    right: -($font-size-base * 0.9375);
+    left: -($font-size-base * 0.9375);
+  }
+
+  .header {
+    width: $font-size-base * 14.375;
+    height: $font-size-base * 1.875;
+    font-weight: var(--cx-font-weight-semi);
+    font-size: $h3-font-size;
```
developpeurweb: * Please use our `—cx-color` variables for colours and `@include type()` mixin for predefined fonts. We are no longer using Bootstrap variables.
* Please make sure all your CSS code follows the guidelines provided on this wiki page https://wiki.one.int.sap/wiki/display/spar/UI+Styling+and+Accessibility+Onboarding 
* Please reach out to Spartacus Theme Owner if any doubts and to follow up with code review: @morganm58

developpeurweb (changes requested):
I see most of the CSS files are using Bootstrap variables, which we don't use anymore.
* Please use our `—cx-color` variables for colours and `@include type()` mixin for predefined fonts. We are no longer using Bootstrap variables.
* Please make sure all your CSS code follows the guidelines provided on this wiki page: https://wiki.one.int.sap/wiki/display/spar/UI+Styling+and+Accessibility+Onboarding 
* Please reach out to Spartacus Theme Owner if any doubts and to follow up with code review: @morganm58

feature-libs/user/profile/styles/_new-profile.scss
```diff
@@ -0,0 +1,57 @@
...
+    font-size: $font-size-base;
+    color: $primary;
+    border: none;
+    background-color: transparent;
+  }
+
+  .myaccount-enhancedUI-text-head {
+    font-weight: var(--cx-font-weight-semi);
+    min-width: 7.5rem;
+    color: $background-dark;
```
developpeurweb: * Please use our `—cx-color` variables for colours and `@include type()` mixin for predefined fonts. We are no longer using Bootstrap variables.

scarai-sap: Updated, Pls check

scarai-sap:
@developpeurweb , Please review it again

developpeurweb:
> @developpeurweb , Please review it again

@scarai-sap it will be @morganm58 taking over the CSS review, as he is the Spartacus Theme Owner.

feature-libs/user/profile/styles/my-account-v2/_my-account-v2-email.scss
```diff
@@ -0,0 +1,81 @@
...
+
+  .btn-group {
+    display: flex;
+    justify-content: space-between;
+    width: 40%;
+  }
+
+  form {
+    label {
+      padding-bottom: 0.75rem !important;
```
morganm58: Avoid using `!important`. Can you make the scss more specific so you won't need to use !important.

feature-libs/user/profile/styles/my-account-v2/_my-account-v2-password.scss
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  .myaccount-password-button-cancel {
+    color: var(--cx-color-primary);
+  }
+
+  .myaccount-password-label {
+    @include type();
+    min-width: 7.5rem;
+    padding: 0 1.2rem 0 0;
```
morganm58: Substitute `padding-left` and `padding-right` for  `padding-inline-start` and `padding-inline-start`. We need these for directionality to work.
`padding-top: 0` 
`padding-inline-end: 1.2rem`
`padding-bottom: 0`
`padding-inline-start: 0`

feature-libs/user/profile/styles/my-account-v2/_my-account-v2-profile.scss
```diff
@@ -0,0 +1,73 @@
...
+    width: 40%;
+  }
+
+  .ng-select .ng-select-container {
+    background-color: var(--cx-color-background);
+  }
+
+  form {
+    label {
+      padding-bottom: 0.75rem !important;
```
morganm58: Avoid using !important. Can you make the scss more specific so you won't need to use !important.

projects/storefrontstyles/scss/components/myaccount/_cx-my-account-v2-consent-management-form.scss
```diff
@@ -0,0 +1,57 @@
...
+  .description {
+    @include type('');
+  }
+
+  .consent-container {
+    display: flex;
+  }
+
+  .consent-details {
+    margin: 0 0 0 2.815rem;
```
morganm58: Substitute margin-left and margin-right for margin-inline-start and margin-inline-start. We need these for directionality to work.
padding-top: 0
padding-inline-end: 0
padding-bottom: 0
padding-inline-start: 2.815rem

projects/storefrontstyles/scss/components/myaccount/_cx-my-account-v2-consent-management-form.scss
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  .custom-switch {
+    margin: 0 0 0 -2.25rem;
```
morganm58: same comment as above

projects/storefrontstyles/scss/components/myaccount/_cx-my-account-v2-notification-preference.scss
```diff
@@ -0,0 +1,52 @@
+@import '../../theme';
+
+%cx-my-account-v2-notification-preference {
+  .np-content-center {
+    -ms-flex-pack: center !important;
```
morganm58: Avoid using `!important`.

projects/storefrontstyles/scss/components/myaccount/_cx-my-account-v2-notification-preference.scss
```diff
@@ -0,0 +1,52 @@
...
+    justify-content: center !important;
```
morganm58: Avoid using `!important` for all instances.

morganm58 (changes requested):
I added comments to  a few of the scss files.

scarai-sap:
@morganm58 , all comments fixed

developpeurweb (approved):
@Platonn LGTM.

morganm58 (approved):

PR #18364, opened 2024-01-12
feat: remove 3rd party typings from public API

Matejk00:
Closes: CXSPA-5634


FollowTheFlo (approved):
make sens to me from our discussion. Sonar and Checkmarx both OK
installation tested successfully on my local machine.
Just double check if moving types in package.json devDependencies instead of dependencies would also work which i believe you already tried.

PR #18372, opened 2024-01-15
chore: SSR contextual logging - changes before Spartacus 7.0

Matejk00:
Closes: CXSPA-3680


core-libs/setup/ssr/engine-decorator/ng-express-engine-decorator.spec.ts
```diff
@@ -185,22 +184,6 @@ describe('decorateExpressEngine', () => {
...
-          providers: [
-            {
-              provide: EXPRESS_SERVER_LOGGER,
-              useValue: new LegacyExpressServerLogger(),
-            },
-          ],
-        },
-        expect.any(Function)
-      );
-    });
```
pawelfras: `should:` Rather than removing the whole test case, we should revert to the previous shape:

```ts
it(`should pass parameters to the original engine instance`, () => {
    expect(originalEngineInstance).toHaveBeenCalledWith(
        mockPath,
        mockOptions,
        expect.any(Function)
    );
});
```

core-libs/setup/ssr/engine-decorator/ng-express-engine-decorator.spec.ts
```diff
@@ -261,22 +244,6 @@ describe('decorateExpressEngine', () => {
...
-
```
pawelfras: `should:` Rather than removing the whole test case, we should revert to the previous shape:

```ts
it(`should pass parameters to the original engine instance`, () => {
    expect(originalEngineInstance).toHaveBeenCalledWith(
        mockPath,
        mockOptions,
        expect.any(Function)
    );
});
```

core-libs/setup/ssr/logger/loggers/express-server-logger.ts
```diff
@@ -39,9 +39,8 @@ export interface ExpressServerLogger {
  * Injection token for ExpressServerLogger used for log message in server side rendering.
  * EXPRESS_SERVER_LOGGER is used to provide proper logger to LoggerService instance.
  *
- * Spartacus is providing two types of server loggers:
+ * Spartacus is providing one type of server logger:
```
pawelfras, line 42: :ok_hand:

core-libs/setup/ssr/optimized-engine/optimized-ssr-engine.ts
```diff
@@ -78,7 +77,7 @@ export class OptimizedSsrEngine {
           ...ssrOptions,
         }
       : undefined;
-    this.logger = this.initLogger(this.ssrOptions);
+    this.logger = new DefaultExpressServerLogger();
```
pawelfras: `should`: We should use the logger from the ssrOptions. In other cases, customers won't be able to provide their implementation.

core-libs/setup/ssr/optimized-engine/optimized-ssr-engine.ts
pawelfras, line 1: `should:` There is also one more place in `logOptions` (line 93) method that requires refactoring before the next-major. I've noticed it doesn't contain the ticket number. Apologies.

core-libs/setup/ssr/engine-decorator/ng-express-engine-decorator.spec.ts
Matejk00: added

core-libs/setup/ssr/engine-decorator/ng-express-engine-decorator.spec.ts
Matejk00: added

core-libs/setup/ssr/engine-decorator/ng-express-engine-decorator.spec.ts
pawelfras: As we discussed offline, we should keep this test without changes due to the logic in the optimized SSR engine.

core-libs/setup/ssr/engine-decorator/ng-express-engine-decorator.spec.ts
pawelfras: As above.

core-libs/setup/ssr/optimized-engine/ssr-optimization-options.ts
```diff
@@ -122,13 +122,12 @@ export interface SsrOptimizationOptions {
    * It enhances the logs in SSR by adding context, including the request's details,
    * and structuring them as JSON.
    *
-   * The `logger` property is optional and accepts two values:
-   * - `true`:         Enables the default logger and enhances the logs.
+   * The `logger` property is optional and accepts:
    * - `ExpressServerLogger`: Interprets the given `ExpressServerLogger` as a custom logger
    *
    * By default, the logger is disabled, meaning that logs in SSR are not enhanced.
```
pawelfras:
```suggestion
   * By default, the DefaultExpressServerLogger is used.
```

core-libs/setup/ssr/optimized-engine/optimized-ssr-engine.ts
```diff
@@ -78,7 +77,7 @@ export class OptimizedSsrEngine {
           ...ssrOptions,
         }
       : undefined;
-    this.logger = this.initLogger(this.ssrOptions);
```
pawelfras, line 81: `could:` After rethinking it once more, I think we can cast the type to `ExpressServerLogger` as we're sure the value won't be `undefined`

Matejk00: done

core-libs/setup/ssr/optimized-engine/optimized-ssr-engine.ts
```diff
@@ -78,7 +76,9 @@ export class OptimizedSsrEngine {
...
+    this.logger =
+      ssrOptions?.logger ??
+      (defaultSsrOptimizationOptions.logger as ExpressServerLogger); // we are sure the logger is defined in this place
```
pawelfras:
```suggestion
    this.logger =
      this.ssrOptions?.logger as ExpressServerLogger; // we are sure the logger is defined in this place
```

pawelfras (approved):
LGTM

PR #18374, opened 2024-01-15
refactor: Remove deprecated method in UserService (CXSPA-2638)

kimhw0630:
Closes: https://jira.tools.sap/browse/CXSPA-2638

Moved address book module from storefront library to user library then removed unused service (userService)


feature-libs/user/profile/components/user-profile-components.module.ts
```diff
@@ -22,6 +23,7 @@ import { UpdateProfileModule } from './update-profile/update-profile.module';
     ForgotPasswordModule,
     ResetPasswordModule,
     CloseAccountModule,
+    AddressBookModule,
```
kpawelczak, line 26: I wonder if this is the correct place to put `AddressBookModule`. The rest of the modules, match with their functionalities in one way or another. Also, I'm not sure if it matters but now when you load the address form it loads the whole profile components module (update profile, register, etc.).

projects/storefrontapp/src/app/spartacus/spartacus-features.module.ts
```diff
@@ -165,7 +164,6 @@ if (environment.requestedDeliveryDate) {
     UserModule,
     UserOccModule,
     // User UI
-    AddressBookModule,
```
kpawelczak, line 168: This import right here makes me wonder if this was intended for an AddressBook to be separate.

feature-libs/user/profile/components/user-profile-components.module.ts
kimhw0630: I moved to that location because the userProfileFacade was present. The objective was to eliminate UserService and utilize userProfileFacade.

kpawelczak (approved):

PR #18379, opened 2024-01-16
add feature toggle for changes to card component

Uli-Tiger:
@Larisa-Staroverova @ChristophHi : Started with the old approach, but i guess we need to change it to the new one as discussed here.
https://sap-cx.slack.com/archives/CAC8PULNN/p1705055286677549

Shall we consolidate all required changes regarding feature toggle in this PR?


ChristophHi (changes requested):
No that's not the way it is supposed to be.
See the slack conversation https://sap-cx.slack.com/archives/CAC8PULNN/p1705055286677549 please. We cannot bind the features to certain releases anymore, we want to behave like ccv2

Uli-Tiger:
@ChristophHi : updated the PR accordingly (with exception the scss file). - However, as you can see, the FEATURE_FLAG_QUOTES is always provided with 'true' i wonder how this shall be switched on/off.

projects/storefrontlib/feature-flags.ts
```diff
@@ -0,0 +1,11 @@
+/*
+ * SPDX-FileCopyrightText: 2024 SAP Spartacus team <spartacus-team@sap.com>
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { InjectionToken } from '@angular/core';
+
+export const FEATURE_FLAG_QUOTES = new InjectionToken<boolean>(
```
ChristophHi: It should rather be named FEATURE_FLAG_COMMENTS_PRODUCT_SUPPORT.
And in order to have these flags also available in core I suggest to put it into a new folder projects/core/src/features

ChristophHi (changes requested):
More comments in slack, I think Kris needs to agree. If my proposal is too simple the core team needs to provide the infrastructure first

PR #18382, opened 2024-01-17
fix: (CXSPA-1030) - Visible focus fixed on mobile/while zoomed

Pio-Bar:
https://jira.tools.sap/browse/CXSPA-1030

Since some images on the storefront span the entire screen width, making an inset outline for mobile seems to be an elegant solution. Otherwise, it will always get clipped. Let me know if this is a viable option.


projects/storefrontstyles/scss/cxbase/mixins/_visible-focus.scss
```diff
@@ -4,4 +4,15 @@
   outline-width: var(--cx-visual-focus-width, 2px);
   outline-offset: 4px;
   transition: none;
+
+  // TODO: Remove feature flag next major
```
Zeyber: We should include the cleanup ticket #.

Zeyber (changes requested):

projects/storefrontstyles/scss/cxbase/mixins/_visible-focus.scss
Pio-Bar: Done: https://jira.tools.sap/browse/CXSPA-5912

Zeyber (dismissed):

developpeurweb (dismissed):

projects/storefrontlib/layout/a11y/keyboard-focus/visible/visible-focus.directive.ts
```diff
@@ -22,6 +29,14 @@ import { VisibleFocusConfig } from '../keyboard-focus.model';
  */
 @Directive() // selector: '[cxVisibleFocus]'
 export class VisibleFocusDirective extends BaseFocusDirective {
+  constructor(
+    protected elementRef: ElementRef<HTMLElement>,
+    protected service: BaseFocusService
+  ) {
+    super(elementRef, service);
```
Zeyber, line 37: I think this call to super could be a breaking change.

Zeyber (changes requested):

projects/storefrontlib/layout/a11y/keyboard-focus/visible/visible-focus.directive.ts
Pio-Bar: This should be fine since we only need this to enable the feature flag. The class behaviour stays the same, and we do not intend on keeping those changes. I have added a comment to make sure the constructor gets removed. :)

Zeyber (approved):

PR #18387, opened 2024-01-18
feat: add opfAddToCart cms cpnt

FollowTheFlo:
created a new cms component: OpfAddToCartComponent
On Spar, OpfAddToCartComponent extends standard AddToCartComponent with 'opf-quick-buy' selector in its template.
on backoffice, OpfAddToCartComponent is set in ProductSummary content slot.
image


package.json
```diff
@@ -85,6 +85,7 @@
     "start:ci": "env-cmd --no-override -e ci,b2c,$SPA_ENV nx serve storefrontapp --configuration=development",
     "start:ci:b2b": "env-cmd --no-override -e ci,b2b,$SPA_ENV nx serve storefrontapp --configuration=development",
     "start:opf": "env-cmd --no-override -e opf,$SPA_ENV nx serve storefrontapp --configuration=development --ssl",
+    "start:opf:vm": "env-cmd --no-override -e opf,$SPA_ENV nx serve storefrontapp --configuration=development",
```
FollowTheFlo, line 88: need to run opf without ssl on azure vm to test ApplePay,. VM already has ssl with nginx.

integration-libs/opf/base/components/opf-add-to-cart/opf-add-to-cart.component.html
FollowTheFlo, line 1: this file is s copy of` add-to-cart.component.ts` with  `<cx-opf-quick-buy></cx-opf-quick-buy>`

Matejk00: The question at hand is whether we should duplicate the content of this "add to cart" template. This would result in two identical copies of the same file, posing potential challenges in maintaining the logic of both files in the future.

An alternative approach could be to introduce an additional outlet position specifically for Quick Buy in the "add to cart" component. This modification is not expected to be a breaking change. For instance:

```
<ng-container *ngIf="hasStock">
  <ng-template [cxOutlet]="CartOutlets.QUICK_BUY"></ng-template>
</ng-container>
```

With this change, the QB module only needs to inject a component into the specified position:

```
providers: [
  provideOutlet({
    id: CartOutlets.QUICK_BUY,
    position: OutletPosition.REPLACE,
    component: OpfQuickBuyComponent,
  }),
],
```
In my view, this approach should be effective. However, I am open to considering any circumstances or considerations that I may not be aware of.

Matejk00 (changes requested):

PR #18388, opened 2024-01-19
feature/CXINT-2714: Display Promotions in PDP

anjana-bl:
https://jira.tools.sap/browse/CXINT-2743


projects/storefrontlib/cms-components/product/product-summary/product-summary.component.ts
```diff
@@ -16,12 +16,20 @@ import { ProductDetailOutlets } from '../product-outlets.model';
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
 export class ProductSummaryComponent {
+  protected featureConfig = inject(FeatureConfigService);
+
   outlets = ProductDetailOutlets;
+  showPromotions = this.featureConfig.isEnabled('showPromotionsInPDP');
```
Platonn: [MEDIUM]
If it's a temporary flag, but not a regular config, then let's not expose additionally unnecessary public API members that would need to be removed as well together with a flag.

So let's remove the class property, but instead use the expression `this.featureConfig.isEnabled('showPromotionsInPDP')` directly inside the method `getProdut()`

Note: in HTML we can use a directive `*cxFeature=""`, e.g. `<cx-promotions *cxFeature="'showPromotionsInPDP'">`

anjana-bl: done

Platonn (approved):
LGTM

PR #18389, opened 2024-01-19
chore: Fix add header workflow by updating deprecated addheader to annotate

Zeyber:
Closes: https://jira.tools.sap/browse/CXSPA-5866


pawelfras:
Hi @Zeyber,
Let's add to this PR a test file to see whether the pipeline still does the job (we can remove it as soon as the workflow is launched and the licence header is added)

pawelfras (approved):

PR #18390, opened 2024-01-19
fix : resolve type incongruities (CXSPA-3088)

kimhw0630:
Closes: https://jira.tools.sap/browse/CXSPA-3088


FollowTheFlo (approved):
good to me

PR #18391, opened 2024-01-19
fix: (CXSPA-1031) - Map overflow on small/zoomed screens

Pio-Bar:
https://jira.tools.sap/browse/CXSPA-1031

The map no longer obscures the reported element, however it still overflows on 320px screen size / 200% zoom.  
This change should prevent that.


feature-libs/storefinder/components/store-finder-store-description/store-finder-store-description.component.html
```diff
@@ -1,7 +1,7 @@
 <ng-container *ngIf="location">
   <div class="container">
     <div class="row">
-      <article class="cx-store col-md-4">
+      <article class="cx-store col-lg-4">
```
Zeyber: Class changes/removals are also considered breaking changes so we still need to address this here.

NOTE: The version these changes will go into will be different than `6.8`, but unfortunately we still don't know what it will be. Just a heads up if you weren't already aware.

Zeyber: I also see some opportunity for clean up here and on line `63`. We would like to remove css bootstrap "one day". But that will be breaking change armageddon at this point. 

We can however start working towards that goal by leveraging those `cx-store` and `cx-storeMap` classes to have the breakpoints in scss instead of relying on those `col-lg-X` bootstrap classes.

Zeyber (changes requested):

Zeyber (commented):
One thing to note is that the PR names should follow convention that our documentation script uses else we will have missing changes in our changelogs.

Example:
- `fix: ...` for something that fixes a bug but doesn't really add functionality.
- `feat: ...` adds functionality on top of existing.
- `chore: ...` changes to workflow or something that doesn't affect the code.
This is my rough explanation but there should be the list in the DoD docs (I think).

feature-libs/storefinder/components/store-finder-store-description/store-finder-store-description.component.html
Pio-Bar: I got rid of bootstrap here altogether and did some clean up. If this is too out of scope let me know!

Pio-Bar: Done. I have left the version as 6.8 for now, once we have the proper number it will probably need to be changed in all the tickets with feature flags.

feature-libs/storefinder/styles/components/_store-finder-store-description.scss
```diff
@@ -49,14 +70,22 @@
     text-align: start;
   }
 
-  .cx-feature-value {
```
Zeyber, line 52: I think we should still keep these `feature-value feature-item` classes for older versions and clean them up later.

feature-libs/storefinder/styles/components/_store-finder-map.scss
```diff
@@ -3,9 +3,19 @@
     width: 100%;
     height: 100%;
 
-    @include media-breakpoint-down(sm) {
-      height: 30rem;
-      width: 100%;
+    // TODO:  Remove feature flag next major
```
Zeyber: Really just a nitpick to add the cleanup ticket number. It probably won't be you doing this cleanup in the future so it really helps ;)

Zeyber (changes requested):

feature-libs/storefinder/styles/components/_store-finder-store-description.scss
Pio-Bar: I've included them in the feature flag

feature-libs/storefinder/styles/components/_store-finder-map.scss
Pio-Bar: Done :)

Zeyber (dismissed):

developpeurweb (dismissed):

Zeyber (dismissed):

Zeyber (approved):

PR #18392, opened 2024-01-19
fix: profile tag event service is not working with the SSR

kpawelczak:
closes: [CXSPA-5867](https://jira.tools.sap/browse/CXSPA-5867)


FollowTheFlo (approved):
as discussed, it is good too me, we only make sure sure values are not undefined, without business logic change.

PR #18393, opened 2024-01-22
Quote list: The tooltip should only contain the truncated name of the quote

Larisa-Staroverova:


feature-libs/quote/components/list/quote-list.component.html
```diff
@@ -71,7 +71,7 @@
                 } | cxUrl
               "
               (keydown.enter)="row.click()"
-              attr.title="{{ 'quote.list.name' | cxTranslate }}: {{
+              attr.aria-lable="{{ 'quote.list.name' | cxTranslate }}: {{
```
steinsebastian: Typo: aria-lable vs aria-label
Currently it does not read the quote info but after fixing typo this works

steinsebastian (changes requested):

steinsebastian (approved):

PR #18399, opened 2024-01-22
feat: Add OPF active cart support for Google Pay

Matejk00:
Closes: CXSPA-5550

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -56,6 +56,7 @@ scripts/install/config.sh
 /connect.lock
 /libpeerconnection.log
 /typings
+/.nx
 
 # Logs
 npm-debug.log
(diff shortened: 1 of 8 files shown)


.gitignore
```diff
@@ -56,6 +56,7 @@ scripts/install/config.sh
 /connect.lock
 /libpeerconnection.log
 /typings
+/.nx
```
FollowTheFlo, line 59: is it intended? or some left over config from your local env?

Matejk00: This change is intended, not sure why, but some files in **.nx/cache** folder appears during the development.

FollowTheFlo (approved):
tested it without problems. code is good to me.
UI/buttons alignment can be addressed in diff task.

PR #18400, opened 2024-01-22
feat: use recent searches (#18400)

lasoh:
CXCDS-8708 use recent searches


integration-libs/cds/src/components/recent-searches/recent-searches.service.ts
```diff
@@ -0,0 +1,44 @@
...
+export class RecentSearchesService {
+  private recentSearchesSource = new ReplaySubject<string[]>();
+  private apiAvailability = false;
+  public get recentSearches$(): Observable<string[]> {
+    this.addRecentSearchesListener();
+    return this.recentSearchesSource.asObservable();
+  }
+  checkAvailability() {
+    return interval(150).pipe(
+      concatMap((_) => of((<any>window).Y_TRACKING)),
```
Platonn: [MEDIUM] using `window` might break SSR

Note : review whole PR for usage of global objects like window or document

integration-libs/cds/src/components/recent-searches/recent-searches.service.ts
```diff
@@ -0,0 +1,44 @@
...
+import { interval, Observable, of, ReplaySubject } from 'rxjs';
+import { concatMap, endWith, take, takeWhile } from 'rxjs/operators';
+
+@Injectable()
+export class RecentSearchesService {
+  private recentSearchesSource = new ReplaySubject<string[]>();
+  private apiAvailability = false;
+  public get recentSearches$(): Observable<string[]> {
+    this.addRecentSearchesListener();
+    return this.recentSearchesSource.asObservable();
```
Platonn: [MEDIUM]
Prone to memory leaks. Here we're creating a subscription every time a getter `recentSearches$()` is accessed.
Did you consider alternative solutions, using `async` pipe and not doing manual subscriptions?

integration-libs/cds/assets/translations/en/cds-recentSearches.ts
Platonn, line 1: [MEDIUM]
Please put translations resources into JSON files, and then import into `index.ts` file - similar as in other libraries.

integration-libs/cds/src/components/recent-searches/recent-searches.module.ts
```diff
@@ -0,0 +1,32 @@
...
+import { I18nModule, UrlModule } from '@spartacus/core';
+import { RecentSearchesComponent } from './recent-searches.component';
+import { RecentSearchesService } from './recent-searches.service';
+
+@NgModule({
+  exports: [RecentSearchesComponent],
+  declarations: [RecentSearchesComponent],
+  imports: [CommonModule, I18nModule, SearchBoxModule, UrlModule, RouterModule],
+  providers: [
+    RecentSearchesService,
```
Platonn: [QUESTION]
is there any reason we're providing a service in module, but not in doing
`@Injectable({providedIn: 'root'}`?

integration-libs/cds/src/components/recent-searches/recent-searches.module.ts
```diff
@@ -0,0 +1,32 @@
...
+    provideOutlet({
+      id: 'RECENT_SEARCHES',
```
Platonn: [MINOR]
Please note that outlet names share a global namespace. So the phrase "recent searches" might be also relevant in some other search functionalites (e.g. search users, etc.) - so let's scope it to "search box"

Also to avoid magic strings, for let's move the value to an enum exported in our public api.

e.g.
```ts
export enum SearchBoxOutlets {
   RECENT_SEARCHES: 'SearchBoxOutlets.RECENT_SEARCHES`
};
```

integration-libs/cds/src/components/recent-searches/recent-searches.service.ts
lasoh: @Platonn do You have any idea what we can use in that case? If we know the Y_TRACKING (recentSeraches) is on `window`. Maybe some proposition?

integration-libs/cds/src/components/recent-searches/recent-searches.module.ts
lasoh: You're right! 👍

integration-libs/cds/src/components/recent-searches/recent-searches.service.ts
lasoh: I know, but we use recentSeraches$ in 
`switchMap` and after that we use asyncPipe inside `recent-searches-component` template (`result$`)

integration-libs/cds/assets/translations/en/cds-recentSearches.ts
lasoh: OK 👍 -> changed

integration-libs/cds/src/components/recent-searches/recent-searches.service.ts
lasoh: I changed to `windowRef from @spartacus/core` instead of `window` -> What do you think?

integration-libs/cds/src/components/recent-searches/recent-searches.component.html
```diff
@@ -0,0 +1,40 @@
+<!--
```
rmch91: You don't have to add a licence header to html files

integration-libs/cds/src/components/recent-searches/recent-searches.component.html
```diff
@@ -0,0 +1,40 @@
...
+  ~ SPDX-FileCopyrightText: 2024 SAP Spartacus team <spartacus-team@sap.com>
+  ~
+  ~ SPDX-License-Identifier: Apache-2.0
+  ~
+  -->
+
+<ng-container *ngIf="(result$ | async)?.length">
```
rmch91: To avoid multiple subscription to `result$` obs, you can use here `*ngIf="(result$ | async) as results"`
and add additional `ng-container` with `*ngIf="results?.length"`
So in line 17 you can also use `*ngFor="let recentSearch of results"` without subscribing with async pipe.

integration-libs/cds/src/components/recent-searches/recent-searches.component.ts
```diff
@@ -0,0 +1,68 @@
...
+              (phrase) =>
+                phrase.toLowerCase().indexOf(context.search.toLowerCase()) >= 0
+            )
+            .slice(0, context.maxRecentSearches ?? MAX_RECENT_SEARCHES);
+        })
+      );
+    })
+  );
+
+  outletContext$: Observable<SearchBoxOutlet>;
```
rmch91, line 51: missing spacing below

integration-libs/cds/src/components/recent-searches/recent-searches.component.ts
```diff
@@ -0,0 +1,68 @@
...
+  constructor(
+    @Optional() protected outletContext: OutletContextData<SearchBoxOutlet>,
+    protected winRef: WindowRef
+  ) {}
```
rmch91, line 55: missing spacing below

integration-libs/cds/src/components/recent-searches/recent-searches.component.ts
```diff
@@ -0,0 +1,68 @@
...
+  ngOnInit() {
+    this.outletContext$ = this.outletContext.context$;
+  }
+  preventDefault(ev: UIEvent): void {
```
rmch91, line 61: missing spacing below

integration-libs/cds/src/components/recent-searches/recent-searches.component.ts
```diff
@@ -0,0 +1,68 @@
...
+  search: string;
+  searchBoxActive: boolean;
+  maxRecentSearches?: number;
+}
+
+const MAX_RECENT_SEARCHES = 5;
+@Component({
+  selector: 'cx-recent-searches',
+  templateUrl: './recent-searches.component.html',
+})
```
rmch91, line 33: Could we use `changeDetection: ChangeDetectionStrategy.OnPush,` here?

integration-libs/cds/src/components/recent-searches/recent-searches.service.ts
rmch91: I'm not sure about the context, but what i understand from code, in `addRecentSearchesListener()` method, if `result` or `recentPhrases` will be `falsy`, this line `this.apiAvailability = true` will not be executed, thats mean that each call of `addRecentSearchesListener()` will create new subscription in memory. I think if you need to get value only one time from `this.checkAvailability()` it should be enough to add `take(1)`

integration-libs/cds/src/components/recent-searches/recent-searches.service.ts
```diff
@@ -0,0 +1,54 @@
...
+    );
+  }
+
+  private addRecentSearchesListener() {
+    if (!this.apiAvailability) {
+      this.checkAvailability().subscribe((result) => {
+        if (result) {
+          const recentPhrases = (<any>(
+            this.winRef.nativeWindow
+          )).Y_TRACKING?.recentSearches?.getPhrases();
```
rmch91: Please add spacing below

projects/storefrontlib/cms-components/navigation/search-box/search-box.component.ts
```diff
@@ -107,7 +113,7 @@ export class SearchBoxComponent implements OnInit, OnDestroy {
   );
 
   ngOnInit(): void {
-    this.subscription = this.routingService
+    const routeStateSubscribtion = this.routingService
```
rmch91: typo in the variable name -> `routeStateSubscription`

projects/storefrontlib/cms-components/navigation/search-box/search-box.component.ts
```diff
@@ -120,6 +126,21 @@ export class SearchBoxComponent implements OnInit, OnDestroy {
           this.chosenWord = '';
         }
       });
```
rmch91, line 128: Please add spacing below

projects/storefrontlib/cms-components/navigation/search-box/search-box.component.ts
```diff
@@ -120,6 +126,21 @@ export class SearchBoxComponent implements OnInit, OnDestroy {
...
+    this.subscriptions.add(routeStateSubscribtion);
+
+    const chosenWordSubsribtion =
```
rmch91: also typo in the name as above

projects/storefrontlib/cms-components/navigation/search-box/search-box.component.ts
```diff
@@ -120,6 +126,21 @@ export class SearchBoxComponent implements OnInit, OnDestroy {
...
+      this.searchBoxComponentService.chosenWord.subscribe((chosenWord) => {
+        this.updateChosenWord(chosenWord);
+      });
```
rmch91, line 135: Please add spacing below

rmch91 (changes requested):

integration-libs/cds/src/components/recent-searches/recent-searches.component.html
lasoh: 👍 -> my ide added  -> sorry ->  changed

integration-libs/cds/src/components/recent-searches/recent-searches.service.ts
rmch91: WindowRef.nativeWindow has a type of `Window | undefined`. On the server it will return `undefined`.  So you need to add `?` before `.Y_TRACKING)`. Also, if `window` will be undefined, i think it will also throw error on `takeWhile` when will try to access `recentSearches`, so probably you need also `result?.recentSearches` here.

integration-libs/cds/src/components/recent-searches/recent-searches.component.html
lasoh: good point -> thanks

integration-libs/cds/src/components/recent-searches/recent-searches.component.ts
lasoh: 👍

integration-libs/cds/src/components/recent-searches/recent-searches.component.ts
lasoh: 👍

integration-libs/cds/src/components/recent-searches/recent-searches.component.ts
lasoh: 👍

integration-libs/cds/src/components/recent-searches/recent-searches.component.ts
lasoh: of course :)

integration-libs/cds/src/components/recent-searches/recent-searches.service.ts
rmch91: I'm not sure i also understand correctly the propose for` endWith(true)`. If after 5th attempt Y_TRACKING will still be `undefined`, it will return `true`. Thats mean that we will try to execute this code 
```
    const recentPhrases = (<any>(
            this.winRef.nativeWindow
          )).Y_TRACKING?.recentSearches?.getPhrases();
```

but we don't have to, if we already know that Y_TRACKING is `undefined`.

integration-libs/cds/src/components/recent-searches/recent-searches.service.ts
```diff
@@ -0,0 +1,54 @@
...
+          if (recentPhrases) {
+            this.recentSearchesSource.next(recentPhrases);
+
+            (<any>window).Y_TRACKING.recentSearches?.addListener(
```
rmch91: Probably leftover, pointing to `window` instead of `winRef`

rmch91 (changes requested):

integration-libs/cds/src/components/recent-searches/recent-searches.service.ts
lasoh: It's not working with `take(1)` :(

projects/storefrontlib/cms-components/navigation/search-box/search-box.component.ts
lasoh: 👍

projects/storefrontlib/cms-components/navigation/search-box/search-box.component.ts
lasoh: 👍

projects/storefrontlib/cms-components/navigation/search-box/search-box.component.ts
lasoh: 👍

integration-libs/cds/src/components/recent-searches/recent-searches.service.ts
lasoh: yeah, eagle eye :) thanks

integration-libs/cds/src/components/recent-searches/recent-searches.service.ts
```diff
@@ -18,11 +18,12 @@ export class RecentSearchesService {
...
-    return this.recentSearchesSource.asObservable();
+    return this.recentSearchesSource;
   }
+
   checkAvailability() {
     return interval(150).pipe(
-      concatMap((_) => of((<any>this.winRef.nativeWindow).Y_TRACKING)),
+      concatMap((_) => of((<any>this.winRef.nativeWindow)?.Y_TRACKING)),
       take(5),
       takeWhile((result: any) => !result.recentSearches),
```
rmch91: One thing commented earlier, if this part `of((<any>this.winRef.nativeWindow).Y_TRACKING))` will return `undefined` i think it will also throw error on takeWhile when will try to access `recentSearches`, so probably you need also `result?.recentSearches` here

integration-libs/cds/src/components/recent-searches/recent-searches.service.ts
```diff
@@ -18,11 +18,12 @@ export class RecentSearchesService {
...
       endWith(true)
```
rmch91: https://github.com/SAP/spartacus/pull/18400#discussion_r1479795911

rmch91 (changes requested):

integration-libs/cds/src/components/recent-searches/recent-searches.service.ts
rmch91: What do you think about this implementation:

```
checkAvailability() {
   return interval(150).pipe(
    concatMap(() => of((<any>this.winRef.nativeWindow)?.Y_TRACKING)),
    map(result => !!result?.recentSearches),
    take(5),
    takeWhile(val => !val, true),
    last(),
}
```

- `takeWhile(val => !val, true)`: This operator is now used to continue taking values until a `truthy` one is found, but importantly, it includes the `inclusive` parameter set to `true`. This means it will emit the last checked value (whether true or false) when it stops taking values. This change is crucial because it ensures that if `Y_TRACKING.recentSearches` is found on the first attempt, the observable will complete correctly after emitting `true`.

- `last()`: Ensures that the last emitted value is taken. This is needed to grab the last value emitted by `takeWhile` — either the first occurrence of `true` or `false` if `Y_TRACKING.recentSearches` is not found within 5 attempts.

This way `checkAvailability()` will emit only one value either `true` or `false`.

In your implementation, if recentSearches will be undefined, if you log values in subscribe you will see 5 logs of undefined and 1 true. Two problems here,  that we are not interested in this undefined values and value `true` in this case should be rather `false`

integration-libs/cds/src/components/recent-searches/recent-searches.service.ts
rmch91: Yeah, now i see. `this.checkAvailability()` could emit values up to 6 times. I don't think this is good idea, cause we are not actually interested in intermediate values but only in final value (that will say is `recentSearches` available or not). 

Probably should work if you implement this one https://github.com/SAP/spartacus/pull/18400#discussion_r1481421048

Please check and let me know.

integration-libs/cds/src/components/recent-searches/recent-searches.service.ts
lasoh: ok, I can secure this element (`? lastSearches ) and as for the whole checkAvailability method, it doesn't work like "infinite check", but it only tries to get data from Y_TRACKING API once in max 5 intervals, and if this API doesn't exists, you should not call it endlessly.

integration-libs/cds/src/components/recent-searches/recent-searches.module.ts
rmch91: As far as i can see, it's already changed. So I'm resolving this threat

integration-libs/cds/src/components/recent-searches/recent-searches.service.ts
```diff
@@ -0,0 +1,57 @@
...
+@Injectable({ providedIn: 'root' })
+export class RecentSearchesService {
+  private readonly recentSearchesSource = new ReplaySubject<string[]>();
+  private apiAvailability = false;
+
+  constructor(protected winRef: WindowRef) {}
+
+  public get recentSearches$(): Observable<string[]> {
+    this.addRecentSearchesListener();
+    return this.recentSearchesSource;
```
rmch91: I'm not sure why you removed `.asObservable();` in previous commits. Now it is possible to call `.next` on `recentSearchesSource` outside of the service, it should be rather encapsulated. And in this case u declared return type as `Observable<string[]>` but in fact this getter returns `ReplaySubject<string[]>`. Let's return proper type and transform replaySubject to observable, as it was before.

integration-libs/cds/src/components/recent-searches/recent-searches.service.ts
```diff
@@ -0,0 +1,57 @@
...
+
+@Injectable({ providedIn: 'root' })
+export class RecentSearchesService {
+  private readonly recentSearchesSource = new ReplaySubject<string[]>();
+  private apiAvailability = false;
+
+  constructor(protected winRef: WindowRef) {}
+
+  public get recentSearches$(): Observable<string[]> {
+    this.addRecentSearchesListener();
```
rmch91: Calling a side-effect method like `addRecentSearchesListener()` inside a getter is generally not considered a best practice for several reasons:

1. **Unexpected Side Effects:** Getters are typically used to access properties or computed values and are expected to be free of side effects. When invoking a getter, consumers do not usually expect it to modify the state or initiate processes such as setting up listeners or fetching data. This can lead to unexpected behavior and makes the code harder to understand and investigate.
2.  **Repeated Calls:** Each time the `recentSearches$` getter is accessed, `addRecentSearchesListener()` is called. And this method sets up event listeners, add subscriptions, calling it multiple times can lead to multiple listeners or subscriptions being created, potentially causing memory leaks or other unintended consequences.

IMO, this method should be called outside of getter, even if it will be a public method and called from the component.

integration-libs/cds/src/components/recent-searches/recent-searches.service.ts
rmch91: Fixed

integration-libs/cds/src/components/recent-searches/recent-searches.service.ts
rmch91: Fixed

integration-libs/cds/src/components/recent-searches/recent-searches.service.ts
rmch91: We still have unsubscribed `Subscription`, it's always a red flag for me, you are calling this method `this.addRecentSearchesListener()` by accessing `results` with `async` pipe in component, so it will call this method each time when angular will trigger the change detection for this component. As far as i see, lifecycles of the component and this service are not the same. If component will be destroyed, service will still exist, and after new component creation, new call to service could create new subscriptions.

Yes, you have a guard `this.apiAvailability`, but as i described above, if this.checkAvailability() will return falsy value, this flag will be not set to true, that's mean the new call to this method will create new Subscription.

What you could possible do here:

1. use take(1)
2. Create a guard, meaning assign this `subscription` to class member, and before each call of `this.checkAvailability().subscribe()` check if subscription exist, if yes -> unsubscribe before creating the new one.
3. Convert this method, and not call `subscribe()` inside of it, but return observable instead of it. In this case you will able to use chaining in the component and in first `switchMap` you can call `addRecentSearchesListener` and receive for example true if all is good and listeners were created successfully, and in second switchMap use `this.recentSearchesService.recentSearches$`. Thanks to this, you will know that api available, and listeners created, before access too recentSearches$ and all subscriptions will be covered by `async` pipe

Right now you are accessing `recentSearches$` even before listeners created.

```
public get recentSearches$(): Observable<string[]> {
    this.addRecentSearchesListener(); // Could take same time because of intervals
    return this.recentSearchesSource; // Will return values instantly
}
```

TL;DR:

Please add a mechanism that will save us from even the slightest possibility of creating memory leaks.

rmch91 (changes requested):

rmch91 (dismissed):

integration-libs/cds/src/components/recent-searches/recent-searches.service.ts
```diff
@@ -0,0 +1,51 @@
...
+@Injectable({ providedIn: 'root' })
+export class RecentSearchesService {
+  private readonly recentSearchesSource = new ReplaySubject<string[]>();
+  public recentSearches$ = this.recentSearchesSource.asObservable();
+  constructor(protected winRef: WindowRef) {
+    this.addRecentSearchesListener();
+  }
+
+  checkAvailability() {
+    return interval(250).pipe(
```
kpawelczak, line 23: In SSR this interval here will slow down the rendering of the page by 2.5s -> 250ms * 100 (take(100)). This delay is likely to cause the SSR engine to timeout and fail to render the page altogether. I would suggest to wrap the whole function `addRecentSearchesListener` in `this.winRef.isBrowser()`.

```typescript
if (this.winRef.isBrowser()) {
this.addRecentSearchesListener();
}
```

integration-libs/cds/src/components/recent-searches/recent-searches.service.ts
```diff
@@ -0,0 +1,51 @@
...
+
+@Injectable({ providedIn: 'root' })
+export class RecentSearchesService {
+  private readonly recentSearchesSource = new ReplaySubject<string[]>();
+  public recentSearches$ = this.recentSearchesSource.asObservable();
+  constructor(protected winRef: WindowRef) {
+    this.addRecentSearchesListener();
+  }
+
+  checkAvailability() {
```
kpawelczak: I'm wondering whether this function should be public. Is it utilized anywhere beyond this service?

kpawelczak (changes requested):

PR #18409, opened 2024-01-23
feat: Memory leak if no cacheSize defined (CXSPA-5497)

kimhw0630:
Closes: https://jira.tools.sap/browse/CXSPA-5497

### DIFF
--- a/core-libs/setup/ssr/optimized-engine/ssr-optimization-options.ts
+++ b/core-libs/setup/ssr/optimized-engine/ssr-optimization-options.ts
@@ -137,6 +137,7 @@ export enum RenderingStrategy {
 }
 
 export const defaultSsrOptimizationOptions: SsrOptimizationOptions = {
+  cacheSize: 3000, // 1.5GiB / 525KB (50% of minimal pod size 3GiB / 150% of OOTB page ~350KB)
   concurrency: 10,
   timeout: 3_000,
   forcedSsrTimeout: 60_000,
--- a/core-libs/setup/ssr/optimized-engine/ssr-optimization-options.ts
+++ b/core-libs/setup/ssr/optimized-engine/ssr-optimization-options.ts
@@ -33,6 +33,8 @@ export interface SsrOptimizationOptions {
    * Can also be use when `cache` option is set to false. It will then limit the
    * number of renders that timeouts and are kept in temporary cache, waiting
    * to be served with next request.
+   *
+   * Default value is set to 3000.
    */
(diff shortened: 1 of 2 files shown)


core-libs/setup/ssr/optimized-engine/ssr-optimization-options.ts
```diff
@@ -137,6 +137,7 @@ export enum RenderingStrategy {
 }
 
 export const defaultSsrOptimizationOptions: SsrOptimizationOptions = {
+  cacheSize: 3000, // 1.5GiB / 525KB (50% of minimal pod size 3GiB / 150% of OOTB page ~350KB)
```
pawelfras: `should:` instead of a comment in this place, let's adjust the description of the jsdoc for this parameter. We can write there what is the default `cacheSize and how we defined such a value by describing a bit the calculations that have been made.

By doing it that way, customers will see all the important information when they want to modify this parameter in their applications.

core-libs/setup/ssr/optimized-engine/ssr-optimization-options.ts
```diff
@@ -137,6 +137,18 @@ export enum RenderingStrategy {
...
+   * Specifies the maximum number of entries to be retained in the cache,
+   * helping to manage memory usage effectively. The calculation is based on the formula:
+   *
+   * Max Entries = 1.5 GiB / 525 KB
+   *
+   * Where:
+   * - 1.5 GiB: Calculated cache size, derived from half of the minimal pod size.
+   * - 525 KB: Assumption that customer pages may have a size of up to 150% of the Out-of-the-Box (OOTB) page size.
+   * Therefore, considering the OOTB page size as ~350 KB, the assumed customer page size is 525 KB.
+   */
```
pawelfras: I was thinking more about the jsdoc extension that already exists in the `SsrOptimizationOptions` interface (lines 27-36). Now I'm thinking that maybe we don't need to provide details of our calculations for the customers. If yes, it can be adjusted later after discussion with @Platonn.
So the jsdoc after adjustment could look like this:
```ts
 /**
   * Limit the cache size
   *
   * Specified number of entries that will be kept in cache, allows to keep
   * memory usage under control.
   *
   * Can also be use when `cache` option is set to false. It will then limit the
   * number of renders that timeouts and are kept in temporary cache, waiting
   * to be served with next request.
   * 
   * Default value is set to 3000.
   */
  cacheSize?: number;

pawelfras (approved):

PR #18426, opened 2024-01-29
fix: Make missing actions column header visible in saved cart list

Pawulon-0:
Closes: https://jira.tools.sap/browse/CXSPA-1185


feature-libs/cart/saved-cart/styles/_saved-cart-list.scss
```diff
@@ -128,6 +128,7 @@ cx-saved-cart-list {
 
     .cx-saved-cart-list-label {
       color: var(--cx-color-secondary);
+      text-align: start;
```
Zeyber: Because we are adding a new style to an existing class, we need to account for breaking changes (ie. some code a user may be using that may change unexpectedly in an upgrade. In this example, they may be using the `cx-saved-cart-list-label` class). 

To account for styling breaking changes, we use `@include forVersion(/* next Spartacus version eg. 6.8 */)` and wrap new changes in this block. This allows customers to be able to "opt-in" to new styling changes by changing their style version from their current version up to the next version (eg. changing their version UP to 6.8 would enable 6.8 version styles.

[Here is an example of using this block](https://github.com/SAP/spartacus/pull/18382/files)

Zeyber (changes requested):
"Breaking Changes" will be a recurring issue in this project. Don't worry, we will work through many examples of this over time. [Here](https://sap.github.io/spartacus-docs/breaking-changes/) is a brief overview of when to notice the change is a breaking change and we will resolve each situation as it occurs until you get the hang of it ;)

A few more points:
- PR titles should be descriptive and follow [the guideline](https://sap.github.io/spartacus-docs/breaking-changes/) (eg. `fix: Make missing actions column visible in saved cart`)
- Please include "Closes: /* link to ticket */" in the PR description.

Zeyber:
In order to test your PR to make sure it is running without issues, hit "Ready for review" to enable pipeline checks (eg. E2E tests, unit tests, link and sonar checks, etc). If this is all green in the end, you are good to go for review. If there is an issue, you can check the results to see what might be wrong with your code.

image

Zeyber (dismissed):

developpeurweb (changes requested):
Since the link is in fact a button without an outline it looks misaligned. Therefore, can you please add a conditional alignment on the label for both scenarios, 320px and 200% zoom. As per screenshot aligning center on flex might fix it.

320px screen


200% zoom

developpeurweb (dismissed):

feature-libs/cart/saved-cart/styles/_saved-cart-list.scss
```diff
@@ -129,6 +129,11 @@ cx-saved-cart-list {
     .cx-saved-cart-list-label {
       color: var(--cx-color-secondary);
 
+      // TODO: (CXSPA-5955) Remove feature flag next major
+      @include forVersion(6.8) {
```
Zeyber: Styling changes here will need to follow the new [style versioning method](https://github.com/SAP/spartacus/pull/18523).

Zeyber (changes requested):

feature-libs/cart/saved-cart/styles/_saved-cart-list.scss
Pio-Bar: Changed for the new method.

Zeyber (dismissed):

Zeyber (approved):

PR #18428, opened 2024-01-29
chore: Fix config:check script checking dependencies.json on pipeline

Zeyber:
Closes: https://jira.tools.sap/browse/CXSPA-5789


tools/config/index.ts
```diff
@@ -287,7 +287,7 @@ if (options.generateDeps) {
   execSync(`npm run generate:deps`);
 } else {
   // collect and generate dependencies.json file.
-  execSync(`npm run generate:deps --compare=true`);
+  execSync(`npm run generate:deps:compare`);
```
pawelfras: Ah... so it looks like a regression bug related to switching to `npm` from `yarn`. The functionality was there before and stopped working due to differences in CLi syntax for passing additional parameters to a script.

We can avoid adding a new `generate:deps:compare` script to our `package.json` by adding a tiny change (--) to the previous version of the script invocation in this line:
```suggestion
  execSync(`npm run generate:deps -- --compare=true`);
```
(some history here: https://github.com/npm/npm/pull/5518)

package.json
```diff
@@ -42,6 +42,7 @@
     "config:update": "ts-node ./tools/config/index.ts --fix",
     "config:check": "ts-node ./tools/config/index.ts",
     "generate:deps": "ts-node ./tools/schematics/dependency-collector.ts",
+    "generate:deps:compare": "ts-node ./tools/schematics/dependency-collector.ts --compare",
```
pawelfras: As in the comment above, we can avoid adding this script.

pawelfras (approved):

PR #18429, opened 2024-01-29
restore server.ts

kpawelczak:
closes: [CXSPA-5855](https://jira.tools.sap/browse/CXSPA-5855)

- updateServerFiles(): This is a schematic rule that calls another rule called removeServer.

- removeServer(). It traverses the file tree and identifies two paths 'server.ts' and 'server.ts.bak'. If both paths are found, it deletes the original server file and renames a backup file to match the original name.


projects/schematics/src/migrations/mechanism/update-ssr/update-ssr-files.ts
```diff
@@ -4,8 +4,42 @@
...
 
 export function updateServerFiles(): Rule {
-  return chain([]);
+  return chain([removeServer]);
+}
+
+function removeServer(): Rule {
+  return (tree: Tree, _context: SchematicContext) => {
+    let serverPath;
+    let serverBakPath;
```
pawelfras: let's add typing as in the case of assigning the value inside the `visit` loop the type won't be inferred:
image

```suggestion
    let serverPath: Path | undefined;
    let serverBakPath: Path | undefined;
```

projects/schematics/src/migrations/mechanism/update-ssr/update-ssr-files.ts
```diff
@@ -4,8 +4,42 @@
...
+    tree.visit((filePath: Path) => {
+      const fileName = filePath.replace(/^.*[\\/]/, '');
+      if (fileName === SERVER_FILENAME) {
+        serverPath = filePath;
+      }
+
+      if (fileName === SERVER_BAK_FILENAME) {
+        serverBakPath = filePath;
+      }
+    });
```
pawelfras, line 40: `nitpick:` Would be great to break the loop as soon as we find both paths. If there is no easy way to do it, let's leave it as it is.

projects/schematics/src/migrations/mechanism/update-ssr/update-ssr-files_spec.ts
```diff
@@ -77,4 +78,20 @@ describe('Update SSR', () => {
       tree
     );
   });
+
+  describe('updateServerFile', () => {
+    it('should rename server.bak to server.ts', async () => {
```
pawelfras:
```suggestion
    it('should restore server.ts based on the server.ts.bak and remove server.ts.bak file', async () => {
```

projects/schematics/src/migrations/mechanism/update-ssr/update-ssr-files_spec.ts
```diff
@@ -77,4 +78,20 @@ describe('Update SSR', () => {
...
+      tree.create(serverBakPath, 'testing');
+      expect(tree.exists(serverBakPath)).toBeTruthy();
+      tree = await updateSsrSchematicRunner.runSchematic(
+        'update-ssr',
+        { name: 'schematics-test' },
+        tree
+      );
+
+      expect(tree.exists(serverBakPath)).toBeFalsy();
+      expect(tree.exists(SERVER_FILENAME)).toBeTruthy();
```
pawelfras, line 103: would be nice to test whether the `server.ts` file is different after restoration.

projects/schematics/src/migrations/mechanism/update-ssr/update-ssr-files_spec.ts
kpawelczak: replaced

projects/schematics/src/migrations/mechanism/update-ssr/update-ssr-files.ts
kpawelczak: I've already thought about it and couldn't find an easy solution.

projects/schematics/src/migrations/mechanism/update-ssr/update-ssr-files.ts
kpawelczak: I've added types

projects/schematics/src/migrations/mechanism/update-ssr/update-ssr-files_spec.ts
kpawelczak: I've added one more expect. It checks whether the restored server.ts has the same content as server.ts.bak.

pawelfras (approved):
LGTM

PR #18431, opened 2024-01-29
Change import paths in server.ts

kpawelczak:
closes: [CXSPA-5856](https://jira.tools.sap/browse/CXSPA-5856)

- updateServerFiles(): This is a schematic rule that calls another rule named modifyServerImports.

- modifyServerImports(): It reads the content of a server file ('server.ts') from the file tree. Checks if the content includes references to two specific imports ('zone.js/dist/zone-node' and '@nguniversal/express-engine').
If either of these imports is found, it updates the content by replacing the old import with a new one. The replacements are specified by constants ('zone.js/node' and '@spartacus/setup/ssr').
Finally, it overwrites the original server file with the updated content.

### DIFF
--- a/projects/schematics/src/migrations/mechanism/update-ssr/update-ssr-files.ts
+++ b/projects/schematics/src/migrations/mechanism/update-ssr/update-ssr-files.ts
@@ -4,8 +4,39 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { chain, Rule } from '@angular-devkit/schematics';
+import {
+  chain,
+  Rule,
+  SchematicContext,
+  Tree,
+} from '@angular-devkit/schematics';
+import {
+  NEW_ZONE_IMPORT,
+  NGUNIVERSAL_IMPORT,
+  OLD_ZONE_IMPORT,
+  SERVER_FILENAME,
+  SSR_SETUP_IMPORT,
+} from '../../../shared/constants';
 
 export function updateServerFiles(): Rule {
-  return chain([]);
+  return chain([modifyServerImports]);
+}
+
+function modifyServerImports(): Rule {
+  return (tree: Tree, _context: SchematicContext) => {
+    let updateContent = tree.read(SERVER_FILENAME)?.toString('utf-8') ?? '';
+    if (
+      updateContent.includes(OLD_ZONE_IMPORT) ||
+      updateContent.includes(NGUNIVERSAL_IMPORT)
+    ) {
+      updateContent = updateContent.includes(OLD_ZONE_IMPORT)
+        ? updateContent.replace(OLD_ZONE_IMPORT, NEW_ZONE_IMPORT)
+        : updateContent;
+      updateContent = updateContent.includes(NGUNIVERSAL_IMPORT)
+        ? updateContent.replace(NGUNIVERSAL_IMPORT, SSR_SETUP_IMPORT)
+        : updateContent;
+      tree.overwrite(SERVER_FILENAME, updateContent);
+    }
+    return tree;
+  };
 }
--- a/projects/schematics/src/migrations/mechanism/update-ssr/update-ssr-files.ts
+++ b/projects/schematics/src/migrations/mechanism/update-ssr/update-ssr-files.ts
@@ -24,17 +24,20 @@ export function updateServerFiles(): Rule {
 
 function modifyServerImports(): Rule {
   return (tree: Tree, _context: SchematicContext) => {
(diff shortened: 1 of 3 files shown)


projects/schematics/src/migrations/mechanism/update-ssr/update-ssr-files.ts
```diff
@@ -4,8 +4,39 @@
...
+      updateContent = updateContent.includes(OLD_ZONE_IMPORT)
+        ? updateContent.replace(OLD_ZONE_IMPORT, NEW_ZONE_IMPORT)
+        : updateContent;
+      updateContent = updateContent.includes(NGUNIVERSAL_IMPORT)
+        ? updateContent.replace(NGUNIVERSAL_IMPORT, SSR_SETUP_IMPORT)
+        : updateContent;
+      tree.overwrite(SERVER_FILENAME, updateContent);
+    }
+    return tree;
+  };
```
pawelfras:
```suggestion
return (tree: Tree, _context: SchematicContext) => {
    let serverFileBuffer = tree.read(SERVER_FILENAME);

    if (!serverFileBuffer) {
      return tree;
    }

    let updateContent = serverFileBuffer.toString('utf-8');
    const hasOldImport =
      updateContent.includes(OLD_ZONE_IMPORT) ||
      updateContent.includes(NGUNIVERSAL_IMPORT);
    if (hasOldImport) {
      updateContent = updateContent.replace(OLD_ZONE_IMPORT, NEW_ZONE_IMPORT);
      updateContent = updateContent.replace(
        NGUNIVERSAL_IMPORT,
        SSR_SETUP_IMPORT
      );

      tree.overwrite(SERVER_FILENAME, updateContent);
    }
    return tree;
  };
```
In this version `includes` is only called once for each import, and replace is called regardless of whether the import is present. There is a small risk that some of these imports will not be on file, but even if they are, the `replace` method will return the original value if there is no search string in the `updateContent`

projects/schematics/src/migrations/mechanism/update-ssr/update-ssr-files_spec.ts
```diff
@@ -77,4 +78,34 @@ describe('Update SSR', () => {
...
+  describe('updateServerFile', () => {
+    it('should change nguniversal import in server.ts', async () => {
+      tree = await updateSsrSchematicRunner.runSchematic(
+        'update-ssr',
+        { name: 'schematics-test' },
+        tree
+      );
+
+      const updatedContent = tree.read('./server.ts')!.toString();
+      expect(updatedContent).toContain('@spartacus/setup/ssr');
```
pawelfras: Let's use the `SSR_SETUP_IMPORT` const once it's merged to the epic branch.

pawelfras: Maybe it's also worth checking whether the old import path has been removed.

projects/schematics/src/migrations/mechanism/update-ssr/update-ssr-files_spec.ts
```diff
@@ -77,4 +78,34 @@ describe('Update SSR', () => {
...
+        tree.overwrite('./server.ts', serverContent);
+      }
+      tree = await updateSsrSchematicRunner.runSchematic(
+        'update-ssr',
+        { name: 'schematics-test' },
+        tree
+      );
+
+      const updatedServerContent = tree.read('./server.ts')!.toString();
+      expect(updatedServerContent).toContain(NEW_ZONE_IMPORT);
```
pawelfras, line 117: Maybe it's also worth checking whether the old import path has been removed.

projects/schematics/src/migrations/mechanism/update-ssr/update-ssr-files.ts
kpawelczak: sure, looks better

projects/schematics/src/migrations/mechanism/update-ssr/update-ssr-files_spec.ts
kpawelczak: fixed

kpawelczak: I've added extra check for the old import path

projects/schematics/src/migrations/mechanism/update-ssr/update-ssr-files_spec.ts
kpawelczak: I've added an extra check for the old import path

pawelfras (approved):
LGTM

PR #18432, opened 2024-01-29
Remove local REQUEST/RESPONSE definitions

kpawelczak:
closes: [CXSPA-5857](https://jira.tools.sap/browse/CXSPA-5857)

The main function, **updateServerFiles()**, is a schematic rule that executes two sub-rules: removeExpressTokensFile() and updateTokenImportPaths().

- removeExpressTokensFile(): This rule goes through the file tree and looks for files ending with the express.tokens.ts, and removes them from the tree.

- getExpressImportPaths(source: ts.SourceFile): string[]: This utility function extracts import paths from a TypeScript source file that include the **express.tokens**. It returns an array of these import paths.

- updateTokenImportPaths(): This rule, similar to the first one, traverses the file tree, reads TypeScript files, and updates import paths. It uses the getExpressImportPaths utility to identify import paths related to **express.tokens**. It then replaces these paths with a different one "@spartacus/setup/ssr" in the TypeScript file content and overwrites the file in the tree.

### DIFF
--- a/projects/schematics/src/migrations/mechanism/update-ssr/update-ssr-files.ts
+++ b/projects/schematics/src/migrations/mechanism/update-ssr/update-ssr-files.ts
@@ -3,9 +3,77 @@
  *
  * SPDX-License-Identifier: Apache-2.0
  */
-
-import { chain, Rule } from '@angular-devkit/schematics';
+import {
+  chain,
+  Rule,
+  SchematicContext,
+  Tree,
+} from '@angular-devkit/schematics';
+import { Path } from '@angular-devkit/core';
+import * as ts from 'typescript';
+import { EXPRESS_TOKENS, SSR_SETUP_IMPORT } from '../../../shared/constants';
 
 export function updateServerFiles(): Rule {
-  return chain([]);
+  return chain([removeExpressTokensFile, updateTokenImportPaths]);
+}
+
+function removeExpressTokensFile(): Rule {
+  return (tree: Tree, _context: SchematicContext) => {
+    tree.visit((filePath: Path) => {
+      if (filePath.endsWith(`${EXPRESS_TOKENS}.ts`)) {
+        tree.delete(filePath);
+      }
+    });
+    return tree;
+  };
+}
+
+function getExpressImportPaths(source: ts.SourceFile): string[] {
+  const importPaths: string[] = [];
+  source.statements.map((node) => {
+    if (
+      ts.isImportDeclaration(node) &&
+      ts.isStringLiteral(node.moduleSpecifier)
+    ) {
+      const importPath = node.moduleSpecifier.text;
+      if (importPath.includes(EXPRESS_TOKENS)) {
+        importPaths.push(importPath);
+      }
+    }
+    return node;
+  });
+
+  return importPaths;
+}
+
+function updateTokenImportPaths(): Rule {
+  return (tree: Tree, _context: SchematicContext) => {
+    tree.visit((filePath) => {
+      if (filePath.endsWith('.ts')) {
+        const fileContent = tree.read(filePath);
+        if (fileContent) {
+          const source = ts.createSourceFile(
+            filePath,
+            fileContent.toString(),
+            ts.ScriptTarget.Latest,
+            true
+          );
+          const importPaths = getExpressImportPaths(source);
+          let updatedFileContent = fileContent.toString('utf-8');
+
+          if (importPaths.length) {
+            importPaths.forEach((path: string) => {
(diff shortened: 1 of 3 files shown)


projects/schematics/src/migrations/mechanism/update-ssr/update-ssr-files.ts
```diff
@@ -4,8 +4,77 @@
...
+        tree.delete(filePath);
+      }
+    });
+    return tree;
+  };
+}
+
+function getExpressImportPaths(source: ts.SourceFile): string[] {
+  const importPaths: string[] = [];
+  source.statements.map((node) => {
```
pawelfras: Do we need to use `map` here? Is mapping values inside source statements required or do you need to iterate over them?

projects/schematics/src/migrations/mechanism/update-ssr/update-ssr-files.ts
pawelfras, line 1: All these steps seem to be very specific for the 2211.19.0 release and I'm wondering whether they should be part of `migrations/mechanism` directory. Let's keep them here for now and we can come back to the topic during the review of the PR with the epic `feature/CXSPA-5613` branch

projects/schematics/src/migrations/mechanism/update-ssr/update-ssr-files.ts
```diff
@@ -4,8 +4,77 @@
...
+
+  return importPaths;
+}
+
+function updateTokenImportPaths(): Rule {
+  return (tree: Tree, _context: SchematicContext) => {
+    tree.visit((filePath) => {
+      if (filePath.endsWith('.ts')) {
+        const fileContent = tree.read(filePath);
+        if (fileContent) {
```
pawelfras: Consider changing the name to `fileContentBuffer` to be more specific.

projects/schematics/src/migrations/mechanism/update-ssr/update-ssr-files.ts
```diff
@@ -4,8 +4,77 @@
...
+          const importPaths = getExpressImportPaths(source);
+          let updatedFileContent = fileContent.toString('utf-8');
+
+          if (importPaths.length) {
+            importPaths.forEach((path: string) => {
+              updatedFileContent = updatedFileContent.replace(
+                path,
+                SSR_SETUP_IMPORT
+              );
+            });
```
pawelfras: Consider using RegExp instead of iterating through all import paths. Thanks to that instead of running ' Replace` each time, you'll call it only once.
```suggestion
            const importPathsRegex = new RegExp(importPaths.join('|'), 'g');
            const updatedFileContent = fileContent.replace(
              importPathsRegex,
              SSR_SETUP_IMPORT
            );
```

projects/schematics/src/migrations/mechanism/update-ssr/update-ssr-files.ts
```diff
@@ -4,8 +4,77 @@
...
+        const fileContent = tree.read(filePath);
+        if (fileContent) {
+          const source = ts.createSourceFile(
+            filePath,
+            fileContent.toString(),
+            ts.ScriptTarget.Latest,
+            true
+          );
+          const importPaths = getExpressImportPaths(source);
+          let updatedFileContent = fileContent.toString('utf-8');
```
pawelfras: Can we do this conversion earlier and use it also in line 58? We will then avoid doing it in two places.

projects/schematics/src/migrations/mechanism/update-ssr/update-ssr-files.ts
kpawelczak: I've changed it to foreach

projects/schematics/src/migrations/mechanism/update-ssr/update-ssr-files.ts
kpawelczak: I've changed the name to fileContentBuffer

projects/schematics/src/migrations/mechanism/update-ssr/update-ssr-files.ts
kpawelczak: I've replaced the foreach.

projects/schematics/src/migrations/mechanism/update-ssr/update-ssr-files.ts
kpawelczak: fixed

pawelfras (approved):

PR #18441, opened 2024-02-01
fix: (CXSPA-924) - Language dropdown aria-label

Pio-Bar:
https://jira.tools.sap/browse/CXSPA-924

Provides an aria-label for dropdown options. Otherwise JAWS was unable to read Chinese and Japanese characters.


Zeyber (dismissed):

developpeurweb (dismissed):
Tested with JAWS, works fine now.

Zeyber (approved):

PR #18444, opened 2024-02-02
fix: (CXSPA-955) - Restore focus after delivery modes are refetched

Pio-Bar:
https://jira.tools.sap/browse/CXSPA-955

This PR only fixes the case#2 from the ticket as i found the case#1 does not occur anymore.
Focus should be restored after selecting an option with the keyboard.


feature-libs/checkout/base/components/checkout-delivery-mode/checkout-delivery-mode.component.ts
```diff
@@ -100,6 +101,33 @@ export class CheckoutDeliveryModeComponent {
...
+    const lastFocusedId = (<HTMLElement>event.target)?.id;
+    this.changeMode(code);
+
+    this.isUpdating$
+      .pipe(
+        filter((isUpdating) => !isUpdating),
+        take(1)
+      )
+      .subscribe(() => {
+        setTimeout(() => {
```
Zeyber: Is there some way we can achieve the goal without `setTimeout`? It is known to cause issues with SSR and testing in cypress sometimes (due to delays between actions). Not critical since this is during a key event but it would be preferable.

https://sap.github.io/spartacus-docs/3.x/server-side-rendering-coding-guidelines/#working-with-timeouts

feature-libs/checkout/base/components/checkout-delivery-mode/checkout-delivery-mode.component.html
```diff
@@ -13,16 +13,34 @@ <h2 class="cx-checkout-title d-none d-lg-block d-xl-block">
...
-          class="form-check-input"
-          role="radio"
-          type="radio"
-          id="deliveryMode-{{ mode.code }}"
-          (change)="changeMode(mode.code)"
-          [value]="mode.code"
-          formControlName="deliveryModeId"
-        />
+        <!-- TODO: (CXSPA-6009) Remove feature flags next major release -->
+        <ng-container *cxFeatureLevel="'6.8'">
```
Zeyber: The only change I see here is adding the key event. I wouldn't consider this as "breaking" so I don't think that we need the feature flags.

feature-libs/checkout/base/components/checkout-delivery-mode/checkout-delivery-mode.component.ts
```diff
@@ -100,6 +101,33 @@ export class CheckoutDeliveryModeComponent {
     protected activeCartFacade: ActiveCartFacade
   ) {}
 
+  onKeydown(code: string | undefined, event: Event): void {
```
Zeyber: I think that we should unit and e2e test this. E2E just to cover this is working in the user's environment.

Zeyber: Could we please have some comments or methods to outline why we are capturing the keydown event?

Zeyber (changes requested):

feature-libs/checkout/base/components/checkout-delivery-mode/checkout-delivery-mode.component.ts
Pio-Bar: I'm not happy with this either, unfortunately I could not get it to work without using `setTimeout`. I've tried a couple of things but could not get it to work. :(

feature-libs/checkout/base/components/checkout-delivery-mode/checkout-delivery-mode.component.html
Pio-Bar: I got rid of feature flags. ;)

feature-libs/checkout/base/components/checkout-delivery-mode/checkout-delivery-mode.component.ts
Pio-Bar: I've provided a unit and e2e test.

Pio-Bar: Comment added for clarity.

projects/storefrontapp-e2e-cypress/cypress/e2e/accessibility/focus-management.e2e.cy.ts
```diff
@@ -0,0 +1,29 @@
...
+ * SPDX-FileCopyrightText: 2024 SAP Spartacus team <spartacus-team@sap.com>
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import * as checkout from '../../helpers/checkout-flow';
+
+describe('Focus managment for a11y', () => {
+  context('Checkout - Delivery modes', () => {
+    before(() => {});
```
Zeyber: Tiny thing, I don't think we need this before hook.

Zeyber (changes requested):

projects/storefrontapp-e2e-cypress/cypress/e2e/accessibility/focus-management.e2e.cy.ts
Pio-Bar: removed

Zeyber (dismissed):

feature-libs/checkout/base/components/checkout-delivery-mode/checkout-delivery-mode.component.ts
```diff
@@ -100,6 +101,32 @@ export class CheckoutDeliveryModeComponent {
     protected activeCartFacade: ActiveCartFacade
   ) {}
 
+  /**
+   * We capture the kayboard activated option,
```
rmch91: typo in kayboard -> keyboard

rmch91 (dismissed):

developpeurweb (dismissed):

feature-libs/checkout/base/components/checkout-delivery-mode/checkout-delivery-mode.component.ts
```diff
@@ -100,6 +101,32 @@ export class CheckoutDeliveryModeComponent {
     protected activeCartFacade: ActiveCartFacade
   ) {}
 
+  /**
+   * We capture the keyboard activated option,
+   * then restore the focus to it after the data is updated.
+   */
+  updateMethod(code: string | undefined, event: Event): void {
```
Zeyber: Is it possible to move this logic into the `changeMode` method? We can add a protected method or something to handle the refocusing. Otherwise it looks like we are in-reality deprecating the `changeMode` method which doesn't really make sense (`updateMethod` still does all `changeMode` does but "fixed"). 

In doing this, we should include the `$event` parameter in the `changeMode` method as an optional parameter and we can extend its logic if it is passed.

Zeyber (changes requested):

feature-libs/checkout/base/components/checkout-delivery-mode/checkout-delivery-mode.component.ts
Pio-Bar: I've refactored it, should be easier to read now.

feature-libs/checkout/base/components/checkout-delivery-mode/checkout-delivery-mode.component.ts
```diff
@@ -100,17 +101,23 @@ export class CheckoutDeliveryModeComponent {
...
       complete: () => this.onSuccess(),
       error: () => this.onError(),
     });
+
+    const isTriggeredByKeyboard = (<MouseEvent>event)?.screenX === 0;
+    if (isTriggeredByKeyboard) {
+      this.restoreFocus(lastFocusedId, code);
+      return;
+    }
+    this.mode.setValue({ deliveryModeId: code });
```
Zeyber: Should we hide this section behind a feature flag?

Zeyber (changes requested):

feature-libs/checkout/base/components/checkout-delivery-mode/checkout-delivery-mode.component.ts
Pio-Bar: Feature flag added ;)

Zeyber (approved):

PR #18447, opened 2024-02-05
chore: create json files for migrations

kpawelczak:
relates to [CXSPA-5990](https://jira.tools.sap/browse/CXSPA-5990)


docs/migration/2211_0/2211-html.md
Platonn, line 1: I can see we're changing folder from 2211 to 2211_0 (which symbolizes `2211.0`).
But our major release is going to be named `2211.19`, therefore IMHO the folder should be named `2211_19`.

I'm afraid our _tool for discovering breaking changes_ might have hardcoded some `_0` (zeros), so our script might need adjusting:
image


Why it's important to no longer stick to `_0`: because we're droping semver, and our future major releases might not have `0` in the second part of the version. For instance in 12 months we might be releasing a "major" release in February 2025 which will have version `2211.31`.

Note: Fixing _our tool_ to support other folder names than `_0` is not urgent. So I suggest to defer it to later, just please create a ticket for it. Hopefully, we can do it even after our nearest February major release is done or when/if we have free time for it.

kpawelczak: Yes, I'm aware of this, hence the change. At the moment I would like to have the script working.  I'll create a ticket.

Platonn (approved):
Approved with comments:
- let's add a prefix for the PR's title (and the merging commit message):  `chore:` (for more see [docs](https://sap.github.io/spartacus-docs/commit-guidelines/#commit-message-format))
- link to another [(comment) ](https://github.com/SAP/spartacus/pull/18447/files#r1478621698)

PR #18448, opened 2024-02-05
fix: (CXSPA-954) - Restored default tab flow for quantity counter

Pio-Bar:
https://jira.tools.sap/browse/CXSPA-954

As discussed on our daily, I've restored the default tab flow for the quantity counter on PDP.


projects/storefrontlib/shared/components/item-counter/item-counter.component.ts
```diff
@@ -78,8 +80,15 @@ export class ItemCounterComponent implements OnInit, OnDestroy {
...
   private sub: Subscription;
 
+  constructor(
+    @Optional() protected featureConfigService: FeatureConfigService
+  ) {}
+
+  // TODO: (CXSPA-6034) Remove HostListener and @ViewChild('qty') next major release
   @HostListener('click') handleClick() {
-    this.input.nativeElement.focus();
+    if (this.featureConfigService.isLevel('6.8')) {
```
Zeyber: Should this be flipped to `!6.8`?

projects/storefrontlib/shared/components/item-counter/item-counter.component.ts
```diff
@@ -78,8 +80,15 @@ export class ItemCounterComponent implements OnInit, OnDestroy {
    */
   private sub: Subscription;
 
+  constructor(
```
Zeyber: Although there is still a lot of legacy code left behind with constructor injection, in general, we have stopped using the constructor to inject dependencies. This is because it is easier to run our migration scripts. Instead, we use injector syntax.

```
protected featureConfigService = inject(FeatureConfigService, {                                                                    
    optional: true,
});
```

**Why is this easier for our migration scripts?**

Because our migration scripts read try to read changes in the constructor where order is important. This is difficult to detect changes and preserve when a client may add/remove/switch any dependency during the upgrade process.

ie. `constructor(firstService, secondService)` IS NOT THE SAME AS `constructor(secondService, firstService)`

With the injection method, we no longer need to worry about order which significantly reduces the complexity of doing our migrations.

ie.

```
firstService = inject(FirstService)
secondService = inject(SecondService)
```

IS THE SAME AS

```
secondService = inject(SecondService)
firstService = inject(FirstService)
```

projects/storefrontlib/shared/components/item-counter/item-counter.component.html
```diff
@@ -7,7 +7,9 @@
 >
   -
 </button>
+<!-- TODO: (CXSPA-6034) Remove feature flag next major release -->
 <input
+  *cxFeatureLevel="'!6.8'"
```
Zeyber: I don't think we need to feature flag this to remove the #qty binding since it is angular-specific and functionality is handled in the class. But lets still leave a comment to remove the binding later.

Zeyber (changes requested):

projects/storefrontlib/shared/components/item-counter/item-counter.component.ts
Pio-Bar: It should, good catch!

projects/storefrontlib/shared/components/item-counter/item-counter.component.ts
Pio-Bar: Switched to constructor syntax. Thanks for the explanation, I appreciate it

projects/storefrontlib/shared/components/item-counter/item-counter.component.html
Pio-Bar: Removed feature flag and left a comment

projects/storefrontlib/shared/components/item-counter/item-counter.component.spec.ts
```diff
@@ -22,7 +23,7 @@ describe('ItemCounterComponent', () => {
     waitForAsync(() => {
       TestBed.configureTestingModule({
         imports: [RouterTestingModule, ReactiveFormsModule, I18nTestingModule],
-        declarations: [ItemCounterComponent],
+        declarations: [ItemCounterComponent, MockFeatureLevelDirective],
```
Zeyber: We don't need this directive anymore ;)

Zeyber (changes requested):

projects/storefrontlib/shared/components/item-counter/item-counter.component.spec.ts
Pio-Bar: removed

Zeyber (dismissed):

developpeurweb (dismissed):

Zeyber (approved):

PR #18454, opened 2024-02-06
feat: new `FeatureFlags` config separate from Spartacus config; new feature flag allowing to toggle "My Account V2" in runtime

Platonn:
We'd like to allow for runtime configurability of all feature flags in Spartacus. The existing `FeaturesConfig` object which is part of the global Spartacus config (`provideConfig({ features: {...} })`) can be easily mapped in runtime from some dynamically loaded external JSON file. But the separate `InjectionToken`s cannot be easily mapped.

Unfortunately the feature of "My Account v2" enable parts of its funcitonality via separate `InjectionToken`s, which cannot be easily mapped from JSON in runtime:
- USE_MY_ACCOUNT_V2_EMAIL
- USE_MY_ACCOUNT_V2_PASSWORD
- USE_MY_ACCOUNT_V2_PROFILE
- USE_MY_ACCOUNT_V2_NOTIFICATION_PREFERENCE
- USE_MY_ACCOUNT_V2_CONSENT

To mitigate this issue, in this PR we introduce a regular feature flag in the `FeaturesConfig` called `config.features.myAccountV2`. We also introduce a mapping from this new feature flag that controls default values of all those `InjectionToken`s, which results in allowing to easily enable them all via an external JSON config file.

Unfortunately, some parts of the Spartacus global config already depend on those `InjectionToken`s, like cms mappings, see example:
https://github.com/SAP/spartacus/blob/eb09a3ed15b5f3b0f5c2cbe36011356ec5b1d5e3/feature-libs/order/components/order-details/order-details.module.ts#L161-L163

Therefore to avoid a circular dependency: 
`Config` (`FeaturesConfig`) -> `USE_MY_ACCOUNT_V2_ORDER` -> `Config` (`CmsConfig`)
... the value of the flag has to be provided not via `provideConfig()` function, but we had to introduce a new provider util function.

So in this PR, we're introducing a higher order mechanism for providing feature flags, called `FeatureFlags` and `provideFeatureFlags` (similarly to `Config` and `provideConfig`). Since now, we'll recommend to customers to enable Spartacus feature flags NOT via:
```ts
provideConfig({ 
  features: { <flags here> }
})
```
but via
```ts
provideFeatureFlags({ <flags here> })
```

Behind the scenes we'll copy whole new object of `FeatureFlags` into the exisitng `FeaturesConfig.features` object for backward compatibility.

**Pros:**
- The feature "My Account v2" can be easily toggled in runtime via an external JSON file, similarly to all other feature flags

**Cons:**
- We introduce a new convention to customers - they have to provide feature flags via a new util function `provideFeatureFlags()`, but not the well known `provideConfig()`, which might be a source of confusion for some customers.

Note: Customers can still use `provideConfig()` for providing feature flags. Only the new flag `myAccountV2` has to be provided via the new function `provideFeatureFlags()`. It's a new flag and a new function, so this PR doesn't introduce any breaking change. That said, for consistency reasons, we'll recommend in our official docs to use only the new function `provideFeatureFlags()` for providing feature flags.

PS. In this PR we also removed an unused interface `FeatureToggles` (it was _truly_ unused and not exported in the public API). Instead of it, we declared all its properties into the `FeatureFlags` interface, via the module augmentation (spread across various libraries).

---
TODO:
- [ ] test manually feature flags
- [ ] test manually providing feature flags via a cookie cxConfig in e2e
- [ ] add unit tests
- [ ] run e2e
- [ ] add QA steps


Platonn:
closing in favor of https://github.com/SAP/spartacus/pull/18670

PR #18456, opened 2024-02-07
feature/CXINT-2714: Show promotions on PDP page

anjana-bl:
user story: https://jira.tools.sap/browse/CXINT-2743


projects/storefrontlib/cms-components/product/product-summary/product-summary.component.ts
```diff
@@ -16,12 +16,19 @@ import { ProductDetailOutlets } from '../product-outlets.model';
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
 export class ProductSummaryComponent {
+  protected featureConfig = inject(FeatureConfigService);
```
Platonn: Not to pollute the Public API of the component with temporary flags (that we'll need to remove in the future - which will be unnecessary breaking change), could you please make this property `private`?

anjana-bl: yes done

Platonn (approved):

PR #18464, opened 2024-02-08
fix: (CXSPA-1794) - List heading order

Pio-Bar:
https://jira.tools.sap/browse/CXSPA-1794

Corrected heading order.


Zeyber (approved):
This is missing a cleanup ticket. BUT I think we may change the way we are catching changes during cleanup so we can address this once the version level has been bumped if needed.

developpeurweb (dismissed):

feature-libs/organization/administration/styles/_list.scss
```diff
@@ -73,15 +73,36 @@
...
-%organizationTitle {
-  .title {
-    h3 {
-      button {
-        border: none;
-        background-color: inherit;
-        cx-icon {
-          pointer-events: none;
-          color: var(--cx-color, var(--cx-color-info));
+@include forVersion(6.8) {
```
Zeyber: The only differences here that I see are that we add the same style to `h2` and include the type. Because I don't see any property changes or removals, I think we can add something like this to the bottom of the class.

```
@include featureStyles('flag') {
    %organizationTitle {
        .title {
            h2, h3 {
                @include type('3', 'normal');
            }
        }
    }
}
```

Zeyber (changes requested):

Zeyber (dismissed):

Zeyber (approved):

PR #18470, opened 2024-02-09
feature/CXINT-2740: Disable editing when service requests are from CDP

anjana-bl:
https://jira.tools.sap/browse/CXINT-2740

When CDP library is installed:
> Disable 'add new message' section and 'close request' button for Open/InProcess requests in details screen
> Disable 'reopen request' button for Closed requests in details screen


projects/storefrontapp/src/app/spartacus/spartacus-features.module.ts
```diff
@@ -114,6 +114,9 @@ if (environment.b2b) {
 if (environment.cdc) {
   featureModules.push(CdcFeatureModule);
 }
+// if (environment.cdp) {
```
FollowTheFlo: is this commented out code portion expected?

projects/schematics/src/shared/schematics-config-mappings.ts
```diff
@@ -128,6 +129,10 @@ export const SCHEMATICS_CONFIGS: SchematicConfig[] = [
   CDC_SCHEMATICS_CONFIG,
   CDC_B2B_SCHEMATICS_CONFIG,
 
+  CDC_B2B_SCHEMATICS_CONFIG,
```
FollowTheFlo: There are 2 occurrences of CDC_B2B_SCHEMATICS_CONFIG, I think we can remove one.

FollowTheFlo:
@anjana-bl Could you also add:
`"dist/cdp:integration-libs/cdp"`
as part of SPARTACUS_PROJECTS for the install script.
in file:
spartacus/scripts/install/config.default.sh
add it between lines 37 and 38, that will be fine, I tested it already.
It will avoid installation to break when using install script, which is essentially use internally.

projects/storefrontapp/src/app/spartacus/spartacus-features.module.ts
anjana-bl: it is not needed as of now, as CdpFeatureModule doesnot have any modules to import.
FYI- there is no CdpFeatureModule or CdpModule created for this library, as they dont have any usage as of now.

projects/schematics/src/shared/schematics-config-mappings.ts
anjana-bl: removed.

FollowTheFlo (approved):
As discussed, I am good with it

PR #18488, opened 2024-02-13
Bug/CXINT-2438: Update Sort Keys in Spartacus PDF Invoices Component 

anjana-bl:
https://jira.tools.sap/browse/CXINT-2438


feature-libs/pdf-invoices/components/invoices-list/invoices-list.component.ts
```diff
@@ -51,8 +51,8 @@ export class InvoicesListComponent implements OnInit, OnDestroy {
   sort = 'byInvoiceIdAsc';
 
   protected sortMapping: { [key: string]: string } = {
-    byCreatedAtAsc: 'invoiceDate:asc', //TODO: (CXINT-2438) update the sort code after the API is changed
-    byCreatedAtDesc: 'invoiceDate:desc', //TODO: (CXINT-2438) update the sort code after the API is changed
+    byCreatedAtAsc: 'createdAt:asc',
+    byCreatedAtDesc: 'createdAt:desc',
```
Platonn: Could you help me understand the backward compatibility consequences of this change?

I can understand that users of the newest backend API and newest Spartacus will get desired behavior.
And users of the old backend API and old Spartacus will also get desired correct behavior.

But does this change mean that some customers that use the old version of the backend API, but the newest Spartacus will encounter problems?

Analogical question: for users of new API but old Spartacus - will they encounter problems?

Should we wrap this change with an opt-in feature flag?

anjana-bl: yes it should be wrapped in opt-in feature flag, i have made the necessary changes

feature-libs/pdf-invoices/components/pdf-invoices-components.module.ts
```diff
@@ -45,6 +45,9 @@ import { InvoicesListComponent } from './invoices-list/invoices-list.component';
           guards: [AuthGuard],
         },
       },
+      features: {
+        useUpdatedSortMappingForInvoices: false,
```
Platonn: I recommend starting with the name of the library, e.g. `pdfInvoices` and if possible describe semantically the change: e.g. `sortByInvoiceDate`.
What do you think about: `pdfInvoicesSortByInvoiceDate`?

anjana-bl: yes, it is better. have changed the name

Platonn (approved):

PR #18493, opened 2024-02-13
fix: Fix unit addresses and users in b2b my company not traversing or notifying on creation

### LINKED ISSUE #18232: [B2B Organization Unit / Delivery addresses & User] create form - incorrect behavior
**Describe the bug**

After submitting the form, it's cleared but still displayed

- Library version: 5.x & 6.x

**To Reproduce**
1. Log in as b2b admin (eg linda.wolf)
1a. Go to http://localhost:4200/powertools-spa/en/USD/organization/units/Rustic%20Retail/addresses/create
1b. Go to http://localhost:4200/powertools-spa/en/USD/organization/units/Rustic%20Retail/users/create
2. Fill out the form
3. Submit
4. See the newly created user/address in the list, but missing redirection and confirmation. The form is still frozen

**Expected behavior**
a) After creating the delivery address, the expected behavior is that the form will be closed and the list (or even better - address details view) will be shown. 
Message `orgUnitAddress.messages.create` should be emitted in the cx-org-message

b) After creating a user, the expected behavior is that the form will be closed and the list will be shown. 
I'm not sure about messages here, it was not defined, but is probably worth having here [improvement].

 - OS: Windows 11
 - Browser: Chrome

**Additional context**
It seems that the issue begins in v5 ([881566a8](https://github.com/SAP/spartacus/commit/881566a8edab8efe4e3bde1dde520cf6ac7e5ae2)) when optional/null has been replaced with undefined/empty string as id
It's worth testing negative scenarios for create/edit actions, where server validation is more strict than the frontend one. 
After deleting the address, it also doesn't trigger the defined confirmation (`orgUnitAddress.messages.deleted`)

Zeyber:
Closes: https://jira.tools.sap/browse/CXSPA-5630


feature-libs/organization/administration/core/store/effects/org-unit.effect.ts
```diff
@@ -384,11 +389,24 @@ export class OrgUnitEffects {
...
-              new OrgUnitActions.CreateAddressSuccess(data),
-              new OrgUnitActions.CreateAddressSuccess({ id: undefined }),
-              new OrganizationActions.OrganizationClearData(),
-            ]),
+            switchMap((data) => {
+              const successActions = [
+                new OrgUnitActions.CreateAddressSuccess(data),
+                new OrgUnitActions.CreateAddressSuccess({ id: undefined }),
+                new OrganizationActions.OrganizationClearData(),
+                new GlobalMessageActions.AddMessage({
```
Zeyber: It would be better to find a way to listen for the effect to complete and then have the message and routing actions performed by that listener. In this way, we can avoid added complexity here and a potential breaking change.

Zeyber:
Have simplified the approach based on the fix provided on [this PR](https://github.com/SAP/spartacus/pull/18360) and following discussion with @Pucek9 about his findings. It leverages the default my company mechanism for generating success messages and proceeding navigation to newly created entities which was broken for these cases in a refactor.

Pucek9 (changes requested):
image
Unfortunately, this change causes another problem

Zeyber:
Note for testers: To enable fixes, add the following provider to the b2b config module `projects/storefrontapp/src/app/spartacus/spartacus-b2b-configuration.module.ts`.

```
    provideConfig(<FeaturesConfig>{
      features: {
        fixMyCompanyUnitAddressCreation: true,
        fixMyCompanyUnitUserCreation: true,
      },
    }),
```

Pucek9 (approved):
Works as expected! Thank you

rmch91 (approved):

PR #18496, opened 2024-02-14
feat: change breaking changes script so it would work with new spa versions

kpawelczak:
closes: [CXSPA-6037](https://jira.tools.sap/browse/CXSPA-6037)


rmch91 (dismissed):

rmch91 (approved):

tools/breaking-changes/common.ts
```diff
@@ -9,13 +9,13 @@ import stringifyObject from 'stringify-object';
 
 // shared configs
 
-export const NEW_MAJOR_VERSION = '2211'; // Increment this number to match the current release for which the script is used.
-export const BREAKING_CHANGES_FILE_PATH = `data/${NEW_MAJOR_VERSION}_0/breaking-changes.json`;
-export const MAJOR_VERSION_DOC_HOME = `../../docs/migration/${NEW_MAJOR_VERSION}_0`;
+export const NEW_MAJOR_VERSION = '2211_19'; // Increment this number to match the current release for which the script is used.
```
Platonn, line 12: IMHO we should replace terms `MAJOR_VERSION` and `NEW_MAJOR_VERSION` to simply `NEW_VERSION`. 

Why: Since we no longer follow the [Semantic Versionsing ](https://semver.org/), IMHO we should no more refer to the word `Major` as it's confusing and suggests a _major_ release in terms of semver meaning (i.e. bumping the first part - x - of the `x.y.z` version). 

Since now, the breaking changes like removing a feature flag can happen potentially in _any release_, provided that a feature flag is at least 12 months old.

PR #18505, opened 2024-02-15
Make-to-stock product variants: Introduce new route for rulebased configurator read-only view

Larisa-Staroverova:


feature-libs/product-configurator/common/assets/translations/en/configurator.json
```diff
@@ -5,6 +5,7 @@
       "complete": "Complete",
       "configId": "Configuration ID",
       "toconfig": "Configure",
+      "toconfigReadOnly": "Show details",
```
ChristophHi:
```suggestion
      "toConfigReadOnly": "Show details",
```

feature-libs/product-configurator/common/components/configure-cart-entry/configure-cart-entry.component.ts
```diff
@@ -141,10 +144,15 @@ export class ConfigureCartEntryComponent {
   /**
    * Retrieves the state of the configuration.
    *
-   *  @returns - 'true' if the configuration is read only, otherwise 'false'
+   *  @returns - 'true' if the configuration is read only and configurator type contains a read-only postfix, otherwise 'false'
```
ChristophHi:
```suggestion
   *  @returns - 'true' if the configuration is read only or configurator type contains a read-only postfix, otherwise 'false'
```

feature-libs/product-configurator/common/components/configure-product/configure-product.component.spec.ts
```diff
@@ -194,6 +197,142 @@ describe('ConfigureProductComponent', () => {
...
+      );
+    });
+
+    it('should return configurator.header.toconfig in case configurator type is CPQCONFIGURATOR', () => {
+      expect(component.getTranslationKey(ConfiguratorType.VARIANT)).toEqual(
+        'configurator.header.toconfig'
+      );
+    });
+
+    it('should return configurator.header.toconfig in case configurator type has suffix readOnly', () => {
```
ChristophHi:
```suggestion
    it('should return configurator.header.toconfig in case configurator type has prefix readOnly', () => {
```

feature-libs/product-configurator/common/components/configure-product/configure-product.component.spec.ts
```diff
@@ -194,6 +197,142 @@ describe('ConfigureProductComponent', () => {
...
+    });
+
+    it('should return configurator.header.toconfigReadOnly in case configurator type has postfix readOnly', () => {
+      expect(
+        component.getTranslationKey(ConfiguratorType.VARIANT + ReadOnlyPostfix)
+      ).toEqual('configurator.header.toconfigReadOnly');
+    });
+  });
+
+  describe('isDisplayRestartDialog', () => {
```
ChristophHi, line 250: Not sure if we need all those tests, as the method internally calls isConfiguratorTypeReadOnly which is tested sufficiently?

feature-libs/product-configurator/common/components/configure-product/configure-product.component.spec.ts
```diff
@@ -194,6 +197,142 @@ describe('ConfigureProductComponent', () => {
...
+      expect(component['isConfiguratorTypeReadOnly']('   ')).toBe(false);
+    });
+
+    it('should return false in case configurator type is CPQCONFIGURATOR', () => {
+      expect(
+        component['isConfiguratorTypeReadOnly'](ConfiguratorType.VARIANT)
+      ).toBe(false);
+    });
+
+    it('should return false in case configurator type has suffix _READ_ONLY', () => {
```
ChristophHi:
```suggestion
    it('should return false in case configurator type has prefix _READ_ONLY', () => {
```

feature-libs/product-configurator/common/components/configure-product/configure-product.component.spec.ts
```diff
@@ -194,6 +197,142 @@ describe('ConfigureProductComponent', () => {
     });
   });
 
+  describe('getTranslationKey', () => {
```
ChristophHi: Not sure if we need _all_ those tests, as the method internally calls isConfiguratorTypeReadOnly which is tested sufficiently?

feature-libs/product-configurator/rulebased/components/add-to-cart-button/configurator-add-to-cart-button.component.spec.ts
```diff
@@ -691,13 +705,13 @@ describe('ConfigAddToCartButtonComponent', () => {
...
 
-  describe('displayOnlyButton', () => {
+  describe('leaveConfigurationOverview', () => {
     it('should navigate to review order', () => {
       setRouterTestDataReadOnlyCart();
       initialize();
       component.leaveConfigurationOverview();
       expect(routingService.go).toHaveBeenCalledWith({
-        cxRoute: 'checkoutReviewOrder',
+        cxRoute: 'cart',
```
ChristophHi: But that's the wrong behaviour. Instead of changing the test, please fix the issue: Navigation from checkout is wrong now, as leaving the OV hits the cart, not the checkout

projects/storefrontapp-e2e-cypress/cypress/helpers/product-configurator-cart.ts
```diff
@@ -24,19 +24,23 @@ export function clickOnEditConfigurationLink(cartItemIndex: number): void {
...
 
 /**
- * Clicks on the 'Display Configuration' link in cart/order or quote for a certain item.
+ * Clicks on the 'Display Configuration' link in cart for a certain cart item.
  *
  * @param {number} cartItemIndex - Index of cart item
  */
 export function clickOnDisplayConfigurationLink(cartItemIndex: number): void {
-  locateCartConfiguratorElement(cartItemIndex);
+  cy.get('cx-cart-item-list .cx-item-list-row')
```
ChristophHi: Just out of curiosity: Why was this change needed?

ChristophHi (changes requested):

feature-libs/product-configurator/common/components/configure-product/configure-product.component.spec.ts
Larisa-Staroverova: You are right. I remove most of them

feature-libs/product-configurator/common/components/configure-product/configure-product.component.spec.ts
Larisa-Staroverova: You are right. I remove most of them

projects/storefrontapp-e2e-cypress/cypress/helpers/product-configurator-cart.ts
Larisa-Staroverova: Good point! I rolled it back. The `locateCartConfiguratorElement(cartItemIndex);` method should be used

feature-libs/product-configurator/rulebased/root/default-rulebased-routing-config.ts
```diff
@@ -12,6 +12,16 @@ export const defaultRulebasedRoutingConfig: RoutingConfig = {
...
         paths: ['configure/vc/:ownerType/entityKey/:entityKey'],
       },
+      configureCPQCONFIGURATOR_READ_ONLY: {
+        paths: [
+          'configure-overview/vc/:ownerType/entityKey/:entityKey/displayOnly/true',
+        ],
+      },
+      configureOverviewCPQCONFIGURATOR_READ_ONLY: {
+        paths: [
+          'configure-overview/vc/:ownerType/entityKey/:entityKey/displayOnly/TRUE',
```
Larisa-Staroverova: Hi @Platonn,
there are 3 routes:
1. **configureCPQCONFIGURATOR_READ_ONLY**,
2. **configureOverviewCPQCONFIGURATOR_READ_ONLY** and 
3. **configureOverviewCPQCONFIGURATOR** that point to the same path
_configure-overview/vc/:ownerType/entityKey/:entityKey/displayOnly/:displayOnly_

We wonder whether there is a more elegant way to point to the same path instead of doing something like:
1. configure-overview/vc/:ownerType/entityKey/:entityKey/displayOnly/**:displayOnly**
2. configure-overview/vc/:ownerType/entityKey/:entityKey/displayOnly/**true** or
3. configure-overview/vc/:ownerType/entityKey/:entityKey/displayOnly/**TRUE**

Could you give us a hint how to declare several routes that point to the same path?
Thanks a lot in advance.

ChristophHi: Hi @Larisa-Staroverova when checking `routing-config.service` it seems that there is not easy way around that. Telling the routes from the path is still used in router.reducer.
So unless Kris has a good idea we could also handle this in configureCartEntry and configureProduct in a way that we generically switch to configureOverview<configuratorType> in case configuratorType has the READ_ONLY postfix.

feature-libs/product-configurator/rulebased/components/add-to-cart-button/configurator-add-to-cart-button.component.spec.ts
```diff
@@ -691,16 +706,7 @@ describe('ConfigAddToCartButtonComponent', () => {
     });
   });
 
-  describe('displayOnlyButton', () => {
```
ChristophHi, line 716: Why is this test deleted? As far as I see it it would reveal an issue because the navigation to overview from checkout and back is still not working

feature-libs/product-configurator/rulebased/components/add-to-cart-button/configurator-add-to-cart-button.component.ts
```diff
@@ -348,6 +348,18 @@ export class ConfiguratorAddToCartButtonComponent implements OnInit, OnDestroy {
         CommonConfigurator.OwnerType.SAVED_CART_ENTRY
       ) {
         this.goToSavedCartDetails(container.routerData.owner);
+      } else if (
+        container.routerData.owner.type ===
+        CommonConfigurator.OwnerType.CART_ENTRY
+      ) {
+        this.routingService.go({ cxRoute: 'cart' });
```
ChristophHi, line 406: that's wrong, we might have initiated the navigation from checkout. Please check the behaviour in develop

ChristophHi (changes requested):

feature-libs/product-configurator/rulebased/components/add-to-cart-button/configurator-add-to-cart-button.component.ts
Larisa-Staroverova: @ChristophHi,
if this condition is removed, then the navigation for dispalyOnly `cartEntry` does not work anymore.
Instead of navigating to the cart the end-user is navigated to the checkout review process.

ChristophHi: It's not about removing this line :-) but the process should work as before

feature-libs/product-configurator/rulebased/components/add-to-cart-button/configurator-add-to-cart-button.component.spec.ts
Larisa-Staroverova: The test has been moved to the `leaveConfigurationOverview` section

feature-libs/product-configurator/common/components/configure-cart-entry/configure-cart-entry.component.spec.ts
```diff
@@ -29,6 +35,20 @@ class MockUrlPipe implements PipeTransform {
   transform(): any {}
 }
 
+const mockRouterState: any = {
+  state: {
+    semanticRoute: 'home',
+  },
+};
+
+let routerStateObservable: any = null;
```
ChristophHi: (1) That's no observable (naming) 
(2) Please add typing where possible
```suggestion
let routerState: RouterState
```

feature-libs/product-configurator/common/components/configure-cart-entry/configure-cart-entry.component.ts
```diff
@@ -163,21 +174,45 @@ export class ConfigureCartEntryComponent {
...
+          let isFound = false;
+          locations.forEach((location) => {
+            if (routerState.state.semanticRoute === location) {
+              isFound = true;
+            }
+          });
+          return isFound;
+        })
+      )
+      .subscribe((value) => (isCartRelevant = value));
```
ChristophHi: Please never subscribe without ending the subscription (via unsubscribe or take(1)). This one will not be terminated and creates a memory leak!

feature-libs/product-configurator/common/components/configure-cart-entry/configure-cart-entry.component.ts
```diff
@@ -163,21 +174,45 @@ export class ConfigureCartEntryComponent {
...
+    return isCartRelevant;
```
ChristophHi: Here you cannot be sure that the subscription actually happened. Please return an observable instead, then you can get rid of the of the subscription at all

feature-libs/product-configurator/common/components/configure-cart-entry/configure-cart-entry.component.ts
```diff
@@ -163,21 +174,45 @@ export class ConfigureCartEntryComponent {
...
+  protected isCartRelevant(): boolean {
+    let isCartRelevant = false;
+    const locations = ['cart', 'product', 'search'];
+    this.routingService
+      .getRouterState()
+      .pipe(
+        map((routerState) => {
+          let isFound = false;
+          locations.forEach((location) => {
+            if (routerState.state.semanticRoute === location) {
```
ChristophHi: Why do product and search location mean that we are in the cart context?

ChristophHi (changes requested):

feature-libs/product-configurator/common/components/configure-cart-entry/configure-cart-entry.component.ts
```diff
@@ -178,6 +192,26 @@ export class ConfigureCartEntryComponent {
     };
   }
 
+  queryParams$: Observable<{
```
ChristophHi: As this is a member: Please move before class methods

feature-libs/product-configurator/common/components/configure-cart-entry/configure-cart-entry.component.ts
```diff
@@ -178,6 +192,26 @@ export class ConfigureCartEntryComponent {
...
+    navigateToCheckout: boolean;
+  }> = this.isCheckoutRelevant().pipe(
+    map((isCheckoutRelevant) => ({
+      forceReload: true,
+      resolveIssues: this.msgBanner && this.hasIssues(),
+      navigateToCheckout: isCheckoutRelevant,
+    }))
+  );
+
+  protected isCheckoutRelevant(): Observable<boolean> {
```
ChristophHi: Please rename
```suggestion
  protected isInCheckout(): Observable<boolean> {
```

ChristophHi (changes requested):

feature-libs/product-configurator/common/components/configure-cart-entry/configure-cart-entry.component.ts
```diff
@@ -163,12 +188,15 @@ export class ConfigureCartEntryComponent {
...
   /**
-   * Compiles query parameters for the router link. 'resolveIssues' is only set if the component is
+   * @deprecated use instead queryParams$
+   * Compiles query parameters for the router link.
+   * 'resolveIssues' is only set if the component is
    * rendered in the context of the message banner, and if issues exist at all
+   *
    * @returns Query parameters
    */
   getQueryParams(): Params {
```
Larisa-Staroverova, line 203: Should this method be deprecated?

feature-libs/product-configurator/common/components/configure-cart-entry/configure-cart-entry.component.ts
```diff
@@ -41,6 +47,20 @@ export class ConfigureCartEntryComponent {
     ? this.abstractOrderContext.key$
     : of({ type: AbstractOrderType.CART });
 
+  queryParams$: Observable<{
```
Larisa-Staroverova, line 50: Is it a breaking change?

ChristophHi: No I guess no

feature-libs/product-configurator/common/components/configure-cart-entry/configure-cart-entry.component.ts
ChristophHi: Yes (with version), to be removed in next major. Mention in respective ticket.

feature-libs/product-configurator/common/components/configure-cart-entry/configure-cart-entry.component.ts
Larisa-Staroverova: Is not breaking!

feature-libs/product-configurator/common/components/configure-cart-entry/configure-cart-entry.component.ts
Larisa-Staroverova: Done!

feature-libs/product-configurator/rulebased/components/product-title/configurator-product-title.component.ts
```diff
@@ -25,37 +25,52 @@ import { Configurator } from '../../core/model/configurator.model';
...
+        )
     )
-    .pipe(
-      tap(() => {
-        this.ghostStyle = false;
-      })
-    );
+  );
+
+  getProductCode(container: {
```
ChristophHi:
```suggestion
  protected getProductCode(container: {
```

feature-libs/product-configurator/rulebased/root/http-interceptors/configurator-bad-request.handler.ts
```diff
@@ -0,0 +1,88 @@
...
+})
+export class ConfiguratorBadRequestHandler extends HttpErrorHandler {
+  responseStatus = HttpResponseStatus.BAD_REQUEST;
+
+  constructor(protected globalMessageService: GlobalMessageService) {
+    super(globalMessageService);
+  }
+
+  getPriority(): Priority {
+    return Priority.HIGH;
```
ChristophHi: Why does the priority need to be high?

ChristophHi (changes requested):

feature-libs/product-configurator/rulebased/root/http-interceptors/configurator-bad-request.handler.ts
Larisa-Staroverova: I will change to `Priority.NORMAL`.
I retested it and it worked as expected.
The reason, why the priority was set to high as far as I remember, war because of the 2 error messages in the case of the the quick-order.

ChristophHi (approved):

PR #18506, opened 2024-02-15
PoC: live stock updates [ DO NOT MERGE ]

Platonn:
## Reloading STOCK product scope

- fix configuration of loading scopes, so `DETAILS` scope semantically _includes_ `STOCK` scope
- bind reloading of `STOCK` scope on the custom event `ReloadProductStockEvent`
- subscribe to various events (e.g. route change, timer or manual invocation) and dispatch `ReloadProductStockEvent`

## Calling different endpoint for STOCK product scope
- provide custom `ProductAdapter`, which calls a fake endpoint for `STOCK` product scope, but the original `OccProductAdapter` for all other scopes

## How to run it
Install:
```
npm install -g http-server
```
and then run:
```
node --watch http-proxy.js
```
## What it does
```
[localhost:9002] - proxies all requests to the OCC dev server, generating random `stockLevel`
                   if the request includes `fields=stock`. Random stock level is between 1-100.
[localhost:9003] - also runs a fake server that returns random stock levels between 100_001-100_100.
```

### DIFF
--- a/feature-libs/cart/base/components/add-to-cart/add-to-cart.component.html
+++ b/feature-libs/cart/base/components/add-to-cart/add-to-cart.component.html
@@ -1,4 +1,11 @@
 <form *ngIf="productCode" [formGroup]="addToCartForm" (submit)="addToCart()">
+  <b
+    >SPIKE STOCK LEVEL:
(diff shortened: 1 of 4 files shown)


Platonn:
Closing as not intended to merge

PR #18507, opened 2024-02-15
Add paymentSessionId in storage state

FollowTheFlo:
context:
FULL_PAGE pattern needs a new sub-pattern:
some PSP have predefined redirect url: eg: https://spartacus/redirect/success , means without parameters so without paymentSessionId value.
Thus we need to handle :
- url without params becomes valid
- a new way to retrieve paymentSessionId -> set in localStorage from CheckoutPage then get it from localStorage in VerificationPage.

### DIFF
--- a/integration-libs/opf/checkout/components/opf-checkout-payment-wrapper/opf-checkout-payment-wrapper.service.spec.ts
+++ b/integration-libs/opf/checkout/components/opf-checkout-payment-wrapper/opf-checkout-payment-wrapper.service.spec.ts
@@ -18,6 +18,7 @@ import { OpfCheckoutFacade } from '../../root/facade';
 import {
   OPF_PAYMENT_AND_REVIEW_SEMANTIC_ROUTE,
   OpfPaymentMethodType,
+  PaymentPattern,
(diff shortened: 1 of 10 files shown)


integration-libs/opf/checkout/components/opf-checkout-payment-wrapper/opf-checkout-payment-wrapper.service.spec.ts
```diff
@@ -147,6 +148,7 @@ describe('OpfCheckoutPaymentWrapperService', () => {
       Promise.resolve()
     );
     spyOn(service, 'renderPaymentGateway').and.callThrough();
+    spyOn<any>(service, 'storePaymentSessionId');
```
FollowTheFlo, line 151: cast to 'any' because storePaymentSessionId is a protected method.
same reason for below 'service a any' (line 199 and more).
Alternative is to define storePaymentSessionId as public but it would impact customer 'non-test' code.
better twick spec file than the code source file.

Matejk00 (approved):

PR #18512, opened 2024-02-19
feat: Generate data attributes for scripts 

Matejk00:
Closes: CXSPA-6138

### DIFF
--- a/integration-libs/opf/base/root/services/opf-resource-loader.service.ts
+++ b/integration-libs/opf/base/root/services/opf-resource-loader.service.ts
@@ -91,7 +91,7 @@ export class OpfResourceLoaderService extends ScriptLoader {
     resources: OpfDynamicScriptResource[],
     resolve: (value: void | PromiseLike<void>) => void
   ) {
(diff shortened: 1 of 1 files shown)


integration-libs/opf/base/root/services/opf-resource-loader.service.ts
```diff
@@ -91,14 +91,21 @@ export class OpfResourceLoaderService extends ScriptLoader {
     resources: OpfDynamicScriptResource[],
     resolve: (value: void | PromiseLike<void>) => void
   ) {
+    let attributes: any = {
```
FollowTheFlo: Sonar warning was raised to use const
detials in https://sonar.tools.sap/project/issues?resolved=false&severities=MINOR&inNewCodePeriod=true&pullRequest=18512&id=composable-storefront

Matejk00: Thanks. Solved it.

FollowTheFlo (approved):
good to me.

PR #18514, opened 2024-02-19
feature: (CXSPA-2190) - Keyboard controls for organization tables

Pio-Bar:
https://jira.tools.sap/browse/CXSPA-2190

Proposed keyboard controls for the organization/units tables.


feature-libs/organization/administration/components/unit/list/toggle-link/toggle-link-cell.component.html
```diff
@@ -1,8 +1,12 @@
...
+  [id]="model.uid"
   class="hide-focus-border"
   [routerLink]="{ cxRoute: route, params: routeModel } | cxUrl"
-  [tabIndex]="tabIndex"
+  (keydown)="onKeydown($event)"
+  [attr.aria-haspopup]="isSwitchable"
+  [attr.data-depth]="model.depthLevel"
+  role="treeitem"
 >
   <button
```
Zeyber, line 11: I know this is very old code and what you have done avoids the breaking change involved. I'm just not sure why we have a button in an anchor tag at all. I don't mind if you fix this or leave it though since its not in scope.

Zeyber (dismissed):
Very nice! It is up to you if you would like to improve the comment above. I don't think it has much value to do so since it won't affect SEO being behind authentication.

I can reapprove if you need to add to the feature toggle doc.

developpeurweb (dismissed):

Zeyber (approved):

PR #18520, opened 2024-02-20
fix: (CXSPA-2186) - Aria label input for generic link

Pio-Bar:
https://jira.tools.sap/browse/CXSPA-2186

I have added an optional input for the generic link. It allows us to set a more relevant content for screen readers to read out.
It should not need feature flags.


rmch91 (dismissed):

rmch91 (dismissed):

developpeurweb (dismissed):
Works well with JAWS/VO. Tested with AMP, no issues found.

Zeyber (approved):

PR #18523, opened 2024-02-20
feat: named feature flags mechanism for css styling

Platonn:
For feature flags in styles we will no longer use mixin `@include forVersion('x.y')`, but `@include forFeature('stringFeatureName')`, where the string feature name is synchronized with Spartacus global config `provideConfig({ features: { ... } })`.

Architecture overview:
- `FeatureStylesService` helper service in TS - applies/removes feature flags CSS classes on the root element, but only  _on demand_, when specific feature flags are _actually used_ on the current page
- `useFeatureStyles(feature)` function in TS  - called directly in components' constructors; it demands setting the relevant feature flag CSS class in the root element. (note: Unsetting the class is done automatically on components' destroy). It uses helper service `FeatureStylesService`, which under the hood counts usage references of each flag. 
- `forFeature(feature){ ... }` mixin in SCSS - adds extra CSS styles, for a selector `.cxFeat_{feature} &` - in other words: when `cxFeat_{feature}` is set on the ancestor element (implicitly: on the root element) then extra CSS styles will apply for the current selector `&`

Notes:
- feature flag CSS class shape: `cxFeat_{feature}`.
- the root element is Angular's root element, but not body. Philosophical concern: Angular app should not own body, but only it's own root element (i.e. cx-storefront)
- using mixin `forFeature()` directly inside styles of the root component (i.e. cx-storefront) won't work. In this speciic edge case, you have to add a special 2nd string param for the mixin: `'currentSelector'`, for example:
  ```scss
  %cx-storefront {
    @include forFeature('darkMode', 'currentSelector') {
      filter: invert(80%);
    }
  }
  ```

**Pros:**
- feature flag CSS is set on the root element **only if it's used anywhere** (only if any currently rendered component declares `useFeatureStyles(feature)`. Thanks to it, we avoid for example setting on on homepage any CSS feature flags related only to /my-account features.

**Cons:**
- developers needs to remember to declare `useFeatureStyles(feature)` in the component's constructor, if they want to use feature's styles
- [edge case] when using `forFeature()` SCSS mixin in the root element's styles, the 2nd param of string value `'currentSelector'` needs to be passed to the SCSS mixin `forFeature(..., 'currentSelector')`

related to https://jira.tools.sap/browse/CXSPA-6042

--
### Preparation of custom code:
1. add to app.module's providers:
```ts
    provideConfig({
      features: {
        cartItemActionsRoundedButtons: true,
        darkMode: true,
      },
    }),
```
2. call `useFeatureStyles()` in the constructors of components:
```ts
// cart-item-list.component.ts
/* ... */

  constructor(
    /* ... */
  ) {
    useFeatureStyles('cartItemActionsRoundedButtons');
  }
```

```ts
// storefront.component.ts
/* ... */

  constructor(
    /* ... */
  ) {
    useFeatureStyles('darkMode');
  }
```

3. Add feature-flagged styles:
```scss
// _cart-item-list.scss, paste it in the line 44:

      @include forFeature('cartItemActionsRoundedButtons') {
        background-color: /* pastel red */ #ff6666;
        color: white;
      }
    }

    .cx-sfl-btn {
      @include forFeature('cartItemActionsRoundedButtons') {
        background-color: /* light pastel blue */ #99ccff !important;
        color: white;
      }
```

```scss
// _storefront.scss, paste it in line 4:

  // It won't work, because cx-storefront is the only one exceptional case - it's the root component.
  @include forFeature('darkMode') {
    color: red;
  }

  @include forFeature('darkMode', 'currentSelector') {
    filter: invert(80%);
  }
```

2. run the storefront
3. play with feature flags config in `app.module.ts`, refresh page, and observe the changes in the UI


projects/core/src/features-config/directives/feature-styles.service.ts
```diff
@@ -0,0 +1,90 @@
...
+
+  protected renderer = this.rendererFactory.createRenderer(null, null);
+  protected targetElement: HTMLElement | undefined;
+  protected usagesCounter = new Map<string, number>();
+
+  init(_rootComponent: ComponentRef<any>): void {
+    this.targetElement = _rootComponent.location.nativeElement;
+
+    // Handle edge-case:
+    // Add classes for all features that were used (called `registerUsage()`) before before `init()` was invoked.
```
pawelfras: `nitpick`
```suggestion
    // Add classes for all features that were used (called `registerUsage()`) before `init()` was invoked.
```

projects/core/src/features-config/directives/feature-styles.service.ts
```diff
@@ -0,0 +1,90 @@
...
+  isDevMode,
+} from '@angular/core';
+import { LoggerService } from '../../logger/logger.service';
+import { FeatureConfigService } from '../services/feature-config.service';
+
+@Injectable({ providedIn: 'root' })
+export class FeatureStylesService {
+  protected featureConfig = inject(FeatureConfigService);
+  private rendererFactory = inject(RendererFactory2); // private, because needed only to create a renderer
+  protected logger = inject(LoggerService);
```
pawelfras: `nitpick`
```suggestion
  private rendererFactory = inject(RendererFactory2); // private, because needed only to create a renderer
  protected featureConfig = inject(FeatureConfigService);
  protected logger = inject(LoggerService);
```

projects/core/src/features-config/features-config.module.ts
```diff
@@ -26,6 +33,15 @@ export class FeaturesConfigModule {
...
           },
         }),
+        {
+          provide: APP_BOOTSTRAP_LISTENER,
+          multi: true,
+          useFactory: (): ((compRef: ComponentRef<any>) => void) => {
+            const featureCssService = inject(FeatureStylesService);
+            return (compRef: ComponentRef<any>) =>
+              featureCssService.init(compRef);
+          },
```
pawelfras: Man... what an excellent usage of `APP_BOOTSTRAP_LISTENER`

projects/core/src/features-config/utils/use-feature-styles.ts
```diff
@@ -0,0 +1,16 @@
...
+import { DestroyRef, inject } from '@angular/core';
+import { FeatureStylesService } from '../directives/feature-styles.service';
+
+export function useFeatureStyles(feature: string): void {
+  const featureStylesService = inject(FeatureStylesService);
+  const destroyRef = inject(DestroyRef);
+
+  featureStylesService.registerUsage(feature);
+  destroyRef.onDestroy(() => featureStylesService.unregisterUsage(feature));
+}
```
pawelfras:
```suggestion
export function useFeatureStyles(feature: string): void {
  assertInInjectionContext(useFeatureStyles);

  const featureStylesService = inject(FeatureStylesService);
  const destroyRef = inject(DestroyRef);

  featureStylesService.registerUsage(feature);
  destroyRef.onDestroy(() => featureStylesService.unregisterUsage(feature));
}
```

Thanks to `assertInInjectionContext`Angular will throw an error using function's name if the function is used outside the injection context:
```
Error: NG0203: useFeatureStyles() can only be used within an injection context such as a constructor, a factory function, a field initializer, or a function used with `runInInjectionContext`.
```

Platonn: good to know. thanks!

projects/core/src/features-config/directives/feature-styles.service.ts
```diff
@@ -0,0 +1,163 @@
...
+/**
+ * Automatically adds/removes CSS classes to/from the root element of the application,
+ * based on the explicitly registered usage of feature flags.
+ *
+ * The CSS classes are added only for flags that are currently used and enabled in the config.
+ *
+ * The CSS classes for flags that are enabled BUT NOT USED are NOT ADDED to the root element.
+ */
+@Injectable({ providedIn: 'root' })
+export class FeatureStylesService {
```
ChristophHi: Shouldn't we put this to parallel folder `projects/core/src/features-config/services `?

projects/storefrontstyles/scss/_features.scss
```diff
@@ -0,0 +1,62 @@
...
+//
+// Use case: Breaking style changes can be added for a specific feature flag
+// and customers can only use those styles if they explicitly enable a feature flag.
+//
+// The mixin requires only a $feature name as argument.
+//
+// The second argument with value `'currentSelector'` should be passed ONLY
+// in one edge case, when the feature-flagged styles are defined
+// for the Angular root element, i.e. `cx-storefront`.
+@mixin useFeatureStyles($feature, $classTarget: 'ancestorSelector') {
```
ChristophHi: I would not use `useFeatureStyles` here. For me `useFeatureStyles` means 'I want to register a feature flag for evaluation' while here it is about a check whether the flag is actually active. Also think about an additional mixin that checks if the flag is _not_ active (btw do you agree that we need that as well? But we could also use an additional parameter then).

So I think the initial `forFeature` was not that bad...

projects/core/src/features-config/directives/feature-styles.service.ts
Platonn: Thanks. I also noticed now it's not exported in public API, not deliberately

Platonn: moved to /services

projects/storefrontstyles/scss/_features.scss
Platonn: 1.
Thank you for feedback. Reverted mixin name to `forFeature`.

2.
> Also think about an additional mixin that checks if the flag is not active (btw do you agree that we need that as well? But we could also use an additional parameter then).
I'm not convinced yet we'd need an opposite mixin. Do you have any specific usecase in mind?
I imagine the normal styles should be defined normally, but only new breaking changes styles should be wrapped within the extra selector.

Theoretically we could add a negative variant of the mixin which under the hood would use `:not()`, but it may impact negatively the CSS performance and in extreme case make the browser slow down on user's interactions.
Sharing snippet just for reference:
```scss
@mixin notForFeature($feature) {
  $featureClass: getFeatureCssClass($feature);
  @at-root :not(.#{$featureClass}) & {
    @content;
  }
}
```
If you have other ideas of implementing a notForFeature, that doesn't use `:not()`, please let me know.

All that said, I'd love to understand an usecase first, in which we couldn't apply simply: default styles by default + conditional styles on top of it (if flag is enabled)

ChristophHi: Here I was wrong and confused view changes (where we frequently worked with a negative variant) with SCSS changes. I confirmed with Larisa, who mostly takes care of our styling: On the SCSS side, we should bei fine with this additive approach.

kpawelczak (approved):

PR #18532, opened 2024-02-21
fix: (CXSPA-1049) - Keep focus on scroll to top button when clicked by keyboard

Pawulon-0:
Closes: https://jira.tools.sap/browse/CXSPA-1049


projects/storefrontlib/layout/main/storefront.component.ts
```diff
@@ -38,7 +38,7 @@ export class StorefrontComponent implements OnInit, OnDestroy {
   @HostBinding('attr.role') role = 'presentation';
 
   // required by esc focus
-  @HostBinding('tabindex') tabindex = '0';
+  @HostBinding('tabindex') tabindex = '-1';
```
Zeyber: Just to understand this issue, would it be possible to remove the `tabindex` binding altogether and have the same? We also need to add breaking change flags to this as its a very destructive change being the root component.

projects/storefrontlib/cms-components/navigation/scroll-to-top/scroll-to-top.component.ts
```diff
@@ -53,25 +57,65 @@ export class ScrollToTopComponent implements OnInit {
...
+  }
+
+  protected setConfig(): void {
+    this.componentData.data$.pipe(take(1)).subscribe((data) => {
+      this.scrollBehavior = data.scrollBehavior ?? this.scrollBehavior;
+      this.displayThreshold = data.displayThreshold ?? this.displayThreshold;
+    });
+  }
+
+  private switchDisplay(): void {
```
Zeyber: We use `protected` instead of `private` because we like to have all our code extendible to our users. It is important as if there are any issues with our library, our users may need to change or extend to solve an issue with our function. With `private` methods, it is not possible and they need to clone the library to be able to modify any private method which is cumbersome.

With `protected` methods, we can still show this is not intended to be a public facing method, but it does exist and the user can modify it if they need to.

projects/storefrontapp-e2e-cypress/cypress/support/utils/a11y-tab.ts
```diff
@@ -20,7 +20,7 @@ export function getNextFocusableElement(
   elements: HTMLElement[],
   activeElementIndex: number
 ): HTMLElement {
-  return activeElementIndex + 1 <= elements.length
+  return activeElementIndex + 1 < elements.length
```
Zeyber: Are we changing this because we are no longer tabbing to the root element? Could be best to comment to understand why we are making this change, what elements we are counting.

projects/storefrontlib/cms-components/navigation/scroll-to-top/scroll-to-top.component.ts
```diff
@@ -53,25 +57,65 @@ export class ScrollToTopComponent implements OnInit {
     this.setConfig();
   }
 
-  protected setConfig(): void {
-    this.componentData.data$.pipe(take(1)).subscribe((data) => {
-      this.scrollBehavior = data.scrollBehavior ?? this.scrollBehavior;
-      this.displayThreshold = data.displayThreshold ?? this.displayThreshold;
-    });
+  @HostListener('window:scroll', ['$event'])
+  onScroll(): void {
```
Zeyber, line 64: We unfortunately have the tricky situation where we need to maintain the old behaviour and allow "opt-in" for clients to new features using the feature flag mechanism. In this situation, we will need to use the new behaviour only when the feature is enabled (eg. `if(this.featureConfigService.isEnabled('myFeatureOrFix') { ...new behaviour... } else { ...old behaviour ...}`). 

This also is relevant when we add new behaviour to an existing function. We are never sure how users may have extended our library apis so we need to be cautious about changing anything. Feel free to ask me or @Pio-Bar if you are unsure about anything :)

[Here is an example of using these feature flags.](https://github.com/SAP/spartacus/pull/18493/files)

projects/storefrontlib/cms-components/navigation/scroll-to-top/scroll-to-top.component.ts
```diff
@@ -53,25 +57,65 @@ export class ScrollToTopComponent implements OnInit {
...
   /**
    * Scroll back to the top of the page and set focus on top most focusable element.
    */
-  scrollToTop(): void {
-    // Focus first focusable element within the html body
-    this.selectFocusUtility
-      .findFirstFocusable(this.winRef.document.body, { autofocus: '' })
-      ?.focus();
-
+  scrollToTop(event: MouseEvent): void {
```
Zeyber: Lets make `event` optional so that it does not cause a breaking change (ie. changing method parameters).

projects/storefrontlib/cms-components/navigation/scroll-to-top/scroll-to-top.component.ts
Zeyber: Maybe we could comment what this function `switchDisplay()` is doing?

projects/storefrontlib/cms-components/navigation/scroll-to-top/scroll-to-top.component.ts
```diff
@@ -53,25 +57,65 @@ export class ScrollToTopComponent implements OnInit {
...
     this.window?.scrollTo({
       top: 0,
       behavior: this.scrollBehavior,
     });
+
+    this.clickEventSource = event.detail;
+  }
+
+  focusOut(): void {
```
Zeyber: Sorry for misunderstanding. What does focus out mean?

projects/storefrontlib/cms-components/navigation/scroll-to-top/scroll-to-top.component.ts
```diff
@@ -53,25 +57,65 @@ export class ScrollToTopComponent implements OnInit {
...
+    if (this.display) {
+      this.switchDisplay();
+    }
+  }
+
+  @HostListener('document:keydown', ['$event'])
+  handleKeyboardEvent(event: KeyboardEvent): void {
+    if (
```
Zeyber, line 110: I think we can hide this function behind the feature flag within this statement.

projects/storefrontlib/cms-components/navigation/scroll-to-top/scroll-to-top.component.ts
Zeyber: I think we can cover ourselves from breaking changes adding a feature flag to the if condition in this method.

Zeyber (changes requested):
E2E test looks good. It looks like it is checking mouse and keyboard events, checking element visibility as well as screen positioning. Nice work!

projects/storefrontlib/layout/main/storefront.component.ts
Pio-Bar: I've reversed this change. This would break too many existing features and it's not necessary for this one to work.

projects/storefrontlib/cms-components/navigation/scroll-to-top/scroll-to-top.component.ts
Pio-Bar: done

projects/storefrontapp-e2e-cypress/cypress/support/utils/a11y-tab.ts
Pio-Bar: Not necessary anymore

projects/storefrontlib/cms-components/navigation/scroll-to-top/scroll-to-top.component.ts
Pio-Bar: I've added comments and refactored for readability.

projects/storefrontlib/cms-components/navigation/scroll-to-top/scroll-to-top.component.ts
Pio-Bar: Name changed to be more clear + feature flag added

projects/storefrontlib/cms-components/navigation/scroll-to-top/scroll-to-top.component.ts
Pio-Bar: done

projects/storefrontlib/cms-components/navigation/scroll-to-top/scroll-to-top.component.ts
Pio-Bar: New feature flags have been added ;)

projects/storefrontapp/src/app/spartacus/spartacus-b2c-configuration.module.ts
```diff
@@ -52,6 +56,11 @@ const baseSite = environment.epdVisualization
         },
       },
     }),
+    provideConfig(<FeaturesConfig>{
```
Zeyber: We can remove this config. We don't want it enabled by default.

projects/storefrontlib/cms-components/navigation/scroll-to-top/scroll-to-top.component.spec.ts
```diff
@@ -43,86 +21,119 @@ const mockData: CmsScrollToTopComponent = {
...
-    fixture.detectChanges();
-    spyOn(focusUtility, 'findFirstFocusable').and.callThrough();
-    spyOn(window, 'scrollTo').and.callThrough();
-    winRef.nativeWindow?.scrollTo(0, 200);
-    winRef.nativeWindow?.dispatchEvent(new Event('scroll'));
+  it('should scroll window to top when clicked', () => {
+    spyOn<any>(component['window'], 'scrollTo');
+
+    component.scrollToTop(new MouseEvent('click'));
+    expect(component['window']?.scrollTo).toHaveBeenCalled();
```
Zeyber: Bit of a nitpick, but should we expect a value that `scrollTo` was called with? ie. zero for top?

projects/storefrontlib/cms-components/navigation/scroll-to-top/scroll-to-top.component.ts
```diff
@@ -7,41 +7,48 @@
...
   templateUrl: './scroll-to-top.component.html',
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
 export class ScrollToTopComponent implements OnInit {
-  iconTypes = ICON_TYPE;
+  readonly iconTypes = ICON_TYPE;
 
   @HostBinding('class.display')
-  display: boolean | undefined;
+  display: boolean = false;
```
Zeyber: Lets avoid changing the type here so that `display` can still be `undefined` which can be treated as `falsey`.

projects/storefrontlib/cms-components/navigation/scroll-to-top/scroll-to-top.component.ts
```diff
@@ -7,41 +7,48 @@
...
+import { ICON_TYPE } from '../../misc/icon/icon.model';
 
 @Component({
   selector: 'cx-scroll-to-top',
   templateUrl: './scroll-to-top.component.html',
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
 export class ScrollToTopComponent implements OnInit {
-  iconTypes = ICON_TYPE;
+  readonly iconTypes = ICON_TYPE;
```
Zeyber: Any reason to change this?

Zeyber (changes requested):

projects/storefrontapp/src/app/spartacus/spartacus-b2c-configuration.module.ts
Pio-Bar: Removed

projects/storefrontlib/cms-components/navigation/scroll-to-top/scroll-to-top.component.spec.ts
Pio-Bar: I've provided the values.

projects/storefrontlib/cms-components/navigation/scroll-to-top/scroll-to-top.component.ts
Pio-Bar: Change Reverted

projects/storefrontlib/cms-components/navigation/scroll-to-top/scroll-to-top.component.ts
Pio-Bar: Can't see a reason why this was changed, I'm reverting it.

Zeyber (dismissed):

Zeyber (dismissed):

Zeyber (approved):

PR #18533, opened 2024-02-21
fix: navigation issue in the paragraph component

kimhw0630:
Closes: https://jira.tools.sap/browse/CXSPA-5944

We had three different solutions, and this is the third solution outlined in Slock.
https://sap-cx.slack.com/archives/C04UZF3QKQ8/p1708469548528539


rmch91 (approved):

projects/storefrontlib/router/on-navigate.service.ts
```diff
@@ -85,17 +85,49 @@ export class OnNavigateService {
...
+    currentRoute: Scroll,
+    position: [number, number] | null
+  ): void {
+    const scrollTo = (
+      anchor: string | null,
+      scrollPosition: [number, number]
+    ) => {
+      if (
+        anchor &&
+        (this.router as any).options?.anchorScrolling === 'enabled'
```
Platonn, line 111: This is using private API of Angular Router. See Angular source code:
https://github.com/angular/angular/blob/1e3bcfecbab260fb66a1ac50e77192bb98cc0f3e/packages/router/src/router.ts#L83

Private API might be changed anytime by Angular team (even in patch releases), therefore by using this API we're risking that potentially our library code may break in runtime anytime when customer updates their Angular version to a newer one, that changes the name of this property `options`. To be honest, I believe there are small chances Angular team will rename this property out of sudden, but they have right to do it. So _in general_ we should avoid using private API of Angular.

Fortunately the injection token `ROUTER_CONFIGURATION` is public, so we can safely inject it in our class and use it. So instead of picking private property `options` from `Router`, let's inject the `ROUTER_CONFIGURATION` ourselves, e.g.
```ts
protected readonly routerConfiguration  = inject(ROUTER_CONFIGURATION, {optional: true}) || {};
```
```ts
...
if (
  ... &&
  this.routerConfiguration.anchorScrolling === 'enabled'
```

PR #18537, opened 2024-02-22
fix: CXSPA-6327 Use ExpressServerLogger type when logger is initialized in OptimizedSsrEngine

pawelfras:
closes [CXSPA-6327](https://jira.tools.sap/browse/CXSPA-6327)

### DIFF
--- a/core-libs/setup/ssr/optimized-engine/optimized-ssr-engine.ts
+++ b/core-libs/setup/ssr/optimized-engine/optimized-ssr-engine.ts
@@ -11,7 +11,6 @@ import { NgExpressEngineInstance } from '../engine-decorator/ng-express-engine-d
 import { getRequestUrl } from '../express-utils/express-request-url';
 import {
   EXPRESS_SERVER_LOGGER,
-  ExpressLoggerService,
   ExpressServerLogger,
   ExpressServerLoggerContext,
 } from '../logger';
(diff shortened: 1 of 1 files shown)


core-libs/setup/ssr/optimized-engine/optimized-ssr-engine.ts
```diff
@@ -78,7 +77,7 @@ export class OptimizedSsrEngine {
         }
       : undefined;
 
-    this.logger = this.ssrOptions?.logger as ExpressLoggerService; // we are sure the logger is defined in this place
+    this.logger = this.ssrOptions?.logger as ExpressServerLogger; // we are sure the logger is defined in this place
```
Platonn: For runtime code safety, what about checking if it's really defined?

```suggestion
    if (!this.ssrOptions?.logger) {
      throw new Error('`SsrOptimizationOptions.logger` is not defined');
    }
    this.logger = this.ssrOptions?.logger as ExpressServerLogger;
```

pawelfras: We're pretty sure it's defined as if it's not provided by customers, it takes default value from default optimization options which (probably) cannot be overwritten by customers.

But I agree - can add this extra step for runtime code safety.

pawelfras: I believe `as ExpressServerLogger;` is no longer needed as we checked if the value is defined.

Platonn (approved):

PR #18539, opened 2024-02-22
fix: (CXSPA-2243) - Improved a11y for popovers in 'organization'

Pio-Bar:
https://jira.tools.sap/browse/CXSPA-2243

Improved screen reader readout for information buttons in 'My Company' subpages.


rmch91 (dismissed):

developpeurweb (changes requested):
This works better now for both JAWS and VO.
There are three ways of closing the popover while using the keyboard, regardless of Screen Reader use, either pressing Spacebar or Enter will toggle the display and the narration, and then continue tabbing to the next element without breaking the order.
However, a third way is pressing the ESC key, and this resets the focus to the top of the page thus breaking the order.
Is there a way to set the ESC key to only close the popover without the focus moving to the top of the page and continue with the natural order expectation?

Pio-Bar:
I've modified it to restore focus to the opener element regardless of how it was closed.

projects/storefrontlib/shared/components/popover/popover.service.ts
```diff
@@ -31,13 +31,9 @@ export class PopoverService {
     return config;
   }
 
-  setFocusOnElement(
-    element: ElementRef,
-    focusConfig: FocusConfig,
-    appendToBody?: boolean
-  ) {
-    if (focusConfig && appendToBody) {
+  setFocusOnElement(element: ElementRef) {
```
Zeyber: Breaking changes need handling :)

Zeyber (changes requested):

projects/storefrontlib/shared/components/popover/popover.service.ts
Pio-Bar: Feature flags added ;)

Zeyber (dismissed):

developpeurweb (dismissed):

Zeyber (approved):

PR #18556, opened 2024-02-27
feat: Add ApplePay Cart support

Matejk00:
Closes: CXSPA-5549

### DIFF
--- a/integration-libs/opf/base/components/opf-quick-buy/apple-pay/apple-pay.component.ts
+++ b/integration-libs/opf/base/components/opf-quick-buy/apple-pay/apple-pay.component.ts
@@ -8,10 +8,10 @@ import {
   ChangeDetectionStrategy,
   Component,
   Input,
-  OnDestroy,
   OnInit,
   inject,
 } from '@angular/core';
+import { Cart } from '@spartacus/cart/base/root';
 import { Product } from '@spartacus/core';
 import {
   OpfCartHandlerService,
@@ -20,15 +20,16 @@ import {
 import {
   ActiveConfiguration,
   DigitalWalletQuickBuy,
-  OpfPaymentError,
   OpfProviderType,
+  OpfQuickBuyLocation,
 } from '@spartacus/opf/base/root';
 import {
   CurrentProductService,
   ItemCounterService,
 } from '@spartacus/storefront';
-import { Observable, Subscription, combineLatest } from 'rxjs';
-import { switchMap } from 'rxjs/operators';
+import { Observable, combineLatest } from 'rxjs';
+import { switchMap, take } from 'rxjs/operators';
+import { OpfQuickBuyService } from '../opf-quick-buy.service';
 import { ApplePaySessionFactory } from './apple-pay-session';
 import { ApplePayService } from './apple-pay.service';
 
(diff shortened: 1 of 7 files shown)


integration-libs/opf/base/components/opf-quick-buy/apple-pay/apple-pay.component.ts
```diff
@@ -68,31 +69,47 @@ export class ApplePayComponent implements OnInit, OnDestroy {
...
-      .pipe(
-        switchMap(([product, _]) =>
-          this.applePayService.start(
-            product as Product,
-            this.itemCounterService.getCounter(),
-            this.applePayDigitalWallet?.countryCode as string
-          )
-        )
+    ]).pipe(
+      take(1),
```
rmch91, line 77: Could we move `take(1)` to the end?
Operators are executed in the order they are declared. This means that data transformations and side effects occur sequentially. By placing `take*` operators at the end, you ensure that all prior operators in the chain have had the chance to act on each emitted value.

integration-libs/opf/base/components/opf-quick-buy/apple-pay/apple-pay.component.ts
```diff
@@ -68,31 +69,47 @@ export class ApplePayComponent implements OnInit, OnDestroy {
...
+    );
   }
 
-  ngOnDestroy(): void {
-    if (this.sub) {
-      this.sub.unsubscribe();
-    }
+  initActiveCartTransaction(): Observable<unknown> {
+    return this.cartHandlerService.getCurrentCart().pipe(
+      take(1),
```
rmch91, line 90: Same as above

rmch91 (changes requested):

integration-libs/opf/base/components/opf-quick-buy/apple-pay/apple-pay.component.ts
FollowTheFlo: are you sure we will get the same behavior by putting take(1) at the end?
the idea is to prevent what is above take(1) to stream additional data.
However, i think if we put concatMap instead of switchMap, we can then put take(1) at the end as we ensure that the first value will be proceed with no chance to be cancelled with switchMap.

integration-libs/opf/base/components/opf-quick-buy/apple-pay/apple-pay.service.ts
```diff
@@ -41,44 +43,77 @@ export class ApplePayService {
...
+    this.transactionDetails = { ...this.initialTransactionDetails };
+
+    if (transactionInput?.cart) {
+      this.transactionDetails = {
+        ...this.transactionDetails,
+        context: OpfQuickBuyLocation.CART,
+        cart: transactionInput.cart,
+        total: {
+          amount: `${transactionInput.cart.totalPrice?.value}`,
+          label: `${transactionInput.cart.code}`,
```
FollowTheFlo, line 74: as discussed, the company name should be the total.label
https://developer.apple.com/documentation/apple_pay_on_the_web/applepaypaymentrequest/1916119-total
I'll create a separate ticket for it.

rmch91 (approved):

integration-libs/opf/base/components/opf-quick-buy/apple-pay/apple-pay.component.ts
Matejk00: After discussion with @rmch91 we agreed that current shape works fine in this case. There is no need to move take(1) operator to the end.

PR #18571, opened 2024-02-29
fix(CXSPA-6418): replace broken references to removed EPD launchpad

Reid-McNaughton:
Also changed CCv2 URL is use a working CCv2 server.


sam-garland:
Reopened as https://github.com/SAP/spartacus/pull/18767

PR #18572, opened 2024-02-29
fix(CXSPA-6418): replace broken references to removed EPD launchpad

Reid-McNaughton:
Also changed CCv2 URL to use a working CCv2 server.


sam-garland:
@hackergil - Can you please review? The existing URLs do not work. The URLs in this PR are needed to be able to run e2e tests.

PR #18574, opened 2024-03-01
MultiCart with switch solution

FollowTheFlo:


integration-libs/opf/base/components/opf-quick-buy/apple-pay/apple-pay.component.ts
```diff
@@ -52,6 +53,9 @@ export class ApplePayComponent implements OnInit {
   isApplePaySupported$: Observable<boolean>;
   applePayDigitalWallet?: DigitalWalletQuickBuy;
 
+  ngOnDestroy(): void {
+    this.cartHandlerService.allowMiniCartUpdate();
```
FollowTheFlo: we want to ensure the cart does not stay blocked. Even though the logic should be handled inside cartHandlerService, we want to avoid any obscured scenario.

integration-libs/opf/base/core/services/opf-cart-handler.service.ts
```diff
@@ -51,23 +50,134 @@ export class OpfCartHandlerService {
...
+    return this.checkStableCart().pipe(
+      switchMap(() => this.getCurrentCartId()),
+      map((cartId) => {
+        this.cartHandlerState.cartId = cartId;
+        return true;
+      })
+    );
+  }
+
+  allowMiniCartUpdate() {
```
Matejk00: Would you consider creating a mirror function that is more reusable within the cart-handler service?

```
blockMiniCartComponentUpdate(decision: boolean) {
    this.opfMiniCartComponentService.blockUpdate(decision);
}
```

integration-libs/opf/base/core/services/opf-cart-handler.service.ts
```diff
@@ -51,23 +50,134 @@ export class OpfCartHandlerService {
...
+    pickupStore?: string | undefined
+  ): Observable<boolean> {
+    this.cartHandlerState = { ...this.initiateCartHandlerState };
+    return combineLatest([
+      this.userIdService.takeUserId(),
+      this.multiCartFacade.getCartIdByType(CartType.ACTIVE),
+      this.activeCartFacade.isStable(),
+    ]).pipe(
+      take(1),
+      tap(() => this.opfMiniCartComponentService.blockUpdate(true)),
```
Matejk00: then we can use a function directly from the service.

```
      tap(() => this.blockMiniCartComponentUpdate(true)),
```

integration-libs/opf/base/core/services/opf-cart-handler.service.ts
```diff
@@ -51,23 +50,134 @@ export class OpfCartHandlerService {
...
+      this.opfMiniCartComponentService.blockUpdate(false);
+      return of(false);
+    }
+    this.multiCartFacade.loadCart({
+      cartId: this.cartHandlerState.previousCartId,
+      userId: this.cartHandlerState.userId,
+      extraData: { active: true },
+    });
+    return this.checkStableCart().pipe(
+      tap(() => this.opfMiniCartComponentService.blockUpdate(false))
```
Matejk00: same here but with the different argument

```
      tap(() => this.blockMiniCartComponentUpdate(false)),
```

integration-libs/opf/base/root/services/opf-mini-cart-component.service.ts
```diff
@@ -0,0 +1,51 @@
...
+import { Observable } from 'rxjs';
+import { filter } from 'rxjs/operators';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class OpfMiniCartComponentService extends MiniCartComponentService {
+  protected isUpdateBlocked = false;
+
+  blockUpdate(val: boolean) {
```
Matejk00: Can we add a comment on top that describes a context of this method?

integration-libs/opf/base/core/services/opf-cart-handler.service.ts
```diff
@@ -51,23 +50,134 @@ export class OpfCartHandlerService {
   protected eventService = inject(EventService);
   protected checkoutBillingAddressFacade = inject(CheckoutBillingAddressFacade);
   protected opfGlobalMessageService = inject(OpfGlobalMessageService);
+  protected opfMiniCartComponentService = inject(OpfMiniCartComponentService);
+  protected initiateCartHandlerState: CartHandlerState = {
```
Matejk00: nitpick:
I think `defaultCartHandlerState` fits better for me or `initialCartHandlerState`.
I believe `initiate` is more appropriate for a function that actually initializes a state, but this is just an assignment.

integration-libs/opf/base/components/opf-quick-buy/apple-pay/apple-pay.service.ts
```diff
@@ -375,6 +379,35 @@ export class ApplePayService {
       );
   }
 
+  protected loadPdpOriginalCart(orderSuccess = false): void {
```
Matejk00: I'm steering clear of using the name "Pdp". In GooglePay and ApplePay services, it was referred to as `SingleProduct`, and the context also resembles `OpfQuickBuyLocation.PRODUCT`.

Perhaps we could consider renaming it to `loadCartAfterSingleProductTransaction`?

Matejk00 (changes requested):

integration-libs/opf/base/core/services/opf-cart-handler.service.ts
```diff
@@ -151,18 +263,55 @@ export class OpfCartHandlerService {
...
+    quantity: number
+  ): Observable<boolean> {
+    if (
+      !this.cartHandlerState.previousCartId ||
+      !this.cartHandlerState.userId
+    ) {
+      return of(false);
+    }
+    return this.multiCartFacade
+      .getEntry(this.cartHandlerState.previousCartId, productCode)
```
FollowTheFlo, line 292: We were previously using 'activeCart' to getEntry, but was getting 'undefined' intermittently.
so, moved to 'multipleCart' to fix it.
I guess root cause was due to race condition where previousCart was not yet loaded as activeCart.

Matejk00 (approved):

PR #18579, opened 2024-03-04
chore: Angular update

Matejk00:
Closes: CXSPA-6404


integration-libs/opf/base/core/facade/opf-global-functions.service.ts
```diff
@@ -204,23 +204,24 @@ export class OpfGlobalFunctionsService implements OpfGlobalFunctionsFacade {
...
+            .submitPayment({
+              additionalData,
+              paymentSessionId,
+              cartId,
+              callbackArray,
+              paymentMethod,
+              returnPath: undefined,
             })
-          )
-          .toPromise();
```
Matejk00, line 223: Because `toPromise` (https://rxjs.dev/deprecations/to-promise) is deprecated I had to refactor two functions in this file. They need to be verified if works as expected.

Matejk00:
Sonar issues are not related to the OPF code. They point out some migration files, and on develop and develop-next-major branches they were whitelisted.

integration-libs/opf/base/core/facade/opf-global-functions.service.ts
FollowTheFlo: Code change is good to me.
Tested succefully, 'finalize' gets calle as expecte and spinner gets closed as expected
image

FollowTheFlo (approved):
good to me. tests good too.

PR #18584, opened 2024-03-04
fix: (CXSPA-2244) - Popover now renders in aria-live

Pio-Bar:
[Ticket](https://jira.tools.sap/browse/CXSPA-2244)

The directive first creates an aria-live wrapper and then appends the popover to it. With this setup the screen reader should be aware of the changes.


projects/storefrontlib/shared/components/popover/popover.directive.ts
```diff
@@ -237,4 +303,6 @@ export class PopoverDirective implements OnInit {
     protected popoverService: PopoverService,
     protected winRef: WindowRef
   ) {}
+  @Optional() featureFlagService = inject(FeatureConfigService);
+  @Optional() selectFocusUtility = inject(SelectFocusUtility);
```
Zeyber: Same comment as [here](https://github.com/SAP/spartacus/pull/18609#discussion_r1521038283). Lets also put these at the top of the class as that's where we would usually expect to see our injected services.

Zeyber (changes requested):

projects/storefrontlib/shared/components/popover/popover.directive.ts
Pio-Bar: Moved to top and added the optional flag

Zeyber: Lets also add our TODO cleanup comments. ie. `featureFlagService` to be removed and `selectFocusUtility` to be required next major.

projects/storefrontlib/shared/components/popover/popover.directive.ts
```diff
@@ -138,41 +141,76 @@ export class PopoverDirective implements OnInit {
...
+  });
+
   /**
    * Method performs open action for popover component.
    */
+  // TODO: (CXSPA-6442) - remove feature flags next major release
   open(event: PopoverEvent) {
     if (!this.cxPopoverOptions?.disable) {
+      if (
+        this.featureFlagService?.isLevel('6.8') &&
```
Zeyber: Need to update these to use `isEnabled` flagging.

Zeyber (changes requested):

projects/storefrontlib/shared/components/popover/popover.directive.ts
Pio-Bar: done

projects/storefrontlib/shared/components/popover/popover.directive.ts
Pio-Bar: Changed to the new feature flag system.

Zeyber (dismissed):

developpeurweb (changes requested):
Tested with JAWS and VO, although I can see the aria-live was included, it's not being vocalized by either screen reader. 
A similar scenario was fixed in ticket CXSPA-2243 and works well in there, can we please review the approach.

PR #18589, opened 2024-03-05
CXINT-1000: Toast Message to show profile update

Akash-SAP:
CXINT-1000: Toast Message to show profile update


integration-libs/cdc/root/service/cdc-js.service.ts
```diff
@@ -574,11 +574,34 @@ export class CdcJsService implements OnDestroy {
...
       });
     }
   }
 
+  /**
+   * handle toast message on profile update
+   * @param response
+   *
+   */
+  handleProfileUpdateTostMessage(response?: any) {
```
vigneshkrish8691: make the method from handleProfileUpdateTostMessage - > handleProfileUpdateToastMessage

integration-libs/cdc/root/service/cdc-js.service.spec.ts
```diff
@@ -620,6 +620,34 @@ describe('CdcJsService', () => {
     });
   });
 
+  describe('handleProfileUpdateTostMessage', () => {
```
vigneshkrish8691: Correct the method name as mentioned below

integration-libs/cdc/assets/translations/en/cdc.json
anjana-bl, line 1: Suggestion: can we rename `"profileDetailUpdateSuccess"` to `"profileUpdateSuccess" `- just to shorten the length. Similarly,  `"profileDetailUpdateFailure"` to  `"profileUpdateFailure"` ?

integration-libs/cdc/root/service/cdc-js.service.ts
```diff
@@ -574,11 +574,34 @@ export class CdcJsService implements OnDestroy {
...
+    if (response.response.errorCode === 0) {
```
anjana-bl: HIGH: This 'reponse' can be undefined. So make it, if (response?.response?.errorCode === 0)

integration-libs/cdc/root/service/cdc-js.service.ts
```diff
@@ -574,11 +574,34 @@ export class CdcJsService implements OnDestroy {
           if (currentEmail !== userDetails.uid) {
             this.logoutUser();
           }
+          this.handleProfileUpdateTostMessage(response);
         });
       });
     }
   }
```
anjana-bl, line 584: HIGH: there is an if block from line 566-582, for this if block. 
can we add an 'else' block to it like below:
`else{this.handleProfileUpdateTostMessage();}` 
so that, error message can be displayed in case you don't receive any response from CDC ??

integration-libs/cdc/root/service/cdc-js.service.ts
```diff
@@ -574,11 +574,34 @@ export class CdcJsService implements OnDestroy {
...
 
+  /**
+   * handle toast message on profile update
+   * @param response
+   *
+   */
+  handleProfileUpdateToastMessage(response?: any) {
```
anjana-bl: MEDIUM: can we make this method `protected` as it is used only within this class?

anjana-bl: Can we change the method name from `handleProfileUpdateToastMessage` to `handleProfileUpdateResponse` ? because 'toast' is not usually used in any libraries in spa.

integration-libs/cdc/root/service/cdc-js.service.spec.ts
Akash-SAP: Done

integration-libs/cdc/root/service/cdc-js.service.ts
Akash-SAP: Done

integration-libs/cdc/assets/translations/en/cdc.json
Akash-SAP: Done

integration-libs/cdc/root/service/cdc-js.service.ts
Akash-SAP: Done

integration-libs/cdc/root/service/cdc-js.service.ts
Akash-SAP: Done

integration-libs/cdc/root/service/cdc-js.service.ts
Akash-SAP: Done

integration-libs/cdc/root/service/cdc-js.service.ts
Akash-SAP: Done

vigneshkrish8691 (dismissed):

integration-libs/cdc/assets/translations/translations.ts
```diff
@@ -12,5 +12,5 @@ export const cdcTranslations: TranslationResources = {
 };
 
 export const cdcTranslationChunksConfig: TranslationChunksConfig = {
-  cdc: ['reconsent'],
+  cdc: ['reconsent', 'profile'],
```
Platonn: If possible, please prefix your main translation keys with `cdc`, e.g. `cdcProfile`. Too bad we have `reconsent` already without `cdc` prefix, but changing it would be a potential breaking change to customers. So let's leave it.

I apologize, such a recommendation should have been documented more clearly before.

Why: If you use the general term `profile`, then no other features can use this term - as you map it to only `"cdc(.json)"` lazy loadable translations chunk.

Akash-SAP: done

Platonn (approved):

PR #18609, opened 2024-03-08
fix: (CXSPA-2798) - Show assistive message when active view changes

Pio-Bar:
https://jira.tools.sap/browse/CXSPA-2798

When the active view changes in the 'split view' component an assistive message should be heard.


projects/storefrontlib/shared/components/split-view/view/view.component.ts
```diff
@@ -73,6 +83,9 @@ export class ViewComponent implements OnInit, OnDestroy {
     protected cd: ChangeDetectorRef
   ) {}
 
+  @Optional() protected globalMessageService = inject(GlobalMessageService);
+  @Optional() featureConfigService = inject(FeatureConfigService);
```
Zeyber: It is also possible to to add the `inject( ... , {optional:true})` flag to the injector in addition to the `@Optional()`. Why? I think there was a case where a linter would react to one of those and not some other tool. I can't remember the specifics unfortunately. This just makes sure that we have covered our bases :)

feature-libs/organization/administration/components/shared/card/card.component.spec.ts
```diff
@@ -41,6 +45,8 @@ describe('CardComponent', () => {
           provide: ItemService,
           useClass: MockItemService,
         },
+        { provide: GlobalMessageService, useClass: MockGlobalMessageService },
+
         SplitViewService,
```
Zeyber, line 50: I wonder if we should cover this scenario in e2e too. Especially in the b2b administration area. The mechanism can be easy to break and we aren't catching errors properly.

This is a recent situation where our e2e test coverage failed us and we introduced issues to the b2b my company panel without realizing: https://github.com/SAP/spartacus/issues/18232

Feel free to ask for help on this topic.

Zeyber (changes requested):

projects/storefrontlib/shared/components/split-view/view/view.component.ts
Pio-Bar: Good to know, I've included the flag too.

feature-libs/organization/administration/components/shared/card/card.component.spec.ts
Pio-Bar: I've provided a short e2e test, let me know what you think

projects/storefrontlib/shared/components/split-view/view/view.component.ts
```diff
@@ -129,6 +152,24 @@ export class ViewComponent implements OnInit, OnDestroy {
     }
   }
 
+  protected showAssistiveMessage() {
```
Zeyber: Since this creates a subscription that does something on some action, lets include this in the title to be more descriptive. Something like `onViewChangeShowAssistiveMessage`.

projects/storefrontlib/shared/components/split-view/view/view.component.ts
```diff
@@ -88,6 +107,10 @@ export class ViewComponent implements OnInit, OnDestroy {
         this.disappeared = view.hidden;
         this.cd.markForCheck();
       });
+    // TODO: (CXSPA-6475) - Remove feature flag next major release
+    if (this.featureConfigService?.isLevel('6.8')) {
```
Zeyber: Needs a new flag name

projects/storefrontlib/shared/components/split-view/view/view.component.ts
```diff
@@ -73,6 +83,15 @@ export class ViewComponent implements OnInit, OnDestroy {
     protected cd: ChangeDetectorRef
   ) {}
 
+  @Optional()
```
Zeyber, line 79: Lets keep our injectors at the top of the class to be consistent.

Zeyber (changes requested):

projects/storefrontlib/shared/components/split-view/view/view.component.ts
Pio-Bar: name updated

projects/storefrontlib/shared/components/split-view/view/view.component.ts
Pio-Bar: Changed to a new flag

projects/storefrontlib/shared/components/split-view/view/view.component.ts
Pio-Bar: I'll keep this in mind ;)

projects/storefrontlib/shared/components/split-view/view/view.component.ts
```diff
@@ -65,6 +72,19 @@ export class ViewComponent implements OnInit, OnDestroy {
   @Output()
   hiddenChange = new EventEmitter();
 
+  @Input()
+  viewTitle?: string;
+
```
Zeyber, line 77: Lets add TODOs here for cleanup. ie. `globalMessageService` is optional now but will be required next major. `featureConfigService` will be removed.

Zeyber (changes requested):

projects/storefrontlib/shared/components/split-view/view/view.component.ts
Pio-Bar: done

Zeyber (dismissed):

developpeurweb (dismissed):
Tested with JAWS, all instances work as expected.
Also tested with VO and narration gets cut, we'll address VO narration as an improvement in the future.

Zeyber (approved):

PR #18645, opened 2024-03-18
test: Spike e2e testing with Access Continuum

Zeyber:
Closes: https://jira.tools.sap/browse/CXSPA-6523

This PR investigates if it is possible and how much would be involved with setting up Access Continuum to continuously perform a11y checks on our e2e tests to avoid regressions. It leverages the [example repo](https://github.tools.sap/I576769/cypress-continuum-tools/tree/LatestCypress) to set up continuum on cypress and can be used as an example for how we can include these checks in spartacus.

Note: API keys and library files have been omitted.

### DIFF
--- a/projects/storefrontapp-e2e-cypress/continuum/continuum.conf.js
+++ b/projects/storefrontapp-e2e-cypress/continuum/continuum.conf.js
@@ -1,0 +1,31 @@
+window.LevelAccess_AccessContinuumConfiguration = {
+  accessEngineType: 'professional',
+  ampInstanceUrl: 'https://sap.levelaccess.net',
+  defaultStandardIds: [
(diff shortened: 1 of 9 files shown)


Zeyber:
**Summary of findings:**

- It is very possible and rather easy to do inline e2e checks with cypress to verify a11y in any given state of the app (eg. starting home page, adding item to cart, etc).

- The Access Continuum library is not public. This means we need a key in order to download the library on fresh installs. This key should not be publicly stored in our repos so we would need to store secrets in github (requires admin) and configure in the actions somehow.

- The results can be shared with AMP. This also requires a key with similar limitations as above.

- We would need to decide on how to address testing to avoid regression (eg. checking a11y for pages or specific app tests, creating new tests or modifying existing, test for new issues only or expand coverage for previous fixes, etc).

PR #18646, opened 2024-03-19
feat: update generating OTP token endpoint call

Matejk00:
Closes: CXSPA-6030


FollowTheFlo (approved):
good to me, thx for adding spec file.

PR #18648, opened 2024-03-19
fix: (CXSPA-2338) - Reorder dialog a11y fixes

Pio-Bar:
[CXSPA-2338](https://jira.tools.sap/browse/CXSPA-2338)

Enables screen reader readout when the modal opens and after its contents change. Additionally it should not lose focus after performing the action.


feature-libs/order/components/order-details/order-detail-reorder/reorder-dialog/reorder-dialog.component.ts
```diff
@@ -40,6 +47,15 @@ export class ReorderDialogComponent {
   showDecisionPrompt$ = new BehaviorSubject(true);
   data$ = this.launchDialogService.data$;
 
+  @Optional() selectFocusUtility = inject(SelectFocusUtility, {
```
Zeyber, line 51: It is good that `selectFocusUtility` and `elementRef` are optional injectors now. But we should include that `TODO` cleanup comment to also make them non-optional injections when we remove the `featureConfigService`.

Zeyber (changes requested):
We should check that the focus is captured in our unit tests too for code coverage.

feature-libs/order/components/order-details/order-detail-reorder/reorder-dialog/reorder-dialog.component.ts
Pio-Bar: Added comments and a unit test to ensure the focus is set.

Zeyber (dismissed):

developpeurweb (dismissed):

Zeyber (approved):

PR #18650, opened 2024-03-19
feat: new `BreakingChangesFlags` config to drive parts of Spartacus config, e.g. cms mapping

Platonn:
The new `BreakingChangesFlags` config is separate from Spartacus config, to be able to control the global Spartacus config. 
The `BreakingChangesFlags` can be a "lightweight serializable wrapper config" which allows for toggling on complex non-serializable breaking changes in the global Spartacus config.
See example:
```ts
provideDefaultBreakingChangesFlags({
  // boolean is a simple, serializable flag, so can be fetched from remote configuration endpoint in runtime
  myAccountV2: true
})
```
```ts
provideConfigFactory(()=>{
  const breakingChangesFlags = inject(BreakingChangesFlags);

  // we use lightweight serializable flag BreakingChangesFlags to induce changes in the cmsMapping
  return breakingChangesFlags.myAccountV2
    ? { 
      cmsComponents: { 
        // CMS mapping contains JS class, so it's a complex non-serializable config,
        // which we could not fetch from remote configuration endpoint in runtime
        component: MyAccountComponentV2
      } 
    } : { 
      cmsComponents: { 
        component: MyAccountComponent
      }
    }
})
```

In other words, the `BreakingChangesFlags` config can be a bridge between a remote configuration of Spratacus (which allows only pure serializable JSON) and making actual changes into the non-serializable global Spartacus config parts like cms mapping with references to components classes.

closes https://jira.tools.sap/browse/CXSPA-6586


feature-libs/order/components/order-details/order-detail-items/consignment-tracking/augmented-features-config.model.ts
```diff
@@ -0,0 +1,16 @@
...
+
+import '@spartacus/core';
+
+declare module '@spartacus/core' {
+  interface FeaturesConfig {
+    /**
+     * Enables displaying consignment tracking info in the order details view
+     */
+    consignmentTracking?: boolean;
+  }
```
Platonn, line 15: Note: It's a real feature toggle (not a breaking changes flag), which is enabled by default since v1.2. To be re-discussed with PO (Steve), whether we should deprecate and eventually remove this toggle.

Platonn:
closing in favor of https://github.com/SAP/spartacus/pull/18670

PR #18655, opened 2024-03-21
fix: GooglePay pickup transaction from PDP page

Matejk00:
Closes: CXSPA-6582

### DIFF
--- a/integration-libs/opf/base/core/services/opf-pickup-in-store-handler.service.ts
+++ b/integration-libs/opf/base/core/services/opf-pickup-in-store-handler.service.ts
@@ -6,7 +6,7 @@
 
 import { Injectable, inject } from '@angular/core';
 import { ActiveCartFacade } from '@spartacus/cart/base/root';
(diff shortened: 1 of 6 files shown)


integration-libs/opf/base/core/services/opf-pickup-in-store-handler.service.ts
```diff
@@ -29,30 +29,37 @@ export class OpfPickupInStoreHandlerService {
...
-                  }
-                : shippingTypeInfo;
-            }),
-            take(1)
-          )
-      )
+            map((intendedLocation) => {
+              if (
+                intendedLocation &&
+                intendedLocation.pickupOption ===
```
Matejk00, line 50: We need this verification:

```
intendedLocation.pickupOption === OpfQuickBuyDeliveryType.PICKUP.toLowerCase()
```

to recognize if pickup option is already selected on PDP page. Otherwise the service returns wrong delivery type.

FollowTheFlo (approved):

PR #18658, opened 2024-03-21
fix payment still on progress on next payment

FollowTheFlo:
issue: after a successful payment with pickup, a 2nd payment with or without pickup was not possible as error 'payment still in progress' was thrown.
reason: finalize rxjs operator was not reached on 2nd call. adding take(1) fixes the prob.
additionally, added conditions to not set unneeded applePay event listeners when pickup mode is selected.


integration-libs/opf/base/components/opf-quick-buy/apple-pay/observable/apple-pay-observable.factory.ts
```diff
@@ -72,7 +76,10 @@ export class ApplePayObservableFactory {
         });
       }
 
-      if (config.shippingMethodSelected) {
+      if (
+        config.shippingMethodSelected &&
+        config.request.shippingType !== ApplePayShippingType.STORE_PICKUP
+      ) {
```
Matejk00, line 103: We duplicate this code two times, maybe is it worth to move it into some function that do such check?

Matejk00 (approved):

PR #18667, opened 2024-03-25
feat: Keyword redirect support in Spartacus

lasoh:


projects/core/src/occ/adapters/product/occ-product-search.adapter.ts
```diff
@@ -22,12 +22,14 @@ import { ConverterService } from '../../../util/converter.service';
...
 import { OCC_HTTP_TOKEN } from '../../utils';
+import { Router } from '@angular/router';
 @Injectable()
 export class OccProductSearchAdapter implements ProductSearchAdapter {
   constructor(
     protected http: HttpClient,
     protected occEndpoints: OccEndpointsService,
-    protected converter: ConverterService
+    protected converter: ConverterService,
+    protected router: Router
```
rmch91: Breaking Change 🧨
Please consider usage of inject() with optional flag

projects/core/src/occ/adapters/product/occ-product-search.adapter.ts
```diff
@@ -44,7 +46,14 @@ export class OccProductSearchAdapter implements ProductSearchAdapter {
 
     return this.http
       .get(this.getSearchEndpoint(query, searchConfig), { context })
-      .pipe(this.converter.pipeable(PRODUCT_SEARCH_PAGE_NORMALIZER));
+      .pipe(
+        this.converter.pipeable(PRODUCT_SEARCH_PAGE_NORMALIZER),
+        tap(
+          (productSearchPage) =>
+            productSearchPage.keywordRedirectUrl &&
+            this.router.navigate([productSearchPage.keywordRedirectUrl])
```
rmch91: After change to inject, add `?` to `this.router?.navigate`, please.

rmch91 (changes requested):

projects/core/src/occ/adapters/product/occ-product-search.adapter.ts
```diff
@@ -22,12 +22,14 @@ import { ConverterService } from '../../../util/converter.service';
...
 import { OCC_HTTP_TOKEN } from '../../utils';
+import { Router } from '@angular/router';
 @Injectable()
 export class OccProductSearchAdapter implements ProductSearchAdapter {
   constructor(
     protected http: HttpClient,
     protected occEndpoints: OccEndpointsService,
-    protected converter: ConverterService
+    protected converter: ConverterService,
+    @Optional() protected router: Router
```
rmch91: Any change of constructor is considered as a breaking change. For the changes in constructor, to avoid breaking changes, we are using constructor overloading. Explained [here](https://sap.github.io/spartacus-docs/breaking-changes/#adding-new-constructor-dependencies-in-minor-versions).

But there is no need in this case to add router to constructor.
You can just declare service as class property using `inject()`

```
protected router = inject(Router, {
    optional: true,
});
```

rmch91 (changes requested):

projects/core/src/occ/adapters/product/occ-product-search.adapter.ts
lasoh: Ok, sure, i misunderstand ;(

rmch91 (approved):

PR #18670, opened 2024-03-25
feat: new `FeatureToggles` config to drive parts of Spartacus config, e.g. cms mapping

Platonn:
The new `FeatureToggles` config is separate from Spartacus config, to be able to control the global Spartacus config. 
The `FeatureToggles` is a simple list of boolean properties which allows for enabling breaking changes in Spartacus, ALSO changes in the default config chunks like cms mapping etc. See example:
```ts
provideDefaultFeatureToggles({
  // boolean is a simple, serializable flag, so can be even potentially
  // fetched from remote configuration endpoint in runtime
  myAccountV2: true
})
```
```ts
// conditionally changing the default config chunk based on the simple boolean feature toggle:
provideConfigFactory(()=>{
  const featureToggles = inject(FeatureToggles);
  return featureToggles.myAccountV2
    ? { cmsComponents: { component: MyAccountComponentV2 } }
    : { cmsComponents: { component: MyAccountComponent } };
})
```

In other words, the `FeatureToggles` config can be a bridge between a simple (and even remote!) feature toggles configuration of Spartacu and making actual changes into the default config chunks in Spartacus (like default cms mapping).

Note: In this PR added also an ESLint rule to prefer using in our libraries the function `provideDefaultFeatureToggles()` over `provideFeatureToggles()` - so our providers have lower precedence than customers' ones.

**ALL FEATURE TOGGLES DEFINED IN THE CORE LIB:**

All new toggles are now added in the file `feature-toggles.ts` for customers and in `spartacus-features.module.ts` for our storefront app.

The centralization of feature toggles was done because:
- for doc writers: otherwise It was hard to track down new toggles each release due to distribution across many libs
- for devs: adding a new toggle lacked a standard process, raising the question: Where to define it? This dilemma leads to issues, **especially when default toggles values were provided by mistake in lazy-loaded modules it was not working** (because the feature toggles mechanism relies on the **root injector**).
- this confusion wasted developers' time, as they "just want to add a toggle" but get stuck figuring out where to place it, with the process varying across different libraries (storefront, core, feature).

**DEPRECATIONS:**

This PR deprecates the `FeaturesConfig` interface in favor of `FeatureToggles`: So please use:
```ts
provideFeatureToggles({
  [string]: boolean
})
```
instead of
```ts
provideConfig({
  features: {
    [string]: boolean
  }
})
```
That said, the old approach will still work until we remove the `FeaturesConfig` eventually.

**QA steps**

I) verify that FeatureToggles mechanism work as expected on the example property `a11yRequiredAsterisks`:
1. open login page
2. verify that the asterisk regarding required form fields IS displayed
3. in `spartacus-features.module.ts` REMOVE the toggle `a11yRequiredAsterisks` and verify asterisk is NOT displayed
4. in `default-feature-toggles.ts` set `a11yRequiredAsterisks: true` and verify asterisk IS displayed
5. in `spartacus-features.module.ts` set `a11yRequiredAsterisks: false` and verify asterisk is NOT displayed
6. in `spartacus-features.module.ts` set `a11yRequiredAsterisks: true` and verify asterisk is IS displayed
7. in `default-feature-toggles.ts` set `a11yRequiredAsterisks: false` and verify asterisk IS STILL displayed
8. in `spartacus-features.module.ts` REMOVE the toggle `a11yRequiredAsterisks` and verify asterisk is NOT displayed

II) verify FeatureToggles mechanism can alter a cms mapping, let's use `a11yRequiredAsterisks` for the spike purposes:
1. in `login-form.module.ts` add to the `providers` array the following code:
```ts
    provideDefaultConfigFactory(() => {
      return inject(FeatureToggles).a11yRequiredAsterisks
        ? {
            cmsComponents: {
              ReturningCustomerLoginComponent: {
                component: SpikeComponent,
              },
            },
          }
        : {};
    }),
```
2. ...and in the top of the same file (`login-form.module.ts`) add the following code:
```ts
@Component({
  selector: 'cx-spike',
  template: 'SPIKE COMPONENT WITH ASTERISK (*.*)',
})
export class SpikeComponent {}
```

3. Visit the login page and verify that the `SPIKE` text is displayed
4. in `spartacus-features.module.ts` REMOVE the toggle `a11yRequiredAsterisks` and verify that the `SPIKE COMPONENT WITH ASTERISK (*.*)` text is NOT displayed

III) verify the old FeaturesConfig mechanism cannot alter cms mapping, but still works for enabling features:
1. ensure that `a11yRequiredAsterisks` in `provideFeatureToggles` is set to `false`
2. in `spartacus-features.module.ts` add to the `providers` array the following code `provideConfig({features: { a11yRequiredAsterisks: true  }})` and verify that the ORIGINAL LOGIN COMPONENT is displayed WITH real a11y asterisk feature displayed
3. in `spartacus-features.module.ts` set `a11yRequiredAsterisks: false` inside the above mentioned `provideConfig()` and verify that the ORIGINAL LOGIN COMPONENT is displayed WITHOUT real a11y asterisk feature displayed
4. in `spartacus-features.module.ts` set `a11yRequiredAsterisks: true` inside the `provideFeatureToggles()` and verify that the the `SPIKE` text is displayed even if `a11yRequiredAsterisks: false` inside the `provideConfig()` is false

closes https://jira.tools.sap/browse/CXSPA-6586


projects/core/src/config/services/augmented-feature-toggles.model.ts
```diff
@@ -4,34 +4,27 @@
...
-     */
-    level?: string;
-    consignmentTracking?: boolean;
+import '../../features-config/config/features-config';
+
+// NOTE: It's EXCEPTIONALLY a part of `FeaturesConfigContent` but not of `FeatureToggles`,
+// because it's a permanent feature enabler, but not a short-living feature toggle
+// covering a breaking change.
+//
+// It's deliberately is disabled by default.
```
Platonn:
```suggestion
// It's disabled by default.
```

projects/core/src/features-config/config/features-config.ts
```diff
@@ -12,10 +12,19 @@ import { Config } from '../../config/config-tokens';
...
-    level?: string;
-    [featureToggle: string]: string | boolean | undefined;
-  };
+  features?: FeaturesConfigContent;
+}
+
+export interface FeaturesConfigContent {
+  /**
+   * Configure feature level.
+   * Value corresponds to minor (feature) release version number: '1.0', '1.1', etc.
```
Platonn:
```suggestion
   * Value corresponds to a specific release version number: '1.0', '1.1', etc.
```

projects/core/src/features-config/feature-toggles/feature-toggles-providers.ts
```diff
@@ -0,0 +1,82 @@
...
+import {
+  DefaultFeatureTogglesChunk,
+  FeatureToggles,
+  FeatureTogglesChunk,
+} from './feature-toggles-tokens';
+
+export type FeatureTogglesFactory = (...props: any[]) => FeatureToggles;
+
+/**
+ * Helper function to provide flags chunk using FlagsChunk token
```
Platonn, line 17:
```suggestion
 * Helper function to provide FlagsChunk token
```

projects/core/src/features-config/feature-toggles/feature-toggles-tokens.ts
```diff
@@ -0,0 +1,76 @@
+/*
+ * SPDX-FileCopyrightText: 2024 SAP Spartacus team <spartacus-team@sap.com>
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { Injectable, InjectionToken, inject } from '@angular/core';
+
+/**
+ * Properties for enabling/disabling breaking changes gradually introduced in Spartacus.
```
Platonn:
```suggestion
 * Properties for gradually enabling/disabling breaking changes introduced in Spartacus.
```

feature-libs/quote/components/quote-components.module.ts
```diff
@@ -37,13 +37,6 @@ import { QuoteSummarySellerEditModule } from './summary/seller-edit/quote-summar
     QuoteSummaryPricesModule,
     QuoteSummarySellerEditModule,
   ],
-  providers: [
-    provideDefaultConfig(defaultQuoteUIConfig),
-    provideDefaultConfig(<FeaturesConfig>{
-      features: {
-        storeFrontLibCardParagraphTruncated: true,
-      },
-    }),
```
Platonn: I believe this flag was enabled by default in v20 by mistake, so reverting it.

projects/core/src/features-config/config/features-config.ts
```diff
@@ -12,10 +12,19 @@ import { Config } from '../../config/config-tokens';
...
 })
 export abstract class FeaturesConfig {
-  features?: {
-    level?: string;
-    [featureToggle: string]: string | boolean | undefined;
-  };
+  features?: FeaturesConfigContent;
+}
+
+export interface FeaturesConfigContent {
```
pawelfras: Is this interface related only to feature togglers? If yes, what do you think about `FeatureTogglesConfig`?

projects/core/src/features-config/feature-toggles/feature-toggles-providers.ts
pawelfras, line 1: `query:` What do you think about creating unit tests for these providers? There are some conditions which may be worth testing.

projects/core/src/features-config/feature-toggles/feature-toggles-tokens.ts
pawelfras, line 1: query: `DefaultFeatureToggles` and `RootFeatureToggles` use factories that may be worth testing. What do you think about this?

projects/core/src/features-config/feature-toggles/feature-toggles-tokens.ts
```diff
@@ -0,0 +1,76 @@
...
+
+import { Injectable, InjectionToken, inject } from '@angular/core';
+
+/**
+ * Properties for gradually enabling/disabling breaking changes introduced in Spartacus.
+ *
+ * Feature toggles facilitate the incremental evolution of the Spartacus product,
+ * allowing Spartacus users to adjust to breaking changes over time.
+ *
+ * Each feature toggle corresponds to a specific breaking change introduced
```
pawelfras: `query:` Wasn't the `FeatureToggles` mechanism supposed to be responsible for enabling/disabling features in general? I thought the mechanism, which is already in the code, would be responsible for breaking changes. I would appreciate being straightened out if I am wrong :)

feature-libs/quote/components/quote-components.module.ts
Platonn: undoing this, becasue in develop its already reverted now

projects/core/src/features-config/feature-toggles/feature-toggles-providers.ts
```diff
@@ -0,0 +1,82 @@
...
+    multi: true,
+  };
+}
+
+/**
+ * Helper function to provide DefaultFeatureTogglesChunk token
+ *
+ * @param flags FeatureToggles object to merge with the default flags
+ */
+export function provideDefaultFeatureToggles(
```
Platonn, line 61: TODO: Let's add an ESLint rule to forbid usage of provide (non-default) Feature Toggles inside our libs, analogically to our rules for `provideConfig` vs `provideDefaultConfig`

projects/core/src/features-config/feature-toggles/feature-toggles-tokens.ts
Platonn: done

projects/core/src/features-config/feature-toggles/feature-toggles-providers.ts
Platonn: done

projects/core/src/features-config/feature-toggles/feature-toggles-providers.ts
Platonn: done

projects/core/src/features-config/config/features-config.ts
Platonn: As discussed offline, this interface was only confusing.
Removed it. And deprecated the whole `FeaturesConfig` in favor of `FeatureToggles`

projects/core/src/features-config/feature-toggles/feature-toggles-tokens.ts
Platonn: Updated text. What do you think?
```
/**
 * Properties for gradually enabling/disabling breaking changes introduced in Spartacus.
 *
 * Feature toggles facilitate the incremental evolution of the Spartacus product,
 * allowing Spartacus users to adjust to breaking changes over time.
 *
 * Each feature toggle corresponds to a specific breaking change introduced
 * in a Spartacus release.
 *
 * Some feature toggles might be only temporary, with a defined duration
 * (specified in the official documentation), after which the related
 * breaking change will be permanently enabled.
 */
 ```

Comment (Platonn) [reply projects/core/src/features-config/feature-toggles/feature-toggles-tokens.ts]: Updated text again and shortened, to be more reader friendly
```
/**
 * Properties for gradual adaptation to breaking changes in Spartacus via feature toggles.
 * Each toggle has its own lifespan defined in the official docs.
 * It should be provided in the root injector (in an eagerly loaded module).
 */
```

File: projects/storefrontlib/feature-toggles.ts
Comment (Platonn) [inline projects/storefrontlib/feature-toggles.ts:1]: Note: It's not a breaking change to remove this file. It was not used anywhere nor exported in public API.

See other complementary files:
- FeaturesConfig (deprecated)
- FeatureToggles

File: projects/core/src/features-config/feature-toggles/feature-toggles-providers.ts [hunk 9]
```diff
@@ -0,0 +1,88 @@
...
+ * For using in a library, call `provideDefaultFeatureToggles() instead.
+ *
+ * @param featureTogglesFactory Factory Function that returns FeatureToggles object
+ */
+export function provideFeatureTogglesFactory(
+  featureTogglesFactory: FeatureTogglesFactory,
+  defaultToggles = false
+): FactoryProvider {
+  return {
+    provide: defaultToggles ? DefaultFeatureTogglesChunk : FeatureTogglesChunk,
```
pawelfras: query: Do we need a possibility for adding default toggles via this util function if we have a dedicated `provideDefaultFeatureTogglesFactory` function for this? If yes, I have two proposals:

- let's add a test case for this condition
- reuse the code in `provideDefaultFeatureTogglesFactory`:
```ts
export function provideDefaultFeatureTogglesFactory(
  featureTogglesFactory: FeatureTogglesFactory
): FactoryProvider {
  return provideFeatureTogglesFactory(featureTogglesFactory, true);
}
```

projects/core/src/features-config/feature-toggles/feature-toggles-providers.ts
```diff
@@ -0,0 +1,88 @@
...
+ * For using in a library, call `provideDefaultFeatureToggles() instead.
+ *
+ * @param toggles FeatureToggles object
+ */
+export function provideFeatureToggles(
+  toggles: FeatureToggles = {},
+  defaultToggles = false
+): ValueProvider {
+  return {
+    provide: defaultToggles ? DefaultFeatureTogglesChunk : FeatureTogglesChunk,
```
pawelfras: query: Do we need a possibility for adding default toggles via this util function if we have a dedicated `provideDefaultFeatureToggles` function for this? If yes, I have two proposals:

- let's add a test case for this condition
- reuse the code in `provideDefaultFeatureToggles`:
```ts
export function provideDefaultFeatureToggles(
  toggles: FeatureToggles = {}
): ValueProvider {
  return provideFeatureToggles(toggles, true);
}
```

projects/core/src/features-config/feature-toggles/feature-toggles-tokens.ts
```diff
@@ -0,0 +1,77 @@
...
+    Object.assign(
+      {},
+      inject(DefaultFeatureToggles),
+      inject(RootFeatureToggles)
+    ),
+})
+export abstract class FeatureToggles {}
+
+export interface FeatureToggles {
+  [featureToggleName: string]: boolean | undefined;
```
pawelfras: TS is complaining 🤔 
image

Platonn: fixed

projects/core/src/features-config/feature-toggles/feature-toggles-providers.ts
Platonn: good catch. it's useless parameter. now I removed it
It was influenced by the copy-pasted implementation of provideConfig() utils, which has this useless parameter

projects/core/src/features-config/feature-toggles/feature-toggles-providers.ts
Platonn: removed. see the same explanation above

projects/core/src/features-config/features-config.module.spec.ts
```diff
@@ -0,0 +1,110 @@
...
+        provideDefaultConfig({ features: { [testToggle]: false } }),
+      ]);
+      const service = TestBed.inject(FeatureConfigService);
+      expect(service.isEnabled(testToggle)).toBe(false);
+    });
+
+    // The following 2 tests confirm a correct behavior, which is might be not intuitive.
+    //
+    // When mixing `FeaturesConfig` and `FeatureToggles` in the same level (both ROOT, or both DEFAULT)
+    // then `FeaturesConfig` takes precedence over `FeatureToggles`, regadless of the order of providers.
```
pawelfras:
```suggestion
    // then `FeaturesConfig` takes precedence over `FeatureToggles`, regardless of the order of providers.
```

projects/core/src/features-config/features-config.module.spec.ts
```diff
@@ -0,0 +1,110 @@
...
+    it('DEFAULT FeatureToggles should be overridden by DEFAULT FeaturesConfig', () => {
+      setProviders([
+        provideDefaultFeatureToggles({ [testToggle]: true } as FeatureToggles),
+        provideDefaultConfig({ features: { [testToggle]: false } }),
+      ]);
+      const service = TestBed.inject(FeatureConfigService);
+      expect(service.isEnabled(testToggle)).toBe(false);
+    });
+
+    // The following 2 tests confirm a correct behavior, which is might be not intuitive.
```
pawelfras:
```suggestion
    // The following 2 tests confirm a correct behavior, which might not be intuitive.
```

pawelfras (dismissed):
QA done. LGTM 👌

projects/storefrontapp/src/app/spartacus/features/quote-feature.module.ts
```diff
@@ -54,11 +54,6 @@ import {
         },
       },
     }),
-    provideConfig(<FeaturesConfig>{
-      features: {
-        storeFrontLibCardParagraphTruncated: true,
```
pawelfras, line 59: the removed flag has a value `true` while the default value of `storeFrontLibCardParagraphTruncated` inside `default-feature-toggles.ts` is `true`. Is this intentional?

Platonn: Oups. My bad. Good catch.

Platonn: Oh, actually everything is fine. Let me clarify:
1. In default-feature-toggles.ts as of today ALL flags should be false (because 6months didn't pass since 2211.20 yet and only then we introduced first toggles)
2. In storefrontapp _some_ toggles have been already enabled for the testing purposes - e.g. QuoteFeatureModule which is a part of the storefront app. Please note that this flag is _still_ enabled in the storefrontapp, but just in a different place - in `SpartacusFeaturesModule`:
image

pawelfras (dismissed):

pawelfras (dismissed):

pawelfras (approved):

PR #18683, opened 2024-03-29
feat: CXSPA-6576 Introduce SERVER_ERROR_RESPONSE_FACTORY

pawelfras:
This PR introduces the SERVER_ERROR_RESPONSE_FACTORY strategy to `ServerRespondingErrorHandler` to generate proper CxServerError errors depending on the incoming HTTP error response during SSR. Two default factories have been introduced together with this PR :
- CmsPageNotFoundServerErrorResponseFactory - responsible for generating errors when the CMS page is not found
- UnknownServerErrorResponseFactory- fallback, generates unknown server error for everything else

QA steps:

  - modify `pages` property in `default-cms-config.ts` to generate `CmsPageNotFoundServerError`
  - modify `components` property in `default-cms-config.ts` to generate `UnknownServerError`
- add `console.log` to `handleError` method inside core-libs/setup/ssr/error-handling/multi-error-handlers/server-responding-error-handler.ts
- run npm `run dev:ssr`
- verify whether the appropriate error is visible in the terminal where server runs

closes [CXSPA-6575](https://jira.tools.sap/browse/CXSPA-6576)


core-libs/setup/ssr/error-handling/server-error-response-transformers/cms-page-not-found-server-error-response-transformer.ts
```diff
@@ -0,0 +1,44 @@
...
+  transform(error: any): CxServerError {
+    const message = 'CMS page not found';
+    return new CmsPageNotFoundServerError({
+      message,
+      originalError: error,
+    });
+  }
+
+  protected isHttpErrorResponse(error: any): error is HttpErrorResponse {
+    return 'headers' in error;
```
Platonn: Did you consider an alternative solution: `return error instanceof HttpErrorResponse` ?

core-libs/setup/ssr/error-handling/server-error-response-transformers/cms-page-not-found-server-error-response-transformer.ts
```diff
@@ -0,0 +1,44 @@
...
+ */
+@Injectable({
+  providedIn: 'root',
+})
+export class CmsPageNotFoundServerErrorResponseTransformer
+  implements ServerErrorResponseTransformer
+{
+  hasMatch(error: unknown): boolean {
+    return (
+      this.isHttpErrorResponse(error) && !!error.url?.includes('cms/pages')
```
Platonn: Now any endpoint that contains `cms/pages` is taken into account, which is prone to mistakes, e.g. `/some/custom/api/cms/pagesStatus` (I made up the example)
What about comparing with precise fully built CMS semantic 'pages' endpoint?
```ts
const pagesEndpoint = this.endpointService.buildUrl('pages');
return ...
  ... (error.url ?? '').startsWith(pagesEndpoint)
```
?

reminder: If we use this solution, then JSDocs of class needs updating

core-libs/setup/ssr/error-handling/server-error-response-transformers/cms-page-not-found-server-error-response-transformer.ts
```diff
@@ -0,0 +1,44 @@
...
+    );
+  }
+
+  getPriority(): number {
+    return Priority.NORMAL;
```
Platonn: Let's use `LOW` by default in our library, so customers' ones have higher priority by default (`NORMAL`)

core-libs/setup/ssr/error-handling/server-error-response-transformers/server-error-response-transformers.ts
```diff
@@ -0,0 +1,27 @@
...
+import { InjectionToken } from '@angular/core';
+import { Applicable } from '@spartacus/core';
+import { CxServerError } from '../server-errors';
+
+/**
+ * A transformer responsible for transforming an error into a {@link CxServerError}.
+ * The transformer is applicable when the error matches the transformer's criteria.
+ * Spartacus provides two default transformers:
+ * - {@link CmsPageNotFoundServerErrorResponseTransformer}
+ * - {@link UnknownServerErrorResponseTransformer}
```
Platonn: Do you think we should really put it in the JSDocs of the generic interface?

Maybe rather in the official docs

core-libs/setup/ssr/error-handling/server-errors/cx-server-error.ts
```diff
@@ -0,0 +1,45 @@
...
+ */
+
+import { CxError } from './cx-errors';
+
+/**
+ * Data that can be attached to a {@link CxServerError}.
+ */
+export interface CxServerErrorData {
+  message?: string;
+  originalError?: unknown;
```
Platonn: Instead of the custom property, let's use the standard `cause` property of the native `Error` object.
see https://nodejs.org/docs/latest-v18.x/api/errors.html#new-errormessage-options

core-libs/setup/ssr/error-handling/server-error-response-transformers/cms-page-not-found-server-error-response-transformer.ts
pawelfras: Good point. I had a gut feeling the check I've proposed may be error-prone. Thank you for the suggestion.

core-libs/setup/ssr/error-handling/server-errors/cx-errors.ts
```diff
@@ -0,0 +1,16 @@
...
+/*
+ *  Error codes used to denote a particular type of error appearing in the application during SSR.
+ *  We provides two default error codes:
+ *  - UnknownServerError SR0001: Used when the error is not recognized by the application.
+ *  - CmsPageNotFoundServerError SR0002: Used when the requested CMS page is not found.
+ * */
+export enum CxError {
+  UnknownServerError = 'SR0001',
+  CmsPageNotFoundServerError = 'SR0002',
+}
```
Platonn: 1. Do we really, really need this enum? if yes, why?

2. If yes, why do we need misterious codes, instead of self-explanatory strings, e.g. 
```ts
export enum CxError {
  UnknownServerError = 'UnknownServerError',
  CmsPageNotFoundServerError = 'CmsPageNotFoundServerError',
}
```


3. `CxError`  sounds very generic. If it should be used everywhere in Spartacus libs, I guess it shouldn't live in `@spartacus/setup/ssr`? What's your long-term vision with this interface? Why you introduced it? Should this approach be spread across all libraries, like Angular does? AFAIK Angular does it also for purposes of treeshaking long error descriptions in production builds of an app - they use Angular's internal isDevMode global variable + provide official docs for each error code. I'm not sure we should dive now into copying an analogical approach from them?

core-libs/setup/ssr/error-handling/server-error-response-transformers/server-error-response-transformers.ts
```diff
@@ -0,0 +1,27 @@
...
+export interface ServerErrorResponseTransformer extends Applicable {
+  transform(error: unknown): CxServerError;
+}
+
+/**
+ * Injection token for server error response transformers.
+ */
+export const SERVER_ERROR_RESPONSE_TRANSFORMERS = new InjectionToken<
+  ServerErrorResponseTransformer[]
+>('SERVER_ERROR_RESPONSE_TRANSFORMERS');
```
Platonn: We proposed `Transformer` from top of head, but knew it's not the best name for it.
Reminder to decide on naming of it.

Some ideas (not exhaustive of course)::
1. ServerErrorResponseTransformer .transform(unknown): CxServerError
1. ServerErrorResponseMapper .map(unknown): CxServerError
1. ServerErrorResponseStrategy .handle(unknown): CxServerError
1. ServerErrorResponseHandler .handle(unknown): CxServerError
1. ... ?

Btw. there is an inconsistency. Sometimes we say ServerErrorResponse, and other times ServerError, which confused me a bit. I guess it wasn't deliberate? I guess we wanted to return `CxServerError**Response**` (e.g. UknownServerError**Response**, CmsPageNotFoundServerError**Response**)?

Anyways, what name would suggest or vote for?

core-libs/setup/ssr/error-handling/server-error-response-transformers/cms-page-not-found-server-error-response-transformer.ts
pawelfras: Well... missed that one 🤦

core-libs/setup/ssr/error-handling/server-errors/cx-errors.ts
pawelfras: I thought enum would be a good solution in terms of extensibility for possible codes in the future. I don't insist on them being enigmatic codes, but I figured it might be worth doing it the "angular way" right now. I agree that the name is too generic at this point. I propose changing it to `CxServerError` or `CxSsrError` - the name can be later adjusted if in the future we see a purpose to spread these codes across the whole project.

core-libs/setup/ssr/error-handling/server-error-response-transformers/server-error-response-transformers.ts
pawelfras: Indeed, we had some doubts about the `Transformer` name. Thank you for your proposals!
- Ad. 1 & 2 - are with me on the same level and I would choose one of these names. I propose to stay with `Transformer` for the time being.
- Ad. 3 - In my opinion, `Strategy` fits more with the abstraction of the level above, which we called the `MultiErrorHandler` (but as far as I remember we talked about `Strategy` when discussing this abstraction
- Ad. 4 I'd omit this name as I have a feeling we have too much `Handler` usage in the whole flow and it may be misleading

Regarding inconsistency - I hear you and like the idea of having `CxServerErrorResponse` as the abstract class 👌

pawelfras: As discussed, this abstraction layer is, in fact, the factory pattern. 
I'll change `SERVER_ERROR_RESPONSE_TRANSFORMERS` into `SERVER_ERROR_RESPONSE_FACTORY` and adjust the related code.

core-libs/setup/ssr/error-handling/server-error-response-factory/cms-page-not-found-server-error-response-factory.ts
```diff
@@ -0,0 +1,81 @@
...
+  protected buildUrlRegex({
+    baseUrl,
+    prefix,
+    endpoint,
+  }: Record<string, string | undefined>): RegExp {
+    const escapedBaseUrl = this.escapeRegExp(baseUrl ?? '');
+    const escapedPrefix = this.escapeRegExp(prefix ?? '');
+    const escapedEndpoint = this.escapeRegExp(endpoint ?? '');
+    return new RegExp(
+      `^${escapedBaseUrl}${escapedPrefix}[^/]+/${escapedEndpoint}`
```
Platonn: It's not obvious why we have a wildcard URL segment here.

Let's make it explicit, e.g. add a code comment why we add it here
Perhaps you might want to also extract `[^/]+/` to a separate variable, e.g. `baseSiteWildcardUrlSegment`

core-libs/setup/ssr/error-handling/server-error-response-factory/server-error-response-factory.ts
```diff
@@ -0,0 +1,24 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { InjectionToken } from '@angular/core';
+import { Applicable } from '@spartacus/core';
+import { CxServerErrorResponse } from '../server-errors';
+
+/**
+ * A factory responsible for creating a {@link CxServerErrorResponse} basedon HTTP error response during SSR.
+ * The fatory is applicable when the error matches the factory's criteria.
```
Platonn, line 13: Perhaps I'm missing something... I'm not sure if this sentence carries any information for me. Perhaps can be dropped?

core-libs/setup/ssr/error-handling/server-error-response-factory/server-error-response-factory.ts
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { InjectionToken } from '@angular/core';
+import { Applicable } from '@spartacus/core';
+import { CxServerErrorResponse } from '../server-errors';
+
+/**
+ * A factory responsible for creating a {@link CxServerErrorResponse} basedon HTTP error response during SSR.
```
Platonn:
```suggestion
 * A factory responsible for creating a {@link CxServerErrorResponse} based on a HTTP error response during SSR.
```

core-libs/setup/ssr/error-handling/server-error-response-factory/server-error-response-factory.ts
Platonn:
```suggestion
 *
 * The factory implements the {@link Applicable} interface. So when handing an original error, only one of the provided server response factories will be used - the one that is the best applicable for the handled error.
```

core-libs/setup/ssr/error-handling/server-error-response-factory/unknown-server-error-response-factory.spec.ts
```diff
@@ -0,0 +1,31 @@
+import { Priority } from '@spartacus/core';
+import { UnknownServerErrorResponseFactory } from './unknown-server-error-response-factory';
+import { TestBed } from '@angular/core/testing';
+
+describe('UnknonnServerErrorResponse', () => {
```
Platonn:
```suggestion
describe('UnknownServerErrorResponse', () => {
```

PS. https://marketplace.visualstudio.com/items?itemName=streetsidesoftware.code-spell-checker for the win

core-libs/setup/ssr/error-handling/server-error-response-factory/unknown-server-error-response-factory.ts
```diff
@@ -0,0 +1,42 @@
...
+  CxServerErrorResponse,
+  UnknownServerErrorResponse,
+} from '../server-errors';
+import { Priority } from '@spartacus/core';
+
+/*
+ * The UnknownServerErrorResponseFactory creates an UnknownServerErrorResponse if HTTP error occur during server-side rendering.
+ *
+ * This factory is used as a fallback to handle any server error that
+ * does not match any other cases.
```
Platonn:
```suggestion
 * does not match the applicability criteria of other Server Error Response Factories.
 ```

File: core-libs/setup/ssr/error-handling/server-error-response-factory/provide-server-error-response-factory.ts [hunk 13]
```diff
@@ -0,0 +1,26 @@
...
+
+import { Provider } from '@angular/core';
+import { CmsPageNotFoundServerErrorResponseFactory } from './cms-page-not-found-server-error-response-factory';
+import { UnknownServerErrorResponseFactory } from './unknown-server-error-response-factory';
+import { SERVER_ERROR_RESPONSE_FACTORY } from './server-error-response-factory';
+
+/**
+ * This file provides the default server error response factories.
+ */
+export const provideServerErrorResponseFactory = (): Provider[] => [
```
Platonn: provideServerErrorResponseFactory vs provideServerErrorResponseFactories?

core-libs/setup/ssr/error-handling/server-errors/cx-errors.ts
```diff
@@ -11,6 +11,6 @@
  *  - CmsPageNotFoundServerError SR0002: Used when the requested CMS page is not found.
  * */
 export enum CxError {
-  UnknownServerError = 'SR0001',
-  CmsPageNotFoundServerError = 'SR0002',
+  UnknownServerError = 'UnknownServerError',
+  CmsPageNotFoundServerError = 'CmsPageNotFoundServerError',
 }
```
Platonn: I'm sorry, I still need a help to understand why we need this enum.

btw. Those values are not about Spartacus _errors_, but rather `ServerErrorResponse`s, so the JSDoc and the enum name are also a bit confusing to me.

pawelfras: As we discussed, let's abandon the enum for now and use hardcoded values inside server error response classes.

core-libs/setup/ssr/error-handling/multi-error-handlers/server-responding-error-handler.ts
```diff
@@ -17,7 +18,13 @@ import { MultiErrorHandler } from '@spartacus/core';
...
 })
 export class ServerRespondingErrorHandler implements MultiErrorHandler {
-  handleError(_error: Error): void {
-    // TODO: CXSPA-6576 use SEVER_ERROR_RESPONSE_TRANSFORMERS
+  protected factories = inject(SERVER_ERROR_RESPONSE_FACTORY);
+
+  handleError(error: unknown): void {
+    //@ts-ignore
+    //TODO:CXSPA-6577 Propagate the error to the OptimizedSsrEngine
+    const cxServerError = resolveApplicable(this.factories, [error]).create(
```
Platonn:
```suggestion
    const cxServerErrorResponse = resolveApplicable(this.factories, [error]).create(
```

Platonn (approved):

PR #18685, opened 2024-03-29
fix: Changing language in saved cart page now reloads saved carts

kpawelczak:
closes: [CXSPA-6551](https://jira.tools.sap/browse/CXSPA-6551)


feature-libs/cart/saved-cart/components/list/saved-cart-list.component.ts
```diff
@@ -82,6 +94,31 @@ export class SavedCartListComponent implements OnInit, OnDestroy {
...
+
+      if (!contexts.length) {
+        return;
+      }
+
+      this.subscription.add(
+        from(contexts)
+          .pipe(
+            mergeMap((context: SiteContextType) => {
+              // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
rmch91: Would you consider assigning `siteContextService` to a variable within the `if block` to avoid using the `non-null assertion operator` and disabling linting rules? By assigning `const siteContextService = this.siteContextService;` inside of the `if block` and using `siteContextService.getActiveItem` instead of the class member `this.siteContextService`, typescript will know that it is not null.

feature-libs/cart/saved-cart/components/list/saved-cart-list.component.ts
```diff
@@ -82,6 +94,31 @@ export class SavedCartListComponent implements OnInit, OnDestroy {
     event.stopPropagation();
   }
 
+  protected observeAndReloadSavedCartOnContextChange() {
```
rmch91, line 97:
```suggestion
  protected observeAndReloadSavedCartOnContextChange() {
    if (this.siteContextService) {
      const contexts: SiteContextType[] = Object.values(SiteContextType);
      const siteContextService = this.siteContextService;

      if (!contexts.length) {
        return;
      }

      this.subscription.add(
        from(contexts)
          .pipe(
            mergeMap((context: SiteContextType) => {
              return siteContextService.getActiveItem(context).pipe(skip(1));
            })
          )
          .subscribe(() => {
            this.savedCartService.loadSavedCarts();
          })
      );
    }
  }
```

rmch91 (changes requested):

rmch91 (commented):
Could we also add some test cases for the new method?

feature-libs/cart/saved-cart/components/list/saved-cart-list.component.ts
kpawelczak: sure

Matejk00 (approved):

rmch91 (approved):

PR #18690, opened 2024-04-02
Handle read-only

Larisa-Staroverova:
If `read-only` is **undefined**, **null** or **false**, then `Edit Configuration` link should be displayed and one can navigate to the editable configuration.
Otherwise, `Display Configuration` link should be displayed and one can navigate to the red.only configuration.

### DIFF
--- a/feature-libs/product-configurator/common/components/configurator-cart-entry-info/configurator-cart-entry-info.component.html
+++ b/feature-libs/product-configurator/common/components/configurator-cart-entry-info/configurator-cart-entry-info.component.html
@@ -37,7 +37,7 @@
       <cx-configure-cart-entry
         *ngIf="(shouldShowButton$ | async) && orderEntry?.product?.configurable"
         [cartEntry]="orderEntry"
-        [readOnly]="(readonly$ | async) ?? true"
+        [readOnly]="(readonly$ | async) ?? false"
         [msgBanner]="false"
         [disabled]="quantityControl.disabled"
       ></cx-configure-cart-entry
--- a/feature-libs/product-configurator/common/components/configurator-issues-notification/configurator-issues-notification.component.html
+++ b/feature-libs/product-configurator/common/components/configurator-issues-notification/configurator-issues-notification.component.html
@@ -13,11 +13,11 @@
             (shouldShowButton$ | async) && orderEntry?.product?.configurable
           "
           [cartEntry]="orderEntry"
-          [readOnly]="(readonly$ | async) ?? true"
+          [readOnly]="(readonly$ | async) ?? false"
           [msgBanner]="true"
           [disabled]="quantityControl.disabled"
-        ></cx-configure-cart-entry
-      ></ng-container>
+        ></cx-configure-cart-entry>
+      </ng-container>
     </div>
   </ng-container>
 </ng-container>
--- a/projects/storefrontapp-e2e-cypress/cypress/helpers/product-configurator-cart.ts
+++ b/projects/storefrontapp-e2e-cypress/cypress/helpers/product-configurator-cart.ts
@@ -24,6 +24,23 @@ export function clickOnEditConfigurationLink(cartItemIndex: number): void {
 }
 
 /**
+ * Clicks on the 'Edit Configuration' link in add-to-cart dialog.
+ *
+ */
+export function clickOnEditConfigurationLinkInAddToCartDialog(): void {
+  cy.get('cx-added-to-cart-dialog').within(() => {
+    cy.get('cx-configurator-cart-entry-info').within(() => {
+      cy.get('cx-configure-cart-entry')
+        .find('a:contains("Edit")')
+        .click()
+        .then(() => {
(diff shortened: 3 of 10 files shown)


feature-libs/product-configurator/common/components/configurator-cart-entry-info/configurator-cart-entry-info.component.html
```diff
@@ -37,7 +37,7 @@
       <cx-configure-cart-entry
         *ngIf="(shouldShowButton$ | async) && orderEntry?.product?.configurable"
         [cartEntry]="orderEntry"
-        [readOnly]="(readonly$ | async) ?? true"
+        [readOnly]="(readonly$ | async) ?? false"
```
Uli-Tiger, line 40: Wouldn't you agree to have unit test asserting the default behavior, so we do not do this mistake again?

feature-libs/product-configurator/common/components/configurator-issues-notification/configurator-issues-notification.component.html
```diff
@@ -13,11 +13,11 @@
             (shouldShowButton$ | async) && orderEntry?.product?.configurable
           "
           [cartEntry]="orderEntry"
-          [readOnly]="(readonly$ | async) ?? true"
+          [readOnly]="(readonly$ | async) ?? false"
```
Uli-Tiger, line 16: Wouldn't you agree to have unit test asserting the default behavior, so we do not do this mistake again?

projects/storefrontapp-e2e-cypress/cypress/helpers/product-configurator-cart.ts
```diff
@@ -23,6 +23,23 @@ export function clickOnEditConfigurationLink(cartItemIndex: number): void {
...
 
+/**
+ * Clicks on the 'Edit Configuration' link in add-to-cart dialog.
+ *
+ */
+export function clickOnEditConfigurationLinkInAddToCartDialog(): void {
+  cy.get('cx-added-to-cart-dialog').within(() => {
+    cy.get('cx-configurator-cart-entry-info').within(() => {
+      cy.get('cx-configure-cart-entry')
+        .find('a:contains("Edit")')
```
Uli-Tiger: we should avoid searching elements using translated text. In this case wouldn't it be sufficient to click for any/first a tag inside? There is only one, or?

projects/storefrontapp-e2e-cypress/cypress/helpers/product-configurator-cart.ts
```diff
@@ -23,6 +23,23 @@ export function clickOnEditConfigurationLink(cartItemIndex: number): void {
...
+        .click()
+        .then(() => {
+          cy.location('pathname').should('contain', '/cartEntry/entityKey/');
```
Uli-Tiger: should we check here that the configuration is in edit mode and  not in read-only mode?

projects/storefrontapp-e2e-cypress/cypress/helpers/product-configurator-cart.ts
```diff
@@ -23,6 +23,23 @@ export function clickOnEditConfigurationLink(cartItemIndex: number): void {
     });
 }
 
+/**
+ * Clicks on the 'Edit Configuration' link in add-to-cart dialog.
+ *
+ */
+export function clickOnEditConfigurationLinkInAddToCartDialog(): void {
```
Uli-Tiger: do you think it would make sense to move this into the common helper, where also some other functions of the add-to-cart dialog are placed?

Uli-Tiger (changes requested):
Good finding 👍 
Although i like that we now have a cypress test for it, i strongly advocate for having it on top of unit tests, so i would be great if you could provide unit tests, as well.

projects/storefrontapp-e2e-cypress/cypress/helpers/product-configurator-cart.ts
Larisa-Staroverova: Agree! Done!

projects/storefrontapp-e2e-cypress/cypress/helpers/product-configurator-cart.ts
Larisa-Staroverova: Depending on the state of the configuration whether it is read-only or not, the corresponding link will be displayed.
To avoid the issue, the link text should be verified before clicking on the link.
if the configuration is not read-only then 'Edit Configuration' link is displayed, otherwise 'Display Configuration' link.
I refactored this method accordingly.

projects/storefrontapp-e2e-cypress/cypress/helpers/product-configurator-cart.ts
Larisa-Staroverova: Agree! I added the check that verifies if the configuration page is rendered.

Uli-Tiger (approved):
> tested it also with CPQ

@Uli-Tiger, I have retested it with CPQ. It works as expected

PR #18696, opened 2024-04-03
Retry service action in case of 401 error

Matejk00:
Closes: CXSPA-6030

### DIFF
--- a/integration-libs/opf/base/core/services/opf-payment-apple-pay.service.ts
+++ b/integration-libs/opf/base/core/services/opf-payment-apple-pay.service.ts
@@ -52,7 +52,7 @@ export class OpfPaymentApplePayService {
       backOff({
         /**
          * We should retry this sequence only if the error is an authorization error.
-         * It menas that `accessCode` (OTP signature) is not valid or expired and we need to refresh it.
+         * It means that `accessCode` (OTP signature) is not valid or expired and we need to refresh it.
          */
         shouldRetry: isAuthorizationError,
         maxTries: opfAuthorizationErrorRetry,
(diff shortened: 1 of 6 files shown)


integration-libs/opf/base/core/services/opf-payment-apple-pay.service.ts
```diff
@@ -43,6 +47,15 @@ export class OpfPaymentApplePayService {
           applePayWebSessionRequest,
           otpKey
         );
+      }),
+      catchError((error) => throwError(error)),
+      backOff({
+        /**
+         * We should retry this sequence only if the error is an authorization error.
+         * It menas that `accessCode` (OTP signature) is not valid or expired and we need to refresh it.
```
FollowTheFlo: typo 'menas' -> 'means', same on below comments

integration-libs/opf/checkout/components/opf-checkout-payment-wrapper/opf-checkout-payment-wrapper.service.ts
```diff
@@ -179,6 +190,10 @@ export class OpfCheckoutPaymentWrapperService {
   protected handlePaymentInitiationError(
     err: HttpErrorModel
   ): Observable<Error> {
+    if (isAuthorizationError(err)) {
+      throwError(err);
```
FollowTheFlo: i believe either we use
`return throwError(()=>err)`
or we use the non rxjs 'throw' without return:
`throw(()=>err`

below explanation:
https://rxjs.dev/api/index/function/throwError

FollowTheFlo (approved):

PR #18700, opened 2024-04-04
Do not display 'Show Details' button for read-only base product

Larisa-Staroverova:
Closes [CXSPA-6484](https://jira.tools.sap/browse/CXSPA-6484)

Please, merge as a bug fix: 
> `fix:` for a bug fix


feature-libs/product-configurator/common/components/configure-product/configure-product.component.ts
```diff
@@ -76,6 +77,38 @@ export class ConfigureProductComponent {
...
+   *
+   * @param product - product
+   * @returns - If a product a base product then returns true, otherwise false.
+   * @protected
+   */
+  protected isBaseProduct(product: Product): boolean {
+    return (
+      product.baseProduct === product.code ||
+      (!product.baseProduct && !!product.code)
+    );
```
Uli-Tiger:
```suggestion
    return (
      product.baseProduct === product.code || !product.baseProduct 
    );
```
I assume `product.code` may never be `undefined`, as we query the product data using the code.

feature-libs/product-configurator/common/components/configure-product/configure-product.component.ts
```diff
@@ -76,6 +77,38 @@ export class ConfigureProductComponent {
...
+      this.isConfiguratorTypeReadOnly(product.configuratorType) &&
+      this.isBaseProduct(product)
+    );
+  }
+
+  /**
+   * Verifies if a product is a base product.
+   *
+   * In the catalog the product is a base one when baseProduct and code are defined and equal.
+   * In the product detail page a product is a base one when only code is defined.
```
Uli-Tiger: the different behavior is really strange.....

i would phrase the comment like this:
```suggestion
   * The `baseProduct` property contains the product code of the base product.
   * So if the baseProduct is undefined or if the provided code equals the code of the product itself,
   * we assume this product to be a base product.
```

Uli-Tiger (dismissed):
In addition to the 2 comments below, do we also need to handle the case when the customer is still on older release, maybe not even using the make-to-stock feature. Will this break anything?

Larisa-Staroverova:
Hi @Uli-Tiger,
thank you for your feedback.

> do we also need to handle the case when the customer is still on older release, maybe not even using the make-to-stock feature. Will this break anything?

If make-to-stock scenario is **not** active, then `configure` button will be displayed instead of `show details`

if make-to-stock scenario is **active**, then configurator type with READ_ONLY postfix will be returned (see `isConfiguratorTypeReadOnly` method)

To react to the older versions a feature toggle like 'productConfiguratorMakeToStockV2' should be defined that's why I asked not to merge this PR because of it.

Uli-Tiger:
> To react to the older versions a feature toggle like 'productConfiguratorMakeToStockV2' should be defined that's why I asked not to merge this PR because of it.

agree - do we have a task for it?

Larisa-Staroverova:
> agree - do we have a task for it?

Not yet. In first line we wanted to find a solution.
The first version of `make-to-stock` has not been merged yet.

AlexanderKuschner (approved):

hackergil (approved):

PR #18701, opened 2024-04-04
remove Unused variable

kimhw0630:
Closes: https://jira.tools.sap/browse/CXSPA-6642


kpawelczak (approved):

projects/storefrontlib/cms-components/misc/site-context-selector/site-context-selector.component.ts
```diff
@@ -17,7 +16,6 @@ import { SiteContextType } from './site-context.model';
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
 export class SiteContextSelectorComponent {
-  siteContextService: SiteContext<any>;
```
kpawelczak, line 20: Would this be a breaking change?

kpawelczak (changes requested):

projects/storefrontlib/cms-components/misc/site-context-selector/site-context-selector.component.ts
kpawelczak: We might not use it but customers could, it's highly unlikely but still possible.

kpawelczak (approved):

PR #18705, opened 2024-04-05
spike: generic link component

Platonn:
related to https://jira.tools.sap/browse/CXSPA-6329


projects/storefrontlib/shared/components/generic-link/generic-link.component.ts
```diff
@@ -91,6 +103,19 @@ export class GenericLinkComponent implements OnChanges {
     return this.routeParts.fragment ?? undefined;
   }
 
+  onClick(event: UIEvent): void {
+    event.preventDefault();
+    if (this.isExternalUrl() || this.target === '_blank') {
+      window.open(this.url as string, this.target ?? '');
```
Platonn, line 109: It feels a bit intrusive to overwrite the native on-click behavior of `<a>` elements. It should be done with caution. I'm wondering if we break other native behaviors of links as a side effect. For instance:
When I click a link with a pressed `Command` button - I'd expect it would open in new tab.

For inspiration you might want to take a look at Angular's `routerLink` [source code](https://github.com/angular/angular/blob/1e3bcfecbab260fb66a1ac50e77192bb98cc0f3e/packages/router/src/directives/router_link.ts#L276-L284). They don't do window.open, but rather just `return true` to allow for native behaviors of on-click in some cases (e.g. when a `Command` button is pressed whili clicking a link)

projects/storefrontlib/shared/components/generic-link/generic-link.component.ts
```diff
@@ -77,6 +77,18 @@ export class GenericLinkComponent implements OnChanges {
...
   }
 
+  get linkUrl(): string {
+    let url: string;
+    if (this.isExternalUrl()) {
+      url = this.url as string;
+    } else {
+      const homeUrlTree = this.router.createUrlTree(['']);
+      const homeUrl = this.router.serializeUrl(homeUrlTree).slice(0, -1);
+      url = `${homeUrl}${this.url}`;
```
Platonn, line 87: 1. will it work correctly also when `this.url` contains site context? e.g. `this.url` is `/electronics-spa/en/USD/faq`. I guess not - I believe it would return `<site.com>/electronics-spa/en/USD/electronics-spa/en/USD/faq` (doubled site context).

2. doe the current solution really compile correctly? `this.url` is of type `string | any[]`. I believe the above implementation won't work when `this.url` is an array of UrlCommands.

3. did you consider instead using `return this.routingService.getFullUrl(this.url)`?

PR #18706, opened 2024-04-05
fix: CXSPA-6581 Race condition in OpenID code flow causes redirect loop or double redirect on login

pawelfras:
This PR contains an alternative fix for the redirect loop caused by race conditions in OpenID code flow when all routes in the application are protected (required login from the beginning). LoginGuard tried to check whether the user was logged in before the code/token was taken from the URL and stored in the application which caused a redirect loop.

To fix that, we run `authStatePersistenceService.initSync()` and `authService.checkOAuthParamsInUrl()` before initial navigation, to ensure the auth state is set. This was achieved by introducing `MULTI_LOCATION_INITIALIZED` being a wrapper on `LOCATION_INITIALIZED`, to which you can pass Promise blocking initial navigation. A wrapper was needed to allow multi-providing. 

QA steps:

build and deploy Spartacus packages locally: 
npx ts-node ./tools/schematics/testing.ts in the SPA root folder
generate fresh ng17 application: 
npx @angular/cli new my-app --standalone=false --ssr=false --style=scss
go to the project's directory and install Spartacus with SSR from local packages:
npx @angular/cli add @spartacus/schematics@latest
in the generated application, add the following config to enable protected routes and Authorization Code Flow
QA steps:
  `npx ts-node ./tools/schematics/testing.ts in the SPA root folder`
  `npx @angular/cli new my-app --standalone=false --ssr=false --style=scss`
`npx @angular/cli add @spartacus/schematics@latest`
- in the generated application, add the following config to enable protected routes and Authorization Code Flow
```
    provideConfig(<OccConfig>{
      backend: {
        occ: {
          baseUrl: 'https://40.76.109.9:9002',
        },
      },
    }),
    provideConfig(<SiteContextConfig>{
      context: {
        urlParameters: ['baseSite', 'language', 'currency'],
        baseSite: ['electronics-spa'],
        currency: ['USD'],
        language: ['en'],
      },
    })
    provideConfig(<RoutingConfig>{
      routing: {
        protected: true,
      },
    }),
    provideConfig(<AuthConfig>{
      authentication: {
        client_id: 'client4kyma',
        client_secret: 'secret',
        tokenEndpoint: '/oauth/token',
        revokeEndpoint: '/oauth/revoke',
        loginUrl: '/oauth/authorize',
        OAuthLibConfig: {
          responseType: 'code',
          scope: '',
          customTokenParameters: ['token_type'],
          strictDiscoveryDocumentValidation: false,
          skipIssuerCheck: true,
          disablePKCE: true,
          oidc: false,
          clearHashAfterLogin: false,
        },
      },
    }),
```
- run app and verify if there is no redirection loop when logging


projects/core/src/multi-location-initialized/multi-location-initialized.module.ts
```diff
@@ -0,0 +1,37 @@
...
+ */
+function locationInitializedFactory() {
+  const initializers =
+    inject(MULTI_LOCATION_INITIALIZED, { optional: true }) ?? [];
+  const promiseInitializers = initializers.map((initializer) => initializer());
+  return Promise.all(promiseInitializers);
+}
+
+@NgModule()
+export class MultiLocationInitializedModule {
```
Platonn: I don't think we need a new module for it, especially with this name.
What do you think about defining the LOCATION_INITIALIZED hook in the core `RoutingModule` ?

projects/core/src/multi-location-initialized/multi-location-initialized.module.ts
```diff
@@ -0,0 +1,37 @@
+/*
+ * SPDX-FileCopyrightText: 2024 SAP Spartacus team <spartacus-team@sap.com>
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { LOCATION_INITIALIZED } from '@angular/common';
+import { ModuleWithProviders, NgModule, inject } from '@angular/core';
+import { MULTI_LOCATION_INITIALIZED } from './multi-location-initialized-token';
```
Platonn: MULTI_LOCATION_INITIALIZED vs LOCATION_INITIALIZED_MULTI?

MULTI_LOCATION_INITIALIZED sounds to me like multiple locations are initialized.

projects/core/src/config/config-initializer/config-initializer.module.ts
```diff
@@ -30,8 +30,11 @@ export function configInitializerFactory(
 
 export function locationInitializedFactory(
   configInitializer: ConfigInitializerService
-): Promise<Config> {
-  return lastValueFrom(configInitializer.getStable());
+): () => Promise<Config> {
+  return () =>
+    lastValueFrom(configInitializer.getStable()).finally(() =>
+      console.log('config initialized')
```
Platonn: Let's clean up console.logs

projects/core/src/config/config-initializer/config-initializer.module.ts
```diff
@@ -30,8 +30,11 @@ export function configInitializerFactory(
 
 export function locationInitializedFactory(
   configInitializer: ConfigInitializerService
-): Promise<Config> {
-  return lastValueFrom(configInitializer.getStable());
+): () => Promise<Config> {
```
Platonn: Too bad, this function is in public API, so we shouldn't change its signature from `Promise` to `()=>Promise`.

I suggest to `@deprecate` this function and create a new one instead, using `inject()` and no `deps[]` array. And in the new function use the return type `() => Promise<Config> {`

PS. let's make sure the new function doesn't land accidentally in the public api

projects/core/src/multi-location-initialized/multi-location-initialized.module.ts
pawelfras: Agreed 👍

Platonn:
QAd various combinations in a fresh NG17 application + Spartacus libs built from sources on this branch and published via verdaccio:
asm/no-asm | csr/ssr/ssr-without-js | password/implicit/code | protected/non-protected

Each time I tested "open new incognito window with a deeplink URL of a PDP. login. verify I'm logged in and redirected back to this same PDP"

Results: looks good ✅ 

The only findings are unrelated to the bug fixed in this PR nor to the fix itself - I've cross-checked it against Spartacus v21 from RBSC. Here are the issues found, for reference:

- oauth implicit or code flow + auto-base-site configuration => after logging in 3rd party auth server we're redirected to `localhost:4200/#some-oauth-url-params-here` . and we get error in Spartacus APP_INITIALIZER: cannot find a BaseSite. Cause: sample data. Those params make matching current URL against regexes from /basesites endpoint to fail. It's just dummy sample data
- SSR without JS + protected storefront + OAuth implicit/code flow => nothing meaningful visible is rendered in SSR (blank page). But when you run SSR with JS, then we're redirected to 3rd party auth server page as soon as JS kicks in. and then works as expected.
- ASM (opening site with ?asm=true) + protected storefront + OAuth implicit/code flow  => functional gap: you cannot eaisly login as ASM agent. Because as soon as JS page renders, you're redirected to 3rd party auth server page. Workaround: you have to login first as SOME USER in 3rd party auth page. Then you are redirected back to Spartacus and you can see the ASM login form in top of the mate. THEN you login as ASM agent in this top form. Then you terminate the session of the logged SOME USER. Now you can use the top menu to impersonate other users if you want. Anyway, if you refresh page, you're redirected to 3rd party page again to login as SOME USER

projects/core/src/routing/routing.module.ts
```diff
@@ -44,6 +51,19 @@ export function initSecurePortalConfig(
   return null;
 }
 
+/**
+ * Factory function that initializes the location.
+ * It retrieves the initializers from the LOCATION_INITIALIZED_MULTI injection token,
+ * invokes each initializer, and returns a promise that resolves when all initializers have completed.
```
Platonn:
```suggestion
* Factory function for Angular's injection token LOCATION_INITIALIZED.
* 
* Note: LOCATION_INITIALIZED is an Angular's API (https://angular.io/api/common/LOCATION_INITIALIZED).
*          Only when the Promise in this injection token is resolved, then Angular
*          will start the initial navigation in the Router.
* 
* Our factory retrieves the initializers from the `LOCATION_INITIALIZED_MULTI` 
* injection token of Spartacus, invokes each initializer, and returns a Promise
* that resolves when all initializers have completed.
* 
```

projects/core/src/auth/user-auth/user-auth.module.ts
```diff
@@ -53,6 +54,22 @@ export function authStatePersistenceFactory(
   return result;
 }
 
+/**
+ * Factory to ensure that auth is initialized and its stated is synced between browser storage and in-memory storage.
```
Platonn:
```suggestion
 * Factory to ensure that auth is initialized and its state is synced between
 * browser storage and in-memory storage.
```

Platonn (dismissed):

Platonn (approved):

aldali:
We recently encountered an issue during our migration from Spartacus `2211.19.0 `to `2211.27.0.` 

After the upgrade, the `spartacus⚿⚿auth` token is no longer stored in browser local storage. As a result, when a user logs in, the access token isn't saved. This causes the user to be logged out upon refreshing the app due to the missing access token. 

After investigating, I identified that the issue stems from `authStatePersistenceFactory` not being invoked by Angular injection tokens as it was previously. 

Here's the original implementation of `authStatePersistenceFactory`: 

```
export function authStatePersistenceFactory( authStatePersistenceService: AuthStatePersistenceService ) 
{ 
  return () => authStatePersistenceService.initSync(); 
} 
```


Previously, it was triggered through: 
```
{ 
  provide: APP_INITIALIZER, 
  useFactory: authStatePersistenceFactory, 
  deps: [AuthStatePersistenceService], 
  multi: true, 
} 
```

After the fix, however, it is invoked via:
```
{ 
  provide: LOCATION_INITIALIZED_MULTI, 
  useFactory: authInitializedFactory, 
  multi: true, 
} 
```

In our app, we have overridden the `AppRoutingModule from @spartacus/storefront `with our own `CustomAppRoutingModule`, where we provide different `RouterConfigOptions` through 

`RouterModule.forRoot([], { RouterConfigOptions }). `

Since we did not explicitly set the `initialNavigation` value, the default (`enabledNonBlocking`) is used, which skips initializing `LOCATION_INITIALIZED`. As a result, `authInitializedFactory()` is not invoked, preventing `authStatePersistenceService.initSync()` from setting the `spartacus⚿⚿auth` token in local storage. 


Previously, since `authStatePersistenceFactory` was provided via `APP_INITIALIZER`, it was always invoked, ensuring the `spartacus⚿⚿auth` token was set. 

The issue is fix when the `initialNavigation` value is set explicitly in our `CustomAppRoutingModule` to `enabledBlocking`

Considering this situation, would it be possible to enhance this implementation to avoid such side effects of switching from `APP_INITIALIZER` to `LOCATION_INITIALIZED_MULTI`? 

Alternatively, could we require the explicit setting of `initialNavigation` to `enabledBlocking` when using a `CustomAppRoutingModule` to prevent this issue (or at least to mention it in the upgrade notes)?

PR #18708, opened 2024-04-05
fix: schematics replace local AppRoutingModule with Spartacus' AppRoutingModule

Platonn:
**BEFORE:**

Spartacus installation schematics silently failed to add the Spartacus' `AppRoutingModule` import to the customer's app.module. Instead the Angular's generated `AppRoutingModule` (yes, the same name) remained in the app.module. This caused the Angular's `RouterModule` to be misconfigured - it didn't meet the requirements of Spartacus (e.g. missing options: `RouterModule.forRoot([], { initialNavigation, anchorScrolling} )`).

**NOW:**

Now we added a fix: we replace the local file `AppRoutingModule` (created by Angular by default) with our Spartacus `AppRoutingModule`

**WHY:**

We noticed it only now. But since ng17 Angular creates new applications with "routing" option enabled by default (see ng17 commit: https://github.com/angular/angular-cli/commit/1a6a139aaf8d5a6947b399bbbd48bbfd9e52372c). Because of that, new Angular 17 applications contain a local file `AppRoutingModule` which clashes (by name) with our schematics' attempt to add our `AppRoutingModule` **from Spartacus**.

The problem did not occur before Angular v17, because apps created didn't use to contain a local `AppRoutingModule` (due to "routing" option being disabled by default in the `ng new` command).

Note: now in all unit tests of our schematics we use an Angular app with implicit default "--routing=true", to follow the defaults of Angular in our default test cases

Note 2: to cover the opposite case too, we added now a special unit test for our schematics, where we use an Angular app with deliberate option "--routing=false"  too

**QA steps:**
**1. With routing implicitly enabled**

- build libs locally
- publish libs to verdaccio, e.g. using `schematics/testing.ts` script
- `npx @angular/cli@17 new test-ng17-spa-AppRoutingModule --standalone=false` (will install with default routing:true)
- verify in `app.module` you have `import { AppRoutingModule } from './app-routing.module';`
- `cd test-ng17-spa-AppRoutingModule`
- `ng add @spartacus/schematics`
- verify in `app.module` you have `import { AppRoutingModule } from '@spartacus/storefront';`
- verify in `app.module` the `imports` array of `AppModule` contains `AppRoutingModule`

Diff of `app.module` focused on `AppRoutingModule` import:
image

Diff of the whole `app.module` (just for reference):
image

**2. With routing explicitly disabled**

- Repeat all steps from point (1), but pass flag `--no-routing` to the command `npx @angular/cli@17 new ....`

Diff of `app.module` focused on `AppRoutingModule` import:
image

Diff of the whole `app.module` (just for reference):
image

closes https://jira.tools.sap/browse/CXSPA-6726


projects/schematics/src/add-spartacus/index_spec.ts
```diff
@@ -20,41 +20,41 @@ import { Schema as SpartacusOptions } from './schema';
...
+  name: 'workspace',
+  newProjectRoot: 'projects',
+  version: '0.5.0',
+};
+
+const appOptions: ApplicationOptions = {
+  name: 'schematics-test',
+  inlineStyle: false,
+  inlineTemplate: false,
+  // routing: true,
```
pawelfras: `nitpick:` Is this line commented on intentionally? or should we remove this part of the options?

projects/schematics/src/add-spartacus/index.ts
```diff
@@ -488,6 +491,42 @@ function updateAppModule(options: SpartacusOptions): Rule {
...
+function addAppRoutingModuleImport(tree: Tree, sourceFile: SourceFile) {
+  // remove import of AppRoutingModule (NgModule import and module path import), if exists
+  removeModuleImport(sourceFile, {
+    importPath: APP_ROUTING_MODULE_LOCAL_PATH,
+    content: APP_ROUTING_MODULE,
+  });
+
+  // delete local file of AppRoutingModule, if exists
+  tree.visit((filePath: Path) => {
+    if (filePath.endsWith(APP_ROUTING_MODULE_LOCAL_FILENAME)) {
```
pawelfras, line 530: `nitpick:` I would consider whether log here that the file has been deleted, in favor of debug mode.

projects/schematics/src/add-spartacus/index.ts
```diff
@@ -488,6 +491,42 @@ function updateAppModule(options: SpartacusOptions): Rule {
...
+ * it will be removed first, and later the `AppRoutingModule` from `@spartacus/storefront` will be imported.
+ *
+ * Note: Since v17 Angular `ng new` command by default creates a local `AppRoutingModule` file in the project.
+ *       So we have to replace it.
+ *
+ * See Angular enabling routing by default in v17: https://github.com/angular/angular-cli/commit/1a6a139aaf8d5a6947b399bbbd48bbfd9e52372c
+ */
+function addAppRoutingModuleImport(tree: Tree, sourceFile: SourceFile) {
+  // remove import of AppRoutingModule (NgModule import and module path import), if exists
+  removeModuleImport(sourceFile, {
```
pawelfras: `nitpick:` I would consider whether log here that the import path has been changed, in favor of debug mode.

pawelfras (dismissed):

pawelfras:
QA done. LGTM.

projects/schematics/src/add-spartacus/index_spec.ts
Platonn:
```suggestion
```

Platonn:
Added logging. Please let me know if it works for you.

I'm attaching below a screenshot from the installation process with `--debug` flag:

image

pawelfras (approved):
I can confirm that logs appear during the installation:
image

LGTM

PR #18713, opened 2024-04-08
fix: (CXSPA-4344) - Quantity update on enter press

Pio-Bar:
[CXSPA-4344](https://jira.tools.sap/browse/CXSPA-4344) I think we can get away with not using feature flags here. I have reused some of the code that have been marked for removal, hence the comment edit. 

- The Item counter updates the value on 'enter' keypress.
- Should restore focus after data gets updated in cart/add to cart modal.


Zeyber (dismissed):

developpeurweb (changes requested):
While on PDP, when changing the quantity and pressing the “Enter” key, the Add To Cart modal gets triggered. This is an unexpected change of context.
Can this be avoided? The expectation is only for the quantity to get updated in the Qty field.

Pio-Bar:
> While on PDP, when changing the quantity and pressing the “Enter” key, the Add To Cart modal gets triggered. This is an unexpected change of context. Can this be avoided? The expectation is only for the quantity to get updated in the Qty field.

I've disabled the PDP 'add to cart' modal after pressing enter on the quantity input field. Let me know if it behaves as intended. ;)

Zeyber (approved):

developpeurweb (approved):

PR #18714, opened 2024-04-09
Handle add-to-cart exception in case of base product

Larisa-Staroverova:
Closes CXSPA-6483
To avoid duplicated error message for the quick order, first this [PR](https://github.com/SAP/spartacus/pull/18780) has to be merge into `develop`.


feature-libs/product-configurator/rulebased/root/http-interceptors/configurator-bad-request.handler.ts
```diff
@@ -0,0 +1,84 @@
...
+      );
+  }
+
+  protected isRelatedToProductConfigurator(
+    response: HttpErrorResponse
+  ): boolean {
+    const isProductConfiguratorIllegalStateError = this.isNotEmpty(
+      this.getIllegalStateErrorsRelatedToProductConfigurator(response)
+    );
+    return isProductConfiguratorIllegalStateError;
```
Uli-Tiger:
```suggestion
    return this.isNotEmpty(this.getIllegalStateErrorsRelatedToProductConfigurator(response));
```
could be inlined

Uli-Tiger (commented):
is this for review? - anyways just left a nitpick comment, looks good to me otherwise.
@Uli-Tiger, yes, it is.

feature-libs/product-configurator/rulebased/root/http-interceptors/configurator-bad-request.handler.ts
Larisa-Staroverova: Sure, good point! Thanks a lot!

feature-libs/product-configurator/rulebased/root/http-interceptors/configurator-bad-request.handler.ts
```diff
@@ -46,7 +46,12 @@ export class ConfiguratorBadRequestHandler extends HttpErrorHandler {
 
   protected handleIllegalArgumentIssues(message: string): void {
     if (this.isIllegalStateErrorRelatedToMakeToStock(message)) {
-      this.globalMessageService.add(message, GlobalMessageType.MSG_TYPE_ERROR);
+      this.globalMessageService.add(
+        {
+          key: 'configurator.httpHandlers.makeToStockBaseProductIssue',
+        },
+        GlobalMessageType.MSG_TYPE_ERROR
+      );
```
Uli-Tiger, line 54: wouldn't a test make sense here -  so checking that we call the message service with the proper key instead of the exception text.

Uli-Tiger (changes requested):

feature-libs/product-configurator/rulebased/root/http-interceptors/configurator-bad-request.handler.ts
Larisa-Staroverova: Good point! I extended the test accordingly.

Uli-Tiger (approved):

PR #18717, opened 2024-04-10
fix: Fix AppRouting configurations not being active when using install script

Zeyber:
Closes: https://jira.tools.sap/browse/CXSPA-6158


projects/storefrontlib/router/on-navigate.service.ts
```diff
@@ -33,8 +27,10 @@ export class OnNavigateService {
 
   protected subscription: Subscription;
 
-  get hostComponent(): ComponentRef<any> {
-    return this.injector.get(ApplicationRef)?.components?.[0];
+  get hostComponent(): HTMLElement | undefined {
+    return <HTMLElement>(
+      this.injector.get(DOCUMENT).getElementsByTagName('cx-storefront')?.[0]
```
Zeyber: Perhaps we can make the root component tag definable in config. It can be introduced as an additional feature rather than a breaking change.

Zeyber:
QA Steps:
1. Set `BRANCH="feature/CXSPA-6158"` in `scripts/install/config.sh`.
2. Install with `cd scripts/install && ./run.sh install`.
3. Open installed csr app on homepage, click login and wait for navigation to complete.
4. Start pressing tab. You should see that the focus starts at the top of the page (and NOT from the login button).
5. To check this fix is responsible, now modify `projects/storefrontlib/router/config/default-on-navigate-config.ts` and remove the `selectedHostElement: 'cx-storefront',` from the config.
6. Repeat the steps above. At step 4, you should notice the focus now starts from the login button.

RadhepS (approved):
Looks good, I also followed the QA steps and it works as expected

PR #18719, opened 2024-04-11
fix: (CXSPA-1029) - Image text alternative is provided to banners

sdrozdsap:


projects/storefrontlib/cms-components/content/banner/banner.component.ts
```diff
@@ -83,4 +83,15 @@ export class BannerComponent {
       }
     }
   }
+
+  getLinkAriaLabel(data: CmsBannerComponent): string | undefined {
```
Zeyber, line 97: We should have a unit test for this method to keep up code coverage.

Zeyber (changes requested):

projects/storefrontlib/cms-components/content/banner/banner.component.ts
sdrozdsap: Done

Zeyber (approved):

developpeurweb (approved):

PR #18727, opened 2024-04-15
feat/CXSPA-6636: OPPS Coupon Codes & Login required

anjana-bl:
https://jira.tools.sap/browse/CXSPA-6636
https://jira.tools.sap/browse/CXSPA-6637

## General

- [x] Epic has sufficient end to end (cypress) test coverage
- [ ] Schematics code deprecations have been provided and tested
- [x] Epic CSS supports directionality (if required)
- [ ] Epic renders successfully in SSR

## Audits/reviews

- [x] Architecture review of the epic
- [x] Security and threat modeling audit/review from the Security team
- [x] Accessibility audit/review from the UX team

## New Library

## Sample data

- [x] Sample data has been added to the sampledata repo and merged to the next version to be released
- [x] New sample data has been loaded and tested on a dev server
- [x] New sample data has been loaded and tested in our CI server

## Documentation

- [ ] Documentation input is provided for it to be published on the doc site.


integration-libs/opps/README.md
```diff
@@ -0,0 +1,8 @@
+# Spartacus Omnichannel Promotion Pricing Service Integration
+
+Spartacus' OPPS library is an enhancement to Personalization library.
+Omnichannel promotion pricing provides a central price and promotion repository and a promotion pricing service for the calculation of effective sales prices in all sales and communication channels. Spartacus' OPPS library enables opps capabilities into Spartacus storefront
+
+It can be added to the existing Spartacus application by running `@spartacus/opps`. For more information about Spartacus schematics, visit the [official Spartacus schematics documentation page](https://sap.github.io/spartacus-docs/schematics/).
```
Platonn:
```suggestion
It can be added to the existing Spartacus application by running `ng add @spartacus/opps`. For more information about Spartacus schematics, visit the [official Spartacus schematics documentation page](https://sap.github.io/spartacus-docs/schematics/).
```

integration-libs/opps/coupon-codes/config/constants.ts
```diff
@@ -0,0 +1,8 @@
+/*
+ * SPDX-FileCopyrightText: 2024 SAP Spartacus team <spartacus-team@sap.com>
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+export const OPPS_COUPON_LOCAL_STORAGE_KEY = 'couponcodes';
+export const OPPS_COUPON_URL_QUERY_PARAM = 'couponcodes';
```
Platonn: 1. let's prefix the local storage key with `opps-` -> `opps-couponcodes`
2. Do you think we should moreover move those both constants also to OppsConfig, to be eaisly configurable for customers?

integration-libs/opps/schematics/add-opps/__snapshots__/index_spec.ts.snap
```diff
@@ -0,0 +1,29 @@
...
+`;
+
+exports[`Spartacus opps schematics: ng-add opps feature general setup should add the feature using the lazy loading syntax 1`] = `
+"import { NgModule } from '@angular/core';
+import { OppsCouponCodesModule } from "@spartacus/opps/coupon-codes";
+
+@NgModule({
+  declarations: [],
+  imports: [
+    OppsCouponCodesModule
```
Platonn: Architectural question: 
Do you think we should make each bit of OPPS integration a separate NgModule within separate secondary entry point (sub-library) and separate schematics installation option?

I guess we expect in the near future the following additions:
- support of `loginrequired` query param
- perhaps? support of the deprecated `segment-ref` query param
- ... perhaps some other features in the future? (do you know the future outllook)?

Did you consider simply creating one umbrella module that would implicitly cover any future additions, without a need to create for those new features separate secondary entry points and schematics options and modules installed in the customer's app? I mean, importing in customer's app only one umbrella NgModule, e.g. `OppsRootModule`. And inside of it we would import all the new added features?

I'm just asking and I'm really happy to understand your point of view.

--
Here's how I'd imagine the **potential** architecture, if we go for an umbrella module:
- import in customer's app a (new) module `OppsRootModule`
- inside `OppsRootModule` we'd import `OppsCouponCodesModule`

If you envision in the future some lazy-loadable parts of the OPPS library, I think it would be good from the start to follow the standard structure to distinguish eagerly-loaded parts from lazy-loaded ones (i.e. create `/root` secondary entry point from the start). **Only if you think that's the case,** I guess the structure would look something like this:
```
integration-libs/
  | opps/
    |- root/              <-- add secondary entry point
      |- ng-package.json 
      |- opps-root.module.ts
      |- coupon-codes/
        |- opps-coupon-codes.module.ts
        |- occ-opps-coupon-codes.interceptor.ts
```

If there is just one OPPS library option to choose in the installation schematics, then also schematics config/schema would look a bit different I guess:
```diff
- "value": "Opps-Coupon-Codes",
- "label": "OPPS Coupon Codes"
+ "value": "OOPS",
+ "label": "OPPS (Omnichannel Promotion Pricing Service)"
```

**But again, I'm not forcing this solution, but just asking, so can be best prepared for the foreseeable upcoming future developments. And happy to learn your point of view.**

integration-libs/opps/coupon-codes/http-interceptors/occ-opps-coupon.interceptor.ts
```diff
@@ -0,0 +1,91 @@
...
+    if (this.winRef.isBrowser()) {
+      if (!this.config.opps?.couponcodes?.httpHeaderName && isDevMode()) {
+        this.logger.warn(
+          `There is no httpHeaderName configured for OPPS Coupon`
+        );
+      }
+      this.requestHeader =
+        this.config.opps?.couponcodes?.httpHeaderName?.toLowerCase?.();
+    }
+  }
```
Platonn: IMHO the inspection of the browser's URL and saving a queryparam from the browser's URL into local storage should happen in the `LOCATION_INITIALIZED_MULTI` hook of Spartacus.

The interceptor should be responsible only for attaching the value from local storage to the outgoing HTTP requests.

projects/schematics/src/shared/lib-configs/integration-libs/opps-schematics-config.ts
```diff
@@ -0,0 +1,39 @@
...
+  },
+  rootModule: {
+    importPath: SPARTACUS_OPPS_COUPON_CODES,
+    name: OPPS_COUPON_CODES_MODULE,
+    content: `${OPPS_COUPON_CODES_MODULE}`,
+  },
+  /* Through Spartacus Opps code doesnot have a dependency on Personalization,
+  backend occ api of Opps requires Personalization to be enabled , hence adding this dependency
+  If Personalization library is not installed, Personalization & Opps won't apply */
+  dependencyFeatures: [TRACKING_PERSONALIZATION_FEATURE_NAME],
```
Platonn: Nice comment!

Let's add it also to the official README.md of the OPPS library. And let's mention there also that the installation schematics of the OPPS library will also implicitly force imply the installation of the Personalization library

projects/schematics/src/shared/utils/feature-utils.ts
```diff
@@ -129,7 +129,7 @@ export function addFeatures<OPTIONS extends LibraryOptions>(
         featureSchematicConfigMapping.get(feature);
       if (!schematicsConfiguration) {
         throw new SchematicsException(
-          `[Internal] No feature config found for ${feature}. ` +
+          `[Internal] Shit No feature config found for ${feature}. ` +
```
Platonn: 👀

projects/schematics/src/shared/utils/schematics-config-utils.ts
```diff
@@ -13,7 +13,9 @@ import { featureSchematicConfigMapping } from '../schematics-config-mappings';
 export function getConfiguredDependencies(feature: string): string[] {
   const featureConfig = featureSchematicConfigMapping.get(feature);
   if (!featureConfig) {
-    throw new SchematicsException(`No feature config found for ${feature}.`);
+    throw new SchematicsException(
+      `Fuck No feature config found for ${feature}.`
```
Platonn: 👀

integration-libs/opps/coupon-codes/http-interceptors/occ-opps-coupon.interceptor.ts
Platonn: Perhaps you might want to also wrap the logic responsible for saving/reading from local storage to some simple service, e.g. `OppsCouponCodesService` with 2 methods `set(value: string)` and `get(): string` - which under the hood communicate with localStorage.
This can help simplify the logic of your Interceptor (and simplify the logic of the potentially used `LOCATION_INITIALIZED_MULTI` hook) - by abstracting out the implementation detail of communicating with localStorage

Platonn: I guess we should implement some logic for clearing the saved value, e.g. on user LOGOUT event. Or at least we should consult it with the product owner of the OPPS feature.

Otherwise if an user logs out, then another user of this same browser will implicitly "inherit" the coupon codes - which might be potentially against the business rules.

integration-libs/opps/README.md
anjana-bl: done

projects/schematics/src/shared/utils/feature-utils.ts
anjana-bl: my bad, very sorry! was doing some debugging, missed to remove it. my appologies!

projects/schematics/src/shared/utils/schematics-config-utils.ts
anjana-bl: my bad, very sorry! was doing some debugging, missed to remove it. my appologies!

integration-libs/opps/coupon-codes/http-interceptors/occ-opps-coupon.interceptor.ts
anjana-bl: Hi Kris, If we use LOCATION_INITIALIZED_MULTI to capture coupon codes, factory method will be called only once and not the second time, when user launches a another url with a different coupon code -> am i missing something where?

integration-libs/opps/coupon-codes/config/constants.ts
anjana-bl: done

projects/schematics/src/shared/lib-configs/integration-libs/opps-schematics-config.ts
anjana-bl: done

integration-libs/opps/schematics/add-opps/__snapshots__/index_spec.ts.snap
anjana-bl: Hi Kris, Initially we expected customers should be able to install each feature independently. But in our later discussions, we felt it should be fine, if they install all features (coupon codes, login required etc.,) in one go. So i have now modified the structure as you have suggested above.

integration-libs/opps/ng-package.json
```diff
@@ -4,5 +4,10 @@
   "lib": {
     "entryFile": "./public_api.ts"
   },
-  "assets": ["**/*.scss", "schematics/**/*.json", "schematics/**/*.js"]
+  "assets": [
+    "**/*.scss",
+    "migrations/**/*.json",
```
Platonn, line 9: Why migrations?

integration-libs/opps/root/coupon-codes/opps-coupon-codes.module.ts
```diff
@@ -0,0 +1,32 @@
...
+): () => void {
+  return () => {
+    service.saveUrlCouponCodes();
+  };
+}
+
+@NgModule({
+  providers: [
+    ...oppsCouponCodesInterceptors,
+    OppsCouponCodesService,
```
Platonn: nitpick: it doesn't have to be provided in a module, becasue it's an eagerly loaded module.
It suffices to make it `@Injectable({providedIn: 'root'})`

integration-libs/opps/root/coupon-codes/opps-coupon-codes.module.ts
```diff
@@ -0,0 +1,32 @@
...
+
+import { NgModule } from '@angular/core';
+import { LOCATION_INITIALIZED_MULTI } from '@spartacus/core';
+import { oppsCouponCodesInterceptors } from './http-interceptors';
+import { OppsCouponCodesService } from './opps-coupon-codes.service';
+
+export function saveCouponCodesFactory(
+  service: OppsCouponCodesService
+): () => void {
+  return () => {
```
Platonn: nitpick: for better type safety, let's use `inject()` function of Angular in the top of the factory funciton body, instead of passing `deps[]` array in line 27 (it's an old way of injecting deps to factories)

```suggestion
export function saveCouponCodesFactory(): () => void {
  const service = inject(OppsCouponCodesService);
  return () => {
```

integration-libs/opps/root/coupon-codes/opps-coupon-codes.service.ts
```diff
@@ -0,0 +1,42 @@
...
+  getCouponCodes(): string | undefined | null {
+    if (this.localStorageKey) {
+      return this.winRef.localStorage?.getItem(this.localStorageKey);
+    }
+  }
+  clearCouponCodes(): void {
+    if (this.localStorageKey) {
+      this.winRef.localStorage?.removeItem(this.localStorageKey);
+    }
+  }
```
Platonn, line 41: I have problem to find any usages of this method. Perhaps you meant to call it on user "log out" event?

projects/storefrontlib/cms-structure/guards/multi-cms-page-guard.service.ts
```diff
@@ -0,0 +1,40 @@
...
+import { Observable, concat, endWith, first, of, skipWhile } from 'rxjs';
+import { CmsGuardsService } from '../services';
+import { CmsActivatedRouteSnapshot, UnifiedInjector } from '@spartacus/core';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class MultiCmsPageGuardService {
+  protected cxPageGuards = inject(CX_PAGE_GUARD_TOKEN);
+  protected cmsGuardService = inject(CmsGuardsService);
```
Platonn, line 19: IMHO we should not depend here the `CmsGuardsService` (which responsibility is to resolve the **cms-mapping**-driven guards).

Instead, I suggest to extract most of the logic from the existing `CmsGuardsService` to a new service like `ManyGuardsService` with 2 methods:
```ts
  canActivateGuards(
    guards: CanActivate[], 
    route: CmsActivatedRouteSnapshot,
    state: RouterStateSnapshot
  }) { ... }

  canActivateGuard(
    guards: CanActivate, 
    route: CmsActivatedRouteSnapshot,
    state: RouterStateSnapshot
  }) { ... }
```

Then we could reuse this new service in the internals of the existing `CmsGuardsService` and also reuse this new service here.

In other words, we only need here an util service for handling many guards, but we don't need resolving **cms-mapping**-driven guards

Platonn: UPDATE: I've noticed only now that the existing CmsPageService uses sophisticated mechanisms like UnifiedInjector, that I believe are not needed here. Therefore the amount of code that could be reuse is much smaller than I thought. I'm sorry, bear with me... I need to think about it more and I'll get back to you...

projects/storefrontlib/cms-structure/guards/multi-cms-page-guard.service.ts
```diff
@@ -0,0 +1,40 @@
...
+import { RouterStateSnapshot, UrlTree } from '@angular/router';
+import { Observable, concat, endWith, first, of, skipWhile } from 'rxjs';
+import { CmsGuardsService } from '../services';
+import { CmsActivatedRouteSnapshot, UnifiedInjector } from '@spartacus/core';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class MultiCmsPageGuardService {
+  protected cxPageGuards = inject(CX_PAGE_GUARD_TOKEN);
```
Platonn, line 18: Let me think a bit about the naming for the token and accompanying service(s) in the core mechanism and I'll get back to you...

integration-libs/opps/root/coupon-codes/opps-coupon-codes.module.ts
anjana-bl: done

integration-libs/opps/root/coupon-codes/opps-coupon-codes.module.ts
anjana-bl: done

anjana-bl:
Creating a new PR based on the https://github.com/SAP/spartacus/pull/18781 - https://github.com/SAP/spartacus/pull/18803

projects/storefrontlib/cms-structure/guards/multi-cms-page-guard.service.ts
Platonn: done. proposed changes in another PR do develop https://github.com/SAP/spartacus/pull/18781

projects/storefrontlib/cms-structure/guards/multi-cms-page-guard.service.ts
Platonn: done. proposed changes in another PR do develop https://github.com/SAP/spartacus/pull/18781

PR #18730, opened 2024-04-15
feat: support of date attribute type

ChristophHi:
Closes [CXSPA-6596](https://jira.tools.sap/browse/CXSPA-6596) user story


feature-libs/product-configurator/common/assets/translations/en/configurator.json
```diff
@@ -147,13 +147,13 @@
...
       "forAttribute": "{{ value }} for attribute {{ attribute }}",
-      "valueOfAttributeFull": "Value {{ value }} of attribute {{ attribute }}",
+      "valueOfAttributeFull": "Value {{ value }} of attribute {{ attribute }}.",
       "valueOfAttributeFullWithPrice": "Value {{ value }} of attribute {{ attribute }}, Surcharge {{ price }}",
       "selectedValueOfAttributeFull": "Selected value {{ value }} of attribute {{ attribute }}",
       "selectedValueOfAttributeFullWithPrice": "Selected value {{ value }} of attribute {{ attribute }}, Surcharge {{ price }}",
       "readOnlyValueOfAttributeFullWithPrice": "Read-only value {{ value }} of attribute {{ attribute }}, Surcharge {{ price }}",
       "readOnlyValueOfAttributeFull": "Read-only value {{ value }} of attribute {{ attribute }}",
-      "valueOfAttributeBlank": "Value of attribute {{ attribute }} is blank",
+      "valueOfAttributeBlank": "Value of attribute {{ attribute }} is blank.",
```
Uli-Tiger: Is it correct that some of these text end with a dot and others not? - looks a bits strange to me.

feature-libs/product-configurator/common/assets/translations/en/configurator.json
```diff
@@ -222,7 +222,8 @@
       "filterOverviewByGroup": "Filter configuration overview by group {{groupName}}",
       "closeConflictSolverModal": "Close conflict solver modal",
       "closeRestartDialog": "Close the \"Unfinished Configuration\" dialog and navigate back to the product details page",
-      "description": "Click to see a description for this value"
+      "description": "Click to see a description for this value",
+      "spaceKeyForDatePicker": " Press space key to enter date picker."
```
Uli-Tiger: I'm not sure whether it is a good idea to have a leading white space here - i see the risk of that it gets lost during translation. May we should add it where the string are combined. See also my comment below for the HTML file.

feature-libs/product-configurator/rulebased/components/attribute/types/input-field/configurator-attribute-input-field.component.html
```diff
@@ -1,7 +1,27 @@
...
+              : {
+                  attribute: attribute.label
+                }) + ('configurator.a11y.spaceKeyForDatePicker' | cxTranslate)
+        : ('configurator.a11y.valueOfAttributeFull'
+            | cxTranslate
+              : {
+                  value: attribute.userInput,
+                  attribute: attribute.label
+                }) + ('configurator.a11y.spaceKeyForDatePicker' | cxTranslate)
+    "
```
Uli-Tiger: Why not defining two variants of the whole text, to have less text combining which is also a bit error-prone (with appending/prefixing dots and white spaces properly). Maybe some parts are than duplicated, but this is not issue. At least I'm not aware of that we need to optimize this.
```
configurator.a11y.valueOfDateAttributeBlank="Value of attribute {{ attribute }} is blank. Press space key to enter date picker."
configurator.a11y.valueOfDateAttributeFull="Value {{ value }} of attribute {{ attribute }}. Press space key to enter date picker."
```

feature-libs/product-configurator/rulebased/occ/variant/converters/occ-configurator-variant-serializer.ts
```diff
@@ -93,34 +93,43 @@ export class OccConfiguratorVariantSerializer
...
+
+    const inputTypesSetValue = [
+      Configurator.UiType.STRING,
+      Configurator.UiType.SAP_DATE,
+    ];
+
+    const inputTypesSetFormattedValue = [Configurator.UiType.NUMERIC];
+
+    if (singleValueTypes.includes(uiType)) {
       this.retractValue(attribute, targetAttribute);
```
Uli-Tiger, line 127: confusing method name as it only eventually retracts it....

Uli-Tiger (changes requested):
By Default shop is running in US Locale (at least en_US is what is set as locale, when SPA sends en language), hence the date picker using GB format is wrong, isn't it?
Short US format is m/d/yy (order MDY, no leading zeros) while GB short format would be dd/mm/yy (so order DMY, leading zeros). 
So we have actually two issues:

- date picker uses GB format, which is wrong for US locale
- date picker uses 4 digits year, while backend uses 2 digit years. this shall be harmonized.

image

ChristophHi:
Hi @Uli-Tiger regarding to your general remark regarding date formats: See https://sap-cx.slack.com/archives/G01KTLWNJ06/p1713362148644349 (which I also discussed with Jutta). But let me again check if we can somehow improve

feature-libs/product-configurator/common/assets/translations/en/configurator.json
ChristophHi: For those texts the screen reader was better understandable with the dots. Sebastian agreed to change it in those 2 occurences (but said that customers are anyhow likely to change those texts because different screen readers handle the punctuations differently)

feature-libs/product-configurator/common/assets/translations/en/configurator.json
ChristophHi: Understand, will take a look

feature-libs/product-configurator/rulebased/components/attribute/types/input-field/configurator-attribute-input-field.component.html
ChristophHi: Agree

feature-libs/product-configurator/rulebased/occ/variant/converters/occ-configurator-variant-serializer.ts
ChristophHi: That's true. Still we shouldn't change this (breaking). This PR anyhow is not interested at all in this method, I just needed to refactor because of new sonar rules

Uli-Tiger (dismissed):
lgtm

Larisa-Staroverova:
Please, align input fields on mobile widgets: 
image

Larisa-Staroverova:
@steinsebastian, could you, please, check a11y for date attribute type? Thanks a lot in advance!

feature-libs/product-configurator/rulebased/components/attribute/types/input-field/configurator-attribute-input-field.component.ts
```diff
@@ -142,9 +171,16 @@ export class ConfiguratorAttributeInputFieldComponent
...
+      return this.debounceForDateActive
+        ? this.config.productConfigurator?.updateDebounceTime?.date ??
+            this.FALLBACK_DEBOUNCE_TIME_DATE
+        : 0;
+    } else {
+      return (
+        this.config.productConfigurator?.updateDebounceTime?.input ??
+        this.FALLBACK_DEBOUNCE_TIME
+      );
+    }
```
Uli-Tiger, line 184: Can we try to simplify this statement? It has 3 paths, but 2 are the same. So we should reduce it to 2 paths. Maybe like this. 
```suggestion
    return !this.isDateBased(this.attribute) || this.debounceForDateActive
      ? this.config.productConfigurator?.updateDebounceTime?.input ??
          this.FALLBACK_DEBOUNCE_TIME
      : 0;
```
because basically we say if it is date and and debounce is not active return 0, in all other case retuurn configured debounce time or fall back.

Uli-Tiger (changes requested):
just a minor remark

feature-libs/product-configurator/rulebased/components/attribute/types/input-field/configurator-attribute-input-field.component.ts
ChristophHi: No, that doesn't work, because we have different debounce times defined for input and date (as it's a bit more complex to enter a date value)

Uli-Tiger (dismissed):

Larisa-Staroverova (approved):

PR #18731, opened 2024-04-16
fix: addToCart takes actual product code into account

ChristophHi:


feature-libs/cart/base/components/added-to-cart-dialog/added-to-cart-dialog-event.listener.ts
```diff
@@ -28,19 +30,33 @@ export class AddedToCartDialogEventListener implements OnDestroy {
...
+          this.closeModal(event);
+        })
+      );
+    }
   }
-
+  /**
+   * Opens modal based on CartUiEventAddToCart.
+   * @param event Signals that a product has been added to the cart.
+   */
```
Platonn, line 64: [MINOR] 
Thank you for adding JSDocs for the old method. What about marking it as @deprecated with a comment recommendation to enable the Feature Toggle `adddedToCartDialogDrivenBySuccessEvent` and use  the method `openModalAfterSuccess` instead? Then let's remove the old method together with the Feature Toggle after 12 months

feature-libs/cart/base/core/facade/multi-cart.service.ts
```diff
@@ -227,7 +227,8 @@ export class MultiCartService implements MultiCartFacade {
     cartId: string,
     productCode: string,
     quantity: number,
-    pickupStore?: string
+    pickupStore?: string,
+    numberOfEntriesBeforeAdd?: number
```
Platonn: [MAJOR]
Can you help me understand why we have this new parameter of the method `addEntry()?

IMHO it should not be the responsibility of the consumer of the MultiCartService to tell how many products were there in the cart before the addition. As a consumer, I want to only tell which _new item_ I'm putting into the cart.

ChristophHi: Agree to your point. It's better to fetch the number of entries before add in the actual multi-cart.service implementation

feature-libs/cart/base/components/added-to-cart-dialog/added-to-cart-dialog-event.listener.ts
```diff
@@ -103,6 +108,19 @@ export class AddedToCartDialogEventListener implements OnDestroy {
     }
   }
 
+  protected calculateEntryWasMerged(
+    successEvent: CartAddEntrySuccessEvent
+  ): boolean {
+    // In case quantityAdded zero, the system could have run into stock issues.
+    // Then we continue and still launch the dialogue in 'merge' mode in order
+    // to not break the existing behaviour.
```
Platonn: Nice catch. Nice you handled it!

feature-libs/cart/base/components/added-to-cart-dialog/added-to-cart-dialog.component.ts
```diff
@@ -36,12 +36,29 @@ import {
...
+  /**
+   * Number of cart entries before addToCart was triggered.
+   * @deprecated since 2211.24. Enable feature toggle 'adddedToCartDialogDrivenBySuccessEvent'
+   * and use attribute addedEntryWasMerged instead.
+   */
+  numberOfEntriesBeforeAdd?: number;
+  pickupStoreName?: string;
+  /**
+   * Was the product added to the cart merged into an existing cart entry (with increased quantity),
+   * or did the system create a new cart entry?
```
Platonn: nitpick: The question mark in the end of the sentence is a bit confusing. What about:
```suggestion
   * Tells whether the product added to the cart was merged into an existing cart entry (with increased quantity),
   * or the system created a new cart entry.
```

feature-libs/cart/base/root/events/cart.events.ts
```diff
@@ -72,6 +72,7 @@ export class CartAddEntrySuccessEvent extends CartEvent {
   entry?: OrderEntry;
   quantityAdded?: number;
   deliveryModeChanged?: boolean;
+  pickupStore?: string;
```
Platonn: Is this change _really_ related to this PR?

projects/core/src/features-config/feature-toggles/config/feature-toggles.ts
```diff
@@ -43,6 +43,12 @@ export interface FeatureTogglesInterface {
    */
   productConfiguratorAttributeTypesV2?: boolean;
 
+  /**
+   * The addedToCart dialog is driven by 'CartAddEntrySuccessEvent'. Previously it was driven
+   * by 'CartUiEventAddToCart event. Code changes affect 'AddedToCartDialogEventListener'
```
```suggestion
   * by 'CartUiEventAddToCart' event. Code changes affect 'AddedToCartDialogEventListener'
```

feature-libs/cart/base/components/added-to-cart-dialog/added-to-cart-dialog-event.listener.ts
```diff
@@ -61,6 +82,45 @@ export class AddedToCartDialogEventListener implements OnDestroy {
...
+
+    if (dialog) {
+      dialog.pipe(take(1)).subscribe();
+    }
+  }
+
+  protected calculateEntryWasMerged(
+    successEvent: CartAddEntrySuccessEvent
+  ): boolean {
+    // In case quantityAdded zero, the system could have run into stock issues.
```
Platonn: From unit tests and implementation I understood that in case of stock issues the property `quantityAdded` is `undefined` (not `0`).

So what about updating the comment:
```suggestion
    // In case `quantityAdded` is `undefined`, the system could have run into stock issues.
```

Btw. IMHO we could turn this inline comment in to a fully-fledged JSDoc of this method.

feature-libs/cart/base/root/events/cart.events.ts
ChristophHi: It is related (because we need to display the pickup store name in the confirmation dialogue).
But still this is a good hint, on a second look I saw that this information can also be derived from the cart modification via entry.deliveryPointOfService.name, then the event does not need to be touched!

Platonn (approved):

PR #18732, opened 2024-04-16
refactor: use Injector to avoid circular DI

pawelfras:


core-libs/setup/ssr/error-handling/server-error-response-factory/cms-page-not-found-server-error-response-factory.ts
```diff
@@ -24,30 +24,15 @@ export class CmsPageNotFoundServerErrorResponseFactory
...
-        prefix,
-        endpoint,
-      });
-      return (
-        error instanceof HttpErrorResponse &&
-        expectedUrlRegex.test(error.url ?? '')
-      );
-    }
-    return false;
+    const occEndpointsService = this.injector.get(OccEndpointsService);
```
Platonn, line 33: It's not obvious why we use this workaround, so let's add a code comment, e.g.
```suggestion
    // OccEndpointsService could not be injected in the constructor, due to a circular dependency:
    //   ErrorHandler -> *CmsPageNotFoundServerErrorResponseFactory* -> OccEndpointsService -> 
    //   -> BaseSiteService -> NgRx -> ErrorHandler
    const occEndpointsService = this.injector.get(OccEndpointsService);
```

Platonn (approved):

PR #18736, opened 2024-04-17
feat: CXCDS-11272 Add Trending Searches to search box component

darvsorceix:


integration-libs/cds/src/trending-searches/trending-searches.component.ts
```diff
@@ -0,0 +1,71 @@
...
+
+import { Component, OnInit, OnDestroy, Optional } from '@angular/core';
+import { OutletContextData } from '@spartacus/storefront';
+import { TrendingSearchesService } from './trending-searches.service';
+import { map, switchMap, takeUntil } from 'rxjs/operators';
+import { EMPTY, Subject } from 'rxjs';
+
+const MAX_TRENDING_SEARCHES = 5;
+
+export interface SearchBoxOutletTrendingSearches {
```
rmch91: do we really need an export here if it is used only inside of this component?

integration-libs/cds/src/trending-searches/trending-searches.component.ts
```diff
@@ -0,0 +1,71 @@
...
+  maxTrendingSearches?: number;
+}
+
+export interface SearchPhrases {
```
rmch91: Let's create a separate file for this interface as it is used in component and service.

integration-libs/cds/src/trending-searches/trending-searches.component.ts
```diff
@@ -0,0 +1,71 @@
...
+
+@Component({
+  selector: 'cx-trending-searches',
+  templateUrl: './trending-searches.component.html',
+})
+export class TrendingSearchesComponent implements OnInit, OnDestroy {
+  public searchPhrases: SearchPhrases[] = [];
+  private destroy$ = new Subject<void>();
+
+  constructor(
```
rmch91: let's use inject instead of constructor. It will safe us from breaking changes in future.

Class property: 
`protected trendingSearchesService = inject(TrendingSearchesService)`

`protected outletContext = inject(OutletContextData, {optional: true}) as OutletContextData | null;`

integration-libs/cds/src/trending-searches/trending-searches.component.ts
```diff
@@ -0,0 +1,71 @@
...
+    @Optional()
+    protected outletContext: OutletContextData<SearchBoxOutletTrendingSearches>,
+    private trendingSearchesService: TrendingSearchesService
+  ) {}
+
+  ngOnInit() {
+    this.listenToContextChanges();
+  }
+
+  private listenToContextChanges() {
```
rmch91: We are using `protected` access modifiers instead of `private`, to make it easier for customers to override the component if it is required

integration-libs/cds/src/trending-searches/trending-searches.service.ts
```diff
@@ -0,0 +1,47 @@
...
+import { Observable } from 'rxjs';
+import { SearchPhrases } from './trending-searches.component';
+
+const TRENDING_SEARCHES_ENDPOINT_KEY = 'searchIntelligence';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class TrendingSearchesService {
+  constructor(
```
rmch91: The same as above, get rid of constructor please and use `inject()` instead. Also check access modifiers to get rid of `private` and use `protected` instead

rmch91 (changes requested):

integration-libs/cds/src/trending-searches/trending-searches.component.ts
rmch91: please create model folder, place there tranding-searches.model.ts and create index.ts. Example cds/src/profiletag

rmch91:
Please add unit tests for new service and component and check coverage

darvsorceix:
> Please add unit tests for new service and component and check coverage

image

rmch91 (dismissed):

rmch91 (dismissed):

integration-libs/cds/src/recent-searches/recent-searches.component.ts
```diff
@@ -26,35 +27,45 @@ export interface SearchBoxOutlet {
...
-              (phrase) =>
-                phrase.toLowerCase().indexOf(context.search.toLowerCase()) >= 0
-            )
-            .slice(0, context.maxRecentSearches ?? MAX_RECENT_SEARCHES)
-        )
-      )
-    )
-  );
-  outletContext$: Observable<SearchBoxOutlet>;
+  public result$: Observable<string[]>;
```
rmch91, line 36: `minor:` Public fields first

integration-libs/cds/src/recent-searches/recent-searches.component.ts
```diff
@@ -26,35 +27,45 @@ export interface SearchBoxOutlet {
...
+  protected destroy$ = new Subject<void>();
 
   constructor(
     @Optional() protected outletContext: OutletContextData<SearchBoxOutlet>
   ) {}
 
   ngOnInit() {
+    this.result$ = this.outletContext?.context$.pipe(
+      switchMap((context: SearchBoxOutlet) =>
+        this.recentSearchesService.recentSearches$.pipe(
```
rmch91: `medium:` Can we get rid of nested pipes with combineLatest?

```
 this.result$ = combineLatest([
    this.outletContext?.context$,
    this.recentSearchesService.recentSearches$
  ]).pipe(
    map(([context, recentSearches]: [SearchBoxOutlet, string[]]) => 
      recentSearches
        .filter(...)
        .slice(...)
    )
  );
```

integration-libs/cds/src/recent-searches/recent-searches.component.ts
```diff
@@ -26,35 +27,45 @@ export interface SearchBoxOutlet {
...
+                  phrase.toLowerCase().indexOf(context.search.toLowerCase()) >=
+                  0
+              )
+              .slice(0, context.maxRecentSearches ?? MAX_RECENT_SEARCHES)
+          )
+        )
+      )
+    );
+
+    this.result$.pipe(takeUntil(this.destroy$)).subscribe((results) => {
```
rmch91: `major:` Creating two parallel subscriptions, one with async pipe in the template and second here.
Duplicated subscriptions could add problems, harder to debug and maintain.
It seems like this logic could be moved to the stream above.

What about adding `tap()` after the `map()` operator in the previous stream?
And we don't need `takeUntil()` more in this situation.

integration-libs/cds/src/trending-searches/trending-searches.component.html
```diff
@@ -0,0 +1,26 @@
+<ng-container *ngIf="contextObservable | async as results">
```
rmch91: `medium:` It seems like `results` is not used and could be removed

integration-libs/cds/src/trending-searches/trending-searches.component.html
```diff
@@ -0,0 +1,26 @@
...
+  <ng-container *ngIf="searchPhrases?.length">
+    <div class="message">
+      {{ 'cdsTrendingSearches.trendingSearches' | cxTranslate }}
+    </div>
+
+    <ul
+      class="recent-searches"
+      [ariaLabel]="'cdsTrendingSearches.ariaTrendingSearches' | cxTranslate"
+    >
+      <li *ngFor="let phrase of searchPhrases">
```
rmch91, line 11: `medium:` Why not to convert this `searchPhrases` to observable and to use async pipe?
In this case we can get rid of this subscription in the component, and probably we will not need ChangeDetectorRef more since async pipe automatically should trigger change detection.

```
    this.searchPhrases$  = this.getSearchPhrases()
      .pipe(tap(() => {
       this.searchBoxComponentService.setTrendingSearches(
          !!this.searchPhrases.length
        );
     }))

TakeUntil probably also redundant, since async pipe automatically handle unsubscribing
```

integration-libs/cds/src/trending-searches/trending-searches.service.ts
```diff
@@ -0,0 +1,93 @@
...
+            .getUrl(TRENDING_SEARCHES_ENDPOINT_KEY)
+            .replaceAll('${cdsSiteId}', cdsSiteId);
+          const httpsPrefix = `https://${this.cdsConfig.cds.tenant}-${cdsSiteId}.`;
+          const modifiedUrl = originalEndpointUrl.replace(
+            /https:\/\//g,
+            httpsPrefix
+          );
+
+          const fetchTrendingSearches = () => {
+            this.httpClient.get<any>(modifiedUrl).subscribe((data) => {
```
rmch91: `major:` Nested subscriptions are not good. It should be possible to achieve desired behaviour with rx.js operators

integration-libs/cds/src/trending-searches/trending-searches.service.ts
```diff
@@ -0,0 +1,93 @@
...
+              this.trendingSearchesSource.next(data?.searchPhrases);
+            });
+          };
+          fetchTrendingSearches();
+          this.interval = setInterval(
```
rmch91: `major:` let's not mix rx.js logic with native intervals, timeouts. There is a `timer` rx.js operator

integration-libs/cds/src/trending-searches/trending-searches.service.ts
```diff
@@ -0,0 +1,93 @@
...
+  private addTrendingSearchesListener() {
+    this.checkAvailability()
+      .pipe(take(1))
+      .subscribe((cdsSiteId) => {
+        if (cdsSiteId) {
+          const originalEndpointUrl = this.cdsEndpointsService
+            .getUrl(TRENDING_SEARCHES_ENDPOINT_KEY)
+            .replaceAll('${cdsSiteId}', cdsSiteId);
+          const httpsPrefix = `https://${this.cdsConfig.cds.tenant}-${cdsSiteId}.`;
+          const modifiedUrl = originalEndpointUrl.replace(
```
rmch91: `medium:` This method breaks single responsibility rules.
Let's move logic responsible for creating endpoint to separate method.

integration-libs/cds/src/trending-searches/trending-searches.service.ts
```diff
@@ -0,0 +1,93 @@
...
+      takeWhile((val) => !val, true),
+      last()
+    );
+  }
+
+  private addTrendingSearchesListener() {
+    this.checkAvailability()
+      .pipe(take(1))
+      .subscribe((cdsSiteId) => {
+        if (cdsSiteId) {
```
rmch91: `major:` if statement should be replaced with `filter()` operator, so stream will not init value if there will be no cdsSiteId

integration-libs/cds/src/trending-searches/trending-searches.service.ts
```diff
@@ -0,0 +1,93 @@
...
+              this.trendingSearchesSource.next(data?.searchPhrases);
+            });
+          };
+          fetchTrendingSearches();
+          this.interval = setInterval(
+            () => fetchTrendingSearches(),
+            15 * 60000
+          );
+        }
+      });
```
rmch91: Since I provided a lot of comments to one method, I'll leave my suggestion below. Please double check my suggestion also, because I didn't dive too deeply into the topic.

```
this.checkAvailability().pipe(
    take(1),
    filter(Boolean),
    switchMap((cdsSiteId) => {
          const url = this.constructTrendingSearchUrl(cdsSiteId); // Check the adapter pattern used in different libraries in sparacus. Maybe it will have more sense to move it to adapter.

  return timer(0, 15 * 60000).pipe(
        switchMap(() => this.httpClient.get<any>(url).pipe(
          catchError((error) => {
            return of(null); // Not sure what is better for you null or empty array. If null, then add additional filter(Boolean) before takeUntil
          })
        ))
      );
    }),
   takeUntil(this.destroy$)
).subscribe((data) => this.trendingSearchesSource.next(data.searchPhrases));
```

projects/storefrontlib/cms-components/navigation/search-box/search-box-component.service.ts
```diff
@@ -31,6 +31,9 @@ export class SearchBoxComponentService {
   chosenWord = new ReplaySubject<string>();
   sharedEvent = new ReplaySubject<KeyboardEvent>();
 
+  private enableRecentSearches: boolean = false;
```
rmch91: `major:` let's make this field `protected` for easier extensibility and customisation

integration-libs/cds/src/trending-searches/trending-searches.service.ts
```diff
@@ -0,0 +1,93 @@
...
+  private readonly trendingSearchesSource = new ReplaySubject<
+    SearchPhrases[]
+  >();
+  private trendingSearches$ = this.trendingSearchesSource.asObservable();
+
+  constructor() {
+    this.addTrendingSearchesListener();
+  }
+
+  private checkAvailability() {
```
rmch91: `major:` let's make this field `protected` for easier extensibility and customisation

integration-libs/cds/src/trending-searches/trending-searches.service.ts
```diff
@@ -0,0 +1,93 @@
...
+    return interval(250).pipe(
+      concatMap((_) => of((this.winRef.nativeWindow as any)?.Y_TRACKING)),
+      map((result) => result?.config?.cdsSiteId),
+      take(100),
+      takeWhile((val) => !val, true),
+      last()
+    );
+  }
+
+  private addTrendingSearchesListener() {
```
rmch91: `major:` let's make this field `protected` for easier extensibility and customisation

integration-libs/cds/src/trending-searches/trending-searches.service.ts
```diff
@@ -0,0 +1,93 @@
...
+  protected httpClient = inject(HttpClient);
+  protected cdsEndpointsService = inject(CdsEndpointsService);
+  protected cdsConfig = inject(CdsConfig);
+  protected baseSiteService = inject(BaseSiteService);
+  protected winRef = inject(WindowRef);
+  protected interval: any;
+  private readonly trendingSearchesSource = new ReplaySubject<
+    SearchPhrases[]
+  >();
+  private trendingSearches$ = this.trendingSearchesSource.asObservable();
```
rmch91: `major:` let's make this field `protected` for easier extensibility and customisation

rmch91 (changes requested):

integration-libs/cds/src/trending-searches/trending-searches.component.ts
```diff
@@ -0,0 +1,57 @@
...
+import { TrendingSearchesService } from './trending-searches.service';
+import { map, switchMap, tap } from 'rxjs/operators';
+import { EMPTY, Observable } from 'rxjs';
+import { SearchBoxOutletTrendingSearches, SearchPhrases } from './model';
+
+const MAX_TRENDING_SEARCHES = 5;
+
+@Component({
+  selector: 'cx-trending-searches',
+  templateUrl: './trending-searches.component.html',
```
rmch91, line 26: Can we add changeDetection: ChangeDetectionStrategy.OnPush?

rmch91 (changes requested):

rmch91 (approved):

PR #18740, opened 2024-04-18
fix: (CXSPA-939) - Buttons should be triggered by space and enter

sdrozdsap:


feature-libs/storefinder/components/store-finder-stores-count/store-finder-stores-count.component.ts
```diff
@@ -16,11 +17,19 @@ export class StoreFinderStoresCountComponent implements OnInit {
   locations$: Observable<any>;
   isLoading$: Observable<boolean>;
 
-  constructor(private storeFinderService: StoreFinderService) {}
+  constructor(
+    private storeFinderService: StoreFinderService,
+    private routingService: RoutingService
```
Zeyber: Two points here:
1. We prefer to use `protected` instead of `private` because protected can still provide visibility to anybody using particular methods and classes in the library so they can extend those methods if they really want to but its clear that it is unintended (not meant to be public facing). With private, it is impossible and users need to modify a local copy of our libraries in order to make changes which is not a great experience.

Yes, we have used `private` in the past in many places such as in this constructor. But we shouldn't have.

2. Modifying the constructor is considered a breaking change and we used to introduce new parameters to constructor classes as `optional` and only include new features when the user passes this optional class as a mechanism to handle that change. We also had schematics to migrate users to new constructors in later versions. But fortunately, that complexity is HISTORY! And now we "inject" new classes instead of touching the constructor.

We can do this by adding the following to the top of our class:

```
  // TODO: (CX-XXXX - cleanup ticket number) Make service required in next major.
  @Optional() routingService? = inject(RoutingService, {
    optional: true,
  });
```
Note that we create a cleanup ticket to have clear instructions on making optional classes required and any other code clean up when we are introducing a new major and finally making breaking changes permanent.

Zeyber (changes requested):

feature-libs/storefinder/components/store-finder-stores-count/store-finder-stores-count.component.ts
sdrozdsap: Thanks for such helpful hints! I've applied these changes, please take a look

feature-libs/storefinder/components/store-finder-stores-count/store-finder-stores-count.component.ts
```diff
@@ -23,4 +28,9 @@ export class StoreFinderStoresCountComponent implements OnInit {
     this.locations$ = this.storeFinderService.getViewAllStoresEntities();
     this.isLoading$ = this.storeFinderService.getViewAllStoresLoading();
   }
+
+  navigateToLocation(isoCode: string, event?: Event): void {
+    event?.preventDefault();
```
Zeyber: Because of `preventDefault()` potentially blocking default `routerLink` behaviour, we surround this functionality with the condition of `this.routingService` being truthy to avoid regression.

Zeyber (changes requested):

Zeyber (approved):

developpeurweb (approved):

PR #18741, opened 2024-04-18
PoC - component logic placed within a directive 

kpawelczak:
ticket: [CXSPA-6486](https://jira.tools.sap/browse/CXSPA-6486)

DO NOT MERGE


eduardoperez2151:
Hi @kpawelczak, what is the reason of this PoC? a new way to develop and extends components?
Thanks in advance for your reply!
Eduardo.

PR #18744, opened 2024-04-19
fix: enable all existing Feature Toggles in new apps via installation schematics

Platonn:
The installation schematics now generates a list of available FeatureToggles in a customer's app.
It's to help customers to see what feature toggles are available in Spartacus.

Moreover, each feature toggle is explicitly enabled in customer's app, to help customers start with the most up-to-date configuration (without a technical debt of any feature toggles disabled). Of course, they can disable any feature toggle they don't want to use.

In particular, our installation schematics generates the list of FeatureToggles right in the `SpartacusFeaturesModule` module, in the `providers` array, in the client's app - all in one place.

For new apps, it's not a breaking change because they're creating a new app, not upgrading an existing one.

Note: we're not enabling any new feature toggles for existing apps. This PR is only about the installation schematics for new apps.

**QA steps**
1. build libs and publish to verdaccio
2. create new app `npx @angular/cli@17 new test-ng17-FeatureToggles
 --standalone=false`
3. add Spartacus  `ng add @spartacus/schematics --baseUrl="https://40.76.109.9:9002"`
4. verify the bottom of the `spartacus-features.module.ts` in the new app contains the enablement of all feature toggles properties (compare it with the the object `defaultFeatureToggles` from Spartacus repo):
```typescript
providers: [
  provideFeatureToggles:({
     <JSON object with all feature toggles here; each with value true>
  })
]
```

fixes https://jira.tools.sap/browse/CXSPA-6783


projects/schematics/src/wrapper-module/index_spec.ts
```diff
@@ -249,7 +249,9 @@ describe('Spartacus Wrapper Module Schematics: ng g @spartacus/schematics:wrappe
   });
 
   describe('Checkout and DP', () => {
-    it('Should order the imports in the wrapper and Spartacus features modules', async () => {
```
Platonn, line 252: Previously this test contained 4 assertions:
```ts
expect(spartacusFeaturesModule.print()).toMatchSnapshot();
expect(checkoutFeatureModule.print()).toMatchSnapshot();
expect(checkoutWrapperModule.print()).toMatchSnapshot();
expect(dpFeaturesModule.print()).toMatchSnapshot();
```

Now the first snapshot test is removed
```diff
- expect(spartacusFeaturesModule.print()).toMatchSnapshot();
```
And instead of it, the test's logic is now explicitly checking the contents of the SpartacusFeaturesModule - i.e. asserting whether the `DigitalPaymentsFeatureModule` is imported after the `CheckoutFeatureModule` within the `import` array of the `@NgModule`

It was done to avoid checking the whole contents of the SpartacusFeaturesModule. It's because this content is now going to change from release to release - as we introduce (and eventually remove after 12 months) the feature toggles.

The rest of 3 snapshot assertions (for other files)
```
expect(checkoutFeatureModule.print()).toMatchSnapshot();
expect(checkoutWrapperModule.print()).toMatchSnapshot();
expect(dpFeaturesModule.print()).toMatchSnapshot();
```
are now splitted to into 3 separate `it()` blocks for better readability of snapshots' descriptions.

projects/schematics/src/add-spartacus/spartacus-feature-toggles.ts
```diff
@@ -0,0 +1,94 @@
...
+  saveAndFormat(sourceFile);
+}
+
+/**
+ * Returns a provider for all existing feature toggles with value `true`.
+ */
+function createFeatureTogglesProvider(): string {
+  const {
+    defaultFeatureToggles,
+  } = require('../feature-toggles.copied-from-core-lib');
```
pawelfras: `nitpick:` I have mixed feelings about this path because, from a code reader's perspective, it's a bit magical and doesn't refer to an existing file "at this point", but I guess we can't create one const `DESTINATION` to use here and in `copy-feature-toggles.js`.

pawelfras (dismissed):
QA done. LGTM

pawelfras (approved):
👌

PR #18750, opened 2024-04-22
fix: (CXSPA-1109) - Styles of links for cart table items

sdrozdsap:
The same styles are shared between quick order, cart and wishlist tables therefore I've fixed this issue for all of those occurrences:
image

image
image

### DIFF
--- a/feature-libs/cart/base/styles/components/_cart-item-list.scss
+++ b/feature-libs/cart/base/styles/components/_cart-item-list.scss
@@ -65,6 +65,12 @@
     text-decoration: none;
     color: var(--cx-color-text);
     font-weight: var(--cx-font-weight-bold);
+
+    // TODO: (CXSPA-6903) - Remove feature flag next major release
+    @include forFeature('a11yCartItemsLinksStyles') {
+      text-decoration: underline;
+      color: var(--cx-color-primary);
+    }
   }
 
   .cx-total {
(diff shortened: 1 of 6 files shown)


feature-libs/cart/base/styles/components/_cart-item-list.scss
```diff
@@ -62,8 +62,8 @@
   }
 
   .cx-link {
-    text-decoration: none;
-    color: var(--cx-color-text);
+    text-decoration: underline;
```
Zeyber: Here we are changing existing css properties which is a breaking change. @Pio-Bar has a few PRs where he has handled this issue. [Here is one example](https://github.com/SAP/spartacus/pull/18603/files).

tldr; Here's what we need to do:


- We add our new properties wrapped in a feature toggle to overwrite old properties along with cleanup ticket. 

Example:
```
    // old properties

    // TODO: (CXSPA-XXXX) - Remove feature flag next major release
    @include forFeature('a11yFeature') {
      // new properties
    }
```

- We need to add `useFeatureStyles` to our component constructor as well to enable them. ie. `useFeatureStyles('a11yFeature');`.
- We need to add documentation about the new flag in `projects/core/src/features-config/feature-toggles/config/feature-toggles.ts`.
- We should add this flag to our feature toggle config in the `app.module.ts` to enable it by default.

Zeyber (changes requested):

feature-libs/cart/base/styles/components/_cart-item-list.scss
sdrozdsap: I've applied these changes

Zeyber (approved):

developpeurweb (approved):

PR #18752, opened 2024-04-22
feat: CXSPA-6689 one time password for login

i333055:
Call OTP Creation API on login form: https://jira.tools.sap/browse/CXSPA-6672
Login with OTP: https://jira.tools.sap/browse/CXSPA-6689

Hi, @SAP/spartacus-codeowners could you help review and approve for [change](https://github.com/SAP/spartacus/pull/18752/files#diff-f5970a8646b27e360071eded1b6a7614f2691bd271f329e371b4137cce332581R117) of projects/core module, thanks!

Hi @morganm58 , @developpeurweb could you help review the UI styles, thanks!


projects/schematics/src/shared/constants.ts
```diff
@@ -865,6 +865,7 @@ export const RESET_PASSWORD_FORM_COMPONENT = 'ResetPasswordFormComponent';
 export const LOGIN_COMPONENT = 'LoginComponent';
 export const LOGIN_MODULE = 'LoginModule';
 export const LOGIN_FORM_MODULE = 'LoginFormModule';
+export const OTP_LOGIN_FORM_MODULE = 'VerificationTokenFormModule';
```
scarai-sap: can't find usage of this const

feature-libs/user/account/assets/translations/en/userAccount.json
```diff
@@ -15,6 +15,26 @@
...
+    "verify": "Verify",
+    "back": "Back"
+  },
+  "verificationTokenDialog": {
+    "title": "Don't receive the code",
+    "noReceiveCode": "If you have not received the code, please try:",
+    "contentLine1": "1. Wait for a while as the email may be delayed.",
+    "contentLine2": "2. Check if it is in the junk mail.",
+    "contentLine3": "3. Go back and make sure your email address and password is correct.",
+    "ok": "ok"
```
scarai-sap: might be Ok or OK? Check with UA for it pls

feature-libs/user/account/root/model/otp-login.model.ts
```diff
@@ -0,0 +1,16 @@
+/*
+ * SPDX-FileCopyrightText: 2024 SAP Spartacus team <spartacus-team@sap.com>
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+export interface LoginForm {
+  purpose: string;
```
scarai-sap, line 8: why we need propose for login form? except login, do we have other proposal to use it?

feature-libs/user/account/components/verification-token-form/verification-token-dialog.component.html
```diff
@@ -0,0 +1,52 @@
+<div
+  class="cx-verification-token-dialog cx-modal-container cx-asm-dialog"
+  [cxFocus]="focusConfig"
+  (esc)="closeModal(VERIFICATION_TOKEN_DIALOG_ACTION.OK)"
+>
+  <div class="cx-modal-content">
+    <!-- Modal Header -->
+    <div class="cx-modal-content">
+      <!-- Modal Header -->
```
scarai-sap: why need two layer of Modal header?

feature-libs/user/account/components/verification-token-form/verification-token-form-component.service.ts
```diff
@@ -0,0 +1,92 @@
...
+      {
+        key: 'verificationTokenForm.sentOTP',
+        params: { target },
+      },
+      GlobalMessageType.MSG_TYPE_CONFIRMATION,
+      10000
+    );
+  }
+
+  sentOTP(loginId: string, password: string, purpose: string) {
```
scarai-sap: should be send? send/sent/sent

Maybe use createToken is more than send?

feature-libs/user/account/components/verification-token-form/verification-token-form-component.service.ts
```diff
@@ -0,0 +1,92 @@
...
+  });
+
+  login() {
+    if (!this.form.valid) {
+      this.form.markAllAsTouched();
+      return;
+    }
+    this.busy$.next(true);
+
+    from(
```
scarai-sap, line 51: Wht you need from here? Code as below is enough?

this.auth.otpLoginWithCredentials(
  this.form.value.tokenId,
  this.form.value.tokenCode
).pipe(
  withLatestFrom(this.auth.isUserLoggedIn()),
  tap(([_, isLoggedIn]) => this.onSuccess(isLoggedIn))
);

feature-libs/user/account/components/verification-token-form/verification-token-form-component.service.ts
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  displayMessage(target: string) {
+    this.globalMessage.add(
+      {
+        key: 'verificationTokenForm.sentOTP',
+        params: { target },
+      },
+      GlobalMessageType.MSG_TYPE_CONFIRMATION,
+      10000
```
scarai-sap: could you move the 10000 to the top of the file with a const?

feature-libs/user/account/components/verification-token-form/verification-token-form.component.html
```diff
@@ -0,0 +1,60 @@
...
+        'verificationTokenForm.verificationCode.placeholder' | cxTranslate
+      }}"
+    />
+    <cx-form-errors [control]="form.get('tokenCode')"></cx-form-errors>
+  </label>
+
+  <input type="hidden" formControlName="tokenId" [(ngModel)]="tokenId" />
+
+  <div class="row resend-link-text">
+    <div class="col-md-6 col-lg-6 left-text">
```
scarai-sap: New component with boost css is not allowed

feature-libs/user/account/styles/_verification-token-dialog.scss
```diff
@@ -0,0 +1,62 @@
...
+    padding-bottom: 0.9rem;
+    padding-inline-start: 1rem;
+    border-width: 0;
+    display: flex;
+
+    box-shadow: 0px 0px 4px rgba(85, 107, 130, 0.16), inset 0px -1px 0px #d9d9d9;
+
+    .info-icon {
+      cx-icon {
+        font-size: 1.4rem;
```
scarai-sap, line 19: use @include type() mixin for adding predefined fonts.

projects/storefrontapp-e2e-cypress/cypress/e2e/regression/user_access/otp-login.e2e-flaky.cy.ts
```diff
@@ -0,0 +1,131 @@
...
+    describe('Create OTP', () => {
+      beforeEach(() => {
+        cy.visit('/login');
+      });
+
+      it('should be able to create a new OTP by customer click Sign In button (CXSPA-6672)', () => {
+        const user = login.registerUserFromLoginPage();
+        listenForCreateVerificationToken();
+
+        cy.log(`🛒 Logging in user ${user.email} from the login form`);
```
i53577, line 29: here is a cute icon, interesting

projects/storefrontapp-e2e-cypress/cypress/e2e/regression/user_access/otp-login.e2e-flaky.cy.ts
```diff
@@ -0,0 +1,131 @@
...
+            .should('eq', 200);
+        });
+      });
+    });
+
+    describe('Failed to Create OTP', () => {
+      beforeEach(() => {
+        cy.visit('/login');
+      });
+      it('should be not able to create OTP with invalid user data (CXSPA-6672)', () => {
```
i53577:
```suggestion
      it('should not be able to create OTP with invalid user data (CXSPA-6672)', () => {
```

feature-libs/user/account/styles/_verification-token-dialog.scss
```diff
@@ -0,0 +1,62 @@
+%cx-verification-token-dialog {
+  .cx-modal-content {
+    max-width: 32rem;
+    margin-inline-start: auto;
```
scarai-sap, line 4: I guess the default value is auto?

feature-libs/user/account/styles/_verification-token-form.scss
```diff
@@ -0,0 +1,45 @@
...
+      }
+
+      .right-text {
+        padding: 0;
+        text-align: var(--cx-text-align, right);
+      }
+
+      a.disabled-link {
+        pointer-events: none;
+        color: #a7cce8;
```
scarai-sap: don't define color

feature-libs/user/account/styles/_verification-token-form.scss
```diff
@@ -0,0 +1,45 @@
...
+        text-decoration: none;
+      }
+
+      a {
+        font-size: 1rem;
+        font-weight: 400;
+        color: #1672b7;
```
scarai-sap: don't define color and font size/weight/...

projects/storefrontapp-e2e-cypress/cypress/e2e/regression/user_access/otp-login.e2e-flaky.cy.ts
```diff
@@ -0,0 +1,131 @@
...
+          .its('response.statusCode')
+          .should('eq', 400);
+
+        cy.get('cx-global-message').within(() => {
+          cy.get('span').contains('Email is not valid.');
+        });
+      });
+    });
+
+    describe('Chould can return back to login page from verify code page', () => {
```
i53577: I guess you want to say this
```suggestion
    describe('Should go back to login page when click back button', () => {
```

projects/storefrontapp-e2e-cypress/cypress/e2e/regression/user_access/otp-login.e2e-flaky.cy.ts
```diff
@@ -0,0 +1,131 @@
...
+      it('should be not able to create OTP with invalid user data (CXSPA-6672)', () => {
```
i53577: not related to this case

projects/storefrontapp-e2e-cypress/cypress/e2e/regression/user_access/otp-login.e2e-flaky.cy.ts
```diff
@@ -0,0 +1,131 @@
...
+        cy.visit('/login/verify-token');
+
+        cy.get('cx-verification-token-form').should('exist');
+
+        cy.get('cx-verification-token-form form').within(() => {
+          cy.get('.btn-secondary').click();
```
i53577: I guess this is the `back` button on verification otp page, right?

projects/assets/src/translations/en/common.json
```diff
@@ -132,7 +132,8 @@
         "password": "Password entered is not valid.",
         "uid": "UID is not valid.",
         "code": "Code is not valid.",
-        "email": "Email is not valid."
+        "email": "Email is not valid.",
+        "loginId": "Email is not valid."
```
i53577, line 136: why need to add a different key with same value?

scarai-sap, line 136: why loginId is same text?

scarai-sap (commented):
Comments added.

Please avoid define color, avoid using bootstrap css definition

feature-libs/user/account/assets/translations/en/userAccount.json
i333055: it's Ok and keep consistence with other button.

projects/storefrontapp-e2e-cypress/cypress/e2e/regression/user_access/otp-login.e2e-flaky.cy.ts
niehuayang: hhh, copy from old login component test

projects/assets/src/translations/en/common.json
niehuayang: The API verificationToken's request body is like:
`
{
  "purpose": "LOGIN",
  "**loginId**": "customer.test@sap.com",
  "password": "Password123."
}
` 
if the loginId is not valid, throw a global error message with subject loginId instad of email

feature-libs/user/account/root/model/otp-login.model.ts
niehuayang: we use it on otp-login-form component, account.core, account.occ and account.root

projects/assets/src/translations/en/common.json
niehuayang: The API verificationToken's request body is like:
`{ "purpose": "LOGIN", "**loginId**": "customer.test@sap.com", "password": "Password123." }`
if the loginId is not valid, throw a global error message with subject loginId instad of emai

projects/storefrontapp-e2e-cypress/cypress/e2e/regression/user_access/otp-login.e2e-flaky.cy.ts
niehuayang: changed

projects/storefrontapp-e2e-cypress/cypress/e2e/regression/user_access/otp-login.e2e-flaky.cy.ts
niehuayang: changed

projects/storefrontapp-e2e-cypress/cypress/e2e/regression/user_access/otp-login.e2e-flaky.cy.ts
niehuayang: it is a  <a>  button, have changed the key

feature-libs/user/account/components/verification-token-form/verification-token-form-component.service.ts
i333055: otpLoginWithCredentials return the promise，from is used to convert Promise to Observable

feature-libs/user/account/assets/translations/en/userAccount.json
scarai-sap: OK, but looks you use. "ok" in here

feature-libs/user/account/root/model/otp-login.model.ts
scarai-sap: I wanna to say, it should hardcode in somewhere as hte loginForm only used for login

feature-libs/user/account/components/verification-token-form/verification-token-form.component.html
i333055: fixed

feature-libs/user/account/styles/_verification-token-dialog.scss
i333055: default is 0，this is used to make dialog in the center of the screen

i333055: default is 0，this is used to make dialog in the center of the screen

feature-libs/user/account/styles/_verification-token-dialog.scss
i333055: for icon can not used  @include type() for the icon is a specific font and will fix others.

feature-libs/user/account/root/model/otp-login.model.ts
niehuayang: Ah, I got, it seems that here is a **name**  misunderstand, LoginForm model is used as payload to call VerificationToken API, instead of otp login html, I have renamed as **VerificationTokenCreation** model to have a correct understanding.  This is a good point, thanks.

feature-libs/user/account/components/otp-login-form/otp-login-form.component.ts
```diff
@@ -0,0 +1,109 @@
...
+  protected onCreateVerificationTokenComplete(): void {
+    this.form.reset();
+    this.busy$.next(false);
+  }
+
+  protected collectDataFromLoginForm(): VerificationTokenCreation {
+    return {
+      // TODO: consider dropping toLowerCase as this should not be part of the UI,
+      // as it's too opinionated and doesn't work with other AUTH services
+      loginId: this.form.value.userId.toLowerCase(),
```
scarai-sap, line 107: What will happen if we remove toLowerCase now?

feature-libs/user/account/components/verification-token-form/verification-token-form.component.html
scarai-sap: col-md-6 col-lg-6 are boots.

feature-libs/user/account/components/verification-token-form/verification-token-form.component.html
```diff
@@ -0,0 +1,66 @@
...
+        </a>
+        <ng-container *ngIf="isResendDisabled">
+          {{
+            'verificationTokenForm.sendRateLime'
+              | cxTranslate: { waitTime: waitTime }
+          }}
+        </ng-container>
+      </span>
+    </div>
+    <div class="col-md-6 col-lg-6 right-text">
```
scarai-sap: boots

feature-libs/user/account/core/connectors/user-account.connector.ts
```diff
@@ -16,4 +20,9 @@ export class UserAccountConnector {
   get(userId: string): Observable<User> {
     return this.adapter.load(userId);
   }
+  createVerificationToken(
```
scarai-sap, line 24: add empty line between methods

feature-libs/user/account/core/facade/facade-providers.ts
```diff
@@ -4,14 +4,23 @@
...
+import { UserAccountService } from './user-account.service';
+import { VerificationTokenService } from './verification-token.service';
 
 export const facadeProviders: Provider[] = [
   UserAccountService,
   {
     provide: UserAccountFacade,
     useExisting: UserAccountService,
   },
+  VerificationTokenService,
```
scarai-sap, line 21: add empty line between methods

feature-libs/user/account/styles/_verification-token-dialog.scss
```diff
@@ -0,0 +1,62 @@
...
+
+  .cx-dialog-header {
+    padding-top: 0.9rem;
+    padding-inline-end: 1rem;
+    padding-bottom: 0.9rem;
+    padding-inline-start: 1rem;
+    border-width: 0;
+    display: flex;
+
+    box-shadow: 0px 0px 4px rgba(85, 107, 130, 0.16), inset 0px -1px 0px #d9d9d9;
```
scarai-sap: Worry for new colors....

scarai-sap (commented):
some small comments added

feature-libs/user/account/components/otp-login-form/otp-login-form.component.ts
i333055: this loginId is email, by default in database is lowercase

feature-libs/user/account/core/facade/facade-providers.ts
i333055: it's in array seems not need add line.

feature-libs/user/account/components/verification-token-form/verification-token-form.component.html
i333055: forget save...

feature-libs/user/account/components/otp-login-form/otp-login-form.component.ts
scarai-sap: It's better let backend(OCC API) to handle it?

feature-libs/user/account/core/facade/facade-providers.ts
scarai-sap: you are right

feature-libs/user/account/components/otp-login-form/otp-login-form.component.ts
niehuayang: maybe here is some strange resaon, I only keep same with old code: https://github.com/SAP/spartacus/blob/8c27c8a72b81535e02694a98a283fe98688a1265/feature-libs/user/account/components/login-form/login-form-component.service.ts#L63

scarai-sap (dismissed):
LGTM

i53577 (dismissed):

feature-libs/user/account/styles/_verification-token-form.scss
```diff
@@ -0,0 +1,45 @@
...
+    .resend-link-text {
+      display: flex;
+      flex-direction: row;
+      width: 100%;
+      margin: auto;
+
+      .left-text {
+        padding: 0;
+        width: 50%;
+        text-align: var(--cx-text-align, left);
```
developpeurweb: Please remove all `--cx-text-align` variables. You can use standard text-align values. We are trying to clean up deprecated variables.

developpeurweb (changes requested):

feature-libs/user/account/styles/_verification-token-form.scss
i333055: fixed

i53577 (dismissed):

feature-libs/user/account/components/verification-token-form/verification-token-dialog.component.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+})
+class MockTranslatePipe implements PipeTransform {
+  transform(): any {}
+}
+
+class MockLaunchDialogService implements Partial<LaunchDialogService> {
+  closeDialog(_reason: any) {}
+}
+
+describe('AsmBindCartDialogComponent', () => {
```
Matejk00: I think it should be `VerificationTokenDialogComponent` instead of  `AsmBindCartDialogComponent`.

feature-libs/user/account/components/verification-token-form/verification-token-dialog.component.ts
```diff
@@ -0,0 +1,33 @@
+/*
+ * SPDX-FileCopyrightText: 2024 SAP Spartacus team <spartacus-team@sap.com>
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { Component } from '@angular/core';
+import { FocusConfig, LaunchDialogService } from '@spartacus/storefront';
+
+export enum VERIFICATION_TOKEN_DIALOG_ACTION {
```
Matejk00: Can we move this enum outside the component file? It should be part of the model file.

feature-libs/user/account/components/verification-token-form/verification-token-form.component.ts
```diff
@@ -0,0 +1,111 @@
...
+@Component({
+  selector: 'cx-verification-token-form',
+  templateUrl: './verification-token-form.component.html',
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class VerificationTokenFormComponent implements OnInit {
+  constructor(
+    protected service: VerificationTokenFormComponentService,
+    protected launchDialogService: LaunchDialogService,
+    private cdr: ChangeDetectorRef
```
Matejk00: Any reason why we need `cdr` to be a `private` instead of the `protected`?

Matejk00 (changes requested):

feature-libs/user/account/components/verification-token-form/verification-token-dialog.component.spec.ts
i333055: fixed

feature-libs/user/account/components/verification-token-form/verification-token-dialog.component.ts
i333055: fixed

feature-libs/user/account/components/verification-token-form/verification-token-form.component.ts
i333055: fixed

i53577 (dismissed):

projects/storefrontapp-e2e-cypress/cypress/e2e/regression/user_access/otp-login.e2e-flaky.cy.ts
```diff
@@ -0,0 +1,131 @@
...
+        cy.get('cx-verification-token-form').should('exist');
+        cy.get('cx-verification-token-form').should('be.visible');
+
+        cy.log('The email being sent is asynchronous, so waiting 10s');
+        cy.wait(10000);
+
+        cy.request({
+          method: 'GET',
+          url:
+            'http://mail-ccv2.westeurope.azurecontainer.io:8025/api/v2/search?query=' +
```
Matejk00: Is it safe to store this URL directly in our codebase? Perhaps it would be worthwhile to consider moving the URL to `cypress.config.ts` and setting it as an environment variable instead.

feature-libs/user/account/components/verification-token-form/verification-token-form-component.service.ts
```diff
@@ -0,0 +1,93 @@
...
+} from '@spartacus/core';
+import { VerificationTokenFacade } from '@spartacus/user/account/root';
+import { BehaviorSubject, from } from 'rxjs';
+import { tap, withLatestFrom } from 'rxjs/operators';
+
+const globalMsgShowTime: number = 10000;
+@Injectable()
+export class VerificationTokenFormComponentService {
+  constructor(
+    protected auth: AuthService,
```
Matejk00: Our recommendation is to use `inject` (https://angular.io/api/core/inject) method instead of passing services into constructor params.
Example:
```
protected auth = inject(AuthService);
```


Can you refactor your code and use such convention?

feature-libs/user/account/components/verification-token-form/verification-token-form.component.html
```diff
@@ -0,0 +1,66 @@
...
+          {{
+            'verificationTokenForm.sendRateLime'
+              | cxTranslate: { waitTime: waitTime }
+          }}
+        </ng-container>
+      </span>
+    </div>
+    <div class="right-text">
+      <a
+        href="javascript:void(0);"
```
Matejk00: Why do we need `javascript:void(0);` as a href attribute?

Matejk00 (changes requested):

developpeurweb:
Please review the following items that are quite likely to come up in future Accessibility Audits:

1. There’s currently a refactoring going on for some UI elements in ticket [CXSPA-1035](https://jira.tools.sap/browse/CXSPA-1035) reported during a previous Accessibility Audit about buttons looking like links and vice-versa. You have similar issues on your verification page. As per screenshot below, the Back “button” although it is a valid link because it takes the user to another page, in ticket CXSPA-1035, auditors don’t like links that look like buttons, therefore we are refactoring all that. Please use `role=“button”` so screen reader announces it as a button and also add button keyboard interactions.

real link but auditor wants button
common keyboard interactions


2. When the timer button is dimmed, it’s still accessible via keyboard-only navigation (with and without a screen reader) and also possible to active it and send a new code while the timer is running. Interaction should be equal for all users (mouse, keyboard, screen reader, touch) elements that are dimmed or disabled should be the same for all users.

dimmed button still triggers with keyboard-only


3. Having a timer requires you to comply with requirement **ACC-278**. My suggestion would be to assist the user with `role="timer" aria-live="polite" aria-atomic="true"` in that `<span>`… also using the word “seconds” will improve clarity, instead of “s”. Confirm this with your UX designer, not sure what’s the intent here, but there’s not enough context for a screen reader user for example. It feels the whole Accessibility layer is missing, perhaps annotations were omitted. Maybe a phrase informing why a timer has been triggered and what happens after the limit will provide context for such users. You can do this using `aria-describedby` like in the password registration field which provides more instructions for the user.

role timer atomic
aria-describedby hint


4. These should be action-links buttons and NOT links. If not possible to convert, use aria-role=“button” and add standard keyboard interactions for buttons. Right now they only respond to the “Enter” key, and all events happen on the same page, so they don’t link other places.

announced as links
button announced as link 1
button announced as link 2
common keyboard interactions


5. There’s an AMP bug of level 6 severity, my suggestion is to change the “Ok” text button to “Close”, because that’s what the button does, but check this with your UX designer if that's the intent. 

AMP bug


6. Button should match dialog title. Past vs present tense (Didn’t vs Don’t).

button and modal title don't match

feature-libs/user/account/components/verification-token-form/verification-token-form.component.html
i333055: Hi @Matejk00 , this is used for Accessibility, when user input tab, the link can be focused.

projects/storefrontapp-e2e-cypress/cypress/e2e/regression/user_access/otp-login.e2e-flaky.cy.ts
niehuayang: fixed

feature-libs/user/account/components/verification-token-form/verification-token-form-component.service.ts
niehuayang: fixed

projects/storefrontapp-e2e-cypress/cypress.config.ci.ts
```diff
@@ -29,6 +29,8 @@ export default defineConfig({
     OCC_PREFIX: '/occ/v2',
     OCC_PREFIX_USER_ENDPOINT: 'users',
     OCC_PREFIX_ORDER_ENDPOINT: 'orders',
+    mail_CCV2_URL: 'http://mail-ccv2.westeurope.azurecontainer.io:8025',
+    mail_CCV2_PREFIX: '/api/v2',
```
Matejk00: Sorry for nitpicking. Can we use capitalise letters for such properties to unify with the rest?

projects/storefrontapp-e2e-cypress/cypress.config.ts
```diff
@@ -23,6 +23,8 @@ export default defineConfig({
     OCC_PREFIX: '/occ/v2',
     OCC_PREFIX_USER_ENDPOINT: 'users',
     OCC_PREFIX_ORDER_ENDPOINT: 'orders',
+    mail_CCV2_URL: 'http://mail-ccv2.westeurope.azurecontainer.io:8025',
+    mail_CCV2_PREFIX: '/api/v2',
```
Matejk00: The same here.

Matejk00 (changes requested):

projects/storefrontapp-e2e-cypress/cypress.config.ci.ts
niehuayang: don't worry, it's my fault.....    have changed

i53577 (approved):
LGTM

Matejk00 (approved):

developpeurweb:
I'll approve but you still have very basic Accessibility issues that will come up in audits for sure, please create a ticket for each issue.
1. You have to complement all `role="button"` elements with full keyboard functionality as if they were real buttons. Both `Enter` and `Spacebar` should work for buttons. Right now only one of them works.
2. To comply with requirement **ACC-278** you still need to provide context for the verification field using `aria-describedby` as shown previously in the password registration example. It could be something very simple, such as: _"You can request a new code by activating the button below. A timer will start indicating the seconds left until you can request a new one"._
3. When using keyboard-only navigation and focusing on the "Resend" button, the focus gets trapped and I can keep pressing `Spacebar` and reset the timer every time. Same thing when using a Screen Reader on read mode, I can still access the button while it's supposed to be disabled. This is an Accessibility issue, but might as well be a Security issue too. My suggestion is to disable the button on activation and remove focus from element so it's not reachable with keyboard and add `aria-disabled` so screen readers will ignore it until it becomes focusable again. Tagging @giancorderoortiz to follow up on Security concerns.

developpeurweb (approved):

PR #18753, opened 2024-04-22
feat: CXSPA-6578 Introduce PROPAGATE_SERVER_ERROR_RESPONSE and default ExpressJS handlers

pawelfras:
This PR introduces the `PROPAGATE_SERVER_ERROR_RESPONSE` injection token which helps to propagate errors that occur during server-side rendering. This covers Angular's loophole in dealing with errors that occur during SSR - by default, Angular returns HTML with a status of 200, regardless of the errors that occur, which can malformed the output.
Thanks to this propagator, we can catch errors in the wrapper for CommonEngine responsible for rendering and return them instead of the malformed rendered view.
Together with this PR, we provide `defaultServerErrorResponseHandlers` function for ExpressJS, which helps to deal with caught errors:
- falls back to the CSR with the status 404:
image
- falls back to the CSR with the status 500:
image

Customer can also provide their handlers and use them inside the `server.js` file instead of the ones we provided. 

QA steps:

1) Tests in storefront app:
  - modify `pages` property in `default-cms-config.ts` to generate `CmsPageNotFoundServerError`
  - modify `components` property in `default-cms-config.ts` to generate `UnknownServerError`
- run npm `run dev:ssr`
- verify whether the app fell back to CSR with the proper status code:
 - `index.html` returned with status `404` if `pages` endpoint has been modified in `default-cms-config.ts`
 - `index.html` returned with status `500` if `components` endpoint has been modified in `default-cms-config.ts`
 
 2) Tests in customer's app:
```npx ts-node ./tools/schematics/testing.ts in the SPA root folder```
```npx @angular/cli new my-app --standalone=false --ssr=false --style=scss --routing=false```
```npx @angular/cli add @spartacus/schematics@latest --ssr```
- use HTTP proxy or modify config for CMS endpoints to cause an error when rendering:
```ts
provideConfig(<OccConfig>{
      backend: {
        occ: {
          endpoints: {
            // components: 'cms/components2', FOR STATUS 500
            pages: 'cms/pages2', // FOR STATUS 404
          },
        },
      },
    }),
```
- in `server.ts` inside `app` function use handlers provided by Spartacus:
```ts 

  /** */

```
- run `npm run build` and `npm run serve:ssr:my-app` to verify if server error responses have been handled as in test case 1
- create a custom handler for errors (you can use the below example):
  - ```npm i -d @types/ejs ejs```
  - handle-page-not-found.ts:
  ```ts
  import { ErrorRequestHandler } from 'express';
  import ejs from 'ejs';
  import { CmsPageNotFoundServerErrorResponse } from '@spartacus/setup/ssr';

  export const handlePageNotFound: ErrorRequestHandler = async (
    err,
    _req,
    res,
    next,
  ) => {
    if (err instanceof CmsPageNotFoundServerErrorResponse) {
      const html = await ejs.renderFile('./404.ejs', {
        title: 'OPS! Page not found',
        text: 'The page you are looking for does not exist',
      });
      res.status(404).send(html);
    } else {
      next(err);
    }
  };
  ```
  - `404.ejs`:
  ```
  <!DOCTYPE HTML>
  <html lang="en">
    <head>
      <meta charset="UTF-8" />
      <meta http-equiv="X-UA-Compatible" content="IE=edge" />
      <meta name="viewport" content="width=device-width, initial-scale=1.0" />
      <title>Page Not Found</title>
    </head>
    <body>
      <h1><%= title %></h1>
      <p><%= text %></p>
    </body>
  </html>
  ```
  - in `server.ts`, use handler inside `app` function:
  ```ts 
   server.use(handlePageNotFound)
  ```
- run `npm run build` and `npm run serve:ssr:my-app` to verify if the custom handler works
 

closes [CXSPA-6578](https://jira.tools.sap/browse/CXSPA-6578)


projects/storefrontapp/server.ts
```diff
@@ -63,6 +66,9 @@ export function app(): express.Express {
     });
   });
 
+  server.use(handleCmsPageNotFoundErrorResponse(indexHtmlContent));
+  server.use(handleUnknownServerErrorResponse(indexHtmlContent));
+
```
Platonn: Strategic question: how do we plan to distribute our error handlers to customers?


- will we be adding them to their server.ts with schematics?
  - if yes, do we plan to put them one by one or somehow wrap them in one uber-error-handler, a.k.a. server.use(defaultServerErrorResponseHandlers()) // import defaultServerErrorResponseHandlers from @spartacus/setup/ssr?
    - thanks to this, over time we will be able to introduce changes/additions if needed - in existing apps. For example, we can easily add a temporaryRedirectErrorHandler (sending 302) if we really want to

projects/storefrontapp/server.ts
```diff
@@ -33,6 +35,7 @@ export function app(): express.Express {
   const indexHtml = existsSync(join(distFolder, 'index.original.html'))
     ? join(distFolder, 'index.original.html')
     : join(distFolder, 'index.html');
+  const indexHtmlContent = readFileSync(indexHtml, 'utf-8');
```
Platonn, line 38: I suggest to do this job in our uber-error-handler, so customers would give us onlu the indexHtml path:

```
server.use(defaultServerErrorResponseHandlers({indexHtml}))
```
(perhaps the long name is yet up to discussion)

Note: I suggest named params inside an object to allow for easisly adding new params in the future potentially

core-libs/setup/ssr/engine/cx-common-engine-wrapper.ts
```diff
@@ -0,0 +1,54 @@
...
+ */
+
+import { StaticProvider } from '@angular/core';
+import { CommonEngine } from '@angular/ssr';
+import { CxServerErrorResponse } from '../error-handling/server-error-response';
+import { PROPAGATE_SERVER_ERROR_RESPONSE } from '../error-handling/server-error-response/propagate-server-error-response';
+import { RenderOptions } from './ng-express-engine';
+
+/**
+ * Wraps the CommonEngine render method to handle propagated server responses caught during server-side rendering.
```
Platonn, line 14:
```suggestion
 * Wraps the CommonEngine render method to handle propagated server (error) responses caught during server-side rendering.
```

core-libs/setup/ssr/engine/cx-common-engine-wrapper.ts
```diff
@@ -0,0 +1,54 @@
...
+ */
+
+export function cxCommonEngineWrapper(
+  engine: CommonEngine,
+  renderOptions: RenderOptions,
+  callback: (err?: Error | null, html?: string) => void
+) {
```
Platonn, line 21: Let's add a return type

Platonn: IMHO this method should not take `callback` as a parameter. IMHO we should build a wrapper ONLY over the `CommonEngine.render` method, but the `callback` interface is an implementation detail of the caller who is the `ngExpressEngine`.

Please note, that theoretically in the future we might get rid of the `ngExpressEngine` (e.g. if we go serverless), but what will last is our wrapper over the `CommonEngine.render` method which is aware of ServerErrorResponses propagation


Instead of exposing a function in our public api `cxCommonEngineWrapper`, what do you think about creating a child class of CommonEngine, named e.g. `CxCommonEngine`. In the child class we could override the method `render()` inside which we could call the `super.render()`.

usage: in ngExpressEngine:
```ts
const commonEngine = new CxCommonEngine(commonEngineOptions); // instead of the CommonEngine
/*...*/
commonEngine.render(renderOptions)
  .then(html => callback(null, html)
  .catch(callback)
```

Please note that such an approach allows also for using our child class also in non-callback-ish environment, e.g. in a pure `server.ts`

```ts
const commonEngine = new CxCommonEngine(commonEngineOptions);
/* ... */
       server.get('*', (req, res, next) => {
          commonEngine
            .render({
              bootstrap: AppServerModule,
              documentFilePath: indexHtml,
              url: req.originalUrl,
              publicPath: distFolder,
            })
            .then((html) => res.send(html))
            .catch((err) => next(err));
        });
```

core-libs/setup/ssr/error-handling/server-error-response/propagate-server-error-response.ts
```diff
@@ -0,0 +1,17 @@
...
+ * SPDX-FileCopyrightText: 2024 SAP Spartacus team <spartacus-team@sap.com>
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { InjectionToken } from '@angular/core';
+import { CxServerErrorResponse } from './cx-server-error-response';
+
+/**
+ * Injection token used to provide a function that will propagate the server error response occurred during server side rendering to the CommonEngine wrapper.
```
Platonn: 1. 
```suggestion
 * Function that will propagate the server error response object
 * created during server side rendering to the CommonEngine wrapper.
```

2. `CommonEngine wrapper` vs `CxCommonEngine`? - as mentioned in a comment above

core-libs/setup/ssr/engine/cx-common-engine-wrapper.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  return engine
+    .render({
+      ...renderOptions,
+      providers: [
+        propagateServerErrorResponseProvider,
+        ...(renderOptions.providers ?? []),
+      ] as StaticProvider[],
+    })
+    .then((html) => propagatedServerErrorResponses.push(html))
```
Platonn: nitpick: 👀 html is not a server **error** response

core-libs/setup/ssr/engine/cx-common-engine-wrapper.ts
```diff
@@ -0,0 +1,54 @@
...
+/**
+ * Wraps the CommonEngine render method to handle propagated server responses caught during server-side rendering.
+ */
+
+export function cxCommonEngineWrapper(
+  engine: CommonEngine,
+  renderOptions: RenderOptions,
+  callback: (err?: Error | null, html?: string) => void
+) {
+  const propagatedServerErrorResponses: (CxServerErrorResponse | string)[] = [];
```
Platonn: We don't need to store an array (with mixed errors and string htmls), but we need to store just one potential element `errorResponse`.

What about simply having a single variable instead of array?


1. define variable `errorResponse`
```diff
- const propagatedServerErrorResponses: (CxServerErrorResponse | string)[] = [];
+ let errorResponse: undefined | CxServerResponse;
```

2. store in this variable `errorResponse` only the first propagated server error response
```diff
-       return (serverErrorResponse: CxServerErrorResponse) => {
-         propagatedServerErrorResponses.push(serverErrorResponse);
-      };
-       return (serverErrorResponse: CxServerErrorResponse) => {
-         errorResponse ??= serverErrorResponse
-      };
```

3. when rendering completes, check if the variable `errorResponse` stores any avlue - then reject Promise with this value . Otherwise resolve with the `html`
```diff
- .then((html) => propagatedServerErrorResponses.push(html))
+ .then((html) => {
+    if (errorResponse) { 
+      throw errorResponse;
+    }
+    return html
+  })
``` 

4. No need to catch errors (APP_INITIALIZERs only). If Angular rejects the promise, we should let this rejection pass through
```diff
- .catch((err) => propagatedServerErrorResponses.push(err))
``` 

5.  No need for `finally` then IMHO. We've already handled 2 cases:
- on rendering success, we'll either resolve Promise with the produced html or reject promise with the stored `errorResponse` (if it's defined)
- on rendering error (i.e.APP_INITIALIZER only), we've already let the error pass through
```diff
- .finally(/* ... */)
```

Please let me know what do you think about it.

core-libs/setup/ssr/engine/cx-common-engine-wrapper.ts
```diff
@@ -0,0 +1,54 @@
...
+    .finally(async () => {
+      const serverResponse = propagatedServerErrorResponses[0];
+      // If the first event is a CxServerErrorResponse, we call the callback with the error.
+      // which will be handled by the error handler in the ExpressJS server.
+      // Otherwise, we call the callback with the rendered html.
+      if (serverResponse instanceof CxServerErrorResponse) {
+        callback(serverResponse, undefined);
+      } else {
+        callback(undefined, serverResponse);
+      }
```
Platonn: nitpick: as far as I understand, the logic is incorrect in case of Errors from APP_INITAILIZERs. Any Error from APP_INITIALIZER is not an instance of CxServerErrorResponse, therefore in line 51 such an Error will be populated via callback as a success parameter. 

Btw. in the comment above I proposed a simplified approach (with a single variable `errorResponse` instead of arrays) which by the way gets rid of this incorrect behavior.

⚠️ Anyways, let's please fix the unit test `'should handle APP_INITIALIZER errors the standard Angular way`. The snapshot of the function call should have the Error object at the first parameter, not second. See [lines 5-7](https://github.com/SAP/spartacus/pull/18753/files?diff=split&w=1#diff-9da85599fc552d5b25ee84670c0544e118f3f62cea8c438ec6ecddc7d351eb89R5-R7) of the file `core-libs/setup/ssr/engine/__snapshots__/cx-common-engine-wrapper.spec.ts.snap`

core-libs/setup/ssr/error-handling/multi-error-handlers/server-responding-error-handler.ts
```diff
@@ -18,13 +19,16 @@ import { SERVER_ERROR_RESPONSE_FACTORY } from '../server-error-response-factory'
   providedIn: 'root',
 })
 export class ServerRespondingErrorHandler implements MultiErrorHandler {
-  protected factories = inject(SERVER_ERROR_RESPONSE_FACTORY);
+  protected transformers = inject(SERVER_ERROR_RESPONSE_FACTORY);
```
Platonn: `transformers` vs. `serverErrorResponseFactories`?

core-libs/setup/ssr/error-handling/multi-error-handlers/server-responding-error-handler.ts
```diff
@@ -18,13 +19,16 @@ import { SERVER_ERROR_RESPONSE_FACTORY } from '../server-error-response-factory'
...
-    //@ts-ignore
-    //TODO:CXSPA-6577 Propagate the error to the OptimizedSsrEngine
-    const cxServerErrorResponse = resolveApplicable(this.factories, [
-      error,
-    ]).create(error);
+    const cxServerError = resolveApplicable(this.transformers, [error])?.create(
+      error
+    );
+
+    cxServerError && this.propagateServerErrorResponse(cxServerError);
```
Platonn: nitpicks:
- cxServerError vs cxServerErrorResponse?
- inline `if` trick (`&& `) vs a regular `if(...) { ... }`? IMHO the latter - for readability

core-libs/setup/ssr/error-handling/server-error-response-factory/cms-page-not-found-server-error-response-factory.spec.ts
```diff
@@ -38,23 +38,34 @@ describe('CmsPageNotFoundServerErrorResponse', () => {
     expect(cmsPageNotFoundServerErrorResponse.getPriority()).toBe(Priority.LOW);
   });
 
-  it('should match if error is an instance of HttpErrorResponse and URL starts with proper OCC endpoint including /cms/pages', () => {
+  it('should match if error is an instance of HttpErrorResponse and, status is 404 and URL starts with proper OCC endpoint including /cms/pages', () => {
```
Platonn, line 41:
```suggestion
  it('should match if: error is an instance of HttpErrorResponse and status is 404 and URL starts with proper OCC endpoint including /cms/pages', () => {
```

core-libs/setup/ssr/error-handling/server-error-response-handlers/server-error-response-handlers.ts
```diff
@@ -0,0 +1,47 @@
...
+  CmsPageNotFoundServerErrorResponse,
+  UnknownServerErrorResponse,
+} from '../server-error-response/cx-server-error-response';
+
+/**
+ * Handles CmsPageNotFoundServerErrorResponse which results in a fallback to client side rendering.
+ * The document content is sent to the client with the appropriate 404 status code.
+ *
+ * @param documentContent The document content to be sent to the client.
+ * @returns The error request handler.
```
Platonn, line 18: nitpick: let's remove `@returns`. this function doesn't return anything.

Note: the same remark applies for the next function

core-libs/setup/ssr/error-handling/server-error-response-handlers/server-error-response-handlers.ts
```diff
@@ -0,0 +1,47 @@
...
+ */
+export const handleCmsPageNotFoundErrorResponse =
+  (documentContent: string): ErrorRequestHandler =>
```
Platonn: nitpick: The name or the JSDoc of the function is a misleading. The function's name and JSDocs tell this function handles the server error response. But it's rather a getter function that returns a handler. 😅 
I'm having hard time finding a better name. `handleCmsPageNotFoundErrorResponse` vs `getCmsPageNotFoundErrorResponseHandler` (I'm not convinced 😕  this name is too long)
But at least we can improve the JSDoc

core-libs/setup/ssr/error-handling/server-error-response/propagate-server-error-response.ts
```diff
@@ -0,0 +1,17 @@
...
+ *
+ * @private
```
Platonn: I believe it should be a public api.

Why: our public class (I hope it's public 😉 ) `ServerRespondingErrorHandler` already injects this token `PROPAGATE_SERVER_ERROR_RESPONSE`, so IMHO the token should be in public API, so our customers can freely customize our `ServerRespondingErrorHandler` if they want -> then they'll need to call the propagating function themselves, but then they need to inject it themselves (so they need the token to be in public API).

PS. At the moment I'm not yet convinced that customers will really need customize the other end of the contract - our CxCommonEngine (wrapper) - and I don't expect them needing to provide any custom strategies of handling propagated error responses there (e.g. forwarding errors immediately instead of waiting until the end of the rendering)

core-libs/setup/ssr/optimized-engine/optimized-ssr-engine.ts
```diff
@@ -473,4 +469,27 @@ export class OptimizedSsrEngine {
       renderCallback(err, html);
     });
   }
+
+  /**
+   * Logs the result of the render.
+   */
+  private logForRenderResult(
+    err: any,
```
```suggestion
    err: unknown,
```

projects/storefrontapp/server.ts
pawelfras: Yes, the idea is to provide it to fresh SPA apps via schematics. `uber-error-handler` is a tempting idea giving us possibilities for easier adjustments. The reason why I decided to introduce separate handlers was the question what if the customer wants to have our handler for unknown server errors, but they custom handler with an error page for 404? 
Actually, now I have an answer... the custom handler can be provided before ours. What do you think about that?

core-libs/setup/ssr/engine/cx-common-engine-wrapper.ts
pawelfras: I didn't consider the serverless approach... (my bad) that's why I thought it'd be better to encapsulate the logic responsible for handling propagated errors inside the `cxCommonEngineWrapper` and call the `callback` function there. But in the case of serverless, your proposal totally makes sense. Thanks!

core-libs/setup/ssr/engine/cx-common-engine-wrapper.ts
pawelfras: Looks cool! Let me implement this and check whether it doesn't affect desired behavior.

core-libs/setup/ssr/error-handling/multi-error-handlers/server-responding-error-handler.ts
pawelfras: Good catch. Leftovers after merge... 🤦

core-libs/setup/ssr/error-handling/server-error-response-factory/cms-page-not-found-server-error-response-factory.spec.ts
pawelfras: Not sure whether I like the proposal more 🤔

core-libs/setup/ssr/error-handling/server-error-response/propagate-server-error-response.ts
pawelfras: > IMHO the token should be in public API, so our customers can freely customize our ServerRespondingErrorHandler if they want -> then they'll need to call the propagating function themselves, but then they need to inject it themselves (so they need the token to be in public API)

Good point. I considered only the second thing you mentioned, that customers probably won't provide their custom implementation of the propagator function. Still, yeah, they might need/want to implement their own ServerREsponsingErrorHandler and then the token for the propagator function is required.

projects/storefrontapp/server.ts
Platonn: [IMHO](https://blog.platis.dev/open-source) the most frequent use cases should be handled with a simple config, or super simple API. Ideally enabled implicitly by default. Medium-frequent use cases should require a tiny bit of coding, but proportionally to how often it is a use case. And for edge cases it's acceptable that customers might need to give up many OOTB default APIs and possibly write a lot of copy-pasted code + some customizations. (https://blog.platis.dev/open-source)

And in this case, in my opinion, default error handlers are good enough for most people. So, it should be a super simple and ergonomic API to use. Ideally implicit, but unfortunately, it's not possible because you have to put it explictly into the server.ts file.

If someone wants to add their error handlers - well, you nicely noticed that they can add their own above ours!! Nice catch!

If someone would like to remove one of our handlers at all, then they might want to remove our pack of error handlers. But for ergonomics, let's expose in our public APi all individuall handlers - so customers can pick and mix whatever they want, explicitly.

core-libs/setup/ssr/engine/cx-common-engine.spec.ts
```diff
@@ -61,102 +58,86 @@ export class TokenComponent {
...
-    );
+    // await cxCommonEngineWrapper(
+    //   engine,
+    //   {
+    //     req: {} as any,
+    //     bootstrap: null as any,
+    //     document: '<cx-response></cx-response>',
+    //   },
+    //   callback
+    // );
```
Platonn: 👀

core-libs/setup/ssr/engine/cx-common-engine.ts
```diff
@@ -0,0 +1,54 @@
...
+/**
+ * The Spartacus extension of the CommonEngine introduced to handle propagated server responses caught during server-side rendering.
+ * @extends {CommonEngine}
+ */
+export class CxCommonEngine extends CommonEngine {
+  /**
+   * Stores any server error response that occurs during the rendering process.
+   * @type {CxServerErrorResponse | undefined}
+   */
+  protected errorResponse: undefined | CxServerErrorResponse;
```
Platonn: CommonEngine's instance is one for all requests. We should not share his state among all requests. Otherwise it may lead to funky bugs.
Let's define it as a variable in the beginning of the `render()` method, so it's local and available only in the method's closure

core-libs/setup/ssr/engine/cx-common-engine.ts
```diff
@@ -0,0 +1,54 @@
...
+    super(options);
+  }
+
+  /**
+   * Renders the given options and handles any server error response.
+   * If an error response exists, it will be thrown after the rendering is complete.
+   * @param {CommonEngineRenderOptions} options - The options to render.
+   * @returns {Promise<string>} The rendered HTML as a string.
+   * @throws {CxServerErrorResponse} The server error response, if one occurred during rendering.
+   */
```
Platonn:
```suggestion
  /**
   * @override
   * Renders for the given options.
   * If a server error response object is populated from the rendered applications
   * (via `PROPAGATE_SERVER_ERROR_RESPONSE` callback), then such an error
   * will be thrown and the result promise rejected - but only AFTER the rendering is complete.
   *
   * @param {CommonEngineRenderOptions} options - The options to render.
   * @returns {Promise<string>} Promise which resolves with the rendered HTML as a string
   *                                                 OR rejects with the server error response object, if any is propagated from the rendered app.
   */
```

core-libs/setup/ssr/engine/cx-common-engine.ts
```diff
@@ -0,0 +1,54 @@
...
+  async render(options: CommonEngineRenderOptions): Promise<string> {
+    return super
+      .render({
+        ...options,
+        providers: [
+          {
+            provide: PROPAGATE_SERVER_ERROR_RESPONSE,
+            useFactory: () => {
+              return (serverErrorResponse: CxServerErrorResponse) => {
+                this.errorResponse ??= serverErrorResponse;
```
Platonn: For readability, IMHO it's worth to leave a hint for future readers in a code comment:
```suggestion
                // We're interested only the first propagated error, so we use `??=` instead of `=`:
                errorResponse ??= serverErrorResponse;
```

core-libs/setup/ssr/error-handling/server-error-response-handlers/server-error-response-handlers.ts
```diff
@@ -11,36 +11,25 @@ import {
...
       res.status(500).send(documentContent);
+    } else if (
+      !res.headersSent &&
+      err instanceof CmsPageNotFoundServerErrorResponse
+    ) {
+      res.set('Cache-Control', 'no-store');
+      res.status(404).send(documentContent);
     } else {
       next(err);
     }
```
Platonn: I apologize I might not have been clear in my last comment.
My intention was to:
- leave 2 separate handlers in our public API (in case customers want to pick and mix)
- and **moreover** expose one "wrapper" handler which composes the other two - that is ergonomic to use
- via schematics let's install only this "wrapper" error handler into `server.ts`

Does it make sense to you?

core-libs/setup/ssr/error-handling/server-error-response/propagate-server-error-response.ts
```diff
@@ -8,9 +8,8 @@ import { InjectionToken } from '@angular/core';
 import { CxServerErrorResponse } from './cx-server-error-response';
 
 /**
- * Injection token used to provide a function that will propagate the server error response occurred during server side rendering to the CommonEngine wrapper.
- *
- * @private
+ * Function that will propagate the server error response object
+ * created during server side rendering to the CxCommonEngine.
```
Platonn:
```suggestion
 * Function that will propagate the server error response object
 * (created during server side rendering) to the CxCommonEngine.
 ```

File: core-libs/setup/ssr/engine/cx-common-engine.spec.ts [hunk 22]
```diff
@@ -0,0 +1,143 @@
...
+
+@NgModule({
+  imports: [BrowserModule, ServerModule],
+  declarations: [WithPropagatedErrorComponent],
+  bootstrap: [WithPropagatedErrorComponent],
+})
+export class WithPropagatedErrorServerModule {}
+
+// Test that the cxCommonEngineWrapper doesn't override providers
+// If SOME_TOKEN not provided, test how the cxCommonEngineWrapper handles APP_INITIALIZER errors
```
Platonn: nitpick: this comment mentions `cxCommonEngineWrapper` which got removed from this PR

Note: there are other places in this file with leftover `cxCommonEngineWrapper` mentions

Let's fix it.

core-libs/setup/ssr/error-handling/server-error-response-handlers/server-error-response-handlers.ts
pawelfras: I was thinking about such a solution, but I haven't found a nice way to return two handlers of type ErrorRequestHandler return in one `defaultServerErrorResponseHandlers` function to use it in ExpressJS API (`server.use`):

```ts
export const defaultServerErrorResponseHandlers =
  (documentContent: string): ErrorRequestHandler => //won't match the type
   getCmsPageNotFoundServerErrorResponseHandler(documentContent)
   getUnknownServerErrorResponseHandler(documentContent)
  };
 ```
  
When I extracted the body of this function to two separate functions for each case, `defaultServerErrorResponseHandlers` looked more complex + it wouldn't make much sense, because these functions and take would not be usable directly as middleware in ExpressJS.

Right now this logic can be overwritten by custom middleware (for 404, 500, or both - if we override behaviour for 404, then handling 500 still will be taken from our default handler).

Anyway, I'm open for proposals.

Comment (Platonn) [reply core-libs/setup/ssr/error-handling/server-error-response-handlers/server-error-response-handlers.ts]: Ah, you're right! My bad. Ok, let's keep it as is!

FYI, I was thinking about something like below, but only now I realized the composition of those 2 handlers doesn't work at all as I indended 🤦  Pasting it only for reference. What you have now is good!

```ts
export const handleUnknownServerError = (documentContent: string) => {
  return (err, _req, res, next) => {
    if (!res.headersSent && err instanceof UnknownServerErrorResponse) {
      res.set('Cache-Control', 'no-store');
      res.status(500).send(documentContent);
    } else {
      next(err);
    }
  };
};

export const handleCmsPageNotFoundError = (documentContent: string) => {
  return (err, _req, res, next) => {
    if (!res.headersSent && err instanceof CmsPageNotFoundServerErrorResponse) {
      res.set('Cache-Control', 'no-store');
      res.status(404).send(documentContent);
    } else {
      next(err);
    }
  };
};

export const defaultServerErrorResponseHandlers =
  (documentContent: string): ErrorRequestHandler =>
  (err, _req, res, next) => {
    handleUnknownServerError(documentContent)(err, _req, res, next);
    handleCmsPageNotFoundError(documentContent)(err, _req, res, next);
  };
```

core-libs/setup/ssr/engine/cx-common-engine-wrapper.ts
Platonn: TBH I'm not sure the "serverless"  approach will be every used in conjunstion with this piece of code. I mentioned it just as an example of different circumstances in which the good separation of concerns becomes more clear to see. 

i.e. The Extended CommonEngine should only care about what extra functionality it provides (i.e. capturing async errors and populating them up as if it was a native feature of the CommonEngine). But it should not care about the context it is used withing. For example, the CommonEngine can be used inside a NgExpressEngine OR inside server.ts.

Platonn (approved):

PR #18758, opened 2024-04-23
feat: (CXSPA-946) - Keyboard navigation for FacetComponent

Pio-Bar:
Ticket: [CXSPA-946](https://jira.tools.sap/browse/CXSPA-946)

Adds keyboard navigation for the facet component. Left and right arrow keys should toggle the facet open/closed. Up and down arrows navigate the options.


projects/storefrontlib/cms-components/product/product-list/product-facet-navigation/facet/facet.component.ts
```diff
@@ -106,4 +114,63 @@ export class FacetComponent {
...
+        this.onArrowDown(event, targetIndex);
+        break;
+      case 'ArrowUp':
+        this.onArrowUp(event, targetIndex);
+        break;
+    }
+  }
+
+  onArrowRight(event: Event): void {
+    if (this.isExpanded) {
```
Zeyber: Not really anything wrong with this. Any argument for not putting the logic in the condition (inverted) rather than returning when the condition is met? Component logic is quite long so it could save some lines without being harder to read IMO.

Zeyber (changes requested):

projects/storefrontlib/cms-components/product/product-list/product-facet-navigation/facet/facet.component.ts
Pio-Bar: I've inverted the conditions. I've also added the feature flag to the template, restoring the default tab flow if the feature is off. I must have missed this earlier! ;)

projects/storefrontlib/cms-components/product/product-list/product-facet-navigation/facet/facet.component.ts
```diff
@@ -53,6 +57,16 @@ export class FacetComponent {
     return this._facet;
   }
 
+  @Optional() featureConfigService = inject(FeatureConfigService, {
+    optional: true,
+  });
+
+  get isFacetKeyboardNavigationEnabled(): boolean {
```
Zeyber, line 66: Interesting way to have a feature flag in an attribute. I think we definitely want a comment here to clean this up and maybe even a `@deprecated` message so its clear that its not meant to be a public api.

Zeyber (changes requested):

projects/storefrontlib/cms-components/product/product-list/product-facet-navigation/facet/facet.component.ts
Pio-Bar: Comment added. I can't use `@deprecated` or SonarCube complains at me for using deprecated code. 🙈

Zeyber: Ah I see. Thanks for adding the comment then :D

Zeyber (approved):

developpeurweb (approved):

PR #18772, opened 2024-04-25
test: Failing scroll-position-restoration spec

sdrozdsap:
In some cases Cypress was trying to scroll to the bottom of the page before it was fully rendered. Additional condition for determining if the content appeared on the page should make it more stable. After these changes I've executed this test 50 consecutive times and no error happened


Zeyber (approved):
Thanks! Lets change the commit title to `test:` so it doesn't appear in the changelog.

PR #18773, opened 2024-04-25
fix:scroll-position-restoration e2e

kimhw0630:
fix scroll-position-restoration e2e

### DIFF
--- a/projects/storefrontapp-e2e-cypress/cypress/e2e/regression/scrolling/scroll-position-restoration.e2e.cy.ts
+++ b/projects/storefrontapp-e2e-cypress/cypress/e2e/regression/scrolling/scroll-position-restoration.e2e.cy.ts
@@ -40,6 +40,7 @@ context('scroll Position Restoration', () => {
 
         cy.log('Go to product details page');
         verifyProductPageLoaded(productName);
+        cy.wait(500);
         cy.window().scrollTo('bottom');
 
         cy.log('Go back to product list');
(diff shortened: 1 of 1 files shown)


projects/storefrontapp-e2e-cypress/cypress/e2e/regression/scrolling/scroll-position-restoration.e2e.cy.ts
```diff
@@ -16,24 +16,35 @@ context('scroll Position Restoration', () => {
...
     cy.get('cx-category-navigation a').eq(0).click();
 
     cy.get('cx-product-list-item').should('exist');
-    cy.get('cx-product-list-item .cx-product-name')
+    cy.get('cx-product-list-item .cx-product-name a')
       .eq(3)
       .then(($productItem) => {
         const productName = $productItem.text();
-        cy.wrap($productItem).scrollIntoView().click();
+        cy.wrap($productItem).scrollIntoView().click({ force: true });
```
Zeyber: I'm not sure that using `force` here is the best solution because we should be asserting that products are actually visible before we click them. Especially since this test is relating to scroll position, I think that it matters.

Zeyber (changes requested):

projects/storefrontapp-e2e-cypress/cypress/e2e/regression/scrolling/scroll-position-restoration.e2e.cy.ts
kimhw0630: 'scrollIntoView' should scroll to the position, but judging from the issue, it seems it wasn't clicking. Therefore, I added a 'force click' to address this. I will remove '{ force: true }'

RadhepS (dismissed):

Zeyber (approved):

PR #18780, opened 2024-04-30
Define cartQuickOrderRemoveListeningToFailEvent feature toggle to avoid showing an unnecessary duplicated error message on the failure of adding to cart

Larisa-Staroverova:
@Platonn, could you please review this PR?
Thanks a lot in advance!


feature-libs/cart/quick-order/components/cart-quick-order-form/cart-quick-order-form.component.ts
```diff
@@ -141,6 +149,13 @@ export class CartQuickOrderFormComponent implements OnInit, OnDestroy {
     );
   }
 
+  /**
+   * @deprecate since SPA_2406
+   *
+   * This method is no longer needed since BadRequestHandler.handleUnknownIdentifierError was introduced.
+   * If this method is used an unnecessary duplicated error message will appear in the UI.
+   * Therefore this method will be removed.
```
Platonn:
```suggestion
   * @deprecated since 2211.24
   *
   * This method is no longer needed since BadRequestHandler.handleUnknownIdentifierError was introduced.
   * If this method is used an unnecessary duplicated error message will appear in the UI.
   * Therefore this method will be removed.
   *
   * You can enable the Feature Toggle 'cartQuickOrderRemoveListenToFailEvent'
   * to stop calling this method by default.
```

projects/core/src/features-config/feature-toggles/config/feature-toggles.ts
```diff
@@ -171,6 +171,12 @@ export interface FeatureTogglesInterface {
    * Determines whether the controls in the `CarouselComponent` are focusable and accessible from the keyboard.
    */
   a11yFocusableCarouselControls?: boolean;
+
+  /**
+   * Determines whether `watchAddEntryFailEvent` method is executed or not in `CartQuickOrderFormComponent` component.
+   * The `watchAddEntryFailEvent` method listens to CartAddEntryFailEvent event.
+   */
+  cartQuickOrderRemoveListenToFailEvent?: boolean;
```
Platonn: IMHO it's worth also mentioning the second-order consequence of this change, which is avoiding an unnecessary doubled error message.

```suggestion
  /**
   * In `CartQuickOrderFormComponent` it stops calling the deprecated method 
   * `watchAddEntryFailEvent()`, which listens to the `CartAddEntryFailEvent`.
   *
   * It avoids showing an unnecessary doubled error message on the failure
   * of adding to cart.
   */
  cartQuickOrderRemoveListenToFailEvent?: boolean;
```

projects/core/src/features-config/feature-toggles/config/feature-toggles.ts
```diff
@@ -203,4 +209,5 @@ export const defaultFeatureToggles: Required<FeatureTogglesInterface> = {
   a11yUnitsListKeyboardControls: false,
   a11yCartItemsLinksStyles: false,
   a11yFocusableCarouselControls: false,
+  cartQuickOrderRemoveListenToFailEvent: false,
```
Platonn: `RemoveListen` sounds quite funky due to 2 verbs in a row. What about `RemoveListening`?

PS. If we go for it, then let's also update the descriptions of unit tests mentioning the old toggle's name

Platonn (approved):

PR #18781, opened 2024-04-30
feat: create `BEFORE_CMS_PAGE_GUARD` injection token allowing for custom pre-`CmsPageGuard` logic

Platonn:
Exposed a new extensible hook `BEFORE_CMS_PAGE_GUARD`, which is needed for injecting composable global guards to be run on every CMS route, just before executing CmsPageGuards's main logic.

Example usages:
- `ProtectedRoutesGuard` (existing)
- `OppsLoginRequiredGuard` (comming soon in another PR https://github.com/SAP/spartacus/pull/18727)
- (perhaps usages coming in the future).

Architectural notes:
- created the injection token `BEFORE_CMS_PAGE_GUARD` and an accompanying service `BeforeCmsPageGuardService` which is used in the `CmsPageGuard`.
- `ProtectedRoutesGuard` is no longer called directly in the `CmsPageGuard`, but indirectly - thanks to being provided via the token `BEFORE_CMS_PAGE_GUARD`
- created a `GuardsComposer` utility class that is able to execute multiple given guards and compose their Observable results into one Observable
  - re-used `GuardsComposer` in `BeforeCmsPageGuardService`
  - re-used `GuardsComposer` in the existing `CmsRoutesService` so this service no longer calls its own deprecated method `canActivateGuard()`. But this new behavior was hidden behind a new Feature Toggle (WHY: Perhaps some customer already customized the method `CmsRoutesService.canActivateGuard()` and expects it to be invoked by default).
  
**QA steps:**

**1. Regression test "protected routing" feature**

- in `app.module` provide: `provideConfig({ routing: { protected: true } })`
- open homepage or PDP and verify you're redirected to login page
- after logging in verify that you're redirected to the anticipated previously homepage or PDP
  
**2. Test adding potentially more `BEFORE_CMS_PAGE_GUARD` classes.**

Disclaimer: all code snippets below should be pasted to the `app.module.ts` file:

- add to `providers` 2 debugging `APP_INITIALIZER`s:
```ts
    // Dump to console all `BEFORE_CMS_PAGE_GUARD` guards:
    {
      provide: APP_INITIALIZER,
      multi: true,
      useFactory: () => {
        const guards = inject(BEFORE_CMS_PAGE_GUARD);
        return () => {
          console.log('Provided `BEFORE_CMS_PAGE_GUARD`:', guards);
        };
      },
    },

    // Dump to console all `Router` events:
    {
      provide: APP_INITIALIZER,
      multi: true,
      useFactory: () => {
        const injector = inject(Injector);
        return () => {
          injector.get(Router).events.subscribe((event) => {
            console.log(`[Router Event: ${event.constructor.name}]`);
          });
        };
      },
    },
```
- In the top of the `app.module` define 3 guard classes (1st and 3rd are brand new; 2nd is an extension of existing `ProtectedRoutesGuard`. They delay their responses by: 3s, 2s, 1s (for respectively: 1st, 2nd, 3rd guard)
```ts
export class SpikeGuardAlpha {
  canActivate() {
    return of(true).pipe(
      tap(() => console.log('SpikeGuardAlpha - START')),
      tap(() => console.log('SpikeGuardAlpha - delay 3_000...')),
      delay(3_000),
      tap(() => console.log('SpikeGuardAlpha - END\n\n'))
    );
  }
}

export class SpikeProtectedRoutesGuard extends ProtectedRoutesGuard {
  override canActivate(
    route: ActivatedRouteSnapshot
    return of(true).pipe(
      tap(() => console.log('SpikeProtectedRoutesGuard - START')),
      tap(() => console.log('SpikeProtectedRoutesGuard - delay 2_000...')),
      delay(2_000),
      switchMap(() => super.canActivate(route)),
      tap(() => console.log('SpikeProtectedRoutesGuard - END\n\n'))
    );
  }
}


export class SpikeGuardBeta {
  canActivate() {
    return of(true).pipe(
      tap(() => console.log('SpikeGuardBeta - START')),
      tap(() => console.log('SpikeGuardBeta - delay 1_000...')),
      delay(1_000),
      tap(() => console.log('SpikeGuardBeta - END\n\n'))
    );
  }
}
```
- define an `@NgModule` that provides the `SpikeGuardAlpha` as `BEFORE_CMS_PAGE_GUARD`:
```ts
@NgModule({
  providers: [
    {
      provide: BEFORE_CMS_PAGE_GUARD,
      multi: true,
      useExisting: SpikeGuardAlpha,
    },
  ],
})
export class SpikeAlphaModule {}
```
- import this `@NgModule` `SpikeAlphaModule` just before `SpartacusModule`:
```diff
      EffectsModule.forRoot([]),
+     SpikeAlphaModule,
      SpartacusModule,
```
- override the `ProtectedRoutesGuard` with `SpikeProtectedRoutesGuard`:
```ts
    {
      provide: ProtectedRoutesGuard,
      useExisting: SpikeProtectedRoutesGuard,
    },
```
- provide `SpikeGuardBeta` as  `BEFORE_CMS_PAGE_GUARD`:
```ts
    {
      provide: BEFORE_CMS_PAGE_GUARD,
      multi: true,
      useExisting: SpikeGuardBeta,
    },
```
- (ensure you're logged out)
- (for clarity you might want to suppress existing warnings from other features:)
image

- open http://localhost:4200/electronics-spa/en/USD/login and verify your console contains the following sequence: 
  - Please notice that guards run **sequentially** (don't race) in the order of being provided as `BEFORE_CMS_PAGE_GUARD` token. The delays in their responses don't matter. It's the providers order that matters.

image
  
- open http://localhost:4200/electronics-spa/en/USD/Open-Catalogue/Cameras/Hand-held-Camcorders/c/584 and verify your console contains the following sequence:
image

- in the implementation of `SpikeGuardAlpha` change `of(true)` to `of(false)`

- open http://localhost:4200/electronics-spa/en/USD/Open-Catalogue/Cameras/Hand-held-Camcorders/c/584 and verify your console contains the following sequence:
image

closes https://jira.tools.sap/browse/CXSPA-6977


projects/storefrontlib/cms-structure/services/guards-composer.ts
```diff
@@ -0,0 +1,100 @@
...
+   * Tells whether the given object has a `canActivate` method.
+   */
+  protected isCanActivate(guard: any): guard is CanActivate {
+    return guard && this.isFunction<CanActivate>(guard.canActivate);
+  }
+
+  /**
+   * Tells whether the given object is a function.
+   */
+  protected isFunction<T>(v: any): v is T {
```
pawelfras: `query:` Is this type guard required? If yes, does it need to be generic?

Btw. I have a problem understanding its logic - if we use it with type `CanActivate` as above, then we're checking whether `v is CanActivate` by verifying `typeof v === function`. The `CanActivate` interface has `canActivate` function definition inside, but this interface isn't a function itself 🤔 
Can you help me understand this?

projects/storefrontlib/cms-structure/guards/cms-page.guard.ts
```diff
@@ -28,11 +29,14 @@ export class CmsPageGuard {
   constructor(
     protected routingService: RoutingService,
     protected cmsService: CmsService,
+    /** since 2211.24 not used anymore, but called indirectly via {@link BeforeCmsPageGuardService} */
```
pawelfras, line 32: `query:` Do we have information if customers extend this guard? If yes, maybe it's worth deprecating the whole constructor and adding new one without `ProtectedRoutesGuard`.

projects/storefrontlib/cms-structure/services/cms-guards.service.ts
```diff
@@ -25,13 +27,36 @@ export class CmsGuardsService {
...
 
+    if (
+      this.featureConfigService.isEnabled('cmsGuardsServiceUseGuardsComposer')
+    ) {
+      const guardsInstances: CanActivate[] = guards
+        .map((guardClass) =>
+          getLastValueSync(this.unifiedInjector.get<CanActivate>(guardClass))
+        )
+        .filter(
+          (guardInstance): guardInstance is CanActivate =>
```
pawelfras: `nitpick:` Consider extracting this logic to a separate function for readability reasons (my subjective opinion)

projects/storefrontlib/cms-structure/guards/cms-page.guard.ts
Platonn: IMHO we should no longer bother deprecating specific constructors. 
I'd rather like to remove all constructors from all classes and use `inject()` everywhere - as a separate initiative

projects/core/src/util/rxjs/wrap-into-observable.spec.ts
```diff
@@ -0,0 +1,47 @@
...
+    const obj = {
+      then: function (): void {},
+    };
+    const promiseValue2 = Promise.resolve(obj);
+    const result2 = wrapIntoObservable(promiseValue2);
+
+    result2.subscribe((value) => {
+      expect(value).toEqual(obj);
+      done();
+    });
```
Platonn: Each of the subscription bodies needs a `done`.

But let's split it to 2 unit tests - each test with one `expect` and one `done` within subscription.

Why: In Jasmine, a test is considered successful when all `expect()` checks pass and when the `done` function is invoked. However, if a single unit test invokes the `done` function in two different places, it can result in false positive results. This is because Jasmine won't wait for the second invocation of `done` after the first one, and any `expect()` checks placed in async operations near the second done may not be properly checked. This can lead to race conditions where the test may finish before all `expect()` checks are completed or the test results may vary depending on the timing of async operations guarded by done calls.

Note: please review other places in this PR for this pattern

projects/storefrontlib/cms-structure/guards/before-cms-page-guard.service.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+    spyOn(guardsComposer, 'canActivate').and.returnValue(of(true));
+
+    service.canActivate(route, state).subscribe((value) => {
+      expect(guardsComposer.canActivate).toHaveBeenCalledWith(
+        service['guards'],
+        route,
+        state
+      );
+      expect(value).toEqual(true);
+    });
```
Platonn, line 50: let's add done() here

projects/storefrontlib/cms-structure/services/guards-composer.spec.ts
```diff
@@ -0,0 +1,91 @@
...
+      { canActivate: () => of(true) },
+      { canActivate: () => of(false) },
+      { canActivate: () => of(urlTree) },
+    ];
+
+    service.canActivate(guards2, route, state).subscribe((result) => {
+      expect(result).toEqual(false);
+      done();
+    });
+  });
```
Platonn: Very nice tests in this PR!

Let's add yet 2 unit tests more to verify the following statement: `should return first encountered false / UrlTree - even if emitted after a delay`. For example"
```ts
    const guards1: CanActivate[] = [
      { canActivate: () => of(true) },
      { canActivate: () => of(urlTree).pipe(delay(2)) },
      { canActivate: () => of(false).pipe(delay(1)) },
    ];

    service.canActivate(guards1, route, state).subscribe((result) => {
      expect(result).toEqual(false);
      done();
    });
```
and

```ts
    const guards1: CanActivate[] = [
      { canActivate: () => of(true) },
      { canActivate: () => of(urlTree).pipe(delay(1)) },
      { canActivate: () => of(false).pipe(delay(2)) },
    ];

    service.canActivate(guards1, route, state).subscribe((result) => {
      expect(result).toEqual(urlTree);
      done();
    });
```

projects/storefrontlib/cms-structure/services/cms-guards.service.ts
Platonn: Done

projects/storefrontlib/cms-structure/services/guards-composer.ts
Platonn: It's a relic of copying previous implementation here.

btw. It should be `this.isFunction<CanActivateFn>(...` but not `...<CanActivate>(...` .
So it seems that this wasn't even serving any purpose - as it didn't raise any errors when used with mistaken type.
So I'm removing this function.

projects/storefrontlib/cms-structure/services/guards-composer.spec.ts
anjana-bl: done

projects/storefrontlib/cms-structure/guards/before-cms-page-guard.service.spec.ts
anjana-bl: done

projects/core/src/util/rxjs/wrap-into-observable.spec.ts
anjana-bl: done

projects/storefrontlib/cms-structure/services/guards-composer.ts
```diff
@@ -0,0 +1,90 @@
...
+      skipWhile((canActivate: boolean | UrlTree) => canActivate === true),
+      endWith(true),
+      first()
+    );
+  }
+
+  /**
+   * Tells whether the given object has a `canActivate` method.
+   */
+  protected isCanActivate(guard: any): guard is CanActivate {
```
pawelfras, line 87: `nitpick:` the same logic is used in CmsGuardsService (line 53). Consider moving it to some utils file and reuse in both places.

Platonn: The same logic in `CmsGuardsService` is going to be removed as soon as we remove the deprecated method there.

pawelfras (approved):
Code LGMT

PR #18785, opened 2024-05-02
feat: support of configuration Bundles

leowangchao:
Hi @Zeyber,
In order to support bundles in Cart Detail Page, we created the `hierarchy-select` folder and created some bundle-related components in it(it's a draft version, still needs polishing).
image
The following is the effect of running:
image

Now, we need to support bundles in more places, such as Order(order Approval, return order, cancel order...)/ Quote/ Checkout/ Saved Cart, etc., you can see it in [Figma](https://www.figma.com/file/iI1Rx0NGhj1vG6QEmEjXhf/Bundles?type=design&node-id=701-65317&mode=design&t=ckKyOVHVQzS3YLC3-0).

There are some questions that need your opinion:
<strike>
1. Now on the cart detail page, we use the `cx-hierarchy-select` component as shown below.
In order to adapt to the scenarios of other pages, if we add some input parameters to the `HierarchySelectComponent` to determine which page should be rendered, is this a feasible solution?
I feel like this will make the `HierarchySelectComponent` very large. Do you have any suggestions?
</strike>

1. Now on the cart detail page, we use the `cx-hierarchy-select` component as shown below.
image
And prepare data through `prepareBundle` during initialization:
image
image

In order to adapt to the scenarios of other pages, I thought of a way:

Now in` hierarchy-node.component.ts`, it supports two types of nodes `TITLE` and `COLLAPSIBLE`.

image

When we want to support bundles in more places, such as `Quote`,  We can create a `Quote` specific `TITLE` and `COLLAPSIBLE` components, maybe called `cx-hierarchy-quote-title`, `cx-hierarchy-quote-collapsible`, then added it into ` hierarchy-node.component.ts`.
So in this way, we can code its corresponding HTML templete separately for Quote and when we prepare data through during initialization we can use `cx-hierarchy-quote-title`/ `cx-hierarchy-quote-collapsible`.
This way, each different page will have its own type of node.
Of course, in the end, we may be able to extract their most basic functions and make them into a parent component, (such as `cx-hierarchy-title`, `cx-hierarchy-collapsible`,) 
and then let other pages inherit it as subclasses. (`cx-hierarchy-quote-title`, `cx-hierarchy-quote-collapsible`)
What do you think about this kind of implementation?

2. Because we started working from cart, so we put the `hierarchy-select` folder under `feature-libs/cart/base/components`, but it is not only used by cart, also by the Order/Quote/Checkout/Saved Cart mentioned above. So, where would be a more appropriate choice for us to put it?

3. As I know, `Bundles` should be the same as `Quote` that customers can choose whether to enable it according to their configuration. So how is this achieved? What should we pay attention to during development? Is there any example you can refer to, thank you.

4. A issue when the cart item delete.
The problem lies in the cart detail page. Like the following:
image

When I click the Remove link in cart item, the quantity becomes 0 and after a while the item is removed from the list. When all the items are removed, a cart empty page is shown:
image

These are all working fine, but now when we add a section for bundle, I find when we delete the last item, it will NOT disappear, the quantity just becomes 0, like this:
image
Debugging shows the remove cart entry / load cart API is completed, but we do not know how and where the UI is synchronized with the store data, and why it does NOT work only for the last item.


Zeyber:
1. I like the way you have come up with a component and especially the idea you have had to make it reusable (ie. cx-hierarchy => cx-hierarchy-quote, etc). In my opinion, it is fine to continue down this path but it would be nice to outline "what is the hierarchy component" and "what does it do exactly" so that we can take it to other features and ask ourselves "will it work for this?". I think this might be the track you are already on though. I would be happy to review this component on its own when you are past the draft stage of it so we can validate it.

2. Yes, since this component is planned to be used in different libraries, I think that we could keep this hierarchical component in `storefrontlib` in the `layout` components.

3. I have no personal experience with Quotes myself. I think you could also direct this question to Team Tiger based on [this feature commit](https://github.com/SAP/spartacus/commit/1cf1663c122e9316362d214ca228fa9ae2b3164a). My personal advise would simply to make the component as dumb as possible (nothing to do with bundles or even cart, leveraging template outlets and using the hierarchy for UI and accessibility). I don't imagine there would be too much similarities between quotes and bundles (could be wrong though, please check with Team Tiger) since quotes is its own feature (ie. not in the cart library but its own one). On top of the features you already need to cover, it would also be great if other teams could use it for their features.

4. For this one, I can't really tell what's going on or wrong without testing it. Lets look into it during our meeting :)

feature-libs/cart/base/components/cart-details/cart-details.component.html
```diff
@@ -1,5 +1,6 @@
 <ng-container *ngIf="cart$ | async as cart">
-  <ng-container *ngIf="entries$ | async as entries">
+  <ng-container *ngIf="{entries: entries$ | async, entryGroups: entryGroups$ | async, standaloneEntries: standaloneEntries$ | async, bundle: entries$ | async} as data">
```
Zeyber: It seems to me that this is causing your issue with the cart entries still being visible when they are deleted. Having an object as your `ngIf` directive means this is mostly going to resolve as true as long as any of those properties resolve to some non-null value (ie. entries or entryGroups). My first suggestion is to treat these is separate parings so that you don't cross the logic of these objects.

Example:
```
<ng-container *ngIf="entries$ | async as entries">
  // ... all entries code as before stays here
</ng-container>

// could be better to refer to this section as entryGroups than bundles to avoid confusion.
<ng-container *ngFor="let bundle of bundles$ | async">
  // ... all entryGroup code here
</ng-container>
```

To follow up on this suggestion, in future we will want to deprecate `entries` and only use `entryGroups` with `STANDALONE` and `BUNDLE` type groups. We will need to account for supporting both options initially by enabling "feature toggles" to handle any breaking changes. To show you what that means:

```
// The user has enabled feature toggle "isEntryGroupsEnabled" which makes the cart use the new entry groups feature to provide bundle support.
<ng-container [cxFeature]="isEntryGroupsEnabled">
  <ng-container *ngFor="let bundle of bundles$ | async">
    // ... all entryGroup code here with STANDALONE (simple entry) and BUNDLE type items
  </ng-container>
</ng-container>

// The user has NOT enabled feature toggle "isEntryGroupsEnabled" which makes the cart use the OLD entries items. So new features that use entryGroups like bundles will not be supported until the user opts-in.
<ng-container [cxFeature]="!isEntryGroupsEnabled">
  <ng-container *ngIf="entries$ | async as entries">
    // ... all entries code as before stays here
  </ng-container>
</ng-container>
```

With this in mind, try to account for both cases where `entries` logic should stay exactly how it is until we eventually deprecate and remove it. And try to create new logic using the `entryGroups` instead. It is a difficult concept and will be challenging so feel free to ask questions :)

leowangchao: image
I tried to change the code and debug it according to the first suggestion, as shown in the picture above, but when the cart has entrygroup and normal products, and delete the normal product, the cart zero issue still exists. You can adapt the code to yours environment and give it a try. BTW, the backend I use is "CX_BASE_URL": "https://40.76.109.9:9002"

https://github.com/SAP/spartacus/assets/153475556/224a4b3e-c7af-4142-b99e-fa36a11731a6

Zeyber: Thanks for the supporting material! 

It is clear that there are still `entries` because you have still have added the `entry` from the `entryGroup`. After deleting the last "visible" entry, this entryGroup's entry still exists so nothing will be refreshed. 

That said, I think you need to change the approach as suggested in my previous comment. That is, support `entries` OR `entryGroups` and not both at the same time. Resolving the above issue will not really contribute to the end architecture which is to support one or the other.

You will need to account for STANDALONE products in entryGroups to handle entry-like groups (ie. adding ui for them in the cart). But in this you can avoid using entries all-together just use entryGroups which is what we will prefer to do in the future anyhow.

leowangchao: > To follow up on this suggestion, in future we will want to deprecate `entries` and only use `entryGroups` with `STANDALONE` and `BUNDLE` type groups. We will need to account for supporting both options initially by enabling "feature toggles" to handle any breaking changes. To show you what that means:
> 
> ```
> // The user has enabled feature toggle "isEntryGroupsEnabled" which makes the cart use the new entry groups feature to provide bundle support.
> <ng-container [cxFeature]="isEntryGroupsEnabled">
>   <ng-container *ngFor="let bundle of bundles$ | async">
>     // ... all entryGroup code here with STANDALONE (simple entry) and BUNDLE type items
>   </ng-container>
> </ng-container>
> 
> // The user has NOT enabled feature toggle "isEntryGroupsEnabled" which makes the cart use the OLD entries items. So new features that use entryGroups like bundles will not be supported until the user opts-in.
> <ng-container [cxFeature]="!isEntryGroupsEnabled">
>   <ng-container *ngIf="entries$ | async as entries">
>     // ... all entries code as before stays here
>   </ng-container>
> </ng-container>
> ```
Hi Caine，
I'm not sure if I understand you correctly. In the cart-detail component, avoid using **`entries`** OR **`entryGroups`** at the same time?
Based on this principle, I modified the original code and did some debugging locally
Previously (left), the **getBundleEntryGroups** method obtained an Observable<OrderEntryGroup[]> containing only `CONFIGURABLEBUNDLE` type;
Now (right), it has become an Observable<OrderEntryGroup[]> containing `STANDALONE` type(normal product).
When entrygroup type is `STANDALONE`, give it's` entryGroup.entries` to a `Observable<OrderEntry[]> `as before, use to render normal cart list as snapshot2.
Because the code for getting `entries` is commented out，in this way, all `entries`' comes from `entrygroups` (including normal product).

But the cart zero issue is still existing. 




image
image

leowangchao: Hi @Zeyber ,
Here are 2 updates:
1. I followed your previous comments and all entries come from entrygroups. 
**The cart zero issue has been fixed.**
2. **The hierarchy-related components have been put into shared/component**, and template outlets were used to solve the code dependency problem.

Please review the code when you're available. Looking forward to your comments. If there are no major issues, I will continue to develop other pages using this structure.

Zeyber: Thanks for the updates! Its looking much cleaner now and you seem to have understood the concepts well :) I won't go into a great deal of comments now since this PR is still in progress and I have no major concerns with the course it is taking now. 

Just one minor remark that you should try to move the styling inputs into scss and use a selector like `nth` to nest paddings. We should try avoid styling logic in our class files unless we really need to to keep it clean.

leowangchao: Thanks for your updates.
Of course, I will try to move the styling into scss in the next update. 
As for using a selector like nth to nest paddings, because the tree does not know how many layers it has before rendering, it does not look very simple. Anyway, when the other pages are almost developed, I will take the time to check it.
Thanks

zhonghaojpg:
Hi @Zeyber,
   I have added the option to display and manipulate bundled products in the 'Saved Cart Details' section of the project.
 
 1,This is the rendering of the project：
Screenshot 2024-11-14 at 17 15 35

  2，List of code files involved in modifications：
      
image

projects/storefrontlib/shared/components/hierarchy/hierarchy/hierarchy.component.ts
```diff
@@ -0,0 +1,55 @@
...
+ */
+@Component({
+  selector: 'cx-hierarchy',
+  templateUrl: './hierarchy.component.html',
+})
+export class HierarchyComponent implements OnInit {
+  /**
+   * Defines the maximum height of visible part of the tree.
+   */
+  @Input() maxHeight: string;
```
Zeyber: There is quite a lot of input properties here. Do you think it might be better to pass in some object like `HierarchyOptions` instead?

projects/storefrontlib/shared/components/hierarchy/hierarchy/hierarchy.component.ts
```diff
@@ -0,0 +1,55 @@
...
+import { ActiveCartFacade } from '@spartacus/cart/base/root';
+
+/**
+ * Selector component that displays a tree-based model.
+ */
+@Component({
+  selector: 'cx-hierarchy',
+  templateUrl: './hierarchy.component.html',
+})
+export class HierarchyComponent implements OnInit {
```
Zeyber, line 17: I see this component has a lot of direct references to the cart library. It seems like its being used specifically for the cart entry groups. If you want to make this component a shared component for use with any library and keep it in `storefrontlib`, you need to remove any reference to the cart and make the component "dumb". 

If you plan to use this component with just the `cart` library, then you should move it to the cart library and rename it to something more targeted (eg. `CartNestedEntryGroupItem` - this example is extremely generalized and should have some more thought put into it).

We can discuss this if you need :)

projects/storefrontlib/shared/components/hierarchy/hierarchy-node-title/hierarchy-node-title.component.ts
```diff
@@ -0,0 +1,28 @@
...
+  templateUrl: './hierarchy-node-title.component.html',
+})
+export class HierarchyNodeTitleComponent {
+  @Input() tree: TitleNode;
+
+  @Input() activeCartService: ActiveCartFacade;
+
+  @Input() readonly = false;
+
+  removeBundle(entryGroupNumber: any) {
```
Zeyber: I see here that this is a very targeted purpose for the `onClick`. I have ellaborated further in another comment of whether this component is "cart-specific" or "dumb". But if you were to go down the dumb component route, you can pass some `doOnClick()` method as an input so that you can declare it earlier.

For example:

```
@Input() doOnClick: Function
...
onClick() {
  doOnClick()
}
...
<button (click)="onClick()">
...
```

Zeyber (changes requested):
Thanks for all your work towards this. You have made a lot of progress and its coming together nicely!

I've left just a few comments for now targeted at the `HierarchyComponent` in particular as I can see that a large amount of this feature relies on it so any modifications will cause many changes (hopefully for the better though!). Please see comments for examples. But in particular, I want to point out for now:
- Lets use `Input() options` properties for components that have lots of input values. This makes it easier to modify and extend them.
- I think that the `HierarchyComponent` should be made dumber and renamed rather than being tied to the cart at all. I believe we could reduce the complexity a lot by passing methods such as `onClick()` and `template` (like you already did :) ). I am happy to assist if you are having any difficulties with techniques on doing this.
- Core is good!

projects/storefrontlib/shared/components/hierarchy/hierarchy/hierarchy.component.ts
leowangchao: Good idea, an object called `HierarchyOptions` has been created instead

projects/storefrontlib/shared/components/hierarchy/hierarchy/hierarchy.component.ts
leowangchao: Very good point, I've removed any reference to the `cart` to make it more dumb.

projects/storefrontlib/shared/components/hierarchy/hierarchy-node-title/hierarchy-node-title.component.ts
leowangchao: Got it, I have made the processing logic of `remove` and `edit` into parameters to receive them with `HierarchyOptions`. If there are new methods in the future, they can also be extended in `HierarchyOptions`. like this：
```
export interface HierarchyOptions{
...
    onItemRemove?: (itemNumber: any) => void; // Function to handle item removal
    onItemEdit?: (itemNumber: any) => void; // Function to handle item edit
...
}
```

projects/storefrontlib/shared/components/hierarchy/hierarchy/hierarchy.component.service.ts
```diff
@@ -0,0 +1,109 @@
...
+import { Observable, map } from 'rxjs';
+
+@Injectable({
+  providedIn: 'root',
+})
+/**
+ * Service responsible for handling hierarchy-related operations.
+ */
+export class HierarchyComponentService {
+  /**
```
leowangchao, line 21: @Zeyber  I would like to discuss with you `where` to put this file? 
First, the main functions of this service are:
1. Construct and prepare the data structure required by the `hierarchyComponent` component; 
2. Filter and construct `entries` of `standalone` type products from `entryGroups`.
The service needs to process and return `OrderEntry` and `OrderEntryGroup`, so it will inevitably reference `cart`.

Then, the reason why I put it in `storefrontlib/shared/component/hierarchy...` is mainly because every place that uses `hierarchy components` needs to call them, not just cart-related pages. I am also using it in `order, pickup-in-store, checkout/saved-cart, etc.` If I put it in one of the functions, such as the `cart directory`, then other functions will need to call the service in the cart when using it. I don't know but it doesn't sound very reasonable？

Finally, according to the idea of ​​making the `hierarchy` component more `dumb` and no any external reference. So it doesn't seem particularly appropriate to leave it here now.

So I would like to discuss this with you, do you have any suggestions on `where to put it `more appropriately, or any other ideas?

leowangchao:
> Thanks for all your work towards this. You have made a lot of progress and its coming together nicely!
> 
> I've left just a few comments for now targeted at the `HierarchyComponent` in particular as I can see that a large amount of this feature relies on it so any modifications will cause many changes (hopefully for the better though!). Please see comments for examples. But in particular, I want to point out for now:
> 
> * Lets use `Input() options` properties for components that have lots of input values. This makes it easier to modify and extend them.
> * I think that the `HierarchyComponent` should be made dumber and renamed rather than being tied to the cart at all. I believe we could reduce the complexity a lot by passing methods such as `onClick()` and `template` (like you already did :) ). I am happy to assist if you are having any difficulties with techniques on doing this.
> * Core is good!

Hi @Zeyber , Thanks for your review, which provides very good ideas, especially making the `hierarchy` component more `general and dumb.`
I have made corresponding changes based on your comments.

Btw, I raised the discussion of the` hierachy-service` path in the previous comment.
I look forward to your reply, thank you. :)

projects/storefrontlib/shared/components/hierarchy/hierarchy-node-title/hierarchy-node-title.component.ts
```diff
@@ -0,0 +1,26 @@
...
+import { TitleNode } from './title-node.model';
+
+/**
+ * Hierarchy Selection node variant to show a title
+ */
+@Component({
+  selector: 'cx-hierarchy-node-title',
+  templateUrl: './hierarchy-node-title.component.html',
+})
+export class HierarchyNodeTitleComponent {
```
Zeyber, line 18: Is there a good reason to decouple this into its own component? I'm not sure if it is adding complexity to split the title part of a node into its own component rather than configure it in its parent.

projects/storefrontlib/shared/components/hierarchy/hierarchy-node/hierarchy-node.component.html
```diff
@@ -0,0 +1,23 @@
+<div>
```
Zeyber, line 1: Lets use css to apply the `display: block` style. No need to add further nesting to the DOM.

projects/storefrontlib/shared/components/hierarchy/hierarchy-node/hierarchy-node.component.html
```diff
@@ -0,0 +1,23 @@
...
+    <cx-hierarchy-node-title
+      [options]="{
+        tree: options.tree,
+        titleReadonly: options.titleReadonly,
+        onItemRemove: options.onItemRemove
+      }">
+    </cx-hierarchy-node-title>
+  </ng-container>
+  <!-- Collapsible Node -->
+  <ng-container *ngIf="type === 'COLLAPSIBLE'">
```
Zeyber, line 13: In general, it looks great! Nice job :D I still think that we can improve this component further.

So similar comment to above regarding the `node-title` component. I see that a node is either going to be collapsable (has some logic) or not-collapsable (little/zero logic). I don't think we need to differentiate these functions so much that they are separate components. 

Currently we have something like:
```
<cx-hierarchy>
  <cx-h-node-title> // Products on the go
  <cx-h-node-collapsable> // Camera component
  <cx-h-node-collapsable>  // Accessory component
```
I still think this might be a little more complicated that it needs to be. Especially given that the tree will be CMS-driven on any button press (ie. removeItem() function as to be passed in and tree is reloaded after making occ call). Also, we seem to stop nesting after the h-node. But this could be present in an example I haven't seen yet.

I think we could do something where we just use the node component and provide a class based on level to style it at different depths similar to the `cx-navigation-ui` component. For example:
```
<cx-hierarchy-node class="h-node-depth-0"> // Products on the go
  <cx-hierarchy-node class="h-node-depth-1"> // Camera component  
  <cx-hierarchy-node class="h-node-depth-1"> // Accessory component
```
This gives the possibility for the component to be more useful if we have even more depths:
```
<cx-hierarchy-node class="h-node-depth-0"> // If we want to wrap the list to make styling easier
  <cx-hierarchy-node class="h-node-depth-1"> // Products on the go
    <cx-hierarchy-node class="h-node-depth-2"> // Camera component  
    <cx-hierarchy-node class="h-node-depth-2"> // Accessory component
  <cx-hierarchy-node class="h-node-depth-1"> // Some other category
    <cx-hierarchy-node class="h-node-depth-2"> // Component  1
    <cx-hierarchy-node class="h-node-depth-2"> // Component 2
      <cx-hierarchy-node class="h-node-depth-3"> // Component 3
```

When it comes to the buttons as I mentioned before, maybe we could also pass these in the templates as they are very targeted in this instance to `bundles`. As in, not every hierarchy node with have an "edit" or "remove" button specifically. I am open to ideas on this however.

I would also encourage you to add the jsdoc-like comments to methods and properties on this component at this point. It can help with understanding the component for yourselves (and me) and identify some redundancy.

Any questions or comments about any of these points, please ask :)

projects/storefrontlib/shared/components/hierarchy/hierarchy/hierarchy.component.service.ts
Zeyber: I can see this component is not really related to the hierarchy component so much apart from the build tree function. IMO this is logic specific to the cart library in order to prepare your data for the hierarchy component. If you have some very generic helper method that can help build structure then it could live in a service like this. But I believe this is rather fine-grained to the cart component and should rather live somewhere there.

Zeyber (changes requested):

PR #18801, opened 2024-05-06
test: Add support for injecting Spartacus TestConfig chunk from ssr testing framework

Zeyber:
We're introducing `TestConfigServerModule`, which reads a cookie from the `REQUEST` (but not `DOCUMENT.cookie` - which is not implemented in runtime of the `Domino` lib). 

Also added a convenient new API to our SSR e2e testing framework for passing the extra `testConfig`.

Closes: [CXSPA-6633](https://jira.tools.sap/browse/CXSPA-6633)


projects/ssr-tests/tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
+  "extends": "../../tsconfig.json",
   "compilerOptions": {
-    "baseUrl": "./",
```
Platonn, line 3: Note:
- inheriting baseUrl being root dir - so all `Config` interface augmentations are taken into account (for type safety of `TestConfig`
- inheriting `paths` mapping for all libs (.e.g. `@sparatcus/core` resolved to `projects/core/src/public_api`) - so we can import `TestConfig` from `@spartacus/core`

core-libs/setup/ssr/testing/test-config-server.module.ts
```diff
@@ -0,0 +1,58 @@
...
+  return (result && decodeURIComponent(result[1])) || '';
+}
+
+export function parseConfigJSON(config: string) {
+  try {
+    return JSON.parse(decodeURIComponent(config));
+  } catch (_) {
+    return {};
+  }
+}
```
pawelfras, line 24: `query:` Are these functions repeated here to not expose them to `@spartacus/core` public API?

pawelfras (dismissed):
LGTM

core-libs/setup/ssr/testing/test-config-server.module.ts
Platonn: yes

pawelfras (dismissed):

pawelfras (approved):

PR #18808, opened 2024-05-07
Fix/added to cart dialog on cart UI event add to cart

ChristophHi:


feature-libs/cart/base/components/added-to-cart-dialog/added-to-cart-dialog-event.listener.ts
```diff
@@ -35,10 +35,18 @@ export class AddedToCartDialogEventListener implements OnDestroy {
       this.featureConfig.isEnabled('adddedToCartDialogDrivenBySuccessEvent')
     ) {
       this.subscription.add(
-        this.eventService
-          .get(CartAddEntrySuccessEvent)
-          .subscribe((successEvent) => {
-            this.openModalAfterSuccess(successEvent);
+        zip(
+          this.eventService.get(CartUiEventAddToCart),
+          this.eventService.get(CartAddEntrySuccessEvent)
```
Platonn, line 40: I believe we should not `zip` events "add to cart" with "add to cart success"

1. we don't have a formal guarantee that those events are related to the same action (i.e. potential race conditions of 2 "add to cart" vs 2 "add to cart success").

2. If an "add to cart failure" happens, then our pairs of events get de-synchronized

Example events sequence in time:
```
add-to-cart-1
add-to-cart-1-failure
add-to-cart-2
add-to-cart2-success // emits a zipped pair: [add-to-cart1, add-to-cart-2-success]
```

PR #18814, opened 2024-05-08
fix: (CXSPA-1795) - Organization table cells not linkable by default

Pio-Bar:
Ticket: [CXSPA-1795](https://jira.tools.sap/browse/CXSPA-1795)

To stop unnecessary links from being rendered I have switched the default linkable property to be false instead.
All config files for organization tables had to be slightly modified to accommodate this change. This ensures that only cells marked as linkable will be rendered in anchor tags.


feature-libs/organization/administration/components/budget/budget.config.ts
```diff
@@ -80,10 +80,57 @@ export const budgetCmsConfig: CmsConfig = {
   },
 };
 
-export function budgetTableConfigFactory(): TableConfig {
-  return budgetTableConfig;
+// TODO: (CXSPA-7155) - Remove feature flag and legacy config next major release
+export function budgetTableConfigFactory(
+  featureConfigService?: FeatureConfigService
+): TableConfig {
+  if (featureConfigService?.isEnabled('a11yOrganizationLinkableCells')) {
```
Zeyber, line 87: While I like the reference to `legacyConfig` and I think it will be easier to clean up, I think we should retain the original names for the legacy configs in case they have been exported and used elsewhere. We should use something like `newConfig` instead to reference our modern changes and change the reference back to `regularConfig` once we can clean up our breaking changes.

feature-libs/organization/administration/components/shared/table/cell.component.ts
```diff
@@ -37,9 +47,15 @@ export class CellComponent {
    * If the cells is linkable, an anchor link is created to the detailed route
    * of the given `_type`.
    *
-   * Defaults to `true`.
+   * Defaults to `false`.
```
Zeyber: I think we should clarify that it only defaults to false while the feature toggle is enabled. At least, until we can clean up breaking changes.

Zeyber (changes requested):

feature-libs/organization/administration/components/budget/budget.config.ts
Pio-Bar: Good point, I've changed the naming around.

feature-libs/organization/administration/components/shared/table/cell.component.ts
Pio-Bar: I've edited the comment to make that clear.

feature-libs/organization/administration/components/budget/budget.config.ts
Zeyber: Please do this for the remaining configs too ;)

feature-libs/organization/administration/components/unit/units.config.ts
```diff
@@ -194,10 +194,141 @@ export const unitsCmsConfig: CmsConfig = {
...
+  featureConfigService?: FeatureConfigService
+): TableConfig {
+  if (featureConfigService?.isEnabled('a11yOrganizationLinkableCells')) {
+    // TODO: (CXSPA-7155) - Remove feature flag and legacy config next major release
+    return NewUnitsTableConfig;
+  }
   return unitsTableConfig;
 }
 
+export const NewUnitsTableConfig: TableConfig = {
```
Zeyber: Wrong casing

Zeyber (changes requested):

Zeyber (dismissed):

developpeurweb (dismissed):

Zeyber (approved):

PR #18815, opened 2024-05-08
refactor: two toggles for strict error handling and server error propagation

pawelfras:


core-libs/setup/ssr/error-handling/multi-error-handlers/server-responding-error-handler.ts
```diff
@@ -26,10 +29,11 @@ export class ServerRespondingErrorHandler implements MultiErrorHandler {
   protected propagateServerErrorResponse = inject(
     PROPAGATE_SERVER_ERROR_RESPONSE
   );
-  protected isSsrErrorHandlingEnabled = inject(ENABLE_SSR_ERROR_HANDLING);
+  protected isServerErrorPropagationEnabled =
+    inject(FeatureToggles).serverErrorPropagation;
```
Platonn: Let's:
- inject FeatureConfigService as `private` property
- call it inline .isEnabled() in line 36

(it's not to leave unnecessary properties in public API `protected` that will need to be removed)

Note: it applies to also other places in this PR

projects/core/src/error-handling/effects-error-handler/cx-error-handler.effect.ts
```diff
@@ -9,22 +9,22 @@ import { Injectable, inject } from '@angular/core';
...
+import { FeatureToggles } from '../../features-config';
 import { ErrorAction } from '../../model/index';
 import { EffectsErrorHandlerService } from './effects-error-handler.service';
 
 @Injectable()
 export class CxErrorHandlerEffect {
   protected actions$ = inject(Actions);
   protected effectErrorHandler = inject(EffectsErrorHandlerService);
-  protected featureConfigService = inject(FeatureConfigService);
+  protected featureToggles = inject(FeatureToggles);
```
Platonn, line 20: same as here https://github.com/SAP/spartacus/pull/18815/files#r1594087591

core-libs/setup/ssr/optimized-engine/ssr-optimization-options.ts
```diff
@@ -164,6 +167,7 @@ export const defaultSsrOptimizationOptions: SsrOptimizationOptions = {
     defaultRenderingStrategyResolverOptions
   ),
   logger: new DefaultExpressServerLogger(),
-  cachingStrategyResolver: (entry) => !entry.error,
-  ssrErrorHandling: true,
+  cacheStrategyResolver: (options, entry) =>
+    !!options.cacheErrors && !!entry.error,
```
Platonn: 1. we have to check if  `cacheErrors` is expliclity `=== true`. Becuase otherwise also `undefined` makes it falsy by default - and we dont want to toggle it falsy by default.
2. IMHO semantics of a flag should be **falsy** when disabled (by default) and **truthy** when enabled. 

What about switching the meaning to `avoidCachingErrors`. IMHO it's more attactive and understandable for customers to toggle it on. Moreover when it's falsy (false/undefined) => the new behavior is disabled

core-libs/setup/ssr/optimized-engine/ssr-optimization-options.ts
```diff
@@ -146,7 +146,7 @@ export interface SsrOptimizationOptions {
   /**
    * Enable caching of errors. By default, errors are not cached.
    */
```
Platonn: let update jsdocs

Platonn (approved):

PR #18816, opened 2024-05-09
Feat: [ASM] Adding link to customer list for navigating to customer360

scarai-sap:
See https://jira.tools.sap/browse/CXSPA-6858


feature-libs/asm/components/asm-main-ui/asm-main-ui.component.ts
```diff
@@ -147,13 +148,44 @@ export class AsmMainUiComponent implements OnInit, OnDestroy {
...
+  protected showC360Dialog(customer: User | undefined): void {
+    const data = { customer: customer };
+    this.launchDialogService?.openDialogAndSubscribe(
+      LAUNCH_CALLER.ASM_CUSTOMER_360,
+      this.element,
+      data
+    );
+
+    this.subscription.add(
+      this.launchDialogService?.dialogClose
```
niehuayang: **?** is redundant.

feature-libs/asm/components/asm-main-ui/asm-main-ui.component.ts
```diff
@@ -147,13 +148,44 @@ export class AsmMainUiComponent implements OnInit, OnDestroy {
...
             }
           }
         })
     );
     this.subscribeForDeeplink();
   }
 
+  protected showC360Dialog(customer: User | undefined): void {
+    const data = { customer: customer };
+    this.launchDialogService?.openDialogAndSubscribe(
```
niehuayang: **?** is redundant.

niehuayang (commented):
just two same small comments

feature-libs/asm/components/asm-main-ui/asm-main-ui.component.ts
scarai-sap: copy from below code, I guess yes, we can remove the ?


  createCustomer(): void {
    this.launchDialogService?.openDialogAndSubscribe(
      LAUNCH_CALLER.ASM_CREATE_CUSTOMER_FORM,
      this.addNewCustomerLink
    );
  }

niehuayang: yes, we can also remove it.

I guess that we fristly add 'launchDialogService' to custructor with a breakchange, then to haddle the break change, we had to difine it as 
```
constructor(
............
    protected launchDialogService**?**: LaunchDialogService
  ) {}
```
and in the major version, we don't remove all redundant code.

i53577 (dismissed):

feature-libs/asm/components/asm-main-ui/asm-main-ui.component.ts
scarai-sap: Yes, I guess so

niehuayang (dismissed):

kimhw0630 (approved):

PR #18820, opened 2024-05-09
feat: (CXSPA-1045) - image gallery zoom accessible by keyboard

Pio-Bar:
Ticket: [CXSPA-1045](https://jira.tools.sap/browse/CXSPA-1045)

Adds a zoom-in/zoom-out button to the PDP gallery that can be triggered by a key press.


feature-libs/product/image-zoom/components/product-image-zoom/product-image-zoom-view/product-image-zoom-view.component.ts
```diff
@@ -58,13 +67,16 @@ export class ProductImageZoomViewComponent implements OnInit, OnDestroy {
...
 
+  protected imageLoaded = new BehaviorSubject<boolean>(false);
   protected subscription = new Subscription();
   protected mainMediaContainer$: Observable<ImageGroup | null> =
     this.mainMediaContainer.asObservable();
   protected defaultImageReady$: Observable<boolean> =
     this.defaultImageReady.asObservable();
   protected zoomReady$: Observable<boolean> = this.zoomReady.asObservable();
 
+  @Optional() featureConfigService = inject(FeatureConfigService);
```
Zeyber: Lets use our new knowledge according to the wiki and change this to `private` here ;)
https://wiki.one.int.sap/wiki/pages/viewpage.action?spaceKey=spar&title=Coding+Guidelines#CodingGuidelines-UseaFeatureToggleinourlibraries

feature-libs/product/image-zoom/components/product-image-zoom/product-image-zoom-view/product-image-zoom-view.component.ts
```diff
@@ -381,4 +406,12 @@ export class ProductImageZoomViewComponent implements OnInit, OnDestroy {
 
     return { positionX, positionY };
   }
+
+  onImageLoad() {
+    this.imageLoaded.next(true);
+  }
+
+  get isMobile(): Observable<boolean> {
```
Zeyber: Is this method going to be called on initial page load? Would we be able to avoid using the `breakpointService` if so (maybe css instead)? It can potentially cause SSR and hydration issues and it is critical to avoid these on PDP, PLP and anywhere where SSO is important. The reason for this is that "If you change DOM structure in the client (based on the browser width recognition) and it's different from the DOM structure produced in SSR -> then very likely you'll have layout shifts on initial page load, affecting core web vitals".

Zeyber (changes requested):

feature-libs/product/image-zoom/components/product-image-zoom/product-image-zoom-view/product-image-zoom-view.component.ts
Pio-Bar: done ;)

feature-libs/product/image-zoom/components/product-image-zoom/product-image-zoom-view/product-image-zoom-view.component.ts
Pio-Bar: I've replaced it with CSS since the result is the same. I will keep the `breakpointService` tip in mind for the future, thanks. 👍

Zeyber (approved):

feature-libs/product/image-zoom/components/product-image-zoom/product-image-zoom-view/product-image-zoom-view.component.ts
Zeyber: Thank you very much for doing this!

developpeurweb (approved):

PR #18833, opened 2024-05-13
fix: (CXSPA-975) - Removed unnecessary aria-live attributes

Pio-Bar:
Ticket: [CXSPA-975](https://jira.tools.sap/browse/CXSPA-975)

Removes unnecessary aria-live attributes. No breaking changes here.


Zeyber (approved):

developpeurweb (approved):

PR #18835, opened 2024-05-13
fix: (CXSPA-1103) - Semantic labels for pagination

Pio-Bar:
Ticket: [CXSPA-1103](https://jira.tools.sap/browse/CXSPA-1103)

Adds semantic labels to pagination.


Zeyber (dismissed):

developpeurweb:
It works for JAWS, but for some reason the aria-current state is not being read by VO.
JAWS is our priority so I suggest leaving the attribute there. However, to improve the solution and avoid a future VO ticket… is it also possible to change the aria-label of selected pages to read: “page x, selected” (including the comma and space after the page number)?

Pio-Bar:
The selected page's aria-label should now say "page X , selected". I have moved the logic to the component since it was getting quite lengthy.

projects/storefrontlib/shared/components/list-navigation/pagination/pagination.component.ts
```diff
@@ -76,13 +80,28 @@ export class PaginationComponent {
    * @param type PaginationItemType
    * @returns string
    */
-  getAriaLabel(label?: string, type?: PaginationItemType): string {
+  getAriaLabel(item: PaginationItem): Observable<string> {
```
Zeyber: I'm wary of changing this function so much because of clear breaking changes. But it might be ok if it was only recently introduced. Can you say when this was created? By us?

Zeyber (changes requested):

projects/storefrontlib/shared/components/list-navigation/pagination/pagination.component.ts
Pio-Bar: I've added a feature flag since its an older method.

projects/storefrontlib/shared/components/list-navigation/pagination/pagination.component.html
```diff
@@ -1,13 +1,35 @@
...
-  [class]="item.type"
-  [class.disabled]="isInactive(item)"
-  [class.current]="isCurrent(item)"
-  [routerLink]="pageRoute"
-  [queryParams]="getQueryParams(item)"
-  [tabIndex]="isInactive(item) ? -1 : 0"
-  (click)="pageChange(item)"
-  [attr.aria-label]="getAriaLabel(item.label, item.type)"
+<!-- TODO: (CXSPA-7289) - Remove feature flag next major release -->
+<ng-container *cxFeature="'a11ySemanticPaginationLabel'"
```
Zeyber, line 2: I don't think we need this feature toggle here because we are not including aria attributes as breaking changes.

Zeyber (changes requested):

projects/storefrontlib/shared/components/list-navigation/pagination/pagination.component.html
Pio-Bar: It probably would be best to have it. Since we are calling a new method with the feature flag enabled we would leave the old version without a proper aria-label.

Zeyber (approved):

developpeurweb (approved):

PR #18857, opened 2024-05-16
Fix failing CPQ cypress tests

Larisa-Staroverova:
Both CPQ cypress tests were failing because `Ship to this address` link is not available anymore in the delivery address tile during the checkout process due to this the corresponding method has been removed.
For `POWERTOOLS-SPA`:
image

`ELECTRONICS-SPA`:
image


Larisa-Staroverova:
@ChristophH and @Uli-Tiger, could you make a review?
Thanks a lot in advance!

ChristophHi (approved):

PR #18858, opened 2024-05-16
fix: (CXSPA-987) - Hide stores list buttons from screen reader.

Pio-Bar:
Ticket: [CXSPA-987](https://jira.tools.sap/browse/CXSPA-987)

Stopping excess information from being read in aria-live area proves to be quite tricky. I have just used aria-hidden since the component already uses it to hide the address. This may not be a good idea since we are hiding intractable elements. It may be better to get rid of aria-live here altogether. @developpeurweb


Zeyber (approved):

developpeurweb (changes requested):
Let's try another approach. I wrote a comment on the ticket.

PR #18866, opened 2024-05-17
fix: (CXSPA-1042) - Empty wishlist notification, heading to paragraph change

Pio-Bar:
Ticket: [CXSPA-1042](https://jira.tools.sap/browse/CXSPA-1042)


Zeyber (dismissed):

feature-libs/cart/wish-list/styles/_wish-list.scss
```diff
@@ -21,6 +21,10 @@
     }
   }
 
+  .cx-empty-wish-list {
+    @include type('2', 'normal');
```
developpeurweb: `type(2)` seems quite big in here, let's use the empty default `type()` instead and aligned center please.

default type and centered

Pio-Bar: Done. I've omitted the 'type()' mixin since result would be the same as default paragraph.

developpeurweb (approved):

PR #18874, opened 2024-05-20
feat: Descriptive error messages + i18n fallback keys mechanism

rmch91:
Closes: [CXSPA-1025](https://jira.tools.sap/browse/CXSPA-1025)


feature-libs/order/components/order-confirmation/order-guest-register-form/order-guest-register-form.component.ts
```diff
@@ -20,16 +24,24 @@ import { Subscription } from 'rxjs';
   templateUrl: './order-guest-register-form.component.html',
 })
 export class OrderGuestRegisterFormComponent implements OnDestroy {
+  // TODO: (CXSPA-7315) Remove feature toggle in the next major
+  @Optional() featureConfigService = inject(FeatureConfigService, {
+    optional: true,
+  });
```
Platonn: [NITPICK]

1. Are those 2 really needed?:
- why @Optional()
- why { optional: true }

2. whenever we need to use featureConfigService, let's make it `private` (see https://wiki.one.int.sap/wiki/display/spar/Coding+Guidelines#CodingGuidelines-UsingFeatureTogglesinTypeScript:)

Note: analogical remarks apply to also other files in this PR

feature-libs/asm/components/csagent-login-form/csagent-login-form.component.html
```diff
@@ -13,7 +13,20 @@
       formControlName="userId"
       [attr.aria-label]="'asm.loginForm.userId.label' | cxTranslate"
     />
-    <cx-form-errors [control]="csAgentLoginForm.get('userId')"></cx-form-errors>
+
+    <!-- TODO: (CXSPA-7315) Remove feature toggle in the next major -->
```
Platonn, line 17: [NITPICK]

Let's not mention "in the next major" in the comment
PS. If needed, please also communicate to a11y team to not include such notice about "major".

Why:
Since v2211, we no longer use semver, i.e. we no longer use a word `minor`/`major` etc.
We will simply remove a flag after >=12 months from introducing it, according to our feature flags rollout policy.

btw. Simply the occurrence of the flag's name should suffice to find all its usages in 12 months from now. Not sure if we need to create TODO comments at all ;)

projects/core/src/features-config/feature-toggles/config/feature-toggles.ts
```diff
@@ -239,6 +239,11 @@ export interface FeatureTogglesInterface {
    */
   a11yCloseProductImageBtnFocus?: boolean;
 
+  /**
+   * When enabled use translation fallback keys enabling to use descriptive error message in non-destructive way
+   */
```
Platonn: [MINOR]

Let's explain in details the breaking changes, so customers' developers can quickly understand the impact on their custom codebases. 

```suggestion
  /**
   * In `FormErrorsComponent` it uses more descriptive validation error messages
   * in all UI form fields existing before v2211.25. 
   *
   * 1. The `FormErrorsComponent` uses new i18n keys: 
   * `formErrors.labeled.<validatorName>` instead of `formErrors.<validatorName>`, 
   * for example `formErrors.labeled.required` instead of `formErrors.required`.
   *
   * 2. The existing usages of `CustomFormValidators.passwordValidator` are replaced with
   * an array of new, more specific validators `CustomFormValidators.passwordValidators`
   * (with the plural `...Validators`)
   */
```

projects/core/src/features-config/feature-toggles/config/feature-toggles.ts
```diff
@@ -239,6 +239,11 @@ export interface FeatureTogglesInterface {
...
+  descriptiveErrorMessages?: boolean;
```
Platonn: [MEDIUM]

Let's make the name of this toggle more specific to FormErrorsComponent. The "errorMessages" may appear in various places (global message, or specific UI elements), but in our case it's very specific - we make breaking change related only to the FormErrorsComponent.
What about `formErrorsDescriptiveMessages`?

projects/core/src/i18n/i18next/i18next-translation.service.ts
```diff
@@ -79,15 +101,42 @@ export class I18nextTranslationService implements TranslationService {
...
+
+    if (Array.isArray(keyOrKeys) && descriptiveErrorMessagesEnabled) {
+      return isDevMode()
+        ? `[${keyOrKeys.join(', ')}]`
+        : this.NON_BREAKING_SPACE;
+    } else {
+      return isDevMode() && descriptiveErrorMessagesEnabled
+        ? `[${keyOrKeys}]`
+        : this.NON_BREAKING_SPACE;
+    }
```
Platonn: [MEDIUM]

Do we _really_ need to add a feature toggle here? What's the breaking change we're trying to cover?

projects/storefrontlib/shared/components/form/form-errors/form-errors.component.ts
```diff
@@ -52,7 +52,12 @@ export class FormErrorsComponent implements DoCheck {
   /**
    * Prefix prepended to the translation key.
    */
-  @Input() prefix = 'formErrors';
+  @Input() prefix = 'formErrors.labeled';
+
+  /**
+   * Fallback prefix prepended to the translation key.
+   */
+  @Input() fallbackPrefix = 'formErrors';
```
Platonn: [MAJOR]

IMHO changing the default value of a class' field is a breaking change. It leads to a _new_ behavior -> if customer already customized the original i18n assets, e.g. they changed the wording of the `formErrors.required`, then they expect their custom text of`formErrors.required` to be used. 
But now we will favor our default new values at new keys `formErrors.labeled.required`. So it will be a behavioral breaking change for a customer.

So IMHO we should cover it with a feature toggle:
```ts
@Input() prefix = this.featureConfigService.isEnabled(...) ? 'formErrors' : 'formErrors.labeled';
```

feature-libs/order/components/order-confirmation/order-guest-register-form/order-guest-register-form.component.ts
```diff
@@ -20,16 +24,24 @@ import { Subscription } from 'rxjs';
...
+
+  private passwordValidators = this.featureConfigService?.isEnabled(
```
Platonn: [NITPICK]

Why not `protected`? (see https://wiki.one.int.sap/wiki/display/spar/Coding+Guidelines#CodingGuidelines-Usepublic/protectedinsteadofprivateinclasses)

projects/core/src/i18n/i18next/i18next-translation.service.ts
```diff
@@ -79,15 +101,42 @@ export class I18nextTranslationService implements TranslationService {
...
+  protected getFallbackValue(key: string): string;
+
+  /**
+   * Returns a fallback value in case when the given keys are missing
+   * @param keys
    */
-  protected getFallbackValue(key: string): string {
-    return isDevMode() ? `[${key}]` : this.NON_BREAKING_SPACE;
+  // eslint-disable-next-line @typescript-eslint/unified-signatures
+  protected getFallbackValue(keys: string[]): string;
```
Platonn: [MINOR]

I think we can live with `keyOrKeys: string | string[]`. So IMHO we don't need to split the constructor signature to 2 different signatures.

projects/storefrontapp/src/app/spartacus/spartacus-features.module.ts
```diff
@@ -308,6 +308,7 @@ if (environment.estimatedDeliveryDate) {
         a11yTruncatedTextForResponsiveView: true,
         a11yMyAccountLinkOutline: true,
         a11yCloseProductImageBtnFocus: true,
+        descriptiveErrorMessages: false,
```
Platonn: [QUESTION]

Why not enabled by default in our storefrontapp? (see https://wiki.one.int.sap/wiki/display/spar/Coding+Guidelines#CodingGuidelines-AddaFeatureToggleinthecodesourcecode)

Platonn:
✅ Did a QA of the fallback keys with "lazy loaded i18n" feature enabled and works great!

### QA steps:
1. (prerequisite: make sure that the feature toggle of this PR, i.e. `descriptiveErrorMessages` is enabled in `spartacus-features.module.ts`)
1. generate i18n JSON files in /dist folder AND copy them to the `/assets` folder of the `storefrontapp`
   ```bash
   npm run build:assets
   cp -r dist/assets/i18n-assets projects/storefrontapp/src/assets
   ```
   Then you should see the following file changes (which are needed only for this test) in your git status:
   image


1. configure lazy loading of i18n jsons in `app.module`:
   ```diff
     provideConfig(<I18nConfig>{
        // we bring in static translations to be up and running soon right away
        i18n: {
   -      resources: translations,
   +      backend: {
   +        loader: (lng: string, chunkName: string) =>
   +           import(`../assets/i18n-assets/${lng}/${chunkName}.json`),
   +      },
          chunks: translationChunksConfig,
          fallbackLang: 'en',
        },
      })
   ```

1. open the Network tab ChromeDev tools and filter by the word `json`
   image

1. visit for example Spartacus login page http://localhost:4200/login - and verify that the following assets are loaded (`...common_json.js` and `...user_json.js`):
   image

1. click Sign in button (without filling the form fields)
   image

1. verify that you can see the descriptive error messages, i.e. which pattern to use (e.g, `Field Email address has not a valid email format. Match pattern: example@yourdomain.com`, but not `Field has not a valid email format`) - proving that the **primary** i18n key was applied i.e. `formErrors.labeled.cxInvalidEmail`, with the param `label`
   image

1. in the HTML file of the usage component, e.g. [login-form.component.html#L24-L26](https://github.com/SAP/spartacus/blob/4d7f603db68e115dd52442aadbb3a1e46ee77907/feature-libs/user/account/components/login-form/login-form.component.html#L24-L26), remove the input `[translationParams]="{ label: ... }:`:
   ```diff
     <cx-form-errors
       *cxFeature="'descriptiveErrorMessages'"
   -   [translationParams]="{
   -     label: 'loginForm.emailAddress.label' | cxTranslate
   -   }"
       [control]="form.get('userId')"
     ></cx-form-errors>
   ```

1. refresh the login page in Spartacus. Click the `Sign in` button (without filling the form fields)
1. verify that the descripting error message is STILL displayed, i.e. which pattern to use  (e.g, `Field Email address has not a valid email format. Match pattern: example@yourdomain.com`, but not `Field has not a valid email format`), despite the fact of missing `label` param  - proving that the **primary** i18n key was applied, i.e. `formErrors.labeled.cxInvalidEmail`, even though the param `label` was missing
   image


1. in the file `projects/storefrontapp/src/assets/i18n-assets/en/myAccount.json` add a new i18n key under the `closeAccount` main key:
   ```diff
     {
       "closeAccount": {
   +     "cxInvalidEmail": "SPIKE it works",
         "confirmAccountClosure": "Confirm Account Closure",
   ```

1. in the file [form-errors.component.ts#L55](https://github.com/SAP/spartacus/blob/4d7f603db68e115dd52442aadbb3a1e46ee77907/projects/storefrontlib/shared/components/form/form-errors/form-errors.component.ts#L55) change the `prefix` property:
   ```diff
   - @Input() prefix = 'formErrors.labeled';
   + @Input() prefix = 'closeAccount';
   ```

1. refresh the login page in Spartacus. Click the `Sign in` button (without filling the form fields) and verify that:
   1. verify now you can see the `SPIKE it works` error message under the Email field (see screenshot below) - proving that the **primary** i18n key was used, i.e. `closeAccount.cxInvalidEmail`
   1. verify, that in the Chrome DevTools Network tab you can see that the file `...myAccount_json.js` has been loaded (see screenshot below) - proving that the i18n JSON file with the **primary** i18n key was lazy loaded
   image

1. in the file [form-errors.component.ts#L55](https://github.com/SAP/spartacus/blob/4d7f603db68e115dd52442aadbb3a1e46ee77907/projects/storefrontlib/shared/components/form/form-errors/form-errors.component.ts#L55) bring back the original value of the `prefix` - `formErrors.labeled`:
   ```ts
   @Input() prefix = 'formErrors.labeled';
   ```
    BUT now change the `fallbackPrefix` property to `'closeAccount'`
      ```diff
   - @Input() fallbackPrefix = 'formErrors';
   + @Input() fallbackPrefix = 'closeAccount';
   ```

1. refresh the login page in Spartacus. Click the `Sign in` button (without filling the form fields) and verify that:
   1. verify now you can see the descriptive error message, i.e. which pattern to use (e.g, `Field Email address has not a valid email format. Match pattern: example@yourdomain.com`, but not `SPIKE it works!`) (see screenshot below) - proving that the **primary** i18n key was used, i.e. `formErrors.labeled.cxInvalidEmail`, but not the fallback key `closeAccount.cxInvalidEmail`
   1. verify, that in the Chrome DevTools Network tab you can see that the file `...myAccount_json.js` has been loaded (see screenshot below) - proving that the i18n JSON file with the **fallback** i18n key was lazy loaded, despite the fact it wasn't used (becasue we could use the primary key)
   image

1. In the file `projects/storefrontapp/src/assets/i18n-assets/en/common.json` remove the entry with the key `formFields.labeled.cxInvalidEmail`:
   ```diff
     "formErrors": {
       "labeled": {
         "required": "Field {{label}} is required",
   -     "cxInvalidEmail": "Field {{label}} has not a valid email format.  Match pattern: example@yourdomain.com",
         "cxInvalidPassword": "Field {{label}} has not a valid password format",
   ```

1. refresh the login page in Spartacus. Click the `Sign in` button (without filling the form fields) and verify that:
   1. verify now you can see `SPIKE it works!` but not the descriptive message `Field Email address has not a valid email format. Match pattern: example@yourdomain.com` - proving that the **fallback** i18n key was used, i.e. `closeAccount.cxInvalidEmail`, because the primary key was absent `formErrors.labeled.cxInvalidEmail` (because we just removed it a moment ago)
   1. verify, that in the Chrome DevTools Network tab you can see that the file `...myAccount_json.js` has been loaded (see screenshot below) - proving that the i18n JSON file with the **fallback** i18n key was lazy loaded
   image

feature-libs/order/components/order-confirmation/order-guest-register-form/order-guest-register-form.component.ts
rmch91: Changed

projects/core/src/features-config/feature-toggles/config/feature-toggles.ts

projects/core/src/features-config/feature-toggles/config/feature-toggles.ts

projects/storefrontlib/shared/components/form/form-errors/form-errors.component.ts
rmch91: Changed

feature-libs/order/components/order-confirmation/order-guest-register-form/order-guest-register-form.component.ts
rmch91: Changed

projects/storefrontapp/src/app/spartacus/spartacus-features.module.ts
rmch91: Changed

projects/core/src/i18n/i18next/i18next-translation.service.ts
rmch91: Yeah, seems like it was redundant. Removed and did a small refactor

projects/core/src/i18n/i18next/i18next-translation.service.ts
rmch91: overloading removed

feature-libs/asm/components/csagent-login-form/csagent-login-form.component.html
rmch91: Did not change yet, thinking about how we should handle this to have more or less consistent approach everywhere.

1. Look for "major" in the whole project and remove it?
2. How to track which exactly feature flags should be removed? I think that not only me, thinking that we will remove all the feature flags in December/January, but actually you have right, we should remove only 12 months old flags. So maybe thats true, that we don't need todos, seems like they will not help.

feature-libs/asm/components/csagent-login-form/csagent-login-form.component.html
```diff
@@ -28,6 +41,9 @@
       cxPasswordVisibilitySwitch
     />
     <cx-form-errors
+      [translationParams]="{
+        label: 'asm.loginForm.password.label' | cxTranslate
+      }"
       [control]="csAgentLoginForm.get('password')"
     ></cx-form-errors>
```
Platonn: [MEDIUM]
Doesn't this case need `*cxFeature=ourFlag` + a counterpart `<cx-form-errors *cxFeature=...`> with `!ourFlag`? 0 as in all other examples?

Note: an analogical remark applies to some places in the files:
```
feature-libs/organization/administration/components/unit/form/unit-form.component.html
feature-libs/organization/administration/components/unit/links/addresses/form/unit-address-form.component.html
```

If possible, please also double-check if I didn't miss any analogical place.

projects/core/src/i18n/i18next/i18next-translation.service.ts
```diff
@@ -76,18 +94,23 @@ export class I18nextTranslationService implements TranslationService {
     return this.i18next.loadNamespaces(chunkNames);
   }
 
-  /**
-   * Returns a fallback value in case when the given key is missing
-   * @param key
-   */
-  protected getFallbackValue(key: string): string {
-    return isDevMode() ? `[${key}]` : this.NON_BREAKING_SPACE;
+  protected getFallbackValue(keyOrKeys: string | string[]): string {
```
Platonn, line 100:
```suggestion
  /**
   * Returns a fallback value in case when the all given keys are missing
   */
  protected getFallbackValue(keyOrKeys: string | string[]): string {
```

projects/core/src/util/regex-pattern.ts
```diff
@@ -9,4 +9,13 @@ export const EMAIL_PATTERN =
   /^(([^<>()\[\]\\.,;:\s@"]+(\.[^<>()\[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/; // eslint-disable-line
 
 export const PASSWORD_PATTERN =
-  /^(?=.*?[A-Z])(?=.*?[0-9])(?=.*?[!@#$%^*()_\-+{};:.,]).{6,}$/;
+  /^(?=.*?[A-Z])(?=.*?\d)(?=.*?[!@#$%^*()_\-+{};:.,]).{6,}$/;
```
Platonn, line 12: [QUESTION]

What do you think about @deprecating in JSDocs this const? + leave a code comment that it should be removed at the moment when we'll remove the feature toggle `formErrorsDescriptiveMessagges` - which will happen in 12 months from now.

projects/core/src/features-config/feature-toggles/config/feature-toggles.ts
```diff
@@ -244,6 +244,20 @@ export interface FeatureTogglesInterface {
    */
   a11yCloseProductImageBtnFocus?: boolean;
 
+  /**
+   * In `FormErrorsComponent` it uses more descriptive validation error messages
+   * in all UI form fields existing before v2211.25.
```
```suggestion
   * in all UI form fields existing before v2211.24.
```

Platonn: My bad. 2211.25 was correct!
Resolving this thread

projects/core/src/util/regex-pattern.ts
rmch91: Changed

projects/core/src/i18n/i18next/i18next-translation.service.ts
rmch91: Changed

feature-libs/asm/components/csagent-login-form/csagent-login-form.component.html
rmch91: Thanks, found additional occurrences. Changed.

developpeurweb:
Great work @rmch91!
Can you please also apply the same solution to the _Checkout > Shipping Address_ form (as you did with the Payment form). I still see the old generic alerts for Shipping.

developpeurweb (dismissed):
all good on my end

feature-libs/organization/administration/components/budget/form/budget-form.component.html
```diff
@@ -26,24 +39,60 @@
...
+      ></cx-date-picker>
+
+      <!-- TODO: (CXSPA-7315) Remove feature toggle in the next major -->
+      <cx-date-picker
+        *cxFeature="'formErrorsDescriptiveMessages'"
         [control]="$any(form.get('startDate'))"
         [max]="form.get('endDate')?.value"
         [required]="true"
         (update)="form.get('endDate')?.updateValueAndValidity()"
+        label="orgBudget.startDate"
```
Zeyber, line 67: Just a side note, I don't think we need to feature toggle the date picker to add the `label`. We don't consider aria-related changes as breaking changes.

projects/storefrontlib/shared/components/form/form-errors/form-errors.component.ts
```diff
@@ -34,6 +35,8 @@ import { map, startWith } from 'rxjs/operators';
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
 export class FormErrorsComponent implements DoCheck {
+  private featureConfigService = inject(FeatureConfigService);
```
Zeyber, line 38: I think this should be a protected Optional() injection.

projects/core/src/i18n/translatable.ts
```diff
@@ -5,11 +5,15 @@
...
+  key?: string | string[];
   params?: TranslatableParams;
   raw?: string;
 }
 
 export interface TranslatableParams {
   [param: string]: any;
 }
+
+export function isTranslatable(input: any): input is Translatable {
```
Zeyber, line 17: We should probably cover this with a unit test.

projects/storefrontlib/shared/components/form/form-errors/form-errors.component.ts
rmch91: We have documented here https://wiki.one.int.sap/wiki/display/spar/Coding+Guidelines#CodingGuidelines-UsingFeatureTogglesinTypeScript (2.18.5.1. Using Feature Toggles in TypeScript:) to use `private` access modifier for feature config service.

image

If to talk about Optional(), I don't see the reason to use it. I think it is not the same case as with constructor. We are declaring this service as a new private property of the class. So it always will be there. Please, let me know if you see any cases for usage of the optional decorator.

projects/core/src/i18n/translatable.ts
rmch91: Add. Thanks for unit tests found one problem in this method :) 
```
typeof input === 'object' &&
    ('key' in input || 'raw' in input)
```

typeof null is also `object` so the part `'key' in input` could make problems.

feature-libs/organization/administration/components/budget/form/budget-form.component.html
rmch91: You've right. Changed

feature-libs/asm/components/csagent-login-form/csagent-login-form.component.html
Platonn: yes, in a separate PR maybe. so people don't copy-paste it blindly to new places

projects/storefrontlib/shared/components/form/form-errors/form-errors.component.spec.ts
```diff
@@ -13,15 +17,31 @@ describe('FormErrors', () => {
...
+  let featureConfigService: jasmine.SpyObj<FeatureConfigService>;
 
   const getContent = () => fixture.debugElement.nativeElement.innerText;
 
   beforeEach(
     waitForAsync(() => {
+      const featureConfigServiceMock = jasmine.createSpyObj(
+        'FeatureConfigService',
+        ['isEnabled']
+      );
```
Platonn: [Nitpick]
It lacks type safety. if ever in the future we'd rename the method, this place will silently fail to rename.
What about `const featureConfigServiceMock: Partial<FeatureConfigService> = ...`?

Platonn (dismissed):

feature-libs/asm/components/csagent-login-form/csagent-login-form.component.html
Platonn: yes, in a separate PR maybe. so people don't copy-paste it blindly to new places

projects/storefrontlib/shared/components/form/form-errors/form-errors.component.ts
Zeyber: Ah I didn't know about this. Thanks!

Zeyber (dismissed):

projects/storefrontlib/shared/components/form/form-errors/form-errors.component.spec.ts
rmch91: Thanks, changed

Zeyber (approved):

PR #18880, opened 2024-05-22
CXSPA-7325: Add scenarios which could not run on pipeline set into WTF D3 run

LindaSAP:
new sh file to add scenarios which could not run on pipeline set into WTF D3 run


scarai-sap (dismissed):
LGTM

run-spartacus-cust-e2es.sh
```diff
@@ -0,0 +1,20 @@
...
+CYPRESS_MYACCOUNT_V2_FOLDER="$CYPRESS_ROOT_FOLDER/regression/my-account"
+CYPRESS_ACCESSIBILITY_FOLDER="$CYPRESS_ROOT_FOLDER/accessibility"
+
+export E2ES_TO_RUN="$CYPRESS_OTP_FOLDER/otp-login.e2e-flaky.cy.ts,$CYPRESS_MYACCOUNT_V2_FOLDER/my-account-v2-consent-management.e2e.cy.ts,$CYPRESS_MYACCOUNT_V2_FOLDER/my-account-v2-email.e2e.cy.ts,$CYPRESS_MYACCOUNT_V2_FOLDER/my-account-v2-notification-preference.e2e.cy.ts,$CYPRESS_MYACCOUNT_V2_FOLDER/my-account-v2-password.e2e.cy.ts,$CYPRESS_MYACCOUNT_V2_FOLDER/my-account-v2-profile.e2e.cy.ts"
+
+# install cypress standlone dependencies
+(cd projects/storefrontapp-e2e-cypress && npm install)
+
+# run spartacus ccv2 e2es for b2c
+npm run e2e:run:ci:ccv2
```
i53577, line 19: `--env CX_MY_ACCOUNT_V2=true` need to be added here. otherwise some test case can't pass.
https://github.com/SAP/spartacus/blob/develop/projects/storefrontapp-e2e-cypress/package.json#L24

i53577: or add `export CX_MY_ACCOUNT_V2=true` in this file, this way we don't need to change the central config. But not sure if this works, you can have a try.

i53577 (approved):

kimhw0630 (approved):

PR #18881, opened 2024-05-22
feat: (CXSPA-1143) - Manual handling of the focus for the facets list

sdrozdsap:


projects/storefrontlib/cms-components/product/product-list/product-facet-navigation/facet-list/facet-list.component.ts
```diff
@@ -49,20 +59,35 @@ export class FacetListComponent {
   @Output() closeList = new EventEmitter();
 
   /** The list of all facet and values related to the products in the list */
-  facetList$: Observable<FacetList> = this.facetService.facetList$;
+  facetList$: Observable<FacetList> = this.facetService.facetList$.pipe(
+    tap((facetList) => {
```
Zeyber: I think we should move where we tap this observable instead of where it is instantiated because now it is doing more than just returning its observable. Are we able to tap `this.facetList$` in `ngOnInit`, `ngOnViewInit` or some other place like that?

projects/storefrontlib/cms-components/product/product-list/product-facet-navigation/facet-list/facet-list.component.ts
```diff
@@ -49,20 +59,35 @@ export class FacetListComponent {
...
+  );
 
   iconTypes = ICON_TYPE;
 
   dialogFocusConfig: FocusConfig = {
     trap: true,
     block: true,
     focusOnEscape: true,
-    autofocus: 'cx-facet',
+    autofocus: 'cx-facet > button',
```
Zeyber, line 73: More just a question, cx-facet is not focusable as-is so autofocus usually has no effect here?

Zeyber (dismissed):

projects/storefrontlib/cms-components/product/product-list/product-facet-navigation/facet-list/facet-list.component.ts
sdrozdsap: Sure, OnInit will be a better place for that

projects/storefrontlib/cms-components/product/product-list/product-facet-navigation/facet-list/facet-list.component.ts
sdrozdsap: Yes, if you will open a demo page and open the filters dialog then nothing will be autofocused, I figured it shouldn't work that way so now the first facets will be focused automatically

Zeyber: That's what I thought. Thanks for clarifying :)

projects/storefrontlib/cms-components/product/product-list/product-facet-navigation/facet-list/facet-list.component.ts
```diff
@@ -49,27 +60,46 @@ export class FacetListComponent {
...
+  });
 
   constructor(
     protected facetService: FacetService,
     protected elementRef: ElementRef,
     protected renderer: Renderer2
   ) {}
 
+  ngOnInit(): void {
+    this.facetList$ = this.facetService.facetList$.pipe(
```
Zeyber: I'm still not really sure about reassigning this property. Perhaps we can tie it to a subscription instead?

Zeyber (changes requested):

projects/storefrontlib/cms-components/product/product-list/product-facet-navigation/facet-list/facet-list.component.ts
sdrozdsap: What do you mean by "reassigning"? This is the first tijme this property will be assigned a value(I've remove initialization in line 63. 
I didn't want to create a separate subscription for it since it seemed a bit redundant. Since we already have the subcription for the facetList$ observable using async pipe, I decided it will be slightly better to perform this side effect here instead of creating a separate subscription. What do you think?

Zeyber: The reason for subscribing to the observable instead is to avoid the observable itself `facetList$` also being responsible for the `tap` whenever it is used by anything that may also use this component. Having the Observable be "dumb" could help us avoid `handleDialogFocus` being called more than once if the `facetList$` observable is used by any external component of service.

I think you might be right in saving performance by giving it the extra function and I think this would be fine in the main case. But I wonder if it could cause issues because of the reason above which could ultimately hinder performance or cause bugs. If we move this to a subscription instead in the component, I believe we would only call `handleDialogFocus` once on any `this.facetService.facetList$` changes instead of potentially more times if the component's property is observed.

This is my opinion though and I am happy to discuss this further :)

sdrozdsap: You convinced me, I will move this to a separate subscription

projects/storefrontlib/cms-components/product/product-list/product-facet-navigation/facet-list/facet-list.component.ts
```diff
@@ -57,19 +70,27 @@ export class FacetListComponent {
...
+  });
 
   constructor(
     protected facetService: FacetService,
     protected elementRef: ElementRef,
     protected renderer: Renderer2
   ) {}
 
+  ngOnInit(): void {
+    this.subscribeToFocusUpdates();
```
Zeyber: Lets move the feature flag here instead to make our method more simple.

```suggestion
    if(this.featureConfigService?.isEnabled('a11yFacetsDialogFocusHandling')) {
        this.subscribeToFocusUpdates();
    }
```

projects/storefrontlib/cms-components/product/product-list/product-facet-navigation/facet-list/facet-list.component.ts
```diff
@@ -109,4 +130,55 @@ export class FacetListComponent {
...
+        const isFeatureEnabled = this.featureConfigService?.isEnabled(
+          'a11yFacetsDialogFocusHandling'
+        );
+        if (!isFeatureEnabled) {
+          return;
+        }
+        this.handleDialogFocus(facetList.facets);
+      }
+    );
+    this.subscriptions.add(updateFocusSub);
```
Zeyber: Now we can make this method simpler.

```suggestion
    this.subscriptions.add(
      this.facetService.facetList$.subscribe(
        (facetList) =>
          this.handleDialogFocus(facetList.facets);
      )
    );
```

projects/storefrontlib/cms-components/product/product-list/product-facet-navigation/facet-list/facet-list.component.ts
```diff
@@ -109,4 +130,55 @@ export class FacetListComponent {
   block(event?: MouseEvent) {
     event?.stopPropagation();
   }
+
+  protected subscribeToFocusUpdates(): void {
```
Zeyber: I think this method could be something more meaningful like:
```suggestion
  protected enableFocusHandlingOnFacetListChanges(): void {
```

Zeyber (changes requested):

sdrozdsap:
Thanks @Zeyber  for useful tips! I've applied the changes you recommended

Zeyber (dismissed):

developpeurweb (dismissed):

Zeyber (approved):

PR #18890, opened 2024-05-23
fix: (CXSPA-1124) - Navigation-ui mobile keyboard improvements

Pio-Bar:
Ticket: [CXSPA-1124](https://jira.tools.sap/browse/CXSPA-1124)

Improves keyboard controls on mobile i.e:
- When expanding a node focus moves inside it.
- When hamburger menu is opened the first node inside it gets focus.
- It should no longer be possible to be stuck on the "Back" button.
- When collapsing a node the focus should return to the node's header.

We should be able to include those fixes within 'a11yNavigationUiKeyboardControls'. This PR only brings improvements to earlier work.


projects/storefrontlib/cms-components/navigation/navigation/navigation-ui.component.ts
```diff
@@ -110,6 +110,26 @@ export class NavigationUIComponent implements OnInit, OnDestroy {
...
+  /**
+   * Focus on the first focusable element in the hamburger menu when the menu is opened.
+   */
+  focusOnMenuExpansion(): void {
+    this.subscriptions.add(
+      this.hamburgerMenuService?.isExpanded.pipe().subscribe((isExpanded) => {
+        if (isExpanded && this.navAriaLabel?.includes('menu')) {
+          setTimeout(() => {
+            this.elemRef.nativeElement.querySelector('[tabindex="0"]').focus();
+          }, 0);
```
Zeyber: Just a question, no need to change anything. Do we still need to pass `0` to `setTimeout`? Does it work fine without it?

projects/storefrontlib/cms-components/navigation/navigation/navigation-ui.component.ts
```diff
@@ -110,6 +110,26 @@ export class NavigationUIComponent implements OnInit, OnDestroy {
...
+      this.focusOnMenuExpansion();
+    }
+  }
+
+  /**
+   * Focus on the first focusable element in the hamburger menu when the menu is opened.
+   */
+  focusOnMenuExpansion(): void {
+    this.subscriptions.add(
+      this.hamburgerMenuService?.isExpanded.pipe().subscribe((isExpanded) => {
```
Zeyber: Do we need to add the empty `pipe()` here?

projects/storefrontlib/cms-components/navigation/navigation/navigation-ui.component.ts
Pio-Bar: Should work fine without and since it defaults to zero, I've removed it. ;)

projects/storefrontlib/cms-components/navigation/navigation/navigation-ui.component.ts
Pio-Bar: nope, removed

Zeyber (approved):

developpeurweb (approved):

PR #18904, opened 2024-05-27
cxspa-4006: implement CHAPTCHA framework with mock CHAPTCHA service

scarai-sap:
Background

The original feature defined as https://jira.tools.sap/browse/CXSPA-805, However, for security reason, we’ll not provide google as default implementation so we decided to provide a mock implementation as https://jira.tools.sap/browse/CXSPA-4006

Also see help doc in 

https://help.sap.com/docs/INTERNAL_COMPOSABLE_STOREFRONT/eaef8c61b6d9477daf75bff9ac1b7eb4/4605061113284c2c92cc558600140aaa.html?locale=en-US&state=DRAFT#loio4605061113284c2c92cc558600140aaa

For local testing, you can add below code to enable  in code captcha.service.ts to override the getting config from server.

`
  fetchCaptchaConfigFromServer(): Observable<CaptchaConfig> {
    return of({
      enabled: true,
      publicKey: '123',
    });
  }
`


projects/storefrontapp-e2e-cypress/cypress/e2e/regression/user_access/mockCaptcha.core-e2e.cy.ts
```diff
@@ -0,0 +1,42 @@
...
+import { viewportContext } from '../../../helpers/viewport-context';
+import { user } from '../../../sample-data/checkout-flow';
+
+describe('Register', () => {
+  viewportContext(['desktop', 'mobile'], () => {
+    before(() => {
+      cy.window().then((win) => win.sessionStorage.clear());
+    });
+
+    it('should register and redirect to login page (CXSPA-805)', () => {
```
i53577: update the ticket number

projects/storefrontlib/shared/components/captcha/mockCaptcha/config/captcha-api-config.ts
```diff
@@ -0,0 +1,22 @@
...
+import { Injectable, Type } from '@angular/core';
+import { Config } from '@spartacus/core';
+import { CaptchaProvider } from '../../captcha.model';
+
+@Injectable({
+  providedIn: 'root',
+  useExisting: Config,
+})
+export abstract class CaptchaApiConfig {
+  captchaProvider?: Type<CaptchaProvider>;
```
i53577: This file should be a part of template, maybe move it from mock folder to projects/storefrontlib/shared/components/captcha/config/captcha-api-config.ts

And I would say let's keep `apiUrl` and `fields` in CaptchaApiConfig

scarai-sap: fixed

projects/storefrontapp-e2e-cypress/cypress/e2e/regression/user_access/mockCaptcha.core-e2e.cy.ts
```diff
@@ -0,0 +1,42 @@
+/*
```
Zeyber: Lets keep file naming standards consistent: `mock-captcha.core-e2e.cy.ts`

projects/storefrontlib/shared/components/captcha/mockCaptcha/mock-captcha.service.ts
```diff
@@ -0,0 +1,84 @@
+/*
```
Zeyber: Can we rename this directory to `.../mock-captcha/...`?

projects/storefrontapp-e2e-cypress/cypress.config.ts
```diff
@@ -9,6 +9,7 @@ import { defineConfig } from 'cypress';
 export default defineConfig({
   defaultCommandTimeout: 10000,
   requestTimeout: 15000,
+  chromeWebSecurity: false,
```
Zeyber, line 12: I know that we likely need to disable `chromeWebSecurity` due to captcha requiring third-parties. But is there a way to isolate it to our captcha tests rather than all tests?

Zeyber: I'm also not sure if this will work in ci e2es since this is only the dev env config. On second thought, if we are using the mock service now, do we need this flag?

Zeyber (changes requested):
I noticed that the e2e tests are failing. It looks like it is due to an a11y update (some text has changed). Would it be possible to get the false positive on sonar accepted to run this pr against ci?

projects/storefrontapp-e2e-cypress/cypress.config.ts
scarai-sap: It already used in https://github.com/SAP/spartacus/blob/6ddc4263feaad8fb1d275ffa06cbcc9f03b815b5/projects/storefrontapp-e2e-cypress/cypress.config.ci.ts#L16 

For me, I need this change, or the server will reject our request as we have. "commerce-cloud-captcha-token" which need to configure to server's allow list.  See  https://help.sap.com/docs/INTERNAL_COMPOSABLE_STOREFRONT/eaef8c61b6d9477daf75bff9ac1b7eb4/4605061113284c2c92cc558600140aaa.html?locale=en-US&state=DRAFT#loio4605061113284c2c92cc558600140aaa

scarai-sap: I have tried to find a solution to dynamic change this settings, looks no way  :(

Zeyber: Ok. If its in the ci config already, I think its fine then.

projects/core/src/model/misc.model.ts
```diff
@@ -128,4 +128,10 @@ export interface BaseSite {
   baseStore?: BaseStore;
   requiresAuthentication?: boolean;
   isolated?: boolean;
+  captchaConfig?: CaptchaConfig;
+}
+
+export interface CaptchaConfig {
```
Zeyber, line 134: I had a look at this and the instructions. It doesn't seem to be indicated how the user can turn on Captcha with this part of the config (ie. set `enabled` to true). I also wonder why the users need to override the CaptchaModule's default config. Wouldn't it be easier for them to enable and provide `CaptchaConfig` in their own spartacus config module like we do in `SpartacusB2cConfigurationModule`?

Zeyber (changes requested):

projects/core/src/model/misc.model.ts
scarai-sap: As you known, Whether captcha is enabled or not is defined in backend and read it via OCC API. We need a configuration to present what's we get from backend.  The is Enabled is a mandatory filed of it. PublicKey is set as optional as extra parameters.

Check the end points"/getBaseSites"  https://api.cg79x9wuu9-eccommerc1-d2-public.model-t.myhybris.cloud/occ/v2/swagger-ui/index.html#/Base%20Sites/getBaseSites.  you can see our OOTB baseSIteInfo contains "enabled" and "publicKey".  


image

 If customer wanna to integrate their backend, they may add other things more than publicKey(like URL, domain, ...) But we just define it as what OCC API provided.

scarai-sap: For your question, how to enable, it mentioned in help doc as "Requirements" https://help.sap.com/docs/INTERNAL_COMPOSABLE_STOREFRONT/eaef8c61b6d9477daf75bff9ac1b7eb4/4605061113284c2c92cc558600140aaa.html?locale=en-US&state=DRAFT#loio4605061113284c2c92cc558600140aaa

You have enabled CAPTCHA for your site. For more information, see [CAPTCHA OCC APIs](https://help.sap.com/docs/INTERNAL_COMPOSABLE_STOREFRONT/eaef8c61b6d9477daf75bff9ac1b7eb4/8a42681cd28549b6a81d6faffc7d6b92.html?locale=en-US&state=DRAFT).

scarai-sap: The full picture of the CAPTCHA should looks as below

1) The feature toogle is in backend;
2) If CAPTCHA enabled, in SPA side, we'll render CAPTCHA component, and  generate token as designed;
3) SPA call APIs with adding CAPTCHA token in the header;
4) The backend side, need add annotation if one API need check the token; then the annotated API will check CAPTCHA token as part of authentication before business process.

In code in SPA, we only can see 2 and 3, Which make it little complex. But I'm not sure where we can make a overall picture in help doc as now we just put a linkage from SPA doc to Commerce(not sure that's enough)

Zeyber: So maybe there is some confusion, sorry. There was two points here I would like to address:

- I understand that it needs to be enabled in the backend and is a baseSite endpoint. But it isn't clear how to enable this in Spartacus from the documentation. Maybe was meant to be in [this link ("see OCC CAPTCHA APIs")](https://help.sap.com/docs/INTERNAL_COMPOSABLE_STOREFRONT/eaef8c61b6d9477daf75bff9ac1b7eb4/8a42681cd28549b6a81d6faffc7d6b92.html?locale=en-US&state=DRAFT) which seems to be dead? 

- I also think its better to tell customers to override the default captcha config from their own configuration module. I think it is unnecessary to override the CaptchaModule itself. I think this is our standard practice. Is there a reason to do it the way in the doc? Maybe it isn't meant to be customer facing?

scarai-sap: The doc update will be handled by UA in coming workdays.

Let's check it again once ready.

projects/storefrontlib/shared/components/captcha/captcha-api-config.ts
```diff
@@ -0,0 +1,24 @@
...
+
+import { Injectable, Type } from '@angular/core';
+import { Config } from '@spartacus/core';
+import { CaptchaProvider } from './captcha.model';
+
+@Injectable({
+  providedIn: 'root',
+  useExisting: Config,
+})
+export abstract class CaptchaApiConfig {
```
Zeyber, line 26: We should add doc comments to these since they will be user facing.

feature-libs/user/profile/occ/adapters/occ-user-profile.adapter.ts
```diff
@@ -39,7 +41,9 @@ export class OccUserProfileAdapter implements UserProfileAdapter {
   constructor(
     protected http: HttpClient,
     protected occEndpoints: OccEndpointsService,
-    protected converter: ConverterService
+    protected converter: ConverterService,
+    protected captchaConfig?: CaptchaApiConfig,
```
Zeyber: Lets use `inject` to include these services instead like the `LoggerService` above. This makes it easier for us to use our schematics when upgrading (ie. its easier to bump injectors than modify constructor).

```
protected captchaConfig = inject(CaptchaApiConfig, { optional: true});
protected injector = inject(Injector, { optional: true});
```
Note: Optional flag would be removed when we bump major versions.

Zeyber (changes requested):

feature-libs/user/profile/occ/adapters/occ-user-profile.adapter.ts
scarai-sap: Thanks, Changed.

projects/storefrontlib/shared/components/captcha/captcha-api-config.ts
scarai-sap: Added

giancorderoortiz (changes requested):
chromeWebSecurity in projects/storefrontapp-e2e-cypress/cypress.config.ts must not be disabled globally.

scarai-sap:
> chromeWebSecurity in projects/storefrontapp-e2e-cypress/cypress.config.ts must not be disabled globally.


As discussed in the channel, is already used in ci configuration. 
As discussed in previous with Caine, there is no way to set to enable/disable it dynamically. :( .
So if we remove it, we have to skip/ignore some e2e tests(to ensure the login succeed with captcha). Is that Acceptable?

projects/storefrontlib/shared/components/captcha/captcha-api-config.ts
Zeyber: I mean for the properties too. For example, its not really clear what some might use `fields` for.

Zeyber (dismissed):

scarai-sap:
@giancorderoortiz ,  May you check it again?

giancorderoortiz (approved):
ok

projects/storefrontlib/shared/components/captcha/captcha.model.ts
```diff
@@ -0,0 +1,35 @@
...
+}
+
+export interface CaptchaProvider {
+  /**
+   * Renders captcha widget
+   *
+   * @params parameters required by a provider
+   * @returns Observable for the key returned by a provider once user validated
+   */
+  renderCaptcha(params: RenderParams): Observable<string>;
```
meeque, line 23: Should rendering really be part of a model?

projects/storefrontlib/shared/components/captcha/captcha.service.ts
```diff
@@ -0,0 +1,102 @@
...
+declare global {
+  interface Window {
+    onCaptchaLoad: () => void;
+  }
+}
+
+@Injectable({
+  providedIn: 'root',
+})
+export abstract class CaptchaService implements CaptchaProvider, OnDestroy {
```
meeque, line 34: I find the break-down into `CaptchaProvider`, `CaptchaService`, and `CaptchaComponent` extremely confusing.

Shouldn't a service be reusable by multiple components? This `CaptchaService` holds `CaptchaProvider` model data that could get be associated with multiple instances of `CaptchaComponent`. Customers might also inject this `CaptchaService` into other components altogether.

Depending on the implementation of `CaptchaService`, the same `token` property could get set twice with differnt values. I guess this might work in the end, since we don't care about a specific `token`, just any `token` that the backend (OCC API) can verify against the captcha service successfully. Still this whole design looks very weird to me.

projects/storefrontlib/shared/components/captcha/captcha.service.ts
```diff
@@ -0,0 +1,102 @@
...
+
+  getCaptchaConfig(): Observable<CaptchaConfig> {
+    return this.captchaConfigSubject$.asObservable();
+  }
+
+  /**
+   * Trigger rendering function configured in CaptchaApiConfig
+   * @param {HTMLElement} elem - HTML element to render captcha widget within.
+   */
+  abstract renderCaptcha(renderParams: RenderParams): Observable<string>;
```
meeque, line 85: Looks like customers will have to implement this method themselves in a custom sub-class. (There's a default implementation in `MockCaptchaService`, but that does not seem to do anything that would be useful for integrating with a real Captcha service provider.)

Imho this abstract method deserves better documentation. From a security perspective, the docs should clearly mention that implementors are responsible for preventing XSS when modifying the `element` in the `RenderParams`.

projects/storefrontlib/shared/components/captcha/captcha.service.ts
```diff
@@ -0,0 +1,102 @@
...
+    ]).pipe(map((result) => result[1]?.captchaConfig as CaptchaConfig));
+  }
+
+  getCaptchaConfig(): Observable<CaptchaConfig> {
+    return this.captchaConfigSubject$.asObservable();
+  }
+
+  /**
+   * Trigger rendering function configured in CaptchaApiConfig
+   * @param {HTMLElement} elem - HTML element to render captcha widget within.
```
meeque, line 83: This parameter does not exist.

I guess this documentation refers to the `element` property of `RenderParams` interface. Maybe that interface needs better documentation, the the docs of the `renderParams` param could just reference it?

projects/storefrontlib/shared/components/captcha/captcha.service.ts
meeque: These help portal docs mention how to implement a custom `CaptchaService`:

https://help.sap.com/docs/INTERNAL_COMPOSABLE_STOREFRONT/eaef8c61b6d9477daf75bff9ac1b7eb4/4605061113284c2c92cc558600140aaa.html?locale=en-US&state=DRAFT#loiofea8e9b575ef447480ee90b5c1b2

But they don't even mention how to implement the `renderCaptcha` method. Not very helpful.

projects/storefrontlib/shared/components/captcha/captcha.model.ts
scarai-sap: yep, maybe a new abstract class for  CaptchaProvider is better

projects/storefrontlib/shared/components/captcha/captcha.service.ts
scarai-sap: Yep, they are complex and confusing. 
CaptchaComponent : Angular component which could be inserted in different places
CaptchaService: Abstract class which provide fetch settings from backend and other common features
CaptchaProvider: Which contains configuration of how it rendering, part of CaptchaApiConfig


> Depending on the implementation of CaptchaService, the same token property could get set twice with differnt values. I guess this might work in the end, since we don't care about a specific token, just any token that the backend (OCC API) can verify against the captcha service successfully. Still this whole design looks very weird to me.

Yep,  same/different tokens may set many times based on how/whether the rendering is called and fetch token from server side, we always communicate with Commerce backend with latest one.

projects/storefrontlib/shared/components/captcha/captcha.service.ts
scarai-sap: Ah, the code was changed, but the doc is not.

projects/storefrontlib/shared/components/captcha/captcha.service.ts
scarai-sap: It's complex for different Captcha provider may has different requirements, some of them need a click, others may need input. It's little complex to have sth in here. Maybe we can write a blog for it.

scarai-sap (commented):
Thanks for your valuable comments. 

@meeque 

I hope we can get them before merging it. As the code is released, I guess we can only do very small changes for compatibility reason

projects/storefrontlib/shared/components/captcha/captcha.service.ts
meeque: Sure, the details differ, but the concepts are often similar. And imho the generic captcha plugin-point on the OCC side and this `CaptchaService` are suitable for integration with many different captcha service providers.

But since we're leaving the captcha service integration to the customer, we should provide better documentation. And ideally provide working code samples for at least one or two captcha service providers. (This should include reCAPTCHA, since that's what the classic accelerators supported.)

Even if we don't ship an out-of-the-box captcha service integration with Spartacus, it would be great to have e2e tests with at least one real-life captcha service. Just to demonstrate feasibility of the framework that we provide.

PR #18923, opened 2024-05-31
fix: workaround for the features option schematics error

kpawelczak:
closes: [CXSPA-7456](https://jira.tools.sap/browse/CXSPA-7456)

This pull request addresses the issue detailed in [Angular CLI GitHub issue #16320](https://github.com/angular/angular-cli/issues/16320). The problem is that using ng add with options of type array may not function correctly in some cases, and might only work on the second attempt.

To resolve this, we have implemented a workaround in the form of a new utility function, getFeaturesOptions. This function ensures that feature options are handled correctly, mitigating the problem described in the issue.

Changes:
- Added `getFeaturesOptions` function to work around the issue with ng add and array options.
- Changed the features type in schema.json

### DIFF
--- a/projects/schematics/src/shared/utils/feature-utils.ts
+++ b/projects/schematics/src/shared/utils/feature-utils.ts
@@ -484,3 +484,29 @@ function createDependentFeaturesLog(
     ', '
   )}: ${notSelectedFeatures.join(', ')}\n`;
 }
+
+/**
+ * This function is a workaround for the GitHub issue
+ * https://github.com/angular/angular-cli/issues/16320.
+ * Using `ng add` with options of type array (--features)
+ * may not work correctly in some situations, and might only work on the
+ * second attempt.
+ */
+export function getFeaturesOptions<OPTIONS extends LibraryOptions>(
+  options: OPTIONS
+): string[] {
+  return getArrayOptions(options.features ?? []);
+}
+
+export function getArrayOptions(values: string[] | string): string[] {
+  let optionsArray: string[] = [];
+  if (values && Array.isArray(values)) {
+    optionsArray = values;
+  } else if (values.includes(',')) {
+    optionsArray = values.split(',');
+  } else {
+    optionsArray = [values];
+  }
+
+  return optionsArray;
+}
--- a/projects/schematics/src/shared/utils/feature-utils.ts
+++ b/projects/schematics/src/shared/utils/feature-utils.ts
@@ -492,13 +492,13 @@ function createDependentFeaturesLog(
  * may not work correctly in some situations, and might only work on the
  * second attempt.
  */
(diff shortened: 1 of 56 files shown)


projects/schematics/src/shared/utils/feature-utils.ts
```diff
@@ -484,3 +484,29 @@ function createDependentFeaturesLog(
...
+ * https://github.com/angular/angular-cli/issues/16320.
+ * Using `ng add` with options of type array (--features)
+ * may not work correctly in some situations, and might only work on the
+ * second attempt.
+ */
+export function getFeaturesOptions<OPTIONS extends LibraryOptions>(
+  options: OPTIONS
+): string[] {
+  return getArrayOptions(options.features ?? []);
+}
```
Platonn: nitpick: Did you consider having a different signature of this util function?
Now we use it as:
`options.features = getFeaturesOptions(options);`, where `getFeaturesOptions` is a pure function that knows about the shape of `OPTIONS`

IMHO more semantic would be using it as:
a) `options.features = convertToArray(options.features)`, where `convertToArray` is a pure function, which doesn't know about the shape of the `OPTIONS` object
OR
b) `normalizeOptionsFeatures(options)` which under the hood mutates the `options.features` property

Personally, I tend toward the option a)

kpawelczak: Thanks, I went with option B as I think it will work better with JSDoc.

projects/schematics/src/shared/utils/feature-utils.ts
```diff
@@ -484,3 +484,29 @@ function createDependentFeaturesLog(
...
+): string[] {
+  return convertToArray(options.features ?? []);
+}
+
+export function convertToArray(values: string[] | string): string[] {
+  let optionsArray: string[] = [];
+  if (Array.isArray(values)) {
+    optionsArray = values;
+  } else if (values.includes(',')) {
+    optionsArray = values.split(',');
```
Platonn, line 506: Question: shall we somehow trim whitespaces, if using as `--features="Cart, Order, Checkout"` (with spaces)?
1. we should test it manually, if it works now (wihtout trimming)
2. if it doesn't work with spaces, then let's fix it + add unit tests for this case

kpawelczak: My initial thought was that the correct command is --features=Cart,Order,Checkout, as this is the syntax used for an array-type option. The command you provided assumes that the feature is a string. However, the type array | string is just a workaround. I believe we should still treat it as an array.

I haven't tested the whitespace theory yet.

kpawelczak:
After a review and testing, we found that this workaround is not viable as it disrupts the prompts

PR #18931, opened 2024-06-04
POC: performance improvements

Uli-Tiger:
* only re-render prices on GET prices
* only re-render changed groups on GET/PATCH
* only re-render changed attributes on GET/PATCH


Uli-Tiger:
final PR: https://github.com/SAP/spartacus/pull/18997

PR #18934, opened 2024-06-04
feat: Remove required validators from coupon and quick order in cart

rmch91:
Closes: https://jira.tools.sap/browse/CXSPA-997

Changes are hidden under the feature toggles.

In the cart, removed required validators for coupon component's input and quick order component's inputs.
Instead of showing errors under the input when user clicks the button, decided to disable buttons when inputs are empty.
It was confusing that near the proceed to checkout button there are couple of fields that were marked as "Required". But those fields are not required for processing to the checkout page. 
Inputs were marked as required only to indicate that it is required for clicking "Add" or "Apply" buttons.


projects/core/src/features-config/feature-toggles/config/feature-toggles.ts
```diff
@@ -251,6 +251,13 @@ export interface FeatureTogglesInterface {
    * `LoginRegisterComponent`, `ConfigureProductComponent`
    */
   a11yUseButtonsForBtnLinks?: boolean;
+
+  /**
+   * When enabled disable "Apply" button in promo code component in cart for empty input,
+   * disable "Add" button in quick order component when input is empty and remove
+   * required validators for both inputs
+   */
+  a11yDisabledButtonInsteadOfRequiredFields?: boolean;
```
Zeyber: This name could describe which buttons are being disabled to distinguish it a bit more.

feature-libs/cart/quick-order/components/cart-quick-order-form/cart-quick-order-form.component.ts
```diff
@@ -89,11 +89,21 @@ export class CartQuickOrderFormComponent implements OnInit, OnDestroy {
   }
 
   protected buildForm(): void {
+    const shouldHaveRequiredValidator = !this.featureConfig.isEnabled(
```
Zeyber, line 93: I think we should have a TODO comment here with the ticket number for breaking change cleanup later.

feature-libs/cart/base/components/cart-coupon/cart-coupon.component.ts
```diff
@@ -84,8 +89,15 @@ export class CartCouponComponent implements OnInit, OnDestroy {
 
     this.cartVoucherService.resetAddVoucherProcessingState();
 
+    const shouldHaveRequiredValidator = !this.featureConfigService?.isEnabled(
```
Zeyber, line 93: I think we should have a TODO comment here with the ticket number for breaking change cleanup later.

feature-libs/cart/base/components/cart-coupon/cart-coupon.component.html
```diff
@@ -7,34 +7,58 @@
     <form (ngSubmit)="applyVoucher()" [formGroup]="couponForm">
       <label class="cx-cart-coupon-title">
         {{ 'voucher.coupon' | cxTranslate }}
-        <abbr
-          *cxFeature="'a11yRequiredAsterisks'"
-          class="text-decoration-none"
-          title="{{ 'common.required' | cxTranslate }}"
-          >*</abbr
-        >
+        <ng-container *cxFeature="'!a11yDisabledButtonInsteadOfRequiredFields'">
```
Zeyber: I think we should have a TODO comment here with the ticket number for breaking change cleanup later.

feature-libs/cart/base/components/cart-coupon/cart-coupon.component.html
```diff
@@ -7,34 +7,58 @@
...
+            *cxFeature="'a11yRequiredAsterisks'"
+            class="text-decoration-none"
+            title="{{ 'common.required' | cxTranslate }}"
+            >*</abbr
+          >
+        </ng-container>
       </label>
 
       <div class="cx-cart-coupon-container">
-        <input
```
Zeyber, line 19: We aren't really doing any DOM changes here but rather changing a few attribute values and removing the <cx-form-errors> tag. Should we use the `a11yDisabledButtonInsteadOfRequiredFields` flag in those attributes and with `*ngIf` on cx-form-errors instead to make this a bit cleaner? I don't mind if you do this or not since it will be cleaned up next major anyway.

feature-libs/cart/quick-order/components/cart-quick-order-form/cart-quick-order-form.component.html
```diff
@@ -8,57 +8,119 @@
         <span class="cx-cart-quick-order-form-productID">
           <label class="cx-cart-quick-order-form-label">
             {{ 'quickOrderCartForm.productCodeLabel' | cxTranslate }}
-            <!-- TODO: (CXSPA-5953) Remove feature flags next major -->
```
Zeyber, line 11: Same comments from `feature-libs/cart/base/components/cart-coupon/cart-coupon.component.html` apply to this template too.

Zeyber (changes requested):

projects/core/src/features-config/feature-toggles/config/feature-toggles.ts

feature-libs/cart/quick-order/components/cart-quick-order-form/cart-quick-order-form.component.ts
rmch91: Changed

feature-libs/cart/base/components/cart-coupon/cart-coupon.component.ts
rmch91: Changed

feature-libs/cart/base/components/cart-coupon/cart-coupon.component.html
rmch91: Changed

feature-libs/cart/base/components/cart-coupon/cart-coupon.component.html
rmch91: I'm not sure I understand exactly what we should do instead. If your propose is to not wrap it with the feature flag at all, then I have some concerns:

- Seems like changing attributes, like removing `required` is a breaking change. Customer could have applied styles for example by attribute selector.
- `cx-from-errors` in the case when toggle is `true` will do nothing, because the input will not have any validators applied. So it's redundant element in the DOM

Zeyber: Ah ok, so this comment is mainly about having easier to read changes. Since we are changing attribute values, we could instead change the properties instead and include the flags in those. This would save us seeing alternative DOM versions with *cxFeature . So like:
```
<button
   ...
   [required]="isFeatureEnabled ? null : true
   ...
>
```
Its not critical to change and may even have issues I didn't consider. I just think it might be cleaner and would love to know what you think :slightly_smiling_face:

Zeyber (dismissed):

developpeurweb:
@rmch91 can you please also toggle visibility in the feature flag for the cautionary legend when the fields are set as not required. It's no longer needed unless we have required fields. Thanks.
cautionary legend

developpeurweb (dismissed):

developpeurweb:
all good on my end

Zeyber (dismissed):

RadhepS (approved):

PR #18936, opened 2024-06-04
fix: (CXSPA-963) - Removed tabindex from storefront component

Pio-Bar:
Ticket: [CXSPA-963](https://jira.tools.sap/browse/CXSPA-963)

Removing this attribute from the storefront component helps to reduce the infamous screen reader bloat. It is better for a screen reader if the focus returns to the body instead.

After my investigation I think there is no consequence to removing it, as i see no effect on the focus directive mentioned in the comment. Some thorough testing may be needed to confirm.


Zeyber (dismissed):

Zeyber (changes requested):
Please follow [these QA steps](https://github.com/SAP/spartacus/pull/18717#issuecomment-2049250540) changing the BRANCH variable to this one to make sure the behaviour is still correct.

Pio-Bar:
> Please follow [these QA steps](https://github.com/SAP/spartacus/pull/18717#issuecomment-2049250540) changing the BRANCH variable to this one to make sure the behaviour is still correct.

I've tried it and the tabbing order starts from the top of the page for 'selectedHostElement' set body.

Zeyber (approved):

developpeurweb (approved):

PR #18948, opened 2024-06-10
feat: A11Y Tab Component for PDP

Zeyber:
Closes: https://jira.tools.sap/browse/CXSPA-2581


projects/storefrontlib/cms-components/content/tab/panel/tab-panel.component.ts
```diff
@@ -0,0 +1,37 @@
...
+
+import { Component, Input, TemplateRef } from '@angular/core';
+import { Tab, TAB_PANEL_CONTENT_TYPE } from '../tab';
+
+@Component({
+  selector: 'cx-tab-panel',
+  templateUrl: './tab-panel.component.html',
+})
+export class TabPanelComponent {
+  @Input() tab: Tab;
```
Zeyber: Need to comment what these inputs mean.

projects/storefrontlib/cms-components/content/tab/tab.component.ts
```diff
@@ -0,0 +1,195 @@
...
+import { Tab, TabConfig, TAB_MODE } from './tab';
+import { wrapIntoBounds } from './tab.utils';
+
+@Component({
+  selector: 'cx-tab',
+  templateUrl: './tab.component.html',
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class TabComponent implements OnInit {
+  @Input() tabs$: Observable<Tab[]>;
```
Zeyber: Not sure about this input being observable.

projects/storefrontlib/cms-components/content/tab/tab.ts
```diff
@@ -0,0 +1,70 @@
...
+   * Content to display in tab panel when open.
+   */
+  content?: TemplateRef<any> | ContentSlotComponentData | any;
+  id?: string;
+  attributes?: TabAttributes;
+}
+
+export interface TabConfig {
+  /**
+   *
```
Zeyber: Missing some comments on these interfaces.

Zeyber: Also can move this file to models.

projects/storefrontlib/cms-components/content/tab/panel/tab-panel.component.html
```diff
@@ -0,0 +1,25 @@
+<div
+  role="tabpanel"
+  [attr.aria-labelledby]="tab?.id"
+  [attr.id]="'section-' + tabNum"
+  [attr.tabindex]="isOpen ? 0 : -1"
+  [class.active]="isOpen"
+  [style.display]="isOpen ? 'block' : 'none'"
+  [cxFocus]="{ focusOnEscape: true }"
+  [ngSwitch]="getContentType()"
```
Zeyber: Could we dumb this down to only render TemplateRefs? This could prevent trying to catch all content-type cases from within this class for compatibility when it might be better to handle this case-by-case (ie. convert any content into templateRef before passing in TabParagraphComponent, Facets, etc).

projects/storefrontlib/cms-components/content/tab-paragraph-container/tab-paragraph-container.component.ts
```diff
@@ -34,8 +41,15 @@ export class TabParagraphContainerComponent implements AfterViewInit, OnInit {
   @ViewChildren(ComponentWrapperDirective)
```
Zeyber, line 51: Add deprecated annotations on all outdated functions.

projects/storefrontlib/cms-components/content/tab/tab.component.ts
```diff
@@ -0,0 +1,178 @@
...
+  templateUrl: './tab.component.html',
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class TabComponent implements OnInit {
+  @Input() tabs: Tab[];
+  @Input() config: TabConfig | any;
+
+  TAB_MODE = TAB_MODE;
+
+  openTabs$: BehaviorSubject<number[]>;
```
Zeyber, line 33: Need to consider benefits of using these observables to control the tab component. Why? Main reason I can think of is to reduce complexity. So far:
- I tried to use static variables of `isOpen` and attach them to the `Tab[]` object but this is NOT responsive in practice.
- I could try to leverage the `openTabs`  property in `TabConfig`. I may still need an observable in the end for responsiveness though.

projects/storefrontlib/cms-components/product/product-intro/product-intro.component.spec.ts
```diff
@@ -192,18 +189,6 @@ describe('ProductIntroComponent in product', () => {
       );
     });
 
-    it('should not display Show Reviews when reviews are not available', () => {
```
Zeyber, line 195: These unit tests were removed because the tab component does not render the reviews component until it is clicked. The "Show Reviews" button behaviour was based on showing the button if the reviews component existed. It now checks whether there is an average reviews and renders the button based on this.

Note to QA: We should test the "Show Reviews" button on the product details page still works correctly based on changes made to fix [this ticket](https://github.com/SAP/spartacus/issues/14030).

developpeurweb:
@Zeyber 
* **Tabsgroup Mode** is all good for JAWS and VO. The only observation I have is that, I can’t recall the original styling for this, I’ll check with designers.
* **Accordion Mode** has a some of issues:
    * It’s an SAP requirement (ACC-255) to have tooltips on non-text controls, like the collapsed/expanded icons.
    * **JAWS:** the plus (+) and minus (-) icons get vocalized along with the button label, let’s please add null alternative text for both (but keep the tooltip).
    * **VO:** For some reason, VO won’t vocalize if an accordion is expanded or collapsed like JAWS does, good to have if possible.

developpeurweb:
@Zeyber thank you for the changes. Can we refine the solution with the following as well:
1. Remove collapsed/expanded icons for Tabsgroup, leave only for Accordion (CSS or JS, whatever feels best for you).
2. We have some aria attributes interference. Marked in red on the screenshots below are the attributes that don't belong to the component type. Is it possible to remove/introduce their presence according to breaking point?

**Foreign to Tabgroup**
tabgroup foreign attributes


**Foreign to Accordion**
accordion foreign element

developpeurweb (dismissed):

developpeurweb:
Tested on JAWS and VO. Works well for both.

projects/assets/src/translations/en/product.json
```diff
@@ -98,7 +98,7 @@
       "deliveryTab": "Shipping",
       "SparePartsTabComponent": " Spare Parts"
     },
-    "tabPanelContainerRegion": "Tab group with more product details"
+    "tabPanelContainerRegion": "Group with more product details"
```
kimhw0630: This is a breaking change. One suggestion is to create a new key and use that instead.

projects/storefrontlib/cms-components/content/tab/panel/tab-panel.component.ts
```diff
@@ -0,0 +1,23 @@
...
+})
+export class TabPanelComponent {
+  TAB_MODE = TAB_MODE;
+
+  /**
+   * Tab object to display content and set attributes to.
+   */
+  @Input() tab: Tab;
+
+  @Input() mode: TAB_MODE;
```
kimhw0630: Providing a default value to avoid undefined? such as @Input() mode: TAB_MODE = TAB_MODE.TAB

projects/storefrontlib/cms-components/content/tab/panel/tab-panel.component.ts
```diff
@@ -0,0 +1,23 @@
...
+
+import { Component, Input } from '@angular/core';
+import { Tab, TAB_MODE } from '../tab.model';
+
+@Component({
+  selector: 'cx-tab-panel',
+  templateUrl: './tab-panel.component.html',
+})
+export class TabPanelComponent {
+  TAB_MODE = TAB_MODE;
```
kimhw0630, line 15:
```suggestion
  readonly TAB_MODE = TAB_MODE;
```

projects/storefrontlib/cms-components/content/tab/panel/tab-panel.component.html
```diff
@@ -0,0 +1,10 @@
+<div
+  [attr.role]="mode === TAB_MODE.ACCORDIAN ? 'region' : 'tabpanel'"
+  [tabindex]="0"
+  [attr.aria-labelledby]="tab?.id"
+  [attr.id]="'section-' + tab?.id"
```
kimhw0630: id will be set to "section-undefined" if tab.id is undefined. do we care?

projects/storefrontlib/cms-components/content/tab/tab.utils.ts
```diff
@@ -0,0 +1,22 @@
...
+  if (index < min) {
+    return max;
+  }
+
+  if (index > max) {
+    return min;
+  }
+
+  return index;
+}
```
kimhw0630, line 20: i think below will be clearer structure and reduced redundancy..
```suggestion
if (index < min) {
    return max;
  } else if (index > max) {
    return min;
  }
```

projects/storefrontlib/cms-components/content/tab/panel/tab-panel.component.ts
```diff
@@ -0,0 +1,23 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { Component, Input } from '@angular/core';
+import { Tab, TAB_MODE } from '../tab.model';
+
+@Component({
+  selector: 'cx-tab-panel',
+  templateUrl: './tab-panel.component.html',
+})
```
kimhw0630, line 13:
```suggestion
changeDetection: ChangeDetectionStrategy.OnPush // Optimizes performance by reducing unnecessary checks
```

projects/storefrontlib/cms-components/content/tab/tab.component.html
```diff
@@ -0,0 +1,66 @@
...
+        [attr.aria-expanded]="mode === TAB_MODE.TAB ? null : isOpen(i)"
+        [attr.aria-controls]="'section-' + i"
+        [attr.tabindex]="getTabIndex(i, mode)"
+        [attr.id]="tab.id"
+        [attr.title]="
+          mode === TAB_MODE.TAB
+            ? null
+            : (isOpen(i) ? 'Collapse' : 'Expand') +
+              ' ' +
+              (tab.headerKey ? (tab.headerKey | cxTranslate) : tab.header)
```
kimhw0630: Complex logic inside the template, like calculating the title or tab indexes, can be moved to the component class to keep the template clean and more readable. what do you think?
example,

getTitle(i: number, mode: string): string {
  if (mode === TAB_MODE.TAB) {
    return null;
  }
  return (this.isOpen(i) ? 'Collapse' : 'Expand') + ' ' + (this.tabs[i].headerKey ? this.translate(this.tabs[i].headerKey) : this.tabs[i].header);
}

projects/storefrontlib/cms-components/content/tab/panel/tab-panel.component.ts
Zeyber: OnPush change detection causes issues with responsiveness when switching between Tab and Accordian modes. Tests fail as a result.

kimhw0630 (approved):

PR #18956, opened 2024-06-12
Poc/configurators performance improvements reactive

Uli-Tiger:


feature-libs/product-configurator/rulebased/components/attribute/composition/configurator-attribute-composition.directive.ts
```diff
@@ -10,34 +10,49 @@ import {
...
   constructor(
     protected vcr: ViewContainerRef,
     protected configuratorAttributeCompositionConfig: ConfiguratorAttributeCompositionConfig
   ) {}
 
-  ngOnInit(): void {
-    const componentKey = this.context.componentKey;
+  ngOnChanges(): void {
+    // Each time we update the configuration a complete new configuration state is emitted, including new attributes objects, regardless of if an attribute actually changed.
+    // Hence we compare the last rendered attribute with the the current state and only re-render the attribute if there are actual changes. This improves performance significantly.
```
Platonn, line 38:
```suggestion
    // Hence we compare the last rendered attribute with the the current state and only destroy and re-create the attribute component, if there are actual changes to its data. This improves performance significantly.
```

feature-libs/product-configurator/rulebased/components/price-async/configurator-price-async.component.ts
```diff
@@ -0,0 +1,142 @@
...
+  valuePrice?: Configurator.PriceDetails;
+  quantity?: number;
+};
+
+@Component({
+  selector: 'cx-configurator-price-async',
+  templateUrl: './configurator-price-async.component.html',
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class ConfiguratorPriceAsyncComponent {
```
Platonn, line 32: Just curious: 
do we really need to introduce a full new copy of the old component, or we can reuse the old one and introduce there a new behavior that is hidden by a feature toggle?

creating a new copy feels a bit drastic. 
But perhaps I'm missing something important and you have good arguments for not reusing the old one in a modified way.

feature-libs/product-configurator/rulebased/core/state/configurator-state-utils.ts
```diff
@@ -117,7 +117,7 @@ export class ConfiguratorStateUtils {
    *
    * If no element of the `array` satisfied the `predicate`, it returns the original `array`.
    */
-  protected static updateArrayElement<T>(
+  static updateArrayElement<T>(
```
Platonn, line 120: nitpick: is this change really needed in this PR?

feature-libs/product-configurator/rulebased/core/state/reducers/configurator.reducer.ts
```diff
@@ -453,19 +452,10 @@ function takeOverPricingChanges(
...
-      ? ConfiguratorStateUtils.mergeGroupsWithSupplements(
-          state.groups,
-          priceSupplements
-        )
-      : state.groups;
 
   const result = {
     ...state,
     ...content,
-    groups: groups,
```
Platonn, line 468: It's for sure a breaking change for customers that relied on the fact that those prices are attached as leafs of the tree.

I believe we should hide this breaking change behind a feature toggle.
That said, I'd need to think a bit on how to introduce `FeaturesConfigService` here as an injectable service - into the reducer 🤔 
IMHO we shold not apss a service via action's payload. Actions should not care about such things.
Maybe we could pass a service _somehow_ as an ambient context, using _some_ Ngrx API, maybe a meta reducer? I'd need to dig deeper and investigate. Any quick ideas from top of your head?

feature-libs/product-configurator/rulebased/components/attribute/composition/configurator-attribute-composition.directive.ts
```diff
@@ -10,34 +10,49 @@ import {
...
   protected logger = inject(LoggerService);
+  protected attribute: Configurator.Attribute;
 
   constructor(
     protected vcr: ViewContainerRef,
     protected configuratorAttributeCompositionConfig: ConfiguratorAttributeCompositionConfig
   ) {}
 
-  ngOnInit(): void {
-    const componentKey = this.context.componentKey;
```
Platonn, line 35: Removing ngOnInit and moveing the whole logic to ngOnChanges is a breaking change. Customers may have relied on `ngOnInit` being implemented, e.g. they could have called in their customized component `super.ngOnInit()`.
Therefore:
- let's leave ngOnInit, but wrap its body with a feature toggle
- let's leave the added ngOnChanges, but wrap it's body with an opposite feature toggle
- (_perhaps_ you might want to reuse some common parts in both ngOnInit and ngOnChanges, as they are pretty similar, to DRY "dont-repeat-yourself")

Uli-Tiger: just skipped the whole FT part for the POC, but we will do it for the productive solution - but good to have this comment as reminder

feature-libs/product-configurator/rulebased/components/price-async/configurator-price-async.component.ts
Uli-Tiger: reasoning:
 - interface / inputs are different. price gets directly the price data, price-async only the key to retrieve the price supplement
 - how data is obtained is differently (binding vs subscribing)
 - price-asnyc only covers some of the cases of price, hence can also be much simple. no need to support formulas
  
 PS: the final impl. will look simplified. for the poc i copied more we actually needed.

Uli-Tiger:
final PR: https://github.com/SAP/spartacus/pull/18997

PR #18961, opened 2024-06-13
fix: (CXSPA-1094) - Store finder alerts

Pio-Bar:
Ticket: [CXSPA-1094](https://jira.tools.sap/browse/CXSPA-1094)

I had a hard time getting the 'aria-live' to cooperate here. I have used the 'role="alert"' instead. This makes it easier to control what has to be read out. @developpeurweb Please let me know if this approach provides appropriate screen reader UX.


Zeyber (dismissed):
All good for me. I think we don't need to worry about making the `TODO:` comments and cleanup tickets for the future ;)

developpeurweb (dismissed):
LGTM, narration is present in both JAWS and VO.

Pio-Bar:
Since we no longer rely on aria-live in this component we can get rid of 'a11yStoreFinderSearchboxBloat' feature flag. The problem it was fixing no longer exists for the new alerts.

Zeyber (dismissed):

Zeyber (approved):

PR #18962, opened 2024-06-13
fix: Redundant reacreation of cart items form

sdrozdsap:
https://jira.tools.sap/browse/CXSPA-7457


Zeyber (changes requested):
Looking at this fix and thinking about our discussion, I think it is perfectly acceptable. The implementation as a while of the cart list item is already questionable but I don't think its worth doing any sort of refactor or "code improvement" since it works and (I think) we aren't getting any complains about it. Thanks for figuring it out :)

Lets just add a feature toggle to this to make sure we are covered and I can approve.

Zeyber (approved):

developpeurweb (approved):

PR #18966, opened 2024-06-14
fix: Increase quick order message default delete timeout

Zeyber:
Closes: https://jira.tools.sap/browse/CXSPA-1111


giancorderoortiz (approved):
ok

developpeurweb (approved):

PR #18980, opened 2024-06-19
fix: (CXSPA-7616) - set color for toolbar buttons

Reid-McNaughton:
closes: [CXSPA-7616](https://jira.tools.sap/browse/CXSPA-7616)

The foreground colors used in the buttons in the floating toolbar are aligned with the colours used for the buttons in this component:
image

**Behaviour with a11yImproveContrast enabled (without the changes in this PR):**
image

**Expected behaviour:**
image

We either need to apply the changes in this PR or alternatively we could keep this unexpected color change (which does not look bad to me) and make other changes to align with the colors used on secondary buttons (adding a colour change on hover for instance).


sam-garland:
This restores the old behaviour.

sam-garland (approved):

sam-garland:
@RadhepS - Would you be able to review this change?

It preserves past behaviour and maximises contrast in common scenarios.
I don't actually mind the unintended colour change introduced by a11yImproveContrast, but we either need to preserve the old behaviour (which is straightforward) or make other some other changes to achieve consistent behaviour.

RadhepS (approved):

RadhepS:
@sam-garland Sorry for the delay! The PR has been approved. Next time you can send me a direct message on slack or an email! I often get too many emails for github so they get ignored.

PR #18984, opened 2024-06-19
fix: (CXSPA-999) - Checkout progress aria attributes

Pio-Bar:
Ticket: [CXSPA-999](https://jira.tools.sap/browse/CXSPA-999)

I've tested with both JAWS and VO and adding those attributes should provide necessary information.


Zeyber (dismissed):

developpeurweb:
This works well on JAWS (although it doesn't convey a step has been completed) but for some reason, especially **between steps on VO**, there's no way to tell if the user has completes a step already, or which step is selected. To make status for each state very clear to screen reader users, let's try to include the status as part of the `aria-label`, as it was done recently for the Results Page pagination. 

E.g. 
- `aria-label="Shipping Address, completed"`
- `aria-label="Delivery Mode, selected"`
- `aria-label="Payment, disabled"`

confusing steps cxspa-999

Zeyber (approved):

developpeurweb (approved):
Works well for both JAWS and VO.

PR #18987, opened 2024-06-20
Fix : Make my account v2 test more simpler to run

scarai-sap:
Set the value instead of adding. "--env CX_MY_ACCOUNT_V2=true" into the test


LindaSAP (approved):
OK

PR #18989, opened 2024-06-20
CXSPA-7418: remove asterisk and instruction text that not necessary in Login and Reset Password page

i53577:
jira: https://jira.tools.sap/browse/CXSPA-7418


scarai-sap (approved):

niehuayang (approved):
LGTM

PR #18999, opened 2024-06-25
feat: CXSPA-7197 Create toggle for strict NgRx and HTTP error handling

pawelfras:
This |PR contains an implementation of a toggle for strict HTTP and NgRx error handling in SSR.

QA steps (examples):
- strict HTTP error handling in storefrontapp
  - go to `default-occ-product-config.ts` and modify the default `productSearch` endpoint
  - run `npm run dev:ssr`
  - go to http://localhost:4200/electronics-spa/en/USD/Brands/all/c/brands and verify if the error log is visible in the server console and not present in the browser's console
   image
  - kill the SSR app and run `npm run start`
  - go to http://localhost:4200/electronics-spa/en/USD/Brands/all/c/brands and verify if the error log from `logger.service.ts` is not present in the browser's console
- strict HTTP error handling in fresh SPA app
  - create new app `npx @angular/cli@17 new my-app --standalone=false --style=scss --routing=false`
  `npx @angular/cli@17 add @spartacus/schematics@latest --baseUrl="https://40.76.109.9:9002"`
  - adjust spartacus config in `spartacus-configuration.module.ts`
  ```ts
  providers: [
    provideConfig(<OccConfig>{
      backend: {
        occ: {
          baseUrl: 'https://40.76.109.9:9002',
          endpoints: {
            productSearch: 'products/search2',
          },
        },
      },
    }),
    / * * /
    provideFeatureToggles({
      strictHttpAndNgrxErrorHandling: true,
    }),
  ]
  ```
  - adjust `server.ts` file:
  ```
  export function app(): express.Express {
      const server = express();
      const serverDistFolder = dirname(fileURLToPath(import.meta.url));
      const browserDistFolder = resolve(serverDistFolder, '../browser');
      const indexHtml = join(browserDistFolder, 'index.html');
    
      / * * /
      
      return server;
  }
  ```
  - run `npm run build && NODE_TLS_REJECT_UNAUTHORIZED=0 run serve:ssr:my-app`
  - go to http://localhost:4200/electronics-spa/en/USD/Brands/all/c/brands and verify if the error log is visible in the server console and not present in the browser's console
  image
  
- strict NgRx error handling in storefrontapp
  - go to `product-search.effect.ts` and modify `searchProducts$` effect:
  ```
  map(() => {
    // return new ProductActions.SearchProductsSuccess(
        //   data,
        //   action.auxiliary
    // );
    throw new Error('Test Error');
  }),
  ```
  - run `npm run dev:ssr`
  - go to http://localhost:4200/electronics-spa/en/USD/Brands/all/c/brands and verify if the error log is visible in the server console and not present in the browser's console
  image

  - kill the SSR app and run `npm run start`
  - go to http://localhost:4200/electronics-spa/en/USD/Brands/all/c/brands and verify if the error log with `Test Error` message from `logger.service.ts` is not present in the browser's console
- strict NgRx error handling in fresh SPA app
  - build and deploy Spartacus packages locally keeping adjustments in `product-search.effect.ts`:
  - create new app `npx @angular/cli@17 new my-app --standalone=false --style=scss --routing=false`
  `npx @angular/cli@17 add @spartacus/schematics@latest --baseUrl="https://40.76.109.9:9002"`
  - adjust spartacus config in `spartacus-configuration.module.ts`
  ```ts
  providers: [
    / * * /
    provideFeatureToggles({
      strictHttpAndNgrxErrorHandling: true,
    }),
  ]
  ```
  - adjust `server.ts` file:
  ```
  export function app(): express.Express {
      const server = express();
      const serverDistFolder = dirname(fileURLToPath(import.meta.url));
      const browserDistFolder = resolve(serverDistFolder, '../browser');
      const indexHtml = join(browserDistFolder, 'index.html');
    
      / *** /
      
      return server;
  }
  ```
  - run `npm run build && NODE_TLS_REJECT_UNAUTHORIZED=0 run serve:ssr:my-app`
  - go to http://localhost:4200/electronics-spa/en/USD/Brands/all/c/brands and verify if the error log is visible in the server console and not present in the browser's console
  image

closes [CXSPA-7197](https://jira.tools.sap/browse/CXSPA-7197)

### DIFF
--- a/projects/core/src/error-handling/effects-error-handler/cx-error-handler.effect.ts
+++ b/projects/core/src/error-handling/effects-error-handler/cx-error-handler.effect.ts
@@ -10,28 +10,25 @@ import { Observable } from 'rxjs';
 import { filter, tap } from 'rxjs/operators';
 import { FeatureConfigService } from '../../features-config';
 import { ErrorAction } from '../../model/index';
-import { WindowRef } from '../../window';
 import { EffectsErrorHandlerService } from './effects-error-handler.service';
 
 @Injectable()
 export class CxErrorHandlerEffect {
   protected actions$ = inject(Actions);
-  protected effectErrorHandler = inject(EffectsErrorHandlerService);
-  protected featureConfigService = inject(FeatureConfigService);
-  protected windowRef = inject(WindowRef);
+  protected effectErrorHandlerService = inject(EffectsErrorHandlerService);
+  private featureConfigService = inject(FeatureConfigService);
 
   error$: Observable<ErrorAction> = createEffect(
     () =>
       this.actions$.pipe(
-        filter(this.effectErrorHandler.filterActions),
+        filter(this.effectErrorHandlerService.filterActions),
         tap((errorAction: ErrorAction) => {
           if (
             this.featureConfigService.isEnabled(
-              'strictHttpAndNgrxErrorHandling'
-            ) &&
-            !this.windowRef.isBrowser() // handle only in SSR
+              'ssrStrictErrorHandlingForHttpAndNgrx'
+            )
           ) {
-            this.effectErrorHandler.handleError(errorAction);
+            this.effectErrorHandlerService.handleError(errorAction);
           }
         })
       ),
--- a/projects/core/src/error-handling/effects-error-handler/cx-error-handler.effect.ts
+++ b/projects/core/src/error-handling/effects-error-handler/cx-error-handler.effect.ts
@@ -4,28 +4,37 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { Injectable } from '@angular/core';
+import { Injectable, inject } from '@angular/core';
 import { Actions, createEffect } from '@ngrx/effects';
 import { Observable } from 'rxjs';
 import { filter, tap } from 'rxjs/operators';
+import { FeatureConfigService } from '../../features-config';
 import { ErrorAction } from '../../model/index';
+import { WindowRef } from '../../window';
 import { EffectsErrorHandlerService } from './effects-error-handler.service';
 
 @Injectable()
 export class CxErrorHandlerEffect {
+  protected actions$ = inject(Actions);
+  protected effectErrorHandler = inject(EffectsErrorHandlerService);
+  protected featureConfigService = inject(FeatureConfigService);
+  protected windowRef = inject(WindowRef);
+
   error$: Observable<ErrorAction> = createEffect(
     () =>
       this.actions$.pipe(
         filter(this.effectErrorHandler.filterActions),
-        tap((errorAction: ErrorAction) =>
-          this.effectErrorHandler.handleError(errorAction)
-        )
+        tap((errorAction: ErrorAction) => {
+          if (
+            this.featureConfigService.isEnabled(
+              'strictHttpAndNgrxErrorHandling'
+            ) &&
+            !this.windowRef.isBrowser() // handle only in SSR
+          ) {
+            this.effectErrorHandler.handleError(errorAction);
+          }
+        })
       ),
     { dispatch: false }
   );
-
-  constructor(
-    protected actions$: Actions,
-    protected effectErrorHandler: EffectsErrorHandlerService
-  ) {}
 }
--- a/projects/core/src/error-handling/http-error-handler/http-error-handler.interceptor.ts
+++ b/projects/core/src/error-handling/http-error-handler/http-error-handler.interceptor.ts
@@ -11,9 +11,11 @@ import {
   HttpInterceptor,
   HttpRequest,
 } from '@angular/common/http';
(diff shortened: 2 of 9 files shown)


projects/core/src/error-handling/effects-error-handler/cx-error-handler.effect.ts
```diff
@@ -4,28 +4,37 @@
...
         filter(this.effectErrorHandler.filterActions),
-        tap((errorAction: ErrorAction) =>
-          this.effectErrorHandler.handleError(errorAction)
-        )
+        tap((errorAction: ErrorAction) => {
+          if (
+            this.featureConfigService.isEnabled(
+              'strictHttpAndNgrxErrorHandling'
+            ) &&
+            !this.windowRef.isBrowser() // handle only in SSR
```
Platonn: [MEDIUM]

1. For better extensibility, let's move this isBrowser condition to the `EffectsErrorHandlerService` 
Why: this effect is private API. There might be edge cases where customers would like to log such actions in browser as well (not necessarily to console, but to their custom CSR telemetry).
So let's make it extendable/customizable and move to `EffectsErrorHandlerService which is in public API

2. It's not obvious why we exclude Browser from this logic, especially for a newcomers to this code. Let's help them. Let's leave here a short notice on _why_ we don't propagate outgoing http request's errors in the browser:
   ```
   // We avoid sending unpredictable errors to the browser's console, to prevent
   // possibly exposing there potentially confidential user's data.
   // This isn't an issue in SSR, where pages are rendered anonymously.
   // Moreover, in SSR we want to capture all app's errors, so we can potentially send
   // a HTTP error response (e.g. 500 error page) from SSR to a client.
   ```
   
    Note: analogical comment applies to `isBrowser` in the http error handling interceptor

projects/core/src/error-handling/http-error-handler/http-error-handler.interceptor.ts
```diff
@@ -24,7 +26,9 @@ import { tap } from 'rxjs/operators';
  */
 @Injectable()
 export class HttpErrorHandlerInterceptor implements HttpInterceptor {
-  constructor(protected errorHandler: ErrorHandler) {}
+  errorHandler = inject(ErrorHandler);
+  featureService = inject(FeatureConfigService);
+  windowRef = inject(WindowRef);
```
Platonn: 1. let's make them protected
2. FeatureConfigService -> make private, as it's only temporary dependency to be removed in the future

projects/core/src/features-config/feature-toggles/config/feature-toggles.ts
```diff
@@ -43,6 +43,11 @@ export interface FeatureTogglesInterface {
    */
   productConfiguratorAttributeTypesV2?: boolean;
 
+  /**
+   * Enable strict error handling in SSR for errors occurred in HTTP calls and in NgRx failure actions.
+   */
```
```suggestion
  /**
   * In SSR, the following errors will be printed to logs (and additionally can also
   * be forwarded to ExpressJS if only the setting 
   * `SsrOptimizationOptions.ssrErrorHandling`  is enabled):
   * 1. any outgoing HTTP request error (4xx-5xx status)
   * 2. any NgRx action with the `error` property
   */
```

projects/core/src/features-config/feature-toggles/config/feature-toggles.ts
```diff
@@ -43,6 +43,11 @@ export interface FeatureTogglesInterface {
...
+  strictHttpAndNgrxErrorHandling?: boolean;
```
Platonn: 1. it's related to only SSR, as we decided only recently. So maybe the flag name should describe it?

`strictHttpAndNgrxErrorHandling` vs

`strictHttpAndNgrxErrorHandlingInSsr` vs

`ssrStrictHttpAndNgrxErrorHandling`

?

2. btw. Do you think it makes sense now to rename to simply `ssrErrorHandling` (same as in SsrOptimizationOptions)?  Sorry for going circles, I know we've been there... but we abandoned this idea due to a few reasons. I only remember (one of?) them being "it's a change for both SSR and CSR" - which is no longer a valid argument.

pawelfras: I had the same considerations yesterday and I've started wondering what if we eventually enable this feature for CSR?

I see two possible scenarios:
* introduce strict error handling for CSR when the toggle for SSR error handling is still valid/active (12 moths rule)
* introduce strict error handling for CSR when SSR have it by default (after 12 months)

I think both scenarios lead to introducing separate toggle `csrErrorHandling`, `strictNgrxAndHttpErrorHandlingInCsr`, you name it, so I would say we can go with renaming this existing toggle to be more related to SSR. I'm still not sure whether `ssrErrorHandling` is ok as in this case, it might be too generic and not describe what exactly we're toggling. But maybe good JSDoc (proposed by you in the comment above) is good enough 🤔

Platonn (approved):

PR #19001, opened 2024-06-25
fix: (CXSPA-1188) - Message component focus management.

Pio-Bar:
Ticket: [CXSPA-1188](https://jira.tools.sap/browse/CXSPA-1188)

If the Message contains an accordion/action button it will get focused after it renders. 
VO navigation in read mode should no longer read the message title twice.


projects/storefrontlib/cms-components/misc/message/message.component.ts
```diff
@@ -67,4 +80,15 @@ export class MessageComponent {
         return ICON_TYPE.SUCCESS;
     }
   }
+
+  ngAfterViewInit(): void {
+    // TODO: (CXSPA-7725) - Remove feature flag next major release.
```
Zeyber: We no longer need to make these TODO comments ;)

projects/storefrontlib/cms-components/misc/message/message.component.ts
```diff
@@ -67,4 +80,15 @@ export class MessageComponent {
...
+    if (
+      (this.featureConfigService.isEnabled('a11yCxMessageFocus') &&
```
Zeyber: It looks like the brackets might be in the wrong place here. I think this would fire regardless of the feature flag setting.

Zeyber: On second look, `#messageContainer` will only exist when the feature flag is active anyway. Can we just use this instead?

```
if(this.accordianText || this.actionButtonText) {
  this.messageContainer?.nativeElement?.focus();
}
```

Note: I'm not sure if the `?.` is needed in both places here.

I think we could also remove `private featureConfigService = inject(FeatureConfigService);` above if this is the case.

Zeyber (changes requested):

projects/storefrontlib/cms-components/misc/message/message.component.ts
Pio-Bar: Comments removed

projects/storefrontlib/cms-components/misc/message/message.component.ts
Pio-Bar: You are right, the template feature flag is enough.

Zeyber (approved):

developpeurweb (approved):

PR #19005, opened 2024-06-26
fix: wrong usage of attribute required

steinsebastian:


ChristophHi (approved):
Should be merged only after CF for 22.11.25

PR #19010, opened 2024-06-27
feat: CXCDS-12079 Improve searchBox handling for Trending Searches in SPA

darvsorceix:


pekura (changes requested):
Somehow the text selection in Recent Searches got broken in this PR:

Bildschirmfoto 2024-07-04 um 08 20 56

(all characters are bold in the Recent Searches list - the 's' should regular because it matches the text in the search input field)



Here the same situation without the PR being applied - this is how it should be:

Bildschirmfoto 2024-07-04 um 08 19 48

pekura (approved):

PR #19012, opened 2024-06-27
CXCDS-12024 fix translation schematics

lasoh:


kpawelczak (approved):

kpawelczak (changes requested):
Currently, `assets` are being exported from `@spartacus/cds` which deviates from our standard practice. We need to phase out this export method and establish a new export entry specifically for `@spartacus/cds/assets`

 _This assumes that the new `@spartacus/cds/assets` sub lib is established_

1. You need to create deprecated entry point from `@spartacus/cds`
```
import {
  cdsTranslations as originaCdsTranslations,
  cdsTranslationChunksConfig as originalCdsTranslationChunksConfig } from '@spartacus/cds/assets';

/** @deprecated - reasons + what to use instead */
export const cdsTranslations: TranslationResources = originalCdsTranslations;

/** @deprecated - reasons + what to use instead */
export const cdsTranslationChunksConfig: TranslationChunksConfig = originalCdsTranslationChunksConfig;
```

2. While `@spartacus/cds/assets` holds the definition
```
// (not deprecated)
export const cdsTranslations: TranslationResources = {
  en,
};
// (not deprecated)
export const cdsTranslationChunksConfig: TranslationChunksConfig = {
  cdsRecentSearches: ['cdsRecentSearches'],
};
```

integration-libs/cds/src/assets/index.ts
```diff
@@ -0,0 +1,18 @@
+/*
+ * SPDX-FileCopyrightText: 2024 SAP Spartacus team <spartacus-team@sap.com>
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { TranslationChunksConfig, TranslationResources } from '@spartacus/core';
+import {
+  cdsTranslations as originaCdsTranslations,
```
kpawelczak:
```suggestion
  cdsTranslations as originalCdsTranslations,
```

integration-libs/cds/src/assets/index.ts
```diff
@@ -0,0 +1,18 @@
...
+  cdsTranslationChunksConfig as originalCdsTranslationChunksConfig,
+} from '@spartacus/cds/assets';
+
+/** @deprecated  Use @spartacus/cds/assets instead */
+export const cdsTranslations: TranslationResources = originaCdsTranslations;
```
kpawelczak:
```suggestion
export const cdsTranslations: TranslationResources = originalCdsTranslations;
```

kpawelczak (approved):

PR #18997, opened 2024-06-24
perf: Improve rendering performance for large configuration models [CXSPA-7477]

Uli-Tiger:
**Issue:** The product configuration UI is completely re-rendered after each UI interaction. This may lead to performance issues for large configuration models, where a lot of attributes (>50) and/or a lot of possible values per attribute (>50) are rendered on the UI. 

**Solution**: When this feature toggle(`productConfigurationDeltaRendering`) is activated, only these parts of the UI are re-rendered, that actually changed, significantly (up to factor 10) improving rendering performance for large models. Please note, this will influence how the pricing requests are processed and rendered. Instead of merging the prices into the configuration model, which effectively triggers re-rendering the whole UI-Component tree, the price supplements are kept in a separate subtree of the model, so that UI components can react independently on pricing chnages
**Hence, it is advised to do full regression testing after activation of this flag and before rolling this out to production.**

**Details**: This optimization consists of 3 main parts:
(1) A smart change detection for attributes implemented in `configurator-attribute-composition.directive.ts`. Basically, the directive will compare last rendered attribute state with the current attribute state and only trigger re-rendering of the attribute if the content of the attribute changed.  Additionally, the `ngFor` generating the attribute list is enhanced with a `trackBy` function based on the attribute key.
(2) The `ngFor` generating the group menu  is enhanced with a `trackBy` function based on the group id.
(3) Decouple pricing updates from configuration changes. Instead of merging pricing changes, which are fetched asynchronously,  into the leaves (values) of the configuration tree within the `configuration.reducer`, the pricing changes are kept in a separate subtree `configuration.priceSupplements`. Unfortunately any attribute component and not only the pricing component should react on pricing changes. The attribute components build the ARIA labels for the attributes including any price related information (surcharges). To help attribute components to track pricing changes a new component service `ConfiguratorDeltaRenderingService` has been introduced. It monitors the configuration observable for pricing supplements fitting the given attribute key. Its `rerender` method will emit whenever a pricing change was detected. So the prices can now be updated independently from the rest of the configuration model. This is also a prerequisite for the optimization (1), as otherwise  fetching the prices would always trigger re-rendering of all attributes.

PR #19015, opened 2024-06-28
fix: Different aria-labels for un-/selected values for radio-buttons

steinsebastian:


Larisa-Staroverova (dismissed):

feature-libs/product-configurator/rulebased/components/attribute/types/base/configurator-attribute-single-selection-base.component.ts
```diff
@@ -247,34 +247,34 @@ export abstract class ConfiguratorAttributeSingleSelectionBaseComponent extends
...
-              value: value.valueDisplay,
-              attribute: attribute.label,
-              price: value.valuePrice.formattedValue,
-            }
-          )
+          .translate(translationKey, {
+            value: value.valueDisplay,
+            attribute: attribute.label,
+            price: value.valuePrice.formattedValue,
+          })
```
Uli-Tiger: @steinsebastian can you do me a favor to refactor the method,

- so there is only one call to the translation service. 
- that multi level if can be replacded by a simple if - else if - else

psydo code
```
if(totalprice != 0) {
key = ...
params = ...
} else if (valueprice != 0){
key = ...
params = ...
} else{
key = ...
params = ...
}
transService.translate.....
return
```

i believe this makes it easier to understand what the method does and that there are exactly 3 paths.

Uli-Tiger (changes requested):

feature-libs/product-configurator/rulebased/components/attribute/types/base/configurator-attribute-single-selection-base.component.ts
Uli-Tiger: see also: https://github.com/SAP/spartacus/pull/18997/files#r1662749706

Uli-Tiger (approved):

PR #19016, opened 2024-07-01
feat : theme switcher and high contrast themes (CXSPA-7604)

kimhw0630:
1. Previous behavior
   Previously, the site theme could be configured via the Spartacus config `config.context.theme` - which could be defined statically in the code or populated automatically from CMS via the feature of "automatic site context configuration",

2. New mechanism of Optional Themes
   Now we introduce a way to define a list of additional, optional themes via a new config `siteTheme.optionalThemes`. The old config `config.context.theme` is still important (can be still defined statically or via CMS) and it still defines the id of the Default theme. If the old config `config.context.theme` is undefined, then the Default theme id is an empty string `''`.

3. Added Theme Switcher UI component 
   This PR introduces a feature of a Theme Switcher - UI dropdown for choosing the preferred theme (and saving it to a local storage to be persisted across page refreshes). 

4. Introduced 2 optional themes for High Contrast
   We introduced now 2 optional themes for a11y reasons: High Contrast Light and High Contrast Dark. Those themes alter the global CSS Custom Properties (aka "CSS variables") like `--cx-color-background` etc. - so all the components that depended on those variables "automagically" change their appearance, when the alternative theme is activated.
For components that didn't depend on those variables, we needed to add extra styles rules there that enforce using the global "CSS variables" just when our high contrast themes are enabled. Those rules say ~"if there is a CSS theme class on an ancestor element like `.cx-theme-high-contrast-dark`, then plesae depend on this CSS global variable", e.g.:
   ```scss
       @include cx-highContrastTheme {
         background-color: var(--cx-color-background);
       }
   ```
   where the util mixin is defined as:
   ```scss
    @mixin cx-highContrastTheme-dark() {
     .cx-theme-high-contrast-dark & {
       @content;
     }
   }
   ```

   NOTE: if customers want to benefit from our OOTB High Contrast themes and the Theme Switcher in their custom components, their styles must depend on our global "CSS variables".

5. How to enable the feature
   To enable the feature of the Theme Switcher and its OOTB Optional Themes for high contrast, you need to do 2 things: 
   - enable the feature toggle `useSiteThemeService` 
   - add the CMS component to your sample data: `SiteThemeSwitcherComponent`.

Closes: https://jira.tools.sap/browse/CXSPA-7604

---

To see the theme switcher:

1.  Go to the backoffice (e.g. https://40.76.109.9:9002/backoffice/)
2.  Navigate to WCMS -> Component.: 
- Add a "CMS Flex Component": 
- enter "Catalog Version" (e.g. "SPA Electronics Content Catalog: Online" )
- enter ID "SiteThemeSwitcherComponent"
After component is created
- Go to Administration tab: enter value for "Flex type" -> "SiteThemeSwitcherComponent"
- Go to Content Slots tab: select "Content Slots"-> Site Context Slot (Id=SiteContextSot, Catalog version = SPA Electronics Content Catalog: Online)


RadhepS (dismissed):

developpeurweb (dismissed):

projects/storefrontlib/cms-components/misc/theme-switcher/theme-switcher.component.ts
```diff
@@ -0,0 +1,119 @@
...
+/**
+ * Component for switching themes.
+ */
+@Component({
+  selector: 'cx-theme-switcher',
+  templateUrl: './theme-switcher.component.html',
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class ThemeSwitcherComponent implements OnInit {
+  private static readonly THEME_ID = 'theme';
```
Platonn: 1. for better extensibility let's make it `protected`
2. the key that is used in the localStorage global namespace -  `theme` - is quite general. Did you consider making it more specific? e.g. `cx-ui-theme`  or `cx-styling-theme`?

projects/storefrontlib/cms-components/misc/theme-switcher/theme-switcher.component.ts
```diff
@@ -0,0 +1,119 @@
...
+
+/**
+ * Component for switching themes.
+ */
+@Component({
+  selector: 'cx-theme-switcher',
+  templateUrl: './theme-switcher.component.html',
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class ThemeSwitcherComponent implements OnInit {
```
Platonn: IMHO this component has too many responsibilities. Here's my recommended split of respobsibilities:
- `ThemeSwitcherComponent` should be just an UI selector. It should not care about local storage
- in a different layer (e.g. a service) we should care about how to persist the `theme` when choosing with a selector. e.g. `ThemePersistenceService`.
- in a different layer (e.g. induced by the `APP_BOOTSTRAP_LISTENER` hook) we should handle setting the theme on page start, based on the localStorage, e.g. `ThemeService.init()`

To justify it more, let's consider the following examples, if we had every responsibility held in `ThemeSwitcherComponent`:
- what if ThemeSwitcherComponent cms component is missing on some route and you do a "full page refresh" on this route? Then the theme won't be set based on localStorage when page starts.
- what if a customer decides to put 2 ThemeSwitcherComponents on the page (e.g. in header and footer). Then both of them would attmpt to set the theme based on the localStorage on page start

projects/storefrontlib/cms-components/misc/theme-switcher/theme-switcher.component.html
```diff
@@ -0,0 +1,18 @@
+<label *ngIf="showSwitcher">
+  <span> {{ 'themeSwitcher.theme' | cxTranslate }}:</span>
+  <select (change)="onSelect($any($event).target.value)">
```
Platonn: [nitpick]
`$any()` - is it _really_ needed?

projects/storefrontstyles/scss/theme/high-contrast/high-contrast.scss
Platonn, line 1: Those theming styles mention plenty of components selectors in the system.
Do we really have to mention them all explicitly in this file? Can't we write more generic styles rules?
Why asking:
- it's very probable we might have missed some UI components from the integration libs (e.g. EPD, CDC,...)
- when we soon add new components in the system, the current styles will likely be missing mentions of those new components. How do we plan to handle this?
- for sure it's missing all customers' custom components (that we don't know)

Maybe alternatively, it should be the responsibility of those components' individual styles to be aware of those theming modes and contain special rules for when being within the specific theme mode? I'm mentioning it only for the a11y purposes.
I don't think any custom theming rules should be a part of components default styles. I'm considering here only by treating a11y as a OOTB capabilities of those components (and part of possible future DOD - to be compatible with dark/light-contrast modes).

Btw. Do we plan to add to DOD checklist that components should be compatible with light/dark high contrast themes?

projects/storefrontlib/cms-components/misc/theme-switcher/config/default-theme-switcher-config.ts
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { ThemeSwitcherConfig } from './theme-switcher-config';
+
+export const defaultThemeSwitcherConfig: ThemeSwitcherConfig = {
+  themeSwitcher: {
+    themes: [
+      { i18nNameKey: 'themeSwitcher.themes.default', className: 'santorini' },
```
Platonn: Why did we hardcoded in our default config the default theme `santorini`?

Will it be a breaking change for existing customers that don't use the `santorini` theme, but their own-named theme, when they just bump Spartacus version to newer?

projects/storefrontlib/cms-components/misc/theme-switcher/config/default-theme-switcher-config.ts
```diff
@@ -0,0 +1,23 @@
...
+      {
+        i18nNameKey: 'themeSwitcher.themes.highContrastDark',
+        className: 'high-contrast-dark',
+      },
+      {
+        i18nNameKey: 'themeSwitcher.themes.highContrastLight',
+        className: 'high-contrast-light',
+      },
```
Platonn: [QUESTION]

Should the "dark contrast" and "light contrast" be really **altearnatives** to the existing themes (santorini or whatever CMS-driven value set in backoffice)?
Or maybe those "dark contrast" and "light contrast" should be only a modifiers of existing themes? aka. "theme mode"
Just asking, what makes more sense to you?

projects/storefrontstyles/scss/theme/high-contrast/high-contrast.scss
```diff
@@ -0,0 +1,258 @@
...
+      background-color: var(--cx-color-background);
+      color: var(--cx-color-text);
+    }
+    cx-icon {
+      color: var(--cx-color-text);
+    }
+  }
+
+  cx-page-layout {
+    background-color: var(--cx-color-background) !important;
```
Platonn: Do we really have to put so many `!important`?
I guess this way we'll make customers' life much harder to extend those styles. 

What other alternative solutions do you see and what are their pros&cons?

projects/storefrontstyles/scss/theme/high-contrast/dark/_variables.scss
```diff
@@ -0,0 +1,14 @@
+.high-contrast-dark {
```
Platonn: Names of those classes in the global namespace are IMHO too general. Can we make them more specific, e.g. 
`cx-high-contrast-dark`
`cx-theme-high-contrast-dark`
?

projects/storefrontstyles/scss/theme/high-contrast/high-contrast.scss
developpeurweb: I agree, let's use specificity as per Styling Guidelines: https://wiki.one.int.sap/wiki/display/spar/UI+Styling+and+Accessibility+Onboarding

projects/storefrontstyles/scss/theme/high-contrast/high-contrast.scss
developpeurweb: @Platonn for your last question I would say YES, at least for what we ship OOTB. This is a [mandatory SAP requirement](https://wiki.one.int.sap/wiki/display/Acc/ACC-262+Details) that goes against the product's Accessibility Score.

projects/storefrontstyles/scss/theme/high-contrast/dark/_variables.scss
developpeurweb: I agree, let's use the `cx-theme-` prefix

projects/storefrontlib/cms-components/misc/theme-switcher/config/default-theme-switcher-config.ts
kimhw0630: Currently, we should focus on modifying existing themes, as we cannot fully address the customer's changes and their components.

projects/storefrontstyles/scss/theme/high-contrast/high-contrast.scss
kimhw0630: removed all important.

projects/storefrontstyles/scss/theme/high-contrast/dark/_variables.scss
kimhw0630: done

projects/storefrontlib/cms-components/misc/theme-switcher/theme-switcher.component.ts
kimhw0630: done

projects/core/src/site-theme/store/site-theme-store.module.ts
```diff
@@ -0,0 +1,42 @@
...
+export function siteThemeStoreConfigFactory(): StateConfig {
+  const config: StateConfig = {
+    state: {
+      ssrTransfer: {
+        keys: { [SITE_THEME_FEATURE]: StateTransferType.TRANSFER_STATE },
+      },
+    },
+  };
+  return config;
+}
```
Platonn, line 29: Did you test it with SSR enabled? 

I'm worrying that perhaps it might break taking the Theme from localStorage on page refresh. I guess the `default` theme calculated during the anonymous SSR rendering and transferred to CSR directly to the ngrx state will take precedence when page starts, and therefore a value from localStorage might be dismissed.

All that said, those just my worries from top of my head. But to be 100% sure, I'd like us to test this scenario it with SSR enabled and have a proof.

projects/storefrontstyles/scss/theme/high-contrast/high-contrast.scss
Platonn: So, do we plan to delete this file and instead, place the relevant rules in their specific components' styles?

e.g. 
```scss
// _searchbox.scss

/*...*/

%cx-searchbox {
  .cx-theme-high-contrast-dark &,
  .cx-theme-high-contrast-light & {
    label {
      background-color: var(--cx-color-background);
    }
    
    /*...*/
```

projects/core/src/site-theme/providers/index.ts
```diff
@@ -0,0 +1,7 @@
+/*
+ * SPDX-FileCopyrightText: 2024 SAP Spartacus team <spartacus-team@sap.com>
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+export * from './theme_storage_Id';
```
Platonn: In this PR sometimes we use term `theme` and sometimes `site theme`.

For consistency, do you think we should rename everywhere in this pr the word `theme` to `site theme`? (preserving the original casing, e.g.  `theme` / `Theme` --> `siteTheme` OR  `site-theme` / `SiteTheme`

projects/core/src/site-theme/facade/site-theme.service.ts
```diff
@@ -0,0 +1,88 @@
...
+
+import { Theme } from '../../model/misc.model';
+import { SiteThemeConfig } from '../config/site-theme-config';
+import { isNotNullable } from '../../util/type-guards';
+import { StateWithSiteTheme } from '../store/state';
+import { SiteThemeSelectors } from '../store/selectors';
+import { SiteThemeActions } from '../store/actions';
+
+@Injectable()
+export class SiteThemeService {
```
Platonn: In general I appreciate the arch design of this feature is consistent with the design of other site-context features.

That said, we already have an analogical service `ThemeService` which is responsible for setting the theme CSS class on the root Angular element, on the page start. However it takes the theme from the `this.config.context?.[THEME_CONTEXT_ID]?.[0]` see https://github.com/SAP/spartacus/blob/17ad050c8e339bca93e9d57b146b2e675fbadbd0/projects/storefrontlib/layout/theme/theme.service.ts#L36

I can see it's used only in the `SiteThemeSwitcherComponent`. So if the component is not present, on specific page, the site theme won't be set on the body (e.g. taking the value from the local storage).

I'd rather like the component not to be responsible for it.

And in general I have problem to understand why/how we're dealing with 2 different Spartacus global configs:
- config.context.theme - the old one; should we deprecate it?
- config.siteTheme - the new one

Could you help me understand your vision on that?

projects/storefrontlib/cms-components/misc/theme-switcher/theme-switcher.component.service.ts
```diff
@@ -0,0 +1,32 @@
...
+
+  getItems(): Observable<Theme[]> {
+    return this.siteThemeService.getAll();
+  }
+
+  getActiveItem(): Observable<string> {
+    return this.siteThemeService
+      .getActive()
+      .pipe(tap((theme) => this.themeService.setTheme(theme)));
+  }
```
Platonn: getter of the component should not set the theme for the page.
Instead maybe the `ThemeService.init()` should use the state from ngrx state that you created?
anuyways, I'm confused that now we set the theme both in the APP_BOOSTRAP_LISTENER ->  ThemeService.init()`
AND
in this component

projects/storefrontstyles/scss/theme/high-contrast/high-contrast.scss
Platonn: As an extra step, to make it easier for people to write high-contrast friendly styles, we could expose an util `@mixin`:

```scss
// _themes.scss

/* ... */

@mixin highContrastTheme() {
  .cx-theme-high-contrast-dark &,
  .cx-theme-high-contrast-light & {
    @content;
  }
}
```

So then we could use this mixin everywhere, in each component's styles e.g.:

```scss
// _searchbox.scss

/*...*/

%cx-searchbox {
  @include highContrastTheme() {
    label {
      background-color: var(--cx-color-background);
    }
  }
  
  label {
      background-color: var(--cx-color-inverted);
    }
    
    /*...*/
```

projects/core/src/site-theme/facade/site-theme.service.ts
Platonn: On the call with Hak, we agreed that
It doesn't make sense the maintain 2 configs that are conflicting a bit. So we will deprecate the old config, which is less capable (it can contain only strings). The new config can contain objects.

Steps to do:
1. deprecate the public const with `THEME_CONTEXT_ID` key for the old config `config.context[THEME_CONTEXT_ID]` 
  image

2. Don't populate this old config anymore deriving from the BaseSitesService response (it's a breaking change in behavior so wrapped in a feature toggle):
  image

3. Don't use the old config to for setting the CSS class on the document, but use the newly introduced ngrx state instead (it's a breaking change in behavior, so wrapped in a feature toggle). Moreover we'll create a long-pending subscription (lasting for the lifetime of the application, that will observe the ngrx state of the SiteTheme and on every emission it would apply the new CSS class on the document - the new value emitted from the ngrx state.
  image

4. We won't set the CSS class in the SiteThemeSwitcher component directly. But we will just set the active theme in the ngrx state. And thanks to this, the subscirption in the ThemeService will cause setting this CSS class on the document (according to above point 3.):
   image

projects/storefrontlib/cms-components/misc/theme-switcher/theme-switcher.component.service.ts
Platonn: see point (4) in the comment: https://github.com/SAP/spartacus/pull/19016#discussion_r1700429306

Platonn:
Adding enhanced diagram from today's meeting with Hak for reference.
high level overview of the proposed architecture:
image

projects/storefrontstyles/scss/theme/high-contrast/high-contrast.scss
kimhw0630: placed the relevant rules in their specific components, except for handling the Autocomplete styles in Chrome

projects/core/src/site-theme/store/site-theme-store.module.ts
kimhw0630: tested SSR mode

projects/core/src/site-theme/providers/index.ts
kimhw0630: done

projects/core/src/site-theme/facade/site-theme.service.ts
kimhw0630: done

projects/core/src/features-config/feature-toggles/config/feature-toggles.ts
```diff
@@ -430,6 +430,11 @@ export interface FeatureTogglesInterface {
    * customization buttons in the BottomHeaderSlot in SmartEdit.
    */
   cmsBottomHeaderSlotUsingFlexStyles?: boolean;
+
+  /**
+   * when enabled a new theme switcher will be used.
```
pawelfras:
```suggestion
   * When enabled, a new theme switcher will be used. It supports default theme together with High Contrast Dark and High Contrast Light.
```

projects/core/src/site-theme/config/site-theme-config.ts
```diff
@@ -0,0 +1,23 @@
...
+import { Config } from '../../config/config-tokens';
+import { SiteTheme } from '../../model/misc.model';
+
+@Injectable({
+  providedIn: 'root',
+  useExisting: Config,
+})
+export abstract class SiteThemeConfig {
+  siteTheme?: {
+    sitethemes?: Array<SiteTheme>;
```
pawelfras:
```suggestion
    siteThemes?: Array<SiteTheme>;
```

projects/core/src/site-theme/config/default-site-theme-config.ts
```diff
@@ -0,0 +1,27 @@
...
+ * SPDX-FileCopyrightText: 2024 SAP Spartacus team <spartacus-team@sap.com>
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { SiteThemeConfig } from './site-theme-config';
+
+export const defaultSiteThemeConfig: SiteThemeConfig = {
+  siteTheme: {
+    sitethemes: [
```
pawelfras:
```suggestion
    siteThemes: [
```

projects/core/src/site-theme/facade/site-theme.service.ts
```diff
@@ -0,0 +1,87 @@
...
+  private _isInitialized = false;
+  constructor(
+    protected store: Store<StateWithSiteTheme>,
+    protected config: SiteThemeConfig
+  ) {}
```
pawelfras: `major:` For DI, let's use the `inject` function instead of the constructor to prevent breaking changes in the future. See: https://wiki.one.int.sap/wiki/display/spar/Coding+Guidelines#CodingGuidelines-Useinject()functioninsteadofconstructorforDependencyInjection

projects/core/src/site-theme/providers/site_theme_storage_Id.ts
pawelfras, line 1: `site_theme_storage_Id.ts` -> `site_theme_storage_id.ts`

projects/core/src/site-theme/providers/site_theme_storage_Id.ts
```diff
@@ -0,0 +1,7 @@
+/*
+ * SPDX-FileCopyrightText: 2024 SAP Spartacus team <spartacus-team@sap.com>
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+export const SITE_THEME_STORAGE_ID = 'theme';
```
pawelfras:
```suggestion
export const SITE_THEME_STORAGE_ID = 'siteTheme';
```

projects/core/src/site-theme/services/site-theme-initializer.ts
```diff
@@ -0,0 +1,85 @@
...
+import { BaseSiteService } from '../../site-context/facade/base-site.service';
+
+@Injectable({ providedIn: 'root' })
+export class SiteThemeInitializer implements OnDestroy {
+  constructor(
+    protected siteThemeService: SiteThemeService,
+    protected siteThemePersistenceService: SiteThemePersistenceService,
+    protected configInit: ConfigInitializerService,
+    protected baseSiteService: BaseSiteService
+  ) {}
```
pawelfras: `major:` For DI, let's use the `inject` function instead of the constructor to prevent breaking changes in the future. See: https://wiki.one.int.sap/wiki/display/spar/Coding+Guidelines#CodingGuidelines-Useinject()functioninsteadofconstructorforDependencyInjection

projects/core/src/site-theme/services/site-theme-persistence.service.ts
```diff
@@ -0,0 +1,48 @@
...
+import { SITE_THEME_STORAGE_ID } from '../providers/site_theme_storage_Id';
+import { SiteThemeService } from '../facade/site-theme.service';
+
+@Injectable({ providedIn: 'root' })
+export class SiteThemePersistenceService {
+  constructor(
+    protected statePersistenceService: StatePersistenceService,
+    protected siteThemeService: SiteThemeService,
+    protected config: SiteThemeConfig
+  ) {}
```
pawelfras: `major:` For DI, let's use the `inject` function instead of the constructor to prevent breaking changes in the future. See: https://wiki.one.int.sap/wiki/display/spar/Coding+Guidelines#CodingGuidelines-Useinject()functioninsteadofconstructorforDependencyInjection

projects/core/src/site-theme/store/effects/site-themes.effect.ts
```diff
@@ -0,0 +1,106 @@
...
+        )
+      )
+    );
+
+  constructor(
+    private actions$: Actions,
+    private state: Store<StateWithSiteTheme>,
+    protected config: SiteThemeConfig,
+    protected baseSiteService: BaseSiteService
+  ) {}
```
pawelfras: `major:` For DI, let's use the `inject` function instead of the constructor to prevent breaking changes in the future. See: https://wiki.one.int.sap/wiki/display/spar/Coding+Guidelines#CodingGuidelines-Useinject()functioninsteadofconstructorforDependencyInjection

projects/core/src/site-theme/store/effects/site-themes.effect.ts
```diff
@@ -0,0 +1,106 @@
...
+
+  loadSiteThemes$: Observable<
+    SiteThemeActions.LoadSiteThemesSuccess | SiteThemeActions.LoadSiteThemesFail
+  > = createEffect(() =>
+    this.actions$.pipe(
+      ofType(SiteThemeActions.LOAD_SITE_THEMES),
+      exhaustMap(() => {
+        return this.getCustomSiteTheme().pipe(
+          map((siteTheme) => {
+            let sitethemes = [];
```
pawelfras:
```suggestion
            let siteThemes = [];
```

projects/storefrontlib/cms-components/misc/theme-switcher/theme-switcher.component.ts
```diff
@@ -0,0 +1,38 @@
...
+  selector: 'cx-theme-switcher',
+  templateUrl: './theme-switcher.component.html',
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class ThemeSwitcherComponent {
+  iconTypes = ICON_TYPE;
+
+  constructor(
+    protected themeSwitcherComponentService: ThemeSwitcherComponentService
+  ) {}
```
pawelfras: `major:` For DI, let's use the `inject` function instead of the constructor to prevent breaking changes in the future. See: https://wiki.one.int.sap/wiki/display/spar/Coding+Guidelines#CodingGuidelines-Useinject()functioninsteadofconstructorforDependencyInjection

This rule corresponds to all new classes across this PR where DI is used.

projects/storefrontlib/cms-components/misc/theme-switcher/theme-switcher.component.service.ts
```diff
@@ -0,0 +1,26 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { Injectable } from '@angular/core';
+import { SiteThemeService, SiteTheme } from '@spartacus/core';
+import { Observable } from 'rxjs';
+
+@Injectable()
+export class ThemeSwitcherComponentService {
+  constructor(protected siteThemeService: SiteThemeService) {}
```
pawelfras: major: For DI, let's use the inject function instead of the constructor to prevent breaking changes in the future. See: https://wiki.one.int.sap/wiki/display/spar/Coding+Guidelines#CodingGuidelines-Useinject()functioninsteadofconstructorforDependencyInjection

This rule corresponds to all new classes across this PR where DI is used.

projects/storefrontlib/cms-components/misc/theme-switcher/theme-switcher.component.html
```diff
@@ -0,0 +1,18 @@
...
+      *ngFor="let theme of themes"
+      value="{{ theme.className }}"
+      [selected]="(activeItem$ | async) === theme.className"
+    >
+      {{
+        theme.i18nNameKey ? (theme.i18nNameKey | cxTranslate) : theme.className
+      }}
+    </option>
+  </select>
+  <cx-icon [type]="iconTypes.CARET_DOWN" class="small"></cx-icon>
```
pawelfras:
```suggestion
  <cx-icon [type]="iconTypes.CARET_DOWN" class="small"/>
```
See: https://wiki.one.int.sap/wiki/display/spar/Coding+Guidelines#CodingGuidelines-Useself-closingcomponenttags

projects/core/src/site-theme/facade/site-theme.service.ts
```diff
@@ -0,0 +1,87 @@
...
+      )
+      .subscribe((activeTheme) => {
+        if (activeTheme !== className) {
+          this.store.dispatch(
+            new SiteThemeActions.SetActiveSiteTheme(className)
+          );
+        }
+      });
+    this._isInitialized = true;
+  }
```
pawelfras: I have two concerns:
- Setting `_isInitialized`: Setting `_isInitialized` outside the RxJS stream might not be the best place because it could be set before the theme is actually changed. Apart from that, mixing reactive and static programming, in general, may lead to data mismatch so maybe let's consider doing this in the reactive way.

- Unsubscription: The lack of unsubscription in `setActive` can indeed cause memory leaks. Based on my knowledge, using `take(1)` inside nested pipe won't prevent from it.

projects/core/src/site-theme/services/site-theme-initializer.ts
```diff
@@ -0,0 +1,85 @@
...
+          )
+      )
+    );
+  }
+
+  protected getCustomSiteTheme(): Observable<string | undefined> {
+    return this.baseSiteService.get().pipe(
+      take(1),
+      map((baseSite) => baseSite?.theme)
+    );
```
pawelfras:
```suggestion
    return this.baseSiteService.get().pipe(
      map((baseSite) => baseSite?.theme)
      take(1),
    );
```

projects/storefrontstyles/scss/_theme.scss
```diff
@@ -8,6 +8,15 @@
...
+@import 'theme/high-contrast/high-contrast';
 // new theme below to uncomment on next major or with impex
 // @import 'theme/santorini-updated/variables';
+
+@mixin cx-highContrastTheme() {
+  .cx-theme-high-contrast-dark &,
+  .cx-theme-high-contrast-light & {
+    @content;
+  }
+}
```
pawelfras: `minor:` I'm not sure whether this is the best place to keep this mixin. I'd consider moving it to `theme/high-contrast/mixins`.

projects/core/src/site-theme/store/effects/site-themes.effect.ts
```diff
@@ -0,0 +1,106 @@
...
+@Injectable()
+export class SiteThemesEffects {
+  protected logger = inject(LoggerService);
+
+  loadSiteThemes$: Observable<
+    SiteThemeActions.LoadSiteThemesSuccess | SiteThemeActions.LoadSiteThemesFail
+  > = createEffect(() =>
+    this.actions$.pipe(
+      ofType(SiteThemeActions.LOAD_SITE_THEMES),
+      exhaustMap(() => {
```
pawelfras: `query:` what's the reason of having `exhaustMap` here?

projects/core/src/site-theme/store/actions/site-themes.action.ts
```diff
@@ -0,0 +1,48 @@
...
+  readonly type = LOAD_SITE_THEMES_SUCCESS;
+  constructor(public payload: SiteTheme[]) {}
+}
+
+export class SetActiveSiteTheme implements Action {
+  readonly type = SET_ACTIVE_SITE_THEME;
+  constructor(public payload: string) {}
+}
+
+export class SiteThemeChange implements Action {
```
pawelfras, line 17: `query:` Can you help me to understand what's the purpose of this action?

projects/core/src/site-theme/providers/site-theme-initializer-providers.ts
```diff
@@ -0,0 +1,26 @@
...
+import { SiteThemeInitializer } from '../services/site-theme-initializer';
+
+export function initializeTheme(
+  siteThemeInitializer: SiteThemeInitializer
+): () => void {
+  const result = () => {
+    siteThemeInitializer.initialize();
+  };
+  return result;
+}
```
pawelfras:
```suggestion
export function initializeTheme(): () => void {
  const siteThemeInitializer = inject(SiteThemeInitializer);
  return () => {
    siteThemeInitializer.initialize();
  };
}
```

then in the provider: 
```
{
  provide: APP_INITIALIZER,
  useFactory: initializeTheme,
  multi: true,
},
```
Such a change helps prevent breaking changes in `initializeTheme` function in case of new injection is needed.

projects/core/src/site-theme/store/actions/site-themes.action.ts
kimhw0630: this is action for theme changes, activateSiteTheme$ is an effect that listens to changes in the site theme, buffers those changes, and then dispatches an action when both previous and current are not same.

projects/core/src/site-theme/providers/site-theme-initializer-providers.ts
kimhw0630: thanks

projects/storefrontstyles/scss/_theme.scss
kimhw0630: thanks I will move it

projects/core/src/site-theme/facade/site-theme.service.ts
kimhw0630: thanks, I will wrap the subscription in setActive with a Subscription, but I will keep _isInitialized separate because it can be called from two places, and the second call might end up being uninitialized due to waiting for the subscription. We only need _isInitialized to check if the theme has been initialized.

projects/core/src/site-theme/store/effects/site-themes.effect.ts
kimhw0630: just ensure one request is processed at a time

projects/core/src/site-theme/store/actions/site-themes.action.ts
```diff
@@ -0,0 +1,48 @@
...
+ */
+
+import { Action } from '@ngrx/store';
+import { SiteTheme } from '../../../model/misc.model';
+
+export const LOAD_SITE_THEMES = '[Site-theme] Load site themes';
+export const LOAD_SITE_THEMES_FAIL = '[Site-theme] Load site themes Fail';
+export const LOAD_SITE_THEMES_SUCCESS = '[Site-theme] Load site themes Success';
+export const SET_ACTIVE_SITE_THEME = '[Site-theme] Set Active site theme';
+export const SITE_THEME_CHANGE = '[Site-theme] site theme Change';
```
pawelfras: `minor:` let's follow the convention and start each word with the capital letter, e.g. `[Site-theme] Load Site Themes`

projects/core/src/site-theme/store/actions/site-themes.action.ts
pawelfras: Ok, now I see that the reason for this might be tracking the whole flow - thanks to dispatching this action we can verify in the Redux DevTools that the theme changed (and that's all). Ok, let's keep it.

projects/core/src/site-theme/facade/site-theme.service.ts
kimhw0630: moved  _isInitialized inside

projects/core/src/site-theme/facade/site-theme.service.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+const mockSiteThemeConfig: SiteThemeConfig = {
+  siteTheme: {
+    themes: [{ i18nNameKey: 'dark', className: 'dark', default: false }],
+  },
+};
+export abstract class SiteThemeConfig {
+  siteTheme?: {
+    themes?: Array<SiteTheme>;
+  };
+}
```
pawelfras: `minor:` let's use class from `site-theme-config.ts`

projects/core/src/site-theme/facade/site-theme.service.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+    TestBed.configureTestingModule({
+      imports: [
+        StoreModule.forRoot({}),
+        EffectsModule.forRoot([]),
+        SiteThemeStoreModule,
+      ],
+      providers: [
+        SiteThemeService,
+        { provide: BaseSiteService, useClass: MockBaseSiteService },
+        { provide: SiteContextConfig, useValue: mockSiteThemeConfig },
```
pawelfras:
```suggestion
        { provide: Config, useValue: mockSiteThemeConfig },
```

projects/core/src/site-theme/store/effects/site-themes.effect.ts
pawelfras: Ok, now I see that in this case, indeed it's good to ignore new values in the source Observable until the previous value is completely processed.

projects/core/src/site-theme/providers/site_theme_id.ts
pawelfras, line 1: `site_theme_id.ts` -> `site-theme-id.ts`
sorry, my bad 🙏

projects/core/src/site-theme/facade/site-theme.service.ts
```diff
@@ -0,0 +1,90 @@
...
+          this._isInitialized = true;
+          this.store.dispatch(
+            new SiteThemeActions.SetActiveSiteTheme(className)
+          );
+        }
+      }),
+      map(() => undefined),
+      take(1)
+    );
+  }
```
pawelfras: `major:` let's clean up this method a bit
```suggestion
  setActive(className: string): void {
    this.isValidTheme(className)
      .pipe(filter(Boolean), take(1))
      .subscribe(() => {
        this.store.dispatch(new SiteThemeActions.SetActiveSiteTheme(className));
      });
    this._isInitialized = true;
  }
```
- use `take(1)` to handle unsubscription
- make a subscription here to make usage of this method cleared
- remove if statement `if (activeTheme !== className)` - I think this micro improvement provides more code complexity than the real value.
- let's set `isInitialized` outside the stream. We have a race condition between `this.siteThemePersistenceService.initSync()` and `this.setFallbackValue()` but I couldn't find the reason. Let's add `TODO` with a link to the separate ticket to investigate the issue and found a better way of checking whether theme is initialized, but let's not block this feature.

projects/core/src/site-theme/services/site-theme-persistence.service.ts
```diff
@@ -0,0 +1,50 @@
...
+  }
+
+  protected onRead(valueFromStorage: string | undefined): void {
+    if (!this.siteThemeService.isInitialized() && valueFromStorage) {
+      this.siteThemeService.setActive(valueFromStorage).subscribe(() => {
+        this.finalizeInitialization();
+      });
+    } else {
+      this.finalizeInitialization();
+    }
```
pawelfras:
```suggestion
    if (!this.siteThemeService.isInitialized() && valueFromStorage) {
        this.siteThemeService.setActive(valueFromStorage);
      });
    }  
    if (!this.initialized$.closed) {
      this.initialized$.next(undefined);
      this.initialized$.complete();
    }
```
- usage of `setActive` doesn't require `subscribe` due to changes in the service itself
- cleanup of `initialized$` has to to be done either way, so let's move it outside `else` block

projects/core/src/site-theme/services/site-theme-initializer.ts
```diff
@@ -0,0 +1,82 @@
...
+  protected setDefaultValue(
+    siteTheme: string | undefined,
+    config: SiteThemeConfig
+  ): void {
+    const defaultTheme =
+      config.siteTheme?.siteThemes?.find((theme) => theme.default)?.className ||
+      siteTheme;
+    if (!this.siteThemeService.isInitialized() && defaultTheme) {
+      this.siteThemeService.setActive(defaultTheme).subscribe();
+    }
```
pawelfras:
```suggestion
    if (!this.siteThemeService.isInitialized && defaultTheme) {
      this.siteThemeService.setActive(defaultTheme);
    }
```
due to changes in `SiteThemeService`

projects/core/src/site-theme/facade/site-theme.service.ts
```diff
@@ -0,0 +1,90 @@
...
+  protected store = inject(Store<StateWithSiteTheme>);
+  protected config = inject(SiteThemeConfig);
+  protected subscription = new Subscription();
+
+  ngOnDestroy(): void {
+    this.subscription.unsubscribe();
+  }
+  // init(className: string): Observable<string> {
+
+  // }
```
pawelfras: `major:` let's remove comments and unused code

projects/storefrontlib/cms-components/misc/theme-switcher/theme-switcher.component.service.ts
```diff
@@ -0,0 +1,26 @@
...
+  getItems(): Observable<SiteTheme[]> {
+    return this.siteThemeService.getAll();
+  }
+
+  getActiveItem(): Observable<string> {
+    return this.siteThemeService.getActive();
+  }
+
+  setActive(value: string): void {
+    this.siteThemeService.setActive(value).subscribe();
```
pawelfras:
```suggestion
    this.siteThemeService.setActive(value)
```

projects/core/src/site-theme/facade/site-theme.service.ts
```diff
@@ -0,0 +1,78 @@
...
+import { SiteTheme } from '../../model/misc.model';
+import { SiteThemeConfig } from '../config/site-theme-config';
+import { isNotNullable } from '../../util/type-guards';
+import { StateWithSiteTheme } from '../store/state';
+import { SiteThemeSelectors } from '../store/selectors';
+import { SiteThemeActions } from '../store/actions';
+
+@Injectable()
+export class SiteThemeService implements OnDestroy {
+  protected _isInitialized = false;
```
pawelfras:
```suggestion
  private _isInitialized = false;
```
I really hope we'll get rid of it soon so let's hide it from the customers :)

pawelfras (dismissed):

pawelfras:
LGTM. Let's consider a11y retesting before merging.

projects/core/src/site-theme/facade/site-theme.service.ts
```diff
@@ -0,0 +1,78 @@
...
+   * Sets the active theme className.
+   */
+  // TODO(#8153): If we move this._isInitialized into an RxJS stream, it may create a race condition between SiteThemePersistenceService:onRead and SiteThemeInitializer:this.setFallbackValue().
+  setActive(className: string): void {
+    this.isValidTheme(className)
+      .pipe(filter(Boolean), take(1))
+      .subscribe(() => {
+        this.store.dispatch(new SiteThemeActions.SetActiveSiteTheme(className));
+      });
+    this._isInitialized = true;
```
pawelfras:
```suggestion
    this.subscription = this.isValidTheme(className)
      .pipe(filter(Boolean), take(1))
      .subscribe(() => {
        this.store.dispatch(new SiteThemeActions.SetActiveSiteTheme(className));
      });
    this._isInitialized = true;
```
`major:` Only now I found this subscription is not handled. We need to adjust the code.

pawelfras: actually, unsubscription is handled by `take(1)` operator. OnDestroy interface and `protected subscription` are redundant here. Let's remove them instead.

pawelfras (dismissed):

projects/storefrontapp-e2e-cypress/cypress/e2e/regression/site-theme/site-theme.e2e.cy.ts
```diff
@@ -0,0 +1,60 @@
...
+    cy.get('cx-storefront').should('have.class', 'cx-theme-high-contrast-dark');
+
+    cy.get('cx-theme-switcher select')
+      .select('HC-Light')
+      .should('have.value', 'cx-theme-high-contrast-light');
+    cy.get('cx-storefront').should(
+      'have.class',
+      'cx-theme-high-contrast-light'
+    );
+  });
```
Platonn: Let's check for selecting back the "Default" theme. It's a special case, so IMHO it should be covered in e2e.

feature-libs/organization/administration/styles/_card.scss
```diff
@@ -93,17 +96,35 @@
     margin: 40px;
     background-color: var(--cx-color-inverse);
     border: solid 1px var(--cx-color-light);
+    @include cx-highContrastTheme-light {
+      border: 1px solid rgba(0, 0, 0, 0.125);
```
pawelfras, line 100: `query:` maybe we should consider moving `rgba(0, 0, 0, 0.125)` to a variable?

feature-libs/organization/administration/styles/_card.scss
```diff
@@ -93,17 +96,35 @@
...
+      }
       a {
         padding: 25px 30px;
         justify-content: flex-start;
         min-height: 74px;
 
         &:not(:last-child) {
           border-bottom: solid 1px var(--cx-color-light);
+          @include cx-highContrastTheme-light {
+            border-bottom: solid 1px rgba(0, 0, 0, 0.125);
```
pawelfras, line 123: `query:` maybe we should consider moving `rgba(0, 0, 0, 0.125)` to a variable?

projects/storefrontstyles/scss/components/product/carousel/_carousel.scss
```diff
@@ -159,4 +159,20 @@
...
+  @include cx-highContrastTheme {
+    .indicators {
+      button {
+        background-color: var(--cx-color-background);
+      }
+    }
+    .indicators button[disabled] .cx-icon {
+      color: var(--cx-color-dark);
+      background-color: var(--cx-color-light);
+      padding: 4px;
```
pawelfras, line 172: `query:` is it fine to have a "magic" value here instead of a variable? should it be easily customizable?

feature-libs/organization/administration/styles/_card.scss
kimhw0630: Currently, we don't have a global common variable for the border color, unless we decide to introduce a local one.

feature-libs/organization/administration/styles/_card.scss
kimhw0630: Currently, we don't have a global common variable for the border color, unless we decide to introduce a local one.

projects/storefrontstyles/scss/components/product/carousel/_carousel.scss
kimhw0630: I think we should create global sizing variables to standardize the entire application, but not in this PR.

projects/core/src/features-config/feature-toggles/config/feature-toggles.ts
```diff
@@ -507,6 +507,12 @@ export interface FeatureTogglesInterface {
    * customization buttons in the BottomHeaderSlot in SmartEdit.
    */
   cmsBottomHeaderSlotUsingFlexStyles?: boolean;
+
+  /**
+   * When enabled, a new theme switcher will be used.
+   * It supports default theme together with High Contrast Dark and High Contrast Light.
```
Platonn, line 513: It's a "breaking change toggle", so let's be empathetic and help customers understand the scope of the breaking changes, so they can estimate the amount of time required for them to adapt to this breaking change.

```suggestion
   * 1. It uses the new `SiteThemeService` as the source of truth for the "site theme" value
   * (this value can change over time, e.g. when selecting new value in the new `ThemeSwitcherComponent`).
   * Previously the "site theme" could be set only on the page start (via the static config `config.context.theme` or via CMS, when using the feature of the "automatic site-context configuration").
   * 2. Now, when no custom theme is selected, the default theme value is an empty string `''`,
   * unless you configure it differently via the global config `config.context.theme` (or via CMS).
   * Previously, there the non-defined theme had a value `undefined`.
```

projects/core/src/features-config/feature-toggles/config/feature-toggles.ts
```diff
@@ -507,6 +507,12 @@ export interface FeatureTogglesInterface {
...
+   */
+  useNewSiteThemeSwitcher?: boolean;
```
Platonn: 1. The name of the feature toggle (aka "breaking change toggle") is a bit misleading. It' doesn't enable the Theme Switcher Component itself (the component is enabled via sample data). The toggle _changes the behavior of the existing mechanism of initializing the site theme_. What about the name `useSiteThemeService`?

2. It's a "breaking change toggle", so let's be empathetic and help customers understand the scope of the breaking changes, so they can estimate the amount of time required for them to adapt to this breaking change. For more, see [coding guidelines](https://wiki.one.int.sap/wiki/pages/viewpage.action?spaceKey=spar&title=Coding+Guidelines#CodingGuidelines-WriteaJSDocdescriptionofaFeatureToggle). Let's describe what the breaking changes in the behavior are, for example:

```suggestion
   * 1. It uses the new `SiteThemeService` as the source of truth for the "site theme" value
   * (this value can change over time, e.g. when selecting new value in the new `ThemeSwitcherComponent`).
   * Previously the "site theme" could be set only on the page start (via the static config `config.context.theme` or via CMS, when using the feature of the "automatic site-context configuration").
   * 2. Now, when no custom theme is selected, the default theme value is an empty string `''`,
   * unless you configure it differently via the global config `config.context.theme` (or via CMS).
   * Previously, there the non-defined theme had a value `undefined`.
   */
  useSiteThemeService?: boolean;
```

kimhw0630: done

projects/storefrontlib/cms-components/misc/theme-switcher/theme-switcher.component.ts
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * Component for switching themes.
+ */
+@Component({
+  selector: 'cx-theme-switcher',
+  templateUrl: './theme-switcher.component.html',
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class ThemeSwitcherComponent {
```
Platonn: [MEDIUM]
As noted in some of previous reviews, **for consistency**, it would be good to align the naming in favor of `SiteTheme` wherever possible (not just `Theme`). But in this PR we have many places using `ThemeSwitcher` instead of the `SiteThemeSwitcher`
- in classes' names
- in component's selector cx-...
- in i18n keys
- file names
- (basically everywhere in this PR, where it makes sense)

projects/core/src/site-theme/config/site-theme-config.ts
```diff
@@ -0,0 +1,23 @@
...
+
+@Injectable({
+  providedIn: 'root',
+  useExisting: Config,
+})
+export abstract class SiteThemeConfig {
+  siteTheme?: {
+    optionalThemes?: Array<SiteTheme>;
+  };
+}
```
Platonn, line 19: [Official docs reminder]
Let's make sure we document in our official docs for customers clearly the Site Theming configuration, e.g.:
- you can configure the default site theme value (css class) either via the Spartacus static config `config.context.theme` or you can auto-populate it from CMS, if you're using the feature of "automatic site-context configuration"
- you can configure the list of the optional themes (to be displayed in the `ThemeSwitcherComponent`) via the config `config.siteTheme.optionalThemes`

feature-libs/checkout/base/styles/components/steps/delivery-method/_delivery-mode.scss
```diff
@@ -65,4 +65,12 @@
       }
     }
   }
+
+  @include cx-highContrastTheme {
```
Platonn, line 69: [Official docs reminder]
1. Let's make sure to document in the official docs, how customers can provide high-contrast capabilities to their custom-made components, so they are compatible with our OOTB high-cotrast themes. Let's explain there that ideally their custom styles should depend on our OOTB global CSS custom properties (aka "CSS variables") like `--cx-color-background` etc. Then there is a higher chance their components will automatically benefit from the OOTB high contrast themes (which alter the values of those global "CSS variables".
And let's explain there that if their components need any custom styles just for the high contrast themes (like some of our OOTB components, including `_delivery-mode.scss` for example), then they can use our mixin: 
```scss
@include cx-highContrastTheme {
   // here styles specific only to the high contrast themes
```
Or if they need to be more specific about the light/dark high contrast themes, they can use other mixins: `cx-highContrastTheme-dark` or `cx-highContrastTheme-light`

2. After merging it, let's communicate to all other teams they should test manually their features in both high-contrast modesThen they should make fixes to their components, if needed. I mean especially the integration features, which you had no chance testing because you didn't have an access to their integraiton servers. CAUTION: we don't have sample data for it yet in the DEV nor CI server , so we need to communicate it only when the sample data are there + note in the announcement to other teams that such-and-such sample data are needed (e.g. in case they run their own Commerce servers with their own sample data fork)

projects/core/src/site-theme/facade/site-theme.service.ts
```diff
@@ -0,0 +1,107 @@
...
+      .subscribe((activeTheme: string | null) => {
+        if (activeTheme !== className && this.isValid(className)) {
+          this.store.dispatch(
+            new SiteThemeActions.SetActiveSiteTheme(className)
+          );
+        }
+      });
+  }
+
+  isValid(className: string): boolean {
```
Platonn: [MEDIUM]
no need to be public
```suggestion
  protected isValid(className: string): boolean {
```

projects/storefrontapp-e2e-cypress/cypress/helpers/site-theme.ts
```diff
@@ -0,0 +1,45 @@
+/*
+ * SPDX-FileCopyrightText: 2024 SAP Spartacus team <spartacus-team@sap.com>
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+export function interceptToAddThemeCompnent() {
```
Platonn, line 8: [MINOR]
Let's leave a code comment here, that this is only a temporary solution, until we deploy the proper sample data to the CI server. And then this code should be removed.
Please also attach in the code comment a relevant JIRA ticket's number - for removing this piece of mocking logic.

projects/storefrontlib/cms-components/misc/theme-switcher/theme-switcher.component.ts
kimhw0630: done

projects/core/src/site-theme/facade/site-theme.service.ts
kimhw0630: done

projects/core/src/features-config/feature-toggles/config/feature-toggles.ts
```diff
@@ -507,6 +507,16 @@ export interface FeatureTogglesInterface {
...
+
+  /**
+   * 1. It uses the new `SiteThemeService` as the source of truth for the "site theme" value
+   * (this value can change over time, e.g. when selecting new value in the new `SiteThemeSwitcherComponent`).
+   * Previously the "site theme" could be set only on the page start (via the static config `config.context.theme` or via CMS, when using the feature of the "automatic site-context configuration").
+   * 2. Now, when no custom theme is selected, the default theme value is an empty string `''`,
+   * unless you configure it differently via the global config `config.context.theme` (or via CMS).
+   * Previously, there the non-defined theme had a value `undefined`.
+   */
+  useNewSiteThemeSwitcher?: boolean;
```
pawelfras: Regarding comment from @Platonn: https://github.com/SAP/spartacus/pull/19016#discussion_r1752088208
I've noticed the JSDoc is adjusted, but the feature toggle name has not changed. I'd opt for changing the name to the proposed one.

kimhw0630: I missed.. i just updated

Platonn (approved):

PR #19020, opened 2024-07-02
fix: (CXSPA-930) - Navigation UI non flyout aria roles.

Pio-Bar:
Ticket: [CXSPA-930](https://jira.tools.sap/browse/CXSPA-930)

Improves JAWS readout for non flyout navigation in the footer. It should not affect other implementations of the component.


Zeyber (changes requested):
We should add unit tests for these attribute conditions.

Zeyber (approved):

developpeurweb (approved):
Works well on both JAWS and VO

PR #19021, opened 2024-07-02
feat: CXSPA-6890 Create toggle and optimization options for propagating errors to the server

pawelfras:
PR contains toggle for propagating async errors captured during SSR to the server. Apart from that, it contains additional optimization options to configure the rendering cache - required as we're going to handle errors, which we discourage caching.

QA steps for `propagateErrorsToServer` toggle: 

1) Tests in storefront app:
- modify `pages` property in `default-cms-config.ts` to generate 'CmsPageNotFound` error.
- run npm `run dev:ssr`
- verify whether the app fell back to CSR and `index.html` returned with status `404`
- stop the app
- go to `spartacus-features.module.ts` and inside `provideFeatureTogglesFactory` set `propagateErrorsToServer` to false
- run app and verify whether the rendered with status `200` (as it used to be before changes around ssr error handling) 
 
 
 2) Tests in customer's app:
```npx ts-node ./tools/schematics/testing.ts in the SPA root folder```
```npx @angular/cli@17 new my-app --standalone=false --ssr=false --style=scss --routing=false```
```npx @angular/cli@17 add @spartacus/schematics@latest --ssr```
- use HTTP proxy or modify config for CMS endpoints to cause an error when rendering:
```ts
provideConfig(<OccConfig>{
      backend: {
        occ: {
          endpoints: {
            pages: 'cms/pages2', // FOR STATUS 404
          },
        },
      },
    }),
```
- provide `provideFeatureToggles({propagateErrorsToServer: true}) in `app.module.ts`
- in `server.ts` inside `app` function use handlers provided by Spartacus:
```ts 

  /** */

```
- run `npm run build` and `npm run serve:ssr:my-app` to verify whether the app fell back to CSR and `index.html` returned with status `404`
- close the app and remove provided toggle
- run `npm run build` and `npm run serve:ssr:my-app` and verify whether the rendered with status `200` (as it used to be before changes around ssr error handling)

QA steps for new optimization options `avoidCachingErrors` and `cacheStrategyResolver`:
- in `server.ts`, pass `cache: true` to the `ssrOptions`
- modify `components` property in `default-cms-config.ts` to generate async error in SSR.
- run `npm run dev:ssr` and run app in the browser
- verify server logs whether request has been resolved with error and in the browser if the `index.html` returned with status `500`:
```
{
  "message": "Request is resolved with the SSR rendering error (/electronics-spa/en/USD/)",
 / ... /
}
```
- reload page and check server logs to verify if page has been taken from cache:
```
{
  "message": "Render from cache (/electronics-spa/en/USD/)",
  / ... /
}
```
- in `server.ts`, add `avoidCachingErrors: true` to the `ssrOptions`
- refresh the page a few times and ensure that log `Render from cache (/electronics-spa/en/USD/)` and rendering was triggered each time the page was reloaded
- in `server.ts`, add any custom `cacheStrategyResolver` to `ssrOptions`, e.g:
```
cacheStrategyResolver: (options, entry) => {
    console.log('MY CUSTOM CACHE STRATEGY RESOLVER');
    return !(options.avoidCachingErrors === true && Boolean(entry.error));
 }
```
- run app and verify if code from the custom `cacheStrategyResolver` was executed (If the above example was used, check server logs)


core-libs/setup/ssr/optimized-engine/rendering-cache.ts
```diff
@@ -37,7 +31,13 @@ export class RenderingCache {
         this.renders.delete(this.renders.keys().next().value);
       }
     }
-    this.renders.set(key, entry);
+    // cache only if cachingStrategyResolver return true
+    if (
+      this.options?.cacheStrategyResolver &&
+      this.options?.cacheStrategyResolver(this.options, entry)
+    ) {
```
Platonn: nitpick
```suggestion
    if (
      this.options?.cacheStrategyResolver?.(this.options, entry)
    ) {
```

core-libs/setup/ssr/optimized-engine/ssr-optimization-options.ts
```diff
@@ -130,6 +131,23 @@ export interface SsrOptimizationOptions {
    * By default, the DefaultExpressServerLogger is used.
    */
   logger?: ExpressServerLogger;
+
+  /**
+   * Caching strategy resolver that determines the behavior of caching in RenderingCache.
+   * By default, the caching strategy is based on the presence of an error.
```
Platonn:
```suggestion
   * When caching is enabled, this function tell whether the given rendering result
   * (html or error) should be cached.
   * 
   * By default, all html rendering results are cached. By default, also all errors are cached
   * unless the separate option `avoidCachingErrors` is enabled.
```

core-libs/setup/ssr/optimized-engine/ssr-optimization-options.ts
```diff
@@ -130,6 +131,23 @@ export interface SsrOptimizationOptions {
...
+    config: SsrOptimizationOptions,
+    entry: Pick<RenderingEntry, 'err' | 'html'>
+  ) => boolean;
+
+  /**
+   * Avoid caching of errors. By default, this value is false.
+   * Caching errors is not recommended because it can lead to serving stale content.
+   *
+   * NOTE: Treat this option as a feature toggle. In a future, such an option is going to be true by default.
+   */
```
Platonn: What about:
```suggestion
  /**
   * Determines if rendering errors should be skipped from caching.
   *
   * NOTE: It's a temporary feature toggle, to be removed in the future.
   *
   * It's recommended to set to `true` (i.e. errors are skipped from caching),
   * which will become the default behavior, when this feature toggle is removed.
   *
   * It only affects the default `cacheStrategyResolver`.
   * Custom implementations of `cacheStrategyResolver` may ignore this setting.
 */
```

projects/core/src/features-config/feature-toggles/config/feature-toggles.ts
```diff
@@ -84,6 +84,12 @@ export interface FeatureTogglesInterface {
    */
   productConfiguratorAttributeTypesV2?: boolean;
 
+  /**
+   * Enables the propagation of errors that occur during server-side rendering (SSR) to the server.
+   * This allows for the errors to be properly handled, ensuring that a correct response is sent to the client instead of a malformed template.
+   */
```
Platonn: What about:
```suggestion
   * In a server environment (SSR or Prerendering) it propagates all errors caught in Angular app
   * (in the Angular's `ErrorHandler` class) to the server layer.
   *
   * In SSR, such a propagation allows the server layer (e.g. ExpressJS) for handling those errors,
   * e.g. sending a proper Error Page in response to the client, 
   * instead of a rendered HTML that is possibly malformed due to the occured error.
   */
```

core-libs/setup/ssr/optimized-engine/rendering-cache.spec.ts
```diff
@@ -151,4 +160,46 @@ describe('RenderingCache with cacheSize', () => {
...
+
+      it('should not cache if avoidCachingErrors is true', () => {
+        renderingCache = new RenderingCache({
+          ...options,
+          avoidCachingErrors: true,
+        });
+        renderingCache.store('a', new Error('err'), 'a');
+        expect(renderingCache.get('a')).toBeFalsy();
+      });
+    });
```
Platonn: 1. To cover all cases, let's create tests for positive HTML results. As a result I expect the  number of tests to double.

2. let's mention `errors` in the test descriptions
```suggestion
    describe('if default cacheStrategyResolver', () => {
      it('should cache errors if avoidCachingErrors is false', () => {
        renderingCache = new RenderingCache({
          ...options,
          avoidCachingErrors: false,
        });
        renderingCache.store('a', new Error('err'), 'a');
        expect(renderingCache.get('a')).toMatchSnapshot();
      });

      it('should not cache errors if avoidCachingErrors is true', () => {
        renderingCache = new RenderingCache({
          ...options,
          avoidCachingErrors: true,
        });
        renderingCache.store('a', new Error('err'), 'a');
        expect(renderingCache.get('a')).toBeFalsy();
      });
    });
```

core-libs/setup/ssr/optimized-engine/ssr-optimization-options.ts
```diff
@@ -130,6 +131,23 @@ export interface SsrOptimizationOptions {
    * By default, the DefaultExpressServerLogger is used.
    */
   logger?: ExpressServerLogger;
+
+  /**
+   * Caching strategy resolver that determines the behavior of caching in RenderingCache.
+   * By default, the caching strategy is based on the presence of an error.
+   */
+  cacheStrategyResolver?: (
```
Platonn:
```
cacheStrategyResolver: (config, entry) => boolean
```
Maybe it's only me that see this name as very ambigious. On the first sight I don't know what is a "cacheStrategy", that it wants to resolve.

What about alternatives? Examples:

```
shouldCachePage: (config, entry) => boolean
```
```
cacheCondition: (config, entry) => boolean
```

core-libs/setup/ssr/optimized-engine/optimized-ssr-engine.spec.ts
```diff
@@ -335,6 +385,68 @@ describe('OptimizedSsrEngine', () => {
...
+      }).request('a');
+
+      tick(200);
+      engineRunner.request('a');
+      tick(200);
+      engineRunner.request('a');
+      tick(200);
+      expect(engineRunner.renders).toMatchSnapshot();
+    }));
+  });
```
Platonn: To cover all cases, let's add also tests for positive HTML - for both of those describe() groups
- 'cacheStrategyResolver option'
- 'avoidCachingErrors option'

As a result, I expect the number of tests to double

core-libs/setup/ssr/optimized-engine/optimized-ssr-engine.spec.ts
```diff
@@ -39,7 +40,7 @@ class MockExpressServerLogger implements Partial<ExpressServerLogger> {
  */
 class TestEngineRunner {
   /** Accumulates html output for engine runs */
-  renders: string[] = [];
+  renders: (string | Error)[] = [];
```
Platonn, line 43: It's my bad I've named it `renders` long time ago.
Now I understand it's a misleading name, because the sequence in this array represents the sequence of responses.
For example, there could be just 1 render that got cached and 3 responses with the same render result.
`[a-0, a-0, a-0]`

What about simple refactoring/renaming it in a separate PR in develop? (can be done anytime)

core-libs/setup/ssr/optimized-engine/optimized-ssr-engine.spec.ts
```diff
@@ -335,6 +385,68 @@ describe('OptimizedSsrEngine', () => {
...
+          cache: true,
+          avoidCachingErrors: true,
+        }).request('a');
+
+        tick(200);
+        engineRunner.request('a');
+        tick(200);
+        engineRunner.request('a');
+        tick(200);
+        expect(engineRunner.renders).toMatchSnapshot();
```
Platonn: For better readability of tests, I suggest to use write explicitly and inline here the expected shape of the array `renders`.

Extracting expected values to separate files would make sense to me only if it would disturb the readability of the tests (due to being too big values to store here AND/OR when indentations need to be reset to column 0)

IMHO for such small arrays it's not justified to require a developer to jump around files and find the expected shape of this array.

Platonn: I mean
```ts
expect(engineRunner.responses).toEqual([
  new Error('a-0'),
  new Error('a-0'),
  new Error('a-0')
]);
```

core-libs/setup/ssr/optimized-engine/optimized-ssr-engine.spec.ts
pawelfras: As we discussed, we can rename it in a separate PR. Here's a separate ticket for this purpose: https://jira.tools.sap/browse/CXSPA-7860

core-libs/setup/ssr/optimized-engine/ssr-optimization-options.ts
pawelfras: I don't mind changing this property name. What do you think about:
```
shouldCacheResponse: (config, entry) => boolean
```
```
shouldCacheRendering: (config, entry) => boolean
```
```
shouldCacheRenderingResult: (config, entry) => boolean
```

Platonn: `shouldCacheRenderingResult` is most clear to me.

+to be future proof (for adding new properties/arguments), I'd wrap args in just a single object `({ config, entry }) => ...`

Platonn (approved):

PR #19023, opened 2024-07-04
bugfix/CXSPA-7818: consider `showAnonymous` and `showLoggedIn` of GigyaRaasComponent

anjana-bl:
https://jira.tools.sap/browse/CXSPA-7818

Issue:
case 1: Consider user is logged in, and now when user opens http://localhost:4200/electronics-spa/en/USD/cdc/login, UI still loads.
case 2: Consider user is not logged in and now when user opens http://localhost:4200/electronics-spa/en/USD/my-account/gigya-profile, UI still loads.
Fix:
case 1: user should be redirected to home page
case 2: user should be redirected to login page


integration-libs/cdc/components/gigya-raas/gigya-raas.component.ts
```diff
@@ -133,4 +151,21 @@ export class GigyaRaasComponent implements OnInit {
     }
     return false;
   }
+
+  canActivate(data: GigyaRaasComponentData): Observable<boolean> {
```
Platonn: In general, it shouldn't be responsibility of components to guard a route. Instead it should be Angular route guards (which are executed earlier that any component is rendered on the anticipated page).

Othwerwise, we may potentially show a flicker of a UI of the anticipated page and then almost immediately redirect to other page.

Did you consider implementing a CMS-mapping-driven route guard instead?

Platonn: I can see the challenge is injecting a CMS component's data just for the Gigya component  🤔 

Component Data can be obtained by `uid` from `cmsService.getComponentData(uid)`

And `uid` of `GigyaRaasComponent` must be found in the whole `Page` object:
```ts
this.routingService.getNextPageContext().pipe(
  filter(isNotUndefined),
  take(1),
  switchMap((pageContext) =>
    this.cmsService.getPage(pageContext).pipe(
      map(page: Page => /* find all (one?) component of type `GigyaRaasComponent` and pick it's `id`)
```

TBH I'm not aware of any different alternative ways to access component data in the cms-mapping driven route guard.

Pros:
- routing logic is kept in Angular Route Guards (to be precise in cms-driven Route Guards). Component implementation is clean from routing resolving logic.

Cons:
- need to implement a custom logic for retrieving Gigya component ID from the anticipated `Page` object (in our custom Route Guard), just to get component's data.

What do you think about it?

anjana-bl: Hi Kris, yes it makes more sense. I have adapted this approach. Please check

integration-libs/cdc/root/guards/gigya-raas.guard.ts
```diff
@@ -0,0 +1,106 @@
...
+  protected semanticPathService = inject(SemanticPathService);
+
+  canActivate(): Observable<boolean | UrlTree> {
+    return this.routingService.getNextPageContext().pipe(
+      filter(isNotUndefined),
+      take(1),
+      switchMap((pageContext) =>
+        this.getComponentsByType(pageContext, 'GigyaRaasComponent')
+      ),
+      switchMap((components) => this.getComponentData(components)),
```
Platonn: [MINOR]
For brevity of the `canActivate()`, what about extracting this logic to a separate method `getComponentData()` ?

integration-libs/cdc/root/guards/gigya-raas.guard.ts
```diff
@@ -0,0 +1,106 @@
...
+  RoutingService,
+  SemanticPathService,
+  isNotUndefined,
+} from '@spartacus/core';
+import { Observable, catchError, filter, map, of, switchMap, take } from 'rxjs';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class GigyaRaasGuard {
```
Platonn: [MINOR]
For type-safety:
```suggestion
export class GigyaRaasGuard implements CanActivate {
```

integration-libs/cdc/root/guards/gigya-raas.guard.ts
```diff
@@ -0,0 +1,106 @@
...
+        ) {
+          return this.router.parseUrl(
+            this.semanticPathService.get('home') ?? ''
+          );
+        }
+        return true;
+      })
+    );
+  }
+}
```
Platonn: [MINOR]
Instead of repeating the logic (of saving redirect url, getting semantic paths for 'home' or 'login) that is _already implemented_ in AuthGuard and NotAuthGuard, what about _reusing_ those guards here?

e.g.
```ts
  protected authGuard = inject(AuthGuard);
  protected notAuthGuard = inject(NotAuthGuard);

  /*...*/
  
  canActivate(
    route: ActivatedRouteSnapshot,
    state: RouterStateSnapshot
  ): Observable<GuardResult> {
  return this.getComponentData().pipe(
    switchMap(componentData => {
      if(data.profileEdit !== 'false') {
        return of(true);
      }
      if (data.showAnonymous === 'false') {
        return this.authGuard.canActivate(route, state);
      }
      if (data.showLoggedIn === 'false') {
        return this.notAuthGuard.canActivate(route, state);
      }
      return of(true);
    })
  );
}  
```

integration-libs/cdc/core/models/cms.model.ts
```diff
@@ -16,4 +16,6 @@ export interface GigyaRaasComponentData extends CmsComponent {
   containerID?: string; // Container name in which CDC screen will be displayed
   linkText?: string; // Text of the button for the screen set when rendered in pop-up way
   advancedConfiguration?: string; // Advanced configuration for the screen set
+  showAnonymous?: string; //whether screenset should be visible for anonymous user
+  showLoggedIn?: string; //whether screenset should be visible for logged in user
```
Platonn: [MINOR]
For our own convenience and convenience of customers, let's change all inline //comment descriptions to /** JSDocs */

integration-libs/cdc/root/guards/gigya-raas.guard.ts
anjana-bl: done

integration-libs/cdc/root/guards/gigya-raas.guard.ts
anjana-bl: done

integration-libs/cdc/root/guards/gigya-raas.guard.ts
anjana-bl: done

integration-libs/cdc/core/models/cms.model.ts
anjana-bl: done

vigneshkrish8691 (approved):
Looks fine to me

integration-libs/cdc/components/gigya-raas/gigya-raas.guard.ts
```diff
@@ -0,0 +1,83 @@
...
+  protected routingService = inject(RoutingService);
+  protected cmsService = inject(CmsService);
+  protected authGuard = inject(AuthGuard);
+  protected notAuthGuard = inject(NotAuthGuard);
+
+  canActivate(): Observable<boolean | UrlTree> {
+    return this.getComponentData().pipe(
+      switchMap((componentData) => {
+        if (Object.keys(componentData).length === 0) {
+          return of(false);
```
dunqan, line 33: @anjana-bl This one is causing a breaking change, as the guards prevent navigation when there’s no componentData for a given component. It’s quite tedious to debug and understand from the perspective of a project that relies on libraries (in our case one page suddenly stopped to work after upgrade, because of this)

PR #19029, opened 2024-07-05
fix: (CXSPA-986) - Exclude StoreFinderHeader from aria-live area

Pio-Bar:
Ticket: [CXSPA-986](https://jira.tools.sap/browse/CXSPA-986)

Moving the search box outside the aria-live area should stop the SR bloat from happening.


Zeyber (approved):

Zeyber:
Note: I would recommend testing this works on a built app since the aria elements are moved to the router outlet.

developpeurweb (approved):

PR #19035, opened 2024-07-09
fix: (CXSPA-944) - Repeated page title fix

Pio-Bar:
Ticket: [CXSPA-944](https://jira.tools.sap/browse/CXSPA-944)

The title was read twice due to the focus returning to the body element and reading both the page title and breadcrumbs component's aria-live. We can conditionally toggle the aria-live area to avoid the repetition.


projects/storefrontlib/cms-components/navigation/breadcrumb/breadcrumb.component.ts
```diff
@@ -36,6 +53,19 @@ export class BreadcrumbComponent extends PageTitleComponent implements OnInit {
   ngOnInit(): void {
     super.ngOnInit();
     this.setCrumbs();
+    if (this.featureConfigService.isEnabled('a11yRepeatedPageTitleFix')) {
+      this.subscription.add(
+        this.router.events
+          .pipe(filter((e) => e instanceof NavigationEnd))
+          .subscribe(() => {
+            if (document.activeElement === document.body) {
+              this.ariaLiveEnabled = false;
```
Zeyber: Maybe we could make this flag an observable instead and use the async pipe for it in the template for `[ariaLive]`. Should save us all the subscription logic.

projects/storefrontlib/cms-components/navigation/breadcrumb/breadcrumb.component.ts
```diff
@@ -21,8 +29,17 @@ import { PageTitleComponent } from '../page-header/page-title.component';
...
 })
-export class BreadcrumbComponent extends PageTitleComponent implements OnInit {
+export class BreadcrumbComponent
+  extends PageTitleComponent
+  implements OnInit, OnDestroy
+{
   crumbs$: Observable<any[]>;
+  protected subscription = new Subscription();
+
+  private router = inject(Router);
```
Zeyber: I think `router` should at most be `protected`. The `featureConfigService` is only set to `private` because we plan to delete it.

Zeyber (changes requested):

projects/storefrontlib/cms-components/navigation/breadcrumb/breadcrumb.component.ts
Pio-Bar: You are right I have simplified it a bunch.

projects/storefrontlib/cms-components/navigation/breadcrumb/breadcrumb.component.ts
```diff
@@ -23,6 +30,10 @@ import { PageTitleComponent } from '../page-header/page-title.component';
 })
 export class BreadcrumbComponent extends PageTitleComponent implements OnInit {
   crumbs$: Observable<any[]>;
+  ariaLiveDisabled$: Observable<boolean> = of(false);
```
Zeyber: Maybe we should flip this to just be `ariaLive$` to be more straight forward.

Zeyber (changes requested):

projects/storefrontlib/cms-components/navigation/breadcrumb/breadcrumb.component.ts
Pio-Bar: fliped ;)

projects/storefrontlib/cms-components/navigation/breadcrumb/breadcrumb.component.ts
```diff
@@ -23,6 +30,10 @@ import { PageTitleComponent } from '../page-header/page-title.component';
 })
 export class BreadcrumbComponent extends PageTitleComponent implements OnInit {
   crumbs$: Observable<any[]>;
+  ariaLive$: Observable<boolean> = of(true);
```
Zeyber:
```suggestion
  ariaLive$: Observable<boolean> = 
    this.featureConfigService.isEnabled('a11yRepeatedPageTitleFix') 
      ? this.router.events.pipe(
        filter((e) => e instanceof NavigationEnd),
        map(() => {
          return document.activeElement !== document.body;
        })
      )
    : of(true);
```

Zeyber: Perhaps we could set the observable like this.

Zeyber (changes requested):

projects/storefrontlib/cms-components/navigation/breadcrumb/breadcrumb.component.ts
Pio-Bar: Done ;)

projects/storefrontlib/cms-components/navigation/breadcrumb/breadcrumb.component.ts
```diff
@@ -36,6 +57,14 @@ export class BreadcrumbComponent extends PageTitleComponent implements OnInit {
   ngOnInit(): void {
     super.ngOnInit();
     this.setCrumbs();
+    if (this.featureConfigService.isEnabled('a11yRepeatedPageTitleFix')) {
```
Zeyber: Not needed anymore

Zeyber (changes requested):

Zeyber (dismissed):

developpeurweb (dismissed):

Zeyber (dismissed):

Zeyber (approved):

PR #19036, opened 2024-07-09
refactor: revert breaking changes made in #17657 

Platonn:
In the past PR https://github.com/SAP/spartacus/pull/17657 within this Epic branch, we made a lot of breaking changes. In that past PR we've also renamed public a lot of properties `public payload: any` to `public error: any` which was also a breaking change.

In this PR, we:
- revert those breaking changes.
  - bring back `public payload` property, if it was renamed to `public error`
  - bring back type `any` for `payload`/`error` properties that were changed to `ErrorActionType`
  - bring back the optional marker (`?`) for `payload`/`error`, that got removed the optional marker (`?`) 
  - bring back the old order of arguments of `EntityScopedFailAction`: `scope, error` vs `error, scope`
- additionally, we:
  - deprecate the signatures with the optional marker on the `payload`/`error` property, in favor of required params
  - widen type of `payload`/`error` to `any` from too-specific types like `ErrorModel`
  - pass missing `payload`/`error` parameter to super actions (like `ErrorActionType` etc.)
- fix some unit tests as a result of all above changes

Note for a reviewer:
The PR https://github.com/SAP/spartacus/pull/19037 (not meant to be merged!) contains a full diff between this branch and the `develop` branch. You may want to check it to verify the current branch `feature/CXSPA-7198--v2` when merged to `epic/ssr-error-handling` will _really_ help to avoid _all_ breaking changes (related to ngrx actions) against `develop`

fixes https://jira.tools.sap/browse/CXSPA-7198


Platonn:
~temporarily closing, to experiment with showing diff of https://github.com/SAP/spartacus/pull/19037 in VSCode~

feature-libs/pickup-in-store/core/store/actions/pickup-location.action.ts
```diff
@@ -80,8 +81,23 @@ export const SetStoreDetailsSuccess = createAction(
...
+// to be recognizable by `CxErrorHandlerEffect` (which expects
+// `error` property in every fail action).
+// However, we keep the original property `payload` as the only
+// one required _input props_ parameter, to avoid introducing a breaking change.
+const _SetStoreDetailsFailure = createAction(
   STORE_DETAILS_FAIL,
-  props<{ error: ErrorActionType }>()
+  props<{ payload: any }>()
 );
+export const SetStoreDetailsFailure = ($props: {
```
pawelfras, line 93: `nitpick:` SonarQube is complaining about the camel case name of const tarting with a capital letter: https://sonar.tools.sap/project/issues?resolved=false&severities=MINOR&inNewCodePeriod=true&pullRequest=19036&id=composable-storefront&open=AZCZwOJEEUQNBNM4DWzB. We can either mark this issue in SonarQube as resolved or change const name.

projects/core/src/error-handling/effects-error-handler/effects-error-handler.service.ts
pawelfras, line 1: 'nitpick:' Do you mind changing the name of the `filterActions` method as part of this PR? I just realised it's a type guard so a name like `isErrorAction` could fit here better.

pawelfras (approved):

feature-libs/pickup-in-store/core/store/actions/pickup-location.action.ts
Platonn: changing name would be a breaking change.
I've only created a higher order function for what already have been a function. I just didn't change the name of the Public API, deliberately.

projects/core/src/error-handling/effects-error-handler/effects-error-handler.service.ts
Platonn: Good point. For brevity of this PR, let me do it in a separate commit, directly in the epic branch.

feature-libs/pickup-in-store/core/store/actions/pickup-location.action.ts
Platonn: I've just left a comment in SonarQube:

> I don't like to rename it to camelCase, becasue:
> - this PR is touching an existing ngrx action that is already part of the public API
> - renaming the action to camelCase would be a breaking change
> - all actions are PascalCase
> - note: original authors of this code already broke the convention of using Class-based actions, but used instead the `createAction` util for creating functinon-based actions. And they named those actions PascalCase
> 
> In other words, this PR is not the source of the problem. And IMHO there is no urgent need to refactor it to comply with SonarQube and pay the price of the API deprecation cycle

and asked Radhep if he can accept dismissing SonarQube complain

PR #19040, opened 2024-07-10
feat: Multidimensional Products

kpawelczak:
closes: [CXSPA-7465](https://jira.tools.sap/browse/CXSPA-7465)

# Display and Functionality Guidelines
In general, multi-dimensional products should behave in the same way as variant products. This means that their attributes, display logic, and user interactions should be consistent. This includes:
- display variant selectors on the Product Detail Page (PDP). Variants should only be changeable from this page.
- display selected variants on cart items and Account pages, such as Order History, Order Page, Wish List, Cancellations and Returns, and so on

1. Cart Items + Account pages
    - Display selected variants the same way as the variants feature library.
    
    _example:
    ...Product details...
    Color: Red
    Size: Medium_
    
    Note:
    
    This is already covered in all places:
```
<ng-container *ngIf="item.product?.baseOptions?.length">
      <div
        *ngFor="
          let variant of item.product?.baseOptions[0]?.selected
            ?.variantOptionQualifiers
        "
        class="cx-property"
      >
        <!-- cx-mobile-header -->
        <div class="cx-label" *ngIf="variant.name && variant.value">
          {{ variant.name }}: {{ variant.value }}
        </div>
      </div>
    </ng-container>
```

2. PDP (Product Detail Page)
    - Display variant selectors (select dropdowns and/or images) like the variants feature library.
    
    Example:
    ```
    ... other product details
    <div>Color:</div>
    <div>
        <cx-media>RED IMAGE</cx-media>
        <cx-media>BLUE IMAGE</cx-media>
    </div
    
    <div>Size:</div>
    <select>
        <option>Medium</option>
        <option>Large</option>
    </select>
    ....
    ```

3. PLP (Product Listing Page)
    - In the Variants section, Style variants are shown as icons. However, for multi-dimensional products, no icons are displayed. This is because the searchable product is only a **baseProduct** and lacks the necessary attributes. Similarly, the variant version of this multi-dimensional product also lacks these attributes. It might be a bug, but in the accelerator app, only the **baseProduct** is displayed without any icons.
    - Products cannot be added to the cart from the PLP; the add to cart button should be hidden [CXSPA-7284](https://jira.tools.sap/browse/CXSPA-7284). In the accelerator app, the order grid button is displayed instead.
4. Cart page
    - Different variants should be displayed as different items. It should be possible to change the quantity and remove the product
    - multi-d variant option cannot be changed from the cart page. Same way the variants feature.
5. Most of the multi-dimensional engine is in the `product-multi-dimensional.service`. It processes the data to ensure the template can easily display it.
6. Order grid will come most likely in the next year.
# Variant Matrix Processing

## Overview
Managing product variants means organizing and filtering different combinations of attributes. Given a scenario with the following attributes:

- **Colors**: red, blue, green
- **Sizes**: small, medium, large
- **Materials**: plastic, metal

Out of 18 possible combinations, the goal is to find 5 valid options plus one selected attribute.


feature-libs/product/multi-dimensional/assets/translations/en/productMultiDimensional.json
```diff
@@ -0,0 +1,9 @@
+{
+  "productMultiDimensional": {
+    "Style": "Style",
+    "Size": "Size",
+    "Color": "Color",
+    "Material": "Material",
+    "Fit": "Fit"
+  }
+}
```
Platonn: [MEDIUM]
As we discovered on a call, names of variant categories and variant values can be translated in CMS in backoffice. no need to translate them in Spartacus

feature-libs/product/multi-dimensional/components/guards/product-multi-dimensional.guard.ts
```diff
@@ -0,0 +1,80 @@
...
+          }
+          return of(true);
+        })
+      );
+  }
+
+  protected findPurchasableProductCode(product: Product): string | undefined {
+    if (product.variantOptions?.length) {
+      const results: VariantOption[] = product.variantOptions.filter(
+        (variant: VariantOption) => variant.stock && variant.stock.stockLevel
```
Platonn: [MEDIUM]
As we discuss on a call, the current implementation allows for entering the PDP of the base, non-purchasable product, which is a bug. It may happen currently when no variant has a positive stock level

feature-libs/product/multi-dimensional/components/guards/product-multi-dimensional.guard.ts
```diff
@@ -0,0 +1,80 @@
...
+  SemanticPathService,
+  VariantOption,
+} from '@spartacus/core';
+import { Observable, of } from 'rxjs';
+import { filter, map, switchMap, take } from 'rxjs/operators';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class ProductMultiDimensionalGuard {
```
Platonn: [MINOR]
At the first sight it's not obvious what this guard does and why it's needed.

Let's add JSDoc explaining for bigger context what would happen if such a guard wasn't implemented.

feature-libs/product/multi-dimensional/components/multi-dimensional-container/multi-dimensional-selector/product-multi-dimensional-selector.component.ts
```diff
@@ -0,0 +1,67 @@
...
+  ngOnChanges(changes: SimpleChanges) {
+    if (changes.product) {
+      this.categories = this.multiDimensionalService.getVariants(this.product);
+    }
+  }
+
+  changeVariant(code: string | undefined): void {
+    if (code) {
+      this.productService
+        .get(code, ProductScope.MULTIDIMENSIONAL)
```
Platonn: [MEDIUM]
Let's use the scope `LIST` to build URLs, as in other places in our system

feature-libs/product/multi-dimensional/components/multi-dimensional-container/product-multi-dimensional-container.component.ts
```diff
@@ -0,0 +1,26 @@
...
+import { Observable } from 'rxjs';
+import { distinctUntilChanged, filter, shareReplay } from 'rxjs/operators';
+import { CurrentProductService } from '@spartacus/storefront';
+
+@Component({
+  selector: 'cx-variants-multi-dimensional',
+  templateUrl: './product-multi-dimensional-container.component.html',
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class ProductMultiDimensionalContainerComponent {
```
Platonn: [MEDIUM]
1. `ProductMultiDimensionalContainerComponent` vs `ProductMultiDimensionalSelectorContainerComponent`?
(if yes, then let's align also filename)

1. Regardless of the above, for consistency let's align component `selector` with the class name

feature-libs/product/multi-dimensional/components/multi-dimensional-container/product-multi-dimensional-container.module.ts
```diff
@@ -0,0 +1,43 @@
...
+        // ProductVariantMultiDimensionalSelectorComponent
+        ProductVariantSelectorComponent: {
+          component: ProductMultiDimensionalContainerComponent,
+          guards: [ProductMultiDimensionalGuard],
+        },
+      },
+    }),
+  ],
+})
+export class ProductMultiDimensionalContainerModule {}
```
Platonn: [MEDIUM]
`ProductMultiDimensionalContainerModule` vs `ProductMultiDimensionalSelectorModule`?

feature-libs/product/multi-dimensional/core/model/augmented-core.model.ts
```diff
@@ -0,0 +1,25 @@
...
+export type VariantCategory = {
+  name: string;
+  variantOptions: VariantCategoryOption[];
+  hasImages: boolean;
+  order: number;
+};
+
+declare module '@spartacus/core' {
+  const enum ProductScope {}
+}
```
Platonn: [MEDIUM]
Models don't have weight of runtime JS. So they can be safely put in to the `/root` entry point. Moreover, it feels right to put them to `/root`, becasue potentially any other library can then simply import types from multidimensional `/root` entry pount without breaking lazy laoding of multidimensional feature.

feature-libs/product/multi-dimensional/public_api.ts
```diff
@@ -0,0 +1,7 @@
+/*
+ * SPDX-FileCopyrightText: 2024 SAP Spartacus team <spartacus-team@sap.com>
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+export * from './index';
```
Platonn: [MAJOR]
As we agreed on the call, let's move Multi-D feature to a standalone feature lib - because this lib will contribute to multiple features/views (PDP, PLP, Cart). What about name `@sparatcus/product-multidimentional` ?

Note: it might affect, among other things, also the schematics

feature-libs/product/multi-dimensional/product-multi-dimensional.module.ts
```diff
@@ -0,0 +1,17 @@
...
+import { ProductMultiDimensionalComponentsModule } from './components/product-multi-dimensional-components.module';
+import { ProductMultiDimensionalOccModule } from './occ/product-multi-dimensional-occ.module';
+
+@NgModule({
+  imports: [
+    ProductMultiDimensionalOccModule,
+    ProductMultiDimensionalComponentsModule,
+  ],
+})
+export class ProductMultiDimensionalModule {}
```
Platonn: [MEDIUM]
For consistency and correctness, lets rename all ocurrences of the word `Multidimensional` to a single word (instead of 2 words). It's a single word.
e.g.
`MultiDimensional` -> `Multidimensional`
`multi-dimensional` -> `multidimensional`
etc.

Note: this comment applies to many places in this PR

feature-libs/product/multi-dimensional/root/variants-multi-dimensional-root.module.ts
```diff
@@ -0,0 +1,29 @@
...
+import { NgModule } from '@angular/core';
+import { provideDefaultConfigFactory } from '@spartacus/core';
+import { PRODUCT_MULTIDIMENSIONAL_FEATURE } from './feature-name';
+
+export function defaultProductMultiDimensionalComponentsConfig() {
+  return {
+    featureModules: {
+      [PRODUCT_MULTIDIMENSIONAL_FEATURE]: {
+        cmsComponents: [
+          'ProductVariantSelectorComponent', // TBD trigger
```
Platonn: [MEDIUM]
As we discussed in call, let's assign a different component type to this library as a Lazy Loading trigger, to allow customers to possibly use any of those options:
(a) only simple Variant products
(b) only Multidimensional products
(c) some simple Variant products and some Multidimensional

feature-libs/product/multi-dimensional/core/services/product-multi-dimensional.service.ts
```diff
@@ -0,0 +1,122 @@
...
+   * }
+   */
+  protected sortAndCheckIfEveryOptionHasImages(
+    variantCategories: VariantCategory[]
+  ): VariantCategory[] {
+    return variantCategories
+      .map((variantCategory: VariantCategory) => {
+        const variantOptions = variantCategory.variantOptions;
+        const hasImages =
+          variantCategory.hasImages ??
```
Platonn: [QUESTION]
Should we really derive the `hasImages` from the presence of images?
As we discussed in the call, the original `hasImages` property comes from the backend CMS and informs whether "should show images", but not "whether there are any images". I guess we want to treat CMS response with the value `hasImages` as a source of truth, right?

feature-libs/product/multi-dimensional/core/services/product-multi-dimensional.service.ts
```diff
@@ -0,0 +1,122 @@
...
+import { inject, Injectable } from '@angular/core';
+import { Product, VariantMatrixElement } from '@spartacus/core';
+import { VariantCategory } from '../model/augmented-core.model';
+import { ProductMultiDimensionalImagesService } from './product-multi-dimensional-images.service';
+
+@Injectable({ providedIn: 'root' })
+export class ProductMultiDimensionalService {
+  protected imagesService = inject(ProductMultiDimensionalImagesService);
+
+  getVariants(product: Product): VariantCategory[] {
```
Platonn: [MEDIUM]
1. TBH at the first sight it's not obvious what this method does, how many responsiblites it has, what type of model conversions it performs under the hood. 
- What about adding JSDocs to it, explaining what the purpose of this conversion and what would happen if we didn't convert it (as a bigger context for the reader).
- And perhaps splitting it to smaller methods could also help readability?

2. Just a question: do you think it makes sense to move this whole logic from the Service layer to the `ProductNormalizer` layer? By gut feeling, I guess it would fit there well. But happy to know your opinion

feature-libs/product/multi-dimensional/core/services/product-multi-dimensional-images.service.ts
```diff
@@ -0,0 +1,47 @@
...
+          format: optionQualifier.image?.format,
+          url: this.getBaseUrl() + optionQualifier.image?.url,
+        };
+      });
+  }
+
+  protected getBaseUrl(): string {
+    return (
+      (this.config as OccConfig).backend?.media?.baseUrl ??
+      (this.config as OccConfig).backend?.occ?.baseUrl ??
```
Platonn: [MINOR]
Is this type casting really needed? I believe `Config` it's already well-typed with `backend` property thanks to module-augmentation which we already have in place for `OccConfig`.

projects/core/src/model/product.model.ts
```diff
@@ -209,6 +209,7 @@ export enum VariantQualifier {
   COLOR = 'color',
   THUMBNAIL = 'thumbnail',
   PRODUCT = 'product',
+  SWATCH = 'styleSwatch',
```
Platonn: [MEDIUM]
For consistency and easier discoverability, let's not shorten original values in the key name
```suggestion
  STYLE_SWATCH = 'styleSwatch',
```

feature-libs/product/multi-dimensional/assets/translations/en/productMultiDimensional.json
kpawelczak: removed

feature-libs/product/multi-dimensional/core/services/product-multi-dimensional.service.ts
kpawelczak: I believe it's in the customer's best interest to present the best possible UI. If the "hasImages" flag is set to true but not every element contains an image, it would be preferable not to display any images. Utilizing the logger, we can notify the customer about which elements do not have an image. This would allow them to fix it more easily.

feature-libs/product/multi-dimensional/core/services/product-multi-dimensional-images.service.ts
kpawelczak: fixed

projects/core/src/model/product.model.ts
kpawelczak: fixed

feature-libs/product/multi-dimensional/components/multi-dimensional-container/multi-dimensional-selector/product-multi-dimensional-selector.component.ts
kpawelczak: I'm not sure I understand this. If I use the `LIST` scope, I'll lose some of the product attributes I need to determine which product to navigate to.
The "variants" library does the same thing.

feature-libs/product/multi-dimensional/product-multi-dimensional.module.ts
kpawelczak: In SAP wikis it's called Multi-Dimensional, should I proceed with this change?

feature-libs/product/multi-dimensional/core/services/product-multi-dimensional.service.ts
kpawelczak: 1. The `variantMatrix` structure is not an easy one, no matter how well I document it, it still might create some confusion. Anyways to address this:

- I moved the existing comments to the `getVariants` method and added some new ones.

- to improve readability, I created two new methods, `createVariantCategory` and `createVariantOptionCategory`.  Any further splitting might complicate the function.

kpawelczak: In addition, I've removed the sorting function as it's not needed -> the variantMatrix is sorted in the backend. Which is not the case for the `baseOptions`

feature-libs/product/multi-dimensional/components/guards/product-multi-dimensional.guard.ts
kpawelczak: I've already addressed it, but need to test it further. I might do it as a sub-ticket.

feature-libs/product/multi-dimensional/components/guards/product-multi-dimensional.guard.ts
kpawelczak: I've added JSDocs

feature-libs/product/multi-dimensional/components/multi-dimensional-container/product-multi-dimensional-container.module.ts
kpawelczak: Honestly, I retained the container component from the previous implementation, which likely had its reasons. I suspect this is related to the order grid functionality.

Let's look into the future: Should this container component be responsible for deciding whether to display "variant option selectors" or the "order grid form"?

Please consider the following:

- The order grid is intended only for base products.

- Variant selectors are meant for variant products of a base product.

feature-libs/product/multi-dimensional/public_api.ts
kpawelczak: From what I understand, you don't want it to be a sub-library of the product library?

feature-libs/product/multi-dimensional/core/model/augmented-core.model.ts
kpawelczak: moved to `root`

feature-libs/product/multi-dimensional/root/variants-multi-dimensional-root.module.ts
kpawelczak: I'll do it a separate sub-ticket

feature-libs/product/multi-dimensional/components/multi-dimensional-container/product-multi-dimensional-container.component.ts
kpawelczak: See the below comment.

feature-libs/product/multi-dimensional/components/multi-dimensional-container/multi-dimensional-selector/product-multi-dimensional-selector.component.ts
Platonn: As far as I can see this method is a `void`. And the Product data are used in the `subscribe()` body only to create an URL (by a semantic name 'product' and params: product) and navigate to this URl.

The Product "loading scope" of type "LIST" should have all necessary details needed to build an URL (e.g. code, name, etc.) -> `/prodcut/:productCode/:name`

(but customers may want to include more meaninful properties into the URL, for SEO reasons, e.g. `/:category1/:category2/:category3/:name/:productCode`. So we need to make sure the `params: product` that we pass to the `routingService.go()` contain all minimal essential data for customers to build custom SEO-friendly URLs and navigate to them

feature-libs/product/multi-dimensional/components/multi-dimensional-container/product-multi-dimensional-container.module.ts
Platonn: Good insight. Thanks!

Should the "order grid form" really be displayed in the same place as the "variant selector"?

Variant selector is displayed in the right hand column
But the full grid view should be full-width I guess?
Therefore should it be really the same component?

feature-libs/product/multi-dimensional/public_api.ts
Platonn: Correct. I meant to not create a sub-library of product lub, but instead: a standalone library.
Any objections? I'm happy to learn your point of view.

feature-libs/product/multi-dimensional/product-multi-dimensional.module.ts
Platonn: In general, I don't mind either option. Just let's be consistent everywhere in our codebase.

My arguments for `multidimensional` as 1 word:
- Backend already returns us a property `product.multidimensional`
- it's correct according to EN dictionary https://dictionary.cambridge.org/dictionary/english/multidimensional

feature-libs/product/multi-dimensional/core/services/product-multi-dimensional.service.ts
Platonn: Nice arguments. I agree with your reasoning.

Additional comments:

1. Because it's not obvious, let's explain it also in a //code comment, _why_ we implement it so

2. To my understanding this implementation is not in line with your intentions described in your answer above.

I thought you want to set `hasImages` with `&&` (when both conditions are true), but not be satisfied only with the original `hasImages` value:
```diff
          const hasImages =
-           variantCategoryGroup.hasImages ??
+           variantCategoryGroup.hasImages &&
            variantOptions.every((option) => option.images.length);
```
Would you agree?

feature-libs/product/multi-dimensional/components/guards/product-multi-dimensional.guard.ts
```diff
@@ -0,0 +1,99 @@
...
+   * @param {Product} product - The product with variant options.
+   * @returns {Observable<boolean | UrlTree>} - An observable that resolves to a `UrlTree` for
+   * redirection if a valid product code is found, or `false` if no valid code is available.
+   *
+   * @description
+   * This method examines the product's variant options to find one with available stock. If a valid
+   * variant is found, it fetches the corresponding product and generates a URL for redirection. If
+   * no valid variant code is found, it resolves to `false`.
+   */
+  protected findPValidProductCodeAndReturnUrlTree(
```
Platonn: typo, unnecesary `P` in `findP...`

feature-libs/product/multi-dimensional/core/services/product-multi-dimensional.service.ts
```diff
@@ -6,66 +6,67 @@
...
+   * Size:      1     2         1      (2)
+   * Fit:     M  L  M   L     M   L   M   (L)
+   *
+   * The method extracts all available colors (e.g., Red and Blue) and recursively
+   * retrieves options until all variant options are exhausted.
+   *
+   * @see checkIfEveryOptionHasImages
+   * @see imagesService.getVariantOptionImages
+   */
+  getVariants(product: Product): VariantCategoryGroup[] {
```
Platonn: Very very nice JSDoc!

**What I'm still not understanding: does this method also filter out variants that should not be displayed in PDP**? If yes, I think it's necessary to document it and also give a bigger context: **why** on PDPs we want to display only specific variants (but others not -> to only allow changing one variant dimension at a time).

Question for bigger context:
Do we plan to have analogical method in the future, but that shouldn't filter out any variants? - e.g. for the purpose of the variant grid view (instead of for variant selectors)?

feature-libs/product/multi-dimensional/core/services/product-multi-dimensional.service.ts
```diff
@@ -74,29 +75,88 @@ export class ProductMultiDimensionalService {
...
-      };
-    });
+    variantCategories: VariantCategoryGroup[]
+  ): VariantCategoryGroup[] {
+    return variantCategories.map(
+      (VariantCategoryGroup: VariantCategoryGroup) => {
+        const variantOptions = VariantCategoryGroup.variantOptions;
+        const hasImages =
+          VariantCategoryGroup.hasImages ??
+          variantOptions.every((option) => option.images.length);
```
Platonn: To my understanding this implementation is not in line with your intentions described in the comment https://github.com/SAP/spartacus/pull/19040#discussion_r1687712623

I thought you want to set `hasImages` with `&&` (when both conditions are true), but not `??` ?:
```suggestion
        const hasImages =
          VariantCategoryGroup.hasImages &&
          variantOptions.every((option) => option.images.length);
```

feature-libs/product/multi-dimensional/core/services/product-multi-dimensional.service.ts
Platonn: Thanks for the readability improvements!

feature-libs/product/multi-dimensional/components/guards/product-multi-dimensional.guard.ts
Platonn: Very nice, thanks!

feature-libs/product/multi-dimensional/components/multi-dimensional-container/multi-dimensional-selector/product-multi-dimensional-selector.component.ts
kpawelczak: Oh God, I didn't realize it is the component. I thought you referred to the component guard...

kpawelczak: fixed

feature-libs/product/multi-dimensional/core/services/product-multi-dimensional.service.ts
kpawelczak: I've added more details:
 ```
 The method extracts all available colors (e.g., Red and Blue) and recursively retrieves options
until all variant options are exhausted, based on the currently selected options. For example the second row
contains multiple sizes of the same value, representing different products. Since the user can change
only one variant at a time, we need to identify only the products that reflect the current selections,
as shown in the tree with rows like (Blue), (2), (L).

Color:        Red             (Blue)
Size:                       1      (2)
Fit:                              M   (L)
 ```
 Question for bigger context:
For the order grid, a different method will be used to filter out values based on different assumptions. For example, the grid might display:

- Columns -> Sizes

- Rows -> Colors

I don't know many details about it, but I know that the exact display of the order grid values has not been finalized and is still to be decided.

feature-libs/product/multi-dimensional/core/services/product-multi-dimensional.service.ts
kpawelczak: thanks, 
I meant to use `variantCategoryGroup.hasImages ? variantOptions.every((option) => option.images.length) : false` and skip checking the `variantOptions` array if `hasImages = false`. However, I don't think it would make a huge difference.

feature-libs/product/multi-dimensional/product-multi-dimensional.module.ts
kpawelczak: I understand why you want to do it, but what I see is that a new term, "Multi-Dimensional," has been created specifically for this feature. It's referred to this way in every wiki. The way I see it, "multidimensional" and "Multi-Dimensional" are different terms, with the latter suggesting a particular feature.
e.g.
https://microlearning.opensap.com/media/Multi-Dimensional+Product+Variants+-+SAP+Commerce+Cloud/1_jmulme9b

feature-libs/product/multi-dimensional/components/guards/product-multi-dimensional.guard.ts
kpawelczak: fixed

feature-libs/product/multi-dimensional/components/multi-dimensional-container/product-multi-dimensional-container.module.ts
kpawelczak: I've made changes in the other pr (https://github.com/SAP/spartacus/pull/19083/files) to rename this module to ProductMultiDimensionalSelectorModule, also removed the container module

feature-libs/product/multi-dimensional/public_api.ts
kpawelczak: I've moved it to the standalone library in a separate pr https://github.com/SAP/spartacus/pull/19083/files

feature-libs/product-multi-dimensional/selector/components/guards/product-multi-dimensional-selector.guard.ts
```diff
@@ -0,0 +1,111 @@
...
+  SemanticPathService,
+  VariantOption,
+} from '@spartacus/core';
+import { Observable, of } from 'rxjs';
+import { filter, map, switchMap, take } from 'rxjs/operators';
+
+/**
+ * Guard that ensures a product is valid and navigable within a multi-dimensional context.
+ *
+ * This service manages navigation and access control for products in a multi-dimensional product catalog.
```
pawelfras:
```suggestion
 * This guard manages navigation and access control for products in a multi-dimensional product catalog.
```

feature-libs/product-multi-dimensional/selector/components/guards/product-multi-dimensional-selector.guard.ts
```diff
@@ -0,0 +1,111 @@
...
+      .pipe(
+        filter(isNotUndefined),
+        switchMap((multiDimensionalProduct: Product) => {
+          const isNotPurchasableAndHasVariantOptions =
+            !multiDimensionalProduct.purchasable &&
+            !!multiDimensionalProduct.variantOptions?.length;
+
+          return isNotPurchasableAndHasVariantOptions
+            ? this.findValidProductCodeAndReturnUrlTree(multiDimensionalProduct)
+            : of(multiDimensionalProduct.purchasable);
```
pawelfras:
```suggestion
          return isNotPurchasableAndHasVariantOptions
            ? this.findValidProductCodeAndReturnUrlTree(multiDimensionalProduct)
            : of(!!multiDimensionalProduct.purchasable);
```
To meet the return type of `canActivate` function

feature-libs/product-multi-dimensional/list/occ/config/default-occ-product-multi-dimensional-list-config.ts
```diff
@@ -0,0 +1,18 @@
...
+ */
+
+import { OccConfig } from '@spartacus/core';
+
+export const defaultOccProductMultiDimensionalListConfig: OccConfig = {
+  backend: {
+    occ: {
+      endpoints: {
+        productSearch:
+          'products/search?fields=products(code,name,summary,configurable,configuratorType,priceRange(maxPrice(formattedValue),minPrice(formattedValue)),multidimensional,price(FULL),images(DEFAULT),stock(FULL),averageRating,variantOptions),facets,breadcrumbs,pagination(DEFAULT),sorts(DEFAULT),freeTextSearch,currentQuery,keywordRedirectUrl',
```
pawelfras, line 15: `major:` The length of this line might cause linting and Sonar issues. Let's add `* eslint-disable max-len */` line before and remember about resolving Sonar issue (if needed) with a proper comment

feature-libs/product-multi-dimensional/list/root/components/price-range/product-multi-dimensional-price-range.component.ts
```diff
@@ -0,0 +1,37 @@
...
+      product.multidimensional && Object.values(priceRange).length;
+
+    if (isMultiDimensionalAndHasPriceRange) {
+      const maxPrice = priceRange.maxPrice.formattedValue;
+      const minPrice = priceRange.minPrice.formattedValue;
+      return `${minPrice} - ${maxPrice}`;
+    }
+
+    return product.price?.formattedValue ?? 0;
+  }
```
pawelfras:
```suggestion
  getProductPrice(product: Product) {
    const defaultValue = '0';

    if (!product.multidimensional) {
      return product.price?.formattedValue ?? '0';
    }

    const priceRange = product.priceRange;
    if (!priceRange || !priceRange.maxPrice || !priceRange.minPrice) {
      return product.price?.formattedValue ?? '0';
    }

    const maxPrice = priceRange.maxPrice.formattedValue;
    const minPrice = priceRange.minPrice.formattedValue;
    return `${minPrice} - ${maxPrice}`;
  }
```
Due to TS issues and few places related to the `priceRange`, I propopose vew improvements:

- Added early return if product is not `multidimensional`.
- Added early return if `priceRange`, `priceRange.maxPrice`, or `priceRange.minPrice` is not defined.
- Moved `0` to a separate constant and returned it as a default value, ensuring a consistent return type.

feature-libs/product-multi-dimensional/selector/components/selector/product-multi-dimensional-selector.component.html
```diff
@@ -0,0 +1,39 @@
...
+      >
+        <cx-media
+          *ngFor="let option of category.variantOptions"
+          (click)="changeVariant(option.code)"
+          [container]="option.images"
+          [ngClass]="{ active: option?.code === product.code }"
+          [title]="option.value"
+          class="image-variant"
+        >
+        </cx-media>
```
pawelfras:
```suggestion
        <cx-media
          *ngFor="let option of category.variantOptions"
          (click)="changeVariant(option.code)"
          [container]="option.images"
          [ngClass]="{ active: option?.code === product.code }"
          [title]="option.value"
          class="image-variant"
        />
```
`minor:` Let's use a self-closing component tag to be in line with SPA coding guidelines. See: https://wiki.one.int.sap/wiki/display/spar/Coding+Guidelines#CodingGuidelines-Useself-closingcomponenttags

feature-libs/product-multi-dimensional/selector/components/selector/product-multi-dimensional-selector.component.html
```diff
@@ -0,0 +1,39 @@
...
+      <label>{{ category.name }}</label>
+
+      <div
+        *ngIf="category.hasImages; else noImages"
+        class="image-variant-container"
+      >
+        <cx-media
+          *ngFor="let option of category.variantOptions"
+          (click)="changeVariant(option.code)"
+          [container]="option.images"
```
pawelfras: `reminder:` https://github.com/SAP/spartacus/pull/19101#discussion_r1705327359

feature-libs/product-multi-dimensional/selector/components/selector/product-multi-dimensional-selector.component.html
```diff
@@ -0,0 +1,39 @@
...
+          [title]="option.value"
+          class="image-variant"
+        >
+        </cx-media>
+      </div>
+
+      <ng-template #noImages>
+        <div class="select-variant-container">
+          <select
+            (change)="changeVariant($event.target.value)"
```
pawelfras: `reminder:` https://github.com/SAP/spartacus/pull/19101#discussion_r1705366637

projects/schematics/src/add-spartacus/schema.json
```diff
@@ -45,6 +45,8 @@
           "Future-Stock",
           "PDF-Invoices",
           "Product-Variants",
+          "Product-multi-dimensional-selector",
+          "Product-multi-dimensional-list",
```
pawelfras:
```suggestion
          "Product-Multi-Dimensional-Selector",
          "Product-Multi-Dimensional-List",
```

projects/storefrontapp-e2e-cypress/cypress/helpers/checkout-multi-dimensional.ts
```diff
@@ -0,0 +1,246 @@
...
+
+    assertAddressForm(
+      {
+        firstName: multiDUser.firstName,
+        lastName: multiDUser.lastName,
+        phone: '',
+        address: multiDUser.address,
+      },
+      'US'
+    );
```
pawelfras: Type issue:
image

projects/storefrontlib/cms-components/product/product-list/product-list-item/product-list-item.component.html
```diff
@@ -47,6 +47,7 @@ <h2 [innerHtml]="product.nameHtml"></h2>
         {{ 'productDetails.noReviews' | cxTranslate }}
       </div>
       <div
+        *ngIf="!product.multidimensional"
```
pawelfras: `query:` could it be the case that customers can already have multidimensional products in their stores despite the lack of SPA support? If so (but I'm guessing it's unlikely) this `ngIf` statement could be a breaking change.

projects/storefrontlib/cms-components/product/product-list/product-grid-item/product-grid-item.component.html
```diff
@@ -26,7 +26,7 @@
       {{ 'productDetails.noReviews' | cxTranslate }}
     </div>
   </div>
-  <div class="cx-product-price-container">
+  <div *ngIf="!product.multidimensional" class="cx-product-price-container">
```
pawelfras: `query:` could it be the case that customers can already have multidimensional products in their stores despite the lack of SPA support? If so (but I'm guessing it's unlikely) this `ngIf` statement could be a breaking change.

package.json
```diff
@@ -25,12 +25,13 @@
...
     "build:order": "npm --prefix feature-libs/order run build:schematics && nx build order --configuration production",
-    "build:libs": "nx build core --configuration production && nx build storefrontlib --configuration production && concurrently --kill-others-on-fail npm:build:schematics npm:build:user && npm run build:cart && npm run build:pdf-invoices && npm run build:order && npm run build:storefinder && concurrently --kill-others-on-fail npm:build:checkout npm:build:asm npm:build:tracking npm:build:customer-ticketing && concurrently --kill-others-on-fail npm:build:organization npm:build:product npm:build:product-configurator npm:build:requested-delivery-date && concurrently --kill-others-on-fail npm:build:estimated-delivery-date && concurrently --kill-others-on-fail npm:build:smartedit npm:build:qualtrics npm:build:assets npm:build:cds npm:build:cdc npm:build:cdp npm:build:digital-payments npm:build:epd-visualization npm:build:s4om npm:build:cpq-quote npm:build:segment-refs npm:build:opps npm:build:pickup-in-store npm:build:quote && npm run build:setup && npm run build:s4-service",
```
pawelfras: `major:` script ends with en error (before I removed `dist` and cleared Nx cache). It might be related to the order of building:
image

feature-libs/product-multi-dimensional/selector/components/guards/product-multi-dimensional-selector.guard.ts
kpawelczak: fixed

feature-libs/product-multi-dimensional/selector/components/guards/product-multi-dimensional-selector.guard.ts
kpawelczak: fixed

feature-libs/product-multi-dimensional/list/occ/config/default-occ-product-multi-dimensional-list-config.ts
kpawelczak: fixed

feature-libs/product-multi-dimensional/list/root/components/price-range/product-multi-dimensional-price-range.component.ts
kpawelczak: fixed

feature-libs/product-multi-dimensional/selector/components/selector/product-multi-dimensional-selector.component.html
kpawelczak: fixed

feature-libs/product-multi-dimensional/list/root/components/price-range/product-multi-dimensional-price-range.component.ts
kpawelczak: on second thought, this is wrong. This is what I propose:
```typescript
  getProductPrice(product: Product): string {
    const defaultValue = '0';
    if (!product.multidimensional) {
      return product.price?.formattedValue ?? defaultValue;
    }
    const priceRange = product.priceRange;

    const maxPrice = priceRange?.maxPrice?.formattedValue;
    const minPrice = priceRange?.minPrice?.formattedValue;

    return maxPrice && minPrice ? `${minPrice} - ${maxPrice}` : '';
  }
``` 
if there is no price range it should not return any price, as it would be misleading

package.json
kpawelczak: fixed

projects/storefrontapp-e2e-cypress/cypress/helpers/checkout-multi-dimensional.ts
kpawelczak: fixed

projects/storefrontlib/cms-components/product/product-list/product-list-item/product-list-item.component.html
kpawelczak: I don't see this as a breaking change. We cannot guarantee the stability of the application, if customers use unsupported features.
Consider a scenario where a customer adds a third-party library, and after updating SPA, it stops working. Would that be a breaking change?

projects/schematics/src/add-spartacus/schema.json
kpawelczak: fixed

feature-libs/product-multi-dimensional/selector/components/selector/product-multi-dimensional-selector.component.html
kpawelczak: fixed

feature-libs/product-multi-dimensional/selector/components/selector/product-multi-dimensional-selector.component.html
kpawelczak: fixed

projects/storefrontlib/cms-components/product/product-list/product-list-item/product-list-item.component.html
pawelfras: Good point! Nevertheless, compared to the usage of third-party lib, we provide a new statement to the template which will change how DOM looks in some conditions. Note that the `multidimensional` field has been part of the `Product` interface for 5 years. I guess nothing blocked users from basing their custom logic on it. That said, I believe it is a breaking change in some corner cases. The question is whether we should care and handle it with a toggle. I don't have a strong opinion on this.

pawelfras: Now I'm wondering whether we have to provide such a condition. Can you help me understand why we need it? If I understand it correctly, without it, the template for non-multidimensional products will behave as it was, which is ok.

kpawelczak: Without this condition, the multi-d product will have both product prices. I've considered many different approaches and the above is the cleanest one.

kpawelczak: As discussed, I've changed the outlet from 'after' to 'replace' and did all necessary updates

pawelfras:
Two unit tests for schematics are failing:
image
image

projects/schematics/src/shared/utils/graph-utils_spec.ts
```diff
@@ -28,7 +28,7 @@ import {
   SPARTACUS_SMARTEDIT,
   SPARTACUS_STOREFINDER,
   SPARTACUS_TRACKING,
-  SPARTACUS_USER,
+  SPARTACUS_USER, SPARTACUS_PRODUCT_MULTI_DIMENSIONAL,
```
pawelfras: `nitpick:` Let's move it to a new line

kpawelczak: I don't see it being on the same line

projects/schematics/src/add-spartacus/schema.json
```diff
@@ -185,6 +187,14 @@
             "value": "Product-Variants",
             "label": "Product - Variants"
           },
+          {
+            "value": "Product-Multi-Dimensional-Selector",
+            "label": "Product Multi-Dimensional - Selector"
+          },
+          {
+            "value": "Product-Multi-Dimensional-List",
+            "label": "Product Multi-Dimensional PLP - price range"
```
pawelfras: `nitpick:`
```suggestion
            "label": "Product Multi-Dimensional - PLP price range"
```

projects/schematics/src/shared/utils/graph-utils_spec.ts
```diff
@@ -167,7 +169,7 @@ describe('Graph utils', () => {
 
   describe('feature dependency graph', () => {
     it('should generate the correct installation order', () => {
-      expect(crossFeatureInstallationOrder).toMatchInlineSnapshot(`
+      expect(crossFeatureInstallationOrder).toMatchSnapshot(`
```
pawelfras: `minor:` I'd rather revert this change and manually modify the inline snapshot to ensure the test works as expected for the previous state before changes.

projects/storefrontapp-e2e-cypress/cypress.config.ci.ts
```diff
@@ -20,7 +20,8 @@ export default defineConfig({
   },
   videoUploadOnPasses: false,
   env: {
-    API_URL: 'https://20.83.184.244:9002',
+    // API_URL: 'https://20.83.184.244:9002',
+    API_URL: 'https://spartacus-dev-asx1.eastus.cloudapp.azure.com:8443',
```
pawelfras: `major:` let's not forget to revert this change

projects/storefrontapp-e2e-cypress/cypress.config.ts
```diff
@@ -17,7 +17,7 @@ export default defineConfig({
   env: {
     CLIENT_ID: 'mobile_android',
     CLIENT_SECRET: 'secret',
-    API_URL: 'https://40.76.109.9:9002',
+    API_URL: 'https://spartacus-dev-asx1.eastus.cloudapp.azure.com:8443',
```
pawelfras: `major:` let's not forget to revert this change

projects/storefrontapp/src/app/app.module.ts
```diff
@@ -63,7 +63,7 @@ if (!environment.production) {
     provideConfig(<OccConfig>{
       backend: {
         occ: {
-          baseUrl: environment.occBaseUrl,
+          baseUrl: 'https://spartacus-dev-asx1.eastus.cloudapp.azure.com:8443',
```
pawelfras: major: let's not forget to revert this change

tsconfig.compodoc.json
pawelfras, line 1: `minor:` It seems there is an issue with formatting in this file

tsconfig.json
pawelfras, line 1: `minor:` It seems there is an issue with formatting in this file

projects/schematics/src/shared/utils/graph-utils_spec.ts
kpawelczak: but `toMatchInlineSnapshot` doesn't work?

feature-libs/product-multi-dimensional/list/occ/config/default-occ-product-multi-dimensional-list-config.ts
pawelfras: I see the `eslint-disable` rule is still not applied. It may cause linting/Sonar issues.

kpawelczak: I  see it?
image

projects/storefrontapp-e2e-cypress/cypress.config.ci.ts
kpawelczak: reverted

projects/storefrontapp-e2e-cypress/cypress.config.ts
kpawelczak: reverted

projects/storefrontapp/src/app/app.module.ts
kpawelczak: reverted

pawelfras (dismissed):

rmch91 (approved):

PR #19044, opened 2024-07-12
fix: (CXSPA-1087) - Role presentation for cx-message wrapper

Pio-Bar:
Ticket: [CXSPA-1087](https://jira.tools.sap/browse/CXSPA-1087)

Adding presentation 'role' to the wrapping div should stop the additional step in VO.


Zeyber (dismissed):

developpeurweb (changes requested):
This fixes the repetition on VO, but now I can hear the repetition on JAWS, can you please check if it's the same for you? 
Is it possible to fix it for both?

Pio-Bar:
> This fixes the repetition on VO, but now I can hear the repetition on JAWS, can you please check if it's the same for you? Is it possible to fix it for both?

I've added additional roles that should help. I'm unsure why JAWS in particular would target those specific elements.

Zeyber (approved):

Zeyber (dismissed):

developpeurweb:
@Pio-Bar still an issue on JAWS, it seems there are two alert roles, as per screenshot, let's remove the alert role at the bottom and just leave the text. If not possible, let's remove the whole arrangement at the bottom and keep only the top alert.
duplicated role alert

Pio-Bar:
I have removed the additional form error at the bottom. Just to summarize: 

- The double stop for VO should no longer appear. 
- JAWS should not repeat the error message twice.

rmch91 (dismissed):

developpeurweb (dismissed):

Zeyber (approved):

PR #19053, opened 2024-07-16
[WIP] feat: Extend session dialog with persist of refresh_token

sdrozdsap:
Here I've commented out the line where refresh_token is removed from the storage for security reasons. It does solve the bug, where after reloading the page we are unable to prolong users session due to lack of refresh_token


sdrozdsap:
Update: the "accessible authentication" a11y requirement does not longer require any changes in Spartacus thus closing this PR

PR #19055, opened 2024-07-17
chore: CXSPA-7900 Ajust schematics to enable  option in fresh apps

pawelfras:
PR contains a small adjustment to schematics that enables `avoidCachingErrors` optimization option in a fresh application to avoid tech debt.

QA steps:
- create new app `npx @angular/cli@17 new my-app --standalone=false --style=scss --routing=false`
  `npx @angular/cli@17 add @spartacus/schematics@latest --ssr --baseUrl="https://40.76.109.9:9002"`
- check in `server.ts` whether `avoidCachingErrors` has been added:
  ```
  const ngExpressEngine = NgExpressEngineDecorator.get(engine, {
    avoidCachingErrors: true, // takes effect only when cache is enabled
  });
  ```

closes [CXSPA-7900](https://jira.tools.sap/browse/CXSPA-7900)

### DIFF
--- a/core-libs/setup/ssr/optimized-engine/optimized-ssr-engine.ts
+++ b/core-libs/setup/ssr/optimized-engine/optimized-ssr-engine.ts
@@ -5,7 +5,6 @@
  */
 
 /* webpackIgnore: true */
-import { deepMerge } from '@spartacus/core';
 import { Request, Response } from 'express';
 import * as fs from 'fs';
 import { NgExpressEngineInstance } from '../engine-decorator/ng-express-engine-decorator';
@@ -73,14 +72,13 @@ export class OptimizedSsrEngine {
     this.ssrOptions = ssrOptions
       ? {
           ...defaultSsrOptimizationOptions,
-
           // overrides the default options
           ...ssrOptions,
-          // deep merge the feature toggles
-          featureToggles: deepMerge(
-            { ...defaultSsrOptimizationOptions.featureToggles },
-            ssrOptions.featureToggles
-          ),
+          // merge feature toggles
+          featureToggles: {
+            ...defaultSsrOptimizationOptions.featureToggles,
+            ...ssrOptions.featureToggles,
+          },
         }
       : undefined;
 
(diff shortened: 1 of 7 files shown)


projects/schematics/src/add-ssr/__snapshots__/index_spec.ts.snap
```diff
@@ -213,7 +213,9 @@ import { dirname, join, resolve } from 'node:path';
 import { fileURLToPath } from 'node:url';
 import AppServerModule from './src/main.server';
 
-const ngExpressEngine = NgExpressEngineDecorator.get(engine);
+const ngExpressEngine = NgExpressEngineDecorator.get(engine, {
+  avoidCachingErrors: true, // takes effect only when cache is enabled
```
Platonn: always takes effect. 
because cache is used in the case of timed out renders

pawelfras: Oh, you're right... I totally missed that.

core-libs/setup/ssr/optimized-engine/optimized-ssr-engine.ts
```diff
@@ -72,8 +73,14 @@ export class OptimizedSsrEngine {
...
       ? {
           ...defaultSsrOptimizationOptions,
+
           // overrides the default options
           ...ssrOptions,
+          // deep merge the feature toggles
+          featureToggles: deepMerge(
+            { ...defaultSsrOptimizationOptions.featureToggles },
+            ssrOptions.featureToggles
+          ),
```
Platonn: `featutreToggles` object is 1-level shallow. No need for `deepMerge`.
We can either `deepMerge` the whole object with ssr options with default ssr options.
OR
just spread explicitly on the level of `featureToggles`. I don't mind either of the options

```suggestion
          featureToggles: {
            ...defaultSsrOptimizationOptions.featureToggles,
            ...ssrOptions.featureToggles
          },
```

Platonn (approved):
LGTM code wise

PR #19058, opened 2024-07-18
fix: CXSPA-6950 Deprecate debug flag in OptimizedSsrOptions

pawelfras:
This PR contains changes in the `debug` flag which is part of `OptimizedSsrOptions`. Due to a request from the Support Team, we enabled all logs produced by the `OptimizedSsrEngine`, regardless of the `debug` flag value. The flag itself has been deprecated.

In addition, we deprecated the protected `log` method from `OptimizedSsrEngine` which depended on the `debug` flag. We want to make it private in the future.

QA steps: 
- in the `server.ts` file, add `maxRenderTIme` with a small value to `SsrOptimizationOptions`:
  ```
  const ssrOptions: SsrOptimizationOptions = {
    timeout: Number(
      process.env['SSR_TIMEOUT'] ?? defaultSsrOptimizationOptions.timeout
    ),
    maxRenderTime: 100,
  };
  ```
- run SSR in dev mode
  `npm run dev:ssr`
- open app in the browser and verify if server logs contain message:
   ```
   {
    "message": "Rendering of /electronics-spa/en/USD completed after the specified maxRenderTime, therefore it was ignored.",
    /* ... */
  }
   ```
- in the `server.ts` file, add `maxRenderTIme` with a small value to `SsrOptimizationOptions`:
  ```
  const ssrOptions: SsrOptimizationOptions = {
    timeout: Number(
      process.env['SSR_TIMEOUT'] ?? defaultSsrOptimizationOptions.timeout
    ),
    debug: false
    maxRenderTime: 100,
  };
  ```
- reload the page and verify that the log is sill present, which means `debug` flag was not taken into consideration

closes [CXSPA-6950](https://jira.tools.sap/browse/CXSPA-6950)

### DIFF
--- a/core-libs/setup/ssr/optimized-engine/optimized-ssr-engine.ts
+++ b/core-libs/setup/ssr/optimized-engine/optimized-ssr-engine.ts
@@ -318,14 +318,15 @@ export class OptimizedSsrEngine {
     });
   }
 
+  /**
+   * @deprecated This method will be private in the future.
+   */
   protected log(
     message: string,
-    debug = true,
+    _ignoredLegacyParameter = true,
     context: ExpressServerLoggerContext
   ): void {
-    if (debug || this.ssrOptions?.debug) {
-      this.logger.log(message, context || {});
-    }
+    this.logger.log(message, context || {});
   }
 
   /** Retrieve the document from the cache or the filesystem */
--- a/core-libs/setup/ssr/optimized-engine/optimized-ssr-engine.ts
+++ b/core-libs/setup/ssr/optimized-engine/optimized-ssr-engine.ts
@@ -319,7 +319,7 @@ export class OptimizedSsrEngine {
   }
 
   /**
-   * @deprecated since v2211.26 - This method will be private in the future.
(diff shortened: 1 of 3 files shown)


core-libs/setup/ssr/optimized-engine/optimized-ssr-engine.ts
```diff
@@ -318,14 +318,15 @@ export class OptimizedSsrEngine {
     });
   }
 
+  /**
+   * @deprecated This method will be private in the future.
```
```suggestion
   * @deprecated since v2211.26 - This method will be private in the future.
```

core-libs/setup/ssr/optimized-engine/optimized-ssr-engine.ts
```diff
@@ -318,14 +318,15 @@ export class OptimizedSsrEngine {
...
+   */
   protected log(
     message: string,
-    debug = true,
+    _ignoredLegacyParameter = true,
```
```suggestion
    _ignoredLegacyDebugParameter = true,
```

Platonn (dismissed):

Platonn (dismissed):

rmch91 (approved):

PR #19067, opened 2024-07-23
feat: Configurable media component

rmch91:
Closes: https://jira.tools.sap/browse/CXSPA-7700
QA[ steps](https://github.com/SAP/spartacus/pull/19067#issuecomment-2415949963) for new application
QA steps for migrated app:
It's worth to perform similar QA steps for migrated app. The following is how to prepare migrated app:

checkout 6.8.0 in Spartacus repo
build and deploy Spartacus packages locally:

- npx ts-node ./tools/schematics/testing.ts in the SPA root folder`

- create new app npx @angular/cli@15 new my-app --style=scss --routing=false

- npx @angular/cli@15 add @spartacus/schematics@latest --baseUrl="https://40.76.109.9:9002"

- checkout feat/CXSPA-7700 in Spartacus repom build and deploy packages with changes

- follow steps from 2211-migration.md to migrate app

perform QA

The idea behind this PR is to provide flexibility for customers, so they can decide whether they want to use the <picture> element or the  element, and to provide a separate config for the <picture> element that will allow them to add specific media queries.

Crucial parts:

1. New input in media component,  `@Input() elementType: 'img' | 'picture' = 'img';`
2.  New input in media component, width `@Input() width: number;`
3. New input in media component, height `@Input() height: number;`
4.  New input in media component, sizes `@Input() sizes: string;`
5. `pictureElementFormats` in media config
6. `pictureFormatsOrder`  in media config
7. `getMediaForPictureElement()` in MediaService
8. `resolveSources()` in MediaService
9. `useExtendedMediaComponentConfiguration?: boolean;` feature toggle
12. deprecation of `useLegacyMediaComponent` config property and `USE_LEGACY_MEDIA_COMPONENT` injection token

**### Important:** 
After activation `useExtendedMediaComponentConfiguration` toggle default HTML element in MediaComponent will be ``
Only BannerComponent has passed `'picture'` value. If you need to use `<picture>` HTML element
you need to pass `[elementType]="'picture'"` to `<cx-media>`

The idea is to move away from the concept that `` is legacy, because from my research, I understand that both elements make sense in different situations. The `<picture>` element is the only way to go in situations when we what to handle different art directions or more complex logic, more complex media queries. That's why I want to provide flexibility, so users can decide where they need `<picture>` element.
In our application I think we could enable it for banner component, since it has different formats, and I see a possibility here to add in future additional formats to handle retina screens and to improve the visibility of this banners on retina mobile/desktop devices.

Current default-media.config is more for representation of the approach, and should be reduced to only formats that we have and queries that we use before merge.


projects/core/src/features-config/feature-toggles/config/feature-toggles.ts
```diff
@@ -397,6 +397,8 @@ export interface FeatureTogglesInterface {
    * customization buttons in the BottomHeaderSlot in SmartEdit.
    */
   cmsBottomHeaderSlotUsingFlexStyles?: boolean;
+
+  useAdvancedMediaComponent?: boolean;
```
rmch91: Better name? + description

projects/storefrontlib/shared/components/media/media.model.ts
```diff
@@ -29,6 +29,13 @@ export interface Media {
    * readers or magnifiers
    */
   role?: string;
+
+  sources?: any[];
+}
+
+export interface PictureElementAttributes {
```
rmch91: + description. Represents attributes valid for Picture element

projects/storefrontlib/shared/components/media/media.model.ts
```diff
@@ -51,6 +58,49 @@ export interface MediaFormatSize {
   width?: number;
 }
 
+/**
+ * Specifies media queries that can be used to generate information for the
+ * browser to resolve the right media for the right layout or device.
+ */
+export interface PictureElementQueries {
```
rmch91: Not sure if it should exist. Maybe it should be 
```
export interface PictureElementQueries {
  [query: string]: string | number;
}
```
In this case user could add additional any queries by adding it to config and by adding proper mapping in config. I think this should be the better way.

projects/storefrontlib/shared/components/media/media.model.ts
rmch91: Removed, i think it is not needed since used only in media.service

projects/storefrontlib/shared/components/media/media.model.ts
rmch91: In current implementation, customers allowed to add to config only queries defined in this interface. But `generateMediaQuery` method from `MediaService` can handle everything that will be provided in config, if only properties will also properly mapped. I see the flexibility here, so users could add any queries they need and not be limited by interface.

projects/core/src/features-config/feature-toggles/config/feature-toggles.ts

projects/storefrontlib/shared/components/media/media.model.ts
rmch91: Removed properties from this interface. Changed to `[query: string]: string | number;` for flexibility.

pawelfras: What about inverting the dependency and using here only queries defined within `mediaQueryMap`. `mediaQueryMap` itself could be a combination of predefined properties with the generic type that allows customers to provide new props to the map. Thanks to that customers will have an intellisence when defining queries.

```
export type DefaultMediaQueryKeys =
  | 'minWidth'
  | 'maxWidth'
  | 'minHeight'
  | 'maxHeight'
  | 'minDevicePixelRatio'
  | 'maxDevicePixelRatio'
  | 'orientation'
  | 'minAspectRatio'
  | 'maxAspectRatio'
  | 'minResolution'
  | 'maxResolution';

export type MediaQueryKeys = DefaultMediaQueryKeys & string;

export interface PictureMediaQuery = {
      [key in MediaQueryKeys]?: string | number;
}
```
in `MediaConfig`: 

```
pictureElementFormats?: {
    [format: string]: PictureMediaQuery
};

mediaQueryMap?: { [key in MediaQueryKeys]?: string };
```

projects/storefrontlib/shared/components/media/media.component.html
```diff
@@ -1,36 +1,77 @@
 <ng-container *ngIf="!!media">
-  <picture *ngIf="media.srcset && !isLegacy; else legacyImgTmp">
-    <source
-      *ngFor="
-        let source of media!.srcset! | cxMediaSources;
-        trackBy: trackByMedia
+  <ng-container *cxFeature="'useMediaComponentWithConfigurableMediaQueries'">
```
pawelfras: If the toggle `useMediaComponentWithConfigurableMediaQueries` controls the template and with the active toggle by default we introduce the use of `` tag and `<picture>` will be used only when we explicitely select it via input `usePictureElement`, we will introduce regression bugs which have been fixed within CXSPA-4114. Due to the lack of `sizes` attribute, the width descriptor in `srcset` will not be taken into account, and consequently, `srcset` used together with `img` tag will have no effect and images will not be responsive (see: https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/srcset)

We need to decide whether we want to include support for `sizes` (and maybe density descriptor) in the scope of this PR or as a separate task - I believe it requires more research.

If not, the introduced feature toggle should not control the method used to retrieve media data - whether from extended media queries or not. The template could stay as it was, so no ``, only `<picture>` element.

pawelfras: It looks that In the case of the `` tag, the `sizes` attribute, if not provided, has a fallback to `100vw`. Apart from that, it depends on the browser how it is interpreted:
- Firefox picks the proper image size on change
- Chrome picks the proper image for the state of the viewport on the refresh, then if an image with a higher resolution is loaded, it will be used even if we decrease the screen width

rmch91:
### Feature flags and config dependencies testing:

| #  | `useImgElement` | `useMediaComponentWithConfigurableMediaQueries` | `useLegacyMediaComponent` | Covered? |
|----|----------------|-----------------------------------------------|----------------------------|----------|
| 1  | **true**        | **false**                                    | **false**           | 👍        |
| 2  | **true**        | **true**                                     | **false**           | 👍        |
| 3  | **false**       | **false**                                    | **false**           | 👍        |
| 4  | **false**       | **true**                                     | **false**           | 👍        |
| 5  | **true**        | **false**                                    | **true**                    | 👍        |
| 6  | **true**        | **false**                                    | **false**                   | 👍        |
| 7  | **true**        | **true**                                     | **true**                    | 👍        |
| 8  | **false**       | **false**                                    | **true**                    | 👍        |

projects/storefrontlib/shared/components/media/media.component.html
```diff
@@ -57,6 +62,9 @@
         [attr.role]="media.role"
         (load)="loadHandler()"
         (error)="errorHandler()"
+        [attr.width]="width"
+        [attr.height]="height"
+        [attr.sizes]="sizes"
```
pawelfras: `major:` As this change might be breaking, I think we should not introduce these attributes to `picture` nor `img` for the scenario where `useMediaComponentWithConfigurableMediaQueries` is disabled. Let's keep the default behavior here.

projects/storefrontlib/shared/components/media/media.component.html
```diff
@@ -71,6 +79,9 @@
...
```
pawelfras: `major:` As this change might be breaking, I think we should not introduce these attributes to `picture` nor `img` for the scenario where `useMediaComponentWithConfigurableMediaQueries` is disabled. Let's keep the default behavior here.

projects/storefrontlib/shared/components/media/media.component.ts
```diff
@@ -61,6 +66,30 @@ export class MediaComponent implements OnChanges {
    */
   @Input() loading: ImageLoadingStrategy | null = this.loadingStrategy;
 
+  @Input() useImgElement: boolean = false;
```
pawelfras:
```suggestion
  @Input() elementType: 'picture' | 'img' = 'picture'
```
`query:` What do you think about changing the type of this input from boolean to enum? Seems to be a better choice semantic-wise and such an approach allows us to adjust the type in case new elements for handling images appear in the future.
TBH. The best solution for similar cases is content projection, but I'm not sure whether it will help here.

projects/storefrontlib/shared/components/media/media.component.ts
```diff
@@ -95,11 +124,20 @@ export class MediaComponent implements OnChanges {
   protected trackByMedia: TrackByFunction<HTMLSourceElement> = (_, item) =>
     item.media;
 
+  /**
+   * @deprecated will be removed
+   *
```
pawelfras:
```suggestion
  /**
   * @deprecated since 2211.30. It will be eventually removed in the future.
   *
```

projects/storefrontlib/shared/components/media/media.component.ts
```diff
@@ -95,11 +124,20 @@ export class MediaComponent implements OnChanges {
...
+   * To use `img` HTML element instead of `picture`
+   * use `useMediaComponentWithConfigurableMediaQueries` feature flag
+   * and pass `[useImgElement]="true"` input to the component
+   */
   protected isLegacy =
     inject(USE_LEGACY_MEDIA_COMPONENT, { optional: true }) ||
     (inject(Config) as any)['useLegacyMediaComponent'] ||
     false;
 
+  protected readonly featureConfigService = inject(FeatureConfigService);
```
pawelfras: `major:` let's change the accesses modifier of `FeatureConfigService` to `private` to remove it easily in the future. For more, see: https://wiki.one.int.sap/wiki/display/spar/Coding+Guidelines#CodingGuidelines-UsingFeatureTogglesinTypeScript

If the logic from the `create` method is moved to `MediaService`, there will then not even be a need to inject this service in the component.

projects/storefrontlib/shared/components/media/media.token.ts
```diff
@@ -6,6 +6,9 @@
 
 import { InjectionToken } from '@angular/core';
 
+/**
+ * @deprecated will be removed
+ */
```
pawelfras:
```suggestion
/**
 * @deprecated since 2211.30 and will be removed
 */
```

projects/storefrontlib/shared/components/media/media.config.ts
```diff
@@ -42,6 +94,7 @@ export abstract class MediaConfig {
   imageLoadingStrategy?: ImageLoadingStrategy;
 
   /**
+   * @deprecated will be removed
```
pawelfras:
```suggestion
   * @deprecated since 2211.30 and will be eventually removed in the future
```

projects/storefrontlib/shared/components/media/media.component.ts
```diff
@@ -110,12 +148,22 @@ export class MediaComponent implements OnChanges {
...
   protected create(): void {
-    this.media = this.mediaService.getMedia(
+    const shouldGetMediaForPictureElement =
+      this.featureConfigService.isEnabled(
+        'useMediaComponentWithConfigurableMediaQueries'
+      ) && !this.useImgElement;
+
+    const getMedia = shouldGetMediaForPictureElement
+      ? this.mediaService.getMediaForPictureElement.bind(this.mediaService)
+      : this.mediaService.getMedia.bind(this.mediaService);
```
pawelfras: `medium:` As we have `MediaService`, let's move this logic there:
- the component doesn't have to decide how to get media
- no need to bind `mediaService`
- component is cleaner (it doesn't need to have notion about the feature toggle)

projects/core/src/features-config/feature-toggles/config/feature-toggles.ts
```diff
@@ -501,6 +501,13 @@ export interface FeatureTogglesInterface {
    * customization buttons in the BottomHeaderSlot in SmartEdit.
    */
   cmsBottomHeaderSlotUsingFlexStyles?: boolean;
+
+  /**
+   * When enabled, allows to provide extended formats and media queries for <picture> element.
+   *
+   * For proper work requires `pictureElementFormats` and `mediaQueryMap` provided in media config.
+   */
```
pawelfras:
```suggestion
  /**
   * When enabled, allows to provide extended formats and media queries for <picture> element if used in MediaComponent.
   *
   * For proper work requires `pictureElementFormats` and `mediaQueryMap` provided in media config:
   * provideConfig({
   *   pictureElementFormats: {
   *     /*...*/
   *   },
   *   mediaQueryMap: {
   *     /* ... */
   *   },
   * }),
   */
```

projects/storefrontlib/shared/components/media/media.component.spec.ts
```diff
@@ -1,13 +1,50 @@
...
+import { InjectionToken } from '@angular/core';
+import { FeatureConfigService } from '@spartacus/core';
 import { MediaComponent } from './media.component';
 import { ImageLoadingStrategy, Media } from './media.model';
 import { MediaService } from './media.service';
 import { USE_LEGACY_MEDIA_COMPONENT } from './media.token';
 
+export const IS_CONFIGURABLE_MEDIA_COMPONENT = new InjectionToken<boolean>(
+  'IS_CONFIGURABLE_MEDIA_COMPONENT'
+);
```
pawelfras: `query:` is this token required? I can see its value is set once and never change from `false`.

projects/storefrontlib/shared/components/media/media.config.ts
```diff
@@ -29,6 +33,54 @@ export abstract class MediaConfig {
...
+   * - `maxWidth` -> `max-width`
+   * - `minHeight` -> `min-height`
+   * - `maxHeight` -> `max-height`
+   * - `minDevicePixelRatio` -> `min-device-pixel-ratio`
+   * - `maxDevicePixelRatio` -> `max-device-pixel-ratio`
+   * - `orientation` -> `orientation`
+   * - `minAspectRatio` -> `min-aspect-ratio`
+   * - `maxAspectRatio` -> `max-aspect-ratio`
+   */
+  mediaQueryMap?: Record<keyof PictureElementQueries, string>;
```
pawelfras: `major:` After secound thougt about `mediaQueryMap`, I'd consider removing it in favour of writing CSS media query features names directly in `pictureElementFormats`. Why:
- less logic in our `MediaService` because mapping is no logner required
- customers don't need to follow our opinionated syntax and can use CSS standards directly in the config

example
```ts
pictureElementFormats: {
    mobile: {
      mediaQueries: {
        'max-width': '767px',
      },
      width: 50,
      height: 50,
    },
    tablet: {
      mediaQueries: { 'min-width': '768px', 'max-width': '1024px' },
    },
}
```

projects/storefrontlib/shared/components/media/media.model.ts
```diff
@@ -51,6 +66,21 @@ export interface MediaFormatSize {
   width?: number;
 }
 
+/**
+ * Specifies media queries that can be used to generate information for the
+ * browser to resolve the right media for the right layout or device.
+ */
+export type PictureElementQueries = {
```
pawelfras: `medium:` let's put a link to official page with all supported queries so customers have it in one place

projects/storefrontlib/shared/components/media/media.component.spec.ts
rmch91: Yes, it is used in function configureTestingModule(). The only thing - it shouldn't be exported, since it is used only in tests to configure testing module on the fly.

rmch91: Actually it changed i different scenarios to false `configureTestingModule(new MockMediaService('srcset'), false, true);`

projects/storefrontlib/shared/components/media/media.component.html
rmch91: Changed

projects/storefrontlib/shared/components/media/media.component.html
rmch91: Changed

projects/storefrontlib/shared/components/media/media.component.ts
rmch91: Make sens, changed

projects/storefrontlib/shared/components/media/media.component.ts
rmch91: Changed

projects/storefrontlib/shared/components/media/media.component.ts
rmch91: Changed

projects/storefrontlib/shared/components/media/media.token.ts
rmch91: Changed

projects/storefrontlib/shared/components/media/media.config.ts
rmch91: Changed

projects/core/src/features-config/feature-toggles/config/feature-toggles.ts

projects/storefrontlib/shared/components/media/media.config.ts
rmch91: Changed

projects/storefrontlib/shared/components/media/media.component.ts
rmch91: Changed

projects/storefrontlib/shared/components/media/media.model.ts
rmch91: Changed

projects/storefrontlib/shared/components/media/media.component.spec.ts
pawelfras: Only now I see it's used in `configureTestingModule` function which is used in tests to create specific env. Apologies!

projects/storefrontlib/shared/components/media/media.config.ts
```diff
@@ -29,6 +33,33 @@ export abstract class MediaConfig {
...
+   * Picture element configuration holds the media queries assigned to
+   * a format.
+   * The order of formats matters.
+   * <source> elements in <picture> will be sorted based on this order.
+   * This is necessary because the browser evaluates each
+   * <source> element in order and uses the first one that matches.
+   */
+  pictureElementFormats?: {
+    [format: string]: {
+      mediaQueries: PictureElementQueries;
```
pawelfras: `query:` shouldn't this propert be optional as well? I'm thinking about a scenario where the user wants to leave the default mediaQueries for a given format, but in the config would like to additionally pass `width` and `height`

rmch91:
QA steps:

1. Checkout feat/CXSPA-7700
2. `ts-node ./tools/schematics/testing.ts` -> Publish
3. Create new angular app `npx -p @angular/cli@17.0.5 ng new test-new-media-cmp --standalone=false`
4. ng add @spartacus/schematics@latest
5. Create custom banner component 
```
@Component({
  selector: 'cx-banner',
  templateUrl: './custom-banner.component.html',
})
export class CustomBannerComponent extends BannerComponent {}
```
```
@NgModule({
  imports: [MediaModule, CommonModule, FormsModule],
  declarations: [CustomBannerComponent],
  exports: [CustomBannerComponent],
})
export class CustomBannerModule {}
```
6. Override banner component to use custom banner component
```
    provideConfig(<CmsConfig>{
      cmsComponents: {
        SimpleResponsiveBannerComponent: {
          component: CustomBannerComponent,
        },
      },
    }),
```
7. Play with custom-banner template by passing different inputs to media component, checking if it reflects on the UI
Example:
```
<ng-container *ngIf="data$ | async as data">
  <cx-media
    [container]="getImage(data)"
    [elementType]="'img'"
    [width]="1000"
    [height]="500"
    [sizes]="'(max-width: 600px) 100vw, (max-width: 1200px) 50vw, 33vw'"
  ></cx-media>
</ng-container>
```
8. Play with media config by adding different media queries, width or height:
```
    provideConfig(<MediaConfig>{
      pictureElementFormats: {
        mobile: {
          mediaQueries: {
            'max-width': '767px',
          },
        },
        desktop: {
          mediaQueries: { 'min-width': '1025px', 'max-width': '1400px' },
          width: 1600,
        },
        widescreen: {
          mediaQueries: {
            'min-width': '1401px',
            '-webkit-min-device-pixel-ratio': 3,
          },
          width: 2200,
          height: 1200,
        },

        retina_desktop: {
          mediaQueries: {
            'min-width': '1440px',
            '-webkit-min-device-pixel-ratio': 2,
          },
        },
      },
    }),
```
9. Add order to config  `pictureFormatsOrder: ['mobile', 'widescreen'],`. Check if order reflects properly in HTML output

pawelfras:
`minor:` I think we can remove usage of `USE_LEGACY_MEDIA_COMPONENT` from `app.module.ts`:
https://github.com/SAP/spartacus/blob/df1fa48a90ba00d7f5361757cd6a403403c60ab1/projects/storefrontapp/src/app/app.module.ts#L107

projects/storefrontlib/shared/components/media/media.config.ts
rmch91: Yeah, You have right, changed

projects/core/src/features-config/feature-toggles/config/feature-toggles.ts
```diff
@@ -626,6 +626,25 @@ export interface FeatureTogglesInterface {
...
+   *     'max-width': '767px',
+   *      ...
+   *    },
+   *    width: 50,
+   *    height: 50,
+   *   },
+   * })
+   * ```
+   */
+  useMediaComponentWithConfigurableMediaQueries?: boolean;
```
pawelfras: `major:` After secound thoughts on this topic, I think we should consider changing the name of this feature toggle as it is responsible not only for support of new config, but also controls whether new `MediaComponent` inputs: `width`, `height` and `sizez` are taken into account.
What do you think about `useExtendedMediaComponent`?
The JSDoc should be adjusted too.

rmch91: Done

projects/storefrontlib/shared/components/media/media.service.ts
```diff
@@ -53,6 +76,67 @@ export class MediaService {
...
+
+    if (!commonMediaProperties) {
+      return;
+    }
+
+    return {
+      ...commonMediaProperties,
+      sources: this.resolveSources(mediaContainer, format),
+    };
+  }
```
pawelfras:
```suggestion
  getMediaForPictureElement(
    mediaContainer?: MediaContainer | Image,
    format?: string,
    alt?: string,
    role?: string
  ): Media {
    if (!mediaContainer) {
      return;
    }

    const commonMediaProperties = this.getCommonMediaObject(
      mediaContainer,
      format,
      alt,
      role
    );

    return {
      ...commonMediaProperties,
      sources: this.resolveSources(mediaContainer, format),
    };
  }
```

projects/core/src/features-config/feature-toggles/config/feature-toggles.ts
```diff
@@ -626,6 +626,28 @@ export interface FeatureTogglesInterface {
...
+   *      ...
+   *    },
+   *    width: 50,
+   *    height: 50,
+   *   },
+   * })
+   * ```
+   *
+   * After activating this toggle, new inputs in `MediaComponent` — specifically
+   * `width`, `height`, and `size` — will be passed to the template as HTML attributes.
```
pawelfras:
```suggestion
   * `width`, `height`, and `sizes` — will be passed to the template as HTML attributes.
```

projects/storefrontlib/shared/components/media/media.service.ts
```diff
@@ -53,6 +76,67 @@ export class MediaService {
...
+   */
+  protected getCommonMediaObject(
+    mediaContainer?: MediaContainer | Image,
+    format?: string,
+    alt?: string,
+    role?: string
+  ): Media | undefined {
+    if (!mediaContainer) {
+      return;
+    }
```
pawelfras:
```suggestion
  protected getCommonMediaObject(
    mediaContainer: MediaContainer | Image,
    format?: string,
    alt?: string,
    role?: string
  ): Media {
```
No need to check again if `mediaContainer` is defined as it was verified in `getMedia` and `getMediaForPictureElement`

projects/storefrontlib/shared/components/media/media.service.ts
```diff
@@ -53,6 +76,67 @@ export class MediaService {
...
+    const commonMediaProperties = this.getCommonMediaObject(
+      mediaContainer,
+      format,
+      alt,
+      role
+    );
+
+    if (!commonMediaProperties) {
+      return;
+    }
```
pawelfras: no longer needed after `getCommonMediaObject` adjustments

projects/storefrontlib/shared/components/media/media.component.ts
```diff
@@ -61,6 +61,36 @@ export class MediaComponent implements OnChanges {
    */
   @Input() loading: ImageLoadingStrategy | null = this.loadingStrategy;
 
+  @Input() elementType: 'img' | 'picture' = 'picture';
```
pawelfras: `major:` Only now I realised this input also relates on the new feature flag. Let's add a descriptive JSDoc for it as well as mention this prop in docs for the feature toggle.

pawelfras:
> QA steps:
> 
> 1. Checkout feat/CXSPA-7700
> 2. `ts-node ./tools/schematics/testing.ts` -> Publish
> 3. Create new angular app `npx -p @angular/cli@17.0.5 ng new test-new-media-cmp --standalone=false`
> 4. ng add @spartacus/schematics@latest
> 5. Create custom banner component
> 
> ```
> @Component({
>   selector: 'cx-banner',
>   templateUrl: './custom-banner.component.html',
>   changeDetection: ChangeDetectionStrategy.OnPush,
> })
> export class CustomBannerComponent extends BannerComponent {}
> ```
> 
> ```
> @NgModule({
>   imports: [MediaModule, CommonModule, FormsModule],
>   declarations: [CustomBannerComponent],
>   exports: [CustomBannerComponent],
> })
> export class CustomBannerModule {}
> ```
> 
> 6. Override banner component to use custom banner component
> 
> ```
>     provideConfig(<CmsConfig>{
>       cmsComponents: {
>         SimpleResponsiveBannerComponent: {
>           component: CustomBannerComponent,
>         },
>       },
>     }),
> ```
> 
> 7. Play with custom-banner template by passing different inputs to media component, checking if it reflects on the UI
>    Example:
> 
> ```
> <ng-container *ngIf="data$ | async as data">
>   <cx-media
>     [container]="getImage(data)"
>     [elementType]="'img'"
>     [width]="1000"
>     [height]="500"
>     [sizes]="'(max-width: 600px) 100vw, (max-width: 1200px) 50vw, 33vw'"
>   ></cx-media>
> </ng-container>
> ```
> 
> 8. Play with media config by adding different media queries, width or height:
> 
> ```
>     provideConfig(<MediaConfig>{
>       pictureElementFormats: {
>         mobile: {
>           mediaQueries: {
>             'max-width': '767px',
>           },
>         },
>         desktop: {
>           mediaQueries: { 'min-width': '1025px', 'max-width': '1400px' },
>           width: 1600,
>         },
>         widescreen: {
>           mediaQueries: {
>             'min-width': '1401px',
>             '-webkit-min-device-pixel-ratio': 3,
>           },
>           width: 2200,
>           height: 1200,
>         },
> 
>         retina_desktop: {
>           mediaQueries: {
>             'min-width': '1440px',
>             '-webkit-min-device-pixel-ratio': 2,
>           },
>         },
>       },
>     }),
> ```
> 
> 9. Add order to config  `pictureFormatsOrder: ['mobile', 'widescreen'],`. Check if order reflects properly in HTML output

It's worth to perform similar QA steps for migrated app. The following is how to prepare migrated app:
- checkout `6.8.0` in Spartacus repo
- create new app `npx @angular/cli@15 new my-app --style=scss --routing=false`
  `npx @angular/cli@15 add @spartacus/schematics@latest --baseUrl="https://40.76.109.9:9002"`
- checkout `feat/CXSPA-7700` in Spartacus repom build and deploy packages with changes
- follow steps from `2211-migration.md` to migrate app
- perform QA

projects/core/src/features-config/feature-toggles/config/feature-toggles.ts
```diff
@@ -626,6 +626,28 @@ export interface FeatureTogglesInterface {
...
+   */
+  useExtendedMediaComponent?: boolean;
```
pawelfras: `nitpick:` What do you think about `useExtendedMediaComponentConfig` or `useExtendedMediaComponentConfiguration` - now as I look at the current name, it suggests what you mentioned offline - that there is some `MediaComponentV2`

pawelfras:
I additionally ran the Spartacus build pipeline step on this branch to run this change against all E2E tests. Here's the related pipeline run: https://github.com/SAP/spartacus/actions/runs/11361460758

projects/core/src/features-config/feature-toggles/config/feature-toggles.ts

projects/storefrontlib/shared/components/media/media.component.ts
rmch91: Changed

projects/storefrontlib/shared/components/media/media.service.ts
rmch91: Removed

projects/storefrontlib/shared/components/media/media.service.ts
rmch91: Removed

projects/core/src/features-config/feature-toggles/config/feature-toggles.ts

projects/storefrontlib/shared/components/media/media.service.ts
rmch91: Changed

feature-libs/product-multi-dimensional/selector/components/selector/product-multi-dimensional-selector.component.html
```diff
@@ -21,7 +21,17 @@
...
+                *cxFeature="'useExtendedMediaComponentConfiguration'"
+                [container]="option.image"
+                role="img"
+                [elementType]="'img'"
+              />
+              <cx-media
+                *cxFeature="'!useExtendedMediaComponentConfiguration'"
+                [container]="option.image"
+                role="img"
+              />
```
pawelfras: `query:` this usage of `cx-media` doesn't contain `format` property. Is the feature toggle wrapper required?

feature-libs/product/image-zoom/components/product-image-zoom/product-image-zoom-product-images/product-image-zoom-product-images.component.html
```diff
@@ -1,5 +1,15 @@
...
+    *cxFeature="'useExtendedMediaComponentConfiguration'"
+    [container]="main"
+    (click)="triggerZoom(true)"
+    [elementType]="'img'"
+  ></cx-media>
+  <cx-media
+    *cxFeature="'!useExtendedMediaComponentConfiguration'"
+    [container]="main"
+    (click)="triggerZoom(true)"
+  ></cx-media>
```
pawelfras: query: this usage of cx-media doesn't contain format property. Is the feature toggle wrapper required?

feature-libs/product/image-zoom/components/product-image-zoom/product-image-zoom-product-images/product-image-zoom-product-images.component.spec.ts
pawelfras, line 1: `nitpick:` This change is not related to the PR and could be reverted.

feature-libs/product/image-zoom/components/product-image-zoom/product-image-zoom-thumbnails/product-image-zoom-thumbnails.component.html
```diff
@@ -10,21 +10,45 @@
...
+    </cx-media>
+    <cx-media
+      *cxFeature="'!a11yCarouselArrowKeysNavigation'"
+      [container]="item.container"
+      tabindex="0"
+      (focus)="openImage(item.container)"
+      [class.is-active]="isActive(item.container) | async"
+    >
+    </cx-media>
+  </ng-container>
```
pawelfras: query: this usage of cx-media doesn't contain format property. Is the feature toggle wrapper required?

feature-libs/product/image-zoom/components/product-image-zoom/product-image-zoom-view/product-image-zoom-view.component.html
```diff
@@ -11,16 +11,43 @@
...
+        <cx-media
+          *ngIf="!isZoomed"
+          [elementType]="'img'"
+          #defaultImage
+          class="cx-default-image-zoom"
+          (loaded)="onImageLoad()"
+          [container]="main"
+        >
+        </cx-media>
+      </ng-container>
```
pawelfras: query: this usage of cx-media doesn't contain format property. Is the feature toggle wrapper required?

projects/core/src/features-config/feature-toggles/config/feature-toggles.ts
```diff
@@ -638,6 +638,30 @@ export interface FeatureTogglesInterface {
...
+   *    height: 50,
+   *   },
+   * })
+   * ```
+   *
+   * After activating this toggle, new inputs in `MediaComponent` — specifically
+   * `width`, `height`, and `sizes` — will be passed to the template as HTML attributes.
+   *
+   * Toggle activates `@Input() elementType: 'img' | 'picture' = 'picture';` in `MediaComponent`
+   */
```
pawelfras:
```suggestion

  /**
   * When enabled, allows to provide extended formats and media queries for <picture> element if used in MediaComponent.
   *
   * For proper work requires `pictureElementFormats`  provided in media config:
   *  ```ts
   * provideConfig({
   *   pictureElementFormats: {
   *    mediaQueries: {
   *     'max-width': '767px',
   *      ...
   *    },
   *    width: 50,
   *    height: 50,
   *   },
   * })
   * ```
   *
   * After activating this toggle, new inputs in `MediaComponent` — specifically
   * `width`, `height`, and `sizes` — will be passed to the template as HTML attributes.
   *
   * Toggle activates `@Input() elementType: 'img' | 'picture' = 'picture';` in `MediaComponent`
   *
   * Toggle sets `elemetType` to `img` for all `MediaComponent` usages, where one of the following format is set: `cartIcon`, `thumbnail`, `product`, `zoom`
   */
```

projects/storefrontlib/cms-components/product/carousel/product-carousel/product-carousel.component.spec.ts
pawelfras, line 1: `nitpick:` not related change, could be reverted

projects/storefrontlib/cms-components/product/product-images/product-images.component.html
```diff
@@ -1,5 +1,13 @@
...
-  <cx-media [container]="main"></cx-media>
+  <cx-media
+    *cxFeature="'useExtendedMediaComponentConfiguration'"
+    [elementType]="'img'"
+    [container]="main"
+  ></cx-media>
+  <cx-media
+    *cxFeature="'!useExtendedMediaComponentConfiguration'"
+    [container]="main"
+  ></cx-media>
```
pawelfras: `query:` this usage of cx-media doesn't contain format property. Is the feature toggle wrapper required?

projects/storefrontlib/shared/components/media/media.component.ts
```diff
@@ -61,6 +61,41 @@ export class MediaComponent implements OnChanges {
    */
   @Input() loading: ImageLoadingStrategy | null = this.loadingStrategy;
 
+  /**
+   * Works only when `useExtendedMediaComponentConfiguration` toggle is true
+   *
+   * @default img
```
pawelfras, line 67: :nice:

pawelfras (approved):
LGTM

PR #19071, opened 2024-07-24
fix: (CXSPA-979) - Dialogs announcement JAWS fix

Pio-Bar:
Ticket: [CXSPA-979](https://jira.tools.sap/browse/CXSPA-979)

Fixes a JAWS specific bug where the title page was announced while dialog was opened. By providing a label to the dialog container we hear it instead.


Zeyber (dismissed):

feature-libs/cart/saved-cart/components/saved-cart-form-dialog/saved-cart-form-dialog.component.html
```diff
@@ -4,12 +4,13 @@
     (esc)="close('Escape clicked')"
     class="cx-saved-cart-form-dialog"
     role="dialog"
+    aria-labbeledby="dialogTitle"
```
developpeurweb: small typo, it should read `aria-labelledby`

feature-libs/quote/components/confirm-dialog/quote-confirm-dialog.component.html
```diff
@@ -5,14 +5,15 @@
   (esc)="dismissModal('Escape clicked')"
   role="dialog"
   aria-modal="true"
+  aria-labelldby="dialogTitle"
```
developpeurweb: small typo, it should read `aria-labelledby`

developpeurweb (changes requested):
Small adjustments please:
1. Small typos were preventing the dialog from being identified.
2. For some reason, the "Add To Cart" dialog doesn't get identified as a dialog, it seems the focus lands one Tab away from the close button still trapped and narrating the page underneath. Once moving focus to the close button, the dialog gets identified. The ideal would be for the focus to move to the close button when opening, as with the other dialogs.

Pio-Bar:
The 'added to cart' dialog should no longer lose focus after the cart is updated and stay on the close button. Typos fixed, good catch ;)

rmch91 (approved):

developpeurweb (approved):

PR #19074, opened 2024-07-25
fix: (CXSPA-942) - Provides titles for all buttons with icons only

Pio-Bar:
Ticket: [CXSPA-942](https://jira.tools.sap/browse/CXSPA-942)

Adds titles to interactive elements with an icon but no text. This includes all 'close' buttons, 'edit' buttons etc.
The default tooltip should show on mouseover.

### DIFF
--- a/feature-libs/asm/components/customer-list/customer-list.component.html
+++ b/feature-libs/asm/components/customer-list/customer-list.component.html
@@ -471,6 +471,7 @@
     class="close"
     attr.aria-label="{{ 'common.close' | cxTranslate }}"
     (click)="closeModal('Cross click')"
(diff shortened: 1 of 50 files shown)


Zeyber (dismissed):

developpeurweb:
@Pio-Bar is it possible to add a "clear filter" tooltip to these elements?
clear filter

Pio-Bar:
> @Pio-Bar is it possible to add a "clear filter" tooltip to these elements? 

I have added the attribute and kept it consistent with the aria-label. I have also found two additional places that could use a title.

developpeurweb (approved):

PR #19077, opened 2024-07-25
fix: update delivery mode text to delivery options

RadhepS:
closes: https://jira.tools.sap/browse/CXSPA-7752


feature-libs/checkout/base/assets/translations/en/checkout.json
```diff
@@ -23,7 +23,7 @@
     "deliveryAddressSelected": "Delivery address selected"
   },
   "checkoutMode": {
-    "deliveryMethod": "Delivery Method",
+    "deliveryMethod": "Delivery Options",
```
kimhw0630: Changing or removing translation keys or values is breaking change
to avoid breaking change add new translation key

feature-libs/pickup-in-store/assets/translations/en/pickupInStore.json
```diff
@@ -56,7 +56,7 @@
     "pickUpInStoreAddress": "Pick Up Store Address",
     "itemsToBeShipped": "Items To Be Shipped",
     "shippingAddress": "Shipping Address",
-    "deliveryMethod": "Delivery Method",
+    "deliveryMethod": "Delivery Options",
```
kimhw0630: Changing or removing translation keys or values is breaking change
to avoid breaking change add new translation key

kimhw0630 (approved):

PR #19083, opened 2024-07-29
chore: Move multi-dimensional products to standalone library

kpawelczak:
closes: [CXSPA-8111](https://jira.tools.sap/browse/CXSPA-8111)
- updated schematics
- moved "selector" to a sub-library - affects PDP
- there will be another sub-library "Order Grid" - which will affect PLP, Order Grid Page, Cart


feature-libs/product-multi-dimensional/README.md
```diff
@@ -0,0 +1 @@
+TBD
```
pawelfras, line 1: `could:` let's consider creating a separate ticket for this matter and putting the ticket number in TODO comment

feature-libs/product-multi-dimensional/selector/components/selector/product-multi-dimensional-selector.module.ts
```diff
@@ -0,0 +1,31 @@
...
+import { ProductMultiDimensionalSelectorComponent } from './product-multi-dimensional-selector.component';
+
+@NgModule({
+  imports: [CommonModule, RouterModule, UrlModule, MediaModule],
+  declarations: [ProductMultiDimensionalSelectorComponent],
+  exports: [ProductMultiDimensionalSelectorComponent],
+  providers: [
+    provideDefaultConfig(<CmsConfig>{
+      cmsComponents: {
+        // ProductVariantMultiDimensionalSelectorComponent
```
pawelfras, line 22: `query:` is this comment left here for a purpose?

feature-libs/product-multi-dimensional/selector/components/selector/product-multi-dimensional-selector.component.html
```diff
@@ -27,7 +27,7 @@
             <option
               *ngFor="let option of category.variantOptions"
               [value]="option.code"
-              [selected]="isSelected(option.code)"
+              [selected]="isSelected(option.code, product)"
```
pawelfras, line 30: `query:` is this component often re-rendered? or it's rather destroyed and recreated? I'm asking in terms of invoking a method in the template - we have to remember that the method is going to be called for each element each time this component re-renders.

kpawelczak: Every time the user changes the variant, it will trigger a new page load. Additionally, invoking a method from the template in this case has a marginal impact on performance.

feature-libs/product-multi-dimensional/selector/components/selector/product-multi-dimensional-selector.module.ts
kpawelczak: Yes, it's on purpose, it's part of another ticket. https://jira.tools.sap/browse/CXSPA-7914

feature-libs/product-multi-dimensional/README.md
kpawelczak: thanks for the suggestion. I've created one https://jira.tools.sap/browse/CXSPA-8114

feature-libs/product-multi-dimensional/selector/components/selector/product-multi-dimensional-selector.module.ts
pawelfras: Thank you for your explanation! Just a proposal - let's consider mentioning this ticket in the comment 👍

feature-libs/product-multi-dimensional/selector/components/selector/product-multi-dimensional-selector.component.html
pawelfras: I appreciate the clarification. That was my assumption but I wanted to double-check

projects/schematics/src/add-spartacus/schema.json
```diff
@@ -184,6 +185,10 @@
             "value": "Product-Variants",
             "label": "Product - Variants"
           },
+          {
+            "value": "Product-multi-dimensional-selector",
+            "label": "Product multi-dimensional selector"
```
pawelfras: What do you think about changing label to
```suggestion
            "label": "Product - Multi-Dimensional Selector"
```
or
```suggestion
            "label": "Product Multi-Dimensional - Selector"
```
to keep the convention we have for other features? I like the first option because this feature applies to the product (like other sub-features in the `product` feature-lib), but I don't have a strong opinion on this.

kpawelczak: thanks, I think the second one is better, as this is not a product sub-lib, but a sub-lib of product multi-dmiensional.

feature-libs/product-multi-dimensional/selector/components/selector/product-multi-dimensional-selector.module.ts
kpawelczak: I've already, replaced it with a proper selector in uncommited changes :)

pawelfras (approved):

PR #19084, opened 2024-07-30
fix: Show error msg when B2B register failed

scarai-sap:
ATT. See https://jira.tools.sap/browse/CXSPA-7721


kimhw0630 (approved):
code looks good to me

PR #19090, opened 2024-07-31
fix: (CXSPA-937) - Options count for ng-select

Pio-Bar:
Ticket: [CXSPA-937](https://jira.tools.sap/browse/CXSPA-937)

- Adds an aria-label with options count to the standard select elements in the header.
- 'NgSelectA11yDirective' will now append a similar aria label to each option in the component.


projects/storefrontlib/shared/components/ng-select-a11y/ng-select-a11y.directive.ts
```diff
@@ -44,4 +62,24 @@ export class NgSelectA11yDirective implements AfterViewInit {
       this.renderer.setAttribute(divCombobox, 'aria-controls', ariaControls);
     }
   }
+
+  appendAriaLabelToOptions(
+    _changes: MutationRecord[],
+    observerInstance: MutationObserver
+  ) {
+    const options =
```
rmch91: We can use `HTMLOptionElement[]` to improve type safety and get rid of `any` below.

rmch91 (dismissed):

developpeurweb (dismissed):

Zeyber (dismissed):

Zeyber (approved):

PR #19101, opened 2024-08-02
feat: Create component selector for Product Multi-Dimensional - Selector feature

kpawelczak:
closes: [CXSPA-7914](https://jira.tools.sap/browse/CXSPA-7914)


feature-libs/product-multi-dimensional/selector/components/selector/product-multi-dimensional-selector.component.html
pawelfras, line 1: `nitpick:` `MediaComponent` doesn't support `Image[]` type for `container` input. I guess we need to extend supported types.
image

feature-libs/product-multi-dimensional/selector/components/selector/product-multi-dimensional-selector.component.ts
```diff
@@ -42,14 +42,17 @@ export class ProductMultiDimensionalSelectorComponent {
...
+    .getProduct(ProductScope.MULTI_DIMENSIONAL)
+    .pipe(
+      filter(isNotNullable),
+      filter((product: Product) => !!product?.multidimensional),
+      distinctUntilChanged(),
+      shareReplay(1),
+      tap((product: Product) => {
+        this.categories = this.getVariants(product);
+      })
+    );
```
pawelfras:
```suggestion
  product$: Observable<Product> = this.currentProductService
    .getProduct(ProductScope.MULTI_DIMENSIONAL)
    .pipe(
      filter(isNotNullable),
      filter((product) => !!product.multidimensional),
      distinctUntilChanged(),
      shareReplay(1),
      tap((product) => {
        this.categories = this.getVariants(product);
      })
    );
```
Types are not required due to type narrowing. In addition, you don't need to use optional chaining as at this point we know `product` is not null.
Apart from that, I'm leaving an alternative usage of the `filter` operator that combines two conditions:
```
filter((product): product is Product => isNotNullable(product) && !!product.multidimensional),
```
Checking the condition during one iteration might be more performant, but I'm not sure whether we suffer on the performance in this place.

feature-libs/product-multi-dimensional/selector/components/selector/product-multi-dimensional-selector.component.html
pawelfras: `nitpick`: `value` does not exist on type `EventTarget`. 
image
Consider changing the implementation of `changeVariant` to:
```
changeVariant(event: Event): void {
    const code = (event.target as HTMLSelectElement).value;
    if (code) {
      /* ... */
    }
}
```

feature-libs/product-multi-dimensional/selector/components/product-multi-dimensional-selector-components.module.ts
```diff
@@ -5,9 +5,22 @@
...
 import { ProductMultiDimensionalSelectorModule } from './selector/product-multi-dimensional-selector.module';
+import { ProductMultiDimensionalSelectorComponent } from './selector/product-multi-dimensional-selector.component';
+import { ProductMultiDimensionalSelectorGuard } from './guards';
 
 @NgModule({
   imports: [ProductMultiDimensionalSelectorModule],
+  providers: [
+    provideDefaultConfig(<CmsConfig>{
+      cmsComponents: {
+        ProductMultiDimensionalSelectorComponent: {
```
pawelfras: `query:` Just thinking out loud - knowing that the customer's application can't have both options - variants and multi-d, what are the benefits of having separate CMS component for multi-d? Wouldn't be enough to use a variant CMS component?

feature-libs/product-multi-dimensional/selector/components/product-multi-dimensional-selector-components.module.ts
```diff
@@ -5,9 +5,22 @@
...
 import { NgModule } from '@angular/core';
+import { CmsConfig, provideDefaultConfig } from '@spartacus/core';
 import { ProductMultiDimensionalSelectorModule } from './selector/product-multi-dimensional-selector.module';
+import { ProductMultiDimensionalSelectorComponent } from './selector/product-multi-dimensional-selector.component';
+import { ProductMultiDimensionalSelectorGuard } from './guards';
 
 @NgModule({
   imports: [ProductMultiDimensionalSelectorModule],
+  providers: [
+    provideDefaultConfig(<CmsConfig>{
```
pawelfras: `major:` I'd rather move this config to the `ProductMultiDimensionalSelectorModule` -`ProductMultiDimensionalSelectorComponent` and `ProductMultiDimensionalSelectorGuard` are declared and provided there and such module can also encapsulate them together with the config. The components module doesn't have to have a notion about this.

feature-libs/product-multi-dimensional/selector/core/services/product-multi-dimensional-selector.service.ts
```diff
@@ -157,7 +157,7 @@ export class ProductMultiDimensionalSelectorService {
   protected createVariantCategoryGroup(
     variantMatrix: VariantMatrixElement[]
   ): VariantCategoryGroup {
-    const parentVariantCategory = variantMatrix[0].parentVariantCategory;
+    const parentVariantCategory = variantMatrix[0]?.parentVariantCategory;
```
pawelfras: `query:` Are we sure at this stage that `variantMatrix` is not null and there is an object at index 0?

feature-libs/product-multi-dimensional/selector/components/selector/product-multi-dimensional-selector.component.html
pawelfras: `minor:` I've noticed the arrow inside the select element is cut.
image
Btw. in our code, in most of the cases, we use `ng-select`. Did you consider using it?

projects/core/src/product/model/product-scope.ts
```diff
@@ -14,5 +14,5 @@ export const enum ProductScope {
   STOCK = 'stock',
   PROMOTIONS = 'promotions',
   LIST_ITEM = 'list_item',
-  MULTIDIMENSIONAL = 'multidimensional',
+  MULTI_DIMENSIONAL = 'multi_dimensional',
```
pawelfras, line 17: `nitpick`: I've checked and both versions, `multidimensional` and `multi-dimensional` are correct. It would be good to ask someone (maybe Bill) what version we want to use in official docs and adjust the code to be in line with documentation and how this feature is going to be introduced to the customers.

kpawelczak: I've already decided to go with `Multi-Dimensional`, here is past conversation https://github.com/SAP/spartacus/pull/19040#discussion_r1682603065

feature-libs/product-multi-dimensional/selector/core/services/product-multi-dimensional-selector.service.ts
kpawelczak: good catch, I thought, I've fixed that.

kpawelczak: I've changed the name to elements as variantMatrix refers to the whole tree
```typescript
  protected createVariantCategoryGroup(
    elements: VariantMatrixElement[]
  ): VariantCategoryGroup {
    const parentVariantCategory = elements.length ? elements[0]?.parentVariantCategory : undefined;
```

feature-libs/product-multi-dimensional/selector/components/product-multi-dimensional-selector-components.module.ts
kpawelczak: Sure, I have seen features implemented both ways. Additionally, I've changed the module name as I was confused for a moment:
`ProductMultiDimensionalSelectorModule` ->`ProductMultiDimensionalSelectorComponentModule`

feature-libs/product-multi-dimensional/selector/components/product-multi-dimensional-selector-components.module.ts
kpawelczak: Good point, but we need it. Either way, customers will have to create a CMS component. 
Moreover, I want to have two selectors for:

- Multi-D Selector -> PDP
- Multi-D Order Grid -> PDP?, Cart, and some others

feature-libs/product-multi-dimensional/selector/components/selector/product-multi-dimensional-selector.component.ts
kpawelczak: TypeScript doesn't like this `filter((product) => !!product.multidimensional),` which is why I've added the check. Also, it breaks the library build if left without the null check. In general, JavaScript is rather fast, so I don't think it makes any difference.

feature-libs/product-multi-dimensional/selector/components/selector/product-multi-dimensional-selector.component.html
kpawelczak: 1. I'll wait for accessibility pr
2. I have considered this and I would need to create two separate functions one for cx-media and the other for option, or add multiple types to single value..anyways I need to wait for accessibity
3. I would rather not use ng-select. The select container should look same as "Variants" feature and this is how they do it.

feature-libs/product-multi-dimensional/selector/components/product-multi-dimensional-selector-components.module.ts
```diff
@@ -5,9 +5,9 @@
  */
 
 import { NgModule } from '@angular/core';
-import { ProductMultiDimensionalSelectorModule } from './selector/product-multi-dimensional-selector.module';
+import { ProductMultiDimensionalSelectorComponentModule } from './selector/product-multi-dimensional-selector-component.module';
```
pawelfras, line 8: `minor:` Based on similar places in the app, I think you don't need to mention `Component` in the name of the class and the file. `ProductMultiDimensionalSelectorModule` would be good enough.

pawelfras: As discussed offline, due to naming conflict with the root module of this sub-feature, let's keep `ProductMultiDimensionalSelectorComponentModule`

pawelfras (approved):

PR #19103, opened 2024-08-02
Add accessibility attributes

rmch91:


developpeurweb:
@rmch91 
Looks good, page title gets narrated on change, that’s perfect. However, I do have some requests please:

* **Color group:** JAWS narration seems to convey the colour of the button instead of the colour of the item, E.g “Red button”, let’s extend the alt text and tooltip with the word “select color” as well, E.g. “select color Red” so it ends up narrating “select color Red button” (Example 2) 
**_Also, is it possible to:_** 
    * Indicate the selected option "as text" next to the visual name of the group? (Example 1 and 2)
    * Add a “selected” state to the button that is selected? (Example 1 and 2)
    * Add `role="list"` and `role="listitem”`? So the number of items gets narrated when accessing the group (Example 1)
* **Size group:** For some reason the Size group label doesn’t get narrated on JAWS, but no issues with VO. Please check the Results Page combobox, that one seems to work fine on both SR.
* **CTA Button:** If possible, please move the Notify Me/Add-To-Cart button under the previous groups so SR users can find the multi-D options before adding the item to cart or to the notification list. That said, customers should be able to place the button wherever they want, if position is not configurable, then leave as is.

rmch91:
@developpeurweb Hello Miguel. Thank you. 
Moving the CTA Button is not possible, at least right now, product page has a lot of dependencies and order could be different based on specific product and features available for this product. So for now we need to leave it as it is.

I'm not sure i understand this point: 
Size group: For some reason the Size group label doesn’t get narrated on JAWS, but no issues with VO. Please check the Results Page combobox, that one seems to work fine on both SR.

On which element it should be pronounced?

feature-libs/product/multi-dimensional/components/multi-dimensional-container/multi-dimensional-selector/product-multi-dimensional-selector.component.html
```diff
@@ -1,43 +1,72 @@
...
+            >
+            </cx-media>
+          </button>
+        </div>
+      </label>
 
       <ng-template #noImages>
         <div class="select-variant-container">
-          <select
+          <ng-select
```
kpawelczak: What are the benefits of using `ng-select`?

feature-libs/product/multi-dimensional/components/multi-dimensional-container/multi-dimensional-selector/product-multi-dimensional-selector.component.html
```diff
@@ -1,43 +1,72 @@
...
-      <label for="{{category.name}}">{{ category.name }}</label>
+      <label>
+        <span>
+          <ng-container
+            *ngIf="
+              getSelectedValue(category.name) as selectedValue;
+              else tmpWithoutLabel
+            "
+          >
+            {{ category.name }}: {{ selectedValue }}
```
kpawelczak: This looks strange when associated with a select element. Wouldn't it be better if we add this only if there are images? Also, would prefer if it was in the method. 
```typescript
  getCategoryName(category): string {
    if (category.hasImages) {
      const selectedValue = this.getSelectedValue(category.name);
      return `${category.name}${selectedValue ? `: ${selectedValue}` : ''}`;
    }

    return category.name;
  }
```

feature-libs/product/multi-dimensional/components/multi-dimensional-container/multi-dimensional-selector/product-multi-dimensional-selector.component.ts
```diff
@@ -61,7 +68,50 @@ export class ProductMultiDimensionalSelectorComponent implements OnChanges {
...
+    }
+
+    return null;
+  }
+
+  isSelected(code: string): boolean {
     return code === this.product.code;
   }
+
+  getAriaLabel(option: VariantCategoryOption, categoryName: string) {
```
kpawelczak: In general, we should avoid subscribing directly in the template using methods. Because subscriptions are asynchronous, this practice can lead to hard-to-find bugs.

```typescript
 onAriaLabel(option: VariantCategoryOption, categoryName: string): Observable<string> {
    const isSelected = this.isSelected(option.code);

    if (isSelected) {
      return this.translationService
        .translate('multiDimensional.selectedVariant')
        .pipe(map(
          (text) =>
            `${text}, ${option.value} ${categoryName}`
        ));
    }

    return this.translationService
      .translate('multiDimensional.variantThumbnailTitle', {
        value: option.value,
        category: categoryName,
      })
      .pipe(map((text) => `${text}`));
  }
```

feature-libs/product/multi-dimensional/components/multi-dimensional-container/multi-dimensional-selector/product-multi-dimensional-selector.component.ts
```diff
@@ -61,7 +68,50 @@ export class ProductMultiDimensionalSelectorComponent implements OnChanges {
...
+    const category = this.categories.find((cat) => cat.name === categoryName);
+
+    if (category && this.selectedVariant) {
+      const selectedOption = category.variantOptions.find(
+        (option) => option.code === this.selectedVariant
+      );
+      return selectedOption ? selectedOption.value : '';
+    }
+
+    return null;
```
kpawelczak: I believe we should avoid using null values. It would be better to return an empty string instead.

feature-libs/product/multi-dimensional/components/multi-dimensional-container/multi-dimensional-selector/product-multi-dimensional-selector.component.html
```diff
@@ -1,43 +1,72 @@
...
+            class="image-variant"
+            [ngClass]="{ active: option?.code === product.code }"
+            [attr.aria-label]="getAriaLabel(option, category.name)"
+            (click)="changeVariant(option.code)"
+            [cxFocus]="{ key: option.code }"
+          >
+            <cx-media
+              [container]="option.images"
+              role="img"
+              [alt]="
```
kpawelczak: I’m wondering if we should consider using the alt text set in the Backoffice. However, keep in mind that customers might not associate it with this specific multi-dimensional case.

kpawelczak (changes requested):
Please take a look at this branch https://github.com/SAP/spartacus/tree/feature/CXSPA-7465-accessibility before adding any changes

feature-libs/product/multi-dimensional/components/multi-dimensional-container/multi-dimensional-selector/product-multi-dimensional-selector.component.html
rmch91: Right, I think it should be alt from backoffice.

feature-libs/product/multi-dimensional/components/multi-dimensional-container/multi-dimensional-selector/product-multi-dimensional-selector.component.ts
rmch91: changed

feature-libs/product/multi-dimensional/components/multi-dimensional-container/multi-dimensional-selector/product-multi-dimensional-selector.component.ts
rmch91: Changed

feature-libs/product/multi-dimensional/components/multi-dimensional-container/multi-dimensional-selector/product-multi-dimensional-selector.component.html
rmch91: changed

feature-libs/product/multi-dimensional/components/multi-dimensional-container/multi-dimensional-selector/product-multi-dimensional-selector.component.html
rmch91: I talk with accessibility team, they prefer to use ng-select everywhere. It is more flexible from the perspective of passing  different attributes and directives. I didn't argue

kpawelczak (approved):

PR #19144, opened 2024-08-21
feat: make cdc oidc login flow get rid of dependency for angular-oauth2-oidc

i53577:


projects/core/src/auth/user-auth/services/oauth-lib-wrapper.service.ts
```diff
@@ -22,15 +23,35 @@ import { AuthConfigService } from './auth-config.service';
...
+    protected winRef: WindowRef,
+    protected baseSiteService?: BaseSiteService
   ) {
     this.initialize();
+    if (baseSiteService) {
+      baseSiteService
+        .get()
+        .pipe(take(1))
+        .subscribe((site) => {
+          if (site?.cdcSiteConfig && site?.uid) {
```
i53577: only MCS has cdcSiteConfig, with this we can set oidc config and don't need to add feature flag.

projects/core/src/auth/user-auth/services/oauth-lib-wrapper.service.ts
```diff
@@ -142,21 +163,34 @@ export class OAuthLibWrapperService {
...
-      this.oAuthService
-        .tryLogin({
-          // We don't load discovery document, because it doesn't contain revoke endpoint information
-          disableOAuth2StateCheck: true,
-        })
-        .then((result: boolean) => {
-          resolve({
-            result: result,
-            tokenReceived: !!tokenReceivedEvent,
+      if (this.currentBaseSite.cdcSiteConfig) {
```
i53577: Only mcs baseSite has cdcSiteConfig. If baseSite has cdcSiteConfig, then trylogin() need to load discovery document.
If not, all the logic won't change.

feature-libs/user/account/components/login-form-oidc/login-form-oidc.component.ts
```diff
@@ -5,49 +5,14 @@
...
-            site.cdcSiteConfig.oidcRpClientId,
-            site.uid,
-            site.cdcSiteConfig.scopes.join(' '),
-            'code'
-          );
-          this.oauthService.configure(authConfig);
-          this.oauthService.loadDiscoveryDocumentAndLogin();
-        }
-      });
+    this.oauthLibWrapper.initLoginFlow();
```
i53577: only need to call initLoginFlow() to start login with code. 
Discovery document is loaded by oAuthLibWrapper.

projects/core/src/auth/user-auth/services/oauth-lib-wrapper.service.ts
i53577: During each time a page is opened, this method will be called by AuthService to log in user. After logging successfully, the page will be redirected.
https://github.com/SAP/spartacus/blob/develop/projects/core/src/auth/user-auth/facade/auth.service.ts#L55

projects/core/src/auth/user-auth/facade/auth.service.ts
```diff
@@ -179,10 +179,4 @@ export class AuthService {
   setLogoutProgress(progress: boolean): void {
     (this.logoutInProgress$ as BehaviorSubject<boolean>).next(progress);
   }
-
-  afterRedirectFromOidcLogin(): void {
```
i53577, line 183: these logic will be done by https://github.com/SAP/spartacus/blob/develop/projects/core/src/auth/user-auth/facade/auth.service.ts#L55

projects/core/src/auth/user-auth/services/oauth-lib-wrapper.service.ts
```diff
@@ -22,15 +23,35 @@ import { AuthConfigService } from './auth-config.service';
...
   // TODO: Remove platformId dependency in 4.0
   constructor(
     protected oAuthService: OAuthService,
     protected authConfigService: AuthConfigService,
     @Inject(PLATFORM_ID) protected platformId: Object,
-    protected winRef: WindowRef
+    protected winRef: WindowRef,
+    protected baseSiteService?: BaseSiteService
   ) {
     this.initialize();
```
i53577: `this.initialize();` will config oAuthService with [defaultAuthConfig](https://github.com/SAP/spartacus/blob/develop/projects/core/src/auth/user-auth/config/default-auth-config.ts). Provide a new defaultOidcAuthConfig will be breaking change. So I add the following logic to load oidc config from baseSite.

projects/core/src/auth/user-auth/services/oauth-lib-wrapper.service.ts
```diff
@@ -22,15 +23,35 @@ import { AuthConfigService } from './auth-config.service';
...
+  currentBaseSite: BaseSite;
 
   // TODO: Remove platformId dependency in 4.0
   constructor(
     protected oAuthService: OAuthService,
     protected authConfigService: AuthConfigService,
     @Inject(PLATFORM_ID) protected platformId: Object,
-    protected winRef: WindowRef
+    protected winRef: WindowRef,
+    protected baseSiteService?: BaseSiteService
```
niehuayang: Could we use inject(BaseSiteService)?

projects/core/src/auth/user-auth/services/oauth-lib-wrapper.service.ts
```diff
@@ -22,15 +23,35 @@ import { AuthConfigService } from './auth-config.service';
...
   events$: Observable<OAuthEvent> = this.oAuthService.events;
+  currentBaseSite: BaseSite;
 
   // TODO: Remove platformId dependency in 4.0
   constructor(
     protected oAuthService: OAuthService,
     protected authConfigService: AuthConfigService,
     @Inject(PLATFORM_ID) protected platformId: Object,
-    protected winRef: WindowRef
+    protected winRef: WindowRef,
```
scarai-sap: use inject?

projects/core/src/auth/user-auth/services/oauth-lib-wrapper.service.ts
```diff
@@ -22,15 +23,35 @@ import { AuthConfigService } from './auth-config.service';
 })
 export class OAuthLibWrapperService {
   events$: Observable<OAuthEvent> = this.oAuthService.events;
+  currentBaseSite: BaseSite;
```
scarai-sap: looks the only way we need this is in tryLogin to judge whether call loadDiscoveryDocumentAndTryLogin. Maybe use a flag instead of keep the whole baseSite?

projects/core/src/auth/user-auth/services/oauth-lib-wrapper.service.ts
i53577: Yes we can. But it seems here using optional baseSiteService or inject make no difference?

niehuayang: Using inject()  function helps to avoid introducing unnecessary breaking changes,  when adding new dependency to a class doesn't require changing the constructor's signature.

niehuayang: see 2.12: https://wiki.one.int.sap/wiki/display/spar/Coding+Guidelines

i53577: Constructor is already used here, and baseSiteService is optional, it should not be breaking change.

i53577: For those customers who are using spartacus already, if using inject, the following logic may stop at `if (site?.cdcSiteConfig && site?.uid) {`. If using optional in constructor, the logic may stop at ` if (baseSiteService) {`. From this perspective, I think constructor is better than inject.
```
    if (baseSiteService) {
      baseSiteService
        .get()
        .pipe(take(1))
        .subscribe((site) => {
          if (site?.cdcSiteConfig && site?.uid) {
```

scarai-sap: but adding parameters for constructor is imcompatbile change?

projects/core/src/auth/user-auth/services/oauth-lib-wrapper.service.ts
i53577: I am trying to avoid introduce new flag. Maybe we can use `this.oAuthService.responseType==='code'` instead of `if (this.currentBaseSite.cdcSiteConfig) {` in [line166](https://github.com/SAP/spartacus/pull/19144/files/5caa4eb8c164e8b47e87b5fc0bc190d93c0d6043#diff-d359f9c3ad517cd184767bb6917d16624704a7b4430cceaaa968ffa457e97301R166) if you think it's better.

projects/core/src/auth/user-auth/services/oauth-lib-wrapper.service.ts
i53577: Here is optional, so it should not be breaking change. 
For those customers who are using spartacus already, if using inject, the following logic may stop at `if (site?.cdcSiteConfig && site?.uid) {`. If using optional in constructor, the logic may stop at ` if (baseSiteService) {`. From this perspective, I think constructor is better than inject.
```
    if (baseSiteService) {
      baseSiteService
        .get()
        .pipe(take(1))
        .subscribe((site) => {
          if (site?.cdcSiteConfig && site?.uid) {
```
Anyway, it doesn't make much difference for me. If you all think inject is better, I can switch it to use `inject` :)

projects/core/src/auth/user-auth/services/oauth-lib-wrapper.service.ts
i53577: it should not be incompatible as along as it's optional.  Correct me if I'm wrong.

i53577: adapt to `inject`

projects/core/src/auth/user-auth/services/oauth-lib-wrapper.service.ts
i53577: adapt to `inject`

projects/core/src/auth/user-auth/services/oauth-lib-wrapper.service.ts
```diff
@@ -31,6 +36,24 @@ export class OAuthLibWrapperService {
     protected winRef: WindowRef
   ) {
     this.initialize();
+    if (this.baseSiteService) {
```
scarai-sap: you can use this.baseSiteService?.get directly. the ? help you to skip if

scarai-sap (approved):
LGTM

niehuayang (approved):

projects/core/src/auth/user-auth/services/oauth-lib-wrapper.service.ts
```diff
@@ -31,6 +36,24 @@ export class OAuthLibWrapperService {
...
+        .get()
+        .pipe(take(1))
+        .subscribe((site) => {
+          if (site?.cdcSiteConfig && site?.uid) {
+            this.currentBaseSite = site;
+            this.oAuthService.configure({
+              clientId: site.cdcSiteConfig.oidcRpClientId,
+              issuer: site.cdcSiteConfig.oidcOpIssuerURI,
+              redirectUri:
+                window.location.origin + '/' + site.uid + '/en/USD/login',
```
scarai-sap: The redirect url might be incorrect as we hardcode lan and currency.

scarai-sap (approved):

feature-libs/user/account/components/login-form-oidc/login-form-oidc.component.ts
vinchen666, line 1: there's no form in this component, shall we rename this component and module to `login-oidc.component`

projects/core/src/auth/user-auth/services/oauth-lib-wrapper.service.ts
```diff
@@ -31,6 +36,39 @@ export class OAuthLibWrapperService {
     protected winRef: WindowRef
   ) {
     this.initialize();
+    if (this.baseSiteService) {
+      this.baseSiteService
```
vinchen666: we should not mix cdc logic into this wrapper, the wrapper should be a generic wrapper for any oidc process.

shall we move to oidc login component?

projects/core/src/auth/user-auth/services/oauth-lib-wrapper.service.ts
```diff
@@ -142,21 +180,34 @@ export class OAuthLibWrapperService {
...
-      this.oAuthService
-        .tryLogin({
-          // We don't load discovery document, because it doesn't contain revoke endpoint information
-          disableOAuth2StateCheck: true,
-        })
-        .then((result: boolean) => {
-          resolve({
-            result: result,
-            tokenReceived: !!tokenReceivedEvent,
+      if (this.currentBaseSite.cdcSiteConfig) {
```
vinchen666: we should not mix cdc logic into this wrapper, the wrapper should be a generic wrapper for any oidc process.

projects/core/src/auth/user-auth/services/oauth-lib-wrapper.service.ts
```diff
@@ -31,6 +36,39 @@ export class OAuthLibWrapperService {
...
+            site?.cdcSiteConfig &&
+            site?.uid &&
+            site.baseStore?.defaultCurrency &&
+            site.defaultLanguage
+          ) {
+            const defaultCurrency = site.baseStore.defaultCurrency.isocode;
+            const defaultLanguage = site.defaultLanguage.isocode;
+
+            this.currentBaseSite = site;
+            this.oAuthService.configure({
```
vinchen666: 1. move this section of code to oidc component, 2. use `AuthConfigService` in oidc component to 

eg:
config = AuthConfigService.getConfig();
config.clientId = site.cdcSiteConfig.oidcRpClientId;

feature-libs/user/account/components/login-form-oidc/login-form-oidc.component.ts
i53577: renamed

vinchen666 (approved):

vinchen666 (approved):
👍

projects/core/src/auth/user-auth/services/oauth-lib-wrapper.service.ts
```diff
@@ -29,9 +29,7 @@ export class OAuthLibWrapperService {
     protected authConfigService: AuthConfigService,
     @Inject(PLATFORM_ID) protected platformId: Object,
     protected winRef: WindowRef
-  ) {
-    this.initialize();
```
i53577, line 33: remove this, since initialize() will be called in https://github.com/SAP/spartacus/pull/19144/files#diff-823fabb03957246defb4097e7b4cbd5dd4f996bc0cce847f71ed8ba173d55109R42

vinchen666 (approved):

PR #19145, opened 2024-08-21
CXSPA-8201: code refactor

scarai-sap:
ATT.

Small code refactor for feature CAPTCHA

### DIFF
--- a/projects/storefrontlib/shared/components/captcha/captcha.renderer.ts
+++ b/projects/storefrontlib/shared/components/captcha/captcha.renderer.ts
@@ -8,7 +8,7 @@ import { Observable } from 'rxjs';
 import { RenderParams } from './captcha.model';
 import { CaptchaConfig } from '@spartacus/core';
 
-export interface CaptchaRender {
+export interface CaptchaRenderer {
   /**
    * Renders captcha widget
    *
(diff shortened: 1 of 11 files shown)


i53577 (dismissed):

Zeyber (approved):

projects/storefrontlib/shared/components/captcha/captcha.renderer.ts
```diff
@@ -0,0 +1,30 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { Observable } from 'rxjs';
+import { RenderParams } from './captcha.model';
+import { CaptchaConfig } from '@spartacus/core';
+
+export interface CaptchaRenderer {
+  /**
+   * Renders captcha widget
```
meeque, line 13: Could we also add something like:

> Implementations must use secure DOM manipulation methods to protect against XSS.

projects/storefrontlib/shared/components/captcha/captcha.renderer.ts
```diff
@@ -0,0 +1,30 @@
...
+   *
+   * @RenderParams parameters required by a provider, includes html element to append CAPTCHA UI element.
+   * @returns Observable for the key returned by a provider once user validated
```
meeque, line 16: Why is it called *key* here? It's called *token* everywhere else. Or are these different things?

PR #19152, opened 2024-08-22
fix: (CXSPA-1128) - NgSelect double stop on mobile alternative fix

Pio-Bar:
Ticket: [CXSPA-1128](https://jira.tools.sap/browse/CXSPA-1128)

To avoid the console error the previous solution caused, we can hide the display value from the screen reader and instead provide it as a part of the aria-label. We can limit this solution to the mobile devices since the bug is only present for such.

I could not come up with a better solution that would satisfy the requirement. This feels a bit hacky to me, I would like @developpeurweb to share his opinion on this.


developpeurweb (dismissed):

projects/storefrontlib/shared/components/ng-select-a11y/ng-select-a11y.directive.ts
```diff
@@ -41,6 +43,10 @@ export class NgSelectA11yDirective implements AfterViewInit {
     observer.observe(this.elementRef.nativeElement, { childList: true });
   }
 
+  @Optional() breakpointService = inject(BreakpointService, { optional: true });
+
+  ariaLabelAttribute = 'aria-label';
```
Zeyber: Is there any reason we have made this global rather than just using the `aria-label` string when we `setAttribute`? I don't see it being reassigned in the class.

projects/storefrontlib/shared/components/ng-select-a11y/ng-select-a11y.directive.ts
```diff
@@ -85,11 +107,37 @@ export class NgSelectA11yDirective implements AfterViewInit {
...
+                ariaLabel
+              );
             }
           );
         });
     }
     observerInstance.disconnect();
   }
+
+  appendValueToAriaLabel(
```
Zeyber, line 123: Could we add some jsdoc comment to explain what this method is doing? We should also cover it with unit tests.

Zeyber (changes requested):

projects/storefrontlib/shared/components/ng-select-a11y/ng-select-a11y.directive.ts
Pio-Bar: I'd be happy with the strings too. However, the Sonarqube is not and gets upset about repeated use :(

projects/storefrontlib/shared/components/ng-select-a11y/ng-select-a11y.directive.ts
Pio-Bar: Added comment + unit test

projects/storefrontlib/shared/components/ng-select-a11y/ng-select-a11y.directive.ts
Zeyber: Ok. Lets move this outside the class then as a constant so it doesn't look like it will change anywhere.

eg.
```
const ARIA_LABEL = 'aria-label';

...

export class ... {
```

projects/storefrontlib/shared/components/ng-select-a11y/ng-select-a11y.directive.ts
```diff
@@ -67,7 +73,19 @@ export class NgSelectA11yDirective implements AfterViewInit {
       this.featureConfigService.isEnabled('a11yNgSelectMobileReadout') &&
       inputElement.readOnly
     ) {
-      this.renderer.setAttribute(inputElement, 'aria-hidden', 'true');
+      const selectObserver = new MutationObserver((changes, observer) => {
```
Zeyber: Couple points with this observer:
- We only start observing in mobile view. Should we create this observer in the breakpoint subscription too?
- If the user resizes their screen from desktop to below the breakpoint, would this observer be created? Should it be removed when returning to desktop size?

Zeyber (changes requested):

projects/storefrontlib/shared/components/ng-select-a11y/ng-select-a11y.directive.ts
Pio-Bar: The bug exists only for mobile devices' screen readers. It is device dependent, the behaviour should not change if the screen is resized. This observer implementation should work, but maybe there is a better way to check if the user is on mobile? 🤔

Zeyber: Ok, lets move this observer into the breakpoint service subscription and leave it at that since we don't need it on non-mobile devices.

There are other ways to detect if devices are using mobile. But I think that breakpoints are enough for this specific case :)

Zeyber (changes requested):

Zeyber (dismissed):

developpeurweb (dismissed):

Zeyber (approved):

PR #19165, opened 2024-08-27
fix: move price range atr to default product search occ endpoint

kpawelczak:
closes: [CXSPA-8245](https://jira.tools.sap/browse/CXSPA-8245)


projects/schematics/src/shared/lib-configs/product-multi-dimensional-schematics-config.ts
```diff
@@ -41,20 +36,11 @@ export const PRODUCT_MULTI_DIMENSIONAL_LIST_SCHEMATICS_CONFIG: SchematicConfig =
     },
     folderName: PRODUCT_MULTI_DIMENSIONAL_FOLDER_NAME,
     moduleName: PRODUCT_MULTI_DIMENSIONAL_LIST_MODULE_NAME,
-    featureModule: [
-      {
-        name: PRODUCT_MULTI_DIMENSIONAL_LIST_MODULE,
-        importPath: SPARTACUS_PRODUCT_MULTI_DIMENSIONAL_LIST,
-      },
-    ],
+    featureModule: [],
```
pawelfras, line 39: `medium:` Let's remove this property if empty.

feature-libs/product-multi-dimensional/list/product-multi-dimensional-list.module.ts
pawelfras, line 1: `major:` instead of keeping this empty module, we can `export {}` in `public_api.ts` and then remove this file as it is used only for "fake" public API purposes.

projects/schematics/src/shared/lib-configs/product-multi-dimensional-schematics-config.ts
kpawelczak: It's required, I can't remove it.
```typescript
export interface SchematicConfig {
...
  featureModule: Module | Module[];
...
}
```

feature-libs/product-multi-dimensional/list/product-multi-dimensional-list.module.ts
kpawelczak: some time ago, I heard it's ok to have empty modules. Anyways I went with your suggestion.

pawelfras: Thanks, Kris! Would you mind sharing the source of such a statement? I'm happy to learn new things!
Nevertheless, we already have the "export {}" pattern in Spartacus so I suggest going this way. Thank you for the adjustment!

PR #19170, opened 2024-08-28
fix: site theme not initialized correctly from storage

Platonn:
**BEFORE:**
- theme value was not restored correctly from local storage on page start
- we performed a mandatory call to OCC endpoint /basesites, even if customers didn't want to use CMS-driven theming

<details>
<summary>
Elaborating more:
</summary>

When we attempted to set a value from storage on page start, we encountered an ASYNC mechanism of validation (which in the background called the OCC endpoint /basesites). Unfortunately, then we immediately expected the value to have been set (despite it was delayed by an async validation) - we checked synchronously the value of the `isInitialized()`. This race condition led to setting improper value to ngrx on the page start.

Besides this problem, we also forced all customers to make in an APP_INITIALIZER an _obligatory_ render-blocking http call to OCC endpount /basesites, (even regardless if they used CMS-driven theming or not). Please note that we already had a different generic mechanism (`SiteContextConfigInitializer`) which performs such a render-blocking call, but customers only if customers decide to load site context data from CMS (alternatively they can configure site context statically in Spartacus config). It's a tradeoff each customer need to make - to have dynamic CMS-driven site context, Spartacus needs to make a http roundtrip.
</details>

**AFTER:**
- theme value is restored correctly from storage on page start
- we let the existing mechanism `SiteContextConfigInitializer` to possibly fetch the theme from CMS - only if a customer decides to load site context data from CMS. Alternatively, a customer can configure the theme statically in Spartacus config `config.context`.

**Side changes that needed to be made:**
1. Reverted the deprecation on the confi  `config.context[THEME_CONTEXT_ID]: string`
2. Renamed the new config `config.siteTheme.siteThemes` to `config.siteTheme.optionalThemes`.
3. The themes are _back again_ being configured via scattered 2 different configs (Note: it was the case at first in the main PR, until I first suggested to unify those configs for simplicity; only now I realized negative consequences unifying them, so reverting this decision), so now:
   - the config `config.context[THEME_CONTEXT_ID]: string` defines the default theme ID (which can be configured manually in Spartacus config OR it can be populated dynamically by the `SiteContextConfigInitializer` if a customers decides so. So we preserve the old behavior.
   - now the new config `config.siteTheme.optionalThemes: SiteTheme[]` defines ONLY the optional themes (like high contrast ones), with their IDs (class names) and a11y labels
4. If the default theme ID defined in `config.context[THEME_CONTEXT_ID]: string` is left undefined, we use an empty string `''` as the default theme ID - resulting in storing `''` in the `localStorage` and setting NO css class in DOM in case when this theme is activated
5. Fixed the mechanism of `StatePersistenceService` (the function `persistToStorage` to be precise) to allow for storing an empty string `''` - which, to my surprise, was previously impossible (due to a condition `if(value)...`). Now only the `undefined` values are prevented. For more, see the JSDoc comment with explanation in `persistToStorage()` function
6. Removed the whole NGRX code for setting/getting possible Themes. Now all Themes can be determined synchronously based on configs.
7. When a customer doesn't configure manually the default theme's ID (`config.context.theme`) and when they don't enable fetching a theme from CMS (via `StatePersistenceService`), then we we use a fallback value - an empty string `''` - as the css class / ID of the default theme. This matches the original behavior on the `develop` branch - to not set any css class when config.context.theme is not defined.

**QA steps:**
Prerequisites:
- you need to display a ThemeSwitcher component in every page (it's a CMS component but currently we're lacking sample data for it in our dev server) - here is a quick way to do it:
  - in `storefront-component.module.ts` line 29, add an import of the module `ThemeSwitcherModule` (from `'../../cms-components/misc/theme-switcher/theme-switcher.module';`)
  - in `storefront.component.html` line 1, add `<cx-theme-switcher />`
- Open Spartacus and open DevTools -> Application tab -> local storage

TEST CASES:

1. TEST CASE: When a default theme ID is NOT configured - `config.context.theme` is undefined
   1. (note: in our storefront app, no default theme ID is configured)
   1. Clear the entry for the key `spartacus⚿⚿siteTheme`, if exists.
   1. Refresh the page
   1. 👉 Verify that the entry `spartacus⚿⚿siteTheme` has a value `''` (empty string)
   1. 👉 Verify that when you run in the console `[...document.querySelector('cx-storefront').classList]`, the list doesn't contain any class related to theming
   1. In the top of the page in the ThemeSwitcher, change the theme to `HC-dark`
   1. 👉 Verify that the entry `spartacus⚿⚿siteTheme` has a value `'cx-theme-high-contrast-dark'`
   1. 👉 Verify that when you run in the console `[...document.querySelector('cx-storefront').classList]`, the list contains `'cx-theme-high-contrast-dark'`
   1. Refresh the page
   1. 👉 Verify that the entry `spartacus⚿⚿siteTheme` has a value `'cx-theme-high-contrast-dark'`
   1. 👉 Verify that when you run in the console `[...document.querySelector('cx-storefront').classList]`, the list contains `'cx-theme-high-contrast-dark'`
   1. In the top of the page in the ThemeSwitcher, change the theme back to `Default`
   1. 👉 Verify that the entry `spartacus⚿⚿siteTheme` has a value `''` (empty string)
   1. 👉 Verify that when you run in the console `[...document.querySelector('cx-storefront').classList]`, the list doesn't contain any class related to theming (e.g. it DOES NOT contain anymore `'cx-theme-high-contrast-dark'`)
1. TEST CASE: When a default theme ID IS configured - `config.context.theme` is defined
   1. in `spartacus-b2c-configuration.module.ts` in line 40, in the `context` config, add a new key: `theme: ['spike']`
   1. 👉👉👉 verify all analogically as in the previous TEST CASE, but instead of `''` you should see `spike`
1. TEST CASE: When a default theme ID is automatically configured, based on CMS Basesites  - `config.context.theme` is populated by the `SiteContextConfigInitializer`
   1. in `spartacus-b2c-configuration.module.ts` comment-out lines 36-40, effectively removing the static `context` config. (This will trigger loading loading this config from CMS Basesites. For more see [docs](https://help.sap.com/docs/SAP_COMMERCE_COMPOSABLE_STOREFRONT/eaef8c61b6d9477daf75bff9ac1b7eb4/9d2e339c2b094e4f99df1c2d7cc999a8.html?q=automatic%20site%20context))
   1. 👉👉👉 verify all analogically as in the previous TEST CASE, but instead of `''` you should see `santorini`

closes https://jira.tools.sap/browse/CXSPA-8153

### DIFF
--- a/projects/core/src/site-context/config/config-loader/site-context-config-initializer.ts
+++ b/projects/core/src/site-context/config/config-loader/site-context-config-initializer.ts
@@ -4,8 +4,8 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { inject, Injectable } from '@angular/core';
-import { Observable, lastValueFrom } from 'rxjs';
+import { Injectable } from '@angular/core';
+import { lastValueFrom, Observable } from 'rxjs';
 import { filter, map, take } from 'rxjs/operators';
 import { ConfigInitializer } from '../../../config/config-initializer/config-initializer';
 import { BaseSite } from '../../../model/misc.model';
@@ -19,11 +19,9 @@ import {
   THEME_CONTEXT_ID,
 } from '../../providers/context-ids';
 import { SiteContextConfig } from '../site-context-config';
-import { FeatureToggles } from '../../../features-config';
 
 @Injectable({ providedIn: 'root' })
 export class SiteContextConfigInitializer implements ConfigInitializer {
-  private featureToggles = inject(FeatureToggles);
   readonly scopes = ['context'];
   readonly configFactory = () => lastValueFrom(this.resolveConfig());
 
@@ -73,9 +71,10 @@ export class SiteContextConfigInitializer implements ConfigInitializer {
           source.baseStore?.currencies,
           source.baseStore?.defaultCurrency
         ),
-        ...(this.featureToggles.useNewSiteThemeSwitcher
-          ? {}
-          : { [THEME_CONTEXT_ID]: [source.theme] }),
+        // Note: The default Site Theme can be driven by the CMS, but additional possible
+        // selectable themes (like a11y high contrast themes) can be configured only
+        // in the separate Spartacus configuration.
+        [THEME_CONTEXT_ID]: [source.theme],
       },
     } as SiteContextConfig;
 
--- a/projects/core/src/site-context/config/config-loader/site-context-config-initializer.ts
+++ b/projects/core/src/site-context/config/config-loader/site-context-config-initializer.ts
@@ -73,7 +73,7 @@ export class SiteContextConfigInitializer implements ConfigInitializer {
         ),
         // Note: The default Site Theme can be driven by the CMS, but additional possible
         // selectable themes (like a11y high contrast themes) can be configured only
-        // in the separate Spartacus configuration.
+        // in the separate Spartacus configuration `config.siteTheme.optionalThemes`.
         [THEME_CONTEXT_ID]: [source.theme],
       },
     } as SiteContextConfig;
--- a/projects/core/src/site-theme/facade/site-theme.service.ts
+++ b/projects/core/src/site-theme/facade/site-theme.service.ts
@@ -87,10 +87,7 @@ export class SiteThemeService implements SiteContext<SiteTheme> {
   }
 
   isValid(className: string): boolean {
-    return (
-      !!className &&
-      this.themes.map((theme) => theme.className).includes(className)
-    );
+    return this.themes.map((theme) => theme.className).includes(className);
   }
 
   isInitialized(): boolean {
--- a/projects/core/src/site-theme/facade/site-theme.service.ts
+++ b/projects/core/src/site-theme/facade/site-theme.service.ts
@@ -52,8 +52,7 @@ export class SiteThemeService implements SiteContext<SiteTheme> {
    * CAUTION: This property should be accessed only when those configs are stable, e.g. `ConfigInitializer.getStable('context','siteTheme'))`
    */
   protected get themes(): SiteTheme[] {
-    // SPIKE TODO: rethink the key names `siteTheme.siteThemes` vs `optionalSiteThemes
-    const optionalThemes = this.config.siteTheme?.siteThemes || [];
+    const optionalThemes = this.config.siteTheme?.optionalThemes || [];
     return [this.getDefault(), ...optionalThemes];
   }
 
--- a/projects/core/src/site-theme/facade/site-theme.service.ts
+++ b/projects/core/src/site-theme/facade/site-theme.service.ts
@@ -28,26 +28,33 @@ export class SiteThemeService implements SiteContext<SiteTheme> {
   protected store = inject(Store<StateWithSiteTheme>);
   protected config = inject(Config);
 
-  protected get themes(): SiteTheme[] {
-    // There are 2 types of themes: default and optional which are configured differently
-    // 1. The default theme ID can be configured only via Spartacus config `config.context.theme`
-    //    (note: its value can be defined statically or fetched dynamically from the CMS backend).
-    const defaultThemeId = getContextParameterDefault(
-      this.config,
-      THEME_CONTEXT_ID
-    );
-    const defaultTheme: SiteTheme = {
-      className: defaultThemeId || '',
+  getDefault(): SiteTheme {
+    const defaultThemeId =
+      getContextParameterDefault(this.config, THEME_CONTEXT_ID) ?? '';
+
+    return {
+      className: defaultThemeId,
       i18nNameKey: 'themeSwitcher.themes.default',
     };
+  }
 
+  /**
+   * List of possible themes.
+   *
+   * There are 2 types of themes: default and optional which are configured differently in Spartacus config.
+   * This property combines both types of themes into a single list.
+   *
+   * 1. The default theme ID can be configured only via Spartacus config `config.context.theme`
+   *    (note: its value can be defined statically or fetched dynamically from the CMS backend).
+   * 2. The optional themes (their IDs and their i18n keys) can be configured
+   *    only via Spartacus config `config.siteTheme.siteThemes`
+   *
+   * CAUTION: This property should be accessed only when those configs are stable, e.g. `ConfigInitializer.getStable('context','siteTheme'))`
+   */
+  protected get themes(): SiteTheme[] {
     // SPIKE TODO: rethink the key names `siteTheme.siteThemes` vs `optionalSiteThemes
-
-    // 2. The optional themes (their IDs and i18n keys) can be configured
-    //    only via Spartacus config `config.siteTheme.siteThemes`
-    const optionalThemes: SiteTheme[] = this.config.siteTheme?.siteThemes || [];
-
-    return [defaultTheme, ...optionalThemes];
+    const optionalThemes = this.config.siteTheme?.siteThemes || [];
+    return [this.getDefault(), ...optionalThemes];
   }
 
   getAll(): Observable<SiteTheme[]> {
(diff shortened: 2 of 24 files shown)


projects/core/src/site-context/config/config-loader/site-context-config-initializer.ts
```diff
@@ -73,9 +71,10 @@ export class SiteContextConfigInitializer implements ConfigInitializer {
           source.baseStore?.currencies,
           source.baseStore?.defaultCurrency
         ),
-        ...(this.featureToggles.useNewSiteThemeSwitcher
-          ? {}
-          : { [THEME_CONTEXT_ID]: [source.theme] }),
+        // Note: The default Site Theme can be driven by the CMS, but additional possible
+        // selectable themes (like a11y high contrast themes) can be configured only
+        // in the separate Spartacus configuration.
```
Platonn:
```suggestion
        // in the separate Spartacus configuration `config.siteTheme.optionalThemes`.
```

projects/core/src/site-theme/facade/site-theme.service.ts
```diff
@@ -4,35 +4,61 @@
...
+  /**
+   * List of possible themes.
+   *
+   * There are 2 types of themes: default and optional which are configured differently in Spartacus config.
+   * This property combines both types of themes into a single list.
+   *
+   * 1. The default theme ID can be configured only via Spartacus config `config.context.theme`
+   *    (note: its value can be defined statically or fetched dynamically from the CMS backend).
+   * 2. The optional themes (their IDs and their i18n keys) can be configured
+   *    only via Spartacus config `config.siteTheme.siteThemes`
```
Platonn:
```suggestion
   *    only via Spartacus config `config.siteTheme.optionalThemes`
```

projects/core/src/site-theme/services/site-theme-initializer.ts
```diff
@@ -40,39 +39,19 @@ export class SiteThemeInitializer implements OnDestroy {
...
+    return this.configInit
+      .getStable('context')
+      .pipe(tap(() => this.setDefaultFromConfig()));
   }
 
   /**
-   * Sets the active theme value based on the default value,
-   * unless the active theme has been already initialized.
+   * Sets the active language value based on the default value from the config,
+   * unless the active language has been already initialized.
```
Platonn:
```suggestion
   * Sets the active theme value based on the default value from the config,
   * unless the active theme has been already initialized.
```

projects/core/src/state/utils/browser-storage.ts
```diff
@@ -35,12 +36,20 @@ export function getStorage(
...
 }
 
+/**
+ * Persists the given value to the given storage, piping it first through `JSON.stringify()`.
+ *
+ * Note: It deliberately doesn't save `undefined` values in storage.
+ *       It's because `JSON.stringify()` returns the string `"undefined"` for `undefined`.
+ *       And this string is not valid JSON and will cause a runtime `SyntaxError` in `readFromStorage()`
+ *       when `JSON.parse()` is called with the string `"undefined"` as an argument.
+ */
```
Platonn:
```suggestion
 * Note: It silently refuses to save `undefined` values in storage.
 *       It's because `JSON.stringify()` returns the string `"undefined"` for `undefined`.
 *       And this string is not valid JSON so it would  cause a runtime `SyntaxError`
 *       in `readFromStorage()` when `JSON.parse()` is called with the string `"undefined"` as an argument.
 */
```

projects/storefrontlib/layout/theme/theme.service.ts
```diff
@@ -60,13 +60,22 @@ export class ThemeService implements OnDestroy {
...
+    const element = this.rootComponent.location.nativeElement;
+
+    if (this.featureConfigService.isEnabled('useNewSiteThemeSwitcher')) {
+      if (this.existingTheme) {
+        this.renderer.removeClass(element, this.existingTheme);
+      }
+      if (theme) {
+        this.renderer.addClass(element, theme);
+        this.existingTheme = theme;
+      }
```
Platonn, line 72: New code accounts for the fact that either `existingTheme` or the new `theme` can be an empty string `''`. But the native DOM API throws an error when we try to add a CSS class `''`.

Moreover, if the old class was non-empty, but the new one is `''`, then we want to just unset the previous one.

projects/core/src/site-theme/facade/site-theme.service.ts
```diff
@@ -48,26 +74,31 @@ export class SiteThemeService {
...
-    this._isInitialized = true;
   }
 
-  isValidTheme(className: string): Observable<boolean> {
-    return this.getAll().pipe(
-      take(1),
-      map((themes) => {
-        return themes.some((theme) => theme.className === className);
-      })
+  isValidTheme(className: string): boolean {
```
pawelfras:
```suggestion
  isThemeValid(className: string): boolean {
```
or
```suggestion
  isValid(className: string): boolean {
```
Apart from that, this method is not used anywhere, but I guess it should be in `setActive` (check the comment below)

projects/core/src/site-theme/facade/site-theme.service.ts
```diff
@@ -48,26 +74,31 @@ export class SiteThemeService {
...
-      .subscribe(() => {
-        this.store.dispatch(new SiteThemeActions.SetActiveSiteTheme(className));
+    this.store
+      .pipe(select(SiteThemeSelectors.getActiveSiteTheme), take(1))
+      .subscribe((activeTheme: string | null) => {
+        if (activeTheme !== className) {
+          this.store.dispatch(
+            new SiteThemeActions.SetActiveSiteTheme(className)
+          );
+        }
```
pawelfras:
```suggestion
        if (this.isValidTheme(className) && activeTheme !== className) {
          this.store.dispatch(
            new SiteThemeActions.SetActiveSiteTheme(className)
          );
        }
```
`major:` I think the check if value is valid is still required before we call an action to set new active theme. Without this, we will still encounter the issue we discussed where action is can be triggered with the invalid value from local storage.

projects/core/src/site-theme/facade/site-theme.service.ts
Platonn: done

projects/core/src/site-theme/facade/site-theme.service.ts
Platonn: oh, good catch. thanks. done

pawelfras (approved):
LGTM. As we discussed offline, unit test adjustments are going to be provided separately

kimhw0630:
@Platonn 
I just checked out this branch and tested it before reviewing the code changes.
The default theme doesn't get set
1. Go to home page and select Dark/light theme.
2. Open devTools from Chrome -> Application -> search for spartacus⚿⚿siteTheme
Check the value for spartacus⚿⚿siteTheme (cx-theme-high-contrast-dark or cx-theme-high-contrast-light) 
3. Remove this key
4. Refresh the browser
5. Go to devTools again to check
spartacus⚿⚿siteTheme value should be “santorini” but spartacus⚿⚿siteTheme isn’t being created.

Also, we can no longer switch back to the default theme after selecting a different one

pawelfras:
> @Platonn I just checked out this branch and tested it before reviewing the code changes. The default theme doesn't get set
> 
> 1. Go to home page and select Dark/light theme.
> 2. Open devTools from Chrome -> Application -> search for spartacus⚿⚿siteTheme
>    Check the value for spartacus⚿⚿siteTheme (cx-theme-high-contrast-dark or cx-theme-high-contrast-light)
> 3. Remove this key
> 4. Refresh the browser
> 5. Go to devTools again to check
>    spartacus⚿⚿siteTheme value should be “santorini” but spartacus⚿⚿siteTheme isn’t being created.
> 
> Also, we can no longer switch back to the default theme after selecting a different one

Good catch. It seems to me that both problems are caused by the lack of a default value of `THEME_CONTEXT_ID`. We could consider adding `santorini` as a default value to the `default-site-context-config.ts`.
**EDIT:** Only now I recalled that we don't need to set `santorini` CSS class to switch back to the default theme as it is default styling for Spartacus which doesn't require setting additional class:
image
Removing classes related to high-contrast themes should be enough.

kimhw0630 (approved):

pawelfras (approved):

PR #19172, opened 2024-08-28
chore: Fix added to cart e2e test interceptor

rmch91:


pawelfras (approved):
The interceptor query had to be changed to a more precise one due to the additional multi-d-related API call. LGTM. QA done

PR #19176, opened 2024-08-28
feat/CXSPA-8061: S/4HANA Service Order UI - Alter Delivery Mode Tab in Checkout

anjana-bl:
https://jira.tools.sap/browse/CXSPA-8061


integration-libs/s4-service/order/components/order-summary/service-details-card.module.ts
```diff
@@ -10,10 +10,17 @@ import { CommonModule } from '@angular/common';
...
+import { ServiceOrderOverviewComponentService } from './service-order-overview-component.service';
+import { OrderOverviewComponentService } from '@spartacus/order/components';
 
 @NgModule({
   imports: [CardModule, CommonModule, I18nModule],
   providers: [
+    ServiceOrderOverviewComponentService,
+    {
+      provide: OrderOverviewComponentService,
+      useClass: ServiceOrderOverviewComponentService,
```
Platonn: [MINOR]
In general, let's use `useExisting` instead of `useClass`, for more straightforward extensibility for customers.

Thanks to this, it doesn't matter if they over-provide:
- `provide: OrderOverviewComponentService` OR
- `provide: ServiceOrderOverviewComponentService`
.. their cutomization will work.

Otherwise (if you leave `useClass`), a customer can only customize the service, by:
- `provide: OrderOverviewComponentService`

feature-libs/order/components/order-details/order-overview/order-overview-component.service.ts
```diff
@@ -0,0 +1,15 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { Injectable } from '@angular/core';
+import { DeliveryMode } from '@spartacus/cart/base/root';
+
+@Injectable()
+export class OrderOverviewComponentService {
+  showDeliveryMode(mode: DeliveryMode | undefined): boolean {
```
Platonn: [MEDIUM]
Let's use `shouldShowDeliveryMode` instead `showDeliveryMode`.
Otherwise, the method name suggest making a side-effectful action (showing something). But in reality it's a boolean indicator _whether_ to show or not.

PS. An analogical comment applies to more places in this PR. Please fix all of them.

integration-libs/s4-service/root/config/default-service-delivery-mode-config.ts
```diff
@@ -0,0 +1,22 @@
...
+ * SPDX-FileCopyrightText: 2024 SAP Spartacus team <spartacus-team@sap.com>
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { PriceType } from '@spartacus/core';
+import { ServiceDeliveryModeConfig } from '../model/augmented-types.model';
+
+export const defaultServiceDeliveryModeConfig: ServiceDeliveryModeConfig = {
+  serviceDeliveryMode: {
```
Platonn: [QUESTION]
`serviceDeliveryMode` sounds like quite a general term to me. Perhaps it's only me.
That said, it's a key in the global config. In general, it would be good to prefix such keys so they are bound to the specific feature (feature library). Do you think that putting `s4Service` prefix to the global config key would be beneficial for understanding its relationship with the `s4-service` feature library? And would it make it less prone to be confused with other possible features' global config keys?

In gerenal, I'd suggest such an approach, but maybe you have good arguments against it.

I mean e.g.:
```suggestion
export const defaultS4ServiceDeliveryModeConfig: S4ServiceDeliveryModeConfig = {
  s4serviceDeliveryMode: {
```

I'm happy to learn your point of view.

anjana-bl: it makes more sense to append s4 in the front. Have made the changes !!!

feature-libs/order/components/order-details/order-overview/order-overview-component.service.ts
anjana-bl: done!

integration-libs/s4-service/order/components/order-summary/service-details-card.module.ts
anjana-bl: understood, done !!

Platonn (approved):
Code LGTM with scope: shared / global things / architecture / standards

Please review and QA the feature details specifically within the Team Fusion. Thanks in advance!

PR #19183, opened 2024-08-29
fix: move price range atr to default product search occ endpoint

kpawelczak:
closes: [CXSPA-8104](https://jira.tools.sap/browse/CXSPA-8104)


pawelfras (approved):
LGTM. QA done

PR #19185, opened 2024-08-29
refactor: OPF sub-libs reorganization

Matejk00:
Closes: CXSPA-7781


core-libs/setup/tsconfig.spec.json
```diff
@@ -637,22 +637,50 @@
...
       "@spartacus/opf/checkout": [
         "../../integration-libs/opf/checkout/public_api"
       ],
-      "@spartacus/opf/checkout/occ": [
-        "../../integration-libs/opf/checkout/occ/public_api"
-      ],
       "@spartacus/opf/checkout/root": [
         "../../integration-libs/opf/checkout/root/public_api"
       ],
+      "@spartacus/opf/cta/core/components": [
```
Platonn: [MEDIUM]
Usually `/components` is a sibling to `/core`.
Here it's nested under `/core`. Was it a deliberate decision or a "typo"?

integration-libs/opf/base/occ/config/default-occ-opf-base-config.ts
```diff
@@ -6,12 +6,11 @@
 
 import { OccConfig } from '@spartacus/core';
 
-export const defaultOccOpfConfig: OccConfig = {
+export const defaultOccOpfBaseConfig: OccConfig = {
   backend: {
     occ: {
       endpoints: {
         getActiveConfigurations: 'active-configurations',
```
Platonn: [MAJOR]
Endpoints that are not present in the OCC backend (but only in the Workbench backend), should not be stored at the `backend.occ` key, but at the `backend.opf` key.
Likewise, the Adapters calling those endpoints (e.g. `/active-configuerations`) should not be prefixed with `Occ` (e.g. `OccOpfAdapter`), because they have nothing to do with Occ.
Similarly, there code for calling those endpoints should not be stored at the `/occ` folder (and sub-library name), but a different name.

TBH I'm having hard time proposing a different name, because I'm lacking knowledge on the nature of the OPF backend. What's its name? `OPF backend`? `OPF Workbench backend`? ...? When we come up with the _name of the backend_, we should use it consistently for places referring to this backend, instead of the term `OCC`.

Note: As far as I can udnerstand, those endpoints are for example `/active-configurations`, `/payments/get-cta-scripts` etc. - they are not OCC endpoints.
But `/cart/accessCode` and `placePaymentAuthorizedOrder` are OCC Commerce's endpoints (so their logic is accurate being put within OCC-prefixed places.

core-libs/setup/tsconfig.spec.json
Matejk00: It was a "typo". I moved "components" folder along the "core" and moved `OpfCtaComponentsModule` import into opf-cta.module.ts

integration-libs/opf/base/occ/config/default-occ-opf-base-config.ts
Matejk00: I clarified with the Bamboo team that all endpoints in a Commerce Cloud Adapter will only affect OPF functionality and they will be never a part of the OCC.

So my proposal is to change a name into:
```
    "opf-api" {
      endpoints: {
        getActiveConfigurations: 'active-configurations',
```

and the same for the folder "opf-api"
We can also use such prefix "OpfApi..." for all files inside the folder, replacing the "Occ" part. What do you think?

integration-libs/opf/base/root/opf-base-root.module.ts
```diff
@@ -5,76 +5,26 @@
...
-          cxRoute: 'paymentVerificationCancel',
-        },
-      },
-    ]),
-    OpfEventModule,
-  ],
+  imports: [OpfEventModule],
   providers: [
     {
       provide: MODULE_INITIALIZER,
```
Platonn: [MEDIUM]
Let's use APP_INITIALIZER, becasue it's not a lazy-laodable module/logic

Matejk00: done

Platonn (approved):

PR #19197, opened 2024-09-02
feat: CXSPA-8273 SSR Error Handling - E2E tests

pawelfras:
This PR provides E2E for SSR Error handling and some improvements for a framework for SSR E2Es.

To run tests, run the following commands:
- `npm run build:libs`
- `npm run build:ssr:local-http-backend`
- `npm run test:ssr`

closes [CXSPA-8273](https://jira.tools.sap/browse/CXSPA-8273)


projects/ssr-tests/src/ssr-testing.spec.ts
```diff
@@ -1,54 +1,126 @@
...
+import { expectLogMessages } from './matchers/matchers';
+import * as LogUtils from './utils/log.utils';
+import * as ProxyUtils from './utils/proxy.utils';
+import * as SsrUtils from './utils/ssr.utils';
 
 const BACKEND_BASE_URL: string = process.env.CX_BASE_URL || '';
 
 describe('SSR E2E', () => {
-  let proxy: any;
+  let proxy: Server;
```
Platonn: [MINOR]
To avoid ambiguity what we're proxying (ssr server vs backend), what about a rename
`proxy` -> `backendProxy`
?

projects/ssr-tests/src/ssr-testing.spec.ts
```diff
@@ -1,54 +1,126 @@
...
+      beforeEach(async () => {
+        await SsrUtils.startSsrServer();
+      });
+
+      it('should receive success response with request', async () => {
+        proxy = await ProxyUtils.startBackendProxyServer({
+          target: BACKEND_BASE_URL,
+        });
+        const response: any =
+          await ProxyUtils.sendRequestToSsrServer(REQUEST_PATH);
```
Platonn: [MINOR]
It's very confusing we're using `ProxyUtils` for sending a request to Ssr server.
Shouldn't it be simply kinda "HttpUtils`?

pawelfras: Are you suggesting only renaming `ProxyUtils` or also moving `sendRequestToSsrServer` outsde `proxy.utils.ts` to a separate file, e.g. `http.utils.ts`?
Maybe we can consider renaming `proxy.utils.ts` -> `http.utils.ts` because IMO function responsible for starting proxy server can be also under `HttpUtils`.

projects/ssr-tests/src/ssr-testing.spec.ts
```diff
@@ -1,54 +1,126 @@
...
+        expect(response.statusCode).toEqual(200);
+
+        // Rendering should not complete in the first request.
```
Platonn: [QUESTION]
Why? Because our `timeout` in `SsrOptimizationOptions` is only 3000s?
If yes, shouldn't we just increase the default `timeout` for e2e tests?

projects/ssr-tests/src/ssr-testing.spec.ts
```diff
@@ -1,54 +1,126 @@
...
+      });
+
+      it('should receive response with 404 when page does not exist', async () => {
+        proxy = await ProxyUtils.startBackendProxyServer({
+          target: BACKEND_BASE_URL,
+        });
+        const response = await ProxyUtils.sendRequestToSsrServer(
+          REQUEST_PATH + 'not-existing-page'
+        );
+        expect(response.statusCode).toEqual(404);
```
Platonn, line 52: [MINOR]
What about adding an `expect` that we want to contain logs:
```
`Request is resolved with the SSR rendering error (${REQUEST_PATH + 'not-existing-page'})`
```
?

projects/ssr-tests/src/ssr-testing.spec.ts
pawelfras: That's a comment left by Caine, but it's actually a good question. I haven't noticed any issues with this by far so maybe the comment is not relevant anymore.

projects/ssr-tests/src/ssr-testing.spec.ts
```diff
@@ -1,54 +1,126 @@
...
+      });
+
+      it('should receive response with status 404 if HTTP error occurred when calling cms/pages API URL', async () => {
+        proxy = await ProxyUtils.startBackendProxyServer({
+          target: BACKEND_BASE_URL,
+          callback: (proxyRes, req) => {
+            if (req.url?.includes('cms/pages')) {
+              proxyRes.statusCode = 404;
+            }
```
Platonn, line 61: nice, looks very readable

projects/ssr-tests/src/ssr-testing.spec.ts
```diff
@@ -1,54 +1,126 @@
...
+          `Rendering started (${REQUEST_PATH})`,
+          `Request is waiting for the SSR rendering to complete (${REQUEST_PATH})`,
+        ]);
+
+        response = await ProxyUtils.sendRequestToSsrServer(REQUEST_PATH);
+        expect(response.statusCode).toEqual(200);
+        expectLogMessages().toContain(`Render from cache (${REQUEST_PATH})`);
+      });
+
+      it('should call rendering for the next request if previous render failed', async () => {
```
Platonn: [NITPICK]
What about:
```suggestion
      it('should render for the next request if previous render failed', async () => {
```

projects/ssr-tests/src/ssr-testing.spec.ts
pawelfras: For some reasons, tests ends before this exact log is generated and written to `.ssr.log` file from which we're taking log messages for assertion purposes. I haven't found a reason of this, but as it's rather framework issue, I don't want to spend more time on it. IMO the test can be adjusted later.

projects/ssr-tests/src/ssr-testing.spec.ts
pawelfras:
```suggestion
      it('should render the next request if previous render failed', async () => {
```

projects/ssr-tests/src/utils/proxy.utils.ts
```diff
@@ -0,0 +1,107 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import * as http from 'http';
+import httpProxy, { ProxyResCallback } from 'http-proxy';
+
+const proxy = httpProxy.createProxyServer({
+  secure: false,
+  // selfHandleResponse: true,
+});
```
Platonn: [MEDIUM]
To avoid being prone to interferences between tests, I suggest to move it to the `startBackendProxyServer`.
Otherwise, multiple calls of this function operate on the same global shared object `proxy`

projects/ssr-tests/src/ssr-testing.spec.ts
Platonn: I'd like to distinguish 3 entities:
- SSR server - that you can `start` (with config), `kill`
- Backend proxy - that you can `start` (with config), and `kill`... ~IMHO we should also kill it after each test to avoid memory leaks in tests? ;) I guess so... - perhaps `close`ing both `httpProxy` server object and `http` server object~ (now I can see we close proxy)
- TestClient - that can `sendRequest` (and get response)

projects/ssr-tests/src/ssr-testing.spec.ts
Platonn: I guess on CI the execution can be much slower and there will be no chance to meet 3000ms timeout there.

projects/storefrontapp/server.ts
```diff
@@ -23,6 +23,7 @@ const ssrOptions: SsrOptimizationOptions = {
   timeout: Number(
     process.env['SSR_TIMEOUT'] ?? defaultSsrOptimizationOptions.timeout
   ),
+  cache: process.env['SSR_CACHE'] === 'true', // Enable caching for SSR E2E purposes
```
Platonn: [MINOR]
It sounds like you'd like to enable caching for the purpose of running (all) E2Es.
btw. Do we really need this comment?

pawelfras: Correct, the comment could be more precise. But indeed, it's not required. I'd remove it 👌

projects/ssr-tests/src/utils/proxy.utils.ts
pawelfras: Good point

projects/ssr-tests/src/ssr-testing.spec.ts
pawelfras: Good point. In that case, I'm going to add another param to `startSsrServer` (with a default value of 10s? 20s?) so we can control the timeout.

projects/ssr-tests/src/utils/proxy.utils.ts
```diff
@@ -6,29 +6,9 @@
...
-
-/**
- * Response from SSR server.
- */
-export interface SsrResponse {
-  statusCode: number | undefined;
-  headers: http.IncomingHttpHeaders;
-  body: string;
-}
+let proxy: httpProxy;
```
Platonn: Why didn't we move the definition of the const to the internals of the function `startBackendProxyServer()`?
Does it really have to be global?

Platonn (approved):

PR #19209, opened 2024-09-05
fix: CXMC-23462 Page display incorrectly when token invalid

i53577:
a property is needed in backend. 
```
corsfilter.commercewebservices.exposedHeaders=x-anonymous-consents occ-personalization-id occ-personalization-time www-authenticate
```

this property not locate in any extension in MCS, but a sample property in cxmc-product-pipeline.
https://github.tools.sap/cx-commerce/cxmc-product-pipeline/blob/develop/galectin/resources/tasks/local.properties.tmpl#L37


projects/core/src/auth/user-auth/http-interceptors/auth.interceptor.ts
```diff
@@ -65,7 +65,8 @@ export class AuthInterceptor implements HttpInterceptor {
                 } else if (
                   // Refresh the expired token
                   // Check if the OAuth endpoint was called and the error is because the refresh token expired
-                  this.errorIsInvalidToken(errResponse)
+                  this.errorIsInvalidToken(errResponse) ||
+                  this.isOidcInvlidToken(errResponse)
```
scarai-sap: maybe insert logic into errorIsInvalidToken?

projects/core/src/auth/user-auth/http-interceptors/auth.interceptor.spec.ts
```diff
@@ -40,6 +41,9 @@ class MockAuthConfigService implements Partial<AuthConfigService> {
   getTokenEndpoint() {
     return '/authorizationserver/token';
   }
+  public getOAuthLibConfig(): AuthLibConfig {
```
scarai-sap: looks the new code is not tested by UT?

integration-libs/cdc/root/service/cdc-auth-http-header.service.ts
```diff
@@ -0,0 +1,32 @@
...
+  providedIn: 'root',
+})
+export class CdcAuthHttpHeaderService extends AuthHttpHeaderService {
+  protected cdcAuthLogOutService = inject(CdcAuthLogOutService);
+  /**
+   * override handleExpiredRefreshToken() method
+   */
+  public handleExpiredRefreshToken(): void {
+    this.authRedirectService.saveCurrentNavigationUrl();
+    this.cdcAuthLogOutService.cdcLogout().finally(() => {
```
scarai-sap: why we need call logout?

scarai-sap (commented):
question raised

integration-libs/cdc/root/service/cdc-auth-http-header.service.ts
i53577: in AuthHttpHeaderService， here calls `this.authService.coreLogout()`, this method won't logout cdc customer. So we need to replace it with `this.cdcAuthLogOutService.cdcLogout()`

projects/core/src/auth/user-auth/http-interceptors/auth.interceptor.spec.ts
i53577: UT added

projects/core/src/auth/user-auth/http-interceptors/auth.interceptor.ts
i53577: if someday we decide to keep spartacus for a single branch, then seems split it to a function is more easy to do adaptation :)

scarai-sap: as you already have judgement as below so i guess it's safe to put them into one routine? BTW. Does it only for OIDC(I just wonder whether it's best name for the logic)? 

 if (!this.authConfigService.getOAuthLibConfig().disablePKCE && authHeader)

integration-libs/cdc/root/guards/cdc-auth-logout.service.ts
```diff
@@ -0,0 +1,17 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { Injectable } from '@angular/core';
+import { CdcLogoutGuard } from './cdc-logout.guard';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class CdcAuthLogOutService extends CdcLogoutGuard {
```
vinchen666: a service should not be a child class of a guard, guard should relies on service. let's follow **composition over Inheritance** principle

i53577: Adapted code to let this service not extends a guard. Please have a check.

projects/core/src/auth/user-auth/http-interceptors/auth.interceptor.ts
i53577: moved logic to errorIsInvalidToken

scarai-sap (approved):
LGTM

vinchen666 (approved):

PR #19222, opened 2024-09-10
feat(a11y): Redesign searchbox to include label

sdrozdsap:
Ticket: https://jira.tools.sap/browse/CXSPA-7990


projects/storefrontstyles/scss/components/product/search/_searchbox.scss
```diff
@@ -52,12 +52,35 @@
     @include media-breakpoint-down(sm) {
       // hide the input on mobile when there's no interaction with searchbox
       cx-searchbox {
-        input {
-          // we cannot use display:none, visible:hidden or opacity: 0
-          // as this will no longer emit a focus event to the controller logic
-          width: 0;
-          padding: 0;
+        // cxFeat_a11ySearchboxLabel class is only applied if a11ySearchboxLabel flag is true
```
Zeyber, line 73: Thinking about this one. Arguments against this approach could be:
- `cxFeat_a11ySearchboxLabel` is being publicly exposed. Users may decide to override this when enabling the feature flag as they will not read these comments.
- This approach is unusual and a bit difficult to consider all possibilities. It will also be a bit difficult for anyone who needs to clean it up as its a non-standard approach. Could you please elaborate more on the issues it resolves?

Zeyber (changes requested):

projects/storefrontstyles/scss/components/product/search/_searchbox.scss
sdrozdsap: I agree with you: the points that you mention are solid arguments against this, but at the moment I didn't have any other thought in my mind. The main reason why I dived in this workaround, is because this component applies styles to the `<body>` element, which is impossible to do with the `forFeature` mixin. Other reason, is that there are big styling changes, and following the standard procedure with the `forFeature` mixin didn't seem doable. Its assumed that with this mixin we would be always able to override the styles if feature flag is enabled, but in this case it would be very nice to have ability to apply styles for when this flag is NOT enabled. Otherwise I need a way to "reset" a lot of styles, that would be applied if the feature flag is disabled(you can take a look at `projects/storefrontstyles/scss/components/product/search/_searchbox.scss` starting line 141 - there are all styles which I don't need). One more point worth mentioning is that nested `forFeature` don't work, but I would end up having them if I would go with this approach.

Considering your points and my thought above, I came up with an idea, but probably it's not a good one: we can create a new component, copying all the functionality from the old one, but having all new styles and a11y adjustments. Then we can decide depending on whether the feature flag is enabled or not, which component to use.
Please let me know what you think

Zeyber: Thank you for clarifying the reasons you took this approach. I think that my concerns around clean up and confusion are not weighted enough to justify rethinking another approach considering the complexity of working something out and the scenario of any major issues coming out of it are very low IMHO. Thank you for making these considerations :)

Zeyber (dismissed):

developpeurweb (dismissed):

Zeyber (approved):

PR #19223, opened 2024-09-10
feat: Add iframe support

Matejk00:
Closes: CXSPA-8085

### DIFF
--- a/integration-libs/opf/checkout/components/opf-checkout-payment-wrapper/opf-checkout-payment-wrapper.component.html
+++ b/integration-libs/opf/checkout/components/opf-checkout-payment-wrapper/opf-checkout-payment-wrapper.component.html
@@ -28,7 +28,9 @@
     >
       <iframe
         class="cx-payment-iframe"
-        *ngIf="!paymentLink.destination?.form?.length"
+        *ngIf="
+          !paymentLink.destination?.form?.length && paymentLink.destination?.url
+        "
         [src]="renderUrl(paymentLink.destination?.url)"
       ></iframe>
 
--- a/integration-libs/opf/checkout/components/opf-checkout-payment-wrapper/opf-checkout-payment-wrapper.component.html
+++ b/integration-libs/opf/checkout/components/opf-checkout-payment-wrapper/opf-checkout-payment-wrapper.component.html
@@ -14,15 +14,23 @@
 <ng-template #paymentData let-paymentLink="paymentLink">
   <div>
     <div
-      *ngIf="paymentLink.renderType === RENDER_TYPES.DESTINATION"
+      *ngIf="paymentLink.renderType === RENDER_PATTERN.FULL_PAGE"
       class="cx-payment-link"
     >
-      <a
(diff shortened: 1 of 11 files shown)


integration-libs/opf/checkout/components/opf-checkout-payment-wrapper/opf-checkout-payment-wrapper.component.html
```diff
@@ -14,15 +14,23 @@
...
+      }}</a>
+    </div>
+    <ng-container
+      *ngIf="
+        paymentLink.data && paymentLink.renderType === RENDER_PATTERN.IFRAME
+      "
+    >
+      <iframe
+        class="cx-payment-iframe"
         *ngIf="!paymentLink.destination?.form?.length"
```
FollowTheFlo: should we also check url not null (paymentLink.destination?.url) in condition? or iframe tag is smart enough to handle empty value?

integration-libs/opf/checkout/styles/components/_opf-checkout-payment-wrapper.scss
```diff
@@ -11,5 +11,10 @@
     .cx-payment-link {
       text-align: end;
     }
+    .cx-payment-iframe {
+      min-height: 400px;
```
FollowTheFlo, line 15: just curiosity how was this min-height value 400px decided? i supposed it is standard?

integration-libs/opf/checkout/components/opf-checkout-payment-wrapper/opf-checkout-payment-wrapper.component.html
Matejk00: I added additional check, to not render an empty iframe.

integration-libs/opf/checkout/components/opf-checkout-payment-wrapper/opf-checkout-payment-wrapper.component.html
```diff
@@ -14,15 +14,25 @@
...
-        [href]="paymentLink.data"
-        >{{ 'opf.checkout.proceedPayment' | cxTranslate }}</a
-      >
+      <a class="btn btn-primary" [href]="paymentLink.data">{{
+        'opf.checkout.proceedPayment' | cxTranslate
+      }}</a>
+    </div>
+    <ng-container
+      *ngIf="
+        paymentLink.data && paymentLink.renderType === RENDER_PATTERN.IFRAME
```
FollowTheFlo, line 26: do we also need to cover the use case where renderType is IFRAME and based on script (not url). In my understanding it was the use case we were covering before this fix.

integration-libs/opf/checkout/styles/components/_opf-checkout-payment-wrapper.scss
FollowTheFlo: as discussed, new jira to handle it as need more analysis and opinions.

PR #19229, opened 2024-09-11
fix: prevent `CmsService.getComponentData()` invoked with an empty string `uid` from unnecesarily calling Ngrx and Backend

Platonn:
Previously:
When the `CmsService.getComponentData()` was invoked with an empty string `uid` argument, then it called Backend (and also Ngrx along the way), which resulted in a Failure Ngrx action. And it emitted `null` value. 

See `componentIds=` in the backend URL in the screenshot below:
image

Observations:
- It doesn't make sense to make a call to Backend with and empty `uid` - we want to prevent it.
- The Failure Ngrx action happening during SSR, made the SSR of PDP fail (return 500) - which breaks SEO - so we want to prevent it.

Fix:
Now we validate the `uid` argument of `CmsService.getComponentData()` and emit `null` immediately if the `uid` is empty. In such case, we don't call the backend anymore (would caused getting empty response from Backend and a failure from Ngrx).

fixes https://jira.tools.sap/browse/CXSPA-8389

### DIFF
--- a/projects/core/src/cms/facade/cms.service.spec.ts
+++ b/projects/core/src/cms/facade/cms.service.spec.ts
@@ -73,6 +73,16 @@ describe('CmsService', () => {
 
   describe('getComponentData', () => {
     describe('when pageContext is NOT provided', () => {
+      describe('when uid is empty', () => {
+        it('should emit null', inject([CmsService], (service: CmsService) => {
+          let result;
+          service.getComponentData('').subscribe((res) => {
+            result = res;
+          });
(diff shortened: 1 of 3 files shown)


Platonn:
Compared e2e result on local BEFORE vs AFTER the fix:

### BEFORE - fail
SSR logs:
image

Cypress:
image


### AFTER - pass
SSR logs:
image

Cypress:
image

pawelfras (approved):
LGTM. QA passed.

PR #19244, opened 2024-09-16
feat: Add a11y tab component to plp facets

Zeyber:


projects/storefrontlib/cms-components/product/product-list/product-facet-navigation/facet-list/facet-list.component.ts
```diff
@@ -94,6 +114,30 @@ export class FacetListComponent implements OnInit, OnDestroy {
     }
   }
 
+  ngAfterViewInit(): void {
+    this.renderFacets();
```
Zeyber: This method of rendering is not friendly and is a bit of a workaround. It has issues not only with facet implementation but also with quick order and most likely other future implementations. We should move this change detection logic to solve the issue into the tab component.

Zeyber:
Note: Best to recheck attributes on tab paragraph component are correct as `collapse` showed a value of `[Object]` in a demo.

developpeurweb:
@Zeyber so far so good, the only thing that I noticed is that two Tab key strokes are needed to get into the content panel, auditors have complained about that in other components. 
Is it possible to go to the first option in the panel on the first stroke but without losing the items count for the group? So, perhaps move the aria-label with the group's items count to the collapsible button.

projects/storefrontlib/cms-components/product/product-list/product-facet-navigation/facet-list/facet-list.component.html
```diff
@@ -26,22 +24,43 @@ <h4>
...
-      'productFacetNavigation.ariaLabelItemsAvailable'
-        | cxTranslate
-          : {
-              name: facet.name,
-              count: facet?.values?.length,
-            }
-    }}"
-  ></cx-facet>
+  <ng-container *cxFeature="'a11yTabComponent'">
+    <cx-tab [tabs$]="tabs$" [config]="tabConfig" *ngIf="tabs$ | async"></cx-tab>
```
sdrozdsap: The `tabs$` is being subscribed by inside the `<cx-tab>` component, I think having this second subscription is redundant(`tabs$ | async`) . Do you think it will work if you just pass `*ngIf="tabs$"`

projects/storefrontlib/cms-components/product/product-list/product-facet-navigation/facet-list/facet-list.component.ts
```diff
@@ -73,6 +78,16 @@ export class FacetListComponent implements OnInit, OnDestroy {
     autofocus: 'cx-facet > button',
   };
 
+  tabConfig: TabConfig = {
+    label: 'Product Facets',
```
sdrozdsap: Maybe we can also internationalise that?

sdrozdsap (dismissed):

projects/storefrontlib/cms-components/product/product-list/product-facet-navigation/facet-list/facet-list.component.html
Zeyber: Yup, seems to work without *ngIf at all

sdrozdsap (dismissed):

sdrozdsap (dismissed):

sdrozdsap (approved):

Zeyber:
**Note for testing**
Feature is disabled by default while we work out some issues. To enable it, you need to open `spartacus-features.module.ts` and set `a11yTabComponent` to `true`.

developpeurweb (approved):

PR #19252, opened 2024-09-17
fix: Sec Val. - additional password validators

kpawelczak:
closes: [CXSPA-6737](https://jira.tools.sap/browse/CXSPA-6737)


feature-libs/user/profile/assets/translations/en/userProfile.json
```diff
@@ -52,7 +52,7 @@
     "emailMarketing": "Use my personal data to receive e-mail newsletters for marketing campaigns. To change your settings, go to Consent Management in My Account.",
     "confirmThatRead": "I am confirming that I have read and agreed with the",
     "selectTitle": "Select Title",
-    "passwordMinRequirements": "Here's a quick guide for a secure password: Use at least 8 characters with a mix of upper and lowercase letters, numbers, and symbols.",
+    "passwordMinRequirements": "Here's a quick guide for a secure password: Use at least 8 characters with a mix of upper and lowercase letters, numbers, and symbols. Make sure not to use the same character consecutively.",
```
kimhw0630: This is a breaking change. One suggestion is to create a new key and use that instead.

projects/assets/src/translations/en/myAccount.json
```diff
@@ -16,7 +16,7 @@
       "label": "New Password",
       "placeholder": "New Password"
     },
-    "passwordMinRequirements": "Here's a quick guide for a secure password: Use at least 8 characters with a mix of upper and lowercase letters, numbers, and symbols.",
+    "passwordMinRequirements": "Here's a quick guide for a secure password: Use at least 8 characters with a mix of upper and lowercase letters, numbers, and symbols. Make sure not to use the same character consecutively.",
```
kimhw0630: same here
This is a breaking change. One suggestion is to create a new key and use that instead.

feature-libs/user/profile/assets/translations/en/userProfile.json
kpawelczak: thanks, I've created a new translation key

projects/assets/src/translations/en/myAccount.json
kpawelczak: thanks I've created a new translation key

projects/core/src/util/regex-pattern.ts
```diff
@@ -19,3 +19,8 @@ export const MIN_ONE_SPECIAL_CHARACTER_PATTERN =
   /^(?=.*?[!@#$%^*()_\-+{};:.,]).*$/;
 
 export const MIN_SIX_CHARACTERS_PATTERN = /^.{6,}$/;
+
+export const CONSECUTIVE_CHARACTERS = /(.)\1+/;
+
+export const STRONG_PASSWORD_PATTERN =
+  /^(?!.*(.)\1)(?=.*?[A-Z])(?=.*?\d)(?=.*?[!@#$%^*()_\-+{};:.,]).{6,}$/;
```
kimhw0630, line 26: By checking this regex
No two consecutive identical characters.
At least one uppercase letter.
At least one digit.
At least one special character from the allowed set.
Minimum length of 6 characters

Does this check match the backend validation? As far as I understand, the validation should match the backend validation.

Limited special characters: Only a small set of special characters is allowed. (?=.*?[!@#$%^*()_\-+{};:.,]). is this correct?

projects/storefrontlib/shared/utils/validators/custom-form-validators.ts
```diff
@@ -58,6 +60,29 @@ export class CustomFormValidators {
...
+   * @deprecated Use passwordValidators instead
+   * @static
+   * @param {AbstractControl} control
+   * @returns {(ValidationErrors | null)} Uses 'cxInvalidPassword' validator error
+   * @memberof CustomFormValidators
+   */
+  static strongPasswordValidator(
+    control: AbstractControl
+  ): ValidationErrors | null {
+    const password = control.value as string;
```
kimhw0630, line 78: If the value might not always be a string (the form control value could be null or undefined), it's better to check the type or handle non-string cases gracefully? if this method get call after null check then might be fine.

projects/core/src/util/regex-pattern.ts
kpawelczak: I'm not sure if the validation should align with the backend requirements. If you take a look at the ticket comments we agreed that the backend would do the proper validation and the frontend will only validate simple checks for UX reasons.

However, I don't believe the backend has been updated to validate the new rule - > consecutive character

Limited special characters -> tbh I haven't touched it, what are you proposing?

projects/storefrontlib/shared/utils/validators/custom-form-validators.ts
kpawelczak: It's the same as every other custom validator. In the case of a password, it should always be a string and will always include a required validator to check for null or undefined values.

projects/core/src/util/regex-pattern.ts
kimhw0630: Limited special characters is a pattern from line 11, so I believe we should be fine.  thanks

kimhw0630 (dismissed):

kimhw0630 (approved):

PR #19259, opened 2024-09-18
fix: context.error was not printed because JSON.stringify(Error) doesn't work by default in JS

Platonn:
**Changes:**
- While stringifying the logged JSON, now we're converting each property being an Error instance into a pretty string. Otherwise, by default JSON.stringify would output by default only `{}`.
- Increased the `depth` option for `formatWithOptions()` util.  Otherwise, the Errors's properties at nested levels higher than 2 (e.g. `error.cause.headers`, where `cause` is `HttpErrorResponse`) were not be visible (e.g. abbreviated as `Map`) in the logs.

**QA steps:**
1. in app.module.ts provide the following config chunk (causing HTTP call to backend to fail):
   ```ts
    provideConfig({
      backend: {
        occ: {
          endpoints: {
            components: 'some-fake-endpoint',
          },
        },
      },
    }),
   ```
2. start SSR server - run `npm run dev:ssr` 
3. in another terminal `curl http://localhost:4200`
4. In the logs of the SSR server, see that under the message `"message": "Request is resolved with the SSR rendering error (/)` there is an property "error" that is not `{}`, but an actual string with Error details

BEFORE:
image

AFTER:
image

Future outlook: in another PR we plan to improve the readability of the error messages by displaying them in multi-line (instead of a single line with `\\n`).

fixes https://jira.tools.sap/browse/CXSPA-8360


core-libs/setup/ssr/logger/get-logger-inspect-options.ts
```diff
@@ -0,0 +1,54 @@
...
+ */
+const DEFAULT_LOGGER_INSPECT_OPTIONS: InspectOptions = {
+  /**
+   * Prevent the depth of the logged object to be limited to 2 (which is a NodeJS default).
+   * Otherwise, the object's properties at nested levels higher than 2 would not be visible in the logs.
+   * e.g. `{ a: { b: { c: { d: 'e' } } } }` would be presented as `{ a: { b: { c: [Object] } } }` in the logs.
+   *
+   * The value 10 was chosen arbitrarily. It can be adjusted in the future, if needed.
+   */
+  depth: 10,
```
pawelfras, line 18: `query:` Shouldn't we consider giving a lower default value? I wonder from the angle of the amount of logs and the app performance.

pawelfras (dismissed):
LGTM. QA done and errors are present in logs for both, dev and prod envs.
Screenshot 2024-09-19 at 15 23 26
Screenshot 2024-09-19 at 15 22 19

tools/config/manage-dependencies.ts
```diff
@@ -378,6 +384,9 @@ function filterNativeNodeAPIs(
     'worker_threads',
     'zlib',
   ];
+  // Node apis might be possibly prefixed with `node:`, e.g. `node:fs`
+  const prefixedNodeAPIs = nodeAPIs.map((nodeAPI) => `node:${nodeAPI}`);
```
pawelfras, line 388: :smart2:

core-libs/setup/ssr/logger/get-logger-inspect-options.ts
Platonn: TBH I had hard time deciding on the arbitrary value.
- 2 (default) seemed too low. especially for `logObject.context.error.cause.headers` it caused `headers` to be displayed as `Map`
- null/Infinity seemed unnecessarily risky in prod mode, if someone accidentally did log a super big object
- When reading logs, especially trying to understand a reason of a system failure, what matters is having the full picture of what happened. It would be a pity, if important details are abbreviated
That's why I've guessed the number 10, which is generous enough (subjectively IMHO), and also prevents too deep inspection levels.

pawelfras: Thank you for your explanation. I do agree with the reasoning.
In that case, maybe we should consider making this API public So customers can use it in their custom loggers and override the `depth` value if needed.

Platonn: good point. Customers might want to re-use this function `getLoggerInspectOptions()` in their `CustomExpressServerLogger`s.

pawelfras (approved):

PR #19266, opened 2024-09-19
fix: in dev mode pretty-print multi-line strings in logs

Platonn:
Now in dev mode we're using `formatWithOptions()` instead of `JSON.stringify()`. So now the output in the console in dev mode is more human-readable (including multi-line strings).

**QA steps**:
1. in app.module.ts provide the following config chunk (causing HTTP call to backend to fail):
   ```ts
    provideConfig({
      backend: {
        occ: {
          endpoints: {
            components: 'some-fake-endpoint',
          },
        },
      },
    }),
   ```
2. start SSR server - run `npm run dev:ssr` 
3. in another terminal `curl http://localhost:4200`
4. In the logs of the SSR server, see the error message is now multi-line human-readable format. Also function's definitions in the `[spartacus] SSR optimization engine initialized` are also multi-line.

BEFORE (dev mode only screenshots):
image

image

image

AFTER:

a) DEV MODE:
image

image

image

b) PROD MODE:
image

image

image

fixes https://jira.tools.sap/browse/CXSPA-8385


pawelfras (approved):
LGTM. QA done.

PR #19271, opened 2024-09-23
test(ssr e2e): redirect also the STDERR stream to the file `.ssr.log`

Platonn:
Otherwise error logs could be not asserted in the e2e. It will be needed in the future tests of logs content in cases when SSR fails.

BEFORE:
image

AFTER:
image

fixes https://jira.tools.sap/browse/CXSPA-8501

### DIFF
--- a/projects/ssr-tests/src/utils/ssr.utils.ts
+++ b/projects/ssr-tests/src/utils/ssr.utils.ts
@@ -51,7 +51,7 @@ export async function startSsrServer({
   timeout = DEFAULT_SSR_TIMEOUT,
 }: SsrServerOptions = {}) {
   child = childProcess.spawn(
-    `NODE_TLS_REJECT_UNAUTHORIZED=0 SSR_CACHE=${cache} SSR_TIMEOUT=${timeout} PORT=${port} npm run serve:ssr --prefix ../../> .ssr.log`,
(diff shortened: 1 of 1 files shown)


projects/ssr-tests/src/utils/ssr.utils.ts
```diff
@@ -51,7 +51,7 @@ export async function startSsrServer({
   timeout = DEFAULT_SSR_TIMEOUT,
 }: SsrServerOptions = {}) {
   child = childProcess.spawn(
-    `NODE_TLS_REJECT_UNAUTHORIZED=0 SSR_CACHE=${cache} SSR_TIMEOUT=${timeout} PORT=${port} npm run serve:ssr --prefix ../../> .ssr.log`,
+    `NODE_TLS_REJECT_UNAUTHORIZED=0 SSR_CACHE=${cache} SSR_TIMEOUT=${timeout} PORT=${port} npm run serve:ssr --prefix ../../> .ssr.log 2>&1`,
```
pawelfras, line 55: `minor:` What do you think about adding a description of what exactly `2>$1` does? It would be helpful to leave a reference for the future.

pawelfras (approved):
LGTM. QA done

PR #19276, opened 2024-09-23
fix: submit API invoked when add to cart or proceed to checkout

FollowTheFlo:
https://jira.tools.sap/browse/CXSPA-8490
submitPayment Observable was never unsubscribed, so getCartId() was still kicking when cart changed.
fix:
using  takeUserId and takeCartId which are non-stream obs (unlike getUserId and getCartId).
We are now sure observable gets completed

I think the issue came when we changed toPromise() into toLastValue().
Using toFirstValue() also fixes the problem but not sure about side effects it can produce

### DIFF
--- a/integration-libs/opf/payment/core/services/opf-payment-hosted-fields.service.ts
+++ b/integration-libs/opf/payment/core/services/opf-payment-hosted-fields.service.ts
@@ -20,16 +20,8 @@ import {
 } from '@spartacus/core';
 import { Order, OrderFacade } from '@spartacus/order/root';
 
-import { combineLatest, EMPTY, from, Observable, throwError } from 'rxjs';
-import {
-  catchError,
-  concatMap,
-  filter,
-  map,
-  switchMap,
-  take,
-  tap,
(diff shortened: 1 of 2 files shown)


integration-libs/opf/payment/core/services/opf-payment-hosted-fields.service.ts
```diff
@@ -237,4 +200,22 @@ export class OpfPaymentHostedFieldsService {
...
+    ]).pipe(
+      switchMap(([userId, activeCartId]: [string, string]) => {
+        submitRequest.cartId = activeCartId;
+        return combineLatest([
+          of(submitRequest),
+          this.cartAccessCodeFacade.getCartAccessCode(userId, activeCartId),
+        ]);
+      })
+    );
+  }
```
FollowTheFlo, line 220: Note that I removed the take(1) as takeUserId and takeActiveCartId already have 1. and getCartAccessCode is an ajax call which alwats get completed

Matejk00 (approved):

PR #19281, opened 2024-09-24
fix: enable by default meta description and canonical tag in CSR

kpawelczak:
closes: [CXSPA-8074](https://jira.tools.sap/browse/CXSPA-8074)


projects/core/src/features-config/feature-toggles/config/feature-toggles.ts
```diff
@@ -557,6 +557,14 @@ export interface FeatureTogglesInterface {
    * Previously, there the non-defined theme had a value `undefined`.
    */
   useSiteThemeService?: boolean;
+
+  /**
+   * When enabled, the default page meta configuration will apply by default
+   * a meta description and canonical URL for client-side rendered (CSR) pages.
+   * This makes sure that meta tags such as the description and the canonical URL
+   * are consistently applied to all pages that are rendered on the client side.
+   */
```
Platonn, line 566:
```suggestion
  /**
   * When enabled, the default page meta configuration will apply by default
   * a meta description and canonical URL for client-side rendered (CSR) pages.
   * This makes sure that meta tags such as the description and the canonical URL
   * are consistently applied to all pages that are rendered on the client side.
   * Previously, they were rendered only in SSR.
   */
```

PR #19293, opened 2024-09-26
feat: revert quick buy to the OPF codebase

Matejk00:
Closes: CXSPA-8172


integration-libs/opf/quick-buy/components/opf-quick-buy-buttons/google-pay/google-pay.service.ts
```diff
@@ -0,0 +1,496 @@
...
+  OpfQuickBuyLocation,
+  QuickBuyTransactionDetails,
+} from '@spartacus/opf/quick-buy/root';
+import { CurrentProductService } from '@spartacus/storefront';
+import { Observable, forkJoin, lastValueFrom, of } from 'rxjs';
+import { catchError, map, switchMap, take, tap } from 'rxjs/operators';
+import { OpfQuickBuyButtonsService } from '../opf-quick-buy-buttons.service';
+
+@Injectable({
+  providedIn: 'root',
```
FollowTheFlo, line 32: is it still in 'root' scope by design?

FollowTheFlo, line 32: is it still in 'root' scope by design?

integration-libs/opf/quick-buy/core/services/opf-quick-buy-transaction.service.ts
```diff
@@ -0,0 +1,177 @@
...
+    return this.activeCartFacade.takeActive();
+  }
+
+  getCurrentCartId(): Observable<string> {
+    return this.activeCartFacade.takeActiveCartId();
+  }
+
+  getCurrentCartTotalPrice(): Observable<number | undefined> {
+    return this.activeCartFacade
+      .getActive()
```
FollowTheFlo: getActive() is a stream that never complete, we should you takeActive() or this.getCurrentCart().

FollowTheFlo: getActive() is a stream that never complete, we should you takeActive() or this.getCurrentCart().

integration-libs/opf/quick-buy/core/services/opf-quick-buy-transaction.service.ts
```diff
@@ -0,0 +1,177 @@
...
+
+  getSupportedDeliveryModes(): Observable<DeliveryMode[]> {
+    return this.checkoutDeliveryModesFacade.getSupportedDeliveryModes();
+  }
+
+  setDeliveryAddress(address: Address): Observable<string> {
+    this.opfGlobalMessageService.disableGlobalMessage([
+      'addressForm.userAddressAddSuccess',
+    ]);
+    return this.checkoutDeliveryAddressFacade.createAndSetAddress(address).pipe(
```
FollowTheFlo, line 104: as an improvement we use createAndSetAddress if no addr yet, then use `PUT addresses/deliver` to update it, it would avoid the temp and stale addresses.
I remember trying that last year, but if occ was improved, we could give a try.

The other thing would be to append
`visibleInAddressBook:false`
in address object.
Same thing, I tried last year but had no impact, I reported it and got vague answers even from the team who refactored, OCC might have fixed that

projects/schematics/src/dependencies.json
```diff
@@ -424,6 +424,7 @@
     "@angular/platform-browser": "^17.0.5",
     "@angular/router": "^17.0.5",
     "@ng-select/ng-select": "^12.0.4",
+    "@ngrx/store": "^17.0.1",
```
FollowTheFlo, line 427: is it still needed?
I don't see any `import @ngrx/store` within opf folder

integration-libs/opf/quick-buy/components/opf-quick-buy-buttons/google-pay/google-pay.service.ts
Matejk00: Yes, it is as it was implemented in our previous code. We need this in root as these services corresponds in a native JS code and we would like to have it at very beginning.

Matejk00: Yes, it is as it was implemented in our previous code. We need this in root as these services corresponds in a native JS code and we would like to have it at very beginning.

integration-libs/opf/quick-buy/core/services/opf-quick-buy-transaction.service.ts
Matejk00: Thanks, I did a change.

Matejk00: Thanks, I did a change.

integration-libs/opf/quick-buy/core/services/opf-quick-buy-transaction.service.ts
Matejk00: We can spike this in another ticket. Current one was rather to revert a previous solution and refactor by organise a code along another sub-libs, but not dig deeper into functionality refactoring direction. Thanks for the suggestion and I will create a separate ticket for this.

projects/schematics/src/dependencies.json
Matejk00: Our `generate:deps` npm command adds this for OPF, so I can imagine that somehow we need it as a dependency.

FollowTheFlo (approved):
good to me as discuss, unit test coverage not sufficient to pass.

PR #19296, opened 2024-09-27
test: print SSR Logs as a context for failed tests to improve troubleshooting of SSR e2e

Platonn:
Wrapped each test function inside `it()` within an util function `LogUtils.attachLogsToErrors()`.
Whenever an error happens inside such a function (e.g. due to a failing `expect()`, the Error is intercepted (via try-catch) and the SSR logs are attached to this error's `cause` property. Then the error is re-thrown. Thanks to this, JestJS will report this `cause` along with the main error message.

Btw. made a refactor in various places related to logs:
- removed custom function `expectLogMessages()` and custom matcher `.toContainLogs` with regular `expect(logMessages).toContain()` - to not obfuscate under the hood what is being expected and how
- moved `LogUtils.clearSsrLogFile();` from top-level `beforeEach` to inside the util `SsrUtils.startSsrServer()` - to be less prone to refactoring mistake of re-ordering the execution of those events (i.e. clear log file VS. start new ssr writing to this file)
- created separate functions to get logs:
   - `getRawLogs(): string[]` - gets raw content of the log file (both JSONs and raw strings like 'Node is listening on port 4000'.
   - `getRawLogsPretty(): string[]` - similar as above, but it parses single line JSON strings and pretty-prints them (Note: it's used in `LogUtils.attachLogsToErrors()` for making the logs more readable when we attach them as a context of the error).
   - `getLogsObjects(): object[]` - gets only JSON logs and parses them to JS objects. Raw strings from logs are skipped
   - `getLogsMessages(): string[]` - gets only JSON logs' `message` properties. Raw strings from logs are skipped
   
**QA steps:**
1. Deliberately make one of the tests fail - e.g. change the expected HTTP status from `200` to `123`, e.g. in line 33, in the test `should receive success response with request`
1. Build libs, app and SSR (with localhost backend):
   `npm run build:libs && npm run build && build:ssr:local-http-backend`
2. Run ssr-tests:
   `npm run test:ssr`
3. Verify that the Error message printed by JestJS contains SSR logs as a context:
image
image

fixes https://jira.tools.sap/browse/CXSPA-8564


pawelfras (approved):

pawelfras (approved):

pawelfras (approved):
LGTM 🔥 QA Done

PR #19302, opened 2024-09-30
test(ssr-tests): print SSR Logs for errors in SSR tests, ALSO when a test times out

Platonn:
Previously, timed out tests didn't get the SSR logs printed. Now it's fixed.

To make it happen, we're changing an approach. Now, instead of wrapping each test callback of `it()`, we're setting up a global Jest Environment class, which intercepts the events of failed tests and attaches SSR logs to those error objects in a generic way.

**QA steps**
1. Prerequisites
   1. Deliberately make one test time out - e.g. add very short timeout `1_000` (1 second) timeout to the first test
   1. Deliberately make one of the tests fail - e.g. change the expected HTTP status from `200` to `123` to the second test
   
   See the example below:
   image

1. Build libs, app and SSR (with localhost backend):
   `npm run build:libs && npm run build && build:ssr:local-http-backend`
3. Run ssr-tests:
   `npm run test:ssr`
4. Verify that BOTH Error messages printed by JestJS contain SSR logs as a context (both the test that timed our and the test that had a failed `expect()`):

BEFORE: (screenshot just for the timed out test)

image

AFTER: (screenshot for both the timed out test and for the failed expect test)

image
image
image
(...)

It's a followup fix for https://jira.tools.sap/browse/CXSPA-8564


projects/ssr-tests/src/environments/custom-test-environment.ts
```diff
@@ -0,0 +1,61 @@
...
+import { Event, State } from 'jest-circus';
+import { TestEnvironment as NodeEnvironment } from 'jest-environment-node';
+import { getRawLogsPretty } from '../utils/log.utils';
+
+/**
+ * This is a custom Jest environment that adds the SSR logs to the test errors.
+ * The logs are added to the error `cause` property for failed tests,
+ * so those logs will be printed by Jest along with the test error.
+ */
+export default class CustomTestEnvironment extends NodeEnvironment {
```
pawelfras, line 16: `query:` How Jest knows that `CustomTestEnvironment` shoudld be used? I didn't find any usage of this class. Is it file-based config, so Jest knows that it should use default export from `environments/custom-test-environment`?

pawelfras (approved):
LGTM. QA done.

projects/ssr-tests/src/environments/custom-test-environment.ts
Platonn: Jest uses the `default export` from the file given in the `jest.config.js`
see https://github.com/SAP/spartacus/blob/1f605852dd43f7a10ad3be0d19a22cb6db34f396/projects/ssr-tests/jest.config.js#L34

PR #19306, opened 2024-09-30
feat: (CXSPA-8032) - Change a into button for "Cancel Orders" and "Request a Return"

petarmarkov9449:
Closed: [CXSPA-8032](https://jira.tools.sap/browse/CXSPA-8032)

Back button was read as link. UI type read was incorrect.


feature-libs/order/components/order-details/order-detail-actions/order-detail-actions.component.html
```diff
@@ -10,7 +10,7 @@
     </div>
 
     <div class="col-xs-12 col-md-4 col-lg-3">
-      <a
+      <button
```
Pio-Bar: I have just taken a look at the jira ticket. Miguel suggested using a `role="button"` attribute. This would be a better solution as its functionally still an anchor element. If we go this route we won't a feature flag too 🙂

Pio-Bar (changes requested):

feature-libs/order/components/order-details/order-detail-actions/order-detail-actions.component.html
petarmarkov9449: I've just updated it.

feature-libs/order/components/order-details/order-detail-actions/order-detail-actions.component.spec.ts
Pio-Bar, line 1: We probably no longer need to modify the tests here 🙂

Pio-Bar (changes requested):

feature-libs/order/components/order-details/order-detail-actions/order-detail-actions.component.spec.ts
petarmarkov9449: I've just restored it.

Pio-Bar (approved):

Zeyber (approved):

developpeurweb (approved):

PR #19315, opened 2024-10-02
fix: (CXSPA-7996) associated a coupon input field with its label

StanislavSukhanov:
* updated a voucher.coupon translation
  * removed an aria-label from an input field and associated the input with it's label
  * removed unneeded translation of an aria-label field
  * closes https://jira.tools.sap/browse/CXSPA-7996


Pio-Bar (dismissed):

feature-libs/cart/base/assets/translations/en/cart.json
```diff
@@ -31,9 +31,8 @@
     "toBeDetermined": "TBD"
   },
   "voucher": {
-    "coupon": "Have a coupon?",
+    "coupon": "Add a coupon",
```
Zeyber: We should be careful about changing or removing properties already in use due to [breaking changes](https://sap.github.io/spartacus-docs/breaking-changes/#avoiding-breaking-changes) (even translations). We can use the `@depreciated` annotation here to show that users should not use `coupon` and `couponLabel` anymore and they will be modified or removed in the future (next major release we do). We can look for the `@deprecated` label when we start removing our breaking changes during cleanup to know where to remove properties.

Note: When adding the @deprecated property, you will likely have the sonarqube check fail in the pipeline. You can ask @RadhepS for an exception here to pass the pipeline.

Lets create a new property for `"Add a coupon"` to avoid making the breaking change.

Zeyber: Sorry, please disregard the first point about using @deprecatied. But lets still use a new property to make sure we are covered :)

StanislavSukhanov: Fixed, thanks for the comments!

Zeyber (approved):

developpeurweb (approved):

PR #19322, opened 2024-10-02
CI Pipeline Runs with CCV2

giancorderoortiz:
DO NOT MERGE!
Trials only.


giancorderoortiz:
No longer needed. 
Will use S2 as base and regenerate if needed. 
https://github.com/SAP/spartacus/pull/19365/files

PR #19325, opened 2024-10-03
fix: (CXSPA-8034) remove aria label from title dropdown

StanislavSukhanov:
* closes https://jira.tools.sap/browse/CXSPA-8034

### DIFF
--- a/feature-libs/user/profile/assets/translations/en/userProfile.json
+++ b/feature-libs/user/profile/assets/translations/en/userProfile.json
@@ -30,6 +30,7 @@
     "confirmNewPassword": "Confirm New Password",
     "resetPassword": "Reset Password",
     "createAccount": "Create an account",
+    "title": "Title",
     "titleOptional": "Title (optional)",
     "firstName": {
       "label": "First name",
--- a/feature-libs/user/profile/assets/translations/en/userProfile.json
+++ b/feature-libs/user/profile/assets/translations/en/userProfile.json
@@ -30,7 +30,6 @@
     "confirmNewPassword": "Confirm New Password",
     "resetPassword": "Reset Password",
     "createAccount": "Create an account",
-    "title": "Title",
     "titleOptional": "Title (optional)",
     "firstName": {
       "label": "First name",
--- a/feature-libs/user/profile/components/register/register.component.html
+++ b/feature-libs/user/profile/components/register/register.component.html
@@ -29,9 +29,6 @@
                 bindValue="code"
                 formControlName="titleCode"
                 id="title-select"
-                [cxNgSelectA11y]="{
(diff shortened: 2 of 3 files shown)


Pio-Bar (dismissed):

Zeyber (dismissed):

developpeurweb:
@StanislavSukhanov for some reason the combobox now gets identified as a "Link".
Also, in VoiceOver, we lost the item count narration. 

link instead of combo

Please compare this with another optional combobox like the one on the _Shipping Address_ step during Checkout. Let's please bring back the item count for VoiceOver and rework the code to make JAWS narrate the element as "edit combo".

edit combo

developpeurweb (changes requested):
Pls see previously attached screenshots

Pio-Bar:
Screenshot 2024-10-10 at 16 26 18

Here's my JAWS output. It's not in the history, but I can hear the word "link" after tabbing to the input.
FYI @StanislavSukhanov The removed `cxNgSelectA11y` directive does more than just add an aria-label. This is why the count is missing on voice over probably.

StanislavSukhanov:
Hi @Pio-Bar. Thanks for the advice, I did suspected that too.  I've put back the directive. The items count is present again in the Voice Over.

Pio-Bar (dismissed):

feature-libs/user/profile/assets/translations/en/userProfile.json
```diff
@@ -30,7 +30,6 @@
     "confirmNewPassword": "Confirm New Password",
     "resetPassword": "Reset Password",
     "createAccount": "Create an account",
-    "title": "Title",
```
Zeyber: I'm not sure about removing this translation key as it would constitute a breaking change. Can you confirm if the key has only been introduced to set the aria-label on ng-select? And nowhere else in the past?

Zeyber (changes requested):

feature-libs/user/profile/assets/translations/en/userProfile.json
StanislavSukhanov: Hi @Zeyber! I have checked in the repository and couldn't find such combination other than in a place it was removed from. 
Please let me know if I need to revert this change. Thanks a lot 👍

StanislavSukhanov: @Zeyber reverted, fixed.

Zeyber (dismissed):

StanislavSukhanov:
@developpeurweb It is fixed. Could you please have a look on that again? Thank you

developpeurweb (dismissed):

Pio-Bar (approved):

PR #19326, opened 2024-10-03
Revert of Revert "test(ssr-tests): print SSR Logs for errors in SSR tests, ALSO when a test times out (#19302)

Platonn:
Original PR: https://github.com/SAP/spartacus/pull/19302
Revert of original PR: https://github.com/SAP/spartacus/pull/19319 becasue of errors on the `develop`'s pipeline https://github.com/SAP/spartacus/actions/runs/11142085195/job/30964328791

Now bringing back the contents of the original PR, but with ensuring SSR Tests don't encounter the following error again on the pipeline:
```
node:internal/event_target:1099
  process.nextTick(() => { throw err; });
                           ^
Error: Unknown worker message type message
This is caused by either a bug in Node.js or incorrect usage of Node.js internals.
Please open an issue with this stack trace at https://github.com/nodejs/node/issues
    at Function.fail (node:internal/assert:20:9)
    at Worker.[kOnMessage] (node:internal/worker:354:12)
    at MessagePort.<anonymous> (node:internal/worker:232:57)
    at _ZoneDelegate.Object.<anonymous>._ZoneDelegate.invokeTask (/<project>/node_modules/zone.js/bundles/zone.umd.js:445:35)
    (... rest of stacktrace ...)
    at MessagePort.<anonymous> (node:internal/per_context/messageport:23:28) {
  code: 'ERR_INTERNAL_ASSERTION'
```

**How fixed the problem:**
Removed `jest-preset-angular` from the Jest setup of SSR Tests. Used simply `ts-jest` instead.

**Why it helped**
I only deduce that the new custom `testEvironment` introduced in the Original PR that extends from `NodeEnvironment` from `jest-environment-node` **conflicted** with the `jest-preset-angular` which under the hood assumes simulating Browser environment (not Node).

**Why we really don't need or want `jest-preset-angular` in SSR Tests project**
- it enforces `jsdom` testEnvironment (see https://github.com/thymikee/jest-preset-angular/blob/ff0895f4b8dfa561c4f4bc9779016d9d4d7213c7/src/presets/index.ts#L6)
- it enforces importing zone.js (see https://github.com/thymikee/jest-preset-angular/blob/4c2b674634b64eed267e7c0ddb715f52e7839cba/setup-jest.js#L1-L2)
- it enforces initializing Jest test environment with Angular's `BrowserDynamicTestingModule` (see https://github.com/thymikee/jest-preset-angular/blob/4c2b674634b64eed267e7c0ddb715f52e7839cba/setup-jest.js#L18)
- in SSR Tests we want to simply run NodeJS processes, but not simulate Angular components in a browser

Note: Finally I could reproduce it on local too, while using JS `testEnvironment` file instead of TS. Moreover, I could reproduce it only on the first run of tests on local. On the second run, everything would pass. Only when I renamed the file with `testEvironment`, on the first run of tests, the error could be reproduced again.

related to https://jira.tools.sap/browse/CXSPA-8564


Platonn:
SSR Tests passed on the pipeline (I've run it manually, because it deliberately doesn't run on every PR)
See https://github.com/SAP/spartacus/actions/runs/11159409151/job/31017762912#step:7:50

pawelfras (approved):
LGTM

PR #19307, opened 2024-09-30
test(SSR e2e): case when a call to Backend timeouts

Platonn:
Added a test case: when a call to Backend timeouts

btw. changes:
- replaced `callback` with `responseInterceptor` in ProxyUtils

fixes https://jira.tools.sap/browse/CXSPA-8386

PR #19331, opened 2024-10-03
test(SSR e2e): re-enable test "should receive response with status 500 if HTTP error occurred when calling other than cms/pages API URL"

Platonn:
Note: the change of the intercepted API URL `cms/pages` to `cms/components` is not the fix of the flakiness.
It's simply reverting an accidental change  that was made in the very recent PR - see https://github.com/SAP/spartacus/pull/19307/files?w=1#diff-799b5da094cfd6cc4192ea13d66afac3f264e0502fe21893c3c4e04f024fae45R73 (change from `cms/components` to `cms/pages`). This accidental change wasn't spotted before, because this test used to be skipped.

Now we're un-skipping it.

The flakiness could have been fix by the PR https://github.com/SAP/spartacus/pull/19307/, which introduced a new way of intercepting requests - `responseInterceptor` (instead of the old `callback`).

I've re-run this test 20+ times on local and it passes.

closes https://jira.tools.sap/browse/CXSPA-8570


projects/ssr-tests/src/utils/log.utils.ts
```diff
@@ -27,25 +27,64 @@ export function clearSsrLogFile(): void {
...
- * Otherwise it throws an error.
+ * Otherwise it warns.
  *
  * Note: multi-line JSONs (printed by SSR in dev mode) cannot be parsed by `JSON.parse`.
  *       That's why we need to to run SSR in prod mode to get single line JSON logs.
  */
-function validateJsonsInLogs(logs: string[]): void {
-  logs.forEach((text) => {
-    if (text.charAt(0) === '{') {
+function validateJsonsInLogs(rawLogs: string[]): void {
```
pawelfras: `medium:` Can you provide steps to verify the warning locally? Btw. the SSR testing framework is getting more complex and maybe we should consider covering by testing at least the most crucial parts of utils

Platonn:
Full pipeline run, with SSR Tests: https://github.com/SAP/spartacus/actions/runs/11181040768

projects/ssr-tests/src/utils/log.utils.ts
Platonn: Good question.
After rethinking, I believe we shouldn't validate logs in runtime of tests to detect if the app is built in dev mode. But instead we should check if the app is built in prod mode, before running the tests.
So I reverted changes to logs validation made in this PR.
And in this PR I'm leaving only the un-skip of the test.

Created a separate PR for checking if app is built in prod mode: https://github.com/SAP/spartacus/pull/19334

pawelfras (approved):
QA done.

PR #19334, opened 2024-10-04
test(SSR e2e): Validate SSR app under test is built in prod mode and using a proxy backend, before running tests

Platonn:
Problem:
To run SSR Tests, it's necessary to execute: `npm run build` and `npm run build:ssr:local-http-backend` - in order to have an app built in prod mode and using a proxy backend. This step can be easily overlooked by newcomers.
Otherwise, some tests might fail:
- some tests asserting JSON SSR messages might fail - if app built in dev mode which prints JSON logs in multi-line (instead of single-line)
- some tests mocking the backend might fail - if app is not configured to call a Proxy Backend 

Previously, we were checking only if the app is built in prod mode. And we were checking it by symptoms in tests runtime, with the following heuristic: if there's any log line starting with character `{` but not possible to parse as a JSON, then it's probably running in dev mode. In such case were throwing an error.

Now, were checking both if the app is in prod mode and if it's using a backend proxy server. We check it statically before running the tests, with the following heuristics:
- if the build output file `dist/storefrontapp-server/main.js` contains a marker string `CX_BASE_URL:"http://localhost:9002"`, then we know the app is using a local proxy backend
- if the build output file `dist/storefrontapp-server/main.js` contains a marker string `ngDevMode`, then we know the app is build in dev mode

**QA steps:**
**Scenario 1 ❌ - missing `main.js`:**
- remove `dist/storefrontapp-server/main.js`
- run `npm run test:ssr`
- verify the following error message appears before running tests:
   image

**Scenario 2 ❌ - dev-mode build:**
- run `npm run dev:ssr` and wait until it fully compiles and then kill the process
- run `npm run test:ssr`
- verify the following error message appears before running tests:
   image

**Scenario 3 ❌ - prod-mode build BUT not using local http proxy backend:**
- run `npm run build && npm run build:ssr`
- run `npm run test:ssr`
- verify the following error message appears before running tests:
   image

**Scenario 4 ✅ - prod-mode build AND  using local http proxy backend:**
- run `npm run build && npm run build:ssr`
- run `npm run test:ssr`
- verify the tests run successfully:
   image

closes https://jira.tools.sap/browse/CXSPA-8586


pawelfras:
Full pipeline run, with SSR Tests: https://github.com/SAP/spartacus/actions/runs/11181902202

pawelfras (approved):
LGTM 🔥 QA done.

PR #19337, opened 2024-10-04
test(SSR e2e): Validate http port 4000 is not used by other process, before running tests

Platonn:
Problem:
Each SSR test starts and kills an SSR process in the background, listening on port 4000.
If you interrupt the tests that you run on local (e.g. by pressing Cmd+C), then the Jest tests process is stopped but the SSR process might be still dangling in the background.
So the next time you try to re-run SSR Tests, they'll fail with because the new SSR process cannot be started on the already occupied port 4000.

Previosuly, the we had to wait for all tests fail (30secons), until we were presented with an vague error message about port being in use. Moreover it didn't get any hint how to solve the problem.

Now, the problem is detected before running any test. The error message is descriptive and provides a hint how to kill a process that is listening on the conflicting port.

Note: Instead of detecting the problem and showing an useful message, my first strategy was to avoid the problem in general. I mean that whenever the parent Jest process is interupted, I wanted the child SSR process to be killed as well. I've tried hooks like `process.on('SIGTERM')` etc. , I've tried tampering the config of spawning the `child` process (remove `detach: true`, add `killSignal`) or calling `child.unref()`, but I had various problems - the child process was still not killed when parent was interupted or the child process couldn't be killed with `process.kill()` in the tests body

**QA steps**
1. Build SSR `npm build:libs && npm build && npm build:local-http-backend`
2. In one terminal run `npm run serve:ssr`
3. In second terminal run `npm run test:ssr`
4. Verify the following error message is presented in the second terminal:
   
   BEFORE:

   image

   AFTER:
   
   image

5. Kill the SSR process in the first terminal (e.g. using the commands from the advice from the console)
6. Re-run ssr tests in the second terminal and verity they pass

Fixes https://jira.tools.sap/browse/CXSPA-8587


pawelfras:
Full pipeline run with SSR tests: https://github.com/SAP/spartacus/actions/runs/11182025201

pawelfras (dismissed):
LGTM. QA done

pawelfras (approved):

PR #19364, opened 2024-10-08
CI Internal Pipeline Runs with CCV2

giancorderoortiz:
DO NOT MERGE!
Trials only. 

Mirrors https://github.com/SAP/spartacus/pull/19322


giancorderoortiz:
No longer needed. Will use s2 PR and regenerate. 
https://github.com/SAP/spartacus/pull/19365/files

PR #19366, opened 2024-10-09
Fix focus after clicking on hamburger menu

Larisa-Staroverova:
Closes [CXSPA-8585](https://jira.tools.sap/browse/CXSPA-8585)

### DIFF
--- a/projects/storefrontlib/layout/main/storefront.component.ts
+++ b/projects/storefrontlib/layout/main/storefront.component.ts
@@ -120,8 +120,9 @@ export class StorefrontComponent implements OnInit, OnDestroy {
   }
 
   protected focusOnFirstNavigationItem() {
-    const closestNavigationUi =
-      this.elementRef.nativeElement.querySelector('cx-navigation-ui');
+    const closestNavigationUi = this.elementRef.nativeElement.querySelector(
+      '.NavigationBar cx-navigation-ui'
+    );
     const focusable = closestNavigationUi?.querySelector<HTMLElement>(
       'li:not(.back) button, [tabindex="0"]'
     );
--- a/projects/storefrontlib/layout/main/storefront.component.ts
+++ b/projects/storefrontlib/layout/main/storefront.component.ts
@@ -121,7 +121,7 @@ export class StorefrontComponent implements OnInit, OnDestroy {
 
   protected focusOnFirstNavigationItem() {
     const closestNavigationUi = this.elementRef.nativeElement.querySelector(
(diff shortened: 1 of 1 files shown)


Larisa-Staroverova:
Hey @sdrozdsap,
due to your [#19264](https://github.com/SAP/spartacus/pull/19264) PR we experience some issues in the product configuration, namely the focus is placed on `About SAP Commerce Cloud` in the footer and not to the active element in the menu. As you probably saw during your implementation `cx-navigation-ui` is used twice in the HTML treen, ones in the navigation bar and ones in the footer.
If you don't mind, I have extended the query selector to specify which `cx-navigation-ui` element is meant. 
Could you, please, approve this PR?
Thank you very much in advance!
Best regards, Larisa

projects/storefrontlib/layout/main/storefront.component.ts
```diff
@@ -120,8 +120,9 @@ export class StorefrontComponent implements OnInit, OnDestroy {
   }
 
   protected focusOnFirstNavigationItem() {
-    const closestNavigationUi =
-      this.elementRef.nativeElement.querySelector('cx-navigation-ui');
+    const closestNavigationUi = this.elementRef.nativeElement.querySelector(
+      '.NavigationBar cx-navigation-ui'
```
sdrozdsap: Following up on your comment there are not 2 `cx-navigation-ui` elements on the page, but 3 if user is logged in
image

The intention of my changes was to find cx-navigation-ui which is the first on the page: either from `.NavigationBar` or the "My Account". The one in footer should still be the last, at least that was my assumption. So the solution which you've proposed will result in the situation which we had before: if the user is logged in, the focus still lands on the first category menu instead of the my account.

sdrozdsap: As far as I understand your issue, after opening the hamburger menu focus landed on the first footer item, which shouldn't happen if there are any `cx-navigation-ui` higher in the DOM, because `document.querySelector` returns first element that matched the selector. Can you please send me a link to the environment where I can reproduce the issue?

sdrozdsap: Maybe this would work better?

```suggestion
      'header cx-navigation-ui'
```

Larisa-Staroverova: It works very well. I retested on our side. 
If it's okay for you, I would go with this fix.

sdrozdsap (approved):

steinsebastian (approved):

PR #19368, opened 2024-10-09
chore: remove TODO comment related to CXSPA-6793, as won't fix

Platonn:
We won't change the syntax to use to `satisfies` keyword

Reason:
- in case of a missing property, the error message is more readable currently. See screenshots below
   CURRENTLY:
   image
   IF WE USED `satisfies` keyword:
   image
- moreover, if we changed the syntax from `provideFeatureTogglesFactory` to `provideFeatureToggles`, it would change the indentation for all lines with toggles, which is prone to code conflicts with other pending branches
   image

All that said, I don't find it justified to refactor it to use `satisfies` keyword, therefore removing the TODO comment as won't fix

closes https://jira.tools.sap/browse/CXSPA-6793


pawelfras (approved):
Thanks, LGTM. I agree with the reasoning to not use `provideFeatureToggles` and `satisfies` keyword.

PR #19369, opened 2024-10-09
refactor(optimized-ssr-engine.spec): don't confuse terms: request, render, HTML, response

Platonn:
Terms used in tests now comply with the following semantics:
- `request` - is sent by client and received by the OptimizedSsrEngine
- `response` - is sent back to the client by the OptimizedSsrEngine
- `render` / `rendering` - the process performed by Angular to produce a final HTML (or error) for given rendering key (note: the rendering key is by default the URL path in the `request`)
- `HTML` - the result of the rendering

closes https://jira.tools.sap/browse/CXSPA-7860


core-libs/setup/ssr/optimized-engine/optimized-ssr-engine.spec.ts
```diff
@@ -839,13 +841,13 @@ describe('OptimizedSsrEngine', () => {
...
 
-        expect(engineRunner.renders).toEqual(['', 'a-1']);
+        expect(engineRunner.responses).toEqual(['', 'a-1']);
       }));
     });
   });
 
   describe('forcedSsrTimeout option', () => {
-    it('should fallback to CSR when forcedSsrTimeout timeout is exceeded for ALWAYS_SSR rendering strategy, and return the timed out render in the followup request', fakeAsync(() => {
+    it('should fallback to CSR when forcedSsrTimeout timeout is exceeded for ALWAYS_SSR rendering strategy, and when after that rendering ends its HTML should be reused in the next request for the same url', fakeAsync(() => {
```
pawelfras:
```suggestion
    it('should fallback to CSR when forcedSsrTimeout timeout is exceeded for ALWAYS_SSR rendering strategy, and after that rendering ends, its HTML should be reused in the next request for the same url', fakeAsync(() => {
```

pawelfras (dismissed):
LGTM 🔥 one nitpick

pawelfras (approved):

PR #19372, opened 2024-10-09
fix: ensure all possible SsrOptimizationOptions have explicitly defined default values

Platonn:
Thanks to this, all the options (with their explicit default values) will printed in the logs on the start of SSR, which should improve developers' confidence while troubleshooting.

closes https://jira.tools.sap/browse/CXSPA-7482


core-libs/setup/ssr/optimized-engine/ssr-optimization-options.ts
```diff
@@ -48,7 +49,7 @@ export interface SsrOptimizationOptions {
    * Time in milliseconds after prerendered page is becoming stale and should
    * be rendered again.
    */
-  ttl?: number;
+  ttl?: number | undefined;
```
Platonn, line 52: Note: because it was an optional property, it could have been effectively `undefined` (and it was indeed `undefined` value in practice by defailt).
So now adding `undefined` to the type here is not a breaking change.

core-libs/setup/ssr/optimized-engine/ssr-optimization-options.ts
```diff
@@ -188,10 +189,30 @@ type DeepRequired<T> = {
...
+> & {
+  ttl: number | undefined; // needed, otherwise we could not set the value `ttl: undefined` value (due to the Required<...>)
+} & DeepRequired<
+    // all nested properties of `ssrFeatureToggles` are required too
+    Pick<
+      SsrOptimizationOptions,
+      //
+      'ssrFeatureToggles'
+    >
+  > = {
```
pawelfras: 'minor:' Type of `defaultSsrOptimizationOptions` gets even more complex so let's move this to a separate `type` property to reduce the noise around default options:
```ts

type DefaultSsrOptimizationOptions = Omit<
  Required<SsrOptimizationOptions>,
  | 'debug' // debug is deprecated and not used anymore
  | 'ttl' // ttl is required but its default value is `undefined`
> & {
  ttl: number | undefined; // needed, otherwise we could not set the value `ttl: undefined` value (due to the Required<...>)
} & DeepRequired<
    // all nested properties of `ssrFeatureToggles` are required too
    Pick<
      SsrOptimizationOptions,
      //
      'ssrFeatureToggles'
    >
  >

export const defaultSsrOptimizationOptions: DefaultSsrOptimizationOptions = { 
   /*...*/
}

```

core-libs/setup/ssr/optimized-engine/optimized-ssr-engine.ts
```diff
@@ -21,13 +20,9 @@ import {
...
   SsrOptimizationOptions,
   defaultSsrOptimizationOptions,
+  getDefaultRenderKey,
 } from './ssr-optimization-options';
 
-/**
- * Returns the full url for the given SSR Request.
- */
-export const getDefaultRenderKey = getRequestUrl;
-
```
pawelfras, line 30: `query:` if we now the default value of `renderKeyResolver`, maybe we can adjust `getRenderingKey` method so it simply uses the value from ssr options?
Here;s its current shape:
```ts
protected getRenderingKey(request: Request): string {
    return this.ssrOptions?.renderKeyResolver
      ? this.ssrOptions.renderKeyResolver(request)
      : getDefaultRenderKey(request); // <--- function same as the default value of `ssrOptions.renderKeyResolver`
  }
  ```

Platonn: I'm afraid our current (wrong) TS typing of `this.ssrOptions?` (which TS see as possibly-undefined) and `this.ssrOptions.renderKeyResolver?` (which TS see as possibly-undefined) doesn't allow for it.

core-libs/setup/ssr/optimized-engine/ssr-optimization-options.ts
Platonn: done

core-libs/setup/ssr/optimized-engine/optimized-ssr-engine.ts
pawelfras: Yes, I faced this issue before when using `ssrOptions` in this class. Let's consider refactoring it in the future 👍

pawelfras (approved):
LGTM

PR #19374, opened 2024-10-09
GHAS. Prod Config. Fixes incomplete string escaping or encoding

giancorderoortiz:
Fixes [https://github.com/SAP/spartacus/security/code-scanning/40](https://github.com/SAP/spartacus/security/code-scanning/40)

To fix the problem, we need to ensure that all occurrences of the characters `>`, `<`, `≥`, and `≤` in the `valueName` string are replaced. This can be achieved by using a regular expression with the global flag (`g`). This change will ensure that every instance of these characters is replaced, not just the first one.

_Suggested fixes powered by Copilot Autofix. Review carefully before merging._


giancorderoortiz:
Fixes https://github.com/SAP/spartacus/security/code-scanning/40

Larisa-Staroverova (approved):

PR #19381, opened 2024-10-10
feat(a11y): add a11y selected label improvements to checkout Address/PaymentMethod components

uroslates:


feature-libs/checkout/base/components/checkout-delivery-address/checkout-delivery-address.component.ts
```diff
@@ -329,4 +335,12 @@ export class CheckoutDeliveryAddressComponent implements OnInit {
   protected shouldUseAddressSavedInCart(): boolean {
     return !!this.checkoutConfigService?.shouldUseAddressSavedInCart();
   }
+
+  private getCardRole(isCardSelected: boolean): 'button' | 'region' {
```
Zeyber: Nice use of the feature flags here! Lets make this method `protected` instead of `private` so that lib users can still access it if they have a need to see/modify it without cloning the repo.

Zeyber: Lets also add a unit test assertion to check that the `role` is set correctly since there is some logic here.

feature-libs/checkout/base/components/checkout-payment-method/checkout-payment-method.component.ts
```diff
@@ -299,9 +303,14 @@ export class CheckoutPaymentMethodComponent implements OnInit, OnDestroy {
...
     const isSelected = selected?.id === paymentDetails.id;
+    const isButtonRole =
+      this.featureConfigService?.isEnabled(
+        'a11ySelectLabelWithContextForSelectedAddrOrPayment'
+      ) && !isSelected;
+    const role = isButtonRole ? 'button' : 'region';
 
     return {
-      role: 'region',
+      role,
```
Zeyber, line 313: Same as above regarding unit test assertion.

Zeyber (changes requested):

feature-libs/checkout/base/components/checkout-delivery-address/checkout-delivery-address.component.ts
uroslates: 👍 Addressed both of these in latest commit!

feature-libs/checkout/base/components/checkout-payment-method/checkout-payment-method.component.ts
uroslates: 👍 Addressed both of these in latest commit!

uroslates:
Closes https://jira.tools.sap/browse/CXSPA-8041

feature-libs/checkout/base/components/checkout-payment-method/checkout-payment-method.component.spec.ts
```diff
@@ -617,5 +617,82 @@ describe('CheckoutPaymentMethodComponent', () => {
...
+            textDefaultPaymentMethod: '✓ DEFAULT',
+            textExpires: 'Expires',
+            textUseThisPayment: 'Use this payment',
+            textSelected: 'Selected',
+          },
+          paymentMethod1
+        ).role).toEqual('region');
+      });
+
+      it('should be set to "region" when feature flag is disabled', () => {
```
Zeyber: We don't really need to test for the feature flags because they will be removed anyway next major and we can always assume they are active for unit tests. The main thing to test here was that `isSelected` sets the role correctly. Is it possible to remove any reference to feature flags here?

feature-libs/checkout/base/components/checkout-delivery-address/checkout-delivery-address.component.spec.ts
```diff
@@ -500,4 +562,11 @@ describe('CheckoutDeliveryAddressComponent', () => {
       expect(getSpinner()).toBeFalsy();
     });
   });
+
+  // describe('getCardRole()', () => {
```
Zeyber: Left over?

feature-libs/checkout/base/components/checkout-delivery-address/checkout-delivery-address.component.spec.ts
```diff
@@ -345,6 +346,67 @@ describe('CheckoutDeliveryAddressComponent', () => {
...
+          'default',
+          'shipTo',
+          'selected',
+          'P',
+          'M'
+        ).role).toEqual('region');
+        expect(component['getCardRole']).toHaveBeenCalledWith(true);
+      });
+
+      it('should be set to "region" when feature flag is disabled', () => {
```
Zeyber: Same as below about testing feature flags.

feature-libs/checkout/base/components/checkout-delivery-address/checkout-delivery-address.component.spec.ts
```diff
@@ -345,6 +346,67 @@ describe('CheckoutDeliveryAddressComponent', () => {
       );
       expect(card.actions?.length).toBe(1);
     });
+    
+    describe('role', () => {
+      it('should be set to "region" for selected addresses when feature flag is enabled', () => {
+        spyOn(featureConfig, 'isEnabled').and.returnValue(true);
+        expect(component.getCardContent(
+          // isSelected = true!
```
Zeyber: Left over reference note?

Zeyber (changes requested):

feature-libs/checkout/base/components/checkout-delivery-address/checkout-delivery-address.component.spec.ts
uroslates: 👍  Addressed all test related comments in latest commit.

Zeyber (approved):

developpeurweb (approved):

PR #19392, opened 2024-10-14
chore: Remove cypress .npmrc

Zeyber:
There are external pipelines that are failing in the e2e step because the environment variable `CONTINUUM_REGISTRY_TOKEN` is not being correctly applied. Lets try to [remove the .npmrc file](https://stackoverflow.com/questions/52015748/npm-failed-to-replace-env-in-config-npm-token) from the e2e directory and see if that helps.

Note that we need to come back for a solution to Access Continuum that works on all pipelines and dev machines.


pawelfras (approved):

pawelfras:
I additionally ran the `Spartacus build pipeline` step on this branch to run this change against all E2E tests and it passed. Here's the related pipeline run: https://github.com/SAP/spartacus/actions/runs/11323263172

PR #19327, opened 2024-10-03
fix: (CXSPA-8033) add aria hidden attribute to 'required' asterisks

StanislavSukhanov:
* closes https://jira.tools.sap/browse/CXSPA-8033

PR #19394, opened 2024-10-14
fix: add * to required billing address form fields on checkout payment details

uroslates:


Zeyber (dismissed):

Zeyber:
A couple things before merge:
- I think we should use `feat:` prefix or better `fix:` in the commit title so that the changelog is generated correctly when we do a release. We should move or drop the "(a11y)" part from the prefix. [Some information on this here](https://sap.github.io/spartacus-docs/commit-guidelines/#type).
- In the commit message, we use `Closes: CXSPA-xxxx` and link it to the ticket in JIRA.

uroslates:
Closes https://jira.tools.sap/browse/CXSPA-8189

developpeurweb (dismissed):

developpeurweb:
@uroslates LGTM, just one last adjustment please, this other PR https://github.com/SAP/spartacus/pull/19327 adds an ARIA attribute to asterisks, can you please sync up with @StanislavSukhanov and update your asterisks to include those attributes as well.

uroslates:
> @uroslates LGTM, just one last adjustment please, this other PR #19327 adds an ARIA attribute to asterisks, can you please sync up with @StanislavSukhanov and update your asterisks to include those attributes as well.

Applied the requested change - * are now not read by the Readers.

developpeurweb (approved):

PR #19398, opened 2024-10-15
GHAS. Fix  Insecure randomness

giancorderoortiz:
Fixes [https://github.com/SAP/spartacus/security/code-scanning/49](https://github.com/SAP/spartacus/security/code-scanning/49)

To fix the problem, we need to replace the use of `Math.random()` with a cryptographically secure random number generator. In a browser environment, we can use `crypto.getRandomValues` to generate secure random values. This method provides a cryptographically secure way to generate random numbers.

We will:
1. Import the necessary `crypto` module.
2. Replace the `Math.random()` call with `crypto.getRandomValues` to generate a secure random value.
3. Convert the generated random value to a string format similar to the original implementation.

_Suggested fixes powered by Copilot Autofix. Review carefully before merging._


feature-libs/cart/base/core/facade/multi-cart.service.ts
```diff
@@ -85,7 +85,9 @@ export class MultiCartService implements MultiCartFacade {
    * Simple random temp cart id generator
    */
   protected generateTempCartId(): string {
-    const pseudoUuid = Math.random().toString(36).substring(2, 11);
+    const array = new Uint32Array(1);
+    window.crypto.getRandomValues(array);
```
kpawelczak: I would be careful with `window` as it could potentially break the SSR rendering process.

kpawelczak: The function is triggered when a new cart is created, which is unlikely to occur during the rendering process. However, it is recommended to check if the window object exists before using it.

kpawelczak: as per @Platonn comment I've switched `window` with `globalThis`.

kpawelczak: Reverted `globalThis.crypto` as it didn't work in SSR.

RadhepS (approved):

PR #19405, opened 2024-10-16
fix: (CXSPA-8156) checkout -  add landmarks 

StanislavSukhanov:
* checkout page add landmarks to: shipping address, delivery mode, payment sections

  * closes https://jira.tools.sap/browse/CXSPA-8156

### DIFF
--- a/feature-libs/checkout/base/components/checkout-delivery-address/checkout-delivery-address.component.html
+++ b/feature-libs/checkout/base/components/checkout-delivery-address/checkout-delivery-address.component.html
@@ -1,8 +1,5 @@
-<section [attr.aria-labelledby]="'shipping-address-title'">
-  <h2
-    id="shipping-address-title"
-    class="cx-checkout-title d-none d-lg-block d-xl-block"
-  >
+<section [attr.aria-label]="'checkoutAddress.shippingAddress' | cxTranslate">
+  <h2 class="cx-checkout-title d-none d-lg-block d-xl-block">
     {{ 'checkoutAddress.shippingAddress' | cxTranslate }}
   </h2>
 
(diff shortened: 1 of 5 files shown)


feature-libs/checkout/base/components/checkout-delivery-address/checkout-delivery-address.component.html
```diff
@@ -1,104 +1,106 @@
-<h2 class="cx-checkout-title d-none d-lg-block d-xl-block">
-  {{ 'checkoutAddress.shippingAddress' | cxTranslate }}
-</h2>
+<section [attr.aria-label]="'checkoutAddress.shippingAddress' | cxTranslate">
```
Pio-Bar: Unfortunately, I think changing the structure here would be breaking.

Pio-Bar (dismissed):

feature-libs/checkout/base/components/checkout-delivery-address/checkout-delivery-address.component.html
StanislavSukhanov: Hi @Pio-Bar! Thanks for your comment 👍 . Fixed

Pio-Bar (dismissed):

Zeyber (dismissed):

developpeurweb:
@StanislavSukhanov works perfect! Can you please apply the same solution to three remaining sections using the following labels:

1. `<section aria-label="Delivery Mode">`
2. `<section aria-label="Payment Method">`
3. `<section aria-label="Review Order">`

Thanks!

Zeyber (approved):

PR #19406, opened 2024-10-16
feat: Use tabs instead of radio group for pickup options

sdrozdsap:
Closes: https://jira.tools.sap/browse/CXSPA-7972

### DIFF
--- a/feature-libs/pickup-in-store/components/container/cart-pickup-options-container/cart-pickup-options-container.component.ts
+++ b/feature-libs/pickup-in-store/components/container/cart-pickup-options-container/cart-pickup-options-container.component.ts
@@ -310,7 +310,10 @@ export class CartPickupOptionsContainerComponent implements OnInit, OnDestroy {
           );
         });
 
-      if (!this.displayNameIsSet) {
+      if (
+        !this.featureConfigService.isEnabled('a11yPickupOptionsTabs') &&
+        !this.displayNameIsSet
+      ) {
         this.openDialog(event.triggerElement);
       }
     } else if (typeof event === 'string') {
@@ -344,7 +347,10 @@ export class CartPickupOptionsContainerComponent implements OnInit, OnDestroy {
           );
         });
 
-      if (!this.displayNameIsSet) {
+      if (
+        !this.featureConfigService.isEnabled('a11yPickupOptionsTabs') &&
+        !this.displayNameIsSet
+      ) {
         this.openDialog();
       }
     }
--- a/feature-libs/pickup-in-store/components/presentational/pickup-options/pickup-options.component.ts
+++ b/feature-libs/pickup-in-store/components/presentational/pickup-options/pickup-options.component.ts
@@ -5,18 +5,26 @@
  */
 
 import {
+  AfterViewInit,
+  ChangeDetectorRef,
   Component,
   ElementRef,
   EventEmitter,
   inject,
   Input,
   OnChanges,
+  OnDestroy,
   Output,
   ViewChild,
+  TemplateRef,
+  Optional,
 } from '@angular/core';
 import { FormControl, FormGroup } from '@angular/forms';
 import { FeatureConfigService, useFeatureStyles } from '@spartacus/core';
 import { PickupOption } from '@spartacus/pickup-in-store/root';
+import { TAB_MODE, Tab, TabComponent, TabConfig } from '@spartacus/storefront';
+import { Subscription, take } from 'rxjs';
+import { PickupOptionsTabs } from './pickup-options.model';
 
 /**
  * The presentational component of a pair of radio buttons for pickup options for a product.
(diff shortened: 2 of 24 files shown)


feature-libs/pickup-in-store/components/presentational/pickup-options/pickup-options.component.ts
```diff
@@ -89,4 +142,46 @@ export class PickupOptionsComponent implements OnChanges {
     // Return false to stop `onPickupOptionChange` being called after this
     return false;
   }
+
+  protected initializeTabs() {
+    this.tabs = [
```
sdrozdsap, line 142: @Zeyber  here I've tried to use the `tabs$` input property here, however it didn't work for me. We can catch up on this issue on slack

sdrozdsap:
Cherry-picked commit from https://github.com/SAP/spartacus/pull/19421 to fix the unit test failures

feature-libs/pickup-in-store/components/presentational/pickup-options/pickup-options.component.ts
```diff
@@ -5,27 +5,40 @@
...
   ViewChild,
+  TemplateRef,
 } from '@angular/core';
 import { FormControl, FormGroup } from '@angular/forms';
 import { FeatureConfigService, useFeatureStyles } from '@spartacus/core';
 import { PickupOption } from '@spartacus/pickup-in-store/root';
+import { TAB_MODE, Tab, TabComponent, TabConfig } from '@spartacus/storefront';
+import { Subscription, take } from 'rxjs';
 
+export enum PickupOptionsTabs {
```
Zeyber: Could we move this to a `model.ts` file for easier import?

feature-libs/pickup-in-store/components/presentational/pickup-options/pickup-options.component.ts
```diff
@@ -46,25 +59,60 @@ export class PickupOptionsComponent implements OnChanges {
...
   pickupId = `pickup-id:${Math.random().toString(16)}`;
   deliveryId = `delivery-id:${Math.random().toString(16)}`;
 
   pickupOptionsForm = new FormGroup({
     pickupOption: new FormControl<PickupOption | null>(null),
   });
+  tabs: Tab[];
+  tabConfig: TabConfig;
+
+  protected cdr = inject(ChangeDetectorRef);
```
Zeyber: We should make this optional for now.

feature-libs/pickup-in-store/components/presentational/pickup-options/pickup-options.component.ts
Zeyber: I can see there is a lot going on with this component with pre-existing change detection calls and emits. I don't think the `TabComponent` can really be designed to account for it. So, the way you update the tabs based on this component's logic is ok :)

Zeyber (changes requested):
I am happy to see that we will be able to clean up a lot of the complexity in future :)

Only minor comments from me. But lets wait to merge [this](https://github.com/SAP/spartacus/pull/19244) since the TabComponent logic overlaps.

feature-libs/pickup-in-store/components/presentational/pickup-options/pickup-options.component.ts
sdrozdsap: Done

feature-libs/pickup-in-store/components/presentational/pickup-options/pickup-options.component.ts
sdrozdsap: Done

Zeyber (dismissed):

developpeurweb:
Hey @sdrozdsap , great work on this component! I have 3 small adjustments, hoping they are not very complicated:

1. Let’s remove the heading level 5 (h5) tag for the Tabgroup legend, since you already have an aria-label for the tablist, let’s just change that `h5` into to a `<div>` otherwise it will show up on the screen reader headings’ list.
2. In the Tabgroup pattern, the Tab key function is missing. On pressing, this key should focus the tabpanel regardless of the type of content it has, please see the wai-aria pattern here: https://www.w3.org/WAI/ARIA/apg/patterns/tabs/examples/tabs-automatic/  I see the Tab key function is also missing in the Tabgroup for the Product Details, maybe it got chopped off at some point.
3. For the cart version only, please set all text to a minimum of 14px, a combination of “type” 7 and 8 as per this guide, 4th item: https://wiki.one.int.sap/wiki/display/spar/UI+Styling+and+Accessibility+Onboarding Also, please make it a bit tighter everywhere, similar to the screenshot below.

cart delivery method

feature-libs/pickup-in-store/components/container/cart-pickup-options-container/cart-pickup-options-container.component.ts
```diff
@@ -121,7 +126,7 @@ export class CartPickupOptionsContainerComponent implements OnInit, OnDestroy {
       cartType: CartType;
     }>
   ) {
-    // Intentional empty constructor
```
Zeyber: I wonder if we should leave a note here to leave this constructor empty when we remove `useFeatureStyles` in the future.

Zeyber (changes requested):

feature-libs/pickup-in-store/components/container/cart-pickup-options-container/cart-pickup-options-container.component.ts
sdrozdsap: WIll do!

Zeyber (dismissed):

developpeurweb (dismissed):

developpeurweb (approved):

PR #19408, opened 2024-10-16
fix: (CXSPA-1128) - Select a11y directive sanitized innerText

Pio-Bar:
Fixes vulnerability introduced by https://github.com/SAP/spartacus/commit/ec946e913298491b43312c165626299a31f27eec 
Sanitizing the text should prevent a potential XSS.


giancorderoortiz (approved):
ok. 

See comments in https://v17.angular.io/api/platform-browser/DomSanitizer
"... 
If the security context is HTML and the given value is a plain string, this method sanitizes the string, removing any potentially unsafe content. For any other security context, this method throws an error if provided with a plain string.
..."

PR #19410, opened 2024-10-17
chore: Add configuration for OPF

Matejk00:
Closes: CXSPA-8562


scripts/install/config.opf.sh
```diff
@@ -0,0 +1,13 @@
+# This file contains default values to use in your config.sh when @spartacus/opf is being included
```
FollowTheFlo, line 1: good idea to have a separate file with pre-filled values

FollowTheFlo (approved):
good to me.

PR #19414, opened 2024-10-18
fix: (CXSPA-8668) - Image zoom legacy media component styles

Pio-Bar:
Closes: [CXSPA-8668](https://jira.tools.sap/browse/CXSPA-8668)

Let's use a more specific selector to make sure the legacy media component remains unchanged. 
This way both versions behave predictably.


rmch91 (approved):
LGTM

PR #19421, opened 2024-10-21
fix: update EPD visualization unit test to stop failures

RadhepS:
closes: https://jira.tools.sap/browse/CXSPA-8696

### DIFF
--- a/integration-libs/epd-visualization/root/util/url-utils.spec.ts
+++ b/integration-libs/epd-visualization/root/util/url-utils.spec.ts
@@ -20,9 +20,7 @@ describe('UrlUtils', () => {
     });
 
     it('should return false for non http-based protocol', () => {
-      expect(
-        isHttpOrHttps(new URL('ws://example.com'))
-      ).toBeFalsy();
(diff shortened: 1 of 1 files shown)


integration-libs/epd-visualization/root/util/url-utils.spec.ts
```diff
@@ -20,9 +20,8 @@ describe('UrlUtils', () => {
     });
 
     it('should return false for data: protocol', () => {
-      expect(
-        isHttpOrHttps(new URL('data://data:text/plain,some%20data/'))
-      ).toBeFalsy();
+      const url = getUrl('data:text/plain,some%20data/');
```
sam-garland: The error seems to be an issue with parsing the URL:
`Failed to construct 'URL': Invalid URL`

This could be because of the trailing `/` (which is actually valid in a data URL according to the MDN docs, but might be causing a validation error in a particular browser version).

This line will return `null` when there is an error parsing the URL and the next line will not actually invoke `isHttpOrHttps()` at all.

I will try using a different URL with a different protocol instead.

sam-garland (approved):

Issue #9680, opened 2020-11-06
cx-media does not work with images with different ratio's


janwidmer:
**Describe the bug**

cx-media does not work with images with different ratio's, but the Simple responsive banner is using it exactly like that.

**Tell us the version of Spartacus**

- Feature Level 3.0, library version 3.0.0-next.3

**To Reproduce**

1. Go to the spartacus demo page: https://spartacus-demo.eastus.cloudapp.azure.com/electronics-spa/en/USD/
2. Open the page in the smallest viewports via dev tools
3. Look at the banners just below the navigation "Save big on..."
4. The tablet / desktop rendition get's used on mobile and the text on the image (which should not be on the image due to seo reasons anyway) is not readable

**Expected behavior**

The correct rendition for mobile get's displayed

**Screenshots**

Actual Mobile View:

![image](https://user-images.githubusercontent.com/3381820/98335821-00c90a00-2006-11eb-9369-7045986c1311.png)

Image renditions which should be used on mobile:

![image](https://user-images.githubusercontent.com/3381820/98335949-3d950100-2006-11eb-9c4e-9ec038475e74.png)

![image](https://user-images.githubusercontent.com/3381820/98335846-12121680-2006-11eb-8d04-3b44a0ce8b8d.png)

**Desktop:**

 - OS: Mac / Windows Actual Versions
 - Browser Chrome / Safari via Dev Tools Mobile View
 - Version Actual versions

**Smartphone:**

 - Device: Android / IOS actual versions
 - OS: Actual Versions]
 - Browser chrome / safari
 - Version Actual versions

**Additional context**

The cx-media uses the srcset property on the image tag. This property should receive different renditions of an image consisting all of the **same ratio**. Because the browser knows about the available space for that element and tries to find the best matching rendition also based on the display density (retina or not) and other factors.

If art direction is needed (different ratio's for different viewports), the picture tag needs to be used instead (I think, that's how it was before #2234 was merged.)

The current usage in the accelerator mixes both ways of using images, which leads to the wrong behaviour as described above

**Considerations**

1. The current implementation is a mix of both approaches and makes the `srcset` approach also not work if the author does not know exactly what he needs to do
2. The simple responsive banners of the accelerator should work correctly because as they are now, the component can not be used in reality (components with such a landscape ratio are just too small on mobile) without having art direction
3. As the authors can add 4 different images in the component dialog, we as developers cannot prevent that he adds different ratio's
[...]

PR #19425, opened 2024-10-22
fix: Tweaks to media service

rmch91:
Closes: https://jira.tools.sap/browse/CXSPA-8699


projects/core/src/model/image.model.ts
```diff
@@ -15,6 +15,8 @@ export interface Image {
   galleryIndex?: number;
   imageType?: ImageType;
   url?: string;
+  width?: number;
+  height?: number;
```
Platonn: For a future newcomer reader of our MediaComponent, let's add JSDoc with more context, for those properties, explaining that those properties currently don't come from CMS, but were added allow for potentially setting manually width & height of the images displayed by the MediaComponent.

projects/storefrontlib/shared/components/media/media.component.ts
```diff
@@ -20,6 +20,25 @@ import { ImageLoadingStrategy, Media, MediaContainer } from './media.model';
 import { MediaService } from './media.service';
 import { USE_LEGACY_MEDIA_COMPONENT } from './media.token';
 
+/**
+ * The HTML element rendered in the template can be either `<img>` or `<picture>`,
+ * depending on the input value of `elementType`, which defaults to `'img'`.
+ *
+ * In the case of a `<picture>` element, the `srcset` attribute will always include only one URL.
```
Platonn:
```suggestion
 * In the case of a `<picture>` element, each `<source>` element will contain `srcset` attribute with exactly one URL.
```

projects/storefrontlib/shared/components/media/media.component.ts
```diff
@@ -20,6 +20,25 @@ import { ImageLoadingStrategy, Media, MediaContainer } from './media.model';
...
+ *
+ * For example, instead of:
```
Platonn:
```suggestion
 * If you need multiple URLs per `<source>` (e.g. for different pixel densities),
 * split it into multiple media queries (e.g. with `min-device-pixel-ratio`). See the following example:
```

projects/storefrontlib/shared/components/media/media.component.ts
```diff
@@ -20,6 +20,25 @@ import { ImageLoadingStrategy, Media, MediaContainer } from './media.model';
...
+ *
+ * ```html
+ * <source media="(min-width: 960px)" srcset="image-1400.jpg 1x, image-2800.jpg 2x">
+ * ```
+ *
+ * Use additional formats and additional media queries, such as:
+ *
+ * ```html
+ * <source media="(min-width: 960px)" srcset="image-1400.jpg">
+ * <source media="(min-width: 960px) and (-webkit-min-device-pixel-ratio: 2)" srcset="image-2800.jpg">
```
Platonn: Please help me understand, shouldn't the <source> tags be swapped? I mean, shouldn't the more specific one be the first?
Currently, I understand that the first `<source>` will be used in all cases (regardless of pixel density), because the constraint is too general.
Please correct me if I'm wrong.

projects/storefrontlib/shared/components/media/media.component.ts
```diff
@@ -20,6 +20,25 @@ import { ImageLoadingStrategy, Media, MediaContainer } from './media.model';
...
+ * ```
+ */
```
Platonn, line 44: Very nice JSDoc. IMHO very helpful

projects/storefrontlib/shared/components/media/media.component.ts
```diff
@@ -69,32 +88,21 @@ export class MediaComponent implements OnChanges {
   @Input() elementType: 'img' | 'picture' = 'img';
```
pawelfras, line 91: `query:` I think we haven't discussed one more thing. Do we want the `MediaComponent` to use the `img` tag by default at this point? That would mean that `` would be used in all our components except `BannerComponent`, where `elementType` is explicitly set to `picture`.

projects/storefrontlib/shared/components/media/media.component.ts
rmch91: Yes, you have right. The more specific should be first

projects/storefrontlib/shared/components/media/media.component.ts
rmch91: I thought we agreed on this. @Platonn right?

rmch91:
QA steps:
 
 1. Checkout feat/CXSPA-7700
 2. `ts-node ./tools/schematics/testing.ts` - Publish
 3. Create new angular app `npx -p @angular/cli@17.0.5 ng new test-new-media-cmp --standalone=false`
 4. ng add @spartacus/schematics@latest
 5. Create custom banner component
 
 ```
 @Component({
   selector: 'cx-banner',
   templateUrl: './custom-banner.component.html',
 })
 export class CustomBannerComponent extends BannerComponent {}
 ```
 
 ```
 @NgModule({
   imports: [MediaModule, CommonModule, FormsModule],
   declarations: [CustomBannerComponent],
   exports: [CustomBannerComponent],
 })
 export class CustomBannerModule {}
 ```
 
 6. Override banner component to use custom banner component
 
 ```
     provideConfig(<CmsConfig{
       cmsComponents: {
         SimpleResponsiveBannerComponent: {
           component: CustomBannerComponent,
         },
       },
     }),
 ```
 7. Override getImage(data) method from BannerComponent and add width&height to returned data. 
 8. Check if `<picture>` element sources has width&height. Note that if it is container that includes  `{desktop: ..., mobile: ...}` you need to pass it to specific Image of container.
 9. Good to test also case when in the component is used  tag, for example cx-product-images. So in this case we need to override ProductImages and extend there the container

rmch91:
8. Check if <picture> element sources has width&height. Note that if it is container that includes  {desktop: ..., mobile: ...} you need to pass it to specific Image of container.

projects/core/src/model/image.model.ts
rmch91: Done

projects/storefrontlib/shared/components/media/media.component.ts
rmch91: Done

projects/storefrontlib/shared/components/media/media.component.ts
rmch91: Done

projects/storefrontlib/shared/components/media/media.component.ts
rmch91: Thanks 🙏

projects/storefrontlib/shared/components/media/media.component.ts
Platonn: As far as I understood, in most places we want to use img - which is product images, which don't need the art direction feature.
The only exception is the banner, for which we want to allow for art direction.

See the [comment](https://github.com/SAP/spartacus/issues/9680#issuecomment-723976093) from @janwidmer:
> So in my opinion, in referral to your [documentation](https://github.com/SAP/spartacus-docs/pull/845/files#diff-784794027f7473410b43510dfaf3c0e4474b8bfea2361ab3760e64e72dc44170R33), if you want to keep the freedom of different ratio's, you would need to adjust the cx-media component to render either img tags with srcset (for e.g. product image with same ratio's) or picture tag with source tags (for e.g. banner with different ratio's per viewport).

That said, it would be good to doublecheck my statement and review all usages of `<cx-media>` component in our codebase and double check those places don't need art direction
image

UPDATE:
I've found there only:
- product images
- mini product images (for a multi-d product variant selector)
- customer avatar
- banner

So the above statement is correct. In OOTB Spartacus, only the banner component would like to benefit from art direction. Product images - to my knowledge - always have the same aspect ratio. Same for customer avatar.

projects/storefrontlib/shared/components/media/media.service.ts
```diff
@@ -84,8 +84,9 @@ export class MediaService {
     );
 
     return {
-      ...commonMediaProperties,
       srcset: this.resolveSrcSet(mediaContainer, format),
+      ...commonMediaProperties,
+      ...this.getWidthAndHeight(mediaContainer, format),
```
Platonn: nitpick question: why not move calling `getWidthAndHeight` to inside `getCommonMediaObject`?

Platonn (dismissed):

projects/storefrontlib/shared/components/media/media.service.ts
rmch91: Make sense, thanks. Changed

Platonn (approved):

PR #19432, opened 2024-10-23
fix: utility function for prototype pollution

kpawelczak:
closes: [CXSPA-8633](https://jira.tools.sap/browse/CXSPA-8633)


giancorderoortiz (approved):
ok

PR #19444, opened 2024-10-25
fix: (CXSPA-8196) combobox cannot be dismissed using ‘ESC’

StanislavSukhanov:
* fix combobox cannot be dismissed using ‘ESC’ when screen reader is on

  * add a role to input element of select to get it narrated by a screen reader

  * add subscription to toggle input's element collapsed state

  * closes https://jira.tools.sap/browse/CXSPA-8196


Pio-Bar (dismissed):

projects/storefrontlib/shared/components/ng-select-a11y/ng-select-a11y.directive.ts
```diff
@@ -51,20 +53,47 @@ export class NgSelectA11yDirective implements AfterViewInit {
...
+
   @Optional() breakpointService = inject(BreakpointService, { optional: true });
 
   @Inject(PLATFORM_ID) protected platformId: Object;
 
   constructor(
     private renderer: Renderer2,
-    private elementRef: ElementRef
+    private elementRef: ElementRef,
+    private selectComponent: NgSelectComponent
```
Zeyber: Better to use the `inject(NgSelectComponent` method to add service rather than in our constructor. This is because we have some schematics that require some extra effort to work correctly when we use the constructor method.

projects/storefrontlib/shared/components/ng-select-a11y/ng-select-a11y.directive.ts
```diff
@@ -51,20 +53,47 @@ export class NgSelectA11yDirective implements AfterViewInit {
...
     const divCombobox =
       this.elementRef.nativeElement.querySelector('[role="combobox"]');
     const inputElement = divCombobox.querySelector('input');
 
+    this.renderer.setAttribute(inputElement, 'role', 'combobox');
+    this.renderer.setAttribute(inputElement, 'aria-expanded', 'false');
+
+    const isOpened$ = this.selectComponent.openEvent.pipe(map(() => 'true'));
+    const isClosed$ = this.selectComponent.closeEvent.pipe(map(() => 'false'));
+    merge(isOpened$, isClosed$).subscribe((state) => {
```
Zeyber: Lets tie this subscription to a parent subscription object that we can unsubscribe from on destroy to remove all our subsciptions at the end of this directive's life.

Zeyber (changes requested):

projects/storefrontlib/shared/components/ng-select-a11y/ng-select-a11y.directive.ts
StanislavSukhanov: @Zeyber fixed

projects/storefrontlib/shared/components/ng-select-a11y/ng-select-a11y.directive.ts
StanislavSukhanov: @Zeyber fixed 👍

projects/storefrontlib/shared/components/ng-select-a11y/ng-select-a11y.directive.ts
```diff
@@ -39,6 +43,8 @@ export class NgSelectA11yDirective implements AfterViewInit {
   protected translationService = inject(TranslationService);
   protected domSanitizer = inject(DomSanitizer);
   private featureConfigService = inject(FeatureConfigService);
+  private selectComponent = inject(NgSelectComponent);
```
Zeyber: Lets make these `protected`. We only make `featureConfigService` private because we don't want to expose it to users.

Zeyber (changes requested):

projects/storefrontlib/shared/components/ng-select-a11y/ng-select-a11y.directive.ts
StanislavSukhanov: @Zeyber, thanks for pointing it out. I's fixed.

Zeyber (dismissed):

developpeurweb:
@StanislavSukhanov I tested this PR and it seems all good, except for the fact that when pressing ESC the menu closes and focus resets to the top of the page, instead of going back to the "closed" instance of the menu.
Earlier I tested something similar for ticket CXSPA-8193. After you finish the HCT tickets, can you pls check if the issue is the same, or similar, and if provided solution also fixes your ticket. Thanks.

StanislavSukhanov:
@developpeurweb I've pulled the latest changes including CXSPA-8193. The only difference in behaviour I see (with or without applying re-focusing logic introduced by @Pio-Bar ) is that to activate the dropdown again when the JAWS is enabled I need to press 'enter' key twice. Could you please have a look into this again? Thank you.

developpeurweb (approved):

PR #19448, opened 2024-10-25
feat: GooglePay guest mode support

Matejk00:
Closes: CXSPA-5535


integration-libs/opf/quick-buy/components/opf-quick-buy-buttons/google-pay/google-pay.service.ts
```diff
@@ -342,6 +342,7 @@ export class OpfGooglePayService {
   private handlePaymentCallbacks(): google.payments.api.PaymentDataCallbacks {
     return {
       onPaymentAuthorized: (paymentDataResponse: any) => {
+        console.log(paymentDataResponse);
```
FollowTheFlo: Left over logs

integration-libs/opf/quick-buy/components/opf-quick-buy-buttons/google-pay/google-pay.service.ts
```diff
@@ -378,6 +379,7 @@ export class OpfGooglePayService {
       },
 
       onPaymentDataChanged: (intermediatePaymentData: any) => {
+        console.log(intermediatePaymentData);
```
FollowTheFlo: Left over logs

FollowTheFlo (changes requested):
Need new logic needed about fetch email addr when user decide to go with 'Proceed to Checkout' as seamlessly created guest user does not have it.
Possible solution:
1 - When guest user with no email addr, redirect to Guest User form before reaching 'Shipping' checkout step. --> complex: we could enhance checkout guard, standard Guest user form would not update guest user, it's only creating it, so would need to overwrite cpnt/service of existing Guestuser form .
2 - in Shipping step, add 'email addr' 'confirm emalil addr' required fields if user is guest  && no email addr

Additionnal use case: shop enforcing user login (I believe only for b2b?)
- it is possible handle ootb as createCartGuestUser would failed, still need to catch the error and return of(false)

To minimize above use cases, could we create guest user only after clicking AP/GP ?

FollowTheFlo (approved):
I tested it successfully with diff scenario.
minor enhancement discussed (stale delivery addr delete request returns 401, guest checkout with no email addr able to access checkout step1), 
to be addressed separatly
Code and UX looks good to me,
unit test failing not due to opf, ignoring it.

PR #19451, opened 2024-10-28
feat(a11y): improve My Account payment-details/address-book components readability support

uroslates:
Closes https://jira.tools.sap/browse/CXSPA-8195


developpeurweb:
@uroslates the combination of `aria-label` + `aria-describedby` provides the expected narration. Just a couple of adjustments please for both, Payments and Addresses:

1. Please remove `role="application"`, you probably meant `presentation`, but even that one is not needed, if I remove the role attribute completely the information still gets narrated properly.
2. You will have to generate differentiated IDs for each container, currently, if a user has more than two payment methods and addresses the content of the first card is always narrated for the rest of the cards.

8195 payment first container gets read for all - also remove role application

8195 addresses have same id for all

uroslates:
After conversation with Miguel we decided to use role application and combination of aria-label+aria-describedby (with dynamic id). This combination vocalises both aria-labels and children's content.

developpeurweb:
@uroslates I've tested this with JAWS on my end and it's all good. 
Let's just wait for a developer to review the code before merging. Thanks.

Pio-Bar (dismissed):

Zeyber (dismissed):

Zeyber (approved):

developpeurweb (approved):

PR #19459, opened 2024-10-29
fix(a11y): Focus outline visible for future stock accordion

sdrozdsap:
Closes: https://jira.tools.sap/browse/CXSPA-2586


Pio-Bar (dismissed):

developpeurweb:
@sdrozdsap The cropped focus ring seems to be fixed now, however there's an underlying problem with the focus order here, or perhaps the order is correct but the component's position is not, as it cannot be under the add-to-cat button, which is the last step. When resizing screen to mobile the component moves to the right place and focus order makes sense. 
I believe this misplacement is mainly due to the CSS grid arrangement.
Can you please try repositioning the component. Thanks!

feature-libs/product/future-stock/styles/_future-stock-accordion.scss
```diff
@@ -21,6 +21,15 @@ cx-future-stock-accordion {
     cx-icon {
       padding-inline-start: 0.375rem;
     }
+
+    @include forFeature('a11yVisibleFocusOverflows') {
```
Pio-Bar: I think I've have fixed this one too here: https://github.com/SAP/spartacus/pull/19579/files.
`a11yVisibleFocusOverflows` has been enabled recently since it's 6 moths old. We probably should not add to it anymore. Do you think we can go with the solution I've linked?

Pio-Bar (changes requested):

feature-libs/product/future-stock/styles/_future-stock-accordion.scss
sdrozdsap: Nice catch! Thanks for pointing that out - I'll switch to the `a11yCroppedFocusRing`

Pio-Bar (approved):

Zeyber (approved):

developpeurweb (approved):

PR #19463, opened 2024-10-30
fix: (CXSPA-8200) - Place 'Delete' button before 'Cancel' button in Card Component

petarmarkov9449:
Closes: [CXSPA-8200](https://jira.tools.sap/browse/CXSPA-8200)

Change order of 'Delete' button and 'Cancel' button in Card Component.
Place 'Delete' button before 'Cancel' button.


Pio-Bar (dismissed):

projects/storefrontlib/shared/components/card/card.component.html
```diff
@@ -86,20 +86,44 @@
...
-    <div *ngIf="editMode" class="row cx-card-body-delete">
-      <div class="col-md-6">
-        <button
-          [cxAtMessage]="'assistiveMessage.actionCancelled' | cxTranslate"
-          class="btn btn-block btn-secondary"
-          (click)="cancelEdit()"
-        >
-          {{ 'common.cancel' | cxTranslate }}
-        </button>
+    <div *ngIf="editMode">
```
Zeyber: I could be wrong here looking at the diff. But it looks like you have removed `class="row cx-card-body-delete"` from the `div` with `*ngIf="editMode"` and placed it on the nested `div`. This would be a breaking change. Could you please verify this?

Zeyber (changes requested):

projects/storefrontlib/shared/components/card/card.component.html
petarmarkov9449: I've updated it.

Zeyber (approved):

developpeurweb (approved):

PR #19467, opened 2024-10-31
fix: (CXSPA-8769) HCT themes adjustment to colors and incomplete areas

StanislavSukhanov:
* closes https://jira.tools.sap/browse/CXSPA-8769
  * closes https://jira.tools.sap/browse/CXSPA-8770

### DIFF
--- a/projects/storefrontstyles/scss/components/product/carousel/_carousel.scss
+++ b/projects/storefrontstyles/scss/components/product/carousel/_carousel.scss
@@ -208,12 +208,18 @@
         background-color: var(--cx-color-background);
       }
     }
-    .indicators button[disabled] .cx-icon {
-      color: var(--cx-color-dark);
-      background-color: var(--cx-color-light);
-      padding: 4px;
+
+    .indicators button[aria-disabled='true'] {
+      .cx-icon {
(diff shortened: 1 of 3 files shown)


Pio-Bar (dismissed):

projects/storefrontstyles/scss/components/product/carousel/_carousel.scss
```diff
@@ -208,12 +208,18 @@
         background-color: var(--cx-color-background);
       }
     }
-    .indicators button[disabled] .cx-icon {
```
Zeyber, line 211: I think we should leave the `button[disabled]` case in here with `[aria-disabled]` so there is no breaking change. We can leave a TODO comment to clean it up later.

Zeyber (changes requested):

projects/storefrontstyles/scss/components/product/carousel/_carousel.scss
StanislavSukhanov: @Zeyber fixed;

Zeyber (approved):

developpeurweb (approved):

PR #19476, opened 2024-11-04
feat: Support payment provider icons

Matejk00:
Closes: CXSPA-6402

### DIFF
--- a/integration-libs/opf/checkout/components/opf-checkout-payments/opf-checkout-payments.component.html
+++ b/integration-libs/opf/checkout/components/opf-checkout-payments/opf-checkout-payments.component.html
@@ -33,6 +33,12 @@
       >
         <div class="cx-payment-type">
           {{ configuration.displayName }}
+          <img
+            *ngIf="configuration?.logoUrl"
+            [src]="configuration?.logoUrl"
+            [alt]="configuration.displayName"
+            class="cx-payment-logo"
+          />
         </div>
       </label>
       <cx-opf-checkout-payment-wrapper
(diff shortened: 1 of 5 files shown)


integration-libs/opf/checkout/components/opf-checkout-payments/opf-checkout-payments.component.spec.ts
```diff
@@ -172,4 +177,37 @@ describe('OpfCheckoutPaymentsComponent', () => {
...
+    mockActiveConfigurations.forEach((configuration) => {
+      const logoElement = el.query(
+        By.css(
+          'label[for=paymentId-' + configuration.id + ']  .cx-payment-logo'
+        )
+      );
+
+      if (configuration?.logoUrl) {
+        expect(logoElement).toBeTruthy();
+        expect(logoElement.nativeElement.attributes['class'].value).toBe(
```
rmch91: Nitpick: I think it is redundant check, since you already use it in selector `By.css('label[for=paymentId-' + configuration.id + '] .cx-payment-logo')`. So this class will be always there, or the element will be undefined at all

integration-libs/opf/checkout/components/opf-checkout-payments/opf-checkout-payments.component.html
```diff
@@ -33,6 +33,12 @@ <h3>
       >
         <div class="cx-payment-type">
           {{ configuration.displayName }}
+          <img
+            *ngIf="configuration?.logoUrl"
+            [src]="configuration?.logoUrl"
```
rmch91: `?` could be removed. You already checked that `configuration` exists in `ngIf`

rmch91 (changes requested):

integration-libs/opf/checkout/components/opf-checkout-payments/opf-checkout-payments.component.spec.ts
Matejk00: good catch, totally agree

rmch91 (approved):

PR #19477, opened 2024-11-04
fix: Wrap review order components with section element

Zeyber:
Closes: https://jira.tools.sap/browse/CXSPA-8796


feature-libs/checkout/base/components/checkout-review/checkout-review-overview/checkout-review-overview.component.ts
```diff
@@ -4,19 +4,68 @@
...
+        .translate('checkoutReview.reviewOrder')
+        .pipe(take(1))
+        .subscribe((label) => {
+          // We need to delay for a tick to let components render before querying.
+          setTimeout(() => {
+            // These are the components that we need to wrap.
+            const els: any[] = [
+              this.document?.querySelector('cx-checkout-review-payment'),
+              this.document?.querySelector('cx-checkout-review-overview'),
+              this.document?.querySelector('cx-checkout-review-shipping'),
```
sdrozdsap: Maybe we can define
```ts
protected readonly CHECKOUT_COMPONENTS = [
  'cx-checkout-review-payment',
  'cx-checkout-review-overview',
  'cx-checkout-review-shipping',
];
```
And then use it like this
```suggestion
              const elements = this.CHECKOUT_COMPONENTS.map(selector => 
                this.document?.querySelector(selector)
              )
```
That way it would be easier to find and potentially change the selector, add new component or remove one.

feature-libs/checkout/base/components/checkout-review/checkout-review-overview/checkout-review-overview.component.ts
```diff
@@ -4,19 +4,68 @@
...
+      this.document &&
+      this.translationService &&
+      this.featureService?.isEnabled('a11yWrapReviewOrderInSection')
+    ) {
+      this.translationService
+        .translate('checkoutReview.reviewOrder')
+        .pipe(take(1))
+        .subscribe((label) => {
+          // We need to delay for a tick to let components render before querying.
+          setTimeout(() => {
```
sdrozdsap: Maybe `requestAnimationFrame` would be a bit more suitable here? I think it should work the same way in this case, but I'd prefer `requestAnimationFrame` over `setTimeout`, since I guess timeout has a bit different purpose, however that's just my 2 cents

feature-libs/checkout/base/components/checkout-review/checkout-review-overview/checkout-review-overview.component.ts
```diff
@@ -4,19 +4,68 @@
...
+            // These are the components that we need to wrap.
+            const els: any[] = [
```
sdrozdsap: Do you think you can get rid of the `any`'s here? Maybe just generic `(HTMLElement | null | undefined)[]` will work? Same comment for other `any`'s here

sdrozdsap (changes requested):
The approach with manipulating the DOM directly looks a bit tricky, could you just add a note here why did you go that way? If we want to just have a wrapper section with aria-label, can't we just modify the template containing those components?

Zeyber:
> The approach with manipulating the DOM directly looks a bit tricky, could you just add a note here why did you go that way? If we want to just have a wrapper section with aria-label, can't we just modify the template containing those components?

Thanks for the review! In regards to the approach:

Unfortunately, we have no parent component to add the section element to in the template. Only the page-slot as shown below. So I chose the `CheckoutReviewOverviewComponent` as the most appropriate element (IMO) to nest wrap logic from. I am certainly open to better approaches however. 

image

sdrozdsap (dismissed):

sdrozdsap:
@Zeyber  Thank you for the explanation! LGTM

developpeurweb (dismissed):

sdrozdsap (approved):

PR #19484, opened 2024-11-04
add explicit T and C

FollowTheFlo:


integration-libs/opf/checkout/root/model/opf-checkout.model.ts
```diff
@@ -6,3 +6,6 @@
 
 export const OPF_PAYMENT_AND_REVIEW_SEMANTIC_ROUTE =
   'opfCheckoutPaymentAndReview';
+
+export const OPF_EXPLICIT_T_AND_C_CMS_CPNT =
```
Matejk00: I am having mixed feelings if we should use a shorter names like "T", "C", or "CPNT". It may make some difficulties in case devs wanna to search for the part of the code and start looking for the "component" for example.

Do you mind to use a full name like: `OPF_EXPLICIT_TERMS_AND_CONDITIONS_COMPONENT`?

integration-libs/opf/checkout/components/opf-checkout-terms-and-conditions-alert/opf-checkout-terms-and-conditions-alert.component.ts
```diff
@@ -16,6 +16,7 @@ export class OpfCheckoutTermsAndConditionsAlertComponent {
   iconTypes = ICON_TYPE;
 
   @Input() isVisible: boolean;
+  @Input() isExplicit: boolean | undefined | null;
```
Matejk00, line 23: Could we describe in a few words in a JS doc comment above what is the difference between "explicit" and "implicit"?

Matejk00 (changes requested):

Matejk00 (approved):

PR #19494, opened 2024-11-05
fix: add FullPage-dynamicScript sub-pattern

FollowTheFlo:


integration-libs/opf/checkout/components/opf-checkout-payment-wrapper/opf-checkout-payment-wrapper.service.ts
```diff
@@ -187,7 +189,13 @@ export class OpfCheckoutPaymentWrapperService {
             this.executeScriptFromHtml(html);
           }
         });
+      return;
     }
+    this.handlePaymentInitiationError({
```
FollowTheFlo, line 194: if didn't fall into the 2 above conditions. It is a config prob, display general error message, rather than infinite spinner.

Matejk00 (approved):

PR #19501, opened 2024-11-06
feat: add guest mode for ApplePay

FollowTheFlo:
- Create guest user only after form is displayed, at session validation time. Otherwise getting and error (InvalidAccessError: Must create a new ApplePaySession from a user gesture handler.)
- Add the email addr to guest user cart when payment gets submitted.


integration-libs/opf/quick-buy/components/opf-quick-buy-buttons/apple-pay/apple-pay.service.spec.ts
```diff
@@ -99,7 +99,7 @@ describe('ApplePayService', () => {
         'getCurrentCartTotalPrice',
         'setDeliveryMode',
         'getSelectedDeliveryMode',
-        'deleteUserAddresses',
```
FollowTheFlo, line 102: deleteUserAddresses was twice in the list.

Matejk00 (approved):

PR #19518, opened 2024-11-08
Fix accessibility issue

Jali-Khusi:
List of bugs resolved in this PR:
https://jira.tools.sap/browse/CXSPA-7980
https://jira.tools.sap/browse/CXSPA-7981


vigneshkrish8691 (dismissed):

projects/storefrontlib/cms-components/navigation/navigation/navigation-ui.component.html
```diff
@@ -1,4 +1,4 @@
-<nav [attr.aria-label]="navAriaLabel">
+<nav [attr.aria-label]="navAriaLabel" [attr.role]="'navigation'">
```
Pio-Bar: The semantic `nav` element makes the use of `role` attribute redundant. It provides the same result.

projects/storefrontlib/cms-components/navigation/navigation/navigation-ui.component.html
```diff
@@ -8,7 +8,7 @@
       *ngIf="flyout && (node?.children?.length ?? 0) > 1"
       class="back is-open"
     >
-      <button (click)="back()">
+      <button (click)="back()" [attr.aria-label]="'common.back' | cxTranslate">
```
Pio-Bar: The back button already has proper text inside it. There is no benefit to give it the same text as a label.

projects/storefrontlib/cms-components/navigation/navigation/navigation-ui.component.html
```diff
@@ -68,7 +65,6 @@
         [tabindex]="getTabIndex(node, depth)"
         (focus)="depth || reinitializeMenu()"
         (keydown.space)="toggleOpen($any($event))"
-        (keydown.esc)="back()"
```
Pio-Bar: Removing this call without an alternative would result in keyboard controls being negatively affected.

projects/storefrontlib/cms-components/navigation/navigation/navigation-ui.component.html
```diff
@@ -125,12 +120,7 @@
         </ng-container>
       </ng-container>
       <ng-template #title>
-        <span
-          role="heading"
-          aria-level="4"
-          *ngIf="node.title"
-          [attr.tabindex]="-1"
-        >
+        <span role="presentation" *ngIf="node.title" [attr.tabindex]="0">
```
Pio-Bar: This text is used as a heading for the following elements and should be recognized as one. Giving it a role of `presentation` would hide it from a11y software, which is not allowed for elements that convey information visible on the screen.

Elements that cannot be interacted with directly should not have tabindex of 0

projects/storefrontlib/cms-components/navigation/navigation/navigation-ui.component.html
```diff
@@ -147,6 +137,9 @@
         [attr.depth]="getTotalDepth(node)"
         [attr.wrap-after]="node.children.length > wrapAfter ? wrapAfter : null"
         [attr.columns]="getColumnCount(node.children.length)"
+        [attr.role]="
+          flyout && (node?.children?.length ?? 0) > 1 ? 'presentation' : 'list'
```
Pio-Bar: The `presentation` role would hide the list from a11y software.

projects/storefrontstyles/scss/components/content/navigation/_my-account-v2-navigation.scss
```diff
@@ -53,3 +56,26 @@
     display: inline-block;
   }
 }
+
+@media screen and (max-width: 544px) {
+  cx-navigation-ui h4,
+  span {
+    width: 100% !important;
+    font-size: 14px;
+  }
```
Pio-Bar: - We may need to keep these within the component's scope. 
- There should be a better way to overwrite these styles then using the nuclear option of `!important`.
- We probably should stick to the breakpoints established by Spartacus already.

Pio-Bar (changes requested):
These changes would be considered breaking. We should put them behind a feature toggle like other a11y ones. They can be recognized by the `a11y` prefix we include for each one.

projects/storefrontlib/cms-components/navigation/navigation/navigation-ui.component.html
```diff
@@ -126,10 +122,9 @@
...
       <ng-template #title>
         <span
-          role="heading"
-          aria-level="4"
+          [attr.aria-label]="node.title"
           *ngIf="node.title"
-          [attr.tabindex]="-1"
+          [attr.tabindex]="0"
         >
           {{ node.title }}
```
Pio-Bar: - The `aria-label` is unnecessary since the element itself contains the exact same text.
- I think this element was meant as a heading displaying the nodes title. Are we getting rid of that extra information?
- Tabbing order changes may affect the keyboard controls of the navigation. Are we giving focus to just a pice of text? Does not look like it's intractable.

projects/storefrontlib/cms-components/navigation/navigation/navigation-ui.component.html
```diff
@@ -147,6 +142,9 @@
         [attr.depth]="getTotalDepth(node)"
         [attr.wrap-after]="node.children.length > wrapAfter ? wrapAfter : null"
         [attr.columns]="getColumnCount(node.children.length)"
+        [attr.role]="
+          flyout && (node?.children?.length ?? 0) > 1 ? 'list' : null
```
Pio-Bar: Do we want the role to be blank for non flyout navigation? I think it can still contain list items?

Pio-Bar (changes requested):
Some of my older concerns are still valid, Please let me know if you have any doubts, thanks.

projects/storefrontlib/cms-components/navigation/navigation/navigation-ui.component.html
```diff
@@ -56,8 +55,6 @@
             | cxTranslate
               : {
                   title: node.title,
-                  position,
-                  listLength,
```
Pio-Bar: Having a proper list count is one of the requirements we need to meet. Let's make sure it's preserved where needed.
https://jira.tools.sap/browse/CXSPA-8035

projects/assets/src/translations/en/common.json
```diff
@@ -63,7 +63,10 @@
     "goTo": "Go to {{location}}",
     "navigateTo": "Navigate to {{nav}}",
     "scrollToTop": "Scroll back to the top of the page",
-    "linkItemInList": "{{title}}. {{position}} of {{listLength}}"
+    "linkItemInList": "{{title}}",
+    "linkItemInLists": "{{title}}. {{position}} of {{listLength}}",
+    "navigation": "navigation",
+    "mainNavigation": "Main Navigation"
```
Pio-Bar: Seems like the uses of some of these have been removed. We don't need them probably

projects/storefrontlib/cms-components/navigation/navigation/navigation-ui.component.html
Jali-Khusi: Thank you for the feedback. Regarding the JIRA ticket https://jira.tools.sap/browse/CXSPA-7979 , I addressed the issue where additional information like "list with 1 item, list with 3 items" was being read out when accessing the ‘Saved Carts’ list item within the ‘Order Information’ group. I’ve updated the aria-label to ensure only relevant information is announced, removing redundant details. The list count is preserved as needed, but without unnecessary repetition for screen readers.

Let me know if further adjustments are needed

projects/storefrontlib/cms-components/navigation/navigation/navigation-ui.component.html
Jali-Khusi: The issue was that group labels like "Order Information" and "Account Information" were not being read when accessing the list items within their respective groups. To resolve this, I updated the tabindex to 0 to ensure these group labels are focusable and properly announced.

I also removed role="heading" and aria-level="4" to avoid unnecessary additional information being read out by screen readers

Let me know if further adjustments are needed

https://jira.tools.sap/browse/CXSPA-7980

projects/storefrontlib/cms-components/navigation/navigation/navigation-ui.component.html
Pio-Bar: The removed translation seems not to be related to the one mentioned in the ticket. ("list with 1 item, list with 3 items"). Each option should include its count within the list ( _n_ of _n options_). However, if this case is different an exception would need to be made for it. We cannot remove this from other implementations as this is a functionality we need to upkeep.

projects/storefrontlib/cms-components/navigation/navigation/navigation-ui.component.html
```diff
@@ -106,7 +104,6 @@
         </ng-container>
         <ng-container *cxFeature="'a11yNavigationUiKeyboardControls'">
           <button
-            aria-level="4"
```
Pio-Bar: If the role is set to `heading` we need to specify the level too.

projects/storefrontlib/cms-components/navigation/navigation/navigation-ui.component.html
```diff
@@ -131,12 +128,7 @@
         </ng-container>
       </ng-container>
       <ng-template #title>
-        <span
-          role="heading"
-          aria-level="4"
-          *ngIf="node.title"
-          [attr.tabindex]="-1"
-        >
```
Pio-Bar: These attributes are required by the main navigation in the header and provide a good a11y experience for that use case. We need to satisfy both cases, we cannot remove one in favour of the other.

projects/storefrontstyles/scss/components/content/navigation/_navigation-ui.scss
```diff
@@ -229,8 +229,6 @@
         padding-top: 10px;
         padding-bottom: 12px;
         padding-inline-start: 0;
-        margin-top: 10px;
-        margin-bottom: 10px;
```
Pio-Bar: I am afraid these changes will influence the look of other implementations as well.

projects/storefrontstyles/scss/layout/page-templates/_myaccount-v2-page.scss
```diff
@@ -17,3 +17,40 @@
     max-width: 75%;
   }
 }
+
+@media screen and (max-width: 1024px) {
+  .LeftContentSlot {
+    max-width: 26% !important;
```
Pio-Bar: We should not use the `!important` to override styles.

Pio-Bar (changes requested):
I do have concerns about some of the changes being made to the `navigation-ui` component. This component is reused within Spartacus for quite important features like the main navigation. We should be especially careful modifying it since the changes impact other implementations. Some a11y improvements have been made already and if we are going to introduce new ones they can't be to the detriment of the existing ones.

I've brought to attention that some of these changes are considered breaking. We need to properly handle those by using feature flags, this mitigates any possible problems our clients might face. More information can be found on the [wiki](https://wiki.one.int.sap/wiki/pages/viewpage.action?spaceKey=spar&title=Spartacus+Onboarding#SpartacusOnboarding-Breakingchanges) and [here](https://github.com/SAP/spartacus/pull/18523) is a nice writeup about the new FF system for CSS changes.

Feel free to reach out to me directly or any a11y team members for help!

projects/storefrontlib/cms-components/navigation/navigation/navigation-ui.component.html
Jali-Khusi: Hi @Pio-Bar 
Thank you for your message. As per my understanding, I have implemented the changes accordingly. Could you kindly suggest if there is an alternative approach I could take that would allow me to make adjustments without affecting the current functionality?

projects/storefrontstyles/scss/components/content/navigation/_navigation-ui.scss
Jali-Khusi: I removed the margin-top: 10px; and margin-bottom: 10px; because they were causing some unwanted gaps between the list items. If you'd prefer, I can revert this change
gap-list

Pio-Bar: Here is screenshot from the current `bugs-acc-issue` branch. If we remove the margins the main navigation will shift. 

I understand this fixes the issue for this particular case. At the same time we can not break other implementations. These changes should be kept within the affected component.

Screenshot 2025-01-24 at 10 17 39

projects/storefrontstyles/scss/components/content/navigation/_navigation-ui.scss
```diff
@@ -321,6 +319,7 @@
 
 %cx-navigation-ui {
   display: flex;
+  display: contents;
```
Pio-Bar: This is a significant change, not sure why flex would be overwritten here. This will surely affect the layout.

image

Pio-Bar (changes requested):

projects/storefrontlib/cms-components/myaccount/my-account-v2/my-account-v2-navigation/my-account-v2-navigation.component.ts
```diff
@@ -6,9 +6,15 @@
...
 @Component({
   selector: 'cx-my-account-v2-navigation',
   templateUrl: './my-account-v2-navigation.component.html',
 })
-export class MyAccountV2NavigationComponent extends NavigationComponent {}
+export class MyAccountV2NavigationComponent extends NavigationComponent {
+  navAriaLabel$: Observable<any> = this.name$.pipe(
+    map((name: any) => name ?? 'MyAccount View Side Navigation')
+  );
+}
```
Pio-Bar: Changing the CMS data would modify the outcome of the logic based on this. If the name was added the count would be present too. It's probably not what we want. 

We should not use a hard coded string for comparison.

projects/storefrontlib/cms-components/navigation/navigation/navigation-ui.component.html
```diff
@@ -52,13 +52,10 @@
...
         "
         [ariaLabel]="
-          'navigation.linkItemInList'
-            | cxTranslate
-              : {
-                  title: node.title,
-                  position,
-                  listLength,
-                }
+          navAriaLabel !== ('navigation.myAccountSideNavigation' | cxTranslate)
```
Pio-Bar: We probably should not base logic on a translation. Maybe we can include `provideOptionCount` input set to `true` by default. This way we would have an easy way to disable the option count where needed.

projects/storefrontlib/cms-components/navigation/navigation/navigation-ui.component.html
```diff
@@ -132,10 +129,25 @@
...
+              : null
+          "
           *ngIf="node.title"
-          [attr.tabindex]="-1"
+          [attr.tabindex]="
+            navAriaLabel !==
+            ('navigation.myAccountSideNavigation' | cxTranslate)
+              ? -1
+              : 0
+          "
```
Pio-Bar: Should the title be stripped of its semantic value? As mentioned previously, role `presentation` hides the element from accessibility software. [In this case it's ignored because the](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/presentation_role#:~:text=Browsers%20ignore%20role%3D%22presentation%22%20and%20role%3D%22none%22%20on%20focusable%20elements%2C%20including%20link%20and%20inputs%2C%20or%20anything%20with%20a%20tabindex%20attribute%20set.) tab index is 0. If these are a11y improvements, the intention here is not clear to me. Are we reducing the title to just a pice of text in this case?

Pio-Bar (changes requested):

suprishi:
Hi @Pio-Bar, I've renamed the input variable to a generic name `a11yOptionCount` and also added the documentation for it and set its default value to `false` in `NavigationUIComponent`. And the issues addressed in this PR are implemented conditionally so that they do not impact wherever they are reused.
Please have a look again and let me know if any additional changes are required.

projects/assets/src/translations/en/common.json
```diff
@@ -68,6 +68,7 @@
     "linkItemInList": "{{title}}. {{position}} of {{listLength}}",
     "menuButonTitle": "{{title}} Menu"
   },
+
```
Pio-Bar:
```suggestion

```

projects/storefrontlib/cms-components/navigation/navigation/navigation-ui.component.html
```diff
@@ -155,7 +155,7 @@
           role="heading"
           aria-level="4"
           *ngIf="node.title"
-          [attr.tabindex]="-1"
+          [attr.tabindex]="a11yOptionCount ? 0 : -1"
```
Pio-Bar: Is the title getting focus relevant to the option count? What if a developer only needs to disable the count itself? We should avoid a side effect here. Making the title intractable should be a separate toggle.

projects/storefrontlib/cms-components/navigation/navigation/navigation-ui.component.ts
```diff
@@ -56,6 +56,11 @@ export class NavigationUIComponent implements OnInit, OnDestroy {
    */
   @Input() resetMenuOnClose: boolean | undefined;
 
+  /**
+   * Flag indicates whether to set option count in aria-label.
+   */
+  @Input() a11yOptionCount: boolean = false;
```
Pio-Bar: Looks like the code actually disabling the option count has been removed within [this commit](https://github.com/SAP/spartacus/pull/19518/commits/924374ae933f9d46348329c5ef0db6ff9e83984b). Is that part no longer relevant? If so, setting up `a11yOptionCount` input might not be needed anymore.

projects/storefrontstyles/scss/components/content/navigation/_navigation-ui.scss
```diff
@@ -321,6 +321,8 @@
 
 %cx-navigation-ui {
   display: flex;
+  flex-direction: column;
+  width: 100%;
```
Pio-Bar: Some changes might be considered breaking. I have mentioned this in the PR comments previously. See [here](https://github.com/SAP/spartacus/pull/19518#pullrequestreview-2549576614) for more information.

Pio-Bar (changes requested):

projects/storefrontlib/cms-components/myaccount/my-account-v2/my-account-v2-navigation/my-account-v2-navigation.component.ts
```diff
@@ -12,4 +12,6 @@ import { NavigationComponent } from '../../../navigation';
   templateUrl: './my-account-v2-navigation.component.html',
   standalone: false,
 })
-export class MyAccountV2NavigationComponent extends NavigationComponent {}
+export class MyAccountV2NavigationComponent extends NavigationComponent {
+  isMyAccount = true;
+}
```
Pio-Bar:
```suggestion
export class MyAccountV2NavigationComponent extends NavigationComponent {}
```

Looks like this variable does not have a purpose, we should be able to remove it.

projects/storefrontlib/cms-components/myaccount/my-account-v2/my-account-v2-navigation/my-account-v2-navigation.component.html
```diff
@@ -4,4 +4,5 @@
   [flyout]="false"
   [ngClass]="(styleClass$ | async) ?? ''"
   [navAriaLabel]="name$ | async"
+  [a11yIntraceableNavSectionTitle]="isMyAccount"
```
Pio-Bar:
```suggestion
  [a11yIntraceableNavSectionTitle]="true"
```

projects/storefrontlib/cms-components/navigation/navigation/navigation-ui.component.ts
```diff
@@ -56,6 +56,11 @@ export class NavigationUIComponent implements OnInit, OnDestroy {
    */
   @Input() resetMenuOnClose: boolean | undefined;
 
+  /**
+   * Flag indicates whether to focus the section titles on tab.
+   */
+  @Input() a11yIntraceableNavSectionTitle: boolean = false;
```
Pio-Bar:
```suggestion
  /**
   * Include non intractable node titles within the tabbing order. 
   */
  @Input() focusableNodeTitles: boolean = false;
```

Let's remove the typo from the input name. We should make our intentions more clear.

Pio-Bar (changes requested):

Pio-Bar (approved):

PR #19520, opened 2024-11-11
fix: style Get Directions link as link

uroslates:
Closes https://jira.tools.sap/browse/CXSPA-8814


feature-libs/storefinder/components/store-finder-list-item/store-finder-list-item.component.html
```diff
@@ -50,7 +50,7 @@
       href="{{ getDirections(location) }}"
       target="_blank"
       rel="noopener noreferrer"
-      class="btn btn-sm btn-secondary btn-block cx-button"
```
Pio-Bar, line 53: A small change, but I think it's still considered breaking. Let's use a feature flag here.

Pio-Bar (changes requested):

feature-libs/storefinder/components/store-finder-list-item/store-finder-list-item.component.html
uroslates: Should be addressed. Added new feature flag for these kind of changes.

projects/core/src/features-config/feature-toggles/config/feature-toggles.ts
```diff
@@ -797,6 +804,7 @@ export const defaultFeatureToggles: Required<FeatureTogglesInterface> = {
   a11yFacetKeyboardNavigation: true,
   a11yUnitsListKeyboardControls: true,
   a11yCartItemsLinksStyles: true,
+  a11yStyleExternalLinksAsLinks: true,
```
Pio-Bar: New feature flags need to be turned off by default for the first 6 months.

Pio-Bar (changes requested):

projects/core/src/features-config/feature-toggles/config/feature-toggles.ts
uroslates: OK thanks for the info. Should be updated now!

Pio-Bar (approved):

Zeyber (approved):

developpeurweb (approved):

PR #19528, opened 2024-11-12
fix: (CXSPA-8813) B2C button My Store not properly styled

StanislavSukhanov:
* replace cx-card-actions buttons with links

* https://jira.tools.sap/browse/CXSPA-8813

### DIFF
--- a/feature-libs/pickup-in-store/components/container/my-preferred-store/my-preferred-store.component.html
+++ b/feature-libs/pickup-in-store/components/container/my-preferred-store/my-preferred-store.component.html
@@ -55,6 +55,30 @@
               [storeDetails]="preferredStore$ | async"
             ></cx-store-schedule>
           </div>
+          <div class="action-buttons">
+            <ng-container *ngFor="let action of actions">
+              <ng-container [ngSwitch]="action.event">
+                <ng-container *ngSwitchCase="'send'">
+                  <a
+                    tabindex="0"
+                    class="link cx-action-link"
+                    (keydown.enter)="getDirectionsToStore()"
+                    (click)="getDirectionsToStore()"
+                    [attr.aria-label]="
+                      action.ariaLabel ? (action.ariaLabel | cxTranslate) : null
+                    "
+                  >
+                    {{ action.name }}
+                  </a>
+                </ng-container>
+                <ng-container *ngSwitchCase="'edit'">
+                  <button class="btn btn-tertiary" (click)="changeStore()">
+                    {{ action.name }}
+                  </button>
+                </ng-container>
+              </ng-container>
+            </ng-container>
+          </div>
         </div>
       </cx-card>
     </div>
--- a/feature-libs/pickup-in-store/components/container/my-preferred-store/my-preferred-store.component.ts
+++ b/feature-libs/pickup-in-store/components/container/my-preferred-store/my-preferred-store.component.ts
@@ -22,6 +22,12 @@ import { ICON_TYPE } from '@spartacus/storefront';
 import { Observable } from 'rxjs';
 import { filter, map, switchMap, take, tap } from 'rxjs/operators';
 
+const getDirectionsAction = {
+  event: 'send',
+  name: 'Get Directions',
+  ariaLabel: 'cardActions.getDirections',
+};
+
 @Component({
   selector: 'cx-my-preferred-store',
   templateUrl: 'my-preferred-store.component.html',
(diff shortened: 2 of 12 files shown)


Zeyber (dismissed):

developpeurweb:
@StanislavSukhanov can you please look into the following issues:
- While testing with keyboard-only navigation, the link element gets skipped from the Tabbing sequence (focus jumps from "Home" breadcrumb to "Search" field).
- When activating the link with the mouse, there is dark background, please remove it and leave as normal link behaviour.
- Add an `aria-label` to the link to announce screen reader users that it will open in a new window, similar to screenshot below
aria label opens in new windows

projects/storefrontlib/shared/components/card/card.component.ts
```diff
@@ -7,14 +7,17 @@
 import { Component, EventEmitter, Input, OnInit, Output } from '@angular/core';
 import { ICON_TYPE } from '../../../cms-components/misc/icon/icon.model';
 
-export interface CardAction {
```
Pio-Bar: Changing or removing interfaces is considered a [breaking change](https://wiki.one.int.sap/wiki/display/spar/Spartacus+Onboarding#:~:text=Changing%20or%20removing%20interface%20properties.).

Pio-Bar (dismissed):

feature-libs/pickup-in-store/components/container/my-preferred-store/my-preferred-store.component.html
```diff
@@ -55,6 +55,30 @@
               [storeDetails]="preferredStore$ | async"
             ></cx-store-schedule>
           </div>
+          <div class="action-buttons">
```
Pio-Bar, line 58: Let's include the feature flag here

feature-libs/pickup-in-store/components/container/my-preferred-store/my-preferred-store.component.ts
```diff
@@ -31,11 +37,14 @@ export class MyPreferredStoreComponent implements OnInit {
   preferredStore$: Observable<PointOfService>;
   content = {
     header: 'My Store',
-    actions: [
-      { event: 'send', name: 'Get Directions' },
-      { event: 'edit', name: 'Change Store' },
-    ],
   };
+  actions = [
```
Pio-Bar, line 41: I think this can be considered breaking too, let's try the `FeatureConfigService` here

feature-libs/pickup-in-store/styles/_my-preferred-store.scss
```diff
@@ -1,6 +1,4 @@
 %cx-my-preferred-store {
-  @import 'bootstrap/scss/buttons';
```
Pio-Bar, line 2: If it does not cause problems I think we should leave this in case of breaking changes. Since, an import cant be toggled with a feature flag let's leave a comment asking to remove it once it's made permanent.

projects/storefrontlib/shared/components/card/card.component.html
```diff
@@ -131,46 +131,84 @@
     </div>
     <!-- Actions -->
     <div *ngIf="content.actions && !editMode" class="cx-card-actions">
-      <div *ngFor="let action of content.actions">
+      <div *ngFor="let action of content.actions" class="cx-card-action-item">
```
Pio-Bar, line 134: Let's include this change within the feature flag too.

projects/storefrontlib/shared/components/card/card.component.html
```diff
@@ -131,46 +131,84 @@
...
         <div>
```
Pio-Bar, line 135: Do you think we can get rid of this extra \<div\>, not sure if it serves a purpose? Let's leave it if it's extra work ;)

projects/core/src/features-config/feature-toggles/config/feature-toggles.ts
```diff
@@ -649,6 +649,11 @@ export interface FeatureTogglesInterface {
    */
   a11yViewHoursButtonIconContrast?: boolean;
 
+  /**
+   * `Cart component` improve action buttons appearance
+   */
+  a11yImproveCardActionButtons?: boolean;
```
Pio-Bar, line 655: [We should make this description more accurate](https://wiki.one.int.sap/wiki/display/spar/Coding+Guidelines#CodingGuidelines-WriteaJSDocdescriptionofaFeatureToggle:~:text=non%2Dvoid%20elements.-,2.18.%20Use%20Feature%20Toggles%20when%20changing%20behavior%20of%20an%20existing%20feature,-When%20introducing%20breaking). The changes to my-preferred-store.component.html feel significant enough

Pio-Bar (changes requested):

PR #17250, opened 2023-04-11
feat: add OPF integration library

Matejk00:

PR #19536, opened 2024-11-13
(DO NOT MERGE) diff between 2 checkpoints of epic/opf branch, but without changes from devleop

Platonn:


feature-libs/cart/base/occ/adapters/occ-cart-guest-user.adapter.ts
```diff
@@ -0,0 +1,99 @@
...
+      new HttpHeaders({
+        ...CONTENT_TYPE_JSON_HEADER,
+      })
+    );
+  }
+
+  createCartGuestUser(
+    userId: string,
+    cartId: string,
+    guestUserDetails?: CartGuestUser
```
Platonn, line 48: [MEDIUM]
Allowing here for the full CartGuestUser looks misleading to me.
In the swagger docs I can see the only allowed property is `email`, but `uid` and `name` cannot be passed in the payload in the request to OCC.
see
https://api.cp96avkh5f-integrati2-s1-public.model-t.cc.commerce.ondemand.com/occ/v2/swagger-ui/index.html#/Carts/createCartGuestUser
and
https://api.cp96avkh5f-integrati2-s1-public.model-t.cc.commerce.ondemand.com/occ/v2/swagger-ui/index.html#/Carts/updateCurrentUserProfile

Note: analogical remark applies to both methods (create and update), and to all the layers (from service/facade, through connector to adapter/occ-adapter)

feature-libs/cart/base/root/facade/cart-guest-user.facade.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  /**
+   * Updates guest user details.
+   */
+  abstract updateCartGuestUser(
+    userId: string,
+    cartId: string,
+    guestUserDetails: CartGuestUser
+  ): Observable<CartGuestUser>;
+}
```
Platonn, line 40: [NITPICK]
When the class name describes a domain (e.g. CartGuestUser), IMHO there is no need to repeat this domain name in each method. So e.g. `CartGuestUserService.create()` vs `CartGuestUserService.createCartGuestUser()` - the first one is more concise and easier to read, yet still precise and correct.

As an example, see [UserProfileFacade](https://github.com/SAP/spartacus/blob/d0b209dae77cd503ed429e3dad1bb46ebfe0af19/feature-libs/user/profile/root/facade/user-profile.facade.ts#L23).

projects/schematics/src/shared/lib-configs/integration-libs/opf-schematics-config.ts
```diff
@@ -70,46 +159,92 @@ export const OPF_CHECKOUT_SCHEMATICS_CONFIG: SchematicConfig = {
...
-    },
-  ],
   customConfig: buildOpfConfig,
+  dependencyFeatures: [
+    OPF_PAYMENT_FEATURE_NAME,
+    OPF_BASE_FEATURE_NAME,
+    OPF_CTA_FEATURE_NAME,
+    OPF_GLOBAL_FUNCTIONS_FEATURE_NAME,
+    OPF_QUICK_BUY_FEATURE_NAME,
+  ],
```
Platonn, line 169: [QUESTION]
Do we really need those `dependencyFeatures` being declared here?
If yes, then why don't we declare them for other sections like OPF_CTA_SCHEMATICS_CONFIG?

I guess it's a "trick" to install all features at once under the "umbrella" of the OPF Checkout, if a customer chooses to install OPF Checkout. And I can see in the main installation schematics (for `ng add spartacus...`, we have just one option to choose for the whole OPF - `OPF-Checkout`, see: https://github.com/SAP/spartacus/blob/d0b209dae77cd503ed429e3dad1bb46ebfe0af19/projects/schematics/src/add-spartacus/schema.json#L37

projects/storefrontapp/src/app/spartacus/spartacus-b2c-configuration.module.ts
```diff
@@ -17,6 +17,7 @@ import { environment } from '../../environments/environment';
 
 const defaultBaseSite = [
   'electronics-spa',
+  'electronics-spa-standalone',
```
Platonn, line 20: [QUESTION]
I'm lacking context and happy to learn: Why `electronics-spa-standalone` is special that we needed to add it for the OPF epic PR?

Matejk00: Yes, this basesite will be dedicated for the OPF checkout testing.

integration-libs/opf/base/core/services/opf-endpoints.service.ts
```diff
@@ -7,20 +7,22 @@
...
   providedIn: 'root',
 })
 export class OpfEndpointsService {
   private _activeBaseSite: string;
 
   constructor(
-    private config: Config,
-    private baseSiteService: BaseSiteService
+    protected opfConfig: OpfConfig,
+    protected opfApiConfig: OpfApiConfig,
```
Platonn, line 24: [MEDIUM]
Do we really need the config injected 2 times? Isn't it the same object in the end of the day?

integration-libs/opf/base/root/services/opf-global-message.service.spec.ts
```diff
@@ -0,0 +1 @@
+// to be done
```
Platonn, line 1: [MEDIUM]
It's a leftover I guess. Let's add tests

integration-libs/opf/base/root/services/opf-global-message.service.ts
```diff
@@ -0,0 +1,65 @@
...
+  StateWithGlobalMessage,
+  Translatable,
+} from '@spartacus/core';
+import { timer } from 'rxjs';
+import { take } from 'rxjs/operators';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class OpfGlobalMessageService extends GlobalMessageService {
```
Platonn, line 21: [MINOR]
This service is a workaround for an issue that is not obvious. Let's explain it in the JSDocs what would happen in OPF, if we didn't have this service.

integration-libs/opf/checkout/assets/translations/en/opfCheckout.json
```diff
@@ -12,7 +12,8 @@
     "itemsToBeShipped": "Items to be shipped",
     "proceedPayment": "Place Order",
     "retryPayment": "Retry to Continue",
-    "checkTermsAndConditionsFirst": "You must agree Terms & Conditions to see available payment options.",
+    "checkTermsAndConditionsExplicit": "You must agree Terms & Conditions to see available payment options.",
+    "checkTermsAndConditionsImplicit": "By submitting the order, you are aggreeing to our Terms & Conditions.",
```
Platonn, line 16: [MEDIUM]
typo `aggreeing` vs `agreeing`

integration-libs/opf/checkout/components/opf-checkout-payment-and-review/opf-checkout-payment-and-review.component.ts
```diff
@@ -78,6 +96,10 @@ export class OpfCheckoutPaymentAndReviewComponent
     );
   }
 
+  protected isCmsComponentInPage(cmsComponentUid: string, page: Page): boolean {
+    return !!page && JSON.stringify(page).includes(cmsComponentUid);
```
Platonn, line 100: [MEDIUM]
I can understand we took shortcut here. But is it really justified?

Coudln't we properly iterate over the Slots/Components array inside the `Page` model and to find if `cmsComponentId` is there?

Theoretically the expected string might be present in some other part of the JSON (not component ID), causing a false positive.

integration-libs/opf/cta/components/opf-cta-element/opf-cta-element.component.ts
```diff
@@ -5,23 +5,45 @@
...
+
+  ngAfterViewInit(): void {
+    this.opfCtaScriptsService.loadAndRunScript(this.ctaScriptHtml);
+  }
   renderHtml(html: string): SafeHtml {
-    return this.sanitizer.bypassSecurityTrustHtml(html);
+    // Display sanitized html in SSR for security concerns
+    return this.windowRef.isBrowser()
+      ? this.sanitizer.bypassSecurityTrustHtml(this.removeScriptTags(html))
+      : html;
```
Platonn, line 38: [MEDIUM]
I'm having problem to understand what threat this line of code prevents in SSR. Is it _really_ needed?
Even if HTML is rendered into DOM, the NodeJS won't execute it. It's only the browser that will execute it. SSR with only output it as a big HTML string sent in the response to the browser.

integration-libs/opf/cta/components/opf-cta-scripts/opf-cta-scripts.service.ts
```diff
@@ -99,84 +127,33 @@ export class OpfCtaScriptsService {
...
+        this.opfStaticCtaService.fillCtaRequestforPagesWithOrder(
+          scriptsLocation
+        ),
       [CtaScriptsLocation.ORDER_CONFIRMATION_PAYMENT_GUIDE]: () =>
-        this.fillCtaRequestforPagesWithOrder(scriptsLocation),
-      [CtaScriptsLocation.CART_MESSAGING]: toBeImplementedException,
-      [CtaScriptsLocation.CART_QUICK_BUY]: toBeImplementedException,
-      [CtaScriptsLocation.CHECKOUT_QUICK_BUY]: toBeImplementedException,
-      [CtaScriptsLocation.PDP_MESSAGING]: toBeImplementedException,
+        this.opfStaticCtaService.fillCtaRequestforPagesWithOrder(
```
Platonn, line 141: [MEDIUM]
PascalCase typo `Requestfor` vs `RequestFor`

integration-libs/opf/cta/core/services/opf-dynamic-cta.service.ts
```diff
@@ -0,0 +1,251 @@
...
+              })
+            );
+          }
+        })
+      )
+      .subscribe();
+  }
+
+  protected productChangedListener(): void {
+    //// need event listener detecting 'pdp counter changed' event and get its qty value.
```
Platonn, line 236: [MEDIUM]
Leftover?

integration-libs/opf/cta/core/services/opf-dynamic-cta.service.ts
```diff
@@ -0,0 +1,251 @@
...
+        GlobalFunctionsDomain.GLOBAL
+      );
+      this.isOnsiteMessagingInit = false;
+    }
+  }
+
+  protected getNewScriptIdentifier(): string {
+    this.scriptIdentifiers.push(String(this.scriptIdentifiers.length + 1));
+    return String(
+      this.scriptIdentifiers[this.scriptIdentifiers.length - 1]
```
Platonn, line 158: [MINOR]
I'm having hard time understanding this code.
- Why +1 in one place, but -1 in another place.
- Why wrapping a value from `scriptIdentifiers[i]` within a `String()` again - in line 158? All values inside this array are already strings, arent they?

integration-libs/opf/global-functions/core/facade/opf-global-functions.service.ts
```diff
@@ -39,17 +40,24 @@ export class OpfGlobalFunctionsService implements OpfGlobalFunctionsFacade {
...
+  readyForScriptEvent$: Observable<string> =
+    this._readyForScriptEvent.asObservable();
 
   registerGlobalFunctions({
     domain,
     paymentSessionId,
     vcr,
     paramsMap,
   }: GlobalFunctionsInput): void {
+    // SSR not supported
```
Platonn, line 57: [MINOR]
Let's explain why it's not supported
a) it's not suported YET (but we plan to add a support)?
vs
b) it's not supported by design - in SSR we don't want to register global functions, becasue they are meant only for the user interactions in the browser

Note: analogical remark applies to other place in this file

integration-libs/opf/global-functions/core/facade/opf-global-functions.service.ts
```diff
@@ -39,17 +40,24 @@ export class OpfGlobalFunctionsService implements OpfGlobalFunctionsFacade {
...
   protected opfPaymentFacade = inject(OpfPaymentFacade);
   protected launchDialogService = inject(LaunchDialogService);
-
+  protected opfCtaFacade = inject(OpfCtaFacade);
   protected loaderSpinnerCpntRef: void | Observable<
     ComponentRef<any> | undefined
   >;
+  protected _readyForScriptEvent: Subject<string> = new Subject();
+  readyForScriptEvent$: Observable<string> =
+    this._readyForScriptEvent.asObservable();
```
Platonn, line 49: [MINOR]
I have problem to find any usages of this observable. Can we remove it?

feature-libs/cart/base/occ/adapters/occ-cart-guest-user.adapter.ts
Matejk00: [OUTDATED] Done.

integration-libs/opf/cta/core/services/opf-dynamic-cta.service.ts
```diff
@@ -0,0 +1,251 @@
...
+    const sub = this.opfCtaFacade.listenScriptReadyEvent().subscribe({
+      next: (id: string) => {
+        this.isCartPage && this.dispatchCartEvents([id]);
+        !this.isCartPage && this.dispatchProductEvents([id]);
+      },
+    });
+    this.subList.push(sub);
+  }
+
+  protected dispatchCartEvents(scriptIdentifiers: Array<string>): void {
```
Platonn, line 179: [MEDIUM]
1. The method names are ambigious to me - whether it dispatches Cart events (e.g. Spartacus events about cart item added etc) OR native OPF/CTA events to communicate with global functions. I believe the latter.
What about adding `Cta` to the name of the method - `dispatchCartEvent()` -> `dispatchCtaCartEvent()` to make it clearer thsoe are native CTA events? If you have other ideas for renaming, I'm also open for it.

2. Why plural `events` but not `event`? I can see we're dispatching just one event.

Note: analogical remark applies also to the method `dispatchProductEvents()` -> `dispatchCtaProductEvents()`

integration-libs/opf/cta/core/services/opf-dynamic-cta.service.ts
```diff
@@ -0,0 +1,251 @@
...
+      this.scriptIdentifiers[this.scriptIdentifiers.length - 1]
+    ).padStart(4, '0');
+  }
+
+  registerScriptReadyEvent() {
+    this.globalFunctionsFacade.registerGlobalFunctions({
+      paymentSessionId: '',
+      domain: GlobalFunctionsDomain.GLOBAL,
+    });
+  }
```
Platonn, line 167: [MEDIUM]
1. Does the registration of the Global Function really have to be called before `CtaScriptsService.fetchCtaScripts()`? If not, I'm in favor of removing this public method, and move the responsibility of registering Global Function to inside of `initiateEvents()`  method - to be analogical to `stopEvents()` method, which already takes care of removing the global function. In other words: if not necessary, I'd like to not leak the responsibility of registering the global function to outside of this class
2. If we _really_ have to expose a public method to be called from outside (which I'm hoping maybe we don't have to), then I'd sugest at least renaming this method  (`registerScriptReadyEvent()` ->`registerGlobalFunction()`). Why: as far as I can understand, this method in its implementation doesn't not touch the scriptReady event - it neither subscribes to it, nor dispatches it. But it regsters a global function.

integration-libs/opf/cta/core/services/opf-dynamic-cta.service.ts
```diff
@@ -0,0 +1,251 @@
...
+  map,
+  merge,
+  Observable,
+  Subscription,
+  take,
+  tap,
+} from 'rxjs';
+
+@Injectable()
+export class OpfDynamicCtaService {
```
Platonn, line 46: [MINOR]
It's not clear at the first sight why we have 3 different services, so let's add JSDocs for them, explaining their purpose, relations and differences
- `CtaService`
- `CtaDynamicService`
- `CtaStaticService`

integration-libs/opf/cta/core/services/opf-dynamic-cta.service.ts
```diff
@@ -0,0 +1,251 @@
...
+  }
+
+  registerScriptReadyEvent() {
+    this.globalFunctionsFacade.registerGlobalFunctions({
+      paymentSessionId: '',
+      domain: GlobalFunctionsDomain.GLOBAL,
+    });
+  }
+
+  protected listenScriptReadyEvent() {
```
Platonn, line 169: [MEDIUM]
IMHO the method name is a bit misleading, especially while being the same as `opfCtaFacade.listenScriptReadyEvent()` , but those 2 methods doing 2 very different things - this one subscribes to a scriptReadt event and dispatches native CTA events; but the other method just returns an observable emitting the the scriptReady event.

1. I suggest to rename the other one - `opfCtaFacade.listenScriptReadyEvent()` to `opfCtaFacade.getScriptReadyEvent()` - because the method itself doens't setup any event _listener_, but just returns the observable, so you can _get_ the event observable.

2. I suggest to also align the naming of the local method `listenScriptReadyEvent()` with `cartChangedListener()` (and `productChangedListener()`). They do analogical things - they all subscribe to some Spartacus events and then emit native CTA events. What about renaming them to:
`addScriptReadyEventListener()`
`addCartChangedEventListener()`
`addProductChangedEventListener()`
or (respectively)
`listenToScriptReadyEvent()`
`listenToCartChangedEvent()`
`listenToProductChangedEvent()`

3. Then we could also align `initiateEvents()` and `stopEvents()` by renaming them to:
`addEventListeners()` and `removeEventListeners()`

4. Moreover, the property `this.isOnsiteMessagingInit` could be renamed to `this.eventListenersAdded`.

integration-libs/opf/cta/core/services/opf-dynamic-cta.service.ts
```diff
@@ -0,0 +1,251 @@
...
+  protected currencyService = inject(CurrencyService);
+  protected languageService = inject(LanguageService);
+  protected activeCartFacade = inject(ActiveCartFacade);
+  protected opfCtaFacade = inject(OpfCtaFacade);
+  protected currentProductService = inject(CurrentProductService);
+
+  protected subList: Array<Subscription> = [];
+  protected isOnsiteMessagingInit = false;
+  protected scriptIdentifiers: Array<string> = [];
+  protected isCartPage: boolean;
```
Platonn, line 59: [MAJOR]
Do we _really_ have to store this static state property? Now we set it in one methods, which causes side effect on other methods - which is hard to reason about (especially in the async flow of calling those methods).

btw. If we add more CTA flows in the future to this service, then a simple true-or-false variable `isCartPage` might be not sufficient and we might need to come up an enum to describe more possibilities (`cart/product/<some_other>/<some_different_other>`).

If possible, I'd suggest removing this static state completely from this service, and instead pass the `CtaScriptsLocation` as an argument of `initiateEvents()` and `stopEvents()`.

integration-libs/opf/quick-buy/core/facade/opf-quick-buy.service.ts
```diff
@@ -0,0 +1,90 @@
...
+  });
+
+  constructor(
+    protected commandService: CommandService,
+    protected opfQuickBuyConnector: OpfQuickBuyConnector,
+    protected cartAccessCodeFacade: CartAccessCodeFacade,
+    protected activeCartFacade: ActiveCartFacade,
+    protected userIdService: UserIdService,
+    protected logger: LoggerService
+  ) {}
```
Platonn, line 83: [MEDIUM]
Let's use `inject()` instead of constructor for DI. Use 3 commands:

```bash
npm add ngxtension
nx g ngxtension:convert-di-to-inject --project=opf
npm remove ngxtension

# Perhaps also need to revert some changes in those 2 files:
git checkout -- package.json package-lock.json
```

PS. Idelally we could run the official migration from Angular 18, instead of ngxtension - `npx @angular/cli@18 generate @angular/core:inject --path integration-libs/opf` - but I somehow couldn't get it work in our NX repo.

feature-libs/cart/base/occ/adapters/occ-cart-guest-user.adapter.ts
Platonn: Already answered offline my Mateusz: in the future other properties might be also possible to pass for the backend.

Platonn:
All comments moved to the main PR https://github.com/SAP/spartacus/pull/17250/
So now closing this temporary PR (that was created only for the purpose of the diff with the last code review checkpoint)

PR #19542, opened 2024-11-13
fix: support all patterns to with dynamic script

FollowTheFlo:
new logic:
- if destination.url exists:
      Iframe pattern and fullPage pattern use destination.url
      
- if no  destination.url and data exists, we display data within DOM, regardless the pattern.

### DIFF
--- a/integration-libs/opf/checkout/components/opf-checkout-payment-wrapper/opf-checkout-payment-wrapper.component.html
+++ b/integration-libs/opf/checkout/components/opf-checkout-payment-wrapper/opf-checkout-payment-wrapper.component.html
@@ -14,10 +14,13 @@
 <ng-template #paymentData let-paymentLink="paymentLink">
   <div>
     <div
-      *ngIf="paymentLink.renderType === RENDER_PATTERN.FULL_PAGE"
+      *ngIf="
+        paymentLink.renderType === RENDER_PATTERN.FULL_PAGE &&
+        paymentLink?.destination?.url
+      "
       class="cx-payment-link"
     >
-      <a class="btn btn-primary" [href]="paymentLink.data">{{
+      <a class="btn btn-primary" [href]="paymentLink?.destination?.url">{{
         'opfCheckout.proceedPayment' | cxTranslate
       }}</a>
     </div>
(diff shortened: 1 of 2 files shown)


integration-libs/opf/checkout/components/opf-checkout-payment-wrapper/opf-checkout-payment-wrapper.component.html
```diff
@@ -14,10 +14,13 @@
...
     <div
-      *ngIf="paymentLink.renderType === RENDER_PATTERN.FULL_PAGE"
+      *ngIf="
+        paymentLink.renderType === RENDER_PATTERN.FULL_PAGE &&
+        paymentLink?.destination?.url
+      "
       class="cx-payment-link"
     >
-      <a class="btn btn-primary" [href]="paymentLink.data">{{
+      <a class="btn btn-primary" [href]="paymentLink?.destination?.url">{{
```
FollowTheFlo: i am wondering if we should bypassSanitizer with this url? afaik we never got problem without it.

FollowTheFlo: note that when
paymentLink?.destination?.url exists, paymentLink.data has the same value as paymentLink.destination.url

integration-libs/opf/checkout/components/opf-checkout-payment-wrapper/opf-checkout-payment-wrapper.component.html
```diff
@@ -55,12 +58,7 @@
         />
       </form>
     </ng-container>
-    <ng-container
-      *ngIf="
-        paymentLink.data &&
-        paymentLink.renderType === RENDER_PATTERN.HOSTED_FIELDS
-      "
-    >
+    <ng-container *ngIf="paymentLink.data && !paymentLink?.destination?.url">
```
FollowTheFlo, line 61: all patterns inject HTML except the ones that have destination.url.
On workbench Iframe and FullPage both have the 2 options 'Create by script' & 'Create by Url'

Matejk00 (approved):

PR #19543, opened 2024-11-13
fix: Prevent blank order history page when using new tab component

sdrozdsap:


feature-libs/cart/base/components/add-to-cart/add-to-cart.component.ts
```diff
@@ -245,19 +245,22 @@ export class AddToCartComponent implements OnInit, OnDestroy {
...
-              | undefined
-          ) => {
-            this.disabled =
-              intendedPickupLocation?.pickupOption === 'pickup' &&
-              !intendedPickupLocation.displayName;
-          }
-        )
-      );
+      // Nesting this `if` statement for keeping feature flag check separately for easier removing
+      if (this.pickupOptionCompRef instanceof ComponentRef) {
```
sdrozdsap: This change is not directly related to the issue, but still its better to have this fix as soon as possible, since there is an error in console for b2b

projects/storefrontlib/cms-components/content/tab-paragraph-container/tab-paragraph-container.component.ts
```diff
@@ -144,7 +150,10 @@ export class TabParagraphContainerComponent implements AfterViewInit, OnInit {
     }
 
     // Render the tabs after the templates have completed loading in the view.
-    this.tabs$ = combineLatest([this.components$, this.tabRefs.changes]).pipe(
+    this.tabs$ = combineLatest([
+      this.components$,
+      this.tabRefs.changes.pipe(startWith(this.tabRefs)),
```
sdrozdsap, line 162: After navigating to different route and back the `tabRefs` are already initialised with values and `this.tabRefs.changes` does not emit any value. This change ensures that at least the initial value will be taken into account. This solves the problem

feature-libs/cart/base/components/add-to-cart/add-to-cart.component.ts
```diff
@@ -244,7 +244,10 @@ export class AddToCartComponent implements OnInit, OnDestroy {
   }
 
   onPickupOptionsCompLoaded() {
-    if (this.featureConfigService.isEnabled('a11yPickupOptionsTabs')) {
+    if (
+      this.featureConfigService.isEnabled('a11yPickupOptionsTabs') &&
+      this.pickupOptionCompRef instanceof ComponentRef
```
sdrozdsap, line 276: This change is not directly related to the issue, but still its better to have this fix as soon as possible, since there is an error in console for b2b

projects/storefrontlib/cms-components/content/tab-paragraph-container/tab-paragraph-container.component.ts
Zeyber: Nice find! Are we able to add an e2e test for the empty order history case just to make sure we have coverage?

Zeyber (changes requested):

projects/storefrontlib/cms-components/content/tab-paragraph-container/tab-paragraph-container.component.ts
sdrozdsap: Added test that checks if tabs are displayed after user navigated to different page and back

Zeyber (approved):

developpeurweb (approved):

PR #19545, opened 2024-11-14
fix: Billing address is not visible in order details

rmch91:
Closes: https://jira.tools.sap/browse/CXSPA-6643


integration-libs/opf/order/ng-package.json
```diff
@@ -0,0 +1,8 @@
+{
+  "$schema": "../../../node_modules/ng-packagr/ng-package.schema.json",
+  "dest": "../../dist/omf",
```
FollowTheFlo: shouldn't it be ../../dist/opf ?

rmch91: Right, changed

FollowTheFlo (approved):
Good to me.
I also performed QA in debug mode and via schematics both successful.

PR #19548, opened 2024-11-14
refactor(opf,cart,checkout): use Angular 'inject' function for DI instead of constructors

Platonn:
Use `inject` in:
- all classes in `opf` project
- new classes in `cart` and `checkout` project

### DIFF
--- a/integration-libs/opf/checkout/components/opf-checkout-payment-and-review/opf-checkout-payment-and-review.component.ts
+++ b/integration-libs/opf/checkout/components/opf-checkout-payment-and-review/opf-checkout-payment-and-review.component.ts
@@ -15,17 +15,9 @@ import {
   UntypedFormGroup,
   Validators,
 } from '@angular/forms';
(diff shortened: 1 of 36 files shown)


integration-libs/opf/checkout/components/opf-checkout-payment-and-review/opf-checkout-payment-and-review.component.ts
```diff
@@ -87,17 +71,6 @@ export class OpfCheckoutPaymentAndReviewComponent
...
-    super(
-      this.checkoutDeliveryAddressFacade,
-      this.checkoutPaymentFacade,
-      this.activeCartFacade,
-      this.translationService,
-      this.checkoutStepService,
-      this.checkoutDeliveryModesFacade
-    );
-  }
-
```
Platonn: We can reuse implicit constructor of the parent class, by skipping a constructor definition in the child class. So no need to re-declare all deps for the parent class. All `protected` properties are inherited anyway.

Matejk00 (approved):

PR #19553, opened 2024-11-15
fix: add sri property to js and css resources

FollowTheFlo:
if sri has a value from OPF server response: in DOM, it is now appearing as 'integrity' attribute:
https://developer.mozilla.org/en-US/docs/Web/Security/Practical_implementation_guides/SRI


rmch91 (approved):
LGTM

PR #19564, opened 2024-11-15
refactor(opf): address some review comments - part 12

Platonn:
Address review comment
 https://github.com/SAP/spartacus/pull/17250#discussion_r1640935216

**How I understood previous behavior:**
The `Promise` returned from `loadProviderResources()`:
- was `resolve`d when all resources were loaded.
- was `resolve`d immediately, when any loading error occurred (for any resource)

**New behavior**
The same as above.

**Concern: should we reject a promise in case of a loading error?**
In case of a loading error, IMHO it would be semantically more correct to `reject` (not `resolve`) the `Promise`.
But I refrained from changing the previous behavior, not to possibly break other consumer classes of the method `loadProviderResources()`.

Anyways, if I understood correctly, due to the previous behavior we had (and still have) a bug - error message not displayed to the user in the payment verification page, when a loading error occur for any resource. See the dependency on the `.catch()` block below:
https://github.com/SAP/spartacus/blob/6a9c76669325222c7178c374694738c4cac2b99a/integration-libs/opf/payment/root/components/opf-payment-verification/opf-payment-verification.service.ts#L242-L253
⬇️
https://github.com/SAP/spartacus/blob/6a9c76669325222c7178c374694738c4cac2b99a/integration-libs/opf/payment/root/components/opf-payment-verification/opf-payment-verification.component.ts#L56-L62
⬇️
https://github.com/SAP/spartacus/blob/6a9c76669325222c7178c374694738c4cac2b99a/integration-libs/opf/payment/root/components/opf-payment-verification/opf-payment-verification.component.ts#L91-L94

By the way, I've found also places which don't depend on the `.catch()` block and don't bother showing any error message in case of a resources loading error, e.g. see the google pay component below:
https://github.com/SAP/spartacus/blob/6a9c76669325222c7178c374694738c4cac2b99a/integration-libs/opf/quick-buy/components/opf-quick-buy-buttons/google-pay/google-pay.service.ts#L140-L144
⬇️
https://github.com/SAP/spartacus/blob/6a9c76669325222c7178c374694738c4cac2b99a/integration-libs/opf/quick-buy/components/opf-quick-buy-buttons/google-pay/google-pay.component.ts#L37-L48

### DIFF
--- a/integration-libs/opf/base/root/services/opf-resource-loader.service.spec.ts
+++ b/integration-libs/opf/base/root/services/opf-resource-loader.service.spec.ts
@@ -111,35 +111,21 @@ describe('OpfResourceLoaderService', () => {
 
       spyOn<any>(opfResourceLoaderService, 'loadScript').and.callThrough();
       spyOn<any>(opfResourceLoaderService, 'loadStyles').and.callThrough();
-      spyOn<any>(
-        opfResourceLoaderService,
-        'markResourceAsLoaded'
-      ).and.callThrough();
 
       opfResourceLoaderService.loadProviderResources([], [mockStyleResource]);
 
       expect(opfResourceLoaderService['loadScript']).not.toHaveBeenCalled();
       expect(opfResourceLoaderService['loadStyles']).not.toHaveBeenCalled();
-      expect(
-        opfResourceLoaderService['markResourceAsLoaded']
-      ).toHaveBeenCalled();
     }));
 
     it('should not load provider resources when no resources are provided', fakeAsync(() => {
       spyOn<any>(opfResourceLoaderService, 'loadScript').and.callThrough();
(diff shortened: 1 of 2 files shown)


FollowTheFlo:
About the verification page, we only want to display generic error message when a resource fails to load, I think current logic does it, I agree it would look better by using 'reject', so replacing resolve(false) by reject() would work.

About GooglePay, if it fails the GooglePay button will not be displayed so we want to keep it quiet for end-user. I think, It would look odd to see an error message without GP btn, we could add console error message eventually.

FollowTheFlo (approved):
I'm happy with the changes, I also ran some tests on CTA and checkout.
As mentioned, verification page could be improved with reject(), it can be done her or in another PR.

PR #19572, opened 2024-11-18
Add pagination to the payment list

Matejk00:
Closes: CXSPA-8484

### DIFF
--- a/integration-libs/opf/base/opf-api/config/default-opf-api-base-config.ts
+++ b/integration-libs/opf/base/opf-api/config/default-opf-api-base-config.ts
@@ -10,7 +10,7 @@ export const defaultOpfApiBaseConfig: OpfApiConfig = {
   backend: {
     opfApi: {
       endpoints: {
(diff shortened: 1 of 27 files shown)


integration-libs/opf/base/opf-api/config/default-opf-api-base-config.ts
```diff
@@ -10,7 +10,7 @@ export const defaultOpfApiBaseConfig: OpfApiConfig = {
   backend: {
     opfApi: {
       endpoints: {
-        getActiveConfigurations: 'active-configurations',
+        getActiveConfigurations: 'v2/active-configurations',
```
Jerify02, line 13: Should remove path fragment /v2, it was just used for your development and testing.
And the v2 api will be removed from our side once your code was deployed.

Matejk00: Thank you, it is removed.

rmch91 (approved):
LGTM

PR #19577, opened 2024-11-19
refactor(opf): rename `callbacksArray` to `callbacks`

Platonn:
In a recent PR we changed it from an array to an object, but forgot to change the property name. The leftover `Array` in the name is confusing.


FollowTheFlo (approved):
good to me as discussed

PR #19594, opened 2024-11-21
chore: Fix storefrontlib unit test

Zeyber:
Closes: https://jira.tools.sap/browse/CXSPA-8727


Zeyber:
Running unit tests 10x in a row with the following tests disabled succeeded without any flakes:
- `should NOT use router navigation for external links`
- `should NOT use router navigation for other protocols`

It can be concluded with reasonable certainty that these tests are the cultprits.

Zeyber:
Ran tests with fix 10x with no failures. See runs here: https://github.com/SAP/spartacus/actions/runs/12035224339/job/33583440720?pr=19594

kimhw0630 (approved):

PR #19598, opened 2024-11-21
fix: (CXSPA-8813) B2C button My Store not properly styled V2

StanislavSukhanov:
* closes: https://jira.tools.sap/browse/CXSPA-8813

### DIFF
--- a/feature-libs/pickup-in-store/components/container/my-preferred-store/my-preferred-store.component.ts
+++ b/feature-libs/pickup-in-store/components/container/my-preferred-store/my-preferred-store.component.ts
@@ -29,8 +29,8 @@ import { ICON_TYPE } from '@spartacus/storefront';
 import { Observable } from 'rxjs';
 import { filter, map, switchMap, take, tap } from 'rxjs/operators';
 
-const getDirectionsName = 'Get Directions';
-const changeStoreName = 'Change Store';
+const GET_DIRECTIONS_NAME = 'Get Directions';
+const CHANGE_STORE_NAME = 'Change Store';
 
 interface PreferredStoreContent {
   header: string;
@@ -50,8 +50,8 @@ export class MyPreferredStoreComponent implements OnInit {
   content: PreferredStoreContent = {
     header: 'My Store',
     actions: [
-      { event: 'send', name: getDirectionsName },
-      { event: 'edit', name: changeStoreName },
+      { event: 'send', name: GET_DIRECTIONS_NAME },
+      { event: 'edit', name: CHANGE_STORE_NAME },
     ],
   };
   openHoursOpen = false;
@@ -60,13 +60,14 @@ export class MyPreferredStoreComponent implements OnInit {
   isStoreFinder = false;
 
   private featureConfigService = inject(FeatureConfigService);
+  protected cdr = inject(ChangeDetectorRef, { optional: true });
+
   constructor(
     private preferredStoreFacade: PreferredStoreFacade,
     protected pickupLocationsSearchService: PickupLocationsSearchFacade,
     protected routingService: RoutingService,
     protected storeFinderService: StoreFinderFacade,
-    protected cmsService: CmsService,
-    protected cdr: ChangeDetectorRef
+    protected cmsService: CmsService
   ) {
     this.preferredStore$ = this.preferredStoreFacade.getPreferredStore$().pipe(
       filter((preferredStore) => preferredStore !== null),
@@ -116,7 +117,7 @@ export class MyPreferredStoreComponent implements OnInit {
               actions: [
                 {
                   link,
-                  name: getDirectionsName,
+                  name: GET_DIRECTIONS_NAME,
                   ariaLabel: 'cardActions.getDirections',
                   target: '_blank',
                 },
@@ -128,7 +129,7 @@ export class MyPreferredStoreComponent implements OnInit {
               actions: [
                 {
                   link,
-                  name: getDirectionsName,
+                  name: GET_DIRECTIONS_NAME,
                   ariaLabel: 'cardActions.getDirections',
                   target: '_blank',
                 },
@@ -136,7 +137,7 @@ export class MyPreferredStoreComponent implements OnInit {
               ],
             };
           }
-          this.cdr.detectChanges();
+          this.cdr?.detectChanges();
         });
     } else {
       this.cmsService
@@ -152,7 +153,7 @@ export class MyPreferredStoreComponent implements OnInit {
           tap(() => {
             this.content = {
               header: '',
-              actions: [{ event: 'send', name: getDirectionsName }],
+              actions: [{ event: 'send', name: GET_DIRECTIONS_NAME }],
             };
           })
         )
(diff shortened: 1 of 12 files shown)


feature-libs/pickup-in-store/components/container/my-preferred-store/my-preferred-store.component.ts
```diff
@@ -69,24 +84,78 @@ export class MyPreferredStoreComponent implements OnInit {
...
+                },
+              ],
+            };
+          }
+          this.content = content;
+        })
+      )
+      .subscribe();
+
+    source$
```
Pio-Bar: Do we need this second subscription? I think we can pull this off with just one

feature-libs/pickup-in-store/components/container/my-preferred-store/my-preferred-store.component.ts
```diff
@@ -69,24 +84,78 @@ export class MyPreferredStoreComponent implements OnInit {
     );
 
     useFeatureStyles('a11yViewHoursButtonIconContrast');
+    useFeatureStyles('a11yImproveButtonsInCardComponent');
   }
 
   ngOnInit(): void {
-    this.cmsService
-      .getCurrentPage()
+    const source$ = this.cmsService.getCurrentPage().pipe(
```
Pio-Bar: let's include it within a feature flag

Pio-Bar (changes requested):

feature-libs/pickup-in-store/components/container/my-preferred-store/my-preferred-store.component.ts
StanislavSukhanov: Re arranged a code. Please have a look

feature-libs/pickup-in-store/components/container/my-preferred-store/my-preferred-store.component.ts
StanislavSukhanov: fixed

Pio-Bar (dismissed):

feature-libs/pickup-in-store/components/container/my-preferred-store/my-preferred-store.component.ts
```diff
@@ -22,31 +29,44 @@ import { ICON_TYPE } from '@spartacus/storefront';
...
 
+  private featureConfigService = inject(FeatureConfigService);
   constructor(
     private preferredStoreFacade: PreferredStoreFacade,
     protected pickupLocationsSearchService: PickupLocationsSearchFacade,
     protected routingService: RoutingService,
     protected storeFinderService: StoreFinderFacade,
-    protected cmsService: CmsService
+    protected cmsService: CmsService,
+    protected cdr: ChangeDetectorRef
```
Zeyber: We need to optionally inject this as it is a breaking change

feature-libs/pickup-in-store/styles/_my-preferred-store.scss
```diff
@@ -77,6 +77,21 @@
         }
       }
     }
+
+    @include forFeature('a11yImproveButtonsInCardComponent') {
```
Zeyber, line 81: I suppose we don't need to change this file anymore?

projects/storefrontlib/shared/components/card/card.component.html
```diff
@@ -131,8 +131,8 @@
     </div>
     <!-- Actions -->
     <div *ngIf="content.actions && !editMode" class="cx-card-actions">
-      <div *ngFor="let action of content.actions">
-        <div>
```
Zeyber, line 135: It might just be the way the change is displayed. But have we removed a nested `div`?

feature-libs/pickup-in-store/components/container/my-preferred-store/my-preferred-store.component.ts
```diff
@@ -22,31 +29,44 @@ import { ICON_TYPE } from '@spartacus/storefront';
 import { Observable } from 'rxjs';
 import { filter, map, switchMap, take, tap } from 'rxjs/operators';
 
+const getDirectionsName = 'Get Directions';
```
Zeyber: Preferably these would be translatable. But lets leave that out of the scope of this fix.

Can we use `GET_DIRECTION_NAME` and `CHANGE_STORE_NAME` names for these variables so its clear that they should be final?

feature-libs/pickup-in-store/components/container/my-preferred-store/my-preferred-store.component.ts
```diff
@@ -69,27 +89,75 @@ export class MyPreferredStoreComponent implements OnInit {
...
+    if (
+      this.featureConfigService.isEnabled('a11yImproveButtonsInCardComponent')
+    ) {
+      this.cmsService
+        .getCurrentPage()
+        .pipe(
+          filter<Page>(Boolean),
+          take(1),
+          map((cmsPage) => {
+            this.isStoreFinder = cmsPage.pageId === 'storefinderPage';
```
Zeyber, line 106: Because we don't use the global `this.isStoreFinder` ourselves (unless it is used in the template?) and rather just the local `isStoreFinder`, lets just return the boolean instead of setting it globally too. I think the global variable should be removed if it doesn't really serve a purpose. Could we please also leave a `//TODO: Remove in major`-like comment if it isn't needed?

Zeyber (changes requested):

projects/storefrontlib/shared/components/card/card.component.html
StanislavSukhanov: Yes, Piotr has suggested to try to remove this div since it seemed it has not affected anything and the intention of this div wasn't clear. I've agreed and removed it. The appearance wasn't affected. Just got rid of redundant div.

feature-libs/pickup-in-store/components/container/my-preferred-store/my-preferred-store.component.ts
StanislavSukhanov: Completely agree with you, however it is used in a template in several places.

feature-libs/pickup-in-store/styles/_my-preferred-store.scss
StanislavSukhanov: It is needed to override the styles applied previously. 

from: 
Screenshot 2024-11-26 at 11 03 27
to:
Screenshot 2024-11-26 at 11 04 35

feature-libs/pickup-in-store/components/container/my-preferred-store/my-preferred-store.component.ts
StanislavSukhanov: fixed

projects/storefrontlib/shared/components/card/card.component.html
Zeyber: Ok, can we please put it back? This changes the DOM structure which is a breaking change.

StanislavSukhanov: @Zeyber done

Zeyber (approved):

developpeurweb (approved):

PR #19617, opened 2024-11-25
fix: (CXSPA-7992) Store, MyPreferredStoreComponent 'In Stock' icon contrast

StanislavSukhanov:
* fixed an opacity of 'In Stock' icon

closes: https://jira.tools.sap/browse/CXSPA-7992


projects/core/src/features-config/feature-toggles/config/feature-toggles.ts
```diff
@@ -644,10 +644,10 @@ export interface FeatureTogglesInterface {
   a11yScrollToReviewByShowReview?: boolean;
 
   /**
-   * `StoreComponent and MyPreferredStoreComponent` an icon in a button that triggers showing
-   * store's opening hours has an acceptable contrast ratio in a default theme
+   * `StoreComponent and MyPreferredStoreComponent` improve a contrast of icons
+   * in `View Hours` button and `In Stock` section
    */
-  a11yViewHoursButtonIconContrast?: boolean;
```
Pio-Bar, line 650: Since this feature flag has been already released we can't remove it/rename it. A client may have already enabled it and that would possibly affect his project. 

The name change makes perfect sense, but we may need to make a new feature flag anyway.

Pio-Bar (changes requested):

projects/core/src/features-config/feature-toggles/config/feature-toggles.ts
StanislavSukhanov: Fixed 👍

Pio-Bar (approved):

Zeyber (approved):

developpeurweb (approved):

PR #19618, opened 2024-11-25
feat: (CXSPA-8041) - Fix status as 'Selected' is not conveyed to the user in CheckoutDeliveryAddressComponent

petarmarkov9449:
Closes: [CXSPA-8041](https://jira.tools.sap/browse/CXSPA-8041)

Fix status as 'Selected' is not conveyed to the user in CheckoutDeliveryAddressComponent


Pio-Bar (dismissed):

projects/storefrontlib/shared/components/card/card.component.html
```diff
@@ -3,10 +3,13 @@
   class="cx-card"
   [tabindex]="border ? 0 : -1"
   [attr.role]="content.role ?? null"
-  [attr.aria-label]="
-    content.label ? (content.label | cxTranslate: { number: this.index }) : null
+  [attr.aria-labelledby]="
+    content.header ? 'content-header' + (index >= 0 ? '-' + index : '') : null
+  "
+  [attr.aria-describedby]="
```
Zeyber: Because we already have the `ariaDescribedBy()` getter, could we have this logic in there? It might also be good to have `ariaLabelledBy()` to do the same with. We should add comments and tests too as the logic is conditional.

Zeyber (changes requested):

projects/storefrontlib/shared/components/card/card.component.html
petarmarkov9449: I've updated it.

projects/storefrontlib/shared/components/card/card.component.html
```diff
@@ -25,11 +23,18 @@
       {{ content.deleteMsg }}
     </div>
     <!-- Card title -->
-    <span *ngIf="content.title" class="cx-card-title">
+    <span
+      *ngIf="content.title"
+      class="cx-card-title"
+      [id]="'content-title' + (index >= 0 ? '-' + index : '')"
```
Zeyber: Could we please use a get method for these `span`s too?

Zeyber (changes requested):

projects/storefrontlib/shared/components/card/card.component.html
petarmarkov9449: I've just updated it.

projects/storefrontlib/shared/components/card/card.component.html
```diff
@@ -3,9 +3,7 @@
   class="cx-card"
   [tabindex]="border ? 0 : -1"
   [attr.role]="content.role ?? null"
-  [attr.aria-label]="
-    content.label ? (content.label | cxTranslate: { number: this.index }) : null
-  "
+  [attr.aria-labelledby]="ariaLabelledBy"
```
Zeyber: Do we still need `ariaLabelledBy` if we have `getLabel`?

Zeyber (changes requested):

projects/storefrontlib/shared/components/card/card.component.html
petarmarkov9449: I've just updated it.

Zeyber (dismissed):

developpeurweb:
@petarmarkov9449 this should get narrated in a similar way as they do in My Account section, please reach out to @uroslates for feedback on how he fixed a similar bug in My Account.
The original bug for the ticket is now fixed, however a new bug was introduced, the narration is now repeated twice when focusing on the card and also while focusing on the button, as per screenshot below. 
While you are it, please exclude the button from the card narration (as in My Account for Address and Payment).

**In short:**
- When focusing the card, everything should be narrated only once, excluding the button.
- When focusing the button, only the button should be narrated.

payment method repetition

shipping address repetition

developpeurweb (approved):

PR #19623, opened 2024-11-26
feat: new translation languages

kimhw0630:
Add translations from TEW team and supporting Japanese, German, and Chinese except for two features: OPF and User Profile.

Technical details:
- in `translations.ts` barrel files now we export each langauge in a separate public API members (to allow for treeshaking unused langauges), e.g. `translationsEn`, `translationsDe`,... - in each library
- deprecate the the old wrapper consts e.g. `translations` (and moved those deprecated tokens to `public_api.ts` files from `translations.ts` files in each library)
- add a new function `extractTranslationChunksConfig()` in `@spartacus/core` and use it in every library for creating chunks config in their `translation.ts` files. It extracts the chunks config from the translations assets automatically. So no more need to type chunks config manually. Aa consequence a few human errors in the chunks configuration were fixed  implicitly by the new util `extractTranslationChunksConfig()` (you won't find it in the code, but only in runtime):
    - added a missing array at `cpqquoteTranslationChunksConfig.cpqQuote`
    - removed unnecessary array at `cpqquoteTranslationChunksConfig.cpqQuoteHeading`
    - removed values from the array `translationChunksConfig.myAccount` - the keys that have been moved to other files long time ago, but the leftover values in the chunks config remained (`"orderDetails",  "orderHistory",  "AccountOrderHistoryTabContainer",  "returnRequestList",  "returnRequest"`)
    - removed the obsolete value `"CMSTabParagraphContainer"` from the array at `translationChunksConfig.product` - such a key no longer exists in translations since some time, but a leftover remained in the chunks config
    - removed the obsolete value `"loginRegister"` from the array at `translationChunksConfig.user` - such a key no longer exists in translations since some time, but a leftover remained in the chunks config
- fix our installation schematics, to use the new recommended syntax in the scaffolded features e.g.:
   ```diff
   - resources: asmTranslations,
   + resources: { en: asmTranslationsEn }
   ```
- use the above new approach in all feature modules in the storefrontapp
- renamed files to match their chunk names: `recentSearches.json` -> `cdcRecentSearches.json`, `trendingSearches.json` -> `cdsTrendingSearches.json`

fixes https://jira.tools.sap/browse/CXSPA-7172


feature-libs/asm/assets/translations/ja/asm.json
```diff
@@ -0,0 +1,175 @@
...
+    "logout": "サインアウト",
+    "hideUi": "Close ASM",
+    "helpPortalUrl": "ヘルプ",
+    "customers": "顧客",
+    "asmCustomer360Button": "Customer 360 View",
+    "asmCustomer360ViewButton": "Customer 360°",
+    "createCustomerSuccessfullyAlert": "The customer account has been created and the customer session has started.",
+    "saveInactiveCartAlertInfo": "The identified cart is an inactive cart. To take further actions on this cart, save it first.",
+    "activeCartAlertInfo": "The identified cart is an active cart.",
+    "startCustomerEmulationAlertInfo": "Customer emulation has started. Any actions you do will reflect the effects on the customer account.",
```
pawelfras: `major:` There are many values across all new translation files that are not translated to the particular language. Do you perhaps know if there is any official process to inform TEW about these gotchas?

pawelfras (commented):
Unfortunately, translation files are not tree-shakeable and they be a part of the main bundle. Right now, main.js of storefrontapp's prod build has grown by 300kb, and it will grow each time we add new translation files/languages. I'm checking if we can load these files dynamically to reduce the initial load. Probably something I'd like to discuss with @Platonn.

Platonn (approved):

PR #19635, opened 2024-11-26
fix: display form for full page pattern

FollowTheFlo:
Conditions top to bottom:
1- FullPage pattern with url and no form -> display url in anchor tag
2- Iframe pattern with url and no form -> display url in iframe tag
3- Any pattern with a form and url -> display form with url in submit btn
4- Any pattern with data and no url -> display data in innerHTML

I believe it covers all configs, 2 options cannot be displayed at same time.


rmch91 (approved):
LGTM

PR #19651, opened 2024-11-28
fix: Use proper keybaord navigation keys for facets

sdrozdsap:
Closes: https://jira.tools.sap/browse/CXSPA-9020

### DIFF
--- a/projects/storefrontlib/layout/a11y/keyboard-focus/keyboard-focus.utils.ts
+++ b/projects/storefrontlib/layout/a11y/keyboard-focus/keyboard-focus.utils.ts
@@ -1,0 +1,26 @@
+/**
+ * Temporarily removes elements from the tabbing flow and restores them after a tick.
+ *
+ * This method sets the `tabIndex` of each element in the provided iterable to `-1`
+ * and resets it back to `0` using `requestAnimationFrame`. While using `requestAnimationFrame`
+ * may seem like a bad code smell, it is justified here as it ensures a natural tabbing flow
+ * in cases where determining the next focusable element is complex, such as when directives
+ * like `TrapFocusDirective` modify the DOM's focus behavior.
+ *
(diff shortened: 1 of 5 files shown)


developpeurweb:
@sdrozdsap I have tested this with keyboard-only navigation, and also with JAWS. Please proceed to merge after code review, no need to retest.

projects/storefrontlib/layout/a11y/keyboard-focus/keyboard-focus.utils.ts
```diff
@@ -0,0 +1,32 @@
...
+ * like `TrapFocusDirective` modify the DOM's focus behavior.
+ *
+ * This utility is especially useful for scenarios like menus, lists, or carousels where
+ * `Tab` navigation is intentionally disabled, but other keyboard keys (e.g., `Arrow` keys)
+ * are used for navigation. It helps prevent these elements from disrupting the tab order
+ * while allowing other key-based interactions.
+ *
+ * @param elements - An iterable of `HTMLElement` objects to temporarily remove from tab navigation.
+ */
+export const temporarilyDisableTabbing = (elements: Iterable<HTMLElement>) => {
```
Zeyber: Thanks for explaining in depth what this is for in the comments :)

I only think to change the name of this to distinguish the purpose of the function. Something like `disableTabbingWhile...` because we don't really know enough context with the current name.

Zeyber (changes requested):

projects/storefrontlib/layout/a11y/keyboard-focus/keyboard-focus.utils.ts
sdrozdsap: Changed to `disableTabbingForTick`

Zeyber (approved):

PR #19653, opened 2024-11-28
fix: (CXSPA-8991) set aria-label attribute for navigation component button

uroslates:
Closes: https://jira.tools.sap/browse/CXSPA-8991


Pio-Bar (changes requested):
It feels like we are going in circles with adding and then removing attributes for this component. It needs to be taken into account that this component is used in a couple different places (header navi, my account dropdown, footer).

This [PR](https://github.com/SAP/spartacus/pull/19359/) removed this exact solution (for the header navigation) while working on the 'my account' dropdown button. Disregarding the other use of the component. We should not repeat this mistake.

Let's figure out a solution that meets the a11y requirements for all instances of this component.

uroslates:
> It feels like we are going in circles with adding and then removing attributes for this component. It needs to be taken into account that this component is used in a couple different places (header navi, my account dropdown, footer).
> 
> This [PR](https://github.com/SAP/spartacus/pull/19359/) removed this exact solution (for the header navigation) while working on the 'my account' dropdown button. Disregarding the other use of the component. We should not repeat this mistake.
> 
> Let's figure out a solution that meets the a11y requirements for all instances of this component.

@Pio-Bar it should be addressed now. Please feel free to check if its OK.

projects/storefrontlib/cms-components/navigation/navigation/navigation-ui.component.ts
```diff
@@ -396,4 +405,12 @@ export class NavigationUIComponent implements OnInit, OnDestroy {
     }
     return depth > 0 && !node?.children ? -1 : 0;
   }
+
+  ariaLabel(node: NavigationNode) {
+    return !this.isHeadersLinksSlotNavigation ? node.title : undefined;
```
Pio-Bar: Looks good, maybe we should include a test to ensure this behaviour works as expected? This could help avoid breaking this feature again and additionally document it for others

Pio-Bar (changes requested):

projects/storefrontlib/cms-components/navigation/navigation/navigation-ui.component.ts
uroslates: Added

Pio-Bar (dismissed):

Zeyber (changes requested):
Thanks for your catch of this issue and workaround!

I think that the current solution we have is much too specific in that we need to react specifically on a "HeaderLinks" node. I believe we might be able to simplify this approach.

What if we had a configuration instead identified by `node.title` for example and added attributes according to that? For example:
```
navigation-node.config.ts

export const NavigationNodeConfig = {
  HeaderLinks: {
    ariaDescribedBy: 'greeting'
  },
  OtherNode: {
    someAttribute: 'some value'
  }
}
```
```
navigation-ui.component.ts

NavigationUiComponent {
  ...
  ariaLabel(node) {
    return this.config?.[node.title]?.ariaLabel ?? node.title
  }

  ariaDescribedBy(node) {
    return this.config?.[node.title]?.ariaDescribedBy ?? undefined  
  }
}
```
Of course, this is just a rough idea and you might have a way to optimize this further. Maybe there is a better identifiable key than `node.title`.

uroslates:
> Thanks for your catch of this issue and workaround!
> 
> I think that the current solution we have is much too specific in that we need to react specifically on a "HeaderLinks" node. I believe we might be able to simplify this approach.
> 
> What if we had a configuration instead identified by `node.title` for example and added attributes according to that? For example:
> 
> ```
> navigation-node.config.ts
> 
> export const NavigationNodeConfig = {
>   HeaderLinks: {
>     ariaDescribedBy: 'greeting'
>   },
>   OtherNode: {
>     someAttribute: 'some value'
>   }
> }
> ```
> 
> ```
> navigation-ui.component.ts
> 
> NavigationUiComponent {
>   ...
>   ariaLabel(node) {
>     return this.config?.[node.title]?.ariaLabel ?? node.title
>   }
> 
>   ariaDescribedBy(node) {
>     return this.config?.[node.title]?.ariaDescribedBy ?? undefined  
>   }
> }
> ```
> 
> Of course, this is just a rough idea and you might have a way to optimize this further. Maybe there is a better identifiable key than `node.title`.

The problem with this issue's requirement is that we either want the aria-label (reading node.title) or aria-describedby ('deviation' case pointing to targeted HTML element whose text we want to read) to be applied for the navigation root node button and in following way:
1. **aria-label** for preferably all cases except for the case 2 (case below)
2. **aria-describedby** which will be set to read the LoggedIn user's name instead of the node's title! The logged in user's name info is encapsulated in the `div#greeting` (which is how it was implemented before - and this is what this hardcoded value is related to). The way to identify where is this going to be applicable is following - this attribute (case) is only applicable for the root navigation node of `uid: MyAccountComponent` (which is a a CMS configured component for the slotId: "HeaderLinksSlot" slot - rendered at the position `position: "HeaderLinks"`). In our codebase [LoginComponent](https://github.com/SAP/spartacus/blob/a4b72abd137d8898669c4d8309965d5b2c7bb9b5/feature-libs/user/account/components/login/login.component.html#L2) is the component that introduces this tight-coupling to our requirement's since its defining both container for page-slot component with position="HeaderLinks" and id="greeting" (both defined within its HTML).

So this PR's implementation, when rendering NavigationUIComponent, is checking if the navigation item: 
1. belongs to the page slot with position="HeaderLinks" (by traversing and searching for the closest parent based on tag cx-page-slot)- in which case we know (since now we know we're in the context of My Profile navigation) its using aria-describedby attribute set to the value of greeting or
2. for all other normal cases just uses aria-label attribute set to node.title value (in Ideal/Utopian world this would be used everywhere).

The only **important improvement** I would add here IMHO is that we should define both of these tight-coupling variables:
1. position: "HeaderLinks" and 
2. describedby element value "greeting"
In some global constants file (if it exists) and use reference them as constants in both files below:
- navigation-ui.component.ts logic as well as in
- login.component.html

Update: The reason why I would avoid using `return this.config?.[node.title]?.ariaLabel ?? node.title` approach is due to i18n support. This is why i would rather use more invariant cms property like position (plus if searched its value is going to be found as used on 2 places - navigation.ui.component.ts and login.component.html).

cc @Zeyber @Pio-Bar

developpeurweb:
@uroslates please also add tooltips to carets. E.g. title="Brands Menu"

feature-libs/user/account/components/login/login.component.ts
```diff
@@ -35,4 +41,22 @@ export class LoginComponent implements OnInit {
       })
     );
   }
+
+  get greeting$() {
```
Zeyber: I'm wary of using getter methods to communicate with the template because AFAIK, they will be called repeatedly to get the latest data. I think that we use it here so that we are always getting the up-to-date DOM and can react to changes. But I think this also has a large performance impact which will be sensitive to all applications (all users will likely have the login button in their storefronts).

Is it possible to use something like a `MutationObserver` or some other way to check for DOM updates that we can have in our `greeting$` pipe?

Zeyber (changes requested):
I really like this solution and its nearly there! I think we could also have an e2e test to check for the label being applied in reality since its sensitive to DOM structure.

feature-libs/user/account/components/login/login.component.ts
uroslates: Hi @Zeyber, I have implemented the solution using `MutationObserver` concept in last commit. Please feel free to dbl check it and let me know if U think its better or not. Thanks!

feature-libs/user/account/components/login/dom-change.directive.ts
```diff
@@ -0,0 +1,40 @@
...
+export class DomChangeDirective implements OnDestroy {
+  private changes: MutationObserver;
+
+  @Output()
+  public cxButtonAdded = new EventEmitter();
+
+  constructor(private elementRef: ElementRef) {
+    this.changes = new MutationObserver((mutations: MutationRecord[]) => {
+      mutations
+        .filter((mutation) => mutation.target?.nodeName === 'BUTTON')
```
Zeyber: Maybe we can pass in the filter for mutations as a directive parameter. That way we can reuse this directive for whatever DOM changes we need to check for.

Zeyber (changes requested):

feature-libs/user/account/components/login/dom-change.directive.ts
Zeyber: Better yet, can we pass in the target to get the changed mutation directly? Something like this:
```
[cxOnDomChange]="
  'cx-navigation-ui nav ul li:first-child button', // target
  callbackFn(target) // returns target element so we can modify it without using querySelector
"
```

uroslates: Hi @Zeyber I have modified the implementation to be more generic (as U suggested). Please let me know if U find it OK now?

feature-libs/user/account/components/login/dom-change.directive.ts
```diff
@@ -0,0 +1,64 @@
...
+ * Optional _cxDomChangeTargetSelector_ Input enables clients to be more specifixc when reacting to
+ * the Host's element DOM changes by allowing them to specify css selector (relative to Host Element)
+ * that narrows down the number of _cxDomChange_ Output executions to be triggered only when
+ * _cxDomChangeTargetSelector_ Input relevant elements changes are detected.
+ */
+@Directive({
+  selector: '[cxDomChange]',
+})
+export class DomChangeDirective implements OnDestroy {
+  private changes: MutationObserver;
```
Zeyber: `protected`

feature-libs/user/account/components/login/dom-change.directive.ts
```diff
@@ -0,0 +1,64 @@
...
+
+  /**
+   * Optional Css Selector Input filtering DOM mutations to those targeting only elements described via provided selector
+   */
+  @Input() cxDomChangeTargetSelector?: string;
+
+  @Output()
+  public cxDomChange = new EventEmitter<MutationRecord>();
+
+  constructor(private elementRef: ElementRef) {
```
Zeyber: Lets use `inject(ElementRef)` instead of passing it through the constructor.

Zeyber (changes requested):
Lets also include an e2e to stress test the login greeting (eg. log in and check, log out and check, log in and check) to make sure nothing funny happens over a few state changes. Because we are introducing a new directive that has to deal with some complexity, we should make sure we haven't overlooked something.

feature-libs/user/account/components/login/dom-change.directive.ts
uroslates: Isn't it better for testing purpose to pass the elementRef? I only saw inject pattern on 2 places in project.

uroslates: Everything is added except e2e test.

uroslates:
@Pio-Bar could you please dbl check the added e2e test so that we could merge this one to develop (in caee all good)?

projects/storefrontapp-e2e-cypress/cypress/e2e/regression/user_access/navigation-login-a11y.e2e.cy.ts
```diff
@@ -0,0 +1,49 @@
...
+    it('should login and logout successfully and have correct Navigation Menu buttons values', () => {
+      login.loginUser();
+
+      const tokenRevocationRequestAlias =
+        login.listenForTokenRevocationRequest();
+
+      cy.get(
+        'cx-login > cx-page-slot > cx-navigation > cx-navigation-ui > nav > ul > li > button'
+      )
+        .contains('My Account')
```
Pio-Bar: Maybe we can shorten these selectors and use an aliases to avoid repetitions?
e.g.:

```suggestion
      cy.get('cx-login button').contains('My Account').as('MyAccountBtn')
     // and later cy.get('@MyAccountBtn')
```

Pio-Bar (changes requested):

projects/storefrontapp-e2e-cypress/cypress/e2e/regression/user_access/navigation-login-a11y.e2e.cy.ts
uroslates: @Pio-Bar  Addressed it and improved the test further.

projects/storefrontapp-e2e-cypress/cypress/e2e/regression/user_access/navigation-login-a11y.e2e.cy.ts
```diff
@@ -0,0 +1,51 @@
...
+
+function assertNavigationButtonsAttributes(buttonsSelector: string) {
+  cy.get(buttonsSelector).each(($btn) => {
+    const btnAriaLabel = $btn.attr('aria-label');
+    cy.wrap($btn)
+      .should('have.attr', 'title', `${btnAriaLabel} Menu`)
+      .should('have.attr', 'aria-label', btnAriaLabel)
+      .should('have.attr', 'aria-controls', btnAriaLabel);
+  });
+}
```
Pio-Bar: Not sure I understand this part: We are grabbing the value of the button's aria label and then checking if it has that value?

projects/storefrontapp-e2e-cypress/cypress/e2e/regression/user_access/navigation-login-a11y.e2e.cy.ts
```diff
@@ -0,0 +1,51 @@
...
+
+      const tokenRevocationRequestAlias =
+        login.listenForTokenRevocationRequest();
+
+      const profileMenuRootBtnSelector =
+        'cx-login > cx-page-slot nav ul li:first-child button';
+      cy.get(profileMenuRootBtnSelector)
+        .contains('My Account')
+        .invoke('attr', 'ariaLabel')
+        .contains(`Hi, ${user.firstName} ${user.lastName}`);
```
Pio-Bar: We can simplify this with shorter selectors and aliases. We should try and follow the Cypress way. 
```suggestion
      cy.get('cx-login button')
        .as('myAccountBtn')
        .invoke('attr', 'ariaLabel')
        .contains(`Hi, ${user.firstName} ${user.lastName}`);
        // ....
        cy.get('@myAccountBtn').should('not.exist');
```

Pio-Bar (changes requested):

projects/storefrontapp-e2e-cypress/cypress/e2e/regression/user_access/navigation-login-a11y.e2e.cy.ts
uroslates: Basically all targeted buttons should have same values for aria-label & aria-control. Same for title except that it has ' Menu' suffix added to it. So this is being asserted here.

projects/storefrontapp-e2e-cypress/cypress/e2e/regression/user_access/navigation-login-a11y.e2e.cy.ts
uroslates: OK thanks, i have applied the changes.

uroslates: @Pio-Bar it should be now addressed. Could you please dbl check?

Pio-Bar (dismissed):

uroslates:
@Zeyber it seems that this can not be merged before your approval. So I'll do it after I come back.

Zeyber (approved):

PR #19660, opened 2024-11-28
fix: disable attribute key restriction for opf

FollowTheFlo:
Note:
A code ahange has been made on script-loader.service.ts which belong to core, I tooke care about not introducing breaking change.


projects/core/src/util/script-loader.service.ts
```diff
@@ -39,6 +40,7 @@ export class ScriptLoader {
     callback?: EventListener;
     errorCallback?: EventListener;
     placement?: ScriptPlacement;
+    disableKeyRestriction?: boolean;
```
FollowTheFlo, line 43: Not a breaking change as adding optional property 'disableKeyRestriction'.
There is no impact on existing behavior when this property is not present (undefined).

Matejk00 (approved):

PR #19669, opened 2024-11-29
fix: add active language code to requests

Matejk00:
Closes: CXSPA-8797


FollowTheFlo (approved):
looks good, also tested successfully with payment, cta and GP.

PR #19672, opened 2024-11-29
fix: Remove unnecessary visible text for screen-readers

steinsebastian:
CXSPA-9063: Screenreader-specific text became visible. This does not make sense in this case.


Larisa-Staroverova (approved):
Thank you for reverting unneccessary changes, @steinsebastian! 
Retested. It worked as expected!

PR #19678, opened 2024-12-02
fix: (CXSPA-9021) Button Menu Focus not looped within the menu items.

StanislavSukhanov:
* closes: https://jira.tools.sap/browse/CXSPA-9021


projects/storefrontlib/layout/main/storefront.component.ts
```diff
@@ -121,6 +134,16 @@ export class StorefrontComponent implements OnInit, OnDestroy {
     this.hamburgerMenuService.toggle(true);
   }
 
+  excludeFromFocus(isExpanded: boolean): void {
+    const tabindex = isExpanded ? '-1' : '0';
+    const logo = this.elementRef.nativeElement.querySelector('cx-banner a');
```
Zeyber: I think your approach of applying `tabindex` to elements is acceptable as I can not think of how we can apply the behaviour correctly otherwise given that the layout is delivered to us from the backend a specific way.

We should keep in mind that user's may have (and likely do) different page slot and layout configurations than us. We need to keep this under consideration with our solution.

In an ideal world, we would have the navigation and the close button in the same slot so that we don't need to modify elements outside of it. We could also use the `dialog` element which mostly has all of the a11y keyboard requirements (ie. tab trap) we need built in.

With all that said, we should consider what we do have and what might solution might cover all edge cases. I see that in particular, section `navigation` needs to all be focusable as well as the `close` button outside. And that's it.
Is it possible to select all elements that could be focusable outside of this criteria (`button` or `a` elements that are visible)? If we can block those indiscriminately with `tabindex`, we don't need to worry as much if they have their page structured as much as we expect.

Zeyber (changes requested):

projects/storefrontlib/layout/main/storefront.component.ts
StanislavSukhanov: Hi @Zeyber. Fixed. Thanks for thinking wider 👍

projects/storefrontlib/layout/a11y/keyboard-focus/skip-focus.directive.ts
```diff
@@ -0,0 +1,67 @@
...
+
+export interface SkipFocusConfig {
+  isEnabled: boolean;
+  skipSelectors?: string[];
+}
+
+@Directive({
+  selector: '[cxSkipFocus]',
+})
+export class SkipFocusDirective implements OnChanges {
```
Zeyber, line 33: I'm trying to think of something to encaptulate what this directive does, while not being confused with the other `TrapFocusDirective` which takes a different approach. I looked at that directive for a while seeing if there is some simpler solution we can leverage. But it does not look built to easily be adjusted to handle our need. So this new directive is justified IMO.

As for naming, I am thinking something similar that highlights our specific implementation (disabling tabindex on all other elements outside of trap). `TrapFocusByTabIndexDirective`? Its long but clear... Maybe there is something just as clear but shorter?

projects/storefrontlib/layout/a11y/keyboard-focus/skip-focus.directive.ts
```diff
@@ -0,0 +1,67 @@
...
+  @Input('cxSkipFocus') config: SkipFocusConfig = { isEnabled: false };
+
+  constructor(
```
Zeyber: Lets inject instead of using constructor

projects/storefrontlib/layout/a11y/keyboard-focus/skip-focus.directive.ts
```diff
@@ -0,0 +1,67 @@
...
+    protected elementRef: ElementRef,
+    protected winRef: WindowRef
+  ) {}
+
+  public ngOnChanges(_changes: SimpleChanges): void {
```
Zeyber: I think using `changes` could be a heavy-handed way to handle this issue. It is going to be present on every storefront component which contains the entire application. Meaning it will be a constant performance impact on every page. I've left a comment below about using an observable make changes to the directive. Maybe this would let us not need to watch `onChanges`?

projects/storefrontlib/layout/main/storefront.component.html
```diff
@@ -3,12 +3,22 @@
...
-      [cxFocus]="{ disableMouseFocus: true }"
+      [cxFocus]="focusConfig"
       [class.is-expanded]="isExpanded$ | async"
       (keydown.escape)="collapseMenu()"
       (click)="collapseMenuIfClickOutside($event)"
     >
-      <cx-page-layout section="header"></cx-page-layout>
+      <ng-container *cxFeature="'a11yHamburgerMenuTrapFocus'">
+        <cx-page-layout
+          [cxSkipFocus]="skipFocusConfig"
```
Zeyber: Because the `skipFocusConfig` will be changing frequently, why don't we use `skipFocusConfig$ | async` instead? Then we could use an `Observable` object instead in our class instead of making another subscription.

Zeyber: On this note and about responsiveness to changes as above and readability, maybe the `isEnabled` should be its own parameter. There could be arguments against this though. What do you think?

Zeyber (changes requested):

projects/storefrontlib/layout/main/storefront.component.html
StanislavSukhanov: I've added `distinctUntilChanged()` to `isExpanded$` observable in storefront component, so the change is'nt triggered when it emits. Tested it. We are safe.

projects/storefrontlib/layout/a11y/keyboard-focus/skip-focus.directive.ts
StanislavSukhanov: I would leave the name as it is. We have a directive named `CxFocus`, so this one does the opposite, hence the name `SkipFocus`. I have added comments what the directive does. But I do agree that naming is a bit challenging in this case

projects/storefrontlib/layout/a11y/keyboard-focus/skip-focus.directive.ts
StanislavSukhanov: fixed

projects/storefrontlib/layout/main/storefront.component.ts
```diff
@@ -105,6 +115,18 @@ export class StorefrontComponent implements OnInit, OnDestroy {
         })
       );
     }
+
+    if (this.featureConfigService.isEnabled('a11yHamburgerMenuTrapFocus')) {
+      this.isExpanded$
```
Zeyber: Lets put this in a protected function to reduce the length `ngOnInit` and so that we can name our function something that describes why we are subscribing.

Zeyber (changes requested):
Thanks for your efforts on this! Just missing tests. Do you think this would warrant an e2e test too?

projects/storefrontlib/layout/a11y/keyboard-focus/skip-focus.directive.ts
StanislavSukhanov: fixed, please see a comment https://github.com/SAP/spartacus/pull/19678#discussion_r1871228321

projects/storefrontlib/layout/main/storefront.component.ts
StanislavSukhanov: done

Zeyber (approved):

developpeurweb (approved):

PR #19680, opened 2024-12-02
fix: (CXSPA-9025) - Add style for active and focused states in SearchBoxComponent

petarmarkov9449:
Closes: [CXSPA-9025](https://jira.tools.sap/browse/CXSPA-9025)

Add style for active and focused states in SearchBoxComponent


projects/storefrontapp/src/app/spartacus/spartacus-features.module.ts
```diff
@@ -405,8 +405,9 @@ if (environment.cpq) {
         a11yTableHeaderReadout: true,
         a11ySearchboxAssistiveMessage: true,
         a11yDifferentiateFocusedAndSelected: true,
+        a11yActiveAndFocusedSearchBoxComponent: true,
         cmsBottomHeaderSlotUsingFlexStyles: true,
-        useSiteThemeService: false,
+        useSiteThemeService: true,
```
Pio-Bar: probably included by accident 😉

projects/storefrontstyles/scss/components/product/search/_searchbox.scss
```diff
@@ -296,6 +296,17 @@
         z-index: 20;
         padding-top: 25px;
       }
+
+      @include forFeature('a11yActiveAndFocusedSearchBoxComponent') {
+        &:focus-within,
+        &:active {
```
Pio-Bar: The outline should be visible when the keyboard is used for navigation only. Wouldn't this selector mean it is visible even when mouse is used? Maybe we should mimic how other inputs in the app work?

projects/storefrontstyles/scss/components/product/search/_searchbox.scss
```diff
@@ -296,6 +296,17 @@
...
+          outline-style: solid;
+          outline-color: var(--cx-color-visual-focus);
+          outline-width: var(--cx-visual-focus-width, 2px);
+          outline-offset: 4px;
+          transition: none;
```
Pio-Bar: We should already have a mixin for the focus outline, can it be used here?
```suggestion
 @include visible-focus();
```

Pio-Bar (changes requested):

projects/storefrontstyles/scss/components/product/search/_searchbox.scss
petarmarkov9449: I've just updated it.

projects/storefrontstyles/scss/components/product/search/_searchbox.scss
petarmarkov9449: I've just updated it.

projects/storefrontapp/src/app/spartacus/spartacus-features.module.ts
petarmarkov9449: I've just fixed it.

projects/storefrontstyles/scss/components/product/search/_searchbox.scss
```diff
@@ -339,6 +353,16 @@
         color: var(--cx-color-secondary);
       }
 
+      @include forFeature('a11yActiveAndFocusedSearchBoxComponent') {
+        .reset,
+        div.search-icon {
+          position: absolute;
```
Pio-Bar: Do we need positioning changes to add the outline? I am having a hard time figuring what is the intention behind it. 

Maybe using sth similar to `.cx-label-inner-container:focus-within` can work?

Pio-Bar (changes requested):

projects/storefrontstyles/scss/components/product/search/_searchbox.scss
petarmarkov9449: In the other component like Search of `QuickOrderComponent`, they didn't use `focus-within` or `focus`.
Instead they made the input to fulfill the parent `div` while adding more `padding` (especially right). And positioned the `button` like `reset` and `search div` in that right padding using `absolute`.

petarmarkov9449: I've just updated it using keyboard-focus module.

projects/storefrontlib/cms-components/navigation/search-box/search-box.component.html
```diff
@@ -127,6 +128,70 @@
...
   </label>
+  <label
+    *cxFeature="'a11yActiveAndFocusedSearchBoxComponent'"
+    class="searchbox cxFeat_a11ySearchboxLabel"
+    [class.dirty]="!!searchInput?.value"
+  >
+    <span class="cx-input-label">{{ 'common.search' | cxTranslate }}</span>
+    <div
+      class="cx-label-inner-container"
+      [cxFocus]="{ disableMouseFocus: true }"
```
Pio-Bar: I think it we shouldn't need to use the VisibleFocusDirective here since it's already applied to the header.

Pio-Bar (changes requested):

projects/storefrontlib/cms-components/navigation/search-box/search-box.component.html
petarmarkov9449: Updated it using` focus-within` pseudo instead.

Pio-Bar (dismissed):

Zeyber (dismissed):

developpeurweb:
@petarmarkov9449 desktop and tablet looks good, can you please optimize for mobile as well:
-  Focus-ring with all 4 sides visible
- Adjust top side so the focus-ring doesn't include the "Search" label, but only the field.

9025 needs mobile fix

developpeurweb (approved):

PR #19683, opened 2024-12-02
fix: (CXSPA-9060) - Limit popover width on sub 375px devices

Pio-Bar:
Closes: [CXSPA-9060](https://jira.tools.sap/browse/CXSPA-9060)


developpeurweb (approved):
@Pio-Bar let's deploy it to retest on real mobile device

Zeyber (approved):

PR #19684, opened 2024-12-02
fix: dynamicScript takes precedence over url on payment option rending

FollowTheFlo:
- renamed property 'data' to 'html' as it is only used for html snippet.
- presence of dynamicScript is now checked before destination url.
- in 'renderPaymentMethodEvent$.next()', emitted object never contains 'destination' and 'html' together, thus no need to do additional check it in template.


Matejk00 (approved):

giancorderoortiz:
No longer needed after consultation with @FollowTheFlo

PR #19685, opened 2024-12-02
fix: add default card params in object

FollowTheFlo:
protected object have type 'any', because we cannot expose any type from AP/GP in our public api (see [CXSPA-5634](https://jira.tools.sap/browse/CXSPA-5634))


Matejk00 (dismissed):

Matejk00 (approved):

Matejk00:
Closing as there are no changes against develop

PR #19691, opened 2024-12-03
fix: (CXSPA-9035) StoreFinder Truncation on applying text spacing settings

StanislavSukhanov:
* closes: https://jira.tools.sap/browse/CXSPA-9035

### DIFF
--- a/feature-libs/storefinder/styles/components/_store-finder-list-item.scss
+++ b/feature-libs/storefinder/styles/components/_store-finder-list-item.scss
@@ -36,6 +36,9 @@
     @include media-breakpoint-down(md) {
       white-space: normal;
     }
+    @include forFeature('a11yTruncatedTextStoreFinder') {
+      white-space: normal;
+    }
   }
   .cx-store-open {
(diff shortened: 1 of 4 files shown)


feature-libs/storefinder/styles/components/_store-finder-list-item.scss
```diff
@@ -36,6 +36,9 @@
     @include media-breakpoint-down(md) {
       white-space: normal;
     }
+    @include forFeature('a11yTruncatedTextStoreFinder') {
+      white-space: normal;
+    }
```
Pio-Bar: Looks good! Small improvement we could make is to put this change below the property it overwrites and use `white-space: unset`. Since ideally we would remove the `white-space: nowrap` and use the default value. This should make it clearer to the dev that will remove this FF later that this is what we intended .

Pio-Bar (dismissed):

Zeyber (dismissed):

giancorderoortiz (approved):

PR #19692, opened 2024-12-03
fix: (CXSPA-9038) add label to store finder search input

uroslates:
Closes: https://jira.tools.sap/browse/CXSPA-9038

### DIFF
--- a/feature-libs/storefinder/components/store-finder-search/store-finder-search.component.html
+++ b/feature-libs/storefinder/components/store-finder-search/store-finder-search.component.html
@@ -33,7 +33,7 @@
   <div class="row align-items-end" *cxFeature="'a11yStoreFinderLabel'">
     <div class="col-md-12 col-lg-6">
       <div class="form-group">
-        <label class="">
+        <label>
           <span class="text-left label-content">
             {{ 'storeFinder.findStoreBy' | cxTranslate }}
           </span>
--- a/feature-libs/storefinder/components/store-finder-search/store-finder-search.component.html
+++ b/feature-libs/storefinder/components/store-finder-search/store-finder-search.component.html
@@ -1,30 +1,41 @@
 <div class="container">
-  <div class="row">
+  <div class="row align-items-end">
     <div class="col-md-12 col-lg-6">
-      <div class="form-group search-wrapper">
-        <input
-          #queryInput
-          [formControl]="searchBox"
-          (keyup)="onKey($event)"
-          type="text"
-          class="form-control"
-          [attr.aria-label]="'storeFinder.searchBoxLabel' | cxTranslate"
-          placeholder="{{ 'storeFinder.searchBox' | cxTranslate }}"
-        />
-        <cx-icon
-          [attr.tabindex]="queryInput.value?.length ? 0 : -1"
-          [type]="iconTypes.SEARCH"
-          role="button"
-          [attr.aria-label]="'storeFinder.searchNearestStores' | cxTranslate"
-          [attr.title]="'storeFinder.searchNearestStores' | cxTranslate"
-          class="search"
-          (keyup)="onKey($event)"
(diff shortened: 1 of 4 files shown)


feature-libs/storefinder/components/store-finder-search/store-finder-search.component.html
```diff
@@ -1,30 +1,41 @@
 <div class="container">
-  <div class="row">
+  <div class="row align-items-end">
```
Zeyber: I can see classes and elements have been added/removed and the DOM structure is changed. Could we please hide these changes behind a feature toggle?

Zeyber (changes requested):

feature-libs/storefinder/components/store-finder-search/store-finder-search.component.html
uroslates: Applied! Hidden behind **a11yStoreFinderLabel** feature flag!

feature-libs/storefinder/components/store-finder-search/store-finder-search.component.html
```diff
@@ -27,25 +27,71 @@
...
-        <div class="col-6">
-          <button
-            [routerLink]="['/store-finder/view-all']"
-            class="btn btn-primary btn-block"
-          >
-            {{ 'storeFinder.viewAllStores' | cxTranslate }}
-          </button>
-        </div>
+      <div class="form-group">
+        <label class="">
```
Pio-Bar: Empty `class` attribute, likely unnecessary ;)

Pio-Bar: Maybe including the label sth like this could help avoid the layout change? Just a thought 😉 

```
<label *cxFeature="'a11yStoreFinderLabel'" for="queryInput" class="text-left label-content">
  {{ 'storeFinder.findStoreBy' | cxTranslate }}
</label>
```

Pio-Bar (changes requested):

feature-libs/storefinder/components/store-finder-search/store-finder-search.component.html
uroslates: Ah sorry its now removed!

uroslates: Unfortunately this is not enough because than we have layout alignment issue as in a screenshoot below.
find-a-store-layout-alignment-issue

To address that I had to add extra class to the parent .row element (and since its new class I made it hidden behind the flag)!

Pio-Bar (dismissed):

Zeyber (dismissed):

feature-libs/storefinder/components/store-finder-search/store-finder-search.component.html
developpeurweb: @uroslates new label is fine. However one last fix needed here, as I'm sure auditors will point out the issue. Right now, JAWS still reads the aria-label in the field and ignores the visible label.
When we have a visible label, we need to programmatically associate it with the field and remove aria-label(s) since they override standard labels. In the case, we had an aria-label instead of a visible label, so we have to remove the aria-label. 
Therefore, can you please:
- Remove the aria-label from the field.
- Associate the visible label to the field.

developpeurweb (approved):

feature-libs/storefinder/components/store-finder-search/store-finder-search.component.html
uroslates: Applied!

uroslates: @Piotr I have applied the changes as @developpeurweb  requested. Could U pls dbl check if its OK prior to merging to develop? Thanks!

Pio-Bar (approved):

PR #19694, opened 2024-12-03
fix: reset focus to the main after navigating

sdrozdsap:
Closes: https://jira.tools.sap/browse/CXSPA-9056

### DIFF
--- a/projects/storefrontlib/layout/main/storefront.component.ts
+++ b/projects/storefrontlib/layout/main/storefront.component.ts
@@ -13,6 +13,7 @@ import {
   inject,
   OnDestroy,
   OnInit,
+  Optional,
   ViewChild,
 } from '@angular/core';
 import {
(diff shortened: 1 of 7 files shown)


projects/storefrontlib/layout/main/storefront.component.ts
```diff
@@ -138,4 +143,29 @@ export class StorefrontComponent implements OnInit, OnDestroy {
...
+  protected onNavigation(isNavigating: boolean): void {
+    this.startNavigating = isNavigating === true;
+    this.stopNavigating = isNavigating === false;
+    if (
+      this.featureConfigService.isEnabled('a11yResetFocusAfterNavigating') &&
+      this.stopNavigating &&
+      // we only need to reset focus after clicking a link
+      this.document?.activeElement !== this.document?.body
+    ) {
+      this.skipFocusTo('cx-main');
```
Zeyber: I think we should comment on why we need to do this.

projects/storefrontlib/layout/main/storefront.component.ts
```diff
@@ -138,4 +143,29 @@ export class StorefrontComponent implements OnInit, OnDestroy {
...
+  protected skipFocusTo(skipLinkKey: string): void {
+    this.skipLinkService
+      ?.getSkipLinks()
+      .pipe(take(1))
+      .subscribe((skipLinks) => {
+        const skipLink = skipLinks.find((link) => link.key === skipLinkKey);
+        if (skipLink) {
+          this.skipLinkService?.scrollToTarget(skipLink);
+        }
+      });
```
Zeyber: Interesting solution to use the skip link service.

Maybe we could have `key` as a possible parameter for the `scrollToTarget` method OR `scrollToTargetByKey` in the skip link service as this method could be useful in other places. I don't think we need `skipFocusTo` in the storefront component.

Zeyber (changes requested):

projects/storefrontlib/layout/main/storefront.component.ts
sdrozdsap: Good suggestion!

Zeyber (dismissed):

developpeurweb (dismissed):

developpeurweb (dismissed):

Pio-Bar (approved):

PR #19711, opened 2024-12-04
fix: elevate scroll to top button when overlaps with banner

sdrozdsap:
Closes: https://jira.tools.sap/browse/CXSPA-9067

### DIFF
--- a/projects/storefrontlib/cms-components/navigation/scroll-to-top/scroll-to-top.component.ts
+++ b/projects/storefrontlib/cms-components/navigation/scroll-to-top/scroll-to-top.component.ts
@@ -20,8 +20,10 @@ import {
   FeatureConfigService,
   ScrollBehavior,
   WindowRef,
+  AnonymousConsentsService,
+  useFeatureStyles,
 } from '@spartacus/core';
-import { take } from 'rxjs/operators';
+import { take, Observable } from 'rxjs';
 import { CmsComponentData } from '../../../cms-structure/page/model/cms-component-data';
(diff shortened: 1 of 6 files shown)


projects/storefrontlib/cms-components/navigation/scroll-to-top/scroll-to-top.component.ts
```diff
@@ -49,15 +56,32 @@ export class ScrollToTopComponent implements OnInit {
...
+    useFeatureStyles('a11yScrollToTopPositioning');
+  }
 
   ngOnInit(): void {
     this.setConfig();
+    if (this.featureConfigService?.isEnabled('a11yScrollToTopPositioning')) {
+      this.subscription.add(
+        this.anonymousConsentsService
+          ?.isBannerVisible()
+          .subscribe((visible) => {
```
Zeyber: Do you think we could use an observable to set the class value instead of making a subscription? Less code and then we can use the async pipe.

Zeyber (changes requested):

projects/storefrontlib/cms-components/navigation/scroll-to-top/scroll-to-top.component.ts
sdrozdsap: Yes, I was able to reimplement it with async pipe utilizing `:has` pseudoselector in scss. Works event better now!

Zeyber (approved):

PR #19720, opened 2024-12-05
fix: resolved the issue of infinite loading in SmartEdit when accessing the PDP

kpawelczak:
closes: [CXSPA-8697](https://jira.tools.sap/browse/CXSPA-8697)

### DIFF
--- a/feature-libs/product-multi-dimensional/selector/components/guards/product-multi-dimensional-selector.guard.ts
+++ b/feature-libs/product-multi-dimensional/selector/components/guards/product-multi-dimensional-selector.guard.ts
@@ -40,8 +40,10 @@ export class ProductMultiDimensionalSelectorGuard {
     activatedRoute: ActivatedRouteSnapshot
   ): Observable<boolean | UrlTree> {
     const productCode = activatedRoute.params?.productCode;
+
     if (!productCode) {
-      return of(false);
+      // Refuse entry unless it is within a SmartEdit environment
+      return of(!!activatedRoute.queryParams?.cmsTicketId);
     }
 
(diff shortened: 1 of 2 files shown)


feature-libs/product-multi-dimensional/selector/components/guards/product-multi-dimensional-selector.guard.ts
```diff
@@ -40,8 +40,10 @@ export class ProductMultiDimensionalSelectorGuard {
     activatedRoute: ActivatedRouteSnapshot
   ): Observable<boolean | UrlTree> {
     const productCode = activatedRoute.params?.productCode;
+
     if (!productCode) {
-      return of(false);
+      // Refuse entry unless it is within a SmartEdit environment
+      return of(!!activatedRoute.queryParams?.cmsTicketId);
```
rmch91, line 46: Can you add additional test scenarios, please?
For example:

1. should return true if no product code is provided but cmsTicketId exists
2. should return false if product code is not provided but queryParams are empty

Also the original one "should return false if no product code is provided" should be adjusted since it shouldn't return false anymore in only there is no product code, but we have additional condition.

rmch91 (changes requested):

feature-libs/product-multi-dimensional/selector/components/guards/product-multi-dimensional-selector.guard.ts
kpawelczak: thanks, fixed

rmch91 (approved):

PR #19731, opened 2024-12-06
fix: (CXSPA-9092) Read out group name when navigating with prev/next button

steinsebastian:


feature-libs/product-configurator/rulebased/core/facade/configurator-groups.service.ts
```diff
@@ -269,6 +269,28 @@ export class ConfiguratorGroupsService {
...
+  /**
+   * Returns the description of the next group.
+   *
+   * @param {Configurator.Configuration} configuration - Configuration
+   * @returns {string} the description of the next group
+   */
+  getNextGroupDescription(configuration: Configurator.Configuration): string {
+    let description: string = '';
+    this.getNextGroupId(configuration.owner)
+      .pipe(take(1))
```
ChristophHi: Let's extract this into a common (protected) function and use it also in getPreviousGroupDescription...

ChristophHi (changes requested):

feature-libs/product-configurator/rulebased/core/facade/configurator-groups.service.ts
```diff
@@ -269,6 +269,28 @@ export class ConfiguratorGroupsService {
     return this.getNeighboringGroupId(owner, 1);
   }
 
+  /**
+   * Returns the description of the next group.
+   *
+   * @param {Configurator.Configuration} configuration - Configuration
+   * @returns {string} the description of the next group
+   */
+  getNextGroupDescription(configuration: Configurator.Configuration): string {
```
ChristophHi: Also: It feels more natural (and is also more robust) to return an observable here. Because you are implicitly assuming that this.getNextGroupId is emitting instantly.
So we would deal with observables in all layers up to the UI and only there use async pipe to subscribe

ChristophHi (changes requested):

feature-libs/product-configurator/rulebased/core/facade/configurator-groups.service.ts
steinsebastian: done.

feature-libs/product-configurator/rulebased/core/facade/configurator-groups.service.ts
steinsebastian: done

ChristophHi (approved):

PR #19735, opened 2024-12-06
fix: move anonymous consent notification inside dialog

sdrozdsap:
Closes: https://jira.tools.sap/browse/CXSPA-9030


projects/storefrontlib/cms-components/myaccount/consent-management/components/consent-form/consent-management-form.component.ts
Pio-Bar, line 1: Should we include those changes within the feature toggle too?

projects/storefrontlib/cms-components/myaccount/consent-management/components/consent-form/consent-management-form.component.html
```diff
@@ -4,7 +4,7 @@
       type="checkbox"
       class="form-check-input"
       (change)="onConsentChange()"
-      [checked]="consentGiven"
```
Pio-Bar: If we add the feature flag to the component file we might need one here too

Pio-Bar (changes requested):

projects/storefrontlib/cms-components/myaccount/consent-management/components/consent-form/consent-management-form.component.html
sdrozdsap: Agree

projects/storefrontlib/cms-components/myaccount/consent-management/components/consent-form/consent-management-form.component.ts
sdrozdsap: Done

Pio-Bar (dismissed):

projects/storefrontlib/shared/components/anonymous-consents-dialog/anonymous-consent-dialog.component.ts
```diff
@@ -214,6 +243,10 @@ export class AnonymousConsentDialogComponent implements OnInit, OnDestroy {
     }
   }
 
+  closeMessage(): void {
```
Zeyber, line 246: Lets add a quick unit test for this method since it isn't covered afaik

projects/storefrontlib/cms-components/myaccount/consent-management/components/consent-form/consent-management-form.component.ts
```diff
@@ -59,7 +87,9 @@ export class ConsentManagementFormComponent implements OnInit {
     this.consentGiven = !this.consentGiven;
 
     this.consentChanged.emit({
-      given: this.consentGiven,
+      given: this.useGetterForIsConsentGiven
```
Zeyber: Is using a getter going to have performance impacts? As in, will the function be repeatedly called to check the consent? 
I think using a`BehaviorSubject` or `Observable` could be a better option.

Zeyber (changes requested):

projects/storefrontlib/cms-components/myaccount/consent-management/components/consent-form/consent-management-form.component.ts
sdrozdsap: The issue here is that this property is a derivative state from the `@Input consent` and theoretically we can either set `this.consentGiven` inside `ngOnChanges` where we would react to the changes for the inputs(we can even push these values to the subject) or use this getter which will automatically return updated value based on current input. Since there are no heavy calculations, or any calculation at all for that matter I've decided to go with the getter approach. 
To reduce number of calls we can create a pipe which will only be fired if the input changes, and not on each re-render, but for implementing ngOnChanges and push values to the subject just doesn't sound like a very clean approach, that's just my opinion - or did you have something else in mind? Please let me know what do you think which approach is the best here

projects/storefrontlib/shared/components/anonymous-consents-dialog/anonymous-consent-dialog.component.ts
sdrozdsap: Done

projects/storefrontlib/cms-components/myaccount/consent-management/components/consent-form/consent-management-form.component.ts
Zeyber: Thank you for highlighting the reasoning and considerations here. I agree that the "getter" method is clean and responsive and I would really love for it to be that simple. But I think we need to emphasise performance as better practice. I think [this approach with signals](https://blog.angular-university.io/angular-signal-inputs/) would be the best solution to our situation. But we might need to fall back to the ngOnChanges method until next major when we support a later version of angular.

sdrozdsap: I agree, I've applied `ngOnChanges` approach

Zeyber (approved):

developpeurweb (approved):

PR #19737, opened 2024-12-06
fix: prevent anonymous consent request from being sent for logged in

sdrozdsap:
After logging in there is an error in console and the banner saying "You are not authorized to perform this action. Please contact your administrator if you think this is a mistake."
image

This is caused by anonymous consents request being sent when the user is logged in


Zeyber (dismissed):

sdrozdsap:
Closed in favor of: https://github.com/SAP/spartacus/pull/19755

PR #19754, opened 2024-12-10
fix: (CXSPA-9031) - Refocus on card after selecting

Pio-Bar:
Closes: [CXSPA-9031](https://jira.tools.sap/browse/CXSPA-9031)

I've found a similar problem exists for delivery checkout step too. I've made a solution that fixes both cases.


projects/storefrontlib/layout/a11y/keyboard-focus/services/select-focus.util.ts
```diff
@@ -92,4 +96,34 @@ export class SelectFocusUtility {
   protected isHidden(el: HTMLElement): boolean {
     return el.offsetParent === null;
   }
+
+  /**
+   * Restores the focus to the Card component after it has been selected and the checkout has finished updating.
+   * It is used for cases where the focus is lost due to DOM changes making it impossible to target elements that have been modified.
+   * @param isUpdating$ An observable that emits a boolean to indicate whether the component is updating.
+   */
+  focusCardAfterSelecting(isUpdating$: Observable<boolean>): void {
```
Zeyber: I don't think this method belongs in this service. It is specific to the Card component and caters to a somewhat niche usecase. Here are some options (but certainly not all):
- We make this method generic and useful for it to belong in the SelectFocusUtility.
- We move it to the Checkout components that this method affects.
- We add it to a shared service between those components.
- Hybrid-approach using some of each option.
- If you think of something better, please go ahead with it :)

Zeyber (changes requested):

projects/storefrontlib/layout/a11y/keyboard-focus/services/select-focus.util.ts
Pio-Bar: I think your are right. I've added it separately for both components. If I strip the specifics there is not enough to warrant a new service or even a reusable method.

Zeyber (dismissed):

developpeurweb (dismissed):

Zeyber (approved):

PR #19779, opened 2024-12-13
CXSPA-8566 bug fix

suprishi:
https://jira.tools.sap/browse/CXSPA-8566

returning `undefined` if `intendedLocation.displayName` is not defined, so that pickup option is reset when navigating between 2 products.


feature-libs/pickup-in-store/components/container/pdp-pickup-options-container/pdp-pickup-options-container.component.ts
```diff
@@ -114,34 +114,15 @@ export class PdpPickupOptionsContainerComponent implements OnInit, OnDestroy {
...
-                    ...storeDetails,
-                    pickupOption: 'delivery',
-                  }
-                );
-              })
-            )
-        )
-      ),
-      tap(() => (this.displayNameIsSet = true))
+      switchMap(({ intendedLocation, productCode }) => {
```
Matejk00, line 117: What do you think about this change proposal?

```
switchMap(({ intendedLocation, productCode }) => {
  if (intendedLocation?.displayName) {
    this.displayNameIsSet = true;
    return of(getProperty(intendedLocation, 'displayName'));
  }

  this.setIntendedPickupLocation(productCode);
  return of(undefined);
})
```

The use of optional chaining (?.) simplifies the null checks and removing the double negation (!!) makes the condition cleaner. Also, dropping the unnecessary else block aligns well with an early return pattern. I think it is improving readability.

feature-libs/pickup-in-store/components/container/pdp-pickup-options-container/pdp-pickup-options-container.component.ts
```diff
@@ -177,6 +158,27 @@ export class PdpPickupOptionsContainerComponent implements OnInit, OnDestroy {
...
+        map(({ name }) => name),
+        tap((storeName) =>
+          this.pickupLocationsSearchService.loadStoreDetails(storeName)
+        ),
+        concatMap((storeName: string) =>
+          this.pickupLocationsSearchService.getStoreDetails(storeName)
+        ),
+        filter((storeDetails) => !!storeDetails)
+      )
+      .subscribe((storeDetails) => {
```
Matejk00: Shouldn't we unsubscribe this subscription somewhere? I am not sure if it not cause a potential memory leaks.

Matejk00 (changes requested):

feature-libs/pickup-in-store/components/container/pdp-pickup-options-container/pdp-pickup-options-container.component.ts
suprishi: Agreed, I've updated the code as per your proposal.

feature-libs/pickup-in-store/components/container/pdp-pickup-options-container/pdp-pickup-options-container.component.ts
suprishi: I've added this to the list of subscriptions which get unsubscribed within `ngOnDestroy()`.

Matejk00 (approved):

PR #19782, opened 2024-12-16
fix: (CXSPA-8997 & 8988) - Resolve Aria Issues in NavigationUIComponent

petarmarkov9449:
Closes: [CXSPA-8997](https://jira.tools.sap/browse/CXSPA-8997), [CXSPA-8998](https://jira.tools.sap/browse/CXSPA-8998)

Removes invalid aria-level usage on button elements and ensures buttons have a proper accessible name via aria-label or aria-labelledby.

Affects: NavigationUIComponent


projects/storefrontlib/cms-components/navigation/navigation/navigation-ui.component.html
```diff
@@ -102,26 +102,50 @@
...
-            <cx-icon [type]="iconType.CARET_DOWN"></cx-icon>
-          </button>
+          <ng-container *cxFeature="'!a11yButtonAriaFixes'">
+            <button
+              aria-level="4"
+              [attr.role]="(isDesktop$ | async) && depth ? 'heading' : 'button'"
+              [attr.aria-haspopup]="true"
+              [attr.aria-expanded]="false"
+              [attr.aria-controls]="node.title"
+              [attr.aria-describedby]="getAriaDescribedbyOfButton(node)"
```
sdrozdsap: You've added it for negated cxFeature therefore if a11yButtonAriaFixes is enabled this won't be used. 
However when I change it on my local it solves the previous errors, but there is one more new one:
image

projects/storefrontlib/cms-components/navigation/navigation/navigation-ui.component.ts
```diff
@@ -396,4 +396,18 @@ export class NavigationUIComponent implements OnInit, OnDestroy {
     }
     return depth > 0 && !node?.children ? -1 : 0;
   }
+
+  /**
+   * Returns the ID for the `aria-describedby` attribute of a button.
+   */
+  getAriaDescribedbyOfButton(node: NavigationNode): string | null {
+    return node.title === 'My Account' ? 'greeting' : null;
```
sdrozdsap: Hm, it doesn't seem right to use the title as an identifier of the node. If this aria described by is used only for my account navigation, maybe we can pass down some input from `my-account-v2-navigation.component.ts` where `<cx-navigation-ui>` is used. Thats just from the top of my head. 
I'm not sure if using 'My Account' would work with i18n...

projects/core/src/features-config/feature-toggles/config/feature-toggles.ts
```diff
@@ -759,6 +759,12 @@ export interface FeatureTogglesInterface {
    */
   a11yAddPaddingToCarouselPanel?: boolean;
 
+  /**
+   * Removes invalid aria-level usage on button elements and ensures buttons have a proper accessible name via aria-label or aria-labelledby.
+   * Affects: NavigationUIComponent
+   */
+  a11yButtonAriaFixes?: boolean;
```
sdrozdsap: maybe we should be more specific like `a11yNavigationButtonsAriaFixes`

sdrozdsap (approved):

Zeyber (approved):

PR #19786, opened 2024-12-17
fix: (CXSPA-8999) - Descriptive repeating buttons

Pio-Bar:
Closes: [CXSPA-8999](https://jira.tools.sap/browse/CXSPA-8999)

Repeating buttons with the same inner text should now be associated with a unique name:

- PLP 'add to cart 'buttons should additionally read out the product name.
- Buttons inside 'store list' items now include the store name in their read out.


feature-libs/pickup-in-store/components/container/set-preferred-store/set-preferred-store.component.html
```diff
@@ -20,6 +20,13 @@
     data-text="setPreferredStore.myStore"
     [attr.data-preferred-store]="pointOfServiceName.name"
     class="set-preferred-heading"
+    [attr.aria-label]="
```
sdrozdsap: Maybe we can move this to the component? you can have a method e.g. `getAriaLabel(storeSelected) {...}` and call it from the template: `getAriaLabel((storeSelected$ | async)?.name)`

sdrozdsap: Although looking at the component again I see there are multiple `(storeSelected$ | async)?.name` which creates redundant subscriptions. My suggestion is to add `*ngIf="{ storeSelected: storeSelected$ | async } as data"` to some top-level div or just create new ng-container. This expression will always return true but you will have only one subscription and then everywhere instead of `(storeSelected$ | async)?.name` you can use `data.storeSelected?.name`

Pio-Bar: I've implemented both ideas, thanks!

feature-libs/pickup-in-store/components/container/set-preferred-store/set-preferred-store.component.html
```diff
@@ -1,30 +1,64 @@
...
+    "
+    class="setpreferredstore-container"
+    (click)="setAsPreferred()"
   >
-    <cx-icon aria-hidden="true" [type]="ICON_TYPE.HEART"></cx-icon>
+    <div
+      [ngClass]="{
+        'icon-selected': pointOfServiceName.name === storeSelected.name,
+        'icon-not-selected': pointOfServiceName.name !== storeSelected.name,
+      }"
```
sdrozdsap:
```suggestion
      [ngClass]="pointOfServiceName.name === storeSelected.name ? 'icon-selected' : 'icon-not-selected'"
```

feature-libs/pickup-in-store/components/container/set-preferred-store/set-preferred-store.component.html
```diff
@@ -1,30 +1,64 @@
...
+    class="setpreferredstore-container"
+    (click)="setAsPreferred()"
+  >
+    <div
+      [ngClass]="{
+        'icon-selected':
+          pointOfServiceName.name === (storeSelected$ | async)?.name,
+        'icon-not-selected':
+          pointOfServiceName.name !== (storeSelected$ | async)?.name,
+      }"
```
sdrozdsap, line 53: Leftover `(storeSelected$ | async)?.name`

```suggestion
     [ngClass]="pointOfServiceName.name === storeSelected.name ? 'icon-selected' : 'icon-not-selected'"
```

feature-libs/pickup-in-store/components/container/set-preferred-store/set-preferred-store.component.html
```diff
@@ -1,30 +1,64 @@
...
+    >
+      <cx-icon aria-hidden="true" [type]="ICON_TYPE.HEART"></cx-icon>
+    </div>
+    <button
+      data-text="setPreferredStore.myStore"
+      [attr.data-preferred-store]="pointOfServiceName.name"
+      class="set-preferred-heading"
+    >
+      {{
+        pointOfServiceName.name === (storeSelected$ | async)?.name
```
sdrozdsap, line 63: Leftover `(storeSelected$ | async)`

feature-libs/pickup-in-store/components/container/set-preferred-store/set-preferred-store.component.html
```diff
@@ -1,30 +1,64 @@
...
->
+<ng-container *cxFeature="'a11yRepeatingButtonsUniqueLabels'">
   <div
-    [ngClass]="{
-      'icon-selected':
-        pointOfServiceName.name === (storeSelected$ | async)?.name,
-      'icon-not-selected':
-        pointOfServiceName.name !== (storeSelected$ | async)?.name,
-    }"
+    *ngIf="storeSelected$ | async as storeSelected"
```
sdrozdsap: I suggest wrapping the whole component in:
```html
<ng-container *ngIf="{ storeSelected: storeSelected$ | async } as data>...</ng-container>
```

As it doesn't introduce any new logic, as in your example `*ngIf="storeSelected$ | async as storeSelected"` will only work if emitted value is truthy. I'm not sure about the case here, but just for safety I prefer the approach above. Also you will have access to `data.storeSelected`  everywhere in the template

feature-libs/pickup-in-store/components/container/set-preferred-store/set-preferred-store.component.html
Pio-Bar: I've included the suggestion, thanks!

feature-libs/pickup-in-store/components/container/set-preferred-store/set-preferred-store.component.html
Pio-Bar: We should leave the old code as is to avoid breaking changes.

feature-libs/pickup-in-store/components/container/set-preferred-store/set-preferred-store.component.html
Pio-Bar: We should leave the old code as is to avoid breaking changes.

sdrozdsap (dismissed):

Zeyber (dismissed):

Zeyber (approved):

PR #19790, opened 2024-12-18
fix: use role button for carousel slide indicators

sdrozdsap:
Closes: https://jira.tools.sap/browse/CXSPA-9000

### DIFF
--- a/projects/storefrontlib/shared/components/carousel/carousel.component.html
+++ b/projects/storefrontlib/shared/components/carousel/carousel.component.html
@@ -96,11 +96,7 @@
     </ng-container>
   </div>
 
-  <div
-    *ngIf="!hideIndicators && size < items.length"
-    role="tablist"
-    class="indicators"
-  >
+  <div *ngIf="!hideIndicators && size < items.length" class="indicators">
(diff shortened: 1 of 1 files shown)


projects/storefrontlib/shared/components/carousel/carousel.component.html
```diff
@@ -104,31 +104,61 @@ <h2 *ngIf="title">{{ title }}</h2>
...
-          "
-          [attr.title]="
-            'carousel.slideNumber'
-              | cxTranslate
-                : {
-                    currentSlideNumber: getSlideNumber(size, i),
-                    numberOfSlides: size,
-                  }
-          "
-          role="tab"
```
Pio-Bar, line 128: I believe this has been included to mimic the [tabs pattern](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/) as we decided this was best at the time. If we are moving away from it we probably should also remove `role="tablist"` from the wrapper div. 

I think removing those roles isn't a breaking change. We can get away with not using a feature flag here at least we did so in the past 😉 .

Pio-Bar (changes requested):

Pio-Bar (approved):

PR #19797, opened 2024-12-20
no bootstrap spartacus installation

kpawelczak:
Moved bootstrap styles to storefront styles `vendor`


extra-webpack.config.js
```diff
@@ -23,7 +23,7 @@ module.exports = {
   ],
   resolve: {
     fallback: {
-      bootstrap: path.join(__dirname, 'projects/styles/bootstrap-copy'),
+      bootstrap: path.join(__dirname, 'projects/storefrontstyles/bootstrap-copy'),
```
pawelfras: `minor:` this `bootstrap` fallback alias is no longer needed as there is no direct reference to `'bootstrap'` location in our app anymore. Bootstrap copy is part of `@spartacus/styles` (handled by the config below).

projects/schematics/src/add-spartacus/index.ts
```diff
@@ -148,7 +148,26 @@ function installStyles(options: SpartacusOptions): Rule {
       styleFilePath
     );
     let insertion =
-      `\n@import '${relativeStyleConfigImportPath}';\n` +
+      `@import '${relativeStyleConfigImportPath}';\n` +
+      `\n// ORDER IMPORTANT: Spartacus core first\n` +
+      `@import '@spartacus/styles/scss/core';\n\n` +
+      `// ORDER IMPORTANT: Bootstrap next\n` +
```
pawelfras:
```suggestion
      `// ORDER IMPORTANT: Copy of Bootstrap files next\n` +
```

extra-webpack.config.js
kpawelczak: removed

projects/schematics/src/add-spartacus/index.ts
kpawelczak: thanks, I've updated it

projects/storefrontstyles/vendor/bootstrap/scss/mixins/_image.scss
```diff
@@ -25,9 +25,9 @@
...
   // Compatibility info: https://caniuse.com/css-media-resolution
-  @media only screen and (min-resolution: 192dpi) {
-    // IE9-11 don't support dppx only screen and (min-resolution: 2dppx) {
-    // Standardized
+  @media only screen and (min-resolution: 192dpi),
+    only screen and (min-resolution: 2dppx) {
+    // IE9-11 don't support dppx
     background-image: url($file-2x);
     background-size: $width-1x $height-1x;
   }
```
Platonn: Is this change really related to this PR?

kpawelczak: our prettier script keeps breaking this file

projects/schematics/src/add-spartacus/index.ts
```diff
@@ -148,7 +148,26 @@ function installStyles(options: SpartacusOptions): Rule {
       styleFilePath
     );
     let insertion =
-      `\n@import '${relativeStyleConfigImportPath}';\n` +
```
Platonn, line 151: Perhaps I'm missing something. But at the first sight this question comes to my mind: what about the installation schematics of other libs? Shouldn't they be also updated, to inject those 3 lines into scss feature-files in customer's app?:

```scss
@import '@spartacus/styles/vendor/bootstrap/scss/functions';
@import '@spartacus/styles/vendor/bootstrap/scss/variables';
@import '@spartacus/styles/vendor/bootstrap/scss/_mixins';
```

kpawelczak: do they import Bootstrap directly in the customer app?

projects/storefrontstyles/vendor/bootstrap/scss/mixins/_image.scss
Platonn: Do I understand correctly that previously the file was broken by prettier and now you're fixing it manually?

projects/schematics/src/add-spartacus/index.ts
Platonn: As you can see in this PR, we have many SCSS files which are separate fine grained entry points for feature libs styles, e.g.
`integration-libs/opf/_index.scss`
`feature-libs/storefinder/styles/_index.scss`
etc.
and in this PR we've added those necessary 3 lines.

When you create a fresh app with our schematics, analogical files will be created in customers' app for each feature lib. I guess we need to inject those 3 lines there too. Otherwise I think the customers' app won't compile becasue of missing stuff expected from botostrap imports that are missing - in those feature SCSS files

projects/storefrontstyles/vendor/bootstrap/scss/mixins/_image.scss
kpawelczak: correct

projects/schematics/src/add-spartacus/index.ts
kpawelczak: I think I'm missing the point here.
this is cart.scss in customer app:
```scss
@import "../../styles-config";
@import "@spartacus/cart";
``` 
and @spartacus/cart looks like:

```scss
@import '@spartacus/styles/scss/cxbase/blocks/modal';
@import '@spartacus/styles/scss/misc/table';

@import '@spartacus/styles/vendor/bootstrap/scss/functions';
@import '@spartacus/styles/vendor/bootstrap/scss/variables';
@import '@spartacus/styles/vendor/bootstrap/scss/_mixins';
....
```

projects/storefrontstyles/vendor/bootstrap/scss/mixins/_image.scss
kpawelczak: I've added the subfolder to the ignore list, in the another pr

projects/schematics/src/add-spartacus/index.ts
Platonn: Ah you're right!

I somehow understood (my bad!) that we already moved the imports of those 3 bootstrap lines already from Spartacus feature libs' internals into the relevant entry point files in the `storefrontapp` (I mean e.g. `projects/storefrontapp/src/styles/lib-asm.scss` etc). With this (wrong) assumption in mind, I was asking if we're going to reflect the same structure in the app created with schematics. 😅 

That said,
Would you agree it would be good to also move those 3 bootstrap lines to relevant files in storefrontapp ? (similarly to how we moved bootstrap imports from `@spartacus/styles`' internals to the main scss file in the storefrontapp - `projects/storefrontapp/src/styles/styles.scss`?

kpawelczak: Thanks for bringing this up. There is a separate ticket for this specific issue. We agreed at some point that this will be an optional task, to be addressed after we merge the must-have changes.

Platonn (approved):

PR #19803, opened 2024-12-23
fix: test and fix dialogs affected by the removal of extended bootstrap clases

Pio-Bar:
Closes: https://jira.tools.sap/browse/CXSPA-9185

Removes commented out `@extend .modal...` rules and get rids of the remaining ones. I've manually tested each affected dialog and made sure it displays as expected.


feature-libs/user/profile/styles/_close-account-modal.scss
```diff
@@ -14,10 +9,6 @@
         }
       }
 
-      .cx-close-account-modal-body {
```
kpawelczak, line 17: I cant find it in the diff. Did you remove the class from the HTML?

feature-libs/user/profile/styles/_suggested-addresses-dialog.scss
```diff
@@ -16,10 +10,6 @@
         }
       }
 
-      .cx-suggested-addresses-body {
```
kpawelczak, line 19: Should we remove `cx-suggested-addresses-body` from the html?

kpawelczak (changes requested):

feature-libs/user/profile/styles/_close-account-modal.scss
Pio-Bar: Removed after changes

feature-libs/user/profile/styles/_suggested-addresses-dialog.scss
Pio-Bar: Removed the class from the HTML too.

kpawelczak (approved):

PR #19825, opened 2025-01-03
fix: add opf-load-once capability on js and css resources

FollowTheFlo:


integration-libs/opf/base/root/services/opf-resource-loader.service.ts
```diff
@@ -22,11 +22,12 @@ export class OpfResourceLoaderService {
   protected document = inject(DOCUMENT);
   protected platformId = inject(PLATFORM_ID);
 
-  protected readonly OPF_RESOURCE_ATTRIBUTE_KEY = 'data-opf-resource';
   protected readonly CORS_DEFAULT_VALUE = 'anonymous';
+  protected readonly OPF_RESOURCE_LOAD_ONCE_ATTR_KEY = 'opf-load-once';
```
Matejk00: Could we use for consistency: `OPF_RESOURCE_LOAD_ONCE_ATTRIBUTE_KEY`?

Matejk00: Also would be good in my opinion to prefix all custom html attributes with "data-".
https://www.w3schools.com/tags/att_data-.asp

Can we change attribute key to: `data-opf-load-once`?

integration-libs/opf/base/root/services/opf-resource-loader.service.ts
```diff
@@ -22,11 +22,12 @@ export class OpfResourceLoaderService {
...
+  protected readonly OPF_RESOURCE_ATTRIBUTE_KEY = 'data-opf-resource';
 
   protected embedStyles(embedOptions: {
-    attributes?: OpfKeyValueMap[];
+    attributes?: { [key: string]: string };
```
Matejk00, line 30: Is it new type different than `OpfKeyValueMap[]`?

integration-libs/opf/base/root/services/opf-resource-loader.service.ts
```diff
@@ -74,6 +71,30 @@ export class OpfResourceLoaderService {
...
 
+  /**
+   * Create attributes intended to script and link elements.
+   *
+   * Return attributes list including keyValueList and OPF specific attribute with below logic:
+   *
+   * 1. Resource loads only once: 'opf-load-once' key detected, no additional attribute added.
+   * 2. Resource deleted at page/payment change: 'data-opf-resource' attribute is added.
+   */
+
```
Matejk00, line 82: Nitpick: There is a space between a comment and function name.

integration-libs/opf/base/root/services/opf-resource-loader.service.ts
```diff
@@ -82,31 +103,21 @@ export class OpfResourceLoaderService {
    */
   protected loadScript(resource: OpfDynamicScriptResource): Promise<void> {
     return new Promise((resolve, reject) => {
-      const attributes: any = {
+      const attributes: { [key: string]: string } = {
```
Matejk00, line 106: Can we create type for it in a model if we cannot use: `OpfKeyValueMap`?

Matejk00 (changes requested):

integration-libs/opf/base/root/services/opf-resource-loader.service.ts
FollowTheFlo: thx for the info, i'll align with Bamboo to add 'data-' prefix.

FollowTheFlo: No need to add 'data-' prefix:
'opf-load-once' is never added as attribute key, it is just a variable that decides if we add 'data-opf-resource' attribute or not.
To sum-up, we receive it in attributes property from server but using it as a flag in this service.

integration-libs/opf/base/root/services/opf-resource-loader.service.ts
FollowTheFlo: No, it was set as 'any'.
OpfKeyValueMap is {key:string,value:string}, we receive the list in this format from resource object (resource.attributes).
We perform a conversion to [key: string]: string to make handling easier and have same type as attributes in ScriptLoader service

integration-libs/opf/base/root/services/opf-resource-loader.service.ts
FollowTheFlo: Mentioned above, we need both:
OpfKeyValueMap is the original format (resource.attributes received from server)
{ [key: string]: string } is format needed for ScriptLoaderService
Thus a conversion is needed.

integration-libs/opf/base/root/services/opf-resource-loader.service.ts
```diff
@@ -74,6 +71,30 @@ export class OpfResourceLoaderService {
...
+   */
+
+  protected createAttributesList(keyValueList?: OpfKeyValueMap[] | undefined): {
+    [key: string]: string;
+  } {
+    const attributes: { [key: string]: string } = {};
+    keyValueList?.forEach((keyValue: OpfKeyValueMap) => {
+      attributes[keyValue.key] = keyValue.value;
+    });
+    if (!attributes[this.OPF_RESOURCE_LOAD_ONCE_ATTR_KEY]) {
```
FollowTheFlo: Just realized, we only check its presence but not value 'true'.
Still valid as server as unique value ('true') associated with this key.
It could be more robust by checking it.

Matejk00 (approved):
LGTM

PR #19833, opened 2025-01-07
CXSPA/9197: Remove unused method in major release

anjana-bl:
https://jira.tools.sap/browse/CXSPA-9197


Platonn (approved):

Platonn:
a year ago we deprecated officially (via JSDocs which are visible to customers) the original method and told customers to use the V2 method instead. Thanks to this advice, some of them them might have moved to V2, to adapt to the future breaking changes in advance.

1. So to be fair, now we should remove the original method which was deprecated and leave the replacement V2.
2. Then we should also document somewhere in the migration docs draft (which are placed rather in the `develop-next-major` but not `develop` yet) that the original method was removed and the V2 should be used instead.

feature-libs/order/core/facade/my-account-v2-order-history.service.ts
```diff
@@ -195,27 +195,7 @@ export class MyAccountV2OrderHistoryService {
...
-    );
-    return using(
-      () => loading$.subscribe(),
-      () => this.getOrderDetailsValue(code)
-    );
-  }
-
-  //TODO: CXINT-2896: Rename this method to `getOrderDetails` in next major release
-  getOrderDetailsV2(code: string): Observable<Order | undefined> {
+  getOrderDetails(code: string): Observable<Order | undefined> {
```
pawelfras, line 198: `major:` Since the `getOrderDetails` method was deprecated with the information to use `getOrderDetailsV2` instead, we should keep the one with V2 suffix to be in line with the deprecation message (and to be fair to our customers).

pawelfras:
In addition to what Kris said, let's change target branch to `develop-next-major` and together with this PR, let's add migration information under `2211_ng18`, similar to what we can find in `2211_19/2211-typescript-manual.doc.md`

PR #19835, opened 2025-01-07
fix: (CXSPA-9041) Popover directive HCT contrast

StanislavSukhanov:
* fixed a contrast of a contend displayed in popover when HCT is applied;

 * closes https://jira.tools.sap/browse/CXSPA-9041


projects/storefrontlib/shared/components/popover/popover.directive.ts
```diff
@@ -225,8 +225,15 @@ export class PopoverDirective implements OnInit {
         this.cxPopoverOptions?.autoPositioning;
 
       if (this.cxPopoverOptions?.appendToBody) {
+        const element = this.featureConfigService?.isEnabled(
+          'a11yPopoverHighContrast'
+        )
+          ? // we need to select first child element,
+            // otherwise HCT theming isn't getting picked up.
+            this.winRef.document.body.firstElementChild
```
Zeyber: Maybe we should let this fall back to body in case `firstElementChild` isn't found for whatever reason. SSR and non-ssr spartacus builds and installations can be a bit inconsistent with their DOM structures immediately after the body tag (eg. sometimes `app-root` is not added to the DOM).

Fallback example:
```this.winRef.document.body.firstElementChild ?? this.winRef.document.body```

Zeyber (changes requested):

projects/storefrontlib/shared/components/popover/popover.directive.ts
StanislavSukhanov: thanks for your comment. it's fixed

Zeyber (approved):

PR #19844, opened 2025-01-09
fix: (CXSPA-9005) fix ng-select options issue - <y of x> is conveyed twice

uroslates:
Closes https://jira.tools.sap/browse/CXSPA-9005


projects/storefrontlib/shared/components/ng-select-a11y/ng-select-a11y.directive.ts
```diff
@@ -40,7 +40,6 @@ export class NgSelectA11yDirective implements AfterViewInit {
    */
   @Input() cxNgSelectA11y: { ariaLabel?: string; ariaControls?: string };
 
-  protected translationService = inject(TranslationService);
```
StanislavSukhanov, line 43: I think we are not allowed to remove a class member as it is considered as a breaking change 

projects/storefrontlib/shared/components/ng-select-a11y/ng-select-a11y.directive.ts
```diff
@@ -145,21 +144,14 @@ export class NgSelectA11yDirective implements AfterViewInit {
...
-              const sanitizedOptionText = this.domSanitizer.sanitize(
-                SecurityContext.HTML,
-                option.innerText
-              );
-              const ariaLabel = `${sanitizedOptionText}, ${+index + 1} ${translation} ${options.length}`;
-              this.renderer.setAttribute(option, ARIA_LABEL, ariaLabel);
-            }
-          );
-        });
+      options.forEach((option: HTMLOptionElement) => {
```
StanislavSukhanov: I think we need a feature toggle here.

StanislavSukhanov (changes requested):

projects/storefrontlib/shared/components/ng-select-a11y/ng-select-a11y.directive.ts
uroslates: Thanks Stan for the heads up & catching this one!
I've reiterated over with Caine and code is updated now.

StanislavSukhanov (dismissed):

uroslates:
Changed the implementation - setting a11yNgSelectOptionsCount to false (per default) will cause ng-select options ariaLabels to working as expected. Marked all relevant directive methods as depricated. Hopefully the depricated code (including the flag) would be removed in next major release (undoing part of old directive which is now introducing issue) - making everything work normal.

projects/storefrontapp/src/app/spartacus/spartacus-features.module.ts
```diff
@@ -380,7 +380,11 @@ if (environment.cpq) {
         a11yLinkBtnsToTertiaryBtns: true,
         a11yRepeatedPageTitleFix: true,
         a11yDeliveryModeRadiogroup: true,
-        a11yNgSelectOptionsCount: true,
+        /**
+         * Defaults to false cause ng-select options ariaLabels are working as expected
+         * since Spartacus 2211.33
```
Zeyber, line 385: Lets also reference the issue CX-9005 here and in `feature-toggles.ts` so anybody curious can reference the discussion about this flag.

Zeyber (changes requested):

projects/storefrontapp/src/app/spartacus/spartacus-features.module.ts
uroslates: Updated comment with reference to the related issue.

Zeyber (approved):

PR #19848, opened 2025-01-09
feat: Migrate existing apps with application builder to use `index.server.html` in server file

pawelfras:
This PR provides an automation that migrates SPA apps using `application` builder (created with SPA >=2211.19) to use `server/index.server.html` instead of `browser/index.html` due to changes in Angular 19 and what's the output of built application.

QA (installation script):

  `npx ts-node ./tools/schematics/testing.ts in the SPA root folder`
- create new app npx @angular/cli@17 new my-app --standalone=false --style=scss --routing=false
- go to the project's directory and install latest SPA 2211.x, e.g. from RBSC
- migrate to Angualr 18, then 19 (following [migration docs](https://github.com/SAP/spartacus/blob/768edc1f4a9dbccaf78f5ca1809f05000b5bd18a/docs/migration/2211_ng19/migration.md)
- update SPA to v2211.35.0:
  `npx @angular/cli@17 update @spartacus/schematics@latest
- verify if migration runs (logs in console):  
  image
- verify `server.ts` file if changes are applied:
  image
- run `npm run build` and `NODE_TLS_REJECT_UNAUTHORIZED=0 run serve:ssr:my-app` to test if app works

Required actions:
 -[ ] change the migration from 2212.32.0 to the proper one, when it's known


kpawelczak (commented):
Tested the changes and the `server.ts` is updated after the migration. The SSR application properly uses the `index.server.html`. 
@pawelfras There are sonar errors complaining about the if statement style

kpawelczak (dismissed):

projects/schematics/src/shared/utils/package-utils_spec.ts
```diff
@@ -104,4 +107,95 @@ describe('Package utils', () => {
       expect(featureLevel).toEqual(version.substring(0, 7));
     });
   });
+
+  // New tests for SSR functions
```
Platonn: [NITPICK]
Since we commit it, those tests won't be "new" anymore, right?
```suggestion
```

projects/schematics/src/shared/utils/package-utils.ts
```diff
@@ -270,3 +332,36 @@ function getCurrentDependencyVersion(
...
+ * @throws SchematicsException if angular.json cannot be found or parsed
+ */
+function readAngularJson(tree: Tree): any {
+  const buffer = tree.read('angular.json');
+  if (!buffer) {
+    throw new SchematicsException('Could not find angular.json');
+  }
+  const angularFileBuffer = buffer.toString(UTF_8);
+  return JSON.parse(angularFileBuffer);
+}
```
Platonn: [QUESTION] 
Why didn't we reuse the existing util function `getWorkspace()`?

https://github.com/SAP/spartacus/blob/be9b4fe7b047ef6a31e922481289752f499799f8/projects/schematics/src/shared/utils/workspace-utils.ts#L51

projects/schematics/src/migrations/2211_35/ssr/update-ssr/update-server-files.ts
```diff
@@ -0,0 +1,191 @@
...
+    };
+
+    return (node: ts.Node) => ts.visitNode(node, visit);
+  };
+
+  // Apply the transformer
+  const result = ts.transform(sourceFile, [transformer]);
+  const transformedSourceFile = result.transformed[0];
+
+  const output = printer.printFile(transformedSourceFile as ts.SourceFile);
```
Platonn: [MEDIUM]
The printed file has all original new-lines removed, which might be a bit annoying for customers. Can we preserve the original new-lines? 

btw. I don't think we should use a printer directly in our schematics, but just output the modifications/`Rule` - the same like we do for example in the function `updateTokenImportPaths()` - https://github.com/SAP/spartacus/blob/be9b4fe7b047ef6a31e922481289752f499799f8/projects/schematics/src/migrations/2211_19/update-ssr/update-ssr-files.ts#L107


See the currently removed new-lines:

BEFORE MIGRATION
image

AFTER MIGRATION
image

Platonn: Proposed changes in the sub-PR:
 https://github.com/SAP/spartacus/pull/19933

projects/schematics/src/shared/utils/package-utils.ts
pawelfras: Good point! Code from `readAngularJson` was used in few functions inside this util file so I extracted it to a separate function and reused, but I wasn't aware of `getWorkspace` util. It works as expected after switching to your proposal 👌

Platonn (approved):

PR #19851, opened 2025-01-10
fix: handle OPF submit payment sequence error

Matejk00:
Closes: CXSPA-9192

### DIFF
--- a/integration-libs/opf/global-functions/core/facade/opf-global-functions.service.ts
+++ b/integration-libs/opf/global-functions/core/facade/opf-global-functions.service.ts
@@ -32,7 +32,7 @@ import {
 } from '@spartacus/opf/payment/root';
 import { LAUNCH_CALLER, LaunchDialogService } from '@spartacus/storefront';
 import { Observable, Subject, lastValueFrom } from 'rxjs';
-import { finalize, take } from 'rxjs/operators';
+import { finalize, last, take } from 'rxjs/operators';
 
 @Injectable()
 export class OpfGlobalFunctionsService implements OpfGlobalFunctionsFacade {
@@ -236,6 +236,13 @@ export class OpfGlobalFunctionsService implements OpfGlobalFunctionsFacade {
               returnPath: undefined,
             })
             .pipe(
+              /**
+               * Needed to to handle empty emissions gracefully without
+               * triggering a sequence error.
+               *
+               * It will resolve with `true` if no values are emitted.
+               */
+              last(() => true, true),
               finalize(() => {
                 if (overlayedSpinner) {
                   this.stopLoaderSpinner(overlayedSpinner);


integration-libs/opf/global-functions/core/facade/opf-global-functions.service.ts
```diff
@@ -236,6 +236,13 @@ export class OpfGlobalFunctionsService implements OpfGlobalFunctionsFacade {
               returnPath: undefined,
             })
             .pipe(
+              /**
+               * Needed to to handle empty emissions gracefully without
+               * triggering a sequence error.
+               *
+               * It will resolve with `true` if no values are emitted.
+               */
+              last(() => true, true),
```
FollowTheFlo, line 245: if i understand, in scenario where error is thrown from `this.opfPaymentFacade.submitPayment`, we return 'true' rather than catching the error?
should we create a logic that if error occurs we returns 'false'?
In GooglePayService (line 406) we have
```
  catchError(() => {
              return of(false);
            })
```
It is the logic we wrote in dociumentation
https://help.sap.com/docs/SAP_COMMERCE_CLOUD_PUBLIC_CLOUD/0996ba68e5794b8ab51db8d25d4c9f8a/1fba7ed786b44700a21435ddc07bbc98.html
```
opfPaymentFacade.submit({
...
}).subscribe(success=>{
    // true when order has been successful placed
});
}
```

Matejk00: Thanks @FollowTheFlo for the suggestion.

This situation arises from using the submit function as a global function, rather than invoking it within the context of a Spartacus.

Screenshot 2025-01-16 at 11 14 48

Key point here isn't about error handling, but rather about managing empty emissions. If `this.opfPaymentFacade.submitPayment` emits no values, we want to ensure we handle such case, otherwise we are getting an error in a browser console: https://stackoverflow.com/questions/41131476/emptyerror-no-elements-in-sequence telling that emission is empty.

The `last` operator is designed to emit the last value from an observable sequence or a default value if the sequence is empty. By providing `true` as a default value, we prevent sequence errors in the console when there are no emitted values. This approach helps maintain a consistent output and avoids any unexpected behavior.

In contrast, in the `catchError` logic we are explicitly handling potential errors that may occur during the payment submission process. Returning `of(false)` to indicate failure is appropriate for error conditions. But, in my case we are not going into the error state, so that is why I can't use such proposal.

Hope it clarifies the difference. In case of additional questions please let me know :)

FollowTheFlo: lgtm, as discussed even though there is no catch in the pipe, partner can catch it from Promise type global function.

FollowTheFlo (approved):
lgtm, see comments

PR #19859, opened 2025-01-13
fix: add resource load-once capability

FollowTheFlo:
Fix a condition logic.

### DIFF
--- a/integration-libs/opf/base/root/services/opf-resource-loader.service.ts
+++ b/integration-libs/opf/base/root/services/opf-resource-loader.service.ts
@@ -87,7 +87,10 @@ export class OpfResourceLoaderService {
     keyValueList?.forEach((keyValue: OpfKeyValueMap) => {
       attributes[keyValue.key] = keyValue.value;
     });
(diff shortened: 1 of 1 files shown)


integration-libs/opf/base/root/services/opf-resource-loader.service.ts
```diff
@@ -87,7 +87,10 @@ export class OpfResourceLoaderService {
     keyValueList?.forEach((keyValue: OpfKeyValueMap) => {
       attributes[keyValue.key] = keyValue.value;
     });
-    if (attributes?.[this.OPF_RESOURCE_LOAD_ONCE_ATTRIBUTE_KEY] === 'true') {
```
FollowTheFlo, line 90: it was the opposite logic, we need:
if OPF_RESOURCE_LOAD_ONCE_ATTRIBUTE_KEY  does not have 'true value' -> we add OPF_RESOURCE_ATTRIBUTE_KEY attribute

Matejk00 (approved):

PR #19870, opened 2025-01-15
fix: add missing role attribute

sdrozdsap:


feature-libs/pickup-in-store/components/presentational/pickup-options/pickup-options.component.html
```diff
@@ -178,7 +178,12 @@
...
-      <span *ngIf="validationError" tabindex="0" class="cx-invalid-message">
+      <span
+        *ngIf="validationError"
+        role="alert"
+        tabindex="0"
+        class="cx-invalid-message"
+      >
         {{ validationError | cxTranslate }}
       </span>
     </div>
```
Pio-Bar, line 189: We can remove the `aria-live="assertive" aria-atomic="true"` since role alert provides the same functionality.

Pio-Bar (dismissed):

feature-libs/pickup-in-store/components/presentational/pickup-options/pickup-options.component.html
sdrozdsap: Since *ngIf and role="alert" doesn't really fit to each other - if we create an element with role alert it is not announced, only when it content changes. 

So for that container div has  `<div aria-live="assertive" aria-atomic="true">`. And role alert on the span is just to suppress AMP warnings

sdrozdsap: Do you know if there is better way that would work?

petarmarkov9449 (approved):

Zeyber (approved):

PR #19871, opened 2025-01-15
fix: (CXSPA-8994) - Prevent forced colors from interfering with contrast themes

Pio-Bar:
Closes: [CXSPA-8994](https://jira.tools.sap/browse/CXSPA-8994)

The Spartacus custom high contrast themes should not be affected by external solutions. These contrast themes are tailored to provide good accessibility without sacrificing readability and usability of the app. The same cannot be guaranteed with one-size-fits-all solutions like the Windows/MacOS contrast themes.

Therefore, we should not make changes to accommodate external stylings that users might apply. These can vary greatly and it would be impossible to satisfy all. However, we allow the default theme to be modified so users can still use other solutions if preferred.


StanislavSukhanov (approved):

Zeyber (approved):
Thanks for the arguments you have left for this one. I think this is a justified approach.

PR #19873, opened 2025-01-16
fix: (CXSPA-9089) - Fix valid ARIA markup for TabComponent

petarmarkov9449:
Closes: [CXSPA-9089](https://jira.tools.sap/browse/CXSPA-9089)

Fix valid ARIA markup for TabComponent.

Affects: TabComponenet, TabPanelComponent


projects/storefrontlib/cms-components/content/tab/tab.component.spec.ts
Pio-Bar, line 1: We should not just remove an assertion to make the test pass. If we have added extra logic we should test it instead. I.e. ensure the aria-controls value is not set when the tab is closed.

Pio-Bar (changes requested):

Pio-Bar (approved):

Zeyber (approved):

PR #19875, opened 2025-01-16
fix: (CXSPA-9075) fix footer region group labels arialLabel issue.

uroslates:
Closes: https://jira.tools.sap/browse/CXSPA-9075

### DIFF
--- a/projects/storefrontlib/cms-components/navigation/navigation/navigation-ui.component.html
+++ b/projects/storefrontlib/cms-components/navigation/navigation/navigation-ui.component.html
@@ -153,6 +153,8 @@
         [attr.depth]="getTotalDepth(node)"
         [attr.wrap-after]="node.children.length > wrapAfter ? wrapAfter : null"
         [attr.columns]="getColumnCount(node.children.length)"
+        role="group"
+        [attr.aria-label]="node.title"
       >
         <ng-container *ngFor="let child of node.children; index as i">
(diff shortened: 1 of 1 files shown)


projects/storefrontlib/cms-components/navigation/navigation/navigation-ui.component.html
```diff
@@ -153,6 +153,8 @@
         [attr.depth]="getTotalDepth(node)"
         [attr.wrap-after]="node.children.length > wrapAfter ? wrapAfter : null"
         [attr.columns]="getColumnCount(node.children.length)"
+        role="group"
```
Pio-Bar: We should keep the role unchanged. This is a list with listitems inside, I am not sure why we would be grouping them all together instead. The aria-label will still be read and this way we are keeping the semantic role. This would affect all implementations of the navigation-ui not just the footer, we need to keep this in mind while testing.

Pio-Bar (changes requested):

projects/storefrontlib/cms-components/navigation/navigation/navigation-ui.component.html
uroslates: Addressed!

Pio-Bar (approved):

Zeyber (approved):

PR #19877, opened 2025-01-16
fix: Move GooglePay url into default config

Matejk00:
Closes: CXSPA-9216


FollowTheFlo (approved):
lgtm, we'll then have to add exception on Checkmrax with [CXSPA-9216](https://jira.tools.sap/browse/CXSPA-9216) as ref

PR #19946, opened 2025-01-28
fix: (CXSPA-9040) Pick Up in store. Product availability Icon

StanislavSukhanov:
* add aria-hidden attribute

* closes https://jira.tools.sap/browse/CXSPA-9040


feature-libs/pickup-in-store/components/presentational/store/store.component.html
```diff
@@ -42,6 +42,7 @@
     <div>
       <cx-icon
         class="cx-store-stock-icon"
+        aria-hidden="true"
```
uroslates, line 45: This looks good. I"m not sure if the expectation was to remove it or add text but I agree with U that removing is better for these cases/icons! 
But maybe would be good to dbl check W the team.

uroslates (approved):

Zeyber (approved):

PR #19952, opened 2025-01-30
CXCDS-14100: move SCI configuration from .env to schematics-config

johannathalmann-SAP:
The value of "sciEnabled" should not be defined in .env_cdmrc file, as that's only available for developers. Instead, it needs to be configured when setting up cds from libs.


klucerofermin (dismissed):

integration-libs/cds/src/schematics/add-cds/schema.json
```diff
@@ -47,6 +47,12 @@
       "type": "string",
       "description": "Configuration URL for ProfileTag. (e.g. https://tag.static.stage.context.cloud.sap/config/my-config123)",
       "x-prompt": "[CDS] What is the ProfileTag configuration URL?"
+    },
+    "sciEnabled": {
+      "type": "boolean",
+      "description": "CDS/ISS is integrated into the SAP Cloud Identity Service (SCI). The downstream services use different domains and URL formats. To enable SCI, enable this parameter and make sure to use the proper profile tag urls (see documentation).",
+      "default": false,
```
Platonn, line 54: [Business question]
I understand we don't want to enable SCI in existing applications, but rather let customers enable it by themselves

But do we really want to disable by default SCI in fresh new applications created by customers? My gut feeling is we'd like new apps to use newest features OOTB. But you know better the business context here, so I'm happy to learn your poiunt of view.

projects/schematics/src/shared/lib-configs/integration-libs/cds-schematics-config.ts
```diff
@@ -97,6 +107,7 @@ function buildCdsConfig(
                   'PROFILE_TAG_CONFIG_URL_PLACEHOLDER'
                 }',
               allowInsecureCookies: true,
+              sciEnabled: ${options.sciEnabled}
```
Platonn: [QUESTION]
I can see here we set the property of the Spartacus global config  `config.cds.profileTag.sciEnabled` based on the argument option `--sci-enabled` of the Installation Schematics, during the installation.

1. But I'm having problem to understand what do we need this Spartacus config option for (`config.cds.profileTag.sciEnabled`)?
I have problem to find any usages of it in runtime. Below is a screenshot from VSCode on the current branch:
image

(See: https://github.com/SAP/spartacus/blob/c5d54c595b63260a38a4d6e883cc1da8d404b0f8/integration-libs/cds/src/config/profile-tag.config.ts#L12 )

2. By the way, I can see we've removed an analogical Spartacus config property in this PR in 1 level above  `config.cds.sciEnabled` (without `.profileTag.` in beteween). So it seems we had previously the config property `sciEnabled` defined in interfaces on both levels:
- `config.cds.sciEnabled` - which is removed in this PR
- `config.cds.profileTag.sciEnabled` - which remained in this PR
Perhaps we wanted to remove also the `profileTag`-nested one?

integration-libs/cds/src/schematics/add-cds/schema.json
johannathalmann-SAP: SCI support for ISS will be activated a little bit later - we’re still in extensive test phase for that (that’s also why we need to be able to activate it already for our test installations) 
But you are right, after SCI is fully integrated into ISS, we could think about making it the default for new setups.

projects/schematics/src/shared/lib-configs/integration-libs/cds-schematics-config.ts
johannathalmann-SAP: We want to have it consistent with our profileTag implementation - https://github.tools.sap/cx-commerce/cds-argonauts-profiletag/blob/master/src/config/config.ts

integration-libs/cds/src/schematics/add-cds/schema.json
Platonn: Thank you. Now I get better the business context.

Do I understand correctly that eventually when we like customers to enable the SCI feature, THEY will have to:
- manually change the endpoints configuration `config.cds.endpoints` in their `CdsFeatureModule` in their apps? 
- they can keep the config `config.cds.profileTag.sciEnabled` false in their  `CdsFeatureModule` in their apps, becasue it is (and will be?) a dead config

btw. as of today new apps crated with schematics will put the config `config.cds.profileTag.sciEnabled` explicitly in customers' apps' `CdsFeatureModule` and [set it there to `false](https://github.com/SAP/spartacus/pull/19952/files#diff-924178d81edea4b6d2638cbf2f6783c4a120785a01d00a833ff3dd21c53882e9R110)`. But it seems to be a **dead config** (as noted in another comment https://github.com/SAP/spartacus/pull/19952/files#r1940901467) - which I believe can be misleading for customers. Changing this config will have no effect for them, right? So I still struggle to understand why we want to expose this config. Perhaps I'm missing something obvious important - then please help me understand.

johannathalmann-SAP: Customers who enable SCI, during their setup, will have to:
- set the sciEnabled = true
- set the correct profiletag urls 

Which strategy endpoint to use, is then decided on the value of sciEnabled in the cds-schematics-config.ts (line 74-76)
Also config.cds.profileTag.sciEnabled will be same value as in the setup (line 110)


Or is your question regarding changing these things MANUALLY in CdsFeatureModule only refering to existing setups/customers?

Platonn: Thank you for your answer.
So I understand that during _initial installation_ of spartacus, the shape of the scaffolded `CdsFeatureModule` in customer's app will look differently depending on the value of sciEnabled in the cds-schematics-config.ts (line 74-76).

And I understand that if customer already created an application with CDS in the past, then the only way enable SCI in the future (if they want it in the future), for them will be they'll manually change the content of their CdsFeatureModule -> then we should put instructions in our official docs for them so they know what changes to peform manually there.

Still, what I dont understand is: why do we keep Spartacus config property `config.cds.profileTag.sciEnabled`. Perhaps we have some future plans for it?
Currently, as far as I can see in the code, it's _not used_ anywhere. It's just _defined_.
Perhaps I'm missing something here, then please correct me.
If this config is not used anywhere, my gut feeling is we should remove it and re-introduce only possibly in the future, if in the future we add some logic in Spartacus that would _use_ this config.

johannathalmann-SAP: Hello, both assumptions for _initial installation_ and _already created application_ are correct. We already gave our input to our documentation team to have that properly written down.

I've deleted the `config.cds.profileTag.sciEnabled` property.

Platonn (approved):

PR #19955, opened 2025-01-30
feat: update jest setup files to remove warning (CXSPA-9288)

kimhw0630:
Closes: https://jira.tools.sap/browse/CXSPA-9288

Notes:
```
import 'jest-preset-angular/setup-jest';
import 'zone.js';
```
The above two lines have been replaced in all instances of setup-jest.ts as below

```
import { setupZoneTestEnv } from 'jest-preset-angular/setup-env/zone';
setupZoneTestEnv();
```

### DIFF
--- a/feature-libs/asm/setup-jest.ts
+++ b/feature-libs/asm/setup-jest.ts
@@ -4,5 +4,6 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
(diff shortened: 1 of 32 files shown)


pawelfras (approved):
LGTM.
Looks like also `import 'zone.js'` are no longer required for these particular test configs.
Let's add info to the PR descriptions that such imports were also removed.

PR #19961, opened 2025-02-04
feat: CXSPA-8113 ESLint rule ensuring 'error' property is initialized for classes inplementing 'ErrorAction'

pawelfras:
this PR contains a new ESLint rule that ensures that NgRx Failure Action which implements `ErrorAction` interface must initialize the `error` property. Due to usage of TS type checker, the rule is able to check initialization across the inheritance tree of the target class.
image

_Note: This ESLint rule doesn't contain autofixer as the fix varies depending on the scenario._

closes [CXSPA-8113](https://jira.tools.sap/browse/CXSPA-8113)

---
QA:
- open any file with actions, e.g. `product-search.action.ts`
- add the following code and verify that there is no error
  ```ts
  export class MyActionFail implements ErrorAction {
    readonly type = SEARCH_PRODUCTS_FAIL;
    public error: any;
  
    constructor(public payload: any) {
      this.error = payload;
    }
  }
  ```
- modify the code in the following way and verify that `@nx/workspace/ngrx-fail-action-must-initialize-error` error occurs
  ```ts
  export class MyActionFail implements ErrorAction {
    readonly type = SEARCH_PRODUCTS_FAIL;
    public error: any;
  
    // constructor(public payload: any) {
    // this.error = payload;
    // }
  }
  ``` 
- modify the code in the following way and verify that `@nx/workspace/ngrx-fail-action-must-initialize-error` error occurs
    ```ts
    export class MyGrandparentActionFail implements ErrorAction {
        readonly type = SEARCH_PRODUCTS_FAIL;
        public error: any;
    }
     
    export class MyParentActionFail extends MyGrandparentActionFail implements ErrorAction {
      readonly type = SEARCH_PRODUCTS_FAIL;
      public error: any;
    }
    
    export class MyActionFail extends MyParentActionFail implements ErrorAction {
      readonly type = SEARCH_PRODUCTS_FAIL;
      public error: any;
  
      // constructor(public payload: any) {
      // this.error = payload;
      // }
    }
    ```
 - modify the code in the following way and verify that `@nx/workspace/ngrx-fail-action-must-initialize-error` is no longer present on `MyActionFail` class
    ```ts
    export class MyGrandparentActionFail implements ErrorAction {
        readonly type = SEARCH_PRODUCTS_FAIL;
        public error: any;
        constructor() {
          this.error = 'error';
        }
     }
     
    export class MyParentActionFail extends MyGrandparentActionFail implements ErrorAction {
      readonly type = SEARCH_PRODUCTS_FAIL;
      public error: any;
    }
    
    export class MyActionFail extends MyParentActionFail implements ErrorAction {
      readonly type = SEARCH_PRODUCTS_FAIL;
      public error: any;
  
      // constructor(public payload: any) {
      // this.error = payload;
      // }
    }
    ```
  - modify the code in the following way and verify that `@nx/workspace/ngrx-fail-action-must-initialize-error` is no longer present on `MyActionFail` class
    ```ts
    export class MyGrandparentActionFail implements ErrorAction {
        readonly type = SEARCH_PRODUCTS_FAIL;
        public error: any;
     }
     
    export class MyParentActionFail extends MyGrandparentActionFail implements ErrorAction {
      readonly type = SEARCH_PRODUCTS_FAIL;
      public error: any;
      constructor() {
        super();
        this.error = 'error';
      }
    }
    
    export class MyActionFail extends MyParentActionFail implements ErrorAction {
      readonly type = SEARCH_PRODUCTS_FAIL;
      public error: any;
  
      // constructor(public payload: any) {
      // this.error = payload;
      // }
    }
    ```
  - modify the code in the following way and verify that `@nx/workspace/ngrx-fail-action-must-initialize-error` is no longer present:
    ```ts
    export class MyActionFail implements ErrorAction {
      readonly type = SEARCH_PRODUCTS_FAIL;
      constructor(public payload: any, public error: Object) {}
    }
    ```


Platonn (dismissed):

tools/eslint-rules/rules/utils/typescript-ast-utils.ts
```diff
@@ -0,0 +1,254 @@
...
+        return true;
+      }
+    }
+  }
+
+  // Check base class recursively
+  const baseTypes = type.getBaseTypes();
+  if (!baseTypes) {
+    return false;
+  }
```
Platonn: I'm having problem to understand why this method is so complex. Wouldn't this util suffice: [`hasImplementsInterface](https://github.com/SAP/spartacus/blob/9278b1092e4068500979316329de3bf5ef29423e/tools/eslint-rules/rules/utils/implements-interface-utils.ts#L14-L26)?

If not, then I'd love if you could help me understand what this method does:
1. `// Check if the type directly implements interface through properties` - I'm having problem to understand what it checks.
2. `// Check if any interface in the heritage chain matches` - although it seems to do its job well (i.e. check if the class name is followed by `implements Xxxx` via [heritageClauses](https://january.sh/posts/gentle-introduction-to-typescript-compiler-api#heritageclauses) property), the //code comment seems to be misleading to me with the word "chain". We don't check the inheritance _chain_ here, as far as I can understand. Moreover I'm wondering if it can be simply replaced by the simple util mentioned above `hasImplementsInterface`.
3. Do we _really_ need to check whether some of the ancestor classes implements this interface? (do we really need it for our usecase). Why asking: I'd like to keep the code as simple as possible, yet matching our needs

tools/eslint-rules/rules/utils/typescript-ast-utils.ts
```diff
@@ -0,0 +1,254 @@
...
+  if (!symbol?.declarations?.[0]) {
+    // Check parent class if no declarations in current class
+    const baseTypes = classType.getBaseTypes();
+    if (!baseTypes) {
+      return false;
+    }
+    return baseTypes.some((baseType) =>
+      isPropertyInitialized(baseType, propertyName)
+    );
+  }
```
Platonn: What doest it mean that a class symbol doesn't have a "declaration"? Can it ever happen in our real case? 
...And why in such case we should check the base type?

btw. I think that instead of getting Symbol from Type (which I understood that we got from Symbol [there](https://github.com/SAP/spartacus/blob/9278b1092e4068500979316329de3bf5ef29423e/tools/eslint-rules/rules/ngrx-fail-action-must-initialize-error.ts#L80)), maybe we could pass the original symbol (`node`) as a param to this function?

PS. if this logic was about to check ancestor classes as a fallback (when didn't find property initialization in this child class), then I understand we're already doing it in line 245.

tools/eslint-rules/rules/utils/typescript-ast-utils.ts
```diff
@@ -0,0 +1,254 @@
...
+  if (
+    declaration.members.some((member) =>
+      hasPropertyInitialization(member, propertyName)
+    )
+  ) {
+    return true;
+  }
+
+  // Check all constructors, including deprecated ones
+  const constructors = declaration.members.filter(ts.isConstructorDeclaration);
```
Platonn: If I understand correctly, we're talking here about such an example:
```ts
constructor(a:number);
constructor(a:number, b:number);

constructor(public a:number, private b:number) {
  ...
}
```
...where first 2 guys are just constructor delcarations, but the 3rd guy is the contructor implementation. And AFAIK only the constructor implementation can have `public/protected/private` modifiers. Then why do we need to check other constructor declarations? Do we _really_ need to do it?

tools/eslint-rules/rules/ngrx-fail-action-must-initialize-error.ts
```diff
@@ -0,0 +1,97 @@
...
+  meta: {
+    type: 'problem',
+    docs: {
+      description:
+        'Ensures declared `error` property in NgRx Failure Actions is initialized',
+    },
+    schema: [],
+    messages: {
+      missingErrorInitialization:
+        '[Spartacus] NgRx Failure Action that implements `ErrorAction` interface must initialize the `error` property. You can do this in the constructor or as a property initializer',
```
Platonn: The instructions of how to fix the problem might be a bit vague for developers, let's give them example.
```suggestion
        '[Spartacus] NgRx Failure Action that implements `ErrorAction` interface must initialize the `error` property. You can do this in the constructor (e.g. `constructor(public error: Object) {}`) or as a property initializer (e.g. `this.error = <some-value>`)',
```

tools/eslint-rules/rules/ngrx-fail-action-must-initialize-error.spec.ts
```diff
@@ -0,0 +1,239 @@
...
+        export interface ErrorAction {
+          error: Object;
+        }
+
+        export class SearchProductsFail implements ErrorAction {
+            readonly type = SEARCH_PRODUCTS_FAIL;
+            error = 'some error'; // Property initializer
+        }
+      `,
+      filename: 'file.ts',
```
Platonn: nitpick: this can be moved to a separate variable

tools/eslint-rules/rules/ngrx-fail-action-must-initialize-error.spec.ts
```diff
@@ -0,0 +1,239 @@
...
+        export interface ErrorAction { error: Object; }
+        export class BaseFailAction implements ErrorAction {
+          error = new Error();
+        }
+        export class DerivedFail extends BaseFailAction {
+          readonly type = DERIVED_FAIL;
+        }
+      `,
+      filename: 'file.ts',
+    },
```
Platonn, line 140: For being more bullet-proof, let's add also a case with a grandparent (to assert it's not only a shallow parent-check but likely a recursive check)
```suggestion
    {
      name: 'Inheritance case',
      code: `
        export interface ErrorAction { error: Object; }
        export class BaseFailAction implements ErrorAction {
          error = new Error();
        }
        export class DerivedFail extends BaseFailAction {
          readonly type = DERIVED_FAIL;
        }
      `,
      filename: 'file.ts',
    },
    {
      name: 'Grandparent inheritance case',
      code: `
        export interface ErrorAction { error: Object; }
        export class GrandparentFailAction implements ErrorAction {
          error = new Error();
        }
        export class ParentFailAction extends GrandparentFailAction {
          readonly type = PARENT_FAIL;
        }
        export class ChildFailAction extends ParentFailAction {
          readonly type = CHILD_FAIL;
        }
      `,
      filename: 'file.ts',
    },
```

Platonn, line 216: For being more bullet-proof, let's add also a case with a grandparent (to assert it's not only a shallow parent-check but likely a recursive check)

```suggestion
    },
    {
      name: 'Grandparent inheritance case',
      code: `
        export interface ErrorAction { error: Object; }
        export class GrandparentFailAction implements ErrorAction {
          error = new Error();
        }
        export class ParentFailAction extends GrandparentFailAction {
          readonly type = PARENT_FAIL;
        }
        export class ChildFailAction extends ParentFailAction {
          readonly type = CHILD_FAIL;
        }
      `,
      filename: 'file.ts',
    },
    {
      name: 'Grandparent inheritance case - with constructor param',
      code: `
        export interface ErrorAction { error: Object; }
        export class GrandparentFailAction implements ErrorAction {
          constructor(public: error: Object);
        }
        export class ParentFailAction extends GrandparentFailAction {
          readonly type = PARENT_FAIL;
        }
        export class ChildFailAction extends ParentFailAction {
          readonly type = CHILD_FAIL;
        }
      `,
      filename: 'file.ts',
    },
```

tools/eslint-rules/rules/utils/typescript-ast-utils.ts
pawelfras: Thanks Kris, `hasImplementsInterface` do the job! I was not aware we have such an util in our repository.

tools/eslint-rules/rules/utils/typescript-ast-utils.ts
pawelfras: Answering the first question, based on my knowledge it can happen in the real case: Syntetic types, union types, anonymous classes, classes from declaration files and. But indeed, we don;t need to rely on that in our case, but we just need to satisfy typing.

We could pass node and adjust the function to do all checks based on it, but to check ancestors, we would have to involve symbol either way.

That said, I'd still use Symbol, but i simplified the function's logic

tools/eslint-rules/rules/utils/typescript-ast-utils.ts
pawelfras: Good point. I didn't find any util that checks whether it's constructor's implementation, but I introduced another filter that looks for access modifiers inside the constructor's params.

pawelfras: Second thoughts confirmed by linter - as access modifiers are optional in TS, we need to iterate over all constructors unless we find better way to verify the constructor implementation. I'm reverting back this change.

tools/eslint-rules/rules/utils/typescript-ast-utils.ts
Platonn: Super readable now. Many thanks!

Platonn (approved):

PR #19964, opened 2025-02-04
chore: prevent translation changes being merged without the correct configurations

RadhepS:
closes: https://jira.tools.sap/browse/CXSPA-9258


ci-scripts/validate-translations.sh
```diff
@@ -35,14 +32,35 @@ check_import_statements() {
...
+  for folder in $translation_folders; do
+    check_export_statements "$folder" "$translations_file"
+    check_import_statements "$folder"
+  done
+}
+
+# Check if folder_path is provided as an argument
+if [ -z "$1" ]; then
+  # Find all translation folders in the repository
+  translation_paths=$(find . -type d -path '*/assets/translations')
```
Platonn: While running the script on local (where having `/dist` folder with already built libs), I've got a failure of the script:
`Translations file not found: ./dist/organization/account-summary/assets/translations/translations.ts`

```suggestion
  # Find all translation folders in the repository, excluding node_modules/ and dist/
  translation_paths=$(find . -type d -path '*/assets/translations' -not -path './node_modules/*' -not -path './dist/*')
``

File: .github/workflows/validate-translations.yml [hunk 2]
```diff
@@ -0,0 +1,44 @@
...
+          node-version: ${{ env.NODE_VERSION }}
+
+      - name: Run Translation Validation Script
+        run: |
+          BASE_COMMIT=${{ github.event.pull_request.base.sha }}
+          HEAD_COMMIT=${{ github.event.pull_request.head.sha }}
+
+          # Get the paths of the changed files
+          changed_files=$(git diff --name-only $BASE_COMMIT $HEAD_COMMIT | grep 'assets/translations')
+
```
Platonn: Here we account only for the pattern `assets/translations`, but it can be also `assets/src/translations` (with `src` in the middle).

When I removed an expected `export` statement or `import` statement in files in the folder `projects/assets/src/translations/` - the script didn't complain, but it should.


We need take it into account:
- in line 32
- in line 37
- in file  `ci-scripts/validate-translations.sh` in line 56

ci-scripts/validate-translations.sh
```diff
@@ -0,0 +1,69 @@
...
+
+  # Ensure the translations file exists
+  [ -f "$translations_file" ] || { echo "Translations file not found: $translations_file"; exit 1; }
+
+  # Get the list of translation folders
+  translation_folders=$(find "$folder_path" -mindepth 1 -maxdepth 1 -type d -exec basename {} \;)
+
+  # Check if each folder and index file is listed in the translations file
+  for folder in $translation_folders; do
+    if echo "$folder" | grep -q -e "dist" -e "node_modules"; then
```
Platonn, line 48: Currently it matches any word containing `dist`, e.g `distribution`, so in edge case it would be false positive when we have a library named e.g.
feature-libs/products-**dist**ribution/translations/

```suggestion
    if echo "$folder" | grep -q -e "^dist/" -e "^node_modules/"; then
```

ci-scripts/validate-translations.sh
```diff
@@ -0,0 +1,69 @@
...
+      continue
+    fi
```
Platonn: 1. Unfortunately, the script still fails, when we're having some libs built in the `/dist` folder, e.g. with the error:
   ```
   Translations file not found: ./dist/organization/account-summary/assets/translations/translations.ts
   ```
   I believe we should validate path against being `dist` or `node_modules` earlier, in the beginning of the function.

1. Currently the pattern is matched for any word containing `dist`, e.g `distribution`. To avoid edge case problems, we can specify a pattern `/dist/`. Example edge case below:
feature-libs/products-**dist**ribution/translations/

I'm attaching a proposal of fixing above 2 points:

```suggestion
process_translation_folder() {
  local folder_path=$1
  local translations_file="$folder_path/translations.ts"
  
  if echo "$folder" | grep -q -e "/dist/" -e "/node_modules/"; then
    return
  fi

  # Ensure the translations file exists
  [ -f "$translations_file" ] || { echo "Translations file not found: $translations_file"; exit 1; }

  # Get the list of translation folders
  translation_folders=$(find "$folder_path" -mindepth 1 -maxdepth 1 -type d -exec basename {} \;)

  # Check if each folder and index file is listed in the translations file
  for folder in $translation_folders; do
```

Platonn (approved):
✅ LGTM
Tested both locally and on an [example PR](https://github.com/SAP/spartacus/pull/20021).
Tested both feature libs and assets lib (which has /src/ subfolder)

PR #19965, opened 2025-02-05
fix: (CXSPA-9266) prevent animation playing when scene loaded

sam-garland:
It should only play when the user clicks the play button.

Fixed by updating from UI5 1.108 to 1.120

JIRA: [CXSPA-9266](https://jira.tools.sap/browse/CXSPA-9266)


Platonn (approved):

Platonn (approved):
Hi Sam.

I cannot test this PR without having an access to EPD-visualisation backend. So I assume you tested it and I approve it as a favor (because currently it's technically needed due to current .CODEOWNERS file).

The changes in schematics should require an approve from a core team member. They look good - we change only how new libraries are created with the installation schematics, no changes to existing customers.

Perhaps you might want to announce via SAP Note (or some other channel?) to customers that we recommend them to update their config in their `epd-visualization-feature.module.ts` in their apps - so to use  newest 1.120 version of UI5. In this case, I guess you might want to contact Matthew Burton, our doc writer.

PR #19966, opened 2025-02-05
[FIX] CXSPA-9334 Terms and condition URL not changed accordingly when…

andreighervan7:
Ticket [CXSPA-9334](https://jira.tools.sap/browse/CXSPA-9334)


rmch91:
Hello @andreighervan7. 
Thank you for preparing this PR.

I'm not sure if you are familiar with the commit guideline and planned to change the PR title to the commit message that complies with our guidelines. If not, please take a look here https://sap.github.io/spartacus-docs/commit-guidelines/

feature-libs/checkout/base/components/checkout-place-order/checkout-place-order.component.html
```diff
@@ -16,8 +16,8 @@
       />
       <span class="form-check-label">
         {{ 'checkoutReview.confirmThatRead' | cxTranslate }}
-        <a
-          [routerLink]="{ cxRoute: 'termsAndConditions' } | cxUrl"
+         <a
+          [attr.href]="termsAndConditionUrl$ | async"
```
rmch91: Could you please describe, why do we need this change from angular mechanism to standard href?
Did you think about any potential complications for the customers, if this navigation will be outside of angular router?
Just from top of my head: User subscribed to `router.events.subscribe` and expect to intercept exactly this navigation event. Will it be registered inside of this stream after this change? I have a gut feeling that the answer is NO (TBC). In this case for me it is a breaking change.

feature-libs/checkout/base/components/checkout-place-order/checkout-place-order.component.ts
```diff
@@ -9,26 +9,36 @@ import {
...
 @Component({
   selector: 'cx-place-order',
   templateUrl: './checkout-place-order.component.html',
   changeDetection: ChangeDetectionStrategy.OnPush,
   standalone: false,
 })
-export class CheckoutPlaceOrderComponent implements OnDestroy {
+export class CheckoutPlaceOrderComponent implements OnDestroy, OnInit {
   placedOrder: void | Observable<ComponentRef<any> | undefined>;
+  currency: string;
```
rmch91: `major:` this property is not used anywhere

feature-libs/checkout/base/components/checkout-place-order/checkout-place-order.component.ts
```diff
@@ -9,26 +9,36 @@ import {
...
+  termsAndConditionUrl$ = new BehaviorSubject<string>('');
+  private currency$ = this.currencyService.getActive();
```
rmch91: `major:` let's make this field `protected` for easier extensibility and customisation

After reading the next lines, and checked where those properties used, I don't think we need to expose it to class properties, since they are not used anywhere outside of `setTermsOfConditionUrl()` method. Let's encapsulate it inside of `setTermsOfConditionUrl()` with `combineLatest([this.currencyService.getActive(), this.languageService.getActive()])`

feature-libs/checkout/base/components/checkout-place-order/checkout-place-order.component.ts
```diff
@@ -9,26 +9,36 @@ import {
...
+  private language$ = this.languageService.getActive();
```
rmch91: `major:` let's make this field `protected` for easier extensibility and customisation

After reading the next lines, and checked where those properties used, I don't think we need to expose it to class properties, since they are not used anywhere outside of `setTermsOfConditionUrl()` method. Let's encapsulate it inside of `setTermsOfConditionUrl()` with `combineLatest([this.currencyService.getActive(), this.languageService.getActive()])`

feature-libs/checkout/base/components/checkout-place-order/checkout-place-order.component.ts
```diff
@@ -43,9 +53,16 @@ export class CheckoutPlaceOrderComponent implements OnDestroy {
     protected routingService: RoutingService,
     protected fb: UntypedFormBuilder,
     protected launchDialogService: LaunchDialogService,
-    protected vcr: ViewContainerRef
+    protected vcr: ViewContainerRef,
+    protected route: Router,
```
rmch91: `major:` For DI, let's use the `inject` function instead of a constructor to avoid introducing unnecessary breaking changes. See: https://wiki.one.int.sap/wiki/display/spar/Coding+Guidelines#CodingGuidelines-Useinject()functioninsteadofconstructorforDependencyInjection

Changes in the constructor are breaking for customers. Here you can check some list of potential breaking changes - https://wiki.one.int.sap/wiki/display/spar/Maintaining+Public+APIs

feature-libs/checkout/base/components/checkout-place-order/checkout-place-order.component.ts
```diff
@@ -78,6 +95,16 @@ export class CheckoutPlaceOrderComponent implements OnDestroy {
     this.routingService.go({ cxRoute: 'orderConfirmation' });
   }
 
+  setTermsOfConditionUrl(): void {
+    combineLatest([this.currency$, this.language$]).subscribe(
```
rmch91: `major:` We need to unsubscribe here to avoid memory leaks

feature-libs/checkout/base/components/checkout-place-order/checkout-place-order.component.ts
```diff
@@ -78,6 +95,16 @@ export class CheckoutPlaceOrderComponent implements OnDestroy {
...
+      ([currency, language]) => {
+        const segments = this.route.url.split('/').filter((s) => !!s);
+        const url = `/${segments[0]}/${language}/${currency}/terms-and-conditions`;
+        this.termsAndConditionUrl$.next(url);
```
rmch91: what do you think about declaring observable instead of BehaviourSubject
`termsAndConditionUrl$: Observable<string>;`
And inside of this method just assign it directly 
```typescript
  setTermsOfConditionUrl(): void {
    this.termsAndConditionUrl$ = combineLatest([
      this.currencyService.getActive(),
      this.languageService.getActive(),
    ]).pipe(
      switchMap(([currency, language]) => {
        const segments = this.route.url.split('/').filter((s) => !!s);
        const url = `/${segments[0]}/${language}/${currency}/terms-and-conditions`;
  
        return url;
      })
    );
  }
``` 

So the `async` pipe will take over the unsubscribing

feature-libs/checkout/base/components/checkout-progress/checkout-progress.component.html
```diff
@@ -10,7 +10,8 @@
           [class.active]="isActive(i)"
           [class.disabled]="isDisabled(i)"
         >
-          <a
+
```
rmch91: Could you please analyze the rest of the code while considering the comments I’ve already provided? I’ve noticed that the remaining code is quite similar and might have similar issues.

rmch91 (changes requested):
Good job @andreighervan7. Let's improve your solution a little bit.

feature-libs/checkout/base/components/checkout-place-order/checkout-place-order.component.html
andreighervan7: i will try another solution with [routerLink]

andreighervan7: i changed now with [routerLink]

feature-libs/checkout/base/components/checkout-place-order/checkout-place-order.component.ts
```diff
@@ -27,13 +33,17 @@ import { Observable } from 'rxjs';
   changeDetection: ChangeDetectionStrategy.OnPush,
   standalone: false,
 })
-export class CheckoutPlaceOrderComponent implements OnDestroy {
+export class CheckoutPlaceOrderComponent implements OnDestroy, OnInit {
   placedOrder: void | Observable<ComponentRef<any> | undefined>;
-
+  currency$ = new Observable<string>();
```
rmch91: `nitpick`: Do we need separate observables, and in the result two subscription? 
Maybe we could have just one observable `params$` and assign value mapped based on this.currencyService.getActive() and this.languageService.getActive() ?

rmch91: Also, you should add new services or mocks to test files. The pipeline is failing

rmch91 (changes requested):

rmch91 (approved):
LGTM

PR #19974, opened 2025-02-07
fix: (CXSPA-9382) improve phone layout

sam-garland:
The list item layout was bad.
The search icon was mispositioned.
The floating toolbar in the viewport was too big. 

image

Most issues were caused by this rule matching too broadly and modifying various divs that should not have had their padding modified:
image

Now looks like this:
image

The spare parts tab layout also needed some tweaks. I added some padding to improve the appearance now that the tabs style has changed.

JIRA: https://jira.tools.sap/browse/CXSPA-9382, https://jira.tools.sap/browse/CXSPA-9265


Platonn (approved):
Hi Sam.
Thank you for providing the fix PR.
Changes in this PR are only related to the EPD lib, no core/shared files, so theoretically no central approve from the core team should be needed. 
As a followup I suggest we should create an an entry in `.CODEOWNERS` that you and your team should be able to self-review and approve PRs that are within the epd-visualistaions library folder.

I cannot test this PR without having an access to EPD-visualisation backend. So I assume you tested it and I approve it as a favor (because currently it's technically needed due to current `.CODEOWNERS` file

PR #20004, opened 2025-02-18
feat: (SSR) define cacheSize in memory units

kpawelczak:
closes: [CXSPA-8964](https://jira.tools.sap/browse/CXSPA-8964)

- Added new mechanism to limit cache usage in bytes: Introduced a feature to manage the cache size by limiting it in bytes, rather than by entries.
     - should handle multiple removals if needed (until new render will not exceed limit)
     - should remove oldest entry when cache size exceeds limit

- Added feature flag: The `cacheSizeInBytes` feature flag is now available in the `ssrFeatureToggles` section of the configuration. When enabled, the cache will be controlled in bytes.

- Deprecated `cacheSize`: The previous cacheSize mechanism is deprecated. This change will be available in the current version, but the exact version number for deprecation is yet to be determined (pending review).
- default value for `cacheLimit` is 3 GB

Summary:
Added logic to check if `cacheSizeInBytes` is enabled via `ssrFeatureToggles`. If true, the cache size is managed in bytes.
Introduced methods `tryRemoveOldestCacheEntries` and `tryToCacheTheEntry` to handle cache eviction and cache size management.

QA Instructions:

-Use the following branch: [CXSPA-8964-testing](https://github.com/SAP/spartacus/tree/feature/CXSPA-8964-testing).
-Start the application in SSR mode.
-Navigate to the website and refresh a few pages.
-Observe the logs for "CACHE IN BYTES" output.


core-libs/setup/ssr/optimized-engine/rendering-cache.ts
```diff
@@ -66,4 +85,45 @@ export class RenderingCache {
...
+    key: string,
+    entry: RenderingEntry
+  ) {
+    if (
+      this.options?.cacheLimit &&
+      htmlSize + this.usedCacheSize <= this.options.cacheLimit
+    ) {
+      this.renders.set(key, entry);
+      this.usedCacheSize += htmlSize;
+    }
```
Platonn, line 137: [QUESTION]
When can it happen that we won't have enough space in the cache?
Shall we print some logger.error() in this case?

core-libs/setup/ssr/optimized-engine/rendering-cache.ts
```diff
@@ -22,9 +23,32 @@ export class RenderingCache {
...
+    if (this.options?.ssrFeatureToggles?.cacheSizeInBytes) {
+      if (html) {
+        htmlSize = this.getHtmlSize(html);
+      }
+
+      this.renders.delete(key);
+      this.tryRemoveOldestCacheEntries(htmlSize);
+
+      if (shouldCache) {
+        this.tryToCacheTheEntry(htmlSize, key, entry);
```
Platonn: [QUESTION]
Shall we also estimate somehow the size of errors, if stored errors entries are stored in the cache?

core-libs/setup/ssr/optimized-engine/rendering-cache.ts
```diff
@@ -66,4 +85,45 @@ export class RenderingCache {
...
+  protected tryRemoveOldestCacheEntries(htmlSize: number): void {
+    if (this.options?.cacheLimit) {
+      while (
+        this.usedCacheSize + htmlSize > this.options?.cacheLimit &&
+        this.usedCacheSize > 0
+      ) {
+        const oldestKey = this.renders.keys().next().value;
+        if (oldestKey !== undefined) {
+          const oldestEntry = this.renders.get(oldestKey);
+          const oldestHtmlSize = this.getHtmlSize(oldestEntry?.html || '');
```
Platonn: [QUESTION]
Instead of calculating the size of the entry each time we deal with it, do you think it makes sense to store the size once calculated as part of the entry (as a new property `entry.size`) so we could later reuse it?

core-libs/setup/ssr/optimized-engine/rendering-cache.ts
```diff
@@ -22,9 +23,32 @@ export class RenderingCache {
...
+
+    if (this.options?.ssrFeatureToggles?.cacheSizeInBytes) {
+      if (html) {
+        htmlSize = this.getHtmlSize(html);
+      }
+
+      this.renders.delete(key);
+      this.tryRemoveOldestCacheEntries(htmlSize);
+
+      if (shouldCache) {
```
Platonn: [MEDIUM]
Don't you think we check `if(shouldCache)` too late, i.e. after already deleting some other cache entries in llines 43-44?
What do you think about adding instead just a simple short-circuit check in the top of this function:
```
if(!shouldCache) {
 return;
}
```

core-libs/setup/ssr/optimized-engine/rendering-cache.ts
```diff
@@ -66,4 +85,45 @@ export class RenderingCache {
...
     return Date.now() - (this.renders.get(key)?.time ?? 0) < this.options?.ttl;
   }
+
+  getHtmlSize(html: string) {
+    return Buffer.byteLength(html, 'utf8');
+  }
+
+  getUsedCacheSize() {
+    return this.usedCacheSize;
+  }
```
Platonn: [NITPICK]
Those methods are not used outside so should be `protected`.
(note: if it's meant only for unit tests, I guess we can workaround it in tests by using a bracket syntax, e.g. `renderingCache['getUsedCacheSize']()`

core-libs/setup/ssr/optimized-engine/ssr-optimization-options.ts
```diff
@@ -189,6 +215,26 @@ type DeepRequired<T> = {
...
 };
 
+/**
+ * Converts a given size (KB, MB, GB) to its equivalent value in bytes.
+ *
+ * @param limit - The numerical value representing the cache size.
+ * @param unit - The unit of measurement ('KB', 'MB', 'GB').
+ * @returns The corresponding cache size in bytes.
+ */
+export function getCacheLimitInBytes(
```
Platonn: [NITPICK]
Let's rename it to `convertToBytes`
(this function has nothing to do with cache, it's a pure calculation util for bytes)
PS. then let's update also JS Docs there: https://github.com/SAP/spartacus/pull/20004/files#diff-356a19d499b69cdf45f9b69dacf25afbaabf54500d3e90aaa2a6406a9f12d69aR54

core-libs/setup/ssr/optimized-engine/ssr-optimization-options.ts
```diff
@@ -234,5 +281,6 @@ export const defaultSsrOptimizationOptions: DefaultSsrOptimizationOptions = {
   renderKeyResolver: getDefaultRenderKey,
   ssrFeatureToggles: {
     avoidCachingErrors: false,
+    cacheSizeInBytes: false,
```
Platonn: [NITPICK]
Let's rename it to `cacheLimitInBytes` - to align with the new config property `....Limit` (instead of the old `....Size`)

kpawelczak: good catch, it was late adjustment and didnt see it, fixed

core-libs/setup/ssr/optimized-engine/ssr-optimization-options.ts
kpawelczak: yeah, I was thinking of naming something like that, fixed.

core-libs/setup/ssr/optimized-engine/rendering-cache.ts
kpawelczak: added `protected `to these functions

core-libs/setup/ssr/optimized-engine/rendering-cache.ts
kpawelczak: I thought about it, but the previous 'caching function' also first removes the entry. Anyway, I changed it to check it earlier.

core-libs/setup/ssr/optimized-engine/rendering-cache.ts
kpawelczak: It does seem to be better, I've added new property

core-libs/setup/ssr/optimized-engine/rendering-cache.ts
kpawelczak: I imagine that once the limit is reached, it will continuously hit the limit, resulting in a large number of logged messages. These logs would confirm that the cache limit is functioning, is this what we want?

core-libs/setup/ssr/optimized-engine/rendering-cache.ts
kpawelczak: I wasn't aware of the errors. But now I think we should calculate the size of the whole entry instead of just the HTML.

I have a question: Can the error object contain properties other than `name`, `message`, and `stack`?
I need a string representation to calculate the byte length, but I don’t fully trust JSON.stringify.
I plan to create `errStr `manually. However, since this string won’t include property names, I need to account for additional bytes. Rough estimate: 10 bytes per property name.

```typescript
    if (entry.err) {
      // errStr = JSON.stringify(entry.err)

      let errStr = '';

      if (entry.err.name) {
        errStr += entry.err.name;
      }

      if (entry.err.message) {
        errStr += entry.err.message;
      }

      if (entry.err.stack) {
        errStr += entry.err.stack;
      }
 
      totalSize += Buffer.byteLength(errStr, 'utf8');
    }
``` 
Would you suggest a better way of handling this or are you ok with strigify?

core-libs/setup/ssr/optimized-engine/ssr-optimization-options.ts
```diff
@@ -27,6 +27,9 @@ export interface SsrOptimizationOptions {
   cache?: boolean;
 
   /**
+   *
+   * @deprecated since ?
```
kpawelczak, line 31: TBD

core-libs/setup/ssr/optimized-engine/rendering-cache.ts
Platonn: Do I understand correctly that in within this function `tryToCacheTheEntry` such a sitiuation should not happen? Becasue we're already called before the function `tryRemoveOldestCacheEntries()` which should ensure that there is enough space for this entry?

core-libs/setup/ssr/optimized-engine/rendering-cache.ts
Platonn: > I have a question: Can the error object contain properties other than name, message, and stack?

yes

The thrown error object can be anything, not only an instance of Error. Or an instance of Error with `cause` property which can contain anything.
in general it can be anything, even containing curcular data structures.
because of circular structures, we can't use a simple stringify
because a real memory representation is not equivalent to the JSON.stringify of an object, I also don't think we should use it as a heuristic.

Moreover we don't want to build any custom calculator **if it would cause additional performance slowdown** to the whole SSR.

To be honest I know no good answer on how to estimate the size of err if it can be anything.

I'm having hard time to find a balanced solution.
What about:
- keep your proposed solution for errStr
- at the same time, log a warning when someone attempts to cache an Error, explaining that errors should not be cached, and it might cause inaccurate calculations for the cache size
?

core-libs/setup/ssr/optimized-engine/rendering-cache.ts
kpawelczak: Yes, and based on my testing, it removes old entries until there is enough available space for new ones.

core-libs/setup/ssr/optimized-engine/rendering-cache.ts
```diff
@@ -66,4 +88,74 @@ export class RenderingCache {
...
+    }
+
+    if (entry.html) {
+      objStr += entry.html;
+    }
+
+    const propSize = 10;
+    const estimatedPropSize = 5 * propSize;
+
+    return Buffer.byteLength(objStr, 'utf8') + estimatedPropSize;
```
Platonn, line 118: [MAJOR]
- Buffer.byteLength(string) is 50% underestimated size of a string in NodeJS. It assumes one byte per character. In practice for our HTMLs it uses 2 bytes per character (which I discovered by taking Heap Snapshots in NodeJS DevTools and investigating the size of cached strings there; and confirmed also with [live pm2 memory monitoring](https://pm2.keymetrics.io/docs/usage/monitoring/) with caching enabled).

So we should use `2* Buffer.byteLength(objStr, 'utf8')`.

I'm not sure we should take into account the marginal size of other properties `estimatedPropSize`.
It's becasue it's size compared to 200-300KB per 1 cache entry is negligible. 
And we don't need to be over-precise here, becasue anyway we need a big big margin for the "working memory" for rendering Angular applications - the memory allocated only for the temporary rendering of an application for the incoming request and then made free by garbage collector.

core-libs/setup/ssr/optimized-engine/ssr-optimization-options.ts
```diff
@@ -216,6 +262,7 @@ type DefaultSsrOptimizationOptions = Omit<
 export const defaultSsrOptimizationOptions: DefaultSsrOptimizationOptions = {
   cache: false,
   cacheSize: 3000,
+  cacheLimit: convertToBytes(3, 'GB'),
```
Platonn, line 265: [MAJOR]
We should not use full 3GB of the NodeJS process for the cache. Let's use at most 500MB by default for cache. Besides the cache entries, in the NodeJS process we need a lot of working memory for rendering applications (that is made free by garbage collector some time after the rendering completes). While [measuring live pm2 memory usage](https://pm2.keymetrics.io/docs/usage/monitoring/), I noticed 500-700 of MB with `concurrency` config being `10` (up to 1-2 GB when `concurrency` config increased to `50`) - that memory is temporarily allocated just for the sake of rendering anuglar applications (while rendering for many parallel requests) and then this memory is made free by the garbage collector after some time.
But if in the meantime the memory exceeds the configured [max_memory_limit](https://pm2.keymetrics.io/docs/usage/memory-limit/) of pm2, then the NodeJS process is restarted (which we don't want to happen).
That's why the default cache size should not be generous and main part of the memory should be reserved for the working memory.
Customers can tweak the concrete cacheLimit value, according to their needs, later.

Platonn:
Closing in favor of https://github.com/SAP/spartacus/pull/20251

PR #20025, opened 2025-02-25
fix(schematics): replace caret (^) with tilde (~) for Spartacus dependencies in package.json

Platonn:
Why it's needed:
Angular CLI by default puts `^` in customer's `package.json` during `ng add @spartacus/schematics@latest`. But since 2211.19 we agreed to break SemVer and release breaking changes in "minor" (x.Y.z) versions, so we need to replace `^` with `~` to prevent breaking changes from being installed automatically in the future and causing peer dependency conflicts when Spartacus updates e.g. Angular versions.

fixes https://jira.tools.sap/browse/CXSPA-9503

---
**QA steps**:
- checkout to this branch, build libs and deploy to verdaccio
- create a fresh ng19 app
- install Spartacus from libs from verdaccio via `ng add @spartacus/schematics@2211.37.0-1` command
- observe `package.json` -> initially `"@spartacus/schematics"` dev dependency will have `^` (as initially created by Angular CLI), but eventually it will have `~` (like all other `@spartacus` deps) after our installation schematics logic kicks in.


pawelfras (approved):
Review and QA done. LGTM

PR #20036, opened 2025-02-27
fix: fix unsanitized untrusted input

andreighervan7:
Ticket: [CXSPA-8078](https://jira.tools.sap/browse/CXSPA-8078)


ercultimate:
I understand this is still in progress, but please ensure that a service is created for HTML string sanitization, and that service is used instead of `sanitize-html` directly.  This will insulate the dependent components and services against changes to the sanitization implementation and will provide a centralized location for security analysis.

projects/core/package.json
```diff
@@ -26,7 +26,8 @@
     "i18next": "^24.2.1",
     "i18next-http-backend": "^3.0.1",
     "i18next-resources-to-backend": "^1.2.1",
-    "rxjs": "^7.8.2"
+    "rxjs": "^7.8.2",
+    "sanitize-html": "^2.14.0"
```
giancorderoortiz, line 30: Approval for this new OSS lib is required. 
Please create compliance ticket. 
Internal process is required.

PR #20043, opened 2025-02-28
OCC Punchout Session & Requisition endpoints

FollowTheFlo:
2 occ endpoint: getPunchoutSession & getPunchoutRequisition.
Both covered with connector, adapter, facade, service.


package.json
```diff
@@ -110,6 +110,7 @@
...
     "test:storefront:lib": "nx test storefrontlib --source-map --code-coverage",
     "test:opf:lib": "nx test opf --source-map --code-coverage",
+    "test:punchout:lib": "nx test punchout --source-map --code-coverage",
```
FollowTheFlo: Ease purpose to run punchout unit test in a single command.
It will be removed at the end when epic/punchout will be merged to develop.

integration-libs/punchout/root/model/punchout.model.ts
```diff
@@ -0,0 +1,24 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+export const PUNCHOUT_SESSION_KEY = 'sid';
+
+export interface PunchoutSession {
+  customerId: string;
+  cartId: string;
+  punchOutLevel: string;
+  punchOutOperation: string;
```
Matejk00: Hint: For fields like `punchOutOperation` if it's not generated value, it may be worth considering using an enum instead of string for better type safety and validation.

integration-libs/punchout/root/facade/punchout.facade.ts
```diff
@@ -0,0 +1,37 @@
...
+@Injectable({
+  providedIn: 'root',
+  useFactory: punchoutFacadeFactory,
+})
+export abstract class PunchoutFacade {
+  /**
+   * Abstract method used to get Punchout Session
+   * @param sId is the sesssion Id given by ARIBA via url param
+   */
+  abstract getPunchoutSession(sId: string): Observable<PunchoutSession>;
```
Matejk00: Could we change `sId` param -> to `sessionId`?
In my opinion will be more readable.

integration-libs/punchout/occ/config/default-occ-punchout-config.ts
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { OccConfig } from '@spartacus/core';
+import { PunchoutOccEndpoints } from '../model/occ-punchout.model';
+
+const punchoutOccEndpoints: PunchoutOccEndpoints = {
+  punchoutSession: 'punchout/sessions/${sId}',
+  punchoutRequisition: 'punchout/sessions/${sId}/requisition',
```
Matejk00: Doesn't requisition affect a specific session?
So we could change a name to `punchoutSessionRequisition`?

Matejk00 (changes requested):

integration-libs/punchout/root/facade/punchout.facade.ts
FollowTheFlo: sounds good

integration-libs/punchout/occ/config/default-occ-punchout-config.ts
FollowTheFlo: done

integration-libs/punchout/root/model/punchout.model.ts
FollowTheFlo: done for both punchOutLevel and punchOutOperation following cxml standard.

Matejk00 (approved):

PR #20047, opened 2025-03-04
chore: Run only core e2e tests for docs commits

rmch91:
Extends logic to not run full e2e tests for docs commits

### DIFF
--- a/ci-scripts/e2e-cypress.sh
+++ b/ci-scripts/e2e-cypress.sh
@@ -77,13 +77,16 @@ echo '-----'
 echo "Building Spartacus storefrontapp"
 npm run build
 
-is_bot_commit() {
+should_run_only_core_e2e() {
     LAST_COMMIT_AUTHOR=$(git log -1 --pretty=format:'%ae')
+    LAST_COMMIT_MESSAGE=$(git log -1 --pretty=format:'%s')
 
     echo "Last commit author: ${LAST_COMMIT_AUTHOR}"
+    echo "Last commit message: ${LAST_COMMIT_MESSAGE}"
 
     if [[ "${LAST_COMMIT_AUTHOR}" == *"dependabot[bot]@users.noreply.github.com" ]] ||
-        [[ "${LAST_COMMIT_AUTHOR}" == *"renovate[bot]@users.noreply.github.com" ]]; then
+        [[ "${LAST_COMMIT_AUTHOR}" == *"renovate[bot]@users.noreply.github.com" ]] ||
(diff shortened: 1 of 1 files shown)


ci-scripts/e2e-cypress.sh
```diff
@@ -77,13 +77,16 @@ echo '-----'
...
     LAST_COMMIT_AUTHOR=$(git log -1 --pretty=format:'%ae')
+    LAST_COMMIT_MESSAGE=$(git log -1 --pretty=format:'%s')
 
     echo "Last commit author: ${LAST_COMMIT_AUTHOR}"
+    echo "Last commit message: ${LAST_COMMIT_MESSAGE}"
 
     if [[ "${LAST_COMMIT_AUTHOR}" == *"dependabot[bot]@users.noreply.github.com" ]] ||
-        [[ "${LAST_COMMIT_AUTHOR}" == *"renovate[bot]@users.noreply.github.com" ]]; then
+        [[ "${LAST_COMMIT_AUTHOR}" == *"renovate[bot]@users.noreply.github.com" ]] ||
+        [[ "${LAST_COMMIT_MESSAGE}" == doc:* ]]; then
```
Platonn, line 89: Do we really need such a workaround now?
I believe ideally we should determine the scope of CI jobs to run based on the changed and affected files, but not commit message.

Platonn: side note: I can see a the number of commit messages in develop starting with `doc` among last 1 000 commits is only 32.
and among last 10 000 is only 93.

(source: `git log -n 10000 --oneline | grep -c "^[a-f0-9]\{10\} doc"`)

rmch91: Maybe you have right, it was just a quick workaround that came to my mind when I saw last commit from Matt, where we run all e2e tests for changes only in md files https://github.com/SAP/spartacus/actions/runs/13249385746. 
But after your comment, I noticed that we have a check already based on changed files, but it seems it doesn't work. 
validate-e2e-execution.sh
```
FILES=`git diff --name-only origin/$GITHUB_BASE_REF`

    for file in $FILES; do

        case "$file" in
        *.md | docs/** | tools/** | *.spec.ts | **/schematics/** )
            ;;
        * )
            # if anything other than `*.md | docs/** | tools/** | *.spec.ts | **/schematics/**` are found, then we should run e2es
            RUN_E2E=true
            ;;
        esac
    done
```
So, yeah, maybe we should investigate it instead of creating this PR

PR #20056, opened 2025-03-06
fix: (CXSPA-9517) - Fix pagination not focused correctly

petarmarkov9449:
Closes: [CXSPA-9517](https://jira.tools.sap/browse/CXSPA-9517)

Fix pagination not focused correctly.


feature-libs/organization/administration/components/shared/list/list.component.ts
Pio-Bar, line 1: Looks like this code is not used anywhere. Probably leftover from previous implementation.

projects/storefrontlib/cms-components/product/product-list/container/product-list.component.ts
Pio-Bar, line 1: Probably same leftovers here.

Pio-Bar (changes requested):
I think we only need unique IDs for pagination components that repeat on the same page. We can let the other fall back to the default id.

projects/storefrontlib/cms-components/product/product-list/container/product-list.component.ts
petarmarkov9449: I've just removed it.

feature-libs/organization/administration/components/shared/list/list.component.ts
petarmarkov9449: Oh, I haven't removed it yet.
Just removed.

Pio-Bar (dismissed):

Zeyber (changes requested):
Its a nice solution! We should definitely include units and e2es to avoid regression.

projects/storefrontlib/shared/components/list-navigation/pagination/pagination.component.html
```diff
@@ -13,6 +13,7 @@
     (click)="pageChange(item)"
     [attr.aria-label]="getAriaLabel$(item) | async"
     [attr.title]="getAriaLabel$(item) | async"
+    [attr.data-focus-key]="paginationID + i.toString()"
```
Zeyber: I don't really think we should add this attribute just for the sake of testing. Can we use `cy.focused()` for the e2e?

Zeyber (changes requested):

projects/storefrontlib/shared/components/list-navigation/pagination/pagination.component.html
petarmarkov9449: thanks for your information, I've just updated it.

projects/storefrontapp-e2e-cypress/cypress/helpers/product-search.ts
```diff
@@ -309,3 +309,11 @@ export function assertNumberOfProducts(alias: string, category: string) {
     checkFirstItem(firstProduct);
   });
 }
+
+export function assertPaginationFocusBehavior(expectedLabel: string): void {
+  cy.focused()
```
Zeyber, line 319: I think you can just use `contains(expectedLabal)` here instead of the larger chain.

projects/storefrontapp-e2e-cypress/cypress/e2e/regression/product-search/product-search.core-e2e.cy.ts
```diff
@@ -24,14 +24,17 @@ context('Product search', { testIsolation: false }, () => {
     describe('Pagination', () => {
       it('should navigate to the next page and display results', () => {
         productSearchFlow.verifyNextPage(2);
+        productSearchFlow.assertPaginationFocusBehavior('2');
```
Zeyber: Would this method fit better in the `verifyXPage` methods themselves?

Zeyber (changes requested):

projects/storefrontapp-e2e-cypress/cypress/e2e/regression/product-search/product-search.core-e2e.cy.ts
petarmarkov9449: I've just updated the `verifyXPage` methods.

projects/storefrontapp-e2e-cypress/cypress/helpers/product-search.ts
petarmarkov9449: As we updated the `verifyXPage` methods to check focus automatically, I cannot set the expected focused pageNumber manually.
So I need to update the logic based on that.

Actually, it is a bit tricky to get the next focused element automatically (for example, if I click 2 when 1,2,3, then the focus remains in 2 but if I click 2 when 2, 3,4, then the focus remains in 1 not 2).

Zeyber (approved):

PR #20064, opened 2025-03-11
CXSPA-9583: 'No option selected' should not appear when 'domainOnDemand' is enabled

Larisa-Staroverova:


ChristophHi (changes requested):
Not sure about this, can't we just do _one_ change in method addRetractValue in the normalizer?

PR #20071, opened 2025-03-12
fix: (CXSPA-9010) Previous field error message read.

StanislavSukhanov:
* closes: https://jira.tools.sap/browse/CXSPA-9010


Pio-Bar (dismissed):

projects/core/src/features-config/feature-toggles/config/feature-toggles.ts
```diff
@@ -1074,6 +1080,7 @@ export const defaultFeatureToggles: Required<FeatureTogglesInterface> = {
   a11yStoreFinderAlerts: true,
   a11yStoreFinderLabel: false,
   a11yFormErrorMuteIcon: true,
+  a11yImprovedErrorMessage: true,
```
Zeyber: I think we should introduce this as `false` by default.

projects/storefrontlib/shared/components/form/form-errors/form-errors.component.ts
```diff
@@ -40,8 +42,22 @@ import { map, startWith } from 'rxjs/operators';
 export class FormErrorsComponent implements DoCheck {
   private featureConfigService = inject(FeatureConfigService);
 
+  private elementRef = inject(ElementRef);
```
Zeyber: Lets use `protected` here.

projects/storefrontlib/shared/components/form/form-errors/form-errors.component.ts
```diff
@@ -40,8 +42,22 @@ import { map, startWith } from 'rxjs/operators';
...
+
   constructor(protected ChangeDetectionRef: ChangeDetectorRef) {
     useFeatureStyles('a11yFormErrorMuteIcon');
+    if (this.featureConfigService.isEnabled('a11yImprovedErrorMessage')) {
```
Zeyber: We could move this logic into its own method just to be a little tidier.

projects/storefrontlib/shared/components/form/form-errors/form-errors.component.ts
```diff
@@ -40,8 +42,22 @@ import { map, startWith } from 'rxjs/operators';
...
+      const ariaLive = inject(new HostAttributeToken('aria-live'), {
+        optional: true,
+      });
+
+      if (!ariaLive) {
+        // If no aria-live value is set add 'polite' as a default. This is preferred over setting
+        // role='alert' so that screen readers do not interrupt the current task to read this aloud.
+        this.elementRef.nativeElement.setAttribute('aria-live', 'polite');
```
Zeyber: We should be careful with DOM manipulation on page instantiation because it can cause issues with SSR and hydration. Is it possible to set the attribute using an Angular-friendly method (eg. template, input value)?
https://angular.dev/guide/hydration#direct-dom-manipulation

Zeyber (changes requested):

projects/core/src/features-config/feature-toggles/config/feature-toggles.ts
StanislavSukhanov: Yes, you are right! Thanks a lot for noticing.

projects/storefrontlib/shared/components/form/form-errors/form-errors.component.ts
```diff
@@ -35,13 +38,24 @@ import { map, startWith } from 'rxjs/operators';
   selector: 'cx-form-errors',
   templateUrl: './form-errors.component.html',
   changeDetection: ChangeDetectionStrategy.OnPush,
+  host: {
```
Zeyber: Do you think we could use `@HostBinding`s for these attributes like you did for `attr.role`? I think it could be easier to understand and modify.

Zeyber (changes requested):

projects/storefrontlib/shared/components/form/form-errors/form-errors.component.ts
StanislavSukhanov: Got it.

StanislavSukhanov: For me, personally, using host in the component decorator is more readable. 
https://angular.dev/guide/components/host-elements
here we have decorators listed as 'alternatively'

projects/storefrontlib/shared/components/form/form-errors/form-errors.component.ts
```diff
@@ -115,6 +137,15 @@ export class FormErrorsComponent implements DoCheck {
     return { ...errorDetails, ...this.translationParams };
   }
 
+  setAriaLiveAttribute() {
```
Zeyber: I wonder if we can also set this directly in the host binding rather than calling from constructor?

projects/storefrontlib/shared/components/form/form-errors/form-errors.component.ts
```diff
@@ -40,8 +43,15 @@ import { map, startWith } from 'rxjs/operators';
 export class FormErrorsComponent implements DoCheck {
   private featureConfigService = inject(FeatureConfigService);
 
+  protected elementRef = inject(ElementRef, { optional: true });
+  protected ariaLiveToken = inject(new HostAttributeToken('aria-live'), {
```
Zeyber, line 46: Is there any reason we need to use both of the Token and Signal vars?

Zeyber (changes requested):

projects/storefrontlib/shared/components/form/form-errors/form-errors.component.ts
StanislavSukhanov: Gotcha. Thank you

projects/storefrontlib/shared/components/form/form-errors/form-errors.component.ts
```diff
@@ -127,5 +146,22 @@ export class FormErrorsComponent implements DoCheck {
...
     return !(this.invalid && (this.touched || this.dirty));
   }
-  @HostBinding('attr.role') role = 'alert';
+  @HostBinding('attr.role') role = this.featureConfigService.isEnabled(
+    'a11yImprovedErrorMessage'
+  )
+    ? null
+    : 'alert';
+
+  @HostBinding('attr.aria-live') get ariaLiveValue() {
```
Zeyber: Forgive my ignorance on this one. Is the `get ariaLiveValue()` a real method? How is it working if so?

projects/storefrontlib/shared/components/form/form-errors/form-errors.component.ts
```diff
@@ -127,5 +146,22 @@ export class FormErrorsComponent implements DoCheck {
...
+    if (!this.featureConfigService.isEnabled(
+      'a11yImprovedErrorMessage'
+    )) {
+      return null;
```
Zeyber: Would this override any preset `aria-live` value to `null`? Or its ok because we don't set it to anything?

Zeyber (changes requested):
We should also unit test the form errors component since for the `aria-live` attribute and the `ngDoCheck` as the logic become more complex.

projects/storefrontlib/shared/components/form/form-errors/form-errors.component.ts
Zeyber: This is completely reasonable point. I was just confused seeing both. I think its better to be consistent with whichever was there first just in my opinion.

projects/storefrontlib/shared/components/form/form-errors/form-errors.component.ts
StanislavSukhanov: It is interesting thing. Let me check it out. 
Initial intent was to ignore the aria-live, because if we do not have ` 'a11yImprovedErrorMessage'` enabled, the role of the host is being set to `alert`

projects/storefrontlib/shared/components/form/form-errors/form-errors.component.ts
StanislavSukhanov: not at all, the reason it was set a a method was because it wasn't working in conjunction with a signal I've previously used. let me correct it

projects/storefrontlib/shared/components/form/form-errors/form-errors.component.ts
StanislavSukhanov: fixed it. now it's a bulletproof and doesnt ignore the aria-live attribute if it's being set from outside

projects/storefrontlib/shared/components/form/form-errors/form-errors.component.ts
```diff
@@ -98,8 +105,20 @@ export class FormErrorsComponent implements DoCheck {
 
   ngDoCheck(): void {
     if (this.control.touched !== this.previousTouchedState) {
-      this.previousTouchedState = this.control.touched;
-      this.ChangeDetectionRef.markForCheck();
+      if (
+        this.featureConfigService.isEnabled('a11yImprovedErrorMessage') &&
+        this.elementRef?.nativeElement?.getAttribute('aria-live') === 'polite'
```
Zeyber, line 110: Can we just check the token value now and remove elementRef?

Zeyber (changes requested):

projects/storefrontlib/shared/components/form/form-errors/form-errors.component.ts
StanislavSukhanov: 🤔 let's try

StanislavSukhanov: well, it doesn't work. Since HostAttribute token injects static attributes of the host node it appears as `null` even if we set aria-live via host binding. 

From my understanding it is being checked once when component is initialised and if the host hasn't had this an attribute during the initialisation, the token's value will remain `null`

Zeyber: Worth a shot, thanks for giving it a go :)

Zeyber (changes requested):
Nice work! Lets just add the unit tests for making sure aria-live is correctly set now on the form errors component.

projects/storefrontlib/shared/components/form/form-errors/form-errors.component.spec.ts
```diff
@@ -104,6 +104,24 @@ describe('FormErrors', () => {
     );
   });
 
+  describe('a11y', () => {
+    it('should have `aria` attributes', () => {
+      fixture.detectChanges();
+      const ariaLive =
```
Zeyber, line 110: I would suggest the `aria-live` attribute is the only thing we need to test because its value is conditional. We should check the value is correct depending on the condition (ie. ariaLiveToken).

Zeyber (changes requested):

projects/storefrontlib/shared/components/form/form-errors/form-errors.component.spec.ts
StanislavSukhanov: fixed. Thank you.

Zeyber (approved):

PR #20072, opened 2025-03-12
fix: CXSPA-8494 check for keyword redirect before loading default results

darvsorceix:


projects/core/src/product/facade/searchbox.service.ts
```diff
@@ -5,18 +5,26 @@
...
 import { ProductSelectors } from '../store/selectors/index';
 import { ProductSearchService } from './product-search.service';
+import { StateWithProduct } from '../store/product-state';
 
 @Injectable({
   providedIn: 'root',
 })
 export class SearchboxService extends ProductSearchService {
+  constructor(
+    protected actionsSubject: ActionsSubject,
```
ercultimate: Wouldn't this be considered a breaking change?  
I also believe the recommended process is to use the `inject()` function for DI (which would resolve the breaking change issue).

rmch91 (dismissed):
LGTM

rmch91 (approved):

projects/storefrontlib/cms-components/navigation/search-box/search-box-component.service.ts
```diff
@@ -261,12 +313,24 @@ export class SearchBoxComponentService {
...
   launchSearchPage(query: string): void {
-    this.routingService.go({
-      cxRoute: 'search',
-      params: { query },
-    });
+    // Reset the completed state before starting new search
+    this.searchCompleted.next(false);
+    this.searchCompleted
+      .pipe(
+        filter((complete) => complete), // Only proceed when true
```
dunqan, line 323: @darvsorceix how this one is supposed to work, if there is no pending searches?
(I'm asking, as it breaks a search functionality in existing implementation)

darvsorceix: This should be fixed in https://github.com/SAP/spartacus/pull/20483

PR #20081, opened 2025-03-13
Login when initial token is invalid

FollowTheFlo:
Workaround to address CXSPA-9608 - Public pages not displayed when token is invalid

Overriding AuthHttpHeaderService.handleExpiredRefreshToken method which was redirecting user to login page.
We want user to stay on Punchout Session page, so the Punchout Session CMS page can be fetched.

### DIFF
--- a/integration-libs/punchout/root/model/punchout.model.ts
+++ b/integration-libs/punchout/root/model/punchout.model.ts
@@ -6,6 +6,7 @@
 
 export const PUNCHOUT_SESSION_KEY = 'sid';
 export const PUNCHOUT_ERROR_PAGE_URL = '/punchout/cxml/error';
+export const PUNCHOUT_SESSION_PAGE_URL = '/punchout/cxml/session';
 export const PUNCHOUT_SESSION_ID = 'punchoutSessionId';
 
 export enum PunchOutLevel {
--- a/integration-libs/punchout/root/services/punchout-detection.service.ts
+++ b/integration-libs/punchout/root/services/punchout-detection.service.ts
@@ -1,3 +1,9 @@
+/*
+ * SPDX-FileCopyrightText: 2025 SAP Spartacus team <spartacus-team@sap.com>
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
 import { Location } from '@angular/common';
 import { Injectable, inject } from '@angular/core';
 import { PUNCHOUT_SESSION_PAGE_URL } from '../model';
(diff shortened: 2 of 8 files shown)


integration-libs/punchout/root/model/punchout.model.ts
```diff
@@ -6,6 +6,7 @@
 
 export const PUNCHOUT_SESSION_KEY = 'sid';
 export const PUNCHOUT_ERROR_PAGE_URL = '/punchout/cxml/error';
+export const PUNCHOUT_SESSION_PAGE_URL = '/punchout/cxml/session';
```
FollowTheFlo, line 9: Those urls are in model file for the mom.
We will more than likely move it into config file as it needs to match the ones in CMS pages.
Discussing with server side to make it shorter too 'punchout/session'.

integration-libs/punchout/root/services/punchout-detection.service.ts
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { Location } from '@angular/common';
+import { Injectable, inject } from '@angular/core';
+import { PUNCHOUT_SESSION_PAGE_URL } from '../model';
+
+@Injectable({ providedIn: 'root' })
+export class PunchoutDetectionService {
```
FollowTheFlo, line 12: 'Detection' is the best wording I could find, feel free to propose others. 'PunchoutAuthService' is already used in core folder.

Matejk00 (approved):

integration-libs/punchout/root/services/punchout-detection.service.ts
```diff
@@ -0,0 +1,27 @@
...
+
+  /**
+   * Check if browser url is the punchout initial session page.
+   * With default config, the expected url shape is '/punchout/cxml/session?abcd'.
+   * @returns boolean
+   */
+  isPunchoutSessionPage(): boolean {
+    const urlSections = this.location.path().split('?');
+    return (
+      urlSections.length > 1 &&
```
Matejk00, line 23: Should we check length first?

Matejk00: Alternative:
```
/**
 * Checks if the browser URL is the punchout session page.
 * Default URL shape: '/punchout/cxml/session?abcd'.
 * @returns boolean
 */
isPunchoutSessionPage(): boolean {
  const [path] = this.location.path().split('?');
  return path.includes(PUNCHOUT_SESSION_PAGE_URL);
}
```

FollowTheFlo: length is checked first, which avoid to run the 'includes' method.
note that location.path never returns null/undefined so I didn't add any check for it.

PR #20089, opened 2025-03-18
refactor: const enums were replaced with enums

npapp-dev002:
An Eslint rule was added to prohibit const enum declarations. Const enums were replaced with enums.

closes [CXSPA-9628](https://jira.tools.sap/browse/CXSPA-9628)


kimhw0630:
Tested a new ESLint rule and searched for missing ones.
image

kimhw0630 (approved):

kimhw0630:
I see an error message when I run npm run lint, but is it possible to see errors immediately after writing code in VSCode?

image

kimhw0630:
Never mind, I see it now.

image

PR #20090, opened 2025-03-18
fix: add error message for invalid query

kpawelczak:
closes: [CXSPA-9632](https://jira.tools.sap/browse/CXSPA-9632)


projects/core/src/product/store/effects/product-search.effect.ts
```diff
@@ -35,14 +41,28 @@ export class ProductsSearchEffects {
...
+                catchError((error) => {
+                  const normalizedError: HttpErrorModel = tryNormalizeHttpError(
+                    error,
+                    this.logger
+                  );
+                  if (
+                    normalizedError?.details &&
+                    normalizedError.details[0].type ===
+                      'ArrayIndexOutOfBoundsError'
+                  ) {
```
ercultimate:
```suggestion
                  if (
                    normalizedError?.details?.[0].type ===
                      'ArrayIndexOutOfBoundsError'
                  ) {
```
Is there a reason not to continue using optional chaining operators?

kpawelczak: I've changed it, thanks for the suggestion

ercultimate (approved):

PR #20092, opened 2025-03-18
fix: fix unsanitized untrusted input v2

andreighervan7:
ticket: [CXSPA-8078](https://jira.tools.sap/browse/CXSPA-8078)


feature-libs/product-configurator/rulebased/components/show-more/configurator-show-more.component.spec.ts
```diff
@@ -35,79 +42,69 @@ describe('ConfiguratorShowMoreComponent', () => {
...
       'span'
     );
-    CommonConfiguratorTestUtilsService.expectElementPresent(
-      expect,
-      htmlElem,
-      'button'
+  });
+
+  // removed unit testing because now i used sanitized dom and i did unit testing for this
+  it('should remove HTML tags from input text', () => {
```
rmch91, line 54: Do we need this comment?

feature-libs/product-configurator/rulebased/components/show-more/configurator-show-more.component.ts
```diff
@@ -29,7 +31,10 @@ export class ConfiguratorShowMoreComponent implements AfterViewInit {
   @Input() productName: string;
   @Input() tabIndex = -1;
 
-  constructor(protected cdRef: ChangeDetectorRef) {}
+  constructor(
+    protected cdRef: ChangeDetectorRef,
+    private sanitizer: DomSanitizer
```
rmch91: `major:` For DI, let's use the `inject` function instead of a constructor to avoid introducing unnecessary breaking changes. See: https://wiki.one.int.sap/wiki/display/spar/Coding+Guidelines#CodingGuidelines-Useinject()functioninsteadofconstructorforDependencyInjection

projects/core/src/occ/adapters/product/converters/product-name-normalizer.ts
```diff
@@ -4,15 +4,19 @@
...
 import { OccConfig } from '../../../config/occ-config';
 import { Occ } from '../../../occ-models/occ.models';
+import { DomSanitizer } from '@angular/platform-browser';
 
 @Injectable({ providedIn: 'root' })
 export class ProductNameNormalizer implements Converter<Occ.Product, Product> {
-  constructor(protected config: OccConfig) {}
+  constructor(
+    protected config: OccConfig,
+    private sanitizer: DomSanitizer
```
rmch91: `major:` For DI, let's use the `inject` function instead of a constructor to avoid introducing unnecessary breaking changes. See: https://wiki.one.int.sap/wiki/display/spar/Coding+Guidelines#CodingGuidelines-Useinject()functioninsteadofconstructorforDependencyInjection

rmch91 (changes requested):

feature-libs/product-configurator/package.json
```diff
@@ -29,6 +29,7 @@
     "@angular/common": "^19.1.8",
     "@angular/core": "^19.1.8",
     "@angular/forms": "^19.1.8",
+    "@angular/platform-browser": "^19.1.8",
```
rmch91, line 32: I'm afraid that this could be breaking change for customers, let's discuss it on team channel.

rmch91 (changes requested):

feature-libs/product-configurator/rulebased/components/show-more/configurator-show-more.component.ts
```diff
@@ -53,7 +60,10 @@ export class ConfiguratorShowMoreComponent implements AfterViewInit {
     this.cdRef.detectChanges();
   }
 
-  protected normalize(text: string = ''): string {
-    return text.replace(/<[^>]*>/g, '');
+  normalize(text: string = ''): string {
+    return (
+      this.sanitizer.sanitize(SecurityContext.HTML, text) ||
```
ercultimate: What is the intended operation here?  Currently the expression evaluates to `this.sanitizer.sanitize(SecurityContext.HTML, text)` or, if that is falsy, `''.replace(/<[^>]*>/g, '')`

projects/core/src/occ/adapters/product/converters/product-name-normalizer.ts
```diff
@@ -29,7 +34,10 @@ export class ProductNameNormalizer implements Converter<Occ.Product, Product> {
    * Sanitizes the name so that the name doesn't contain html elements.
    */
   protected normalize(name: string): string {
-    return name.replace(/<[^>]*>/g, '');
+    return (
+      this.sanitizer.sanitize(SecurityContext.HTML, name) ||
```
ercultimate: Same as above, what is the intended operation here?

feature-libs/product-configurator/package.json
rmch91: Adding this dependency is safe.
Every Spartacus customer already has @angular/platform-browser in their dependencies (becasue every Spartacus customer has @spartacus/core in their dependencies, which has @angular/platform-browser as a peer dependency, so customer must have already installed also @angular/platform-browser ).
Therefore introducing @angular/platform-browser as a new peer dependency of the product-configurator feature lib should not break any Spartacus customer.

projects/core/src/occ/adapters/product/converters/product-name-normalizer.ts
```diff
@@ -4,14 +4,17 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { Injectable } from '@angular/core';
+import { inject, Injectable } from '@angular/core';
 import { Product } from '../../../../model/product.model';
 import { Converter } from '../../../../util/converter.service';
 import { OccConfig } from '../../../config/occ-config';
 import { Occ } from '../../../occ-models/occ.models';
+import { SanitizeService } from '@spartacus/core';
```
rmch91: `ProductNameNormalizer` is a part of core library, and now you add dependency on  `'@spartacus/core'`, so it dependent from itself. Because of this CI failed, you can check it in github and reproduce also locally by running `npm run build:core`
image

projects/core/src/config/services/sanitize.service.ts
```diff
@@ -0,0 +1,19 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { Injectable } from '@angular/core';
+import { DomSanitizer, SafeHtml } from '@angular/platform-browser';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class SanitizeService {
```
rmch91: Before I told only that if adding `@angular/platform-browser` as a peerDependency is a breaking change, to workaround we can create a wrapper in the core lib. But Kris confirmed that this peerDependency is safe, in this case we don't need a wrapper. 
IMO creating new services files should solve some problem, if we don't have any problem with `'@angular/platform-browser'` as a peerDependency of product-configurator, there is no need to produce additional code.

projects/core/src/config/services/sanitize.service.ts
```diff
@@ -0,0 +1,19 @@
...
+  constructor(private sanitizer: DomSanitizer) {}
```
rmch91: As wrote above, I think we don't need this service, but just a reminder for the future - we should use inject instead of constructors for the DI. It will help to avoid breaking changes in feature, since any change in the constructor is breaking for customers

rmch91 (changes requested):

rmch91 (approved):
LGTM

PR #20096, opened 2025-03-19
test: add PDP accessibility e2e test

uroslates:
Closes: https://jira.tools.sap/browse/CXSPA-9644


feature-libs/cart/base/components/add-to-cart/add-to-cart.component.html
```diff
@@ -43,7 +43,6 @@
     "
     type="submit"
     [disabled]="disabled || quantity <= 0 || quantity > maxQuantity"
-    [attr.aria-describedby]="productCode + '_header'"
```
Pio-Bar, line 46: Are we sure about removing this? I'm guessing the ID is invalid on PDP but, on PLP this provides additional context. The button can be repeated many times on the same page and to differentiate between them we add the products name as a description. The output looks sth like: `Add to cart, <product name>` 

To fix this we can provide it with a valid ID on PDP as well or make some kind of exception here.

Pio-Bar (changes requested):
Let's pull the latest develop changes and use the new a11yRunContinuumTest command.

feature-libs/cart/base/components/add-to-cart/add-to-cart.component.html
uroslates: The problem here is that PDP is CMS driven with quite a few slots. I decided adding the necessary id (with product title) to product-intro.component.html. Now test is passing W/O need to remove above mentioned attribute (aria-describedby).

projects/storefrontapp-e2e-cypress/cypress/e2e/accessibility/product-details-page.a11y-e2e.cy.ts
```diff
@@ -0,0 +1,51 @@
...
+  beforeEach(() => {
+    cy.a11yContinuumSetup();
+    cy.visit(siteContextSelector.PRODUCT_PATH_2);
+  });
+
+  context('PDP main section', () => {
+    it('should pass a11y check', () => {
+      cy.get('cx-page-slot[position="Tabs"] .tab button:nth-child(1)')
+        .click()
+        .wait(1000);
```
Pio-Bar: We should get loaded elements instead of waiting an arbitrary amount of time.

projects/storefrontapp-e2e-cypress/cypress/e2e/accessibility/product-details-page.a11y-e2e.cy.ts
```diff
@@ -0,0 +1,51 @@
...
+ * This test checks accessibility concerns on the PDP page using Access Continuum
+ */
+describe('Product Details Page Accessibility', () => {
+  beforeEach(() => {
+    cy.a11yContinuumSetup();
+    cy.visit(siteContextSelector.PRODUCT_PATH_2);
+  });
+
+  context('PDP main section', () => {
+    it('should pass a11y check', () => {
```
Pio-Bar:
```suggestion
    it('PDP main section', () => {
```
I think we don't need separate context for each test.

projects/storefrontapp-e2e-cypress/cypress/e2e/accessibility/product-details-page.a11y-e2e.cy.ts
```diff
@@ -0,0 +1,51 @@
...
+        .click()
+        .wait(1000);
+      cy.get('main').a11yRunContinuumTest();
+    });
+  });
+
+  context('PDP with Reviews', () => {
+    it('should pass a11y check', () => {
+      cy.get('cx-product-intro > .rating button').click().wait(1000);
+      cy.get('main').a11yRunContinuumTest();
```
Pio-Bar: Let's scope the test to only include the changed tab. No need to scan the whole page again.

projects/storefrontlib/cms-components/content/tab/panel/tab-panel.component.html
```diff
@@ -1,10 +1,8 @@
 <div
   [attr.role]="mode === TAB_MODE.ACCORDIAN ? 'region' : 'tabpanel'"
   [tabindex]="tab.disableBorderFocus ? -1 : 0"
-  [attr.aria-labelledby]="tab.id ?? null"
-  [attr.id]="
-    tab.id !== null && tab.id !== undefined ? 'section-' + tab.id : null
-  "
+  [attr.aria-labelledby]="ariaLabelledby"
```
Pio-Bar: This ID should match the button that triggered the panel. This change will break that connection.

projects/storefrontlib/cms-components/product/product-intro/product-intro.component.html
```diff
@@ -1,4 +1,8 @@
 <ng-container *ngIf="product$ | async as product">
+  <!-- a11y related - below satisfies add-to-cart.component.html button attr.aria-describedby requirement -->
+  <h4 [attr.id]="product.code + '_header'" style="display: none">
+    {{ product.name }}
+  </h4>
```
Pio-Bar: We recently removed visually hidden elements as these were found to be a11y issues. Might be best to not introduce them back. 

Hmm, if we do not have a product name that's visible to the user, we can conditionally remove the description from the button instead. I think these were only necessary when more than one "add to cart" button was present.

If that's not an option we can try adding the product name to the "add to cart" button's aria-label directly. without using "aria-describedby".

Pio-Bar (changes requested):
Should we also scan the "write a review" form and the "image zoom" modal?

projects/storefrontlib/cms-components/product/product-intro/product-intro.component.html
uroslates: Removed that element and old implementation where elements with IDs product.code + '_header' were expected.
Instead used dynamically constructed ariaLabel attribute attribute value. IMHO that one seems as a cleaner solution.

projects/storefrontapp-e2e-cypress/cypress/e2e/accessibility/product-details-page.a11y-e2e.cy.ts
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import * as siteContextSelector from '../../helpers/site-context-selector';
+
+/**
+ * This test checks accessibility concerns on the PDP page using Access Continuum
+ */
+describe('Product Details Page Accessibility', () => {
```
Pio-Bar:
```suggestion
describe('Product Details Page Accessibility', { testIsolation: false }, () => {
```
As we mentioned on our daily. We can speed up the test by turning the test isolation off.

projects/storefrontapp-e2e-cypress/cypress/e2e/accessibility/product-details-page.a11y-e2e.cy.ts
```diff
@@ -0,0 +1,46 @@
...
+    cy.get('cx-page-slot[position="Tabs"] .tab button:nth-child(2)').click();
+    cy.get(
+      'cx-tab-panel [role="tabpanel"].active cx-product-attributes .container'
+    ).a11yRunContinuumTest();
+  });
+
+  it('PDP with Shipping', () => {
+    cy.get('cx-page-slot[position="Tabs"] .tab button:nth-child(4)').click();
+    cy.get(
+      'cx-tab-panel [role="tabpanel"].active cx-paragraph'
```
Pio-Bar:
```suggestion
      '[position="Tabs"] cx-tab-panel .active''
```

We need to run this on the whole panel not just a single element.

projects/storefrontapp-e2e-cypress/cypress/e2e/accessibility/product-details-page.a11y-e2e.cy.ts
```diff
@@ -0,0 +1,46 @@
...
+    );
+    cy.get('main').a11yRunContinuumTest();
+  });
+
+  it('PDP with Reviews', () => {
+    cy.get('cx-page-slot[position="Tabs"] .tab button:nth-child(3)').click();
+    cy.get(
+      'cx-tab-panel [role="tabpanel"].active cx-product-reviews .container .header'
+    ).a11yRunContinuumTest();
+  });
```
Pio-Bar: This selector would mean the test runs in the header only. We need to make sure all the content loads and run the test on a element that contains it.
Screenshot 2025-04-10 at 13 28 21

feature-libs/cart/base/components/add-to-cart/add-to-cart.component.html
Pio-Bar: Like we discussed during the meeting, the label needs to match what's on the button and additional information can be provided after a comma. `Add to cart, <product name>` This is a work around that emulates the proper behaviour. 

If not in use remove the matching ID's from : `product-list-item` and `product-grid-item`. Please make sure we are not leaving any other dead code behind.

Pio-Bar (changes requested):
I got no answer last time so, I'll repeat for visibility.
Should we also scan the "write a review" form and the "image zoom" modal?

feature-libs/cart/base/components/add-to-cart/add-to-cart.component.html
uroslates: Yes I've added aria-label (removed aria-describedby as we mentioned that it seems like a cleaner solution).
I would have to dbl check and remove ids containing '_header' and dbl check those though! Thanks!

projects/storefrontapp-e2e-cypress/cypress/e2e/accessibility/product-details-page.a11y-e2e.cy.ts
uroslates: Applied.

projects/storefrontapp-e2e-cypress/cypress/e2e/accessibility/product-details-page.a11y-e2e.cy.ts
uroslates: Applied.

projects/storefrontapp-e2e-cypress/cypress/e2e/accessibility/product-details-page.a11y-e2e.cy.ts
uroslates: Applied!

feature-libs/cart/base/components/add-to-cart/add-to-cart.component.html
uroslates: I have removed the id from those 2 places product-list-item and product-grid-item. 
I'm a bit confused why is the same type of concatenated id also used on cart-item-list-row.component and wish-list-item.component title links!?

feature-libs/cart/base/components/add-to-cart/add-to-cart.component.html
```diff
@@ -43,7 +43,11 @@
...
+    [attr.aria-label]="
+      product?.name
+        ? product?.name + ' ' + ('addToCart.addToCart' | cxTranslate)
+        : ('addToCart.addToCart' | cxTranslate)
+    "
```
Pio-Bar:
```suggestion
    [attr.aria-label]="
      product?.name
        ? ('addToCart.addToCart' | cxTranslate) + ', ' + product?.name 
        : ('addToCart.addToCart' | cxTranslate)
    "
```

Text displayed in the screen should come first, additional information can be provided after a comma.

projects/storefrontapp-e2e-cypress/cypress/e2e/accessibility/product-details-page.a11y-e2e.cy.ts
```diff
@@ -0,0 +1,50 @@
...
+    cy.get(
+      'cx-tab-panel [role="tabpanel"].active cx-product-details-tab .container'
+    );
+    cy.get('main').a11yRunContinuumTest();
+  });
+
+  it('PDP with Reviews', () => {
+    cy.get('cx-page-slot[position="Tabs"] .tab button:nth-child(3)').click();
+    cy.get(
+      'cx-tab-panel [role="tabpanel"].active cx-product-reviews .container .header'
```
Pio-Bar: Does grabbing the header here serve a purpose?

projects/storefrontapp-e2e-cypress/cypress/e2e/accessibility/product-details-page.a11y-e2e.cy.ts
uroslates: >> Should we also scan the "write a review" form and the "image zoom" modal?
Added!

projects/storefrontapp-e2e-cypress/cypress/e2e/accessibility/product-details-page.a11y-e2e.cy.ts
```diff
@@ -0,0 +1,69 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import * as siteContextSelector from '../../helpers/site-context-selector';
+
+/**
+ * This test checks accessibility concerns on the PDP page using Access Continuum
+ */
+describe('Product Details Page Accessibility', { testIsolation: false }, () => {
+  beforeEach(() => {
```
Pio-Bar:
```suggestion
  before(() => {
```
Not navigating each time will make the test run a lot faster.

projects/storefrontapp-e2e-cypress/cypress/e2e/accessibility/product-details-page.a11y-e2e.cy.ts
```diff
@@ -0,0 +1,69 @@
...
+    cy.get('cx-page-slot[position="Tabs"] cx-tab').a11yRunContinuumTest();
+  });
+
+  it('PDP with Shipping', () => {
+    cy.get('cx-page-slot[position="Tabs"] .tab button:nth-child(4)').click();
+    cy.get('cx-tab-panel [role="tabpanel"].active cx-paragraph');
+    cy.get('cx-page-slot[position="Tabs"] cx-tab').a11yRunContinuumTest();
+  });
+
+  it('PDP write a Review Form', () => {
```
Pio-Bar: Since test isolation is off, If we put this test after the review tab we won't need to navigate to it again.

Pio-Bar (changes requested):
I think we should switch to testing a product that has more than one image otherwise we are not testing the UI at it's fullest.

feature-libs/cart/base/components/add-to-cart/add-to-cart.component.html
uroslates: 👍 Applied!

projects/storefrontapp-e2e-cypress/cypress/e2e/accessibility/product-details-page.a11y-e2e.cy.ts
uroslates: 👍 Removed!

projects/storefrontapp-e2e-cypress/cypress/e2e/accessibility/product-details-page.a11y-e2e.cy.ts
uroslates: Applied 👍

projects/storefrontapp-e2e-cypress/cypress/e2e/accessibility/product-details-page.a11y-e2e.cy.ts
uroslates: 👍 Applied!

uroslates:
>> I think we should switch to testing a product that has more than one image otherwise we are not testing the UI at it's fullest.

True applied and found concerns for the image carousel navigation buttons having:
`[attr.aria-selected]="i === activeSlide"`
`Accessibility Concern: The aria-selected attribute is not allowed on this BUTTON`

Fixing it.

Pio-Bar (approved):

Zeyber (approved):

PR #20108, opened 2025-03-21
Keep storage cleaned when Punchout is unused

FollowTheFlo:


integration-libs/punchout/root/services/punchout-state-persistence.service.ts
```diff
@@ -38,12 +39,19 @@ export class PunchoutStatePersistanceService implements OnDestroy {
    * Gets and transforms state into the form that should
    * be saved in storage.
    */
-  protected getPunchoutSessionId(): Observable<PunchoutState> {
+  protected getPunchoutSessionId(): Observable<string | undefined> {
```
FollowTheFlo, line 42: As we only have 1 string to stored.
I thought it is better to store the string directly.
before --> {punchoutSessionId:'abcd'}
now --> 'abcd'

Matejk00 (approved):
LGTM

PR #20109, opened 2025-03-21
Add Cart Loading feature

FollowTheFlo:
Added:
- Load cart from cartId given by occ api
- PunchoutCartInterceptor interceptor to add punchoutsId key/value in header for request with url : carts/{cartId}, if not there server returns error.
- Added routing logic: when EDIT punchoutOperation and no ItemId -> route user to Cart Page.


rmch91 (approved):
LGTM

PR #20116, opened 2025-03-24
fix: (CXSPA-9641) - Header and footer Continuum tests

Pio-Bar:
Closes: [CXSPA-9641](https://jira.tools.sap/browse/CXSPA-9641)

Includes a Cypress command that introduces scoped a11y testing.


projects/storefrontapp-e2e-cypress/cypress/support/continuum.commands.ts
```diff
@@ -89,27 +89,6 @@ const a11yContinuumSetup = withContinuum(
   }
 );
 
-// We verify Access Engine is loaded, loading it again only if necessary, before running our accessibility tests using `Continuum.runAllTests`
-const a11yContinuumRunAllTests = withContinuum((includeiframe = false) => {
```
sdrozdsap, line 96: I guess we have 2 options:
**1. Remove this method.**
In this case we also need to update `continuum.example-e2e.cy.ts`, probably remove `a11y-continuum.utils.ts`(and update `cart.a11y-e2e.cy.ts` which uses the util) and update documentation in `a11y-testing.md`. 
**2. Mark this as deprecated**
In this case all of the above can be done in the next iteration PR

I don't have strong preference here, but in either of this options I guess we need to make adjustments to your current PR. So if you feel like you can also include changes I mentioned in **1.** I think that would be nice, otherwise you can revert deletion and add `@deprecated` mark

sdrozdsap (changes requested):
Generally - beautifully done! Just one small issue I think we need to address

projects/storefrontapp-e2e-cypress/cypress/support/continuum.commands.ts
Pio-Bar: I went with the first one 😉. The docs and the example have been updated to reflect the changes. I've additionally tidied up the leftover commands.

sdrozdsap (approved):
Really nice!

projects/storefrontapp-e2e-cypress/cypress/e2e/accessibility/header-and-footer.a11y-e2e.cy.ts
```diff
@@ -0,0 +1,42 @@
...
+    it('Main header body', () => {
+      cy.get('@header').get('a').contains('Brands');
+      cy.get('@header').a11yRunContinuumTest();
+    });
+
+    it('My Account dropdown', () => {
+      cy.requireLoggedIn(standardUser);
+      cy.reload();
+      cy.get('@header')
+        .get('nav[aria-label="My Account"]')
```
Zeyber, line 26: I'm not sure we are actually testing the dropdown like this. I think we need to open the dropdown first.

projects/storefrontapp-e2e-cypress/cypress/e2e/accessibility/header-and-footer.a11y-e2e.cy.ts
```diff
@@ -0,0 +1,42 @@
...
+describe('Header and Footer Continuum tests', () => {
+  beforeEach(() => {
+    cy.a11yContinuumSetup();
+    cy.visit('/');
+    cy.get('[section="header"]').as('header');
+    cy.get('[section="footer"]').as('footer');
+  });
+  context('Header', () => {
+    it('Main header body', () => {
+      cy.get('@header').get('a').contains('Brands');
```
Zeyber, line 18: We could additionally test when we open a category dropdown. Ideally a more complex one.

Zeyber (changes requested):

projects/storefrontapp-e2e-cypress/cypress/e2e/accessibility/header-and-footer.a11y-e2e.cy.ts
Pio-Bar: In this case the dropdown is fully rendered just hidden from the user with CSS. This means it not necessary to click anything since the structure is already there. 

This brings a good point though. Anything that uses structural directives will need to be interacted with. I have briefly mentioned this in the doc.

projects/storefrontapp-e2e-cypress/cypress/e2e/accessibility/header-and-footer.a11y-e2e.cy.ts
Pio-Bar: It's the same scenario here. Actually, all categories are getting tested!

Zeyber (approved):

PR #20126, opened 2025-03-27
feat: add Account Settings e2e continuum tets for Email, Password, Personal Details, Notifications Pages

uroslates:
Closes: 
- https://jira.tools.sap/browse/CXSPA-9722
- https://jira.tools.sap/browse/CXSPA-9723
- https://jira.tools.sap/browse/CXSPA-9724
- https://jira.tools.sap/browse/CXSPA-9725
- https://jira.tools.sap/browse/CXSPA-9726


projects/storefrontapp-e2e-cypress/cypress/e2e/accessibility/account-settings-close-account.a11y-e2e.cy.ts
```diff
@@ -0,0 +1,45 @@
...
+import * as updateEmail from '../../helpers/update-email';
+import { isolateTests } from '../../support/utils/test-isolation';
+
+const CLOSE_ACCOUNT_URL = '/my-account/close-account';
+
+/**
+ * This test checks accessibility concerns on the Account Settings Close Account page using Access Continuum
+ */
+describe('Account Settings / Close Account Page Accessibility', () => {
+  isolateTests();
```
Pio-Bar: The test isolation is on on by default, I think this may be unnecessary.

projects/storefrontapp-e2e-cypress/cypress/e2e/accessibility/account-settings-close-account.a11y-e2e.cy.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  before(() => {
+    cy.a11yContinuumSetup();
+  });
+
+  it('initial page load', () => {
+    updateEmail.registerAndLogin();
+    cy.visit(CLOSE_ACCOUNT_URL);
+    cy.get('cx-breadcrumb h1').should('contain', 'Close Account');
```
Pio-Bar:
```suggestion
    cy.get('cx-breadcrumb h1').contains('Close Account');
```
We should not use assertions. Ideally we want failures to happen only if Access Continuum finds issues. Otherwise we are risking false negatives. This goes for all the tests in this PR.

projects/storefrontapp-e2e-cypress/cypress/e2e/accessibility/account-settings-close-account.a11y-e2e.cy.ts
```diff
@@ -0,0 +1,45 @@
...
+    cy.get('cx-close-account-modal .cx-close-account-modal-title').should(
+      'contain',
+      'Confirm Account Closure'
+    );
+    cy.get(
+      'cx-close-account-modal .cx-close-account-modal-footer button.btn-secondary'
+    ).click();
+
+    cy.get('main').a11yRunContinuumTest();
+  });
```
Pio-Bar: Not sure why the modal is opened and then closed? Access Continuum can only see what's part of the DOM while the test is run.
The `main` has already been tested in the previous test we should only target the modal here.

projects/storefrontapp-e2e-cypress/cypress/e2e/accessibility/account-settings-close-account.a11y-e2e.cy.ts
```diff
@@ -0,0 +1,45 @@
...
+ */
+describe('Account Settings / Close Account Page Accessibility', () => {
+  isolateTests();
+
+  before(() => {
+    cy.a11yContinuumSetup();
+  });
+
+  it('initial page load', () => {
+    updateEmail.registerAndLogin();
```
Pio-Bar: There is no point of having Cypress manually go through the forms. We can login with a request using `cy.requireLoggedIn(standardUser);` command. This should make our test run a lot faster. 

If we are not making any changes to the account we should be able to just reuse one.

projects/storefrontapp-e2e-cypress/cypress/e2e/accessibility/account-settings-notification-preference.a11y-e2e.cy.ts
```diff
@@ -0,0 +1,62 @@
...
+  isEmailNotificationChannel: boolean;
+}) {
+  cy.log(
+    `🛒 Updating notification preference ${JSON.stringify({ isEmailNotificationChannel })}!`
+  );
+  const checkbox = cy
+    .get('cx-notification-preference')
+    .find('input.form-check-input');
+  // TODO: Why is check causing error!?
+  return isEmailNotificationChannel ? checkbox.check() : checkbox.uncheck();
```
Pio-Bar: You should not use variables with cypress. Trying to access sth outside of it's queue will usually cause problems. Use aliases instead. [You can read more here](https://docs.cypress.io/app/core-concepts/variables-and-aliases;).

projects/storefrontapp-e2e-cypress/cypress/e2e/accessibility/account-settings-notification-preference.a11y-e2e.cy.ts
```diff
@@ -0,0 +1,62 @@
...
+    cy.get('cx-notification-preference input[type="checkbox"]').should(
+      'not.be.checked'
+    );
+    updateNotificationPreferencesForm({ isEmailNotificationChannel: true });
+    cy.get('cx-notification-preference input[type="checkbox"]').should(
+      'be.checked'
+    );
+
+    cy.get('main').a11yRunContinuumTest();
+  });
```
Pio-Bar: Are there unique elements here compared to the previous test? Looks like we are just scanning the same thing twice.

Pio-Bar (changes requested):
So far only two files run AccessContinuum tests. Let's circle back once the rest is completed and existing comments have been resolved with changes applied throughout. 

You may also want to check out `/storefrontapp-e2e-cypress/docs/a11y-testing.md` for some reasoning behind the comments here. There should also be examples on how to specifically target modals and such.

projects/storefrontapp-e2e-cypress/cypress/e2e/accessibility/account-settings-close-account.a11y-e2e.cy.ts
uroslates: Addressed.

projects/storefrontapp-e2e-cypress/cypress/e2e/accessibility/account-settings-close-account.a11y-e2e.cy.ts
uroslates: Addressed.

projects/storefrontapp-e2e-cypress/cypress/e2e/accessibility/account-settings-close-account.a11y-e2e.cy.ts
uroslates: Addressed.

projects/storefrontapp-e2e-cypress/cypress/e2e/accessibility/account-settings-close-account.a11y-e2e.cy.ts
uroslates: Addressed.

projects/storefrontapp-e2e-cypress/cypress/e2e/accessibility/account-settings-notification-preference.a11y-e2e.cy.ts
uroslates: Addressed.

projects/storefrontapp-e2e-cypress/cypress/e2e/accessibility/account-settings-notification-preference.a11y-e2e.cy.ts
uroslates: Addressed.

projects/storefrontapp-e2e-cypress/cypress/e2e/accessibility/account-settings-close-account.a11y-e2e.cy.ts
```diff
@@ -0,0 +1,32 @@
...
+  });
+
+  it('initial page load', () => {
+    cy.requireLoggedIn(standardUser);
+    cy.visit(CLOSE_ACCOUNT_URL);
+
+    cy.get('main').a11yRunContinuumTest();
+  });
+
+  it('confirm Account Closure Modal', () => {
```
Zeyber: I think we could make this test run faster by disabling test isolation so that we don't need to reload pages..

projects/storefrontapp-e2e-cypress/cypress/e2e/accessibility/account-settings-close-account.a11y-e2e.cy.ts
```diff
@@ -0,0 +1,32 @@
...
+ */
+describe('Account Settings / Close Account Page Accessibility', () => {
+  before(() => {
+    cy.a11yContinuumSetup();
+  });
+
+  it('initial page load', () => {
+    cy.requireLoggedIn(standardUser);
+    cy.visit(CLOSE_ACCOUNT_URL);
+
```
Zeyber, line 28: We should probably use some `cy.get()` for some element here to make sure that the page has loaded before running the check.

projects/storefrontapp-e2e-cypress/cypress/e2e/accessibility/account-settings-personal-details.a11y-e2e.cy.ts
```diff
@@ -0,0 +1,59 @@
...
+ * This test checks accessibility concerns on the Account Settings Personal Details page using Access Continuum
+ */
+describe('Account Settings / Personal Details Page Accessibility', () => {
+  before(() => {
+    cy.a11yContinuumSetup();
+  });
+
+  it('initial page load', () => {
+    cy.requireLoggedIn(standardUser);
+    cy.visit(UPDATE_PROFILE_URL);
```
Zeyber: Doesn't seem to be any check here.

projects/storefrontapp-e2e-cypress/cypress/e2e/accessibility/account-settings-close-account.a11y-e2e.cy.ts
Zeyber: This is relevant for the other test suites too.

projects/storefrontapp-e2e-cypress/cypress/e2e/accessibility/account-settings-close-account.a11y-e2e.cy.ts
Zeyber: This is relevant for the other test suites too.

Zeyber (changes requested):

projects/storefrontapp-e2e-cypress/cypress/e2e/accessibility/account-settings-close-account.a11y-e2e.cy.ts
uroslates: @Zeyber I think its disabled - I have removed explicit isolation method call I had. Is there any other way of doing ii?

projects/storefrontapp-e2e-cypress/cypress/e2e/accessibility/account-settings-close-account.a11y-e2e.cy.ts
uroslates: There is get below - like 23. It would fail in case its not loaded. Or am I missing something?

projects/storefrontapp-e2e-cypress/cypress/e2e/accessibility/account-settings-personal-details.a11y-e2e.cy.ts
uroslates: True, nice catch! I have just added get.

projects/storefrontapp-e2e-cypress/cypress/e2e/accessibility/account-settings-close-account.a11y-e2e.cy.ts
Zeyber: It wouldn't fail because it is possible to run the test on the `main` before inner DOM content has loaded due to the async nature. So you would be testing an empty DOM which of course wouldn't have errors.

You can see an example here where `cy.get('list')` is used to make sure inner content loaded before running the test: https://github.com/SAP/spartacus/commit/c6409049389f66fa200c46a5556eaad8ec557c29#diff-de52f169b851ae64d6fe1f046c41693ce0bda320c48b86ee8737d714793c336aR20

We would need to use this often after a `cy.visit()` or click to load some modal. We won't need to do it after filling out form details as that would already assure DOM elements loaded.

Zeyber (changes requested):

projects/storefrontapp-e2e-cypress/cypress/e2e/accessibility/account-settings-close-account.a11y-e2e.cy.ts
uroslates: OK thanks for the link. I have applied the change to all the tests.

Zeyber (approved):

Zeyber (approved):

Pio-Bar (approved):

PR #20135, opened 2025-04-01
fix: CXSPA-9398:Reset checkbox state before rendering captcha

young-binary:


projects/storefrontlib/shared/components/captcha/mock-captcha/mock-captcha.service.ts
```diff
@@ -76,6 +76,13 @@ export class MockCaptchaService extends CaptchaService {
    */
   renderCaptcha(renderParams: RenderParams): Observable<string> {
     if (renderParams.element instanceof HTMLElement) {
+      // Reset checkbox state before rendering
+      this.checkbox.disabled = false;
+      this.checkbox.checked = false;
```
i53577, line 82: will this value become true when calling `onCheckBoxClicked()`?

projects/storefrontlib/shared/components/captcha/mock-captcha/mock-captcha.service.ts
```diff
@@ -76,6 +76,13 @@ export class MockCaptchaService extends CaptchaService {
...
+      this.label.textContent = "I'm not a robot";
+      this.token = '';
+      this.retVal = new Subject<string>();
```
scarai-sap, line 85: remove the code in line 28 as we always create new for it? but it's not critical.

projects/storefrontlib/shared/components/captcha/mock-captcha/mock-captcha.service.ts
young-binary: it seems the value `checked` won't be set to true when onCheckBoxClicked is called.

I will add the code to change the value.

> will this value become true when calling `onCheckBoxClicked()`?

scarai-sap (dismissed):

i53577 (dismissed):

scarai-sap (approved):
LGTM

Zeyber (approved):

PR #20141, opened 2025-04-01
Punchout Cancel Requisition feature

FollowTheFlo:
added cancelRequisition:boolean is state.
cancel btn set cancelRequisition as true into state.
When occ api punchout requisition call is made, value is taken form state.


rmch91 (approved):
LGTM

PR #20142, opened 2025-04-02
fix: CMS requests issued by a logged-in user have been re-played when logged out

kpawelczak:
closes: [CXSPA-9754](https://jira.tools.sap/browse/CXSPA-9754)


rmch91 (approved):
LGTM

PR #20102, opened 2025-03-20
test: add PDP/PLP add to cart dialog e2e test

uroslates:
Closes: https://jira.tools.sap/browse/CXSPA-9645

PR #20147, opened 2025-04-03
fix: (CXSPA-9649) Continuum E2Es - Order History

StanislavSukhanov:
* introduced a new test suite for checking accessibility concerns for order history
* adressed accessibility concerns found by a test.

* closes: https://jira.tools.sap/browse/CXSPA-9649


feature-libs/cart/base/components/add-to-cart/add-to-cart.component.html
```diff
@@ -43,7 +43,7 @@
     "
     type="submit"
     [disabled]="disabled || quantity <= 0 || quantity > maxQuantity"
-    [attr.aria-describedby]="productCode + '_header'"
+    [attr.aria-describedby]="productCode + '_productDescription'"
```
Pio-Bar: I agree with the name change here but, there are PRs currently making changes to this as well.  ([CXPA-9645](https://github.com/SAP/spartacus/pull/20102/files#diff-6bec5c108e4434baf908a2cde0c1552249b74f4f6d976b7bfca3119b302ca096L46), [CXSPA-9643](https://github.com/SAP/spartacus/pull/20149/files#diff-2d5e09de743d7b53e9f979f9de1c8fdd754560ed9971b3a1f3f8bf9a36b37972R17)) Maybe let's stick to the old one for now to avoid a mismatch?

projects/storefrontapp-e2e-cypress/cypress/e2e/accessibility/order-history.cy.ts
```diff
@@ -0,0 +1,67 @@
...
+  waitForResponse,
+} from '../../helpers/order-history';
+import { viewportContext } from '../../helpers/viewport-context';
+import { product } from '../../sample-data/checkout-flow';
+import { waitForOrderWithConsignmentToBePlacedRequest } from '../../support/utils/order-placed';
+import { isolateTestsBefore } from '../../support/utils/test-isolation';
+
+describe('Order History Page accessibility', { testIsolation: false }, () => {
+  viewportContext(['mobile', 'desktop'], () => {
+    isolateTestsBefore();
```
Pio-Bar: We should probably keep test isolation on.

It should be enough to only test the differences between mobile and desktop. We avoid re-running things if possible.

projects/storefrontapp-e2e-cypress/cypress/e2e/accessibility/order-history.cy.ts
```diff
@@ -0,0 +1,67 @@
...
+        );
+        cy.visit('/my-account/orders');
+        cy.get('.cx-order-history-code > .cx-order-history-value').then(
+          (el) => {
+            const orderNumber = el.text().match(/\d+/)[0];
+            waitForOrderWithConsignmentToBePlacedRequest(orderNumber);
+          }
+        );
+        cy.wait(300);
+      });
```
Pio-Bar: You can use the login command with (`test-user-with-orders@sap.cx.com`) account. 
It already has some orders, this should save us some time on the setup. Just make sure we don't modify any data there since it's used for other tests.

projects/storefrontapp-e2e-cypress/cypress/e2e/accessibility/order-history.cy.ts
Pio-Bar, line 1: We should follow the naming convention for AC tests.

projects/storefrontapp-e2e-cypress/cypress/e2e/accessibility/order-history.cy.ts
```diff
@@ -0,0 +1,67 @@
...
+  clickOnActionLink,
+  doPlaceOrder,
+  interceptAddToCartEndpoint,
+  interceptCartPageEndpoint,
+  orderHistoryTest,
+  verifyActionLinkHasText,
+  waitForResponse,
+} from '../../helpers/order-history';
+import { viewportContext } from '../../helpers/viewport-context';
+import { product } from '../../sample-data/checkout-flow';
```
Pio-Bar: Most of these imports aren't used. Let's remove them.

projects/storefrontapp-e2e-cypress/cypress/e2e/accessibility/order-history.cy.ts
```diff
@@ -0,0 +1,67 @@
...
+    context('Order list', () => {
+      it('Page loaded', () => {
+        cy.get('main').a11yRunContinuumTest();
+      });
+    });
+
+    context('Order details', () => {
+      it('Page loaded', () => {
+        cy.get('.cx-order-history-code > .cx-order-history-value')
+          .should('exist')
```
Pio-Bar: We should not use assertions, ideally we want tests to be failed only by Continuum itself.

projects/storefrontapp-e2e-cypress/cypress/e2e/accessibility/order-history.cy.ts
```diff
@@ -0,0 +1,67 @@
...
+          .first()
+          .click();
+
+        cy.wait(3000);
```
Pio-Bar: We shouldn't wait for arbitrary amounts of time. We can just get() an element that gets loaded instead.

Pio-Bar (changes requested):
Should we also test canceling the order within this file?

projects/storefrontapp-e2e-cypress/cypress/e2e/accessibility/order-history.e2e.cy.ts
```diff
@@ -0,0 +1,41 @@
...
+  viewportContext(['mobile', 'desktop'], () => {
+    before(() => {
+      cy.a11yContinuumSetup();
+      cy.login('test-user-with-orders@sap.cx.com', 'pw4all');
+      cy.visit('/my-account/orders');
+    });
+
+    context('Order list', () => {
+      it('Page loaded', () => {
+        cy.get('main').a11yRunContinuumTest();
```
Pio-Bar: This test will run before the content is rendered. Let's get some element that's being loaded to make Cypress wait long enough.

image

projects/storefrontapp-e2e-cypress/cypress/e2e/accessibility/order-history.e2e.cy.ts
```diff
@@ -0,0 +1,41 @@
...
+describe('Order History Page accessibility', { testIsolation: false }, () => {
+  viewportContext(['mobile', 'desktop'], () => {
+    before(() => {
+      cy.a11yContinuumSetup();
+      cy.login('test-user-with-orders@sap.cx.com', 'pw4all');
+      cy.visit('/my-account/orders');
+    });
+
+    context('Order list', () => {
+      it('Page loaded', () => {
```
Pio-Bar:
```suggestion
      it('Order list', () => {
```

Is there a need for the extra context here?

projects/storefrontapp-e2e-cypress/cypress/e2e/accessibility/order-history.e2e.cy.ts
```diff
@@ -0,0 +1,41 @@
+/*
+ * SPDX-FileCopyrightText: 2025 SAP Spartacus team <spartacus-team@sap.com>
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { viewportContext } from '../../helpers/viewport-context';
+
+describe('Order History Page accessibility', { testIsolation: false }, () => {
+  viewportContext(['mobile', 'desktop'], () => {
```
Pio-Bar:
```suggestion
describe('Order History Page accessibility', { testIsolation: false }, () => {
```

If I'm correct there is no difference between mobile and desktop content. If that's the case no need to test it twice. 

I can see having the test isolation off makes the test much faster, maybe we should encourage this in all our Continuum tests?

Pio-Bar (changes requested):

projects/storefrontapp-e2e-cypress/cypress/e2e/accessibility/order-history.e2e.cy.ts
```diff
@@ -0,0 +1,40 @@
...
+      cy.get('#order-history-table'); // wait until content is loaded
+      cy.get('main').a11yRunContinuumTest();
+    });
+
+    it('Order details', () => {
+      cy.get('.cx-order-history-code > .cx-order-history-value')
+        .contains('00001092')
+        .click();
+
+      cy.url().should('contain', '00001092');
```
Pio-Bar:
```suggestion
```
Additional assertions should be unnecessary. Ideally we want Access Continuum to be the only failure point.

projects/storefrontapp-e2e-cypress/cypress/e2e/accessibility/order-history.e2e.cy.ts
```diff
@@ -0,0 +1,40 @@
...
+      cy.get('.cx-order-details-cards'); // wait until content is loaded
+      cy.get('main').a11yRunContinuumTest();
+    });
+
+    it('Cancel order', () => {
+      cy.get('cx-order-details-actions .btn-secondary')
+        .contains(' Cancel Items ')
+        .click();
+      cy.url().should('contain', 'cancel');
```
Pio-Bar:
```suggestion
```
Additional assertions should be unnecessary. Ideally we want Access Continuum to be the only failure point.

Let's grab some loaded content instead. We should make sure everything is loaded each time.

projects/storefrontapp-e2e-cypress/cypress/e2e/accessibility/order-history.e2e.cy.ts
Pio-Bar, line 1: Use the suffix specified in the a11y-testing doc.`*.a11y-e2e.cy.ts` We are likely to use it in the CI setup later.

projects/storefrontapp-e2e-cypress/cypress/e2e/accessibility/order-history.e2e.cy.ts
```diff
@@ -0,0 +1,40 @@
...
+      cy.get('.AccountPageTemplate').a11yRunContinuumTest();
```
Pio-Bar: This selector returns two elements. Maybe let's try a unique selector? I'm not sure how AC will handle more then one.

Pio-Bar (changes requested):

projects/storefrontapp-e2e-cypress/cypress/e2e/accessibility/order-history.e2e.cy.ts
StanislavSukhanov: done

projects/storefrontapp-e2e-cypress/cypress/e2e/accessibility/order-history.e2e.cy.ts
StanislavSukhanov: yes, it was redundant. Thank you

projects/storefrontapp-e2e-cypress/cypress/e2e/accessibility/order-history.e2e.cy.ts
StanislavSukhanov: yes. It was better to wait for the table. fixed

Pio-Bar (approved):

Zeyber (approved):

PR #20153, opened 2025-04-07
chore: CXSPA-9746 Optimize product-search.core-e2e.cy.ts

i306028:
Merge tests to reduce overall # of tests while keeping the coverage


KateChuen (commented):
I would only change the description of this it "should navigate to the next page and display results" to something more general: "should be able to navigate to other pages and display results" or "should test pagination and display results"

KateChuen (approved):

PR #20155, opened 2025-04-07
Punchout Close Session button

FollowTheFlo:
added styling.
Close Session btn
scenario on EDIT punchout mode, we want to revert cart modification:
- when session starts, Requisition is fetched from server, cxml containing initial cart info is stored in Punchout State
- user clicks 'close session' -> we send this 'cxml containing initial cart' info to ARIBA


integration-libs/punchout/components/punchout-close-session/punchout-close-session.component.ts
```diff
@@ -0,0 +1,39 @@
...
+          ? this.punchoutStoreService.getPunchoutState()
+          : of({ punchoutSessionId: undefined });
+      }),
+      map((punchoutState) => {
+        return !!punchoutState.punchoutSessionId;
+      })
+    );
+
+  clickCloseSessionButton(): void {
+    this.punchoutFacade.closePunchoutSession().subscribe();
```
FollowTheFlo, line 37: this obs runs only once, no need to unsubscribe afterwards.

rmch91 (approved):
LGTM

PR #20161, opened 2025-04-09
feat: nightly e2e test job

andreighervan7:
ticket: [CXSPA-9739](https://jira.tools.sap/browse/CXSPA-9739)


rmch91:
I see a lot of code duplications between `ci.yml` and your workflow, and between `e2e-run-all.sh` and `e2e-cypress.sh`
I would say that it is better to encapsulate logic responsible for running e2e tests in on bash script.
And I'm not sure if we need a separate workflow for this.
What do you think if we will use ci.yaml and e2e-cypress.sh only?

What if we will add a cron job to ci.yml and in e2e-cypress.sh extend if statement with 

`elif [ "${GITHUB_EVENT_NAME}" == "schedule" ]; then ` npm run e2e:run:ci:vendor

We can create separate job in ci.yml something like   `vendor_e2e_tests:` . thanks to this, all tests will run as usual, and additionally we will run vendor tests, only if it is weekly run. We can achieve it by adding `if: ${{ github.event_name == 'schedule' }}` to the job. and inside of this job we can call e2e-cypress.sh with a new parameter  
```
run: |
          ci-scripts/e2e-cypress.sh -s vendor
```
And in bash script we can extend if statement with `elif [ "${GITHUB_EVENT_NAME}" == "schedule" ]; then ` npm run e2e:run:ci:${SUITE}"

and last step will be to extend slack trigger if statement, this one:
 ```
       if: |
          needs.build_conclusion.result == 'failure' &&
          github.event_name == 'push' &&
          github.ref_name == github.event.repository.default_branch
```

by adding event `schedule`

Please consider this as a question and the beginning of a discussion, not as an implementation guide. My suggestion might also have some drawbacks — I haven’t looked deeply into the topic. Feel free to share your thoughts.

File: .github/workflows/ci.yml [hunk 1]
```diff
@@ -333,3 +333,4 @@ jobs:
           needs.build_conclusion.result == 'failure' &&
           github.event_name == 'push' &&
           github.ref_name == github.event.repository.default_branch
+
```
rmch91: we can remove this change to have clear diff

.github/workflows/e2e-vendor.yml
```diff
@@ -0,0 +1,335 @@
+name: E2E vendor tests
+
+on:
+  push:
+  #schedule:
+  #   Runs every day at 10:00 (UTC)
```
rmch91: Don' forget to uncomment cron job

.github/workflows/e2e-vendor.yml
```diff
@@ -0,0 +1,335 @@
...
+    runs-on: ubuntu-latest
+    if: ${{ always() }}
+    steps:
+      - name: Required build failed
+        uses: actions/github-script@v7
+        with:
+          script: |
+            core.setFailed('Build failed')
+        if: |
+          needs.no_retries.result == 'failure' || needs.no_retries.result == 'cancelled' ||
```
rmch91: we don't have `no_retries` job in this workflow

.github/workflows/e2e-vendor.yml
```diff
@@ -0,0 +1,335 @@
...
+                      "value": "buid_history",
+                      "url": "https://github.com/SAP/spartacus/actions/workflows/ci.yml?query=branch%3Adevelop"
+                    }
+                  ]
+                }
+              ]
+            }
+        if: |
+          needs.build_conclusion.result == 'failure' &&
+          github.event_name == 'push' &&
```
rmch91: this slack notification will never be triggered because event_name will be not push in our case

.github/workflows/e2e-vendor.yml
```diff
@@ -0,0 +1,335 @@
...
+          github.ref_name == github.event.repository.default_branch
```
rmch91: this condition I think also redundant

ci-scripts/e2e-cypress.sh
```diff
@@ -126,6 +126,7 @@ if [[ "${SSR}" = true ]]; then
         echo "Running full Cypress end-to-end tests"
         npm run e2e:run:ci:ssr
     fi
+
```
rmch91: revert this change to not include unrelated changes to the PR

ci-scripts/e2e-cypress.sh
```diff
@@ -157,4 +158,4 @@ else
         echo "Running full Cypress end-to-end tests"
         npm run e2e:run:ci"${SUITE}"
     fi
-fi
+fi
```
rmch91, line 193: same as above

package.json
```diff
@@ -71,17 +71,24 @@
...
     "e2e:run:ci:asm": "npm --prefix ./projects/storefrontapp-e2e-cypress run cy:run:ci:asm",
     "e2e:run:ci:ccv2": "npm --prefix ./projects/storefrontapp-e2e-cypress run cy:run:ci:ccv2",
     "e2e:run:ci:ccv2-b2b": "npm --prefix ./projects/storefrontapp-e2e-cypress run cy:run:ci:ccv2-b2b",
     "e2e:run:ci:ccv2-product-configurator": "npm --prefix ./projects/storefrontapp-e2e-cypress run cy:run:ci:ccv2-product-configurator",
     "e2e:run:ci:cdc-b2b": "npm --prefix ./projects/storefrontapp-e2e-cypress run cy:run:ci:cdc-b2b",
+    "e2e:run:ci:cdc-b2b-v2": "npm --prefix ./projects/storefrontapp-e2e-cypress run cy:run:ci:cdc-b2b-v2",
     "e2e:run:ci:segment-refs": "npm --prefix ./projects/storefrontapp-e2e-cypress run cy:run:ci:segment-refs",
+    "e2e:run:ci:segment-refs-v2": "npm --prefix ./projects/storefrontapp-e2e-cypress run cy:run:ci:segment-refs-v2",
     "e2e:run:ci:opps": "npm --prefix ./projects/storefrontapp-e2e-cypress run cy:run:ci:opps",
+    "e2e:run:ci:opps-v2": "npm --prefix ./projects/storefrontapp-e2e-cypress run cy:run:ci:opps-v2",
```
rmch91: I don't think `-v2` is understandable. Looking ot it for the first time, I thought that there are two implementations of module, regular one and modernised v2, like we have in `myaccount`. But it is not the case.
What do you think about `e2e:run:ci:vendor:opps`, `e2e:run:ci:vendor:cdc` etc. And similar approach in cypress package.json

projects/storefrontapp-e2e-cypress/package.json
```diff
@@ -16,26 +16,33 @@
...
-    "cy:run:ci:opps": "cypress run --config-file cypress.config.ci.ts --env API_URL=https://api.cg79x9wuu9-eccommerc1-s5-public.model-t.myhybris.cloud/ --tag \"opps\" --group OPPS --spec \"cypress/e2e/vendor/opps/*.e2e.cy.ts\"",
+    "cy:run:ci:cds": "cypress run --config-file cypress.config.ci.ts --record --key $CYPRESS_KEY --tag \"2011,b2c,all-cds\" --group CDS --parallel --spec \"cypress/e2e/vendor/cds/**/*.core-e2e.cy.ts\"",
+    "cy:run:ci:cdc": "cypress run --config-file cypress.config.ci.ts  --record --env API_URL=https://api.cg79x9wuu9-eccommerc1-s1-public.model-t.myhybris.cloud/ --tag \"2005,cdc\" --parallel --group CDC --spec \"cypress/e2e/vendor/cdc/b2c/*.e2e.cy.ts\"",
+    "cy:run:ci:cdc-v2": "cypress run --config-file cypress.config.ci.ts --record  --key $CYPRESS_KEY --env API_URL=https://api.cg79x9wuu9-eccommerc1-s1-public.model-t.myhybris.cloud/ --tag \"2005,cdc\" --parallel --group CDC --spec \"cypress/e2e/vendor/cdc/b2c/*-e2e.cy.ts\"",
```
rmch91: same comment as above about -v2. I would rather group all scripts that we are using in the workflow under :vendor:

rmch91 (changes requested):

.github/workflows/e2e-vendor.yml
```diff
@@ -0,0 +1,342 @@
+name: E2E vendor tests
```
rmch91, line 1: image
I don't see other tests, like asm, product-configurator.

.github/workflows/e2e-vendor.yml
```diff
@@ -0,0 +1,342 @@
...
+                      "type": "button",
+                      "text": {
+                        "type": "plain_text",
+                        "text": "Failed build",
+                        "emoji": true
+                      },
+                      "value": "failed_build",
+                      "url": "${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}"
+                    },
+                    {
```
rmch91, line 161: Work please on slack message. It is not relevant. 
We don't need a link to "What was committed". There are no commits in our case.

I thinks slack message "Broken build on feature/CXSPA-9739" Also could be more informative. For example "Vendor E2E Tests Failed".

.github/workflows/e2e-vendor.yml
```diff
@@ -0,0 +1,342 @@
...
+                      "type": "button",
+                      "text": {
+                        "type": "plain_text",
+                        "text": "Build history",
+                        "emoji": true
+                      },
+                      "value": "buid_history",
+                      "url": "https://github.com/SAP/spartacus/actions/workflows/ci.yml?query=branch%3Adevelop"
```
rmch91: You should link proper file

projects/storefrontapp-e2e-cypress/package.json
```diff
@@ -16,26 +16,33 @@
...
-    "cy:run:ci:opps": "cypress run --config-file cypress.config.ci.ts --env API_URL=https://api.cg79x9wuu9-eccommerc1-s5-public.model-t.myhybris.cloud/ --tag \"opps\" --group OPPS --spec \"cypress/e2e/vendor/opps/*.e2e.cy.ts\"",
+    "cy:run:ci:cds": "cypress run --config-file cypress.config.ci.ts --record --key $CYPRESS_KEY --tag \"2011,b2c,all-cds\" --group CDS --parallel --spec \"cypress/e2e/vendor/cds/**/*.core-e2e.cy.ts\"",
+    "cy:run:ci:cdc": "cypress run --config-file cypress.config.ci.ts  --record --env API_URL=https://api.cg79x9wuu9-eccommerc1-s1-public.model-t.myhybris.cloud/ --tag \"2005,cdc\" --parallel --group CDC --spec \"cypress/e2e/vendor/cdc/b2c/*.e2e.cy.ts\"",
+    "cy:run:ci:vendor:cdc": "cypress run --config-file cypress.config.ci.ts --record  --key $CYPRESS_KEY --env API_URL=https://api.cg79x9wuu9-eccommerc1-s1-public.model-t.myhybris.cloud/ --tag \"2005,cdc\" --parallel --group CDC --spec \"cypress/e2e/vendor/cdc/b2c/*-e2e.cy.ts\"",
```
rmch91, line 34: When we are creating separate set of scripts, :vendor:, I think we should revert changes from the original scripts like cy:run:ci:cdc, because we don't know what code owners do with their initial scripts.

rmch91 (changes requested):

.github/workflows/e2e-vendor.yml
```diff
@@ -0,0 +1,395 @@
...
+   # - cron: "0 10 * * *"
+  workflow_dispatch:
+
+env:
+  CYPRESS_KEY: ${{ secrets.CYPRESS_KEY }}
+  NODE_VERSION: '20'
+  CONTINUUM_REGISTRY_TOKEN: ${{ secrets.CONTINUUM_REGISTRY_TOKEN }}
+
+jobs:
+  no_retries:
```
rmch91: I don't think we need no_retries job here.

.github/workflows/e2e-vendor.yml
rmch91: You don't fix this comment

.github/workflows/e2e-vendor.yml
```diff
@@ -247,8 +248,58 @@ jobs:
           BUILD_NUMBER: ci-build-number-${{ github.event.pull_request.head.sha || github.run_id }}
         run: |
           ci-scripts/e2e-cypress.sh -s cds
+   asm_e2e_tests:
```
rmch91: This line is underlined on my local, problem with indents. Isn't it highlighted on your?

.github/workflows/e2e-vendor.yml
```diff
@@ -247,8 +248,58 @@ jobs:
...
+          restore-keys: nodemodules-${{ github.event.pull_request.base.sha }}
+      - name: Run e2es
+        env:
+          SPA_ENV: b2c
+          BUILD_NUMBER: ci-build-number-${{ github.event.pull_request.head.sha || github.run_id }}
+        run: |
+          ci-scripts/e2e-cypress.sh -s product-configuration
   build_conclusion:
-    needs: [cdc_b2c_e2e_tests, cdc_b2b_e2e_tests, cdp_b2c_e2e_tests, segment_refs_e2e_tests, digital_payments_e2e_tests, opps_e2e_tests, s4_service_e2e_tests, epd_visualization_e2e_tests, cds_e2e_tests]
+    needs: [cdc_b2c_e2e_tests, cdc_b2b_e2e_tests, cdp_b2c_e2e_tests, segment_refs_e2e_tests, digital_payments_e2e_tests, opps_e2e_tests, s4_service_e2e_tests, epd_visualization_e2e_tests, cds_e2e_tests, asm_e2e_tests, product_configurator_e2e_tests]
```
rmch91: Missing libs from `vendor` folder:

1. cpq
2. estimated-delivery-date
3. omf
4. opf
5. pdf-invoices
6. pickup-in-store
7. requested-delivery-date
8. s4om

.github/workflows/e2e-vendor.yml
```diff
@@ -294,7 +347,7 @@ jobs:
                   "type": "section",
                   "text": {
                     "type": "mrkdwn",
-                    "text": ":nuclear-bomb: :fireduck: *Broken build on ${{ github.ref_name }}*"
+                    "text": ":nuclear-bomb: :fireduck: *Vendor E2E Tests Failed on ${{ github.ref_name }}*"
```
rmch91: I don't think we need branch name in the message. It will be triggered always on develop

rmch91 (changes requested):

.github/workflows/e2e-vendor.yml
```diff
@@ -0,0 +1,592 @@
+name: E2E vendor tests
```
rmch91, line 1: It seems like we are running everything what we need now.
Do you think we can refactor this workflow somehow? Do you see any fields for improvements?

rmch91 (changes requested):

.github/workflows/e2e-vendor.yml
andreighervan7: for refactoring, i tried to put everything in one job, but didn't work

rmch91: 1. You have a lot of unrelated changes on your PR. Something with prettier? Fix this please, leave only changes related to your task
2. Please merge latest develop to you branch and resolve conflicts
3. About refactoring, how did you try to achieve this? Did you try to do this with [Matrix](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow) maybe?

andreighervan7: yes i tried with matrix

rmch91: Can you share any commit where I can check how you did it? 
What specific issues did you encounter? What exactly isn't working?

andreighervan7: https://github.com/SAP/spartacus/pull/20161/commits/32895a735ae088af61d453536f0814d1c6e598a7

rmch91: What specific issues did you encounter? What exactly isn't working?

rmch91: Ok, I checked it on commit you provided. 
It appears that by default github cancel all jobs in matrix when one of them failed.
But i think there is a possibility to change the default behaviour. Can you check [this](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow#handling-failures) section in the docs please and try in your solution?

andreighervan7: ok, thanks

.github/workflows/e2e-vendor.yml
```diff
@@ -0,0 +1,169 @@
...
+jobs:
+  vendor_e2e_tests:
+    name: E2E tests for vendor
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - suite: vendor:cdc
+            spa_env: cdc,b2c
```
rmch91, line 23: image
The name of specific test is not readable as it could be. My propose is to add additional parameter to each matrix element, for example:
```shell
- test_name: E2E tests for S4 Service
spa_env: s4-service,b2b
suite: vendor:s4-service
``` 

And to use it in 
`- name: Run ${{ matrix.test_name }}` instead of `Run E2E suite ${{ matrix.suite }}`.

.github/workflows/e2e-vendor.yml
```diff
@@ -0,0 +1,169 @@
+name: E2E tests
```
rmch91: Please make the name more specific. This is a workflow name, and it should clearly indicate what it does when viewed in the list of workflow names.

.github/workflows/e2e-vendor.yml
rmch91: and in this part also use Run ${{ matrix.test_name }
```
vendor_e2e_tests:
    name: E2E tests for vendor
```

rmch91 (changes requested):

PR #20169, opened 2025-04-11
feat: add punchout navigation guard

FollowTheFlo:


integration-libs/punchout/root/model/punchout.model.ts
```diff
@@ -56,3 +56,16 @@ export interface PunchoutState {
...
+
+export type PunchoutNavigationGuardConfig = {
+  [key in PunchOutOperation]: {
+    urls?: string[];
+    cxRoutes?: string[];
+    homePage?: boolean;
+  };
+};
+
+export interface punchoutStorageInfo {
```
Platonn, line 68: [NITPICK]
For consistency, let's keep interface names PascalCase

integration-libs/punchout/root/services/punchout-state-persistence.service.ts
```diff
@@ -20,51 +23,68 @@ export class PunchoutStatePersistanceService implements OnDestroy {
...
-  protected getPunchoutSessionId(): Observable<string | undefined> {
+  protected getPunchoutStorageInfo(): Observable<
+    punchoutStorageInfo | undefined
+  > {
     return this.punchoutStoreService.getPunchoutState().pipe(
       map((punchoutState) => {
-        if (punchoutState?.punchoutSessionId) {
+        if (
+          punchoutState?.punchoutSessionId &&
+          punchoutState?.punchoutSession?.punchOutOperation
```
Platonn, line 50: [NITPICK]
`punchout` vs `punchOut`
For consistency, lets keep only one variant throughout the whole epic feature - `punchout` as one word, or `punchOut`/`PunchOut`/`punch-out`/`PUNCH_OUT` as 2 words

integration-libs/punchout/guards/punchout-navigation.guard.ts
```diff
@@ -0,0 +1,126 @@
...
+        urls: [...this.allowedUrlsForAll, ...this.allowCxRoutesForEdit],
+        cxRoutes: [...this.allowCxRoutesForEdit],
+        homePage: true,
+      },
+      [PunchOutOperation.CREATE]: {
+        urls: [...this.allowedUrlsForAll, ...this.allowCxRoutesForEdit],
+        cxRoutes: [...this.allowCxRoutesForEdit],
+        homePage: true,
+      },
+    };
```
Platonn, line 59: [MEDIUM]
The property names of `PunchoutNavigationGuardConfig` are quite confusing to me.
- `urls` - does it mean allowedUrls?
- `cxRoutes` - does it mean allowedCxRoutes
- `homePage: boolean` - what's the intention/meaning behind it? Did you consider some alternative more descriptive name? (I'm just guessing, I might be totally wrong: shouldRedirectToHomepage? isHomePageAllowed? something else?

integration-libs/punchout/root/model/punchout.model.ts
```diff
@@ -9,6 +9,7 @@ export const PUNCHOUT_ERROR_PAGE_URL = '/punchout/cxml/error';
 export const PUNCHOUT_SESSION_ID = 'punchoutSessionId';
 export const PUNCHOUT_SESSION_PAGE_URL = '/punchout/cxml/session';
 export const PUNCHOUT_REQUISITION_PAGE_URL = '/punchout/cxml/requisition';
+export const PUNCHOUT_INSPECT_PAGE_URL = '/punchout/cxml/inspect';
```
Platonn, line 12: [QUESTION]
1. I'm sorry I forgot, what is the `Inspect` page in the punchout user flow? I'm just curious.
2. Why we added it only in this PR? Was it a direct result of adding an Angular Guard? or it was needed anyway to be added, and we just added it in this PR by the way?

integration-libs/punchout/root/model/punchout.model.ts
```diff
@@ -33,11 +33,11 @@ export interface PunchoutSessionInput {
...
-  customerId: string;
-  cartId: string;
-  punchOutLevel: PunchOutLevel;
-  punchOutOperation: PunchOutOperation;
-  selectedItem: string;
+  customerId?: string;
+  cartId?: string;
+  punchOutLevel?: PunchOutLevel;
+  punchOutOperation?: PunchOutOperation;
+  selectedItem?: string;
```
Platonn, line 40: [QUESTION]
Why we need to change all those properties to optional in this PR? Is it really related to adding an Angular Guard? Or it was needed anyway and we just added it in this PR by the way?

integration-libs/punchout/guards/punchout-navigation.guard.ts
```diff
@@ -0,0 +1,126 @@
...
+  ) {
+    if (!cxRoutes) {
+      return false;
+    }
+    return !!route.data['cxRoute'] && cxRoutes.includes(route.data['cxRoute']);
+  }
+
+  protected isPunchoutSessionActive(): Observable<
+    PunchOutOperation | undefined
+  > {
```
Platonn, line 114: [NITPICK]
The method name is a bit confusing - it suggests it returns a boolean `is..`, but it returns a punchout operation.

integration-libs/punchout/root/model/punchout.model.ts
```diff
@@ -24,7 +25,6 @@ export enum PunchOutOperation {
   CREATE = 'CREATE',
   EDIT = 'EDIT',
   INSPECT = 'INSPECT',
-  SOURCE = 'SOURCE',
```
Platonn: [QUESTION]
Why we removed `SOURCE` punchout operation type in this PR? Was it a direct result of adding an Angular Guard? or it was needed anyway to be added, and we just removed it in this PR by the way?

integration-libs/punchout/root/services/punchout-state-persistence.service.ts
```diff
@@ -20,51 +23,68 @@ export class PunchoutStatePersistanceService implements OnDestroy {
...
       punchoutSessionId &&
+      punchOutOperation &&
       !this.punchoutDetectionService.isPunchoutSessionPage()
     ) {
-      this.punchoutStoreService.setPunchoutState({ punchoutSessionId });
+      this.punchoutStoreService.setPunchoutState({
+        punchoutSessionId,
+        punchoutSession: {
+          punchOutOperation: punchOutOperation as PunchOutOperation,
+        },
```
Platonn, line 86: [QUESTION]
Why we need store punchout operation in local storage? Is it really related to adding an Angular Guard? Or it was needed anyway and we just added it in this PR by the way?

integration-libs/punchout/guards/punchout-navigation.guard.ts
```diff
@@ -0,0 +1,126 @@
...
+  ];
+  protected readonly allowCxRoutesForEdit: string[] = [
+    'category',
+    'quickOrder',
+    'product',
+    'cart',
+  ];
+  protected readonly allowedUrlsForInspect: string[] = [
+    PUNCHOUT_INSPECT_PAGE_URL,
+  ];
```
Platonn, line 42: [NITPICK]
`allowed...` vs `alllow...` vs `allowed...`

For consistency, shall we rename `allowCxRoutesForEdit` to `allowedCxRoutesForEdit`,

integration-libs/punchout/root/model/punchout.model.ts
FollowTheFlo: 1. Inspect Cart page, the cart is read-only, user can only see the PunchoutInspectcart page, this page has not been implemented yet (CXSPA-8597 | Punchout INSPECT CART mode). On ARIBA, the cart is read-only because it is on pending-approval status, user still have the right to punch-in to inspect his/her cart.
2. To anticipate, so the full logic is here, covers all use cases.

integration-libs/punchout/root/model/punchout.model.ts
FollowTheFlo: Accelerator did not support this operation, neither us.
https://github.tools.sap/cx-commerce/subscriptionbundles/blob/develop/b2bpunchout/resources/b2bpunchout-beans.xml#L22
```
<enum class="de.hybris.platform.b2b.punchout.constants.PunchOutSetupOperation">
		<description>Supported PunchOut operations</description>
		<value>CREATE</value>
		<value>EDIT</value>
		<value>INSPECT</value>
	</enum>
```
	It is used to initiate a quote with a sourcing application. More info on
	https://xml.cxml.org/current/cXMLReferenceGuide.pdf page 79

FollowTheFlo: i removed it as TypeScripot was enforcing to define SOURCE in below config:
protected readonly punchoutNavigationGuardConfig: PunchoutNavigationGuardConfig

As the type enforce to enumerate all punchoutOperation
```
export type PunchoutNavigationGuardConfig = {
  [key in PunchOutOperation]: {
    urls?: string[];
    cxRoutes?: string[];
    homePage?: boolean;
  };
};
```
I believe it avoid confusion, as we won't use SOURCE operation, i just removed it.

integration-libs/punchout/root/services/punchout-state-persistence.service.ts
FollowTheFlo: punchOutOperation property is returned by cx-server (punchout session occ api) , it is to have the exact match.
I personally prefer to keep 'punchout' everywhere. I had to do an exception for punchOutLevel and punchOutOperation

integration-libs/punchout/root/services/punchout-state-persistence.service.ts
FollowTheFlo: to get the punchoutOperation value, we have to call the punchout Session Api and wait for its response.
It was for performance matter. It avoid calling a facade method.

integration-libs/punchout/root/model/punchout.model.ts
FollowTheFlo: I added punchOutOperation value in local storage. 
When punchout info value is read from local storage,  punchout state is updated and contains only 2 values at this stage (when app starts and reads local storage): punchoutSessionId and punchoutSession. 
punchOutOperation, punchoutSession obj need to have the other properties as optional to be able to be partial.

integration-libs/punchout/guards/punchout-navigation.guard.ts
FollowTheFlo: yes, i couldn't find a proper name. At first it was called getPunchoutOpeation, but it is a bit more than that as it checks user login status too.

integration-libs/punchout/guards/punchout-navigation.guard.ts
FollowTheFlo: thank I'll add
allowedUrls
allowedCxRoutes
allowHomePage

The redirection is not done yet, at the mom it only block the page without any message.

PR #20184, opened 2025-04-16
feat: navigation restriction with guard using occ request

FollowTheFlo:


integration-libs/punchout/root/facade/punchout.facade.ts
```diff
@@ -33,7 +34,8 @@ export function punchoutFacadeFactory() {
 })
 export abstract class PunchoutFacade {
   /**
-   * Abstract method used to get Punchout Session
+   * Abstract method used to get Punchout Session, worflow:
+   * occ api request, logout previous user, login punchout user, load cart
    * @param sessionId is the sesssion Id given by ARIBA via url param
    */
   abstract getPunchoutSession(
```
FollowTheFlo, line 41: planning to rename as 'initPunchoutSession' as it does much more than requesting server punchout session.

integration-libs/punchout/core/facade/punchout.service.ts
```diff
@@ -49,6 +49,18 @@ export class PunchoutService implements PunchoutFacade {
   protected multiCartFacade = inject(MultiCartFacade);
   protected userIdService = inject(UserIdService);
 
+  /**
+   * punchoutSession ajax request will always return same response during a punchout session.
+   * To avoid multiple server calls, input (punchoutSessionId) and output (punchoutSession) values are stored into punchoutSessionRequest.
+   */
+  protected readonly punchoutSessionRequest: {
```
FollowTheFlo: could be rename as punchoutSessionRequestSnapshot, cachedPunchoutSessionRequest or punchoutSessionRequestState if not explicit enough.

integration-libs/punchout/core/facade/punchout.service.ts
```diff
@@ -67,55 +79,44 @@ export class PunchoutService implements PunchoutFacade {
       this.displayErrorPage();
       return throwError(() => new Error('Punchout Session Id missing'));
     }
-    return this.punchoutConnector
-      .getPunchoutSession(payload.punchoutSessionId)
-      .pipe(
-        map((punchoutSession) => {
+
+    return this.requestPunchoutSession(payload.punchoutSessionId).pipe(
+      switchMap((punchouSession) => {
```
Platonn: typo `punchou` vs `punchout`

integration-libs/punchout/core/facade/punchout.service.ts
Platonn: I like the word "cached" in this context with this explanatory JSDocs

Do we cache the request payload? Or the response contents? 
If the response, what about "cachedPunchoutSession**Response**"?
or just "cachedPunchoutSession", if it makes more sense to you?

integration-libs/punchout/root/guards/punchout-navigation.guard.ts
```diff
@@ -0,0 +1,181 @@
...
+  protected readonly punchoutNavigationGuardConfig: PunchoutNavigationGuardConfig =
+    {
+      [PunchOutOperation.INSPECT]: {
+        allowedUrls: [...this.allowedUrlsForAll, ...this.allowedUrlsForInspect],
+        redirectPage: PUNCHOUT_INSPECT_PAGE_URL,
+      },
+      [PunchOutOperation.EDIT]: {
+        allowedUrls: [
+          ...this.allowedUrlsForAll,
+          ...this.allowedCxRoutesForEdit,
```
Platonn: [QUESTION]
Should `allowedUrls` really contain `this.allowedCxRoutesForEdit`? I understood that we should treat separately Urls and cxRoutes.

Note: analogical comment applies to lines 80-82:
```
     [PunchOutOperation.CREATE]: {
        allowedUrls: [
          ...this.allowedUrlsForAll,
          ...this.allowedCxRoutesForEdit,
```

integration-libs/punchout/root/guards/punchout-navigation.guard.ts
```diff
@@ -0,0 +1,181 @@
...
+            this.punchoutNavigationGuardConfig[punchoutOperation]
+              .allowedCxRoutes
+          ) ||
+          this.findByUrls(
+            route,
+            this.punchoutNavigationGuardConfig[punchoutOperation].allowedUrls
+          ) ||
+          (!!this.punchoutNavigationGuardConfig[punchoutOperation]
+            ?.allowHomePage &&
+            this.findHomePage(route));
```
Platonn: [NITPICK]
I'm having hard time understanding what this method does `findByCxRoute` from it's name. Does it return an element of an array?
After analyzing the whole implementation, I guess this method answers the question, whether the given operation is allowed for some cxRoute.

 What about simplifying the methods signatures and renaming them so the condition is simplified as below:
```
        const canActivate =
          !punchoutOperation ||
          this.isAllowedCxRoute(route, punchoutOperation) ||
          this.isAllowedUrl(route, punchoutOperation)
```
(and moving the access to `this.punchoutNavigationGuardConfig` into internals/implementation of those new methods `this.isAllowedCxRoute()` and `this.isAllowedUrl())

btw. to simplify logic even further, I guess we can simply put the `''` (homepage url) into the array of relevant allowedUrls (instead of creating a special property for `allowHomePage:boolean`)?

integration-libs/punchout/root/guards/punchout-navigation.guard.ts
FollowTheFlo: yep, .`..this.allowedCxRoutesForEdit` should be removed from allowedUrls.

integration-libs/punchout/core/facade/punchout.service.ts
FollowTheFlo: I realized we don't need to cache the sessionId (payload) since there is no scenario where sessionId would be different during a user punchout instance without leaving spartacus.
so now only caching the response:
` protected cachedPunchoutSessionResponse?: PunchoutSession;`

integration-libs/punchout/root/facade/punchout.facade.ts
FollowTheFlo: renaming tasks covered in 
CXSPA-9873 | Improve source code naming

integration-libs/punchout/core/facade/punchout.service.ts
```diff
@@ -67,55 +73,44 @@ export class PunchoutService implements PunchoutFacade {
...
+          logout: this.punchoutAuthService.logout(),
+        });
+      }),
+      map(({ punchoutSession }) => {
+        if (punchoutSession?.token?.accessToken) {
+          this.punchoutAuthService.loginWithToken(
+            punchoutSession.token.accessToken,
+            punchoutSession.customerId
+          );
+          this.loadCart(punchoutSession.cartId).subscribe();
```
Pucek9, line 90: I know it's already existing code, but nested subscription seems potentially unpredictable

FollowTheFlo: thanks,
I'm adding this concerns to 
[CXSPA-9871](https://jira.tools.sap/browse/CXSPA-9871)
Token not revoked when silently logout user
this pipe will need to be enhanced.

Pucek9:
Work as expected
1a,1b:
image
2a, 2b:
image

Pucek9 (approved):

PR #20199, opened 2025-04-18
fix: Get Directions button redirects to incorrect page CXSPA-9546

Pucek9:
Closes https://jira.tools.sap/browse/CXSPA-9546


feature-libs/storefinder/core/service/geolocation.service.ts
```diff
@@ -0,0 +1,41 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { Injectable } from '@angular/core';
+import { PointOfService } from '@spartacus/core';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class GeolocationService {
```
Pucek9: Please vote for a name of that service (or give your own proposal):
👎🏻 GeolocationService
👍🏻 StoreLocationService
❤️ StoreGeolocationService
👀 PointOfServiceLocationService
😄 StoreMapLinkService
🚀 StoreCoordinatesService
🎉 LocationUrlService
😕 GoogleMapsUrlService/GoogleMapsDirectionsService

feature-libs/pickup-in-store/components/container/my-preferred-store/my-preferred-store.component.ts
```diff
@@ -70,32 +84,102 @@ export class MyPreferredStoreComponent implements OnInit {
...
+          tap((store: PointOfService) => {
+            this.pointOfService = store;
+          })
+        );
+    }
     useFeatureStyles('a11yViewHoursButtonIconContrast');
     useFeatureStyles('a11yImproveButtonsInCardComponent');
   }
 
   ngOnInit(): void {
```
rmch91, line 114: `Nitpick`: IMO ngOnInit is too much overloaded with different logic, and conditions. 
Do you think it is reasonable to split it to separate methods for better readability?

rmch91 (dismissed):
The rest LGTM. If you will decide to make changes in `ngOnInit`, ping me please for re-approve

feature-libs/pickup-in-store/components/container/my-preferred-store/my-preferred-store.component.ts
Pucek9: I wanted to do as little modification as possible. Most of the code can be removed by cleaning up featureFlags (the code looks big because there are 2 flags causing branches). I didn't notice that we created new methods because of flags, so I avoided that.

rmch91 (approved):

PR #20204, opened 2025-04-22
refactor: removal of `useExtractedBillingAddressComponent` feature toggle

anjana-bl:
https://jira.tools.sap/browse/CXSPA-9866


feature-libs/checkout/base/components/checkout-payment-method/checkout-payment-form/checkout-payment-form.component.ts
```diff
@@ -150,66 +93,15 @@ export class CheckoutPaymentFormComponent implements OnInit {
     protected launchDialogService: LaunchDialogService,
     protected translationService: TranslationService
   ) {}
-  useExtractedBillingAddressComponent: boolean = this.featureConfig.isEnabled(
```
pawelfras, line 153: `medium:` Unfortunately, this field was not marked as deprecated when `useExtractedBillingAddressComponent` feature flag was introduced and I think we cannot simply remove it now as it is part of public API and there is a risk (small, but still) that some customers use it in their custom components. let's mark this field as `@deprecated wth a proper information, like you did [here](https://github.com/SAP/spartacus/pull/20202/files#diff-2fc52ffa14ee4ce9a4bc99a4aa8e626d99f718f76ac534bb3134a829794ca704)

anjana-bl: agreed, done!

pawelfras (approved):

vigneshkrish8691 (approved):

PR #20208, opened 2025-04-22
fix: nx update

npapp-dev002:
updates NX and some dependencies


.eslintrc.json
```diff
@@ -78,6 +78,7 @@
           "error",
           {
             "allow": [],
+            "allowCircularSelfDependency": true,
```
Platonn, line 81: [QUESTION]
Do we really need this option?
What alternative solutions can you see? And what pros and cons of each solution can you see?

integration-libs/cdc/core/auth/facade/cdc-auth.service.spec.ts
```diff
@@ -1,5 +1,6 @@
 import { TestBed } from '@angular/core/testing';
 import { Store, StoreModule } from '@ngrx/store';
+// eslint-disable-next-line @nx/enforce-module-boundaries
```
Platonn: [QUESTION]
Could you help me understand why do we need this line?
My understanding was that a Core part of a lib is allowed to access a Root part of other lib.

Note: analogical comment applies to the line https://github.com/SAP/spartacus/pull/20208/files#diff-0ddfb0a42baefeec1952ffccde64aa129a4e9c4a4b8e44114fba322a39b732bdR9

package.json
```diff
@@ -122,6 +122,11 @@
     "dev:cds:ssr": "env-cmd --no-override -e dev,b2c,cds cross-env NODE_TLS_REJECT_UNAUTHORIZED=0 nx run storefrontapp:serve-ssr",
     "prerender:cds:dev": "env-cmd --no-override -e dev,cds cross-env NODE_TLS_REJECT_UNAUTHORIZED=0 nx run storefrontapp:prerender --routes-file projects/storefrontapp/prerender.txt"
   },
+  "overrides": {
+    "axios": "1.8.4",
+    "eazy-logger": "4.1.0",
+    "http-proxy-middleware": "3.0.4"
+  },
```
Platonn: [QUESTION]
Is force-bumping those transitive dependencies really related to this PR?
Why I ask: I understood that this PR is related to bumping only Nx, with regards to this ticket: https://jira.tools.sap/browse/CXSPA-9425
Perhaps I'm missing something important.

npapp-dev002: You are right. These overrides are not strictly necessary for this PR, and can be provided by other contributions.

integration-libs/cdc/core/auth/facade/cdc-auth.service.spec.ts
npapp-dev002: [ANSWER]
It comes from the misconfiguration of @nx/enforce-module-boundaries. 
image
I tried to turn off the force-module-boundaries plugin for *.spec.ts files but that didn't help. Therefore, this one line solution was chosen.

.eslintrc.json
npapp-dev002: [ANSWER]
That option was turned on to silence the following errors: 
image
Another solution would be to rewrite those imports, but I think that might be outside the scope of this PR. Therefore, I introduced this as temporary solution.

Platonn: Thank you, now I understand.

That said, I believe `allowCircularSelfDependency: false` is a healthy option for the codebase. It prevents in "eslint check time" errors that eventually happen while building libraries (often with ambigious errors) when a source code file (not spec file) import something from itself (the same library) via a symbolic import (.e.g. `from '@spartacus/core'`) while it should have been importing via relative path - within the same library. Otherwise the production library build will fail due to circular dependency.

Note: the production build doesn't fail when such imports happen in spec files. And we never had an issue with it. However I understand that after updating Nx, this eslintrule complains - because we have such circular symbolic imports in spec files.

What do you think about creating a separate ticket (not necessarily you implementing it, but just creating for now) to:
- remove the option `"allowCircularSelfDependency": true,` that was introduced in this PR
- fixing problem in the roots -> importing things via relative imports also in spec files
You might want to mention in the ticket description that removing `allowCircularSelfDependency: true` will help spot the late production build-time issues earlier and easier thanks to eslint hints

integration-libs/cdc/core/auth/facade/cdc-auth.service.spec.ts
Platonn: > I tried to turn off the force-module-boundaries plugin for *.spec.ts files but that didn't help. Therefore, this one line solution was chosen.

I'm sorry, I commented on the spec file of cdc.service. But perhaps I should look at the TS file of the service first. There we also import from `'@spartacus/asm/root';`. IMHO such and import should not be forbidden by eslint (also in TS service file).

It's strange that Eslint reports that the _reason_ of the error is that that ASM library is lazy-loaded in the Spec file _of Schematics(!)_ `integration-libs/cdc/schematics/add-cdc/index_spec.ts`. To my understanding it's not lazy lazy loaded there. Moreover it shouldn't be lazy loaded there. It looks like some misconfiguration issue just for the CDC library.

Because the problem happens only for CDC Spec file, I only guess there might be some misconfiguration problem for the CDC library only (e.g. misconfigured tsconfig, tsconfig.spec, or misconfigured Nx project's tags, etc.)

All that said, I agree it might be time consuming to find the root cause in the CDC library's configuration now and we can descope it to separate ticket.

Would you mind creating a ticket for it and then passing to the Team Fusion (responsible for the CDC library)?
And in the `eslint-ignore` section please add [a comment section](https://eslint.org/docs/latest/use/configure/rules#using-configuration-comments-1) after `--`, e.g.
```
// eslint-disable-next-line @nx/enforce-module-boundaries -- workaround, see <ticket number here
```
(...with the ticket number created and passed to Team Fusion)

package-lock.json
```diff
@@ -7547,1041 +7385,2232 @@
...
+        "npm": ">= 8"
+      },
+      "peerDependencies": {
+        "@angular/common": "^19.0.0",
+        "@angular/core": "^19.0.0",
+        "@angular/forms": "^19.0.0"
+      }
+    },
+    "node_modules/@ngrx/effects": {
+      "version": "19.1.0",
```
Platonn: I can see we still bumped the _used_ version of ngrx in our monorepo to 19.1.0. 
I guess reason might be that on your local you've already bumped the versions in the package-lock and already commited it. Then you reverted just the change of the constraints in the package.json - which didn't revert the changes in the package-lock.
_Perhaps_ one of the solutions might be to start the Nx update from scratch (without touching ngrx at all).

btw. I'm sorry I noticed only now that in `package.json` we're still bumping `"@ngrx/store-devtools": "^19.0.1",`

PS. To avoid similar problems in the future, I've just created a ticket https://jira.tools.sap/browse/CXSPA-9974

.eslintrc.json
npapp-dev002: Thank you very much for your input. 

Ticket was created: [CXSPA-9978](https://jira.tools.sap/browse/CXSPA-9978)

integration-libs/cdc/core/auth/facade/cdc-auth.service.spec.ts
npapp-dev002: Thank you very much.

I created the ticket: [CXSPA-9982](https://jira.tools.sap/browse/CXSPA-9982)

Platonn (approved):

PR #20210, opened 2025-04-23
feat: OPF B2B Checkout support

Matejk00:
Closes: CXSPA-9698


integration-libs/opf/b2b-checkout/components/opf-b2b-checkout-cost-center/opf-b2b-checkout-cost-center.component.html
```diff
@@ -0,0 +1,30 @@
...
+    <div class="col-md-12 col-xl-10">
+      <ng-container *ngIf="costCenters$ | async as costCenters">
+        <div *ngIf="costCenters.length !== 0">
+          <label>
+            <span class="label-content required">{{
+              'checkoutB2B.costCenter' | cxTranslate
+            }}</span>
+            <select
+              #costCenterSelect
+              (change)="setCostCenter(costCenterSelect.value)"
```
ercultimate, line 12: Why use a component reference `costCenterSelect` instead of `$event.target.value`?

Matejk00: In a previous approach I'm getting a linter error: 'Property 'value' does not exist on type 'EventTarget'.
Using a template reference Angular can properly infer the type of the element.

integration-libs/opf/b2b-checkout/components/opf-b2b-checkout-payment-and-review/opf-b2b-checkout-payment-and-review.component.ts
```diff
@@ -0,0 +1,92 @@
...
+})
+export class OpfB2bCheckoutPaymentAndReviewComponent
+  extends OpfCheckoutPaymentAndReviewComponent
+  implements OnInit
+{
+  protected fb = inject(UntypedFormBuilder);
+  protected opfMetadataStoreService = inject(OpfMetadataStoreService);
+  protected cmsService = inject(CmsService);
+  protected checkoutPaymentTypeFacade = inject(CheckoutPaymentTypeFacade);
+  protected opfBaseFacade = inject(OpfBaseFacade);
```
ercultimate: These are injected in the parent class.  Is it necessary to duplicate here?

integration-libs/opf/b2b-checkout/components/opf-b2b-checkout-payment-and-review/opf-b2b-checkout-payment-and-review.component.ts
```diff
@@ -0,0 +1,92 @@
...
+
+  checkoutStepTypePaymentType = CheckoutStepType.PAYMENT_TYPE;
+
+  get paymentType$(): Observable<PaymentType | undefined> {
```
ercultimate: Is this getter required over an assigned property?  A new observable is created on each access.

integration-libs/opf/b2b-checkout/components/opf-b2b-checkout-payment-type/opf-b2b-checkout-payment-type.component.spec.ts
```diff
@@ -0,0 +1,175 @@
...
+  });
+
+  it('should navigate to next step', () => {
+    spyOn(checkoutStepService, 'next');
+    spyOn(checkoutPaymentTypeFacade, 'setPaymentType').and.returnValue(of({}));
+    spyOn(activeCartFacade, 'reloadActiveCart');
+
+    // Set required state
+    component['selectedPaymentOption'] = '1';
+    component['poNumberInputElement'] = {
```
ercultimate: It's not great to work with private properties in tests.

integration-libs/opf/b2b-checkout/components/opf-b2b-checkout-payment-and-review/opf-b2b-checkout-payment-and-review.component.ts
Matejk00: They are not needed I removed them.

integration-libs/opf/b2b-checkout/components/opf-b2b-checkout-payment-and-review/opf-b2b-checkout-payment-and-review.component.ts
Matejk00: You make a good point. it's more efficient to create the Observable once and store it as a property. I've converted the `paymentType$` getter to a property.

integration-libs/opf/b2b-checkout/components/opf-b2b-checkout-payment-type/opf-b2b-checkout-payment-type.component.spec.ts
Matejk00: Right, I made some changes to the test and component logic, now we're using a public method to update the component's state during testing

integration-libs/opf/b2b-checkout/components/opf-b2b-checkout-payment-and-review/opf-b2b-checkout-payment-and-review.component.ts
```diff
@@ -0,0 +1,72 @@
...
+  extends OpfCheckoutPaymentAndReviewComponent
+  implements OnInit
+{
+  checkoutStepTypePaymentType = CheckoutStepType.PAYMENT_TYPE;
+
+  paymentType$ = this.activeCartFacade
+    .getActive()
+    .pipe(map((cart: Cart) => cart.paymentType));
+
+  get poNumber$(): Observable<string | undefined> {
```
ercultimate: Getter or assigned property?

integration-libs/opf/b2b-checkout/components/opf-b2b-checkout-review/opf-b2b-checkout-review.component.ts
```diff
@@ -0,0 +1,145 @@
...
+      .getActive()
+      .pipe(map((cart: Cart) => cart.paymentType));
+  }
+
+  get poNumber$(): Observable<string | undefined> {
+    return this.checkoutPaymentTypeFacade.getPurchaseOrderNumberState().pipe(
+      filter((state) => !state.loading && !state.error),
+      map((state) => state.data)
+    );
+  }
```
ercultimate: Getter or assigned property?

integration-libs/opf/b2b-checkout/components/opf-b2b-checkout-review/opf-b2b-checkout-review.component.ts
```diff
@@ -0,0 +1,145 @@
...
+
+  get costCenter$(): Observable<CostCenter | undefined> {
+    return this.checkoutCostCenterFacade.getCostCenterState().pipe(
+      filter((state) => !state.loading && !state.error),
+      map((state) => state.data)
+    );
+  }
```
ercultimate: Getter or assigned property?

integration-libs/opf/b2b-checkout/components/opf-b2b-checkout-review/opf-b2b-checkout-review.component.ts
```diff
@@ -0,0 +1,145 @@
...
+        return {
+          title: textTitle,
+          textBold: poNumber ? poNumber : noneTextTitle,
+        };
+      })
+    );
+  }
+
+  protected isCmsComponentInPage(cmsComponentUid: string, page: Page): boolean {
+    return !!page && JSON.stringify(page).includes(cmsComponentUid);
```
ercultimate, line 121: Would you explain why we need to use JSON.stringify here?

integration-libs/opf/b2b-checkout/components/opf-b2b-checkout-review/opf-b2b-checkout-review.component.ts
```diff
@@ -0,0 +1,145 @@
...
+  selector: 'cx-opf-b2b-checkout-review',
+  templateUrl: './opf-b2b-checkout-review.component.html',
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  standalone: false,
+})
+export class OpfB2bCheckoutReviewComponent
+  extends CheckoutReviewSubmitComponent
+  implements OnInit
+{
+  protected fb = inject(UntypedFormBuilder);
```
ercultimate: Shouldn't we use a typed form?

integration-libs/opf/checkout/components/opf-checkout-payment-and-review/opf-checkout-payment-and-review.component.ts
```diff
@@ -66,11 +73,14 @@ export class OpfCheckoutPaymentAndReviewComponent
     return Boolean(this.checkoutSubmitForm.get('termsAndConditions')?.value);
   }
 
-  get paymentType$(): Observable<PaymentType | undefined> {
```
ercultimate, line 69: Is this removal a breaking change?

integration-libs/opf/checkout/components/opf-checkout-payments/opf-checkout-payments.component.ts
```diff
@@ -68,6 +93,16 @@ export class OpfCheckoutPaymentsComponent implements OnInit, OnDestroy {
 
   iconTypes = ICON_TYPE;
 
+  paymentChange = output<OpfActiveConfiguration>();
```
ercultimate: Here is a use of `output()`.  Why is this used over an `Output` decorator?

integration-libs/opf/checkout/components/opf-checkout-review-cart-details/opf-checkout-review-cart-details.component.html
```diff
@@ -0,0 +1,29 @@
+<div class="cx-review-cart-total d-none d-lg-block d-xl-block">
+  {{
+    'cartItems.cartTotal' | cxTranslate: { count: cart?.deliveryItemsQuantity }
+  }}:
+  {{ cart?.totalPrice?.formattedValue }}
+</div>
+<div class="cx-review-cart-item col-md-12">
+  <cx-promotions
+    [promotions]="
+      (cart?.appliedOrderPromotions || []).concat(
```
ercultimate, line 10: Since the component is OnPush, this calculation will only be done on cart and/or entry change, right?

integration-libs/opf/checkout/styles/components/_opf-checkout-payments.scss
```diff
@@ -29,13 +30,6 @@
       background-color: var(--cx-color-light);
     }
 
-    > cx-opf-checkout-terms-and-conditions-alert {
```
ercultimate, line 32: Aren't changes like this considered breaking changes?  I was under the impression that the CSS is all public interface.

integration-libs/opf/payment/core/connectors/opf-payment.connector.ts
```diff
@@ -20,6 +20,11 @@ import {
 import { Observable } from 'rxjs';
 import { OpfPaymentAdapter } from './opf-payment.adapter';
 
+/**
+ * This connector acts as between Storefront and the OPF Commerce Cloud Adapter.
```
ercultimate: This could use some wordsmithing.

integration-libs/opf/b2b-checkout/root/guards/opf-b2b-checkout-steps-set.guard.ts
```diff
@@ -0,0 +1,186 @@
...
+            this.routingConfigService.getRouteConfig(step.routeName)?.paths?.[0]
+          }`;
+          return stepRouteUrl === currentRouteUrl;
+        });
+        // get current step
+        let currentStep;
+        if (currentIndex >= 0) {
+          currentStep = steps[currentIndex];
+        }
+        if (Boolean(currentStep)) {
```
ercultimate, line 96: We don't need a boolean conversion in an if condition.

integration-libs/opf/b2b-checkout/root/guards/opf-b2b-checkout-steps-set.guard.ts
```diff
@@ -0,0 +1,186 @@
...
+          } else {
+            return this.getUrl(step.routeName);
+          }
+        } else {
+          if (deliveryAddress && Object.keys(deliveryAddress).length) {
+            return true;
+          } else {
+            return this.getUrl(step.routeName);
+          }
+        }
```
ercultimate, line 182: Could this be simplified?  Something more like this?
```
        const addressKeys = Object.keys(deliveryAddress ?? {}).length;
        if (isAccountPayment) {
          if (addressKeys && !!costCenter) {
            return true;
          }
        } else {
          if (addressKeys) {
            return true;
          }
        }
        return this.getUrl(step.routeName);
```

integration-libs/opf/b2b-checkout/root/opf-b2b-checkout-root.module.ts
```diff
@@ -0,0 +1,48 @@
...
+import { defaultOpfB2bCheckoutOccEndpointsConfig } from './config/default-opf-b2b-checkout-occ-endpoints-config';
+
+export const OPF_B2B_CHECKOUT_CMS_COMPONENTS: string[] = [
+  'OpfCheckoutPaymentType',
+  'OpfCheckoutDeliveryAddress',
+  'OpfCheckoutReview',
+  'OpfB2bCheckoutPaymentAndReview',
+];
+
+export function defaultOpfB2bCheckoutComponentsConfig() {
```
ercultimate, line 26: I believe the config factories are supposed to have the `-Factory` suffix.

ercultimate (commented):
Mostly looks good.  Just awaiting a response to the current comments.

integration-libs/opf/b2b-checkout/components/opf-b2b-checkout-payment-and-review/opf-b2b-checkout-payment-and-review.component.ts
Matejk00: Switched `poNumber$` to assigned property.

integration-libs/opf/b2b-checkout/components/opf-b2b-checkout-review/opf-b2b-checkout-review.component.ts
Matejk00: Switched `poNumber$` to assigned property.

integration-libs/opf/b2b-checkout/components/opf-b2b-checkout-review/opf-b2b-checkout-review.component.ts
Matejk00: Switched `costCenter$` to assigned property.

integration-libs/opf/b2b-checkout/components/opf-b2b-checkout-review/opf-b2b-checkout-review.component.ts
Matejk00: The use of `JSON.stringify(page)` is a bit of a hacky approach to search for a CMS component ID within the entire page object. 

`JSON.stringify` searches through the entire page object structure, including nested properties
It can find the component ID even if it's deeply nested in the page object structure.

This approach was used mainly becauseI t's a one-liner and making the code concise. It avoids the need to understand or traverse the complex structure of the page object manually and eliminates the need to handle multiple levels of nesting.

integration-libs/opf/b2b-checkout/components/opf-b2b-checkout-review/opf-b2b-checkout-review.component.ts
Matejk00: I changed to the typed form.

integration-libs/opf/checkout/components/opf-checkout-payment-and-review/opf-checkout-payment-and-review.component.ts
Matejk00: As far as I know, we don't need to worry about breaking changes, since we don't follow major/minor release versions. However, introducing B2B OPF Checkout also impacts B2C components of the OPF, so customers need to be aware that they'll have to adjust their code to use the new OPF functionalities.

integration-libs/opf/checkout/components/opf-checkout-payments/opf-checkout-payments.component.ts
Matejk00: I changed it to use a decorator to keep it consistent with the other outputs in the file.

integration-libs/opf/checkout/components/opf-checkout-review-cart-details/opf-checkout-review-cart-details.component.html
Matejk00: That's correct. It's in an OnPush component, so the calculation only happens on cart changes.
The template binding is optimized by Angular's change detection.

integration-libs/opf/checkout/styles/components/_opf-checkout-payments.scss
Matejk00: The same aspect as: https://github.com/SAP/spartacus/pull/20210/#discussion_r2079157877

integration-libs/opf/b2b-checkout/root/guards/opf-b2b-checkout-steps-set.guard.ts
Matejk00: Since we are extending from CheckoutStepsSetGuard (which already exists in the SPA codebase), I aimed to keep the original code as untouched as possible, adding only the minimal adjustments required for OPF. The part you mentioned was directly copied from the parent guard. In my view, this approach helps ensure code stability by preserving the well-tested functionality of the base guard, while also improving maintainability.

integration-libs/opf/b2b-checkout/root/guards/opf-b2b-checkout-steps-set.guard.ts
Matejk00: It is the same case as in: https://github.com/SAP/spartacus/pull/20210#discussion_r2079196788

integration-libs/opf/b2b-checkout/root/opf-b2b-checkout-root.module.ts
Matejk00: In SPA I think we are not using this convention. Wanted to be aligned with another features config files:
image

integration-libs/opf/b2b-checkout/components/opf-b2b-checkout-review/opf-b2b-checkout-review.component.ts
```diff
@@ -0,0 +1,137 @@
...
+
+  get termsAndConditionInvalid(): boolean {
+    return this.checkoutSubmitForm?.invalid ?? true;
+  }
+
+  get termsAndConditionsFieldValue(): boolean {
+    return Boolean(this.checkoutSubmitForm?.get('termsAndConditions')?.value);
+  }
+
+  get paymentType$(): Observable<PaymentType | undefined> {
```
ercultimate, line 77: This one is still a getter.  Is that intentional?

integration-libs/opf/b2b-checkout/components/opf-b2b-checkout-review/opf-b2b-checkout-review.component.ts
Platonn: Sorry to jump in,
Did we try the alternative approach - picking the value from proper nested properties? If yes, what specific challenges did we face?

btw. the current solution is prone to edge-case name clashes i.e. if the same string ID used in a different property (e.g. as an ID of some cms slot - in that case the _component with such ID_ is not present of the page, but we would think it is). We don't have control over how customers name their other CMS components and slots.

integration-libs/opf/b2b-checkout/root/opf-b2b-checkout-root.module.ts
ercultimate: I see.  CMS configurations do not use that convention.

integration-libs/opf/b2b-checkout/components/opf-b2b-checkout-review/opf-b2b-checkout-review.component.ts
Matejk00: This one is intentional.

integration-libs/opf/b2b-checkout/components/opf-b2b-checkout-review/opf-b2b-checkout-review.component.ts
Matejk00: As we agreed on the slack conversation I will create a ticket to find the better solution for this.

ercultimate (approved):

PR #20213, opened 2025-04-23
chore: Automation to check changes in the peerDependencies

rmch91:


ercultimate:
Shouldn't this also check for changes in `projects/schematics/src/dependencies.json`?

rmch91:
> Shouldn't this also check for changes in `projects/schematics/src/dependencies.json`?

@ercultimate I think it is not needed to check dependencies.json. There is a config check step in the CI which checks for any inconsistency in versions. `dependencies.json` is generated by `npm run generate:deps` command. If someone will change directly version in this file the config check will complain that `dependencies.json file is not up to date`

ercultimate (dismissed):

ercultimate (approved):

PR #20218, opened 2025-04-24
fix: fix the problem when adding products to cart from saved cart (CXSPA-9781)

kimhw0630:
Closes: https://jira.tools.sap/browse/CXSPA-9781


feature-libs/cart/base/components/add-to-cart/add-to-cart.component.ts
```diff
@@ -288,6 +298,13 @@ export class AddToCartComponent implements OnInit, OnDestroy {
     }
   }
 
+  protected getQuantityFromSavedCart(productCode: string): number {
+    return (
+      this.savedCart?.entries?.find((e) => e?.product?.code === productCode)
+        ?.quantity ?? 0
```
ercultimate, line 304: Should we use a default of 1?

ercultimate: I see, if the quantity is not located, the add-to-cart process is cancelled.

ercultimate (approved):

PR #20226, opened 2025-04-25
feat: add Punchout Inspect Cart feature

FollowTheFlo:


integration-libs/punchout/styles/_punchout-template.scss
FollowTheFlo, line 1: PunchoutTemplate is a CMS template used in all our punchout pages.
It displays:
- context slot which add context bar with 'language' 'currency' 'theme switcher' drop-downs. those menus stays in top bar even in mobile dimension as burger menu does not exist on punchout pages.
- body slot filled with a punchout component.
- loginSite slot with empty component to hide 'login button'. (all handle cms server side, no change in our repo)

integration-libs/punchout/components/punchout-buttons/punchout-buttons.component.html
```diff
@@ -7,6 +7,7 @@
     {{ 'punchout.backToRequisition' | cxTranslate }}
   </button>
   <button
+    *ngIf="!removeCancelButton"
```
FollowTheFlo, line 10: We only display 'Back To Requisition' btn on Inspect Cart page.

feature-libs/cart/base/components/cart-shared/cart-item-list/cart-item-list.component.ts
```diff
@@ -56,6 +56,8 @@ export class CartItemListComponent implements OnInit, OnDestroy {
 
   @Input() readonly: boolean = false;
 
+  @Input() disableLink: boolean = false;
```
FollowTheFlo, line 59: I had to add a new toggle to inactive product link. On Inspect Cart page, user should not be able to leave the page including navigation to pdp.
Even though navigation guard would prevent it.
I belive it is not a breaking change as this new toggle is optional.

feature-libs/cart/base/root/models/cart.model.ts
```diff
@@ -118,6 +118,7 @@ export interface CartItemComponentOptions {
   displayAddToCart?: boolean;
   addToCartString?: string;
   cartType?: CartType;
+  disableItemLink?: boolean;
```
FollowTheFlo, line 121: introducing optional new toggle to active/unactive anchor tag on item title.
context: we don't want user to navigate to pdp.

feature-libs/cart/base/components/cart-shared/cart-item-list-row/cart-item-list-row.component.html
```diff
@@ -10,6 +10,7 @@
     <div class="cx-table-item-container">
       <!-- Item Image -->
       <a
+        *ngIf="!options.disableItemLink"
```
FollowTheFlo, line 13: see below comments about introduction of new 'disableItemLink' toggle

FollowTheFlo: Not a breaking change in my opinion. The toggle is optional,  DOM stays unchanged for existing customer not using this toggle.

Pucek9:
looks as expected
image

Pucek9 (approved):

PR #20242, opened 2025-04-29
feat: introduce esbuild bundler

npapp-dev002:
Related to [CXSPA-9318](https://jira.tools.sap/browse/CXSPA-9318)

We want our example `storefrontapp` in our monorepo to look very similar to our customers' applications. In order to achieve that the following changes were introduced:

- Storefrontapp was updated according to guidelines: 
  - [modernize-apps-migrated-from-6.8-to-2211.19.md](https://github.com/SAP/spartacus/blob/f51aa3ec4fb0bf2b9f82f2eb3eb792bbe0c7a611/docs/migration/2211_35/modernize-apps-migrated-from-6.8-to-2211.19.md)
  -   [modernize-apps-migrated-from-2211.32-to-2211.35.md](https://github.com/SAP/spartacus/blob/f51aa3ec4fb0bf2b9f82f2eb3eb792bbe0c7a611/docs/migration/2211_35/modernize-apps-migrated-from-2211.32-to-2211.35.md)

The following documentation outlines the changes we made and the challenges we faced during the modernization.

# Challenges:
As it described in the previous [PR](https://github.com/SAP/spartacus/pull/20011), we faced the following challenges:
- handle esbuild warnings about sideEffects
- handle symbolic imports in .scss files
- handle csr and prerender builds

## Esbuild warnings about sideEffects
image

This PR introduces a new plugin designed to filter and remove `sideEffect` warnings after the build process completes.

## Symbolic imports in .scss files
Symbolic imports in .scss files (like `@import '@spartacus/styles';`) allow you to reference styles from npm packages or workspace libraries using a package name instead of a relative or absolute file path. Previously `extra-webpack.config.js` handled these imports, but it looked like that esbuild does not have similar support.

Many workarounds were attempted:
- An attempt was made to use a Sass plugin to address the issue, but Angular's build system does not permit customization of Sass handling.
- Rewriting every single symbolic import to their relative counterparts and changing them back after the application build is done. This was a quite invasive and extensive workaround that affected all of the libraries.
- The last and successful attempt uses `npm workspaces`. It creates symlinks to each library's source code folder inside the `node_modules/@spartacus/<library-name>`.

In order to make your library to be able to handle symbolic imports, it has to be listed under the workspaces section of the main `package.json` file.

## Handle csr and prerender builds
Previously the SSR related configuration was handled by different builders, but now the new Angular application builder provides the integrated functionality for all of the following preexisting builders (prerender, server, ssr-dev-server). The storefrontapp was updated with new build targets to reflect the changes. For example, `build-csr` target was introduced to execute the application builder with `noSsr` and `production` configuration. 

# Important updates to note

## `npm run build` changes
`npm run build` creates two folders: 
- The browser folder contains the client-side JavaScript, CSS, and assets for the Angular application, with index.csr.html serving as the entry point. To generate index.html for your http server, execute npm run build:csr.
- The server folder contains the server-side bundle, which is used for server-side rendering (SSR) with Node.js. This code runs on the server to pre-render pages before sending them to the browser.

## NPM workspaces
npm workspaces allow you to manage multiple packages (like libraries and apps) in a single monorepo. Each workspace is a subproject with its own package.json, and dependencies between them are symlinked for easy local development.

Usage:
- Check if your packages is under the listed folders in root folder's `package.json`. It should be either under `feature-libs/* `or listed directly as `projects/core`.
- Run npm install at the root to install all dependencies and link local packages.
- Test your package.

## Removed config files
`tsconfig.server.json` and `tsconfig.server.prod.json` were removed because they are no longer used. The default install script was modified to reflect these changes.

## Validate ssr changes
Please note that SSR markers such as `USING_DEV_MODE_MARKER` and `USING_PROXY_BACKEND_MARKER` are no longer located in fixed positions as they were with Webpack. With the new bundler, these markers may appear in different chunk files and at arbitrary locations. As a result, the search logic was updated to scan both the server and browser directories comprehensively.

# Remaining work

These issues might be out of scope for the current PR, because they require a certain level of refactoring to introduce.

## Isolated Modules
IsolatedModules is a TypeScript compiler option (--isolatedModules) that enforces each file to be treated as a separate module during compilation. It ensures that every file can be safely transpiled in isolation, without relying on information from other files.

Reasons to use:
- Requires every file to be a module (must have at least one import or export).
- Disallows certain TypeScript features that can't be safely transpiled in isolation (e.g., const enum, namespaces, and type-only imports/exports without code).
- Commonly used with build tools like Babel or Esbuild, which transpile files one at a time.
- Helps catch issues early when using transpilers that do not perform full type-checking or cross-file analysis.  

It is currently not added because of the errors that occur for `dev-server`.
image

## Use Define Class Fields
The `useDefineForClassFields` flag is a TypeScript compiler option that controls how class fields are emitted in the compiled JavaScript:

Reasons to remove it
- When set to true (the default in recent TypeScript versions), class fields are defined using the standard ECMAScript semantics. This means fields are defined with Object.defineProperty or as direct field declarations, and are initialized before the constructor runs.
- When set to false, TypeScript uses legacy behavior, assigning fields inside the constructor after super() is called.

It is currently set to `false` in the main `package.json`, because it requires extensive refactoring:
image

## Side effects warnings
Side effect warnings are filtered only after the build is completed, because the onEnd hook in esbuild plugins only runs after a successful build (when there are no fatal errors). If there is a compilation error, esbuild does not call onEnd, so `filterWarningsPlugin` never gets a chance to filter the warnings. As a result, all warnings are shown alongside errors.
image


package.json
```diff
@@ -126,6 +126,7 @@
   },
   "private": false,
   "dependencies": {
+    "@angular-builders/custom-esbuild": "^19.1.0",
```
Platonn: [nitpick]
lets move new deps to `devDependencies`

tools/schematics/testing.ts
```diff
@@ -275,6 +275,9 @@ async function publishAllPackages(): Promise<void> {
 
   /** Number of packages published, successfully or not */
   let completedCount = 0;
+  /** Transform storefrontstyles and dist to use symbolic paths before publish */
+  execSync("ts-node tools/scripts/regex-scss-path-replace.ts ./dist");
+  execSync("ts-node tools/scripts/regex-scss-path-replace.ts ./projects/storefrontstyles");
```
Platonn: [medium]
1. Ideally, let's to run the new regex-replace as part of the regular nx `build` task of each library.
Thanks to this it won't matter from where you'll build or publish libs.
This script `tools/schematics/testing.ts` is used only for internal testing.

2. if we make changes inside the folder `storefrontstyles`, then we're changing the sources, which is a bummer. I believe, instead we should copy those styles to /dist folder (as an "artificial build" step), and then regex-replace only there in /dist folder

tools/scripts/paths.ts
```diff
@@ -0,0 +1,59 @@
...
+  'product-multi-dimensional',
+  'product',
+  'qualtrics',
+  'quote',
+  'requested-delivery-date',
+  's4-service',
+  's4om',
+  'storefinder',
+  'user',
+];
```
Platonn: [question]
1. Don't we have too little libs listed there? I had an impression we have much more libs, don't we? (maybe it's my bad and we indeed we have only them).
2. can we have `asm` defined only once? against a patter `@import 'asm'` (with ending `'`) or `@import 'asm/` (with ending `/`). I mean not specifically `asm`, but for all libs we could replace just a prefix. What do you think?

`@import 'lib-name/whatever/nested` -> `@import @spartacus/lib-name/whatever/nested`
`@import 'lib-name'` -> `@import @spartacus/lib-name`

tools/scripts/paths.ts
```diff
@@ -0,0 +1,59 @@
...
+ * and update their `@import` statements to use the correct paths.
+ *
+ * Example:
+ * - For `{ key: 'feature-libs', value: '@spartacus' }`:
+ *   `@import 'feature-libs/asm/styles';` becomes `@import '@spartacus/asm/styles';`.
+ * - For `{ key: 'projects/schematics/index', value: '@spartacus/schematics' }`:
+ *   `@import 'projects/schematics/index';` becomes `@import '@spartacus/schematics';`.
+ */
+export const pathsMapping = [
+  { key: ' cart/base/styles/components/cart-item-list', value: '@spartacus/cart/base/styles/components/cart-item-list' },
```
Platonn: [question]
why the key has preceding space ` cart/.....`

npapp-dev002: Corrected. Thank you.

tools/scripts/paths.ts
npapp-dev002: 1. I first included those who were referenced with symbolic imports in .scss files. I updated the code to contain every project. 
2. You are right. There were some issues where `@import 'lib-name'` was not handled properly.

package.json
npapp-dev002: It is not a nitpick. 🥇 It is completely valid point. Dependency moved to `devDependecies`.

projects/storefrontapp/project.json
```diff
@@ -171,6 +193,151 @@
...
     },
+    "serve-ssr": {
+      "executor": "nx:run-commands",
+      "options": {
+        "commands": [
+          {
+            "command": "nx run storefrontapp:build:production --watch"
+          },
+          {
+            "command": "sh -c 'while [ ! -f dist/storefrontapp/server/server.mjs ]; do sleep 1; done; nodemon dist/storefrontapp/server/server.mjs --watch dist/storefrontapp/server/server.mjs --signal SIGTERM --verbose'"
```
Platonn: 1. Do we really have to install `nodemon`? Can't we use native `node --watch`?
2. New builder outputs build artifacts to many chunked JS files, so we should not watch only the server.mjs but the whole folder. I believe this can be also achieved with `node --watch-path folder`

for more see https://medium.com/@moaidmoaidrazhy/node-watch-vs-nodemon-4c8f0fc8a1af

npapp-dev002: Done.

projects/ssr-tests/validate-ssr-build.ts
```diff
@@ -84,20 +84,34 @@ export default async function validateSsrBuild() {
...
       `
 SSR app is not using a local backend proxy as a base OCC url.
 ${BUILD_COMMAND_ADVICE}`
     );
   }
 
+  const fileContents = fs.readFileSync(SSR_APP_PATH, 'utf8');
   if (fileContents.includes(USING_DEV_MODE_MARKER)) {
     throw new Error(
       `
```
Platonn: For `USING_PROXY_BACKEND_MARKER` we're now searching the whole folder.
Shouldn't we search for `USING_DEV_MODE_MARKER` also in the whole folder?

projects/storefrontapp/project.json
```diff
@@ -7,21 +7,29 @@
   "generators": {},
   "targets": {
     "build": {
-      "executor": "@angular-builders/custom-webpack:browser",
+      "executor": "@angular-builders/custom-esbuild:application",
+      "logLevel": "error",
```
Platonn: 1. Do we really have to change the default logLevel?
2. And what was the previous default logLevel?

projects/storefrontapp/project.json
```diff
@@ -7,21 +7,29 @@
...
-          "path": "./extra-webpack.config.js"
-        },
+        "preserveSymlinks": true,
+        "plugins": ["./projects/storefrontapp/esbuild/plugins.ts"],
         "aot": true,
+        "allowedCommonJsDependencies": [
+          "whatwg-url",
+          "encoding",
+          "cross-fetch"
+        ],
```
Platonn, line 19: Why in this PR we have to add 3 packages to this allow-list `allowedCommonJsDependencies`? Why we didn't need this config before?

projects/storefrontapp/project.json
```diff
@@ -171,6 +193,151 @@
...
+          {
+            "command": "nx run storefrontapp:build:production --watch"
+          },
+          {
+            "command": "sh -c 'while [ ! -f dist/storefrontapp/server/server.mjs ]; do sleep 1; done; node --watch-path=dist/storefrontapp/server/ dist/storefrontapp/server/server.mjs'"
+          }
+        ]
+      }
+    },
+    "server": {
```
Platonn: Do we really have to maintain the architect section "server"? Won't "build" suffice?

One downside of maintaining it is that it contains a copy of "styles" array config, the same as in the "build" architect section?

Btw. I cannot find architect section "server" it in fresh Angular+Spartacus app with SSR.

projects/ssr-tests/validate-ssr-build.ts
npapp-dev002: Good question. If SSR is built in development mode, the whole `dist/storefrontapp/server` folder will be empty. This would make `validate-ssr-build.ts` to fail at the first check where the existence of `dist/storefrontapp/server/server.mjs` is checked. I think that makes the `USING_DEV_MODE_MARKER` check obsolete. What do you think? Should I remove this check, or extent the search to include both browser and server folders? (Note: only browser folder has content when ssr build is invoked with development configuration, and the marker can be found there)

projects/storefrontapp/tsconfig.app.json
Platonn, line 1: Out modernization docs mention adding the option `isolatedModules`.
IMHO not critical to add it, but just curious, did we have reasons to not add it in this PR?

https://help.sap.com/docs/SAP_COMMERCE_COMPOSABLE_STOREFRONT/10a8bc7f635b4e3db6f6bb7880e58a7d/9bb7dd061da14cfc9c5e3f36388c77e4.html

projects/storefrontapp/project.json
npapp-dev002: 1. No, we don't have to. That is just the residue of some testing before `filterWarningsPlugin` have been introduced.
2. info
I remove this logLevel setting.

projects/storefrontapp/project.json
npapp-dev002: This config was added suppress warnings like these: 
image
They occur for webpack configuration as well
image

projects/storefrontapp/project.json
npapp-dev002: You are right. We don't have to. I am going to remove it.

projects/ssr-tests/validate-ssr-build.ts
npapp-dev002: I've decided to check both folders for the marker.

ci-scripts/e2e-cypress.sh
```diff
@@ -95,7 +95,7 @@ else
 fi
 echo '-----'
 echo "Building Spartacus storefrontapp"
-npm run build
+npm run build:csr
```
Platonn: [MINOR]
Let's add a `# code comment` explaining _why_ we needed to use :csr here. It's not obvious for a newcomer to this code.

I mean, that with a common csr-and-ssr build, the index.html is absent, but index.csr.html is there instead. But it won't work, unless SSR is enabled. But for e2e we want to run a pure CSR app.

docs/libs/creating-lib.md
Platonn, line 1: [MINOR]
Please let's find the best place in this markdown to share also the following advice (this is just a proposal, feel free to use it or edit it):


After creating your library, please run `npm install` to update `package-lock.json` with information about the new lib. Then please commit changes in the `package-lock.json`.

Note: The [npm workspaces](https://docs.npmjs.com/cli/v11/using-npm/workspaces) feature will create symlinks between the source code folder of your lib and their virtual representations inside `node_modules/@spartacus/<your lib>`. It's needed for symbolic import paths in SCSS to work, e.g. `@import '@spartacus/cart/x/y/z'`

Note 2: If you can't see any changes in `package-lock.json` after adding your library and running `npm install`, it might be the case that your library folder created outside of the registered _npm workspaces_. Please make sure your library folder is placed in side one of the directories listed in the "workspaces" property of the main `/package.json` of our monorepo.

package.json
```diff
@@ -107,26 +118,27 @@
...
     "test:storefront:lib": "nx test storefrontlib --source-map --code-coverage",
     "test:opf:lib": "nx test opf --source-map --code-coverage",
-    "dev:ssr": "env-cmd --no-override -e dev,b2c,$SPA_ENV cross-env NODE_TLS_REJECT_UNAUTHORIZED=0 nx run storefrontapp:serve-ssr",
-    "serve:ssr": "node dist/storefrontapp-server/main.js",
-    "serve:ssr:ci": "NODE_TLS_REJECT_UNAUTHORIZED=0 SSR_TIMEOUT=0 node dist/storefrontapp-server/main.js",
-    "serve:ssr:dev": "cross-env NODE_TLS_REJECT_UNAUTHORIZED=0 node dist/storefrontapp-server/main.js",
+    "serve:ssr": "node dist/storefrontapp/server/server.mjs",
+    "serve:ssr:watch": "node --watch dist/storefrontapp/server/server.mjs",
```
Platonn: [MEDIUM]
Let's watch the whole dist /server folder for changes, because it can contain other JS chunks that can change. and then we should restart the server

```suggestion
    "serve:ssr:watch": "node --watch dist/storefrontapp/server dist/storefrontapp/server/server.mjs",
```

projects/storefrontapp/project.json
```diff
@@ -141,36 +157,57 @@
...
+          "optimization": true,
+          "outputHashing": "all",
+          "sourceMap": false,
+          "namedChunks": false,
+          "serviceWorker": false,
+          "server": "projects/storefrontapp/src/main.server.ts",
+          "prerender": {
+            "discoverRoutes": false,
+            "routesFile": "projects/storefrontapp/prerender.txt"
+          }
```
Platonn: [MINOR]
If possible, let's not duplicate here config properties specific for "production" or "development" build target (e.g. let's not include `optimization: true`, etc.) in the "prerender" section. Let's leave here only things needed for the prerendering - which is just  one sub-property: "prerender".

And then let's update other "scripts" in `package.json`, to include 2 suffixes: both the `:prerender` and the `:development`/`:production` , e.g. `storefrontapp:build:production:prerender`

In other words, I suggest to leave here only:
```suggestion
        "prerender": {
          "prerender": {
            "discoverRoutes": false,
            "routesFile": "projects/storefrontapp/prerender.txt"
          }
```

projects/storefrontapp/project.json
```diff
@@ -141,36 +157,57 @@
...
+        "commands": [
+          {
+            "command": "nx run storefrontapp:build:production --watch"
+          },
+          {
+            "command": "sh -c 'while [ ! -f dist/storefrontapp/server/server.mjs ]; do sleep 1; done; node --watch-path=dist/storefrontapp/server/ dist/storefrontapp/server/server.mjs'"
+          }
+        ]
+      }
+    },
```
Platonn: [MEDIUM]
I can see we've removed `"dev:ssr"` script from package.json and left only `"dev:cds:ssr"` which is the only script that uses this "serve-ssr" architect.
1. Was it deliberate that we removed `"dev:ssr"` script from package.json? If, yes, why did we leave its cds variant?
2. If we keep this `"serve-ssr"` architect, IMHO we should not run with with `:production` suffix, but with `:development`. That's becasue we want to run ssr in **dev** mode

```diff
- "command": "nx run storefrontapp:build:production --watch"
+ "command": "nx run storefrontapp:build:development --watch"
```

or just alternatively:
```diff
- "command": "nx run storefrontapp:build:production --watch"
+ "command": "npm run watch"
```

where the "watch" script in package.json should have `--configuration development`, just like it is in fresh Angular 19 apps:

```diff
"scripts":
    ...
-   "watch": "npm run build -- --watch"
+   "watch": "npm run build -- --watch --configuration development"
```

see screenshot from a fresh app:

image


Note: by relying on `npm run watch` command (which transitively relies on `npm run build`), we're leveraging cross-env definitions which already live in the `npm run build` command

projects/storefrontapp/src/main.server.ts
```diff
@@ -14,4 +14,4 @@ if (environment.production) {
   enableProdMode();
 }
 export { renderModule } from '@angular/platform-server';
-export { AppServerModule as default } from './app/app.server.module';
+export { AppModuleServer as default } from './app/app.module.server';
```
Platonn: [MINOR]
Although angular renamed the path, the module name remained unchanged `AppServerModule` 🙃 

```suggestion
export { AppServerModule as default } from './app/app.module.server';
```

tools/config/tsconfig-paths.ts
```diff
@@ -371,55 +371,7 @@ function handleAppConfigs(
...
-      {} as { [key: string]: [string] }
-    );
-
-  const hadErrorsServerProd = handleConfigUpdate(
-    serverProdEntryPoints,
-    'projects/storefrontapp/tsconfig.server.prod.json',
-    options
-  );
-
-  if (hadErrorsApp || hadErrorsServer || hadErrorsServerProd) {
```
Platonn, line 422: [QUESTION]
Why we removed this?

tsconfig.json
```diff
@@ -4,19 +4,17 @@
...
     "typeRoots": ["node_modules/@types"],
-    "lib": ["ES2022", "dom"],
     "strictPropertyInitialization": false,
     "strict": true,
     "noUnusedLocals": true,
     "noUnusedParameters": true,
     "stripInternal": true,
     "resolveJsonModule": true,
+    "moduleResolution": "bundler",
+    "skipLibCheck": true,
```
Platonn: [MEDIUM]
This is a global tsconfig, that also our libraries' tsconfigs extend.
The modernization guideline suggests to add "skipLibCheck: true" for the apps.
So I'd do it only in our `tsconfig.app.json`.
I believe we shouldn't do "skipLibCheck: true" for our libraries tsconfigs. Otherwise we can loose some type safety there in terms of their 3rd party deps (see [skipLibCheck docs](https://www.typescriptlang.org/tsconfig/#skipLibCheck))

projects/storefrontapp/project.json
npapp-dev002: In NX, configurations cannot be merged (https://github.com/nrwl/nx/discussions/22710). That is why there are some duplicated values here for prerender.

projects/storefrontapp/project.json
npapp-dev002: I removed it because of the migration guide: 
image
I think I should have removed `dev:cds:ssr` as well. They can use `serve:ssr` and `serve:ssr:watch` as the guidelines suggested.

projects/storefrontapp/src/main.server.ts
npapp-dev002: You are right. Lol

tools/config/tsconfig-paths.ts
npapp-dev002: It checks `tsconfig.server.prod.json` and `tsconfig.server.json` which is no longer available.
image

tsconfig.json
npapp-dev002: You are right.

docs/libs/creating-lib.md
```diff
@@ -36,6 +36,12 @@ These are some naming guidelines for libraries:
 
 Run `nx g @schematics/angular:library <lib-name> --prefix=cx`, move it to the appropriate directory (`feature-libs` or `integration-libs`), and commit.
 
+After creating your library, please run npm install to update package-lock.json with information about the new lib. Then please commit changes in the package-lock.json.
+
+Note: The npm workspaces feature will create symlinks between the source code folder of your lib and their virtual representations inside node_modules/@spartacus/<your lib>. It's needed for symbolic import paths in SCSS to work, e.g. @import '@spartacus/cart/x/y/z'
```
Platonn: Formatting
```suggestion
Note: The npm workspaces feature will create symlinks between the source code folder of your lib and their virtual representations inside `node_modules/@spartacus/<your lib>`. It's needed for symbolic import paths in SCSS to work, e.g. `@import '@spartacus/cart/x/y/z'`
```

projects/storefrontapp/project.json
Platonn: Now I can see we've removed `"serve-ssr"` target from `project.json`. but we've added `"dev:ssr"` script to `package.json` that relies of `"serve-ssr"` target.

npapp-dev002: Removed.

projects/storefrontapp/project.json
Platonn: Thank you for linking the discussion in the Nx repo. I've found there a workaround that allows merging configurations and avoid duplication of properties among configurations - proposed by the Nx maintainer.

I've prepared a proposal sub-PR for this PR with this workaround implementation: 
https://github.com/SAP/spartacus/pull/20355 . See that sub-PR description for more details.

I'm looking forward to hearing what you think about it.
If by any chance you like it, feel free to merge. If not, no worries, just let me know.

projects/storefrontapp/tsconfig.app.json
```diff
@@ -1,15 +1,22 @@
...
+    "src/main.ts",
+    "src/main.server.ts",
+    "src/server.ts",
+    "src/environments/models/build.process.env.d.ts"
+  ],
   "include": ["projects/storefrontapp/src/**/*.d.ts"],
   "angularCompilerOptions": {
-    "strictTemplates": false
+    "strictTemplates": false,
+    "skipLibCheck": true
```
Platonn: [MINOR]
My bad, I didn't notice it before.
We already have skipLibCheck in this file - in "compilerOptions"
Let's remove it from the section "angularCompilerOptions". It should be not there

package.json
```diff
@@ -42,10 +52,11 @@
...
-    "build:ssr": "env-cmd --no-override -e dev,b2c,$SPA_ENV nx run storefrontapp:server:production",
-    "build:ssr:opf": "env-cmd --no-override -e opf,b2c,$SPA_ENV nx run storefrontapp:server:production",
-    "build:ssr:ci": "env-cmd -e ci,b2c,$SPA_ENV nx run storefrontapp:server:production",
-    "build:ssr:local-http-backend": "env-cmd -e local-http,b2c,$SPA_ENV nx run storefrontapp:server:production",
+    "build:csr": "env-cmd --no-override -e dev,b2c,$SPA_ENV nx run storefrontapp:build:noSsr",
+    "build:ssr": "env-cmd --no-override -e dev,b2c,$SPA_ENV npm run build",
+    "build:ssr:opf": "env-cmd --no-override -e opf,b2c,$SPA_ENV npm run build",
+    "build:ssr:ci": "env-cmd -e ci,b2c,$SPA_ENV npm run build",
+    "build:ssr:local-http-backend": "env-cmd -e local-http,b2c,$SPA_ENV npm run build",
     "build:storefinder": "npm --prefix feature-libs/storefinder run build:schematics && nx build storefinder --configuration production",
```
Platonn, line 56: [MEDIUM]
Let's not call transitively `npm run build`, because then we loose all the customized values of `env-cmd` (like `opf` or `local-http`) in favor of the `env-cmd` values defined in the `"build"` script of package.json.

In particular the the `npm run test:ssr` command fails on the `validate-ssr-build.ts` script, because there is no `USING_PROXY_BACKEND_MARKER` present in the dist /server folder. And it's because the command `build:ssr:local-http-backend` (which I've run before) didn't populate correctly the value `local-http` into `env-cmd` (because it was shadowed by `env-cmd` values defined inside the `"build"` script)

In other words, let me suggest the following changes:
```suggestion
 "build:ssr": "env-cmd --no-override -e dev,b2c,$SPA_ENV nx run storefrontapp:build:production",
    "build:ssr:opf": "env-cmd --no-override -e opf,b2c,$SPA_ENV nx run storefrontapp:build:production",
    "build:ssr:ci": "env-cmd -e ci,b2c,$SPA_ENV nx run storefrontapp:build:production",
    "build:ssr:local-http-backend": "env-cmd -e local-http,b2c,$SPA_ENV nx run storefrontapp:build:production",
```

Platonn, line 56: [MEDIUM]
By running `storefrontapp:build:noSsr` I believe the build happens without production optimizations. But before this PR, the build for the e2e purposes was a production build.
In other words, the `:noSsr` configuration is not merged with the `production` configuration.

To address it, I propose to apply the analogical workaround that [we applied](https://github.com/SAP/spartacus/pull/20242/commits/400cf67e472906bd509fba14f39b62349f95e5bf) for the prerendering, adding a new special target that merges configurations "production" and "noSsr":

1. Let's change this line to call the new target `storefrontapp:build-csr-production`:
```suggestion
    "build:csr": "env-cmd --no-override -e dev,b2c,$SPA_ENV nx run storefrontapp:build-csr-production",
```
1. And let's add the following target in the `project.json` (e.g. as a sibling under the target `"prerender-production"`):
```json
    "build-csr-production": {
      "executor": "@angular-devkit/architect:concat",
      "options": {
        "targets": [{ "target": "storefrontapp:build:production,noSsr" }]
      }
    },
```

package.json
```diff
@@ -107,26 +118,23 @@
...
-    "prerender:dev": "env-cmd --no-override -e dev,$SPA_ENV cross-env NODE_TLS_REJECT_UNAUTHORIZED=0 nx run storefrontapp:prerender --routes-file projects/storefrontapp/prerender.txt",
-    "prerender:cds": "env-cmd --no-override -e cds nx run storefrontapp:prerender --routes-file projects/storefrontapp/prerender.txt",
+    "prerender": "env-cmd --no-override -e dev,$SPA_ENV cross-env NODE_TLS_REJECT_UNAUTHORIZED=0 nx run storefrontapp:prerender-production",
     "publish:docs": "npx --yes gh-pages@^2.2.0 -d documentation -m \"Compodocs updates\"",
     "test:cds:lib": "nx test cds --code-coverage",
-    "build:cds:ssr": "env-cmd --no-override -e dev,b2c,cds nx run storefrontapp:server:production",
-    "dev:cds:ssr": "env-cmd --no-override -e dev,b2c,cds cross-env NODE_TLS_REJECT_UNAUTHORIZED=0 nx run storefrontapp:serve-ssr",
-    "prerender:cds:dev": "env-cmd --no-override -e dev,cds cross-env NODE_TLS_REJECT_UNAUTHORIZED=0 nx run storefrontapp:prerender --routes-file projects/storefrontapp/prerender.txt"
+    "build:cds:ssr": "env-cmd --no-override -e dev,b2c,cds nx run storefrontapp:build:production",
+    "watch": "npm run build -- --watch"
```
Platonn: [MEDIUM]
The watch build IMHO should use dev mode, so let's change it to 
```suggestion
    "watch": "npm run build -- --watch --configuration development"
```

(and btw. we might want to move it e.g. right under the "build" script, so they are close to each other in the package.json file)

projects/storefrontapp/tsconfig.app.json
npapp-dev002: You're absolutely right, I missed that too.

package.json
npapp-dev002: I assume that storefinder was accidentally removed. I added it back

npapp-dev002: The changes were added.
Shouldn't we create development and production builds separately for this target?
Something like: 
```
    "build-csr": {
      "executor": "@angular-devkit/architect:concat",
      "configurations": {
        "development": {
          "targets": [{ "target": "storefrontapp:build:development,noSsr" }]
        },
        "production": {
          "targets": [{ "target": "storefrontapp:build:production,noSsr" }]
        }
      },
      "defaultConfiguration": "production"
    },
```
What do you think?

Platonn: I really like your idea! It makes no sense to have `-production` suffix hardcoded in the name of the target, if it can be achieved by the `--configuration production` argument or `:production` suffix. (e.g. `nx run storefrontapp:build-csr:production`.
TBH I tried such a configuraiton in project.json on Thursday, but couldn't get it to work (perhaps becasue of some typo in my config).
Now I used your above snippet and it works on my machine.

Would you mind then updating the targets `build-csr-production` and `prerender-production` according to your proposal? (so they have "configuration" as a param instead of hardcoded in the name?).

npapp-dev002: Done. Please check.

projects/storefrontapp/project.json
```diff
@@ -141,32 +157,59 @@
...
       "configurations": {
+        "development": {
+          "targets": [{ "target": "storefrontapp:build:development,noSsr" }]
+        },
         "production": {
-          "buildTarget": "storefrontapp:build:production"
+          "targets": [{ "target": "storefrontapp:build:production,noSsr" }]
+        }
+      },
+      "defaultConfiguration": "production"
```
Platonn: typo (I guess due to a copy-paste): for "prerender" we wanted to use `,withPrerender` (but not `,noSsr`)

npapp-dev002: Yes, you are right.

Platonn (dismissed):

Platonn (dismissed):

Platonn (approved):

PR #20250, opened 2025-04-30
fix: add cx-read-more component (for product reviews)

uroslates:
Closes https://jira.tools.sap/browse/CXSPA-9967


projects/storefrontlib/shared/components/read-more/read-more.component.ts
```diff
@@ -0,0 +1,41 @@
...
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  standalone: true,
+  imports: [CommonModule, I18nModule],
+})
+export class ReadMoreComponent implements OnChanges {
+  @Input() readMoreTranslation?: string = 'common.readMore';
+  @Input() readLessTranslation?: string = 'common.readLess';
+  @Input() maxLength = 360;
+
+  @Input() text: string = '';
```
Zeyber, line 37: To make this component more useful in other cases, should we include a translation key as an option for this?

projects/storefrontlib/shared/components/read-more/read-more.component.ts
```diff
@@ -0,0 +1,41 @@
...
+  ChangeDetectionStrategy,
+  Component,
+  Input,
+  OnChanges,
+  SimpleChanges,
+} from '@angular/core';
+import { I18nModule } from '@spartacus/core';
+
+/**
+ * Renders cx-read-more.
```
Zeyber: This is a nice approach to the read more case. Simple enough to understand quickly and works well. Could we expand on jsdocs for the class and properties?

projects/storefrontlib/cms-components/product/product-tabs/product-reviews/product-reviews.component.ts
```diff
@@ -43,12 +46,16 @@ export class ProductReviewsComponent {
   @ViewChild('writeReviewButton', { static: false })
   writeReviewButton: ElementRef;
 
+  @Input() inputCharactersForReviewTitle = 255;
```
Zeyber: We should keep this behaviour feature toggled as I would consider it a breaking change. If it falls back to `null`, then I think the `maxLength` property wouldn't be applied.

Zeyber: I think the property name could also be changed to something closer to the attribute it affects like `maxLengthReviewTitle`. Just an example, but I think it could be more clear.

projects/storefrontlib/cms-components/product/product-tabs/product-reviews/product-reviews.component.ts
```diff
@@ -127,12 +134,53 @@ export class ProductReviewsComponent {
     }
   }
 
+  get reviewTitleCharacterLeft(): number {
```
Zeyber: I'm not a huge fan of using getters this way as they retrigger on every component interaction (eg. clicking on the field triggers all, typing a character triggers all, etc). Its a big impact on performance at scale. It would be better if they could trigger only when the needed property changes.

Note: It could be an opportunity to look into the new angular signals.

image
image

projects/storefrontlib/cms-components/product/product-tabs/product-reviews/product-reviews.component.html
```diff
@@ -147,6 +166,16 @@ <h3>{{ 'productReview.overallRating' | cxTranslate }}</h3>
             *cxFeature="'!formErrorsDescriptiveMessages'"
             [control]="reviewForm.get('comment')"
           ></cx-form-errors>
+
+          <p
```
Zeyber, line 203: Could you please consult with UI designers on the styling details of this component? You can tag them on the onboarding chat with links to the PR (and maybe ticket for context).

Zeyber (changes requested):

projects/storefrontlib/shared/components/read-more/read-more.component.ts
uroslates: We alread have that - the labels are inputs (accepting translation keys) -  [here](https://github.com/SAP/spartacus/pull/20250/files/f1b78384c70716293ab3341bb287d4eacefeb4fe#diff-01919cde46266351fcd1fb69fe20e645f2402cb58cca75177dc6199e8a6baaffR28-R29)
```
@Input() readMoreTranslation?: string = 'common.readMore';
@Input() readLessTranslation?: string = 'common.readLess';
```

projects/storefrontlib/shared/components/read-more/read-more.component.ts
uroslates: Applied! 👍

projects/storefrontlib/cms-components/product/product-tabs/product-reviews/product-reviews.component.html
uroslates: If we're talking about the characters limiting feature & form errors - I don't think this is necessary since this kind of feature was already implemented on few other places, so the implementation here is fully matching the other usages & styles. 
But if needed I could do that. (or if its related to the read-more cmp)

projects/storefrontlib/cms-components/product/product-tabs/product-reviews/product-reviews.component.ts
uroslates: Addressed! 👍

projects/storefrontlib/cms-components/product/product-tabs/product-reviews/product-reviews.component.ts
uroslates: Applied 👍. Removed those and moved to template.

projects/storefrontlib/cms-components/product/product-tabs/product-reviews/product-reviews.component.ts
uroslates: Applied! 👍

projects/storefrontlib/shared/components/read-more/read-more.component.ts
Zeyber: Sorry, to be clear I mean `@Input() text: I18nKey | string = '';`. I suppose the other inputs should also have the `I18nKey` typing too right?

uroslates: Ah that. sorry the text is the actual text coming from the DB (so its should not be translatable - thats done in through the BE system).
As for the other 2 inputs NP I'll update it to have the name suffix I18nKey (to be more intuitive). As for they I18nKey type - i haven't found it and wouldn't like to introduce it now since its just a string. And introducing it now (and enforcing consistency) would require tons of changes on other places I guess. We could to that type of improvement globally as some other issue (if really required).

Zeyber: Thanks for clarification on `text`. That makes sense now.
Sorry, about the other inputs the class should be `Translatable` not `I18nKey.

projects/storefrontlib/cms-components/product/product-tabs/product-reviews/product-reviews.component.html
Zeyber: From the points discussed with UI, I think these are the ones they made that are in scope:
2. lets do a limit of 300 and then introduce "read more"
3. the placing of the action buttons is usually on the right side as well :slightly_smiling_face: (not sure which action button they are talking about here though).

projects/storefrontlib/shared/components/read-more/read-more.component.ts
uroslates: OK applied.

projects/storefrontlib/cms-components/product/product-tabs/product-reviews/product-reviews.component.html
uroslates: Points 2 and 3 are implemented. Additionally I've added translations UX suggested to the comment field ([commentPlaceholder](https://github.com/SAP/spartacus/pull/20250/files#diff-ef57707d2f11db7bd02093b4b7973d8f0afaaab963889f2a0855e7e636dcea5fR63)) .

Zeyber (approved):

PR #20251, opened 2025-04-30
 feat: limit of SSR in-memory cache is defined in bytes instead of number of cache entries

Platonn:
**Context**
------------

Each NodeJS pod has some available memory (`max_memory_restart` option in the `pm2` process manager) and when the total process memory exceeds this limit, the process is restarted. Ideally we'd like to avoid such situations, becasue restarts disrupt the continual responsiveness of the SSR server.

Angular in SSR mode needs a lot of "working memory" for the rendering for all incoming requests (e.g. creating the DOM structure in-memory of NodeJS, storing temporarily big responses from backend endpoints, creating the state of the whole Angular application). When the rendering is done, Angular destroys its application instance, and after a moment NodeJS decides to free up this memory with the Garbage Collector. But it's important to notice that temporarily during the renderings high peaks of "working memory" is needed, but ideally it should not exceed `max_memory_restart` limit of `pm2` process manager).

Besides that, some memory is needed for the in-memory cache of rendered HTML strings, stored in Spartacus `OptimizedSsrEngine`, or to be precise, stored in Spartacus `RenderingCache` class.
Currently this cache size could be limited only by the **number of entries** via the `cacheSize` option. It was not very precise method of limiting cache, becasue each HTML can have different size and it's not possible to predict in advance surely the actual space taken by all those cache entries. In other words, you could only assume in advance how much in average take one HTML cache entry and then calculate how many such average entries you contain inside a NodeJS process, so it doesn't restart becasue a memory limit. And a few exceptionally big HTML cache entries (big compared to an average) in practice could lead to crossing the memory limit for the cache, and together with high peaks of "working memory" needed for actual renderings, a process could be restarted due to exceeding the `max_memory_restart` limit. 

Although it's difficult to predict "working memory" for rendering each page (it's possible only with experimentation), the memory allocated for the string cache entries, should be predictable. So it would be much better and bullet-proof to limit the cache size in actual memory units (like `800 MB`) but not arbitrary cache entries number (like `3000 entries`).

**New feature**
-----------------

Because the total available memory for the NodeJS process is defined in memory units (e.g. 3GB) in the configuration of the NodeJS pod, it would be good to also define a hard precise limit of the cache size defined in memory units. This PR introduces such a new config option `cacheSizeMemory` (and deprecates the old `cacheSize` property), while introducing the new ssr feature toggle `ssrFeatureToggles.limitCacheByMemory`. And this PR sets the default value to `800MB` (which takes effect only when the new feature toggle is enabled).

A new class `RenderingCacheSizeManager` is introduced in this PR, which constantly tracks the summed size of each cache entry, ensuring the configured max `cacheSizeMemory` won't be exceeded. When attempting to add new entry that would exceed this limit, we keep removing oldest entries, until there is enough free memory for the new entry to be stored.

With experimentation, we measured that each character of the cached HTML string takes 2 bytes per character (more on this topic in the next section). That said, we exposed a new SSR option `cacheEntrySizeCalculator` which allows for customizing the strategy how the size of a cache entry is calculated. By default it assumes 2 bytes for each character of the HTML string. Theoretically it's possible to cache also error objects (which is not recommended in our docs), but for completeness our default calculator roughly approximates the size of the error, by summing up its 3 string properties: name, message, trace, which is not ideal and prone to under-estimation, especially when the error object has much more properties or even is not an instance of an Error object (theoretically `cacheEntry.err` has type `any`). For most customers who don't cache errors (as recommended), the default `cacheEntrySizeCalculator` should suffice. But for customers who - due to some reasons - deliberately want to cache some error objects, we allow them to customize the non-ideal default logic of calculating the size of the cached errors.

Future plans: in another PR I plan to move all files related to the rendering cache to the relative subfolder `/rendering-cache` (when I tried to do it initially in this PR, then `git diff` treated some files as removed/created instead of edited+moved, which caused harder review of changes).

**How "2 bytes per character" estimation was determined**
-------------------------------------------------------------
NodeJS V8 engine can represent strings with various data structures. Some of them (like `SeqOneByteString`) allocates 1 byte per character (plus a few bytes of overhead). Other structures (like `SeqTwoByteString`) allocate 2 bytes per character (plus a few bytes of overhead). In practice a single HTML string of OOTB Spartacus Homepage _in prod mode_ takes now in V8 ~520kB of "Allocated Size", so the overhead of _few bytes_ per string can be ignored in calculations. Instead, we can just assume some general additional safety margin, e.g. additional free 50 MB "just in case".

For Spartacus SSR the HTML strings in cache entries were consistently represented with two-byte-per-character representations (`SeqTwoByteString`) - it was verified through heap snapshots and memory measurements. See the example screenshot of NodeJS DevTools inspecting the Heap Snapshot for the string with HTML using `SeqTwoByteString`:

image

The hypothesis of using "2 byte per character" was confirmed also by comparing the "Allocated Size" reported in NodeJS for a specific string (512kB) against the expression `2 * string.length`, where `string.length` is the number of characters in the string. Various experiments gave evidences that the expression `2 * string.length` described accurately the "Allocated Size" reported by NodeJS DevTools. The experiments included our HTMLs of OOTB Homepage (512kB per HTML string), custom-made-longer HTMLs of OOTB Homepage (1.2M per HTML string) or big strings having only complex characters of UTF8 like `ąŌ∑ę∂™ŘŽĘŽ` repeated 208 times (so total length was 2080 chars, which gives 4160 expected Bytes of "Allocated Size", assuming 2 bytes per character, which matches the measured 4.2kB measurement screenshot below):

image

(note: NodeJS dev tools use the SI standard of memory units based on powers of 10, not powers of 2, so `800 MB` equals `800 * 1_000_000`)

I've made above measurements using `--inspect-brk` argument, connecting to the process via remote NodeJS debugger in Chrome and taking `Heap Snapshot`s. I'm not aware of any method of inspecting how much Heap Size V8 engine allocates in practice for each string, **from within the running NodeJS process**. In other words, I can see not option of Spartacus SSR inspecting accurately in runtime how much memory take each entry. So it's only by prior experimentation and investigating Heap Snapshots to understand how much memory is allocated for each HTML string in cache entries. And later on this knowledge was implemented in the logic of Spartacus `DefaultCacheEntrySizeCalculator` (`2* str.length`) to estimate and control the size of each string in cache entry in runtime.

**1GB of "working memory" is needed for 10 parallel renderings of OOTB Homepage**
--------------------------------------------------------------------------------------
To be precise, I've noticed that on my machine ~500MB of "working memory" is needed for rendering 10 parallel requests for OOTB Homepage as of today (note: in the experiment each parallel request for the Homepage had a distinct query param, to trigger separate renderings, even though all requests were for the Homepage). I logged `process.memoryUsage()` to CSV file every 10ms, and then from that data I observed the peak memory usage at the level of ~500MB (by total memory I mean Resident Set Size; not just Heap Size that is a subset of Resident Set Size).
For that experiment, I've ensured no CSR fallbacks happen and disabled cache (so the total memory doesn't grow due to cache size increase). See the "rss" (Resident Set Size) total process memory in the chart below (basd on the data in CSV):

image

Even though the observed total memory usage was ~500MB, to be on the safe side let's add a margin of yet another 500MB to it - becasue each customer's custom page can have different size than our OOTB Homepage, the data fetched from backend and stored temporarily in-memory can be different (bigger) and Garbage Collector might kick in after a delay (so unused memory is not unallocated immediately causing a temporary bigger `Heap Size` and `Resident Set Size` until GC kicks in). So let's guess-assume 1GB is needed for withstanding 10 parallel renderings.

This is just a general estimation, but each customer might need to do their own estimation to their specific needs.
The amount of "working memory" needed for rendering each custom page might be different. Moreover, customers can adjust the `concurrency` limit option of `OptimizedSsrEngine` (currently allowing for at most 10 parallel renderings by default).

Moreover, the virtual machine can be possibly scaled up to allow for more total available memory for the NodeJS process and then the `--max-memory-restart` limit of `pm2` process manager might be increased.

**Why we're setting the default `cacheSizeMemory` to 800MB**
--------------------------------------------------------
Assuming the `--max-memory-restart` config of `pm2` process manager set to `1.8GB` (`3GB * 0.6 max_memory_restart_factor = 1.8GB`) and assuming we should leave 1GB for "working memory" (as explained above) for ongoing renderings for incoming requests, then at most 800MB can be utilzed for the cache. That said, each customers may want to tweak this value to their specific needs. 

Note: As of today, the default NodeJS pod in CCV2 can have 3GB of available system memory and the default `max_memory_restart_factor` configuration in CCV2 is `x0.6` which translates to `1.8GB` for the actual value of the `--max-memory-restart`. That said, in the future those default limits might be up to change.

To give you a general feeling how many average cache entries can fit into 800MB limit, let's assume average to be ~520kB of "Allocated Size" in V8 (it's the "Allocated Size" for a single OOTB Homepage string HTML in prod mode, as of today) and let's assume 800MB of total `cacheSizeMemory` (note: I'm using SI standard of memory units based on powers of 10, not powers of 2). Then it gives ~1538 entres = 800 000 000 Bytes / 520 000 Bytes-Per-Entry. That said, each page can be different, so it's just to give you a general feeling.

**Why bother setting any default `cacheSizeMemory`, as enabling cache is not recommended anyway?**
-----------------------------------------------
Spartacus OptimizedSsrEngine needs some space for the implicit in-memory cache of HTML strings that finished rendering after their original requests timed out. That cached HTML entries are reused for the next incoming request for this url, to not waste that rendering result. Then such cache entry is evicted. This takes place, even when `cache: false` is configured (which is a recommended value). If customer enables `cache: true`, then we need cache size limit anyway too.

fixes https://jira.tools.sap/browse/CXSPA-8964

Co-authored-by: kpawelczak <42094017+kpawelczak@users.noreply.github.com>
Co-authored-by: Krzysztof <lbkrpaw87@gmail.com>

---

**QA steps**:
--------------

- [x] PREREQUISITES
	- [x] build libs on this branch
	- [x] publish to verdaccio
	- [x] create fresh ng19 app
	- [x] nstall Spartacus with SSR, using libs from verdaccio
	- [x] verify in `server.ts` that it has `limitCacheByMemory: true`
	- [x] add to top of `server.ts` the line to ignore wrong TLS cert on backend `process.env['NODE_TLS_REJECT_UNAUTHORIZED'] = '0';`
	- [x] add SsrOptimizationOptions to avoid CSR fallbacks during the testing
		- [x] timeout: 30_000, // to avoid CSR fallbacks due to a request timeout   concurrency: 1000, // to avoid CSR fallback due to too many parallel renderings
	- [x] in `server.ts` enable cache in SsrOptimizationOptions:
		- [ ] `cache: true`
	- [x] create a dir for QA results - `mkdir qa-resuts` 
	- [x] create an util script for making requests to SSR
		- [x] `touch make-requests.sh` and paste the following content:
```bash
#!/bin/bash

# Parse named parameters
while [[ $# -gt 0 ]]; do
  case $1 in
    --start-number)
      START_NUMBER="$2"
      shift 2
      ;;
    --range)
      RANGE="$2"
      shift 2
      ;;
    --parallel)
      PARALLEL=true
      shift
      ;;
    *)
      echo "Unknown parameter: $1"
      exit 1
      ;;
  esac
done

# Fallback to default values
START_NUMBER=${START_NUMBER:-1} # start from NUMBER=1
RANGE=${RANGE:-10} # request 10 pages
PARALLEL=${PARALLEL:-false}

function process_data() {
  NUMBER=$1
  curl -s http://localhost:4000/electronics-spa/en/USD/\?$NUMBER | node -e '
  let content = "";
  process.stdin.on("data", chunk => { content += chunk });
  process.stdin.on("end", () => {
    const Buffer_byteLength = 2* Buffer.byteLength(content, "utf8");
    const expectedHeapSizeIncreaseBytes = Buffer_byteLength;
    const expectedHeapSizeIncreaseKb = (expectedHeapSizeIncreaseBytes / 1000).toFixed(2);
    console.log({ expectedHeapSizeIncrease: { Bytes: expectedHeapSizeIncreaseBytes, kB: expectedHeapSizeIncreaseKb } });
    require("fs").writeFileSync(`qa-results/page-${process.argv[1]}.html`, content);
  });
  ' $NUMBER
}

END=$((START_NUMBER + RANGE - 1))

echo "Making requests for range $START_NUMBER to $END"
if [ "$PARALLEL" = true ]; then
  echo "Executing requests in parallel"
  for NUMBER in $(seq $START_NUMBER $END); do
    (
      process_data $NUMBER
    ) &
  done
  wait
else
  echo "Executing requests sequentially"
  for NUMBER in $(seq $START_NUMBER $END); do
    process_data $NUMBER
  done
fi

echo "All commands completed"
```
- [x] ... prerequisites continued:
	- [x] add the following lines to "scripts" section in the `package.json` of the fresh app, to manage a NodeJS process with a pm2 process manager
```json
    "pm2:start": "npx pm2 start dist/<your-app-name>/server/server.mjs --node-args='--expose-gc'",
    "pm2:delete": "npx pm2 delete 0",
    "pm2:monit": "npx pm2 monit"
```

- [x] TESTS - Verify cacheSizeMemory works (i.e. cache size doesn't exceed this limit)
	- [x] for easier testing, set small cacheSizeMemory, e.g. 20MB
		- [x] cacheSizeMemory: 20_000_000,
		- [x] note: assuming the "Allocated size" in V8 for storing a HTML of OOTB homepage generated prod being 519812 , it allows only for 38,5 cache entries of OOTB Homepage HTML to be stored in practice in cache (so in practice 38 full entries)
	- [x] build app
		- [x] `npm run build`
	- [x] open a new terminal window (ideally full screen)
		- [x] start SSR process and its monitoring
			- [x] `npm run pm2:start && npm run pm2:monit`
	- [x] in the bottom left pane you should see `Used Heap Size` with a value around 21-23MiB (if not the case, wait a few seconds)
	- [x] attack with a first singe request (that will be cached)
		- [x] in another terminal window (in the directory of the fresh app) run our util script for making requests
			- [x] `bash ./make-requests.sh --start-number 1 --range 1`
			- [x] after a moment you should see the logged value in console:
				- [x] { expectedHeapSizeIncrease: { Bytes: 519812, kB: '519.81' } }
	- [x] in the terminal window with the monitoring the Used Heap Size can temporarily raise above 100MiB during the rendeing, but please wait a few seconds - it should go down to around 30-32MiB when Garbage Collector kicks in
		- [x] the huge difference between initial 21MiB and 30MiB now is not caused by a memory leak, not by the significant cache size increase, but by NodeJS V8 engine being "warmed by" and remembering various data structures needed to execute Spartacus SSR for the first time. it will reuse this "knowledge" in the later renderings, so such huge jumps should be not observed later
	- [x] attack with the same request with the same url - the result should come immediately, becasue it will come from the in-memory cache of Spartacus SSR
		- [x] moreover in the monitoring terminal in top right pane you'll see the logged message "Render from cache (/electronics-spa/en/USD/?1) "
	- [x] now let's attack with 38 requests, with query params from 1 to 38, in sequence (not in paralllel)
		- [x] (note: the first request's response will be immediate, for next requests we'll need to wait for their rendering. On my machine I needed to wait around 70 seconds)
			- [x] on MacOS you can append `&& say "requests completed"` to get a voice notification when it all finishes
				- [x] `bash ./make-requests.sh --start-number 1 --range 38 && say "requests completed"`
	- [x] in the meantime in the monitoring terminal, the temporary `Used Heap Size` can increase to 200-300MiB, but it's normal. When all requests' renderings finish, we'll wait a few seconds for Garbage Collector to kick in
	- [x] please wait a few seconds after all the requests complete, then the Garbage Collector kicks in, and the Used Heap Size should go down to around 56-58 MiB
		- [x] please note that Used Heap Size ~56 MiB now is ~20MiB (which is also ~20MB) bigger than the initial Used Heap Size ~30MiB (after we "warmed up" V8 with the first request)
			- [x] to be precise it's 56-30 is 26 MiB
				- [x] I can only guess that those extra 6MiB is not a result of some memory leak, but of V8 engine warming up event more and remembering  various data structures needed to execure Spartacus SSR. I don't expect such jumps in Used Heap Size later...
	- [x] let's repeat all those 38 requests for those urls. the responses should come immediately, because their entries are already in the cache
		- [x] `bash ./make-requests.sh --start-number 1 --range 38 && say "requests completed"`
	- [x] we expect that the the limit of 20MB `cacheSizeMemory` can contain at most 38 OOTB Homepage HTMLs, therefore the request for 39th Homepage and caching its result should cause evicting the cache of the last cached entry - the one with query param "1"
	- [x] let's make the 39th request
		- [x] `bash ./make-requests.sh --start-number 39 --range 1 && say "requests completed"`
	- [x] now let's requests all pages from query param 2 to 39 - they should be all in cache, so responses should be immediate
		- [x] `bash ./make-requests.sh --start-number 2 --range 38 && say "requests completed"`
	- [x] now let's verify that the entry for the query param "1" was evicted from cache - let's request it and see the response is rendered, but not served immediately from cache
		- [x] `bash ./make-requests.sh --start-number 1 --range 1 && say "requests completed"`
	- [x] assuming that the cache could contain at most 38 distinct copies of OOTB Homepage, then adding the entry 1 to the cache (after having there entries from 2 to 39) should cause the entry 2 to be evicted
	- [x] let's verify that
		- [x] let's request pages from 3 to 39 - they should all be still served from cache immediately
			- [x] `bash ./make-requests.sh --start-number 3 --range 36 && say "requests completed"`
		- [x] ok now lets request URL with number 2 - and see that it's not served immediately from cache, but rendered with a delay
			- [x] `bash ./make-requests.sh --start-number 2 --range 1 && say "requests completed"`
	- [x] the rendering of the entry 2 and caching it should cause evicting of the entry with number 3 (as the oldest cache entry)
		- [x] let's verify that
			- [x] ok now lets request URL with number 3 - and see that it's not served immediately from cache, but rendered with a delay
				- [x] `bash ./make-requests.sh --start-number 3 --range 1 && say "requests completed"`
	- [x] Let's see in the monitoring that the Used Heap Size is still on the level of 55-57MiB
	- [x] Note: In all the next steps we'll request rounds of "fresh new 38 URLs" so they render and those new entries will be cached, causing old entries to be evicted. We'll observe the `Used heap size` stays at the same level (however intially it will grow a bit with each round, I guess likely becasue of NodeJS will be still "warming up", but eventually we'll see the `Used Heap Size` stabilizes after the ~5th round of requests
	- [x] OK, now let's render requests with numbers from 101 to 138 -
		- [x] (by the way causing them all to be cached, and clearing the cache for entries with numbers from 1 to 39). It should take around 70seconds
			- [x] `bash ./make-requests.sh --start-number 101 --range 38 && say "requests completed"`
		- [x] In the meantime the Used Heap Size can temporarily go up above 100-300MiB
	- [x] after all renderings finish, let's wait a few seconds for Garbage Collector to kick in and see that the `Used Heap Size` went down to around 58-60MiB
		- [x] please note it didn't increase for the next 20MiB, so the max cache memory size is kept under control
		- [x] but we can also see that the `Used Heap Size` increased from the last measurement by around 3MiB (from 55 to 58MiB). I can only guess it's not a memory leak in the app, but just again V8 warming up even more and remembering some data structures for optimization reasons
	- [x] OK, let's try now to render entries with numbers 201 to 238 (which should cause evicting entries with numbers 101-138). You'll need to wait another 70seconds
		- [x] `bash ./make-requests.sh --start-number 201 --range 38 && say "requests completed"`
		- [x] In the meantime the Used Heap Size can temporarily go up above 100-300MiB
	- [x] after all renderings finish and after waiting few seconds for Garbage Collector we can see the `Used Heap Size` went down to 59-61MiB
		- [x] please note it didn't increase for the next 20MiB, so the max cache memory size is kept under control
		- [x] but we can also see that the `Used Heap Size` increased from the last measurement by around 1MiB (from 58 to 59MiB). I can only guess it's not a memory leak in the app, but just again V8 warming up even more and remembering some data structures for optimization reasons
	- [x] OK, let's try now to render entries with numbers 301 to 338 (which should cause evicting entries with numbers 201-238). You'll need to wait another 70seconds
		- [x] `bash ./make-requests.sh --start-number 301 --range 38 && say "requests completed"`
		- [x] In the meantime the Used Heap Size can temporarily go up above 100-300MiB
	- [x] after all renderings finish and after waiting few seconds for Garbage Collector we can see the `Used Heap Size` went down to 60-62MiB
		- [x] please note it didn't increase for the next 20MiB, so the max cache memory size is kept under control
		- [x] but we can also see that the `Used Heap Size` increased from the last measurement by around 1MiB (from 59 to 60MiB). I can only guess it's not a memory leak in the app, but just again V8 warming up even more and remembering some data structures for optimization reasons
	- [x] OK, let's try now to render entries with numbers 401 to 438 (which should cause evicting entries with numbers 301-338). You'll need to wait another 70seconds
		- [x] `bash ./make-requests.sh --start-number 401 --range 38 && say "requests completed"`
		- [x] In the meantime the Used Heap Size can temporarily go up above 100-300MiB
	- [x] after all renderings finish and after waiting few seconds for Garbage Collector we can see the `Used Heap Size` went down to 61-63MiB
		- [x] please note it didn't increase for the next 20MiB, so the max cache memory size is kept under control
		- [x] but we can also see that the `Used Heap Size` increased from the last measurement by around 1MiB (from 60 to 61MiB). I can only guess it's not a memory leak in the app, but just again V8 warming up even more and remembering some data structures for optimization reasons
	- [x] OK, let's try now to render entries with numbers 501 to 538 (which should cause evicting entries with numbers 401-438). You'll need to wait another 70seconds
		- [x] `bash ./make-requests.sh --start-number 501 --range 38 && say "requests completed"`
		- [x] In the meantime the Used Heap Size can temporarily go up above 100-300MiB
	- [x] after all renderings finish and after waiting few seconds for Garbage Collector we can see the `Used Heap Size` went down to 61-63MiB
		- [x] please note it didn't increase for the next 20MiB, so the max cache memory size is kept under control
		- [x] and we can also see that the `Used Heap Size` didn't increase from the last measurement (it was 61MiB and still is 61MiB). So I can only guess V8 warmed up enough
	OK, let's try now to render entries with numbers 601 to 638 (which should cause evicting entries with numbers 501-538). You'll need to wait another 70seconds
		- [x] `bash ./make-requests.sh --start-number 601 --range 38 && say "requests completed"`
		- [x] In the meantime the Used Heap Size can temporarily go up above 100-300MiB
	- [x] after all renderings finish and after waiting few seconds for Garbage Collector we can see the `Used Heap Size` went down to 61-63MiB
		- [x] please note it didn't increase for the next 20MiB, so the max cache memory size is kept under control
		- [x] and we can also see that the `Used Heap Size` didn't increase from the last measurement (it was 61MiB and still is 61MiB). So I can only guess V8 warmed up enough
	- [x] OK, let's try now to render entries with numbers 701 to 738 (which should cause evicting entries with numbers 601-638). You'll need to wait another 70seconds
		- [x] `bash ./make-requests.sh --start-number 701 --range 38 && say "requests completed"`
		- [x] In the meantime the Used Heap Size can temporarily go up above 100-300MiB
	- [x] after all renderings finish and after waiting few seconds for Garbage Collector we can see the `Used Heap Size` went down to 61-63MiB
		- [x] please note it didn't increase for the next 20MiB, so the max cache memory size is kept under control
		- [x] and we can also see that the `Used Heap Size` didn't increase from the last measurement (it was 61MiB and still is 61MiB). So I can only guess V8 warmed up enough
	OK, let's try now to render entries with numbers 801 to 838 (which should cause evicting entries with numbers 701-738). You'll need to wait another 70seconds
		- [x] `bash ./make-requests.sh --start-number 801 --range 38 && say "requests completed"`
		- [x] In the meantime the Used Heap Size can temporarily go up above 100-300MiB
	- [x] after all renderings finish and after waiting few seconds for Garbage Collector we can see the `Used Heap Size` went down to 61-63MiB
		- [x] please note it didn't increase for the next 20MiB, so the max cache memory size is kept under control
		- [x] and we can also see that the `Used Heap Size` didn't increase from the last measurement (it was 61MiB and still is 61MiB). So I can only guess V8 warmed up enough
	- [x] OK, let's try now to render entries with numbers 901 to 938 (which should cause evicting entries with numbers 801-838). You'll need to wait another 70seconds
		- [x] `bash ./make-requests.sh --start-number 901 --range 38 && say "requests completed"`
		- [x] In the meantime the Used Heap Size can temporarily go up above 100-300MiB
	- [x] after all renderings finish and after waiting few seconds for Garbage Collector we can see the `Used Heap Size` went down to 61-63MiB
		- [x] please note it didn't increase for the next 20MiB, so the max cache memory size is kept under control
		- [x] and we can also see that the `Used Heap Size` didn't increase from the last measurement (it was 61MiB and still is 61MiB). So I can only guess V8 warmed up enough
	- [x] OK, let's try now to render entries with numbers 1001 to 1038 (which should cause evicting entries with numbers 901-938). You'll need to wait another 70seconds
		- [x] `bash ./make-requests.sh --start-number 1001 --range 38 && say "requests completed"`
		- [x] In the meantime the Used Heap Size can temporarily go up above 100-300MiB
	- [x] at the end, stop SSR process and its monitoring
		- [ ] control+C in monitoring
		- [ ] `npm run pm2:delete`


core-libs/setup/ssr/optimized-engine/default-cache-entry-size-calculator.ts
```diff
@@ -0,0 +1,73 @@
...
+
+/**
+ * Default implementation of the cache entry size calculator.
+ *
+ * For HTML string, it returns the size of the string in bytes, assuming 2 bytes per each character
+ * (an upper-bound estimation assuming V8 is using `SeqTwoByteString` for string cache entries).
+ *
+ * For error object, it only roughly approximates the size - it sums sizes of the `name`, `message` and `stack` string properties.
+ *
+ * CAUTION: the error object can be potentially anything and can have more properties (not necessarily strings),
```
pawelfras:
```suggestion
 * ***CAUTION:*** the error object can be potentially anything and can have more properties (not necessarily strings),
```

core-libs/setup/ssr/optimized-engine/default-cache-entry-size-calculator.spec.ts
```diff
@@ -0,0 +1,43 @@
...
+          Buffer.byteLength(error.message, 'utf8') +
+          Buffer.byteLength(error.stack || '', 'utf8'));
+      expect(calculator.calculateSize(entry)).toBe(expectedSize);
+    });
+
+    it('should handle error with missing properties', () => {
+      const error = { message: 'test' }; // 4 chars
+      const entry: RenderingEntry = { err: error };
+      expect(calculator.calculateSize(entry)).toBe(8); // 4 chars * 2 bytes
+    });
```
pawelfras: `minor:` let's consider a test case for scenario, where error is defined, but it's e.g. an empty object
```typescript
const entry: RenderingEntry = { err: {} };
expect(calculator.calculateSize(entry)).toBe(0); // no expected properties
```

core-libs/setup/ssr/optimized-engine/index.ts
```diff
@@ -4,10 +4,12 @@
...
 
+export * from './default-cache-entry-size-calculator';
 export * from './optimized-ssr-engine';
 export * from './rendering-cache';
+export * from './rendering-cache-size-manager';
 export * from './rendering-cache.model';
 export * from './rendering-strategy-resolver';
 export * from './rendering-strategy-resolver-options';
-export { RequestContext, getRequestContext } from './request-context';
+export { getRequestContext, RequestContext } from './request-context';
```
pawelfras, line 14: less barrel files, better tree shaking - nice 👌

core-libs/setup/ssr/optimized-engine/ssr-optimization-options.ts
```diff
@@ -7,10 +7,19 @@
...
 import { defaultRenderingStrategyResolver } from './rendering-strategy-resolver';
 import { defaultRenderingStrategyResolverOptions } from './rendering-strategy-resolver-options';
 
+/**
+ * Custom strategy for calculating the size of a cache entry.
+ * It's needed to keep track of the used cache size.
+ */
+export interface CacheEntrySizeCalculator {
+  calculateSize(entry: RenderingEntry): number;
+}
```
pawelfras: `query:` Was this interface placed here intentionally? I suggest moving it to a separate file. TBH I'm surprised it doesn't cause any circular dependency issues
`./ssr-optimization-options` -> `./default-cache-entry-size-calculator` -> `./ssr-optimization-options` 🤔

core-libs/setup/ssr/optimized-engine/ssr-optimization-options.ts
```diff
@@ -37,9 +46,40 @@ export interface SsrOptimizationOptions {
...
 
+  /**
+   * Limits the cache size memory in bytes.
+   *
+   * Specifies the maximum memory (in bytes) allocated for cached entries,
+   * helping to keep memory usage under control.
+   *
+   * The default value is set to 800 MB (meaning 800 000 000 in International System of Units).
+   *
+   * IMPORTANT: Your server should have much more available memory than the configured `cacheSizeMemory`,
```
pawelfras:
```suggestion
   * ***IMPORTANT:*** Your server should have much more available memory than the configured `cacheSizeMemory`,
```

core-libs/setup/ssr/optimized-engine/ssr-optimization-options.ts
```diff
@@ -37,9 +46,40 @@ export interface SsrOptimizationOptions {
...
+   *            because the NodeJS process needs a lot of operational memory also for the rendering activities,
+   *            such as creating instances of the Angular applications for each incoming requests.
+   *            The more parallel requests are allowed (which can be limited with the `concurrency` option),
+   *            the more operational memory is needed.
+   *
+   * Note: for calculating the size of the cache entry, the `cacheEntrySizeCalculator` option is used.
```
pawelfras:
```suggestion
   * *Note:* for calculating the size of the cache entry, the `cacheEntrySizeCalculator` option is used.
```

core-libs/setup/ssr/optimized-engine/ssr-optimization-options.ts
```diff
@@ -37,9 +46,40 @@ export interface SsrOptimizationOptions {
...
+   */
+  cacheSizeMemory?: number;
+
+  /**
+   * Strategy for calculating the size of a cache entry. It's needed to keep track of the used cache size,
+   * so the oldest entries can be removed when the cache size memory limit is reached.
+   *
+   * Note: This config option is used only when the `ssrFeatureToggles.limitCacheByMemory` is set to true.
```
pawelfras:
```suggestion
   * *Note:* This config option is used only when the `ssrFeatureToggles.limitCacheByMemory` is set to true.
```

core-libs/setup/ssr/optimized-engine/index.ts
Platonn: Excuse me, I'm having problem to understand what you mean,
1. this line just got automatically sorted the properties alphabetically
2. Tree shaking of code that runs in the browser is important (for Web Vitals), but I don't think it's that important for the code that runs on the server.

core-libs/setup/ssr/optimized-engine/ssr-optimization-options.ts
Platonn: 👍 moved to `./rendering-cache.model.ts`

core-libs/setup/ssr/optimized-engine/default-cache-entry-size-calculator.spec.ts
Platonn: added

core-libs/setup/ssr/optimized-engine/ssr-optimization-options.ts
Platonn: done

core-libs/setup/ssr/optimized-engine/ssr-optimization-options.ts
Platonn: done

core-libs/setup/ssr/optimized-engine/ssr-optimization-options.ts
Platonn: done

core-libs/setup/ssr/optimized-engine/default-cache-entry-size-calculator.ts
Platonn: done

pawelfras (approved):
LGTM

PR #20253, opened 2025-04-30
chore: CXSPA-9963 restrict facade-descriptor

Pucek9:
Facade validation - all methods should return void or an Observable type only 

Closes [CXSPA-9963](https://jira.tools.sap/browse/CXSPA-9963)


Pucek9:
Currently applied a new restriction only for the method keys array (no issues for building) 



Type for Facade is also ready, but not set (commented) because it requires removing methods from code (breaking changes)

Pucek9:
Applied new restriction with 2 ts-ignore for existing facades with deprecated methods

rmch91 (approved):

PR #20255, opened 2025-05-01
Land on Login page when error

FollowTheFlo:
a- Removed punchout error page
b- Instead we now end punchout session by: logout, clear punchout storage, navigate to login page with authorization message.
c- added facade method 'endPunchoutSession()' to address 'b'.
d- fixed token revoking prob by setting login user with 'current' rather than the actual userId string.
e- better designed by silentLogout before calling punchout occ api (fixed probs that were triggered token expiration in specific use cases)
f- improve/simplified  accesstoken and refreshtoken expiration event by just ending punchout session.

### DIFF
--- a/integration-libs/punchout/core/facade/punchout.service.spec.ts
+++ b/integration-libs/punchout/core/facade/punchout.service.spec.ts
@@ -172,7 +172,7 @@ describe('Punchoutservice', () => {
     });
   });
 
-  it('should getPunchoutSessionRequisition with empty param opens error page', (done) => {
+  it('should getPunchoutSessionRequisition with empty param ends punchout session', (done) => {
     spyOn(routingService, 'go').and.returnValue(Promise.resolve(true));
     service.getPunchoutSession({ punchoutSessionId: '' }).subscribe({
       error: () => {
@@ -182,7 +182,7 @@ describe('Punchoutservice', () => {
     });
   });
 
-  it('should getPunchoutSessionRequisition without logged-in user opens error page', (done) => {
+  it('should getPunchoutSessionRequisition without logged-in user ends punchout session', (done) => {
     spyOn(routingService, 'go').and.returnValue(Promise.resolve(true));
     spyOn(punchoutAuthService, 'isUserLoggedIn').and.returnValue(of(false));
     service.getPunchoutSessionRequisition().subscribe({
@@ -208,7 +208,7 @@ describe('Punchoutservice', () => {
       });
   });
 
-  it('should getPunchoutSession opens error page when request failed', (done) => {
+  it('should getPunchoutSession ends punchout session when request failed', (done) => {
     spyOn(routingService, 'go').and.returnValue(Promise.resolve(true));
     spyOn(connector, 'getPunchoutSession').and.returnValue(
       of({ ...mockPunchoutSessionResponse, token: undefined })
@@ -221,7 +221,7 @@ describe('Punchoutservice', () => {
     });
   });
 
-  it('should getPunchoutSession opens error page when no auth token', (done) => {
+  it('should getPunchoutSession ends punchout session when no auth token', (done) => {
     spyOn(routingService, 'go').and.returnValue(Promise.resolve(true));
     spyOn(connector, 'getPunchoutSession').and.returnValue(
       of({ ...mockPunchoutSessionResponse, token: undefined })
@@ -293,7 +293,7 @@ describe('Punchoutservice', () => {
     });
   });
 
-  it('should logoutPunchoutUser calls punchoutAuthService logout method', (done) => {
+  it('should logoutPunchoutUser calls punchoutAuthService silentLogout method', (done) => {
     spyOn(connector, 'getPunchoutSessionRequisition').and.returnValue(
       of(mockPunchoutRequisitionResponse)
     );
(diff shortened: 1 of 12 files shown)


integration-libs/punchout/root/services/punchout-detection.service.ts
```diff
@@ -24,4 +26,7 @@ export class PunchoutDetectionService {
       urlSections[0].includes(PUNCHOUT_SESSION_PAGE_URL)
     );
   }
+  isPunchoutSession(): boolean | undefined {
```
FollowTheFlo, line 29: we voluntary don't check if user is logged-in because it is used in auth-http-header service where user already has an invalid/expired token.

integration-libs/punchout/core/services/punchout-auth.service.ts
```diff
@@ -47,19 +57,36 @@ export class PunchoutAuthService {
...
+  loginWithToken(accessToken: string): void {
     // Code mostly based on auth lib we use and the way it handles token properties
     this.authStorageService.setItem(ACCESS_TOKEN, accessToken);
     this.authStorageService.setItem(
       ACCESS_TOKEN_STORED_AT,
       Date.now().toString()
     );
-    this.userIdService.setUserId(userId);
+    this.authStorageService.setItem(TOKEN_TYPE, 'Bearer');
+    this.userIdService.setUserId(OCC_USER_ID_CURRENT);
```
FollowTheFlo, line 68: setting 'current' fixed all the 'token revoking failure': CXSPA-9871 Token not revoked when silently logout user.
It sems to be the standard practice to set current rather than user id string. Not sure about the reason.

integration-libs/punchout/core/services/punchout-auth.service.ts
```diff
@@ -47,19 +57,36 @@ export class PunchoutAuthService {
...
   }
+
+  endPunchoutSession(): void {
+    this.punchoutStoreService.clearPunchoutState();
+    this.globalMessageService.add(
+      { key: 'httpHandlers.unauthorized.common' },
+      GlobalMessageType.MSG_TYPE_ERROR
+    );
+
+    if (this.punchoutDetectionService.isPunchoutSessionPage()) {
```
FollowTheFlo, line 84: setting redirect to homepage if issue happens on punchout page.
we don't want the user to be redirected to punchout page after a manual login, user would get back to login page again, stuck in infinite loop until user navigates somewhere else.

integration-libs/punchout/core/facade/punchout.service.spec.ts
```diff
@@ -168,22 +172,22 @@ describe('Punchoutservice', () => {
     });
   });
 
-  it('should getPunchoutSessionRequisition with empty param opens error page', (done) => {
+  it('should getPunchoutSessionRequisition with empty param ends punchout session', (done) => {
     spyOn(routingService, 'go').and.returnValue(Promise.resolve(true));
     service.getPunchoutSession({ punchoutSessionId: '' }).subscribe({
       error: () => {
-        expect(routingService.go).toHaveBeenCalledWith(PUNCHOUT_ERROR_PAGE_URL);
+        expect(punchoutAuthService.endPunchoutSession).toHaveBeenCalled();
```
Pucek9: Better verification includes parameters - even void; it's not a big cost for us, but can point to much more behaviors
```suggestion
        expect(punchoutAuthService.endPunchoutSession).toHaveBeenCalledWith();
```

integration-libs/punchout/core/facade/punchout.service.spec.ts
```diff
@@ -168,22 +172,22 @@ describe('Punchoutservice', () => {
...
   });
 
-  it('should getPunchoutSessionRequisition without logged-in user opens error page', (done) => {
+  it('should getPunchoutSessionRequisition without logged-in user ends punchout session', (done) => {
     spyOn(routingService, 'go').and.returnValue(Promise.resolve(true));
     spyOn(punchoutAuthService, 'isUserLoggedIn').and.returnValue(of(false));
     service.getPunchoutSessionRequisition().subscribe({
       error: () => {
-        expect(routingService.go).toHaveBeenCalledWith(PUNCHOUT_ERROR_PAGE_URL);
+        expect(punchoutAuthService.endPunchoutSession).toHaveBeenCalled();
```
Pucek9, line 190:
```suggestion
        expect(punchoutAuthService.endPunchoutSession).toHaveBeenCalledWith(true);
```

integration-libs/punchout/core/facade/punchout.service.ts
```diff
@@ -151,16 +143,32 @@ export class PunchoutService implements PunchoutFacade {
...
+  protected logoutPunchoutUserCommand: Command<void, boolean> =
+    this.commandService.create(() => this.punchoutAuthService.silentLogout());
+
+  /**
+   * endUserSession workflow:
+   * clear punchout state,logout, redirect to login page, display message
+   */
+  protected endPunchoutSessionCommand: Command<void> =
+    this.commandService.create<void>(() =>
+      of(true).pipe(
```
Pucek9, line 166: is it necessary to hack it with of(true)?

FollowTheFlo: yes, with of(), `next: () =>{}`  is not triggered, it goes to `complete:()=>{}` directly
using EMPTY also completes the observable

PR #20273, opened 2025-05-07
fix: Punchout build throws circular dependency error

Pucek9:
- fix circular dependency
-  rename `PunchoutComponentsService` to `PunchoutUiRestrictionService` + move it to services

Closes [CXSPA-10023](https://jira.tools.sap/browse/CXSPA-10023)


FollowTheFlo (approved):
lgtm

PR #20274, opened 2025-05-07
chore: Improve Punchout Level logic

Pucek9:
Closes [CXSPA-10016](https://jira.tools.sap/browse/CXSPA-10016)


Pucek9:
Rejected as keyboarding is still accessible, and no way to avoid that.

FollowTheFlo (approved):
As discussed, happy with it

PR #20276, opened 2025-05-07
fix: Hide pdp link in Inspect Cart using CSS solution

Pucek9:
Closes [CXSPA-10039](https://jira.tools.sap/browse/CXSPA-10039)


Pucek9:
Rejected as keyboarding is still accessible, and no way to avoid that.

PR #20279, opened 2025-05-08
fix: Sanitize Redirect URL to Prevent Reflected XSS in Payment Form

anjana-bl:
https://jira.tools.sap/browse/CXSPA-10026

### DIFF
--- a/integration-libs/digital-payments/src/checkout/cms-components/dp-payment-method/dp-payment-form/dp-payment-form.component.ts
+++ b/integration-libs/digital-payments/src/checkout/cms-components/dp-payment-method/dp-payment-form/dp-payment-form.component.ts
@@ -6,12 +6,14 @@
 
 import { DpLocalStorageService } from '../../../facade/dp-local-storage.service';
 import {
+  ExternalRoutesGuard,
   GlobalMessageService,
   GlobalMessageType,
   WindowRef,
 } from '@spartacus/core';
 import { DpCheckoutPaymentService } from '../../../facade';
-import { Component, OnInit, Output, EventEmitter } from '@angular/core';
+import { Component, OnInit, Output, EventEmitter, inject } from '@angular/core';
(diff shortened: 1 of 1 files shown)


integration-libs/digital-payments/src/checkout/cms-components/dp-payment-method/dp-payment-form/dp-payment-form.component.ts
```diff
@@ -33,7 +36,10 @@ export class DpPaymentFormComponent implements OnInit {
     this.dpPaymentService.getCardRegistrationDetails().subscribe((request) => {
       if (request?.url) {
         this.dpStorageService.syncCardRegistrationState(request);
-        this.redirect(request.url);
+        this.externalRoutesGuard.canActivate(
+          {} as ActivatedRouteSnapshot,
+          request as RouterStateSnapshot
+        );
```
Platonn, line 42: ExternalRoutesGuard is meant for the External Routes feature, e.g. to navigate to an Accelerator page in a Hybrid Storefront (having some routes on Spartacus, some on Accelerator). Is this really our usecase in this PR?

IMHO it looks stange we call Angular Router Guard inside ngOnInit of component. Ideally Guards should be called only by the router.

Could you help me understand how this PR addresses the root cause of the issue reported in the original ticket https://jira.tools.sap/browse/CXSPA-10026 ? To my understanding, we're just replacing the old implementation with new implementation which does (nearly) exactly the same (the only thing that guard does additionally is checking `isBrowser` to avoid using `window` object in SSR context - which we should have done also here in DpPaymentForm, but forgot about it). 

What do you think should be the fix for the root cause of the problem reported in the ticket https://jira.tools.sap/browse/CXSPA-10026?

anjana-bl:
continuing this in https://github.com/SAP/spartacus/pull/20280

PR #20283, opened 2025-05-08
fix: CXSPA-9778 Return order request loads unshipped items, then fail…

andreighervan7:
closes [CXSPA-9778](https://jira.tools.sap/browse/CXSPA-9778)

The backend response provides an unreliable value for `returnableQuantity`, so a new observable was introduced to retrieve data related to consignments. The `shippingQuantity` value from the consignments observable is now used to determine the number of items eligible for return and to filter items that have already been shipped.

Additionally, the form is now reset after a successful change, and an error handler has been implemented to display a meaningful error message when the return quantity is invalid.


npapp-dev002:
### QA steps:
- in `feature-libs/order/components/amend-order/returns/return-order/return-order.component.ts` remove `consigments$` observable and replace the declaration of `entries$` observable with `entries$: Observable<OrderEntry[]> = this.orderAmendService.getEntries();`
- in `feature-libs/order/components/amend-order/returns/order-return.service.ts` remove `form.reset()` from `save()`method and place it somewhere before its original position
- create an order (for example, 2 from 816802 and 4 from 2054947)
- Go to the Backoffice service and set a specific number of product items to the "Shipped" status. Make sure the order itself is also updated to "Shipped".(for example, 2 from 2054947) 
- go back to Spartacus and select the order and click on 'return request' button

---
### Before:
All products, including those that are not eligible for return, can be seen on the return order page.
image
Error messages contains keys that have no translation: 
image
### After:
Correct number of elements are visible on the page:
image
The correct error message is visible if a return request has already been submitted.
image

rmch91 (approved):
LGTM

PR #20284, opened 2025-05-08
chore: remove authorizationCodeFlowDefault feature flag

ercultimate:


projects/storefrontapp/src/app/spartacus/spartacus-features.module.ts
```diff
@@ -286,6 +287,7 @@ if (environment.cpq) {
       provide: USE_MY_ACCOUNT_V2_NOTIFICATION_PREFERENCE,
       useValue: environment.myAccountV2,
     },
+    useAuthorizationCodeFlowByDefaultProvider(false),
```
Platonn: let's not over-provide defailt value explicitly here in the module
the factory `() => false` inside the injection token definition suffices.

now if customer calls `useAuthorizationCodeFlowByDefaultProvider(true)` before importing our module, our module will win - as the last provided, and it might be confusing for customers

for more see https://dev.to/krisplatis/the-importance-of-order-in-angular-dependency-injection-providers-4g31

projects/core/src/auth/user-auth/config/default-auth-config.ts
```diff
@@ -4,10 +4,41 @@
...
+ * Server only supports Authorization Code flow for public clients from
+ * that version and onwards.
+ *
+ * @usageNotes
+ * Add to the root module providers:
+ * ```
+ * useAuthorizationCodeFlowByDefaultProvider(true)
+ * ```
+ */
+export function useAuthorizationCodeFlowByDefaultProvider(
```
Platonn: Angular API (and we try to mimic it) uses by convention the word `provide` as the first word in such providers utils
Angular: e.g. `provideClientHydration()`, `provideRoutes()`, ...
Spartacus: e.g. `provideConfig()`, `provideFeatureToggles()`, ...

What about `provideAuthorizationCodeFlowByDefault()` ?

Platonn (dismissed):

projects/storefrontapp/src/app/spartacus/spartacus-features.module.ts
ercultimate: I think there is some confusion here.  The use highlighted here is in the storefrontapp and would be how the customer toggles the feature in their repo.

The token is not provided in any of our modules (and `useAuthorizationCodeFlowByDefaultProvider` is not used either), so the only default from the token's factory would apply normally.  I don't think we can overwrite the customer's specification currently.

kimhw0630 (approved):

PR #20294, opened 2025-05-12
fix: cxspa 10039 remove too strict selector

Pucek9:
Fix the issue for the app built by schematics 
Related to: https://jira.tools.sap/browse/CXSPA-10039


FollowTheFlo (approved):
lgtm

PR #20299, opened 2025-05-12
feat: added global messages for loading status (CXSPA-10017)

kimhw0630:
Closes: https://jira.tools.sap/browse/CXSPA-10017

### DIFF
--- a/integration-libs/punchout/components/punchout-session/punchout-session.component.ts
+++ b/integration-libs/punchout/components/punchout-session/punchout-session.component.ts
@@ -43,18 +43,8 @@ export class PunchoutSessionComponent implements OnInit {
         take(1)
       )
       .subscribe({
-        error: (error) => {
-          this.globalMessageService.add(
-            error.message
-              ? error.message
-              : { key: 'punchout.failToInitiateUserSession' },
-            GlobalMessageType.MSG_TYPE_ERROR
-          );
-        },
         complete: () => {
-          this.globalMessageService.remove(
-            GlobalMessageType.MSG_TYPE_CONFIRMATION
-          );
+          this.globalMessageService.remove(GlobalMessageType.MSG_TYPE_INFO);
         },
       });
   }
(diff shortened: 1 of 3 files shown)


integration-libs/punchout/components/punchout-session/punchout-session.component.ts
```diff
@@ -11,29 +11,51 @@ import {
...
+        error: (error) => {
+          this.globalMessageService.add(
+            error.message
+              ? error.message
+              : { key: 'punchout.failToInitiateUserSession' },
+            GlobalMessageType.MSG_TYPE_ERROR
+          );
+        },
+        complete: () => {
+          this.globalMessageService.remove(
```
FollowTheFlo: which message do we want to hide?
in punchoutFacade we already saliently logout and login, did you see some other unwanted messages?
did you mean MSG_TYPE_INFO to make  punchout.initiatingUserSession meassge disappear? it is actually what we need.

integration-libs/punchout/components/punchout-session/punchout-session.component.ts
```diff
@@ -11,29 +11,51 @@ import {
...
             punchoutSessionId: param?.[PUNCHOUT_SESSION_KEY],
-          })
-        ),
+          });
+        }),
         take(1)
       )
-      .subscribe();
+      .subscribe({
+        error: (error) => {
```
FollowTheFlo: i think we can remove the full error callback as we andling the error (global message and redirection) in 
this.punchoutFacade.getPunchoutSession

integration-libs/punchout/components/punchout-session/punchout-session.component.ts
kimhw0630: sorry  it should MSG_TYPE_INFO

integration-libs/punchout/components/punchout-session/punchout-session.component.ts
kimhw0630: i will take look and if we need to remove error handler then i will thanks.

FollowTheFlo (approved):
LGTM

PR #20301, opened 2025-05-12
fix: invalid token throwing error

FollowTheFlo:
steps to replicate:
- set invalid token on previous instance, eg: login with standard user, modify access token in storage, close browser
- start punchout session
---> inside Punchout Session page, user gets redirected to Login page before even fetch punchout occ api.
technically: when authorisation error occurs, we always end punchout session (logout, redirect to login page)

expected:
Punchout session to be initialized.
technically expected: silently logout user (no redirection) when authorization error occurs on punchout session page. PunchoutSessionComponent will take care of initializing punchout session.

### DIFF
--- a/integration-libs/punchout/root/services/punchout-auth-http-header.service.ts
+++ b/integration-libs/punchout/root/services/punchout-auth-http-header.service.ts
@@ -4,20 +4,17 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { HttpEvent, HttpHandler, HttpRequest } from '@angular/common/http';
 import { inject, Injectable } from '@angular/core';
 import {
   AuthHttpHeaderService,
   AuthRedirectService,
   AuthService,
   AuthStorageService,
-  AuthToken,
   GlobalMessageService,
   OAuthLibWrapperService,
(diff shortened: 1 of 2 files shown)


integration-libs/punchout/root/services/punchout-auth-http-header.service.ts
```diff
@@ -50,37 +47,21 @@ export class PunchoutAuthHttpHeaderService extends AuthHttpHeaderService {
...
+      return;
     }
-  }
-
-  /**
-   * @override
-   * Refreshes access_token and then retries the call with the new token.
-   * When punchout session exists,  punchout session gets ended, user is redirected to login page
-   */
-  public handleExpiredAccessToken(
```
FollowTheFlo, line 73: Redundancy. 
We don't override handleExpiredAccessToken anymore.
UI actions (logout, end session) need to be done on handleExpiredRefreshToken (not handleExpiredAccessToken which is an interceptor)
It is not needed anymore since we refactored punchoutService.getPunchoutSession where we now always logout user before fetching occ api punchout session

Pucek9 (approved):

PR #20304, opened 2025-05-13
chore: adjust auth server endpoints to match server values

ercultimate:
- update configured endpoints
- fix reverted default configuration option
- fix revoke endpoint error


projects/core/src/auth/user-auth/config/default-auth-config.ts
```diff
@@ -52,9 +51,10 @@ export const defaultAuthConfig: AuthConfig = {
       customTokenParameters: ['token_type'],
       strictDiscoveryDocumentValidation: false,
       skipIssuerCheck: true,
-      disablePKCE: true,
+      disablePKCE: false,
```
rmch91, line 54: I think changes in default config values can lead to unexpected behaviour. 
Seems to be a breaking change, isn't it?

rmch91 (changes requested):

projects/core/src/auth/user-auth/config/default-auth-config.ts
```diff
@@ -52,9 +51,10 @@ export const defaultAuthConfig: AuthConfig = {
...
       oidc: false,
       clearHashAfterLogin: false,
+      responseType: 'code',
```
rmch91, line 57: same question as above

rmch91 (changes requested):

projects/core/src/auth/user-auth/config/default-auth-config.ts
ercultimate: This default object is internal, so changes will not affect customers.  It is not exported out of the module and is only used when applied in the `UserAuthModule`.  Even this is not strictly true anymore now that we use a temporary provider factory to set the default configuration, but we will eventually return to using the object directly.

The changes to the default object are to reduce ambiguity when the JDK17 authorization server support is dropped. 
 The default object is set to the "future" JDK21 default, and the provider factory converts the object to match the "legacy" version.

projects/core/src/auth/user-auth/config/auth-config.ts
```diff
@@ -55,6 +55,10 @@ export abstract class AuthConfig {
      * Endpoint url for revoking tokens.
      */
     revokeEndpoint?: string;
+    /**
+     * Determine if the `Authorization` header should be sent with revocation requests.
+     */
+    enableTokenRevocationInterceptor?: boolean;
```
Platonn: just nitpicks: 

1. the name of this config describes the implementation (e.g. mentions the concept of interceptor).
Other config properties are related to the characteristics of the auth standard/process itself.

What do you think about renaming the config property, to rather describe the effect on the auth process?
e.g. `shouldSendAuthHeaderOnRevoke`

(If only you agree, then let's rename also analogical places, like the method in AuthConfigService)

2. If this config property is needed only for JDK21, for clarity let's mention this fact also in this the JSDoc of this config property

projects/core/src/auth/user-auth/config/default-auth-config.ts
rmch91: If, in customer's application, customer provided authConfig with only this values (just a dummy example):
```
 provideConfig(<AuthConfig>{
      authentication: {
        clientId: 'spartacus-spa',
        OAuthLibConfig: {
          strictDiscoveryDocumentValidation: true,
        },
      },
    }),
```
Customer provided only these values, because the rest values from the default config are suitable for him,
in the result his config will looks like this:
```
const oAuthConfig = {
  OAuthLibConfig: {
    clearHashAfterLogin: false,
    customTokenParameters: ['token_type'],
    disablePKCE: true,
    oidc: false,
    scope: '',
    skipIssuerCheck: true,
    strictDiscoveryDocumentValidation: true, // From CUSTOM config
  },
  clientId: 'spartacus-spa', // From CUSTOM config
  client_id: 'mobile_android',
  client_secret: 'secret',
  loginUrl: '/oauth/authorize',
  revokeEndpoint: '/oauth/revoke',
  tokenEndpoint: '/oauth/token',
};
```
Are we sure that changes in the default config that we are doing for 100% cannot affect his application?
Can customer's app broke because of not expected `resposeType: code` for example?

ercultimate: The factory emits the default configuration exactly as it did before this change.  This is proven by the spec test that was added for the factory.  The normal merging rules applying, so the customer's property will be merged into the configuration with a higher priority.

The only thing that could break something is the new property `sendAuthHeaderOnRevoke` conflicting with the customer's custom properties.

rmch91: new property `sendAuthHeaderOnRevoke` and removal of `responseType;` - `delete config.authentication.OAuthLibConfig.responseType`, and changing value of `disablePKCE`, no?
I can miss the context here, talking more in general about configs, than about this exact case. So if you think that it is safe change - feel free to ignore me

rmch91 (approved):
LGTM

PR #20312, opened 2025-05-14
feat: hide write review button (CXSPA-10137)

kimhw0630:
Closes: https://jira.tools.sap/browse/CXSPA-10137


FollowTheFlo (approved):
happy with it, tested successfully

PR #20314, opened 2025-05-14
chore: prevent manual navigation to Requisition page

Pucek9:
Closes: [CXSPA-10019](https://jira.tools.sap/browse/CXSPA-10019)

### DIFF
--- a/integration-libs/punchout/components/punchout-requisition/punchout-requisition.component.ts
+++ b/integration-libs/punchout/components/punchout-requisition/punchout-requisition.component.ts
@@ -61,12 +61,6 @@ export class PunchoutRequisitionComponent implements OnInit, OnDestroy {
     );
 
   ngOnInit(): void {
-    this.globalMessageService.add(
-      {
-        key: 'punchout.redirectToProcurementSystem',
-      },
-      GlobalMessageType.MSG_TYPE_INFO
-    );
     this.punchoutFormGroup = this.formBuilder.group({
       [this.FORM_CONTROL_NAME.ORDER]: [''],
     });
--- a/integration-libs/punchout/core/facade/punchout.service.ts
+++ b/integration-libs/punchout/core/facade/punchout.service.ts
@@ -9,6 +9,8 @@ import { inject, Injectable } from '@angular/core';
 import {
   Command,
   CommandService,
+  GlobalMessageService,
+  GlobalMessageType,
   RoutingService,
   UserIdService,
 } from '@spartacus/core';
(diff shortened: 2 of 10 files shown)


integration-libs/punchout/components/punchout-requisition/punchout-requisition.component.ts
```diff
@@ -61,12 +61,6 @@ export class PunchoutRequisitionComponent implements OnInit, OnDestroy {
...
   ngOnInit(): void {
-    this.globalMessageService.add(
-      {
-        key: 'punchout.redirectToProcurementSystem',
-      },
-      GlobalMessageType.MSG_TYPE_INFO
-    );
     this.punchoutFormGroup = this.formBuilder.group({
       [this.FORM_CONTROL_NAME.ORDER]: [''],
     });
```
Pucek9, line 66: do we really need that (here)?

``` 
 ngOnDestroy(): void {
    this.globalMessageService.remove(GlobalMessageType.MSG_TYPE_INFO);
  }
```

integration-libs/punchout/core/facade/punchout.service.ts
```diff
@@ -127,6 +130,20 @@ export class PunchoutService implements PunchoutFacade {
       }),
       take(1),
       switchMap((punchoutState: PunchoutState) => {
+        // prevent manual navigation to requisition page
+        if (
+          punchoutState?.closePunchoutSession === undefined &&
+          punchoutState?.cancelRequisition === undefined
+        ) {
```
Pucek9, line 137: I didn't add new properties, just observed possible condition can catch an event - is there any other edge case that I don't know, and it causes this condition can't be used here?

FollowTheFlo: All use cases works fine for me. great! i haven't thought that boolean can actually have 3 values (undefined,true,false), so no need to add more properties.

integration-libs/punchout/components/punchout-requisition/punchout-requisition.component.ts
FollowTheFlo: i aggree, no need to keep the message removal.
use case 1: when success, we leave spartacus
use case2: if failure, land in login page, info message would disappear after 3 secs automatically.

integration-libs/punchout/core/facade/punchout.service.ts
```diff
@@ -269,6 +286,11 @@ export class PunchoutService implements PunchoutFacade {
     return this.requestPunchoutSessionCommand.execute(punchoutSessionId);
   }
 
+  submitRequisition(cancelRequisition: boolean): void {
```
FollowTheFlo, line 289: i always thought facade need to return Observable, if it works fine with void that's better.

FollowTheFlo (approved):
lgtm

integration-libs/punchout/core/facade/punchout.service.ts
Pucek9: Yup, observable or void only, feel free to see my PR about restrict facade-descriptor: https://github.com/SAP/spartacus/pull/20253/files

PR #20317, opened 2025-05-15
feat: Punchout

Pucek9:
Closes [CXSPA-254](https://jira.tools.sap/browse/CXSPA-254)


feature-libs/cart/base/components/cart-shared/cart-item-list-row/cart-item-list-row.component.html
```diff
@@ -10,6 +10,7 @@
     <div class="cx-table-item-container">
       <!-- Item Image -->
       <a
+        *ngIf="!options.disableItemLink"
```
FollowTheFlo, line 13: for info: 
not considered as breaking change as added property is optional and DOM structure stayed unchanged for existing customers.

feature-libs/cart/base/root/models/cart.model.ts
```diff
@@ -118,6 +118,7 @@ export interface CartItemComponentOptions {
   displayAddToCart?: boolean;
   addToCartString?: string;
   cartType?: CartType;
+  disableItemLink?: boolean;
```
FollowTheFlo, line 121: same as above
not considered as breaking change as added property is optional and DOM structure stayed unchanged for existing customers.

FollowTheFlo (approved):
fulfill all  FF requirements.

PR #20321, opened 2025-05-15
fix: logout page accessible during punchout session

Pucek9:
Closes [CXSPA-10176](https://jira.tools.sap/browse/CXSPA-10176)


integration-libs/punchout/root/guards/punchout-navigation.guard.ts
```diff
@@ -149,6 +147,9 @@ export class PunchoutNavigationGuard {
       take(1),
       switchMap((punchoutState: PunchoutState | undefined) => {
         if (punchoutState?.punchoutSessionId) {
+          if (cxRoute === 'logout') {
```
FollowTheFlo: put it in var same way as HOME_PAGE_URL, eg:
`protected readonly LOGOUT_CX_ROUTE='logout';`

integration-libs/punchout/root/guards/punchout-navigation.guard.ts
```diff
@@ -149,6 +147,9 @@ export class PunchoutNavigationGuard {
...
+            this.handleWarning();
```
FollowTheFlo, line 153: as discussed i'll need to test first, i'm having hard time to understand the work flow. if logout guard run before, should isLoggedIn have value false?

FollowTheFlo: my guess is that we want to catch it before 'this.punchoutFacade.requestPunchoutSession' run?

Pucek9: ---

In Angular, **route guards** (`CanActivate`, `CanActivateChild`, `CanLoad`, `CanMatch`, `CanDeactivate`) are executed **sequentially (one after another)** – **not in parallel**.

### Execution Order:

1. **Sequential Execution**:

    * If you have multiple guards on a route (e.g., `canActivate: [AuthGuard, RoleGuard]`), Angular will execute them **in the order they're listed**.
    * If any guard returns `false` or a `UrlTree` (for redirection), **subsequent guards will not be executed**.

2. **Return Types**:
   A guard can return:

    * `boolean` – allow or block navigation;
    * `UrlTree` – redirect;
    * `Observable<boolean | UrlTree>`;
    * `Promise<boolean | UrlTree>`.

   Angular waits for the result before continuing to the next guard.

3. **Common Use Cases**:

    * `CanActivate` – protect a specific route;
    * `CanActivateChild` – protect child routes;
    * `CanLoad` – prevent lazy-loaded module from loading;
    * `CanMatch` – conditionally match a route (Angular 15+);
    * `CanDeactivate` – warn or prevent the user from leaving a route/component.

---

If `AuthGuard` returns `false` or a `UrlTree`, `RoleGuard` **won’t be called at all**.

---

### Summary:

* Guards are executed **one by one**, **not in parallel**.
* Angular waits for each guard’s result before proceeding.
* A single guard can block the entire navigation flow.

FollowTheFlo: cool! would you mind to add an explanation comment above the condition
eg:
// handle race condition with logout guard, it needs to be assessed before punchout session is requested

Pucek9:
replaced by https://github.com/SAP/spartacus/pull/20338

PR #20325, opened 2025-05-16
fix: resolve reverse tabnabbing in order attachment window

anjana-bl:


anjana-bl:
https://github.com/SAP/spartacus/pull/20324 will be backported to develop after release

PR #20331, opened 2025-05-16
fix: CXSPA-10184 Punchout Inspect cart page header too wide

Pucek9:
Closes [CXSPA-10184](https://jira.tools.sap/browse/CXSPA-10184)


FollowTheFlo (dismissed):
lgtm

Pucek9:
Replaced by https://github.com/SAP/spartacus/pull/20337

integration-libs/punchout/styles/_punchout-template.scss
```diff
@@ -17,7 +19,9 @@
   }
 
   .SiteContext {
-    padding-inline-start: 1.5rem;
+    $padding: max(1.5rem, calc((100vw - var(--cx-page-width-max)) / 2));
+    padding-inline-start: $padding;
+    padding-inline-end: $padding;
```
tobi-or-not-tobi, line 24: this would be the same:

```css
padding-inline: max(1.5rem, calc((100vw - var(--cx-page-width-max)) / 2));
```

PR #20332, opened 2025-05-19
fix: removing the extra <div> introduced by PR #20204

anjana-bl:
This PR fixes an issue introduced in [PR #20204](https://github.com/SAP/spartacus/pull/20204), where the following conditional template was removed:
```
<div
  *ngIf="!useExtractedBillingAddressComponent; else showExtractedBillingAddressForm"
  class="cx-payment-form-billing"
>
</div>
<ng-template #showExtractedBillingAddressForm>
  <cx-checkout-billing-address-form></cx-checkout-billing-address-form>
</ng-template>
```
The intention was to always show the content previously rendered in the else block. However, the extracted component `<cx-checkout-billing-address-form>` was mistakenly wrapped in a `<div>`, assuming the same structure would be maintained. Since Angular does not wrap the contents of an <ng-template> with a parent element, this change unintentionally introduced an extra wrapper element.
This led to a CI pipeline failure on the `develop` branch due to inconsistent DOM structure and test expectations.

This PR corrects that by removing the unnecessary `<div>` wrapper and directly rendering `<cx-checkout-billing-address form>` as originally intended.

### DIFF
--- a/feature-libs/checkout/base/components/checkout-payment-method/checkout-payment-form/checkout-payment-form.component.html
+++ b/feature-libs/checkout/base/components/checkout-payment-method/checkout-payment-form/checkout-payment-form.component.html
@@ -1,10 +1,6 @@
 <!-- FORM -->
 <ng-container *ngIf="!loading; else spinner">
-  <div
-    *cxFeature="'!a11yRemoveStatusLoadedRole'"
-    role="status"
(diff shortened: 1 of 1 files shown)


pawelfras (approved):
Good catch, thank you. LGTM

PR #20337, opened 2025-05-19
fix: CXSPA-10184 Punchout Inspect cart page header too wide

Pucek9:
Closes [CXSPA-10184](https://jira.tools.sap/browse/CXSPA-10184)


FollowTheFlo (approved):
lgtm

PR #20347, opened 2025-05-21
fix: CXSPA-3698 remove asm header after failed login as customer

i53577:
If customer want to perform any action on asm UI, the asm UI will be collapsed and a global error message will appear.
Screenshot 2025-05-21 at 4 48 38 PM


feature-libs/asm/core/store/reducers/asm-ui.reducer.ts
```diff
@@ -18,6 +18,12 @@ export function reducer(state: AsmUi = initialState, action: Action): AsmUi {
         ...(action as AsmActions.AsmUiUpdate).payload,
       };
     }
+    case AsmActions.CUSTOMER_SEARCH_FAIL: {
```
i53577, line 21: searching by id or order will trigger this action

feature-libs/asm/components/customer-list/customer-list.component.ts
```diff
@@ -121,6 +121,14 @@ export class CustomerListComponent implements OnInit, OnDestroy {
       this.asmCustomerListFacade.getCustomerListsState().pipe(
         tap((state) => (this.listsError = !!state.error)),
         map((state) => {
+          if (
+            state.error &&
+            typeof state.error === 'object' &&
+            'status' in state.error &&
+            state.error.status === 403
+          ) {
+            this.launchDialogService.closeDialog('forbidden');
```
i53577: trying to get customer list will trigger this logic

feature-libs/asm/components/asm-main-ui/asm-main-ui.component.ts
```diff
@@ -125,6 +125,13 @@ export class AsmMainUiComponent implements OnInit, OnDestroy {
           uiState.collapsed === undefined ? false : uiState.collapsed
         )
       );
+    this.asmService.getAsmUiState().subscribe((state) => {
+      if (state.status === 403) {
```
scarai-sap: avoid using magic number 403.

feature-libs/asm/root/model/asm.models.ts
```diff
@@ -23,6 +23,7 @@ export interface CustomerSearchOptions {
 }
 
 export interface AsmUi {
+  status?: number;
```
scarai-sap, line 26: maybe rename it as statusCode as it's number in here?

feature-libs/asm/components/asm-main-ui/asm-main-ui.component.ts
```diff
@@ -165,6 +172,10 @@ export class AsmMainUiComponent implements OnInit, OnDestroy {
                 });
             }
           }
+          if (result === 'forbidden') {
+            this.logout();
+            this.hideUi();
```
scarai-sap, line 180: why not calling. "        this.asmService.updateAsmUiState({ status: undefined });"?

Maybe consider we have a routine for 

logout/hideUI/set status?

feature-libs/asm/components/customer-list/customer-list.component.ts
```diff
@@ -121,6 +121,14 @@ export class CustomerListComponent implements OnInit, OnDestroy {
       this.asmCustomerListFacade.getCustomerListsState().pipe(
         tap((state) => (this.listsError = !!state.error)),
         map((state) => {
+          if (
+            state.error &&
+            typeof state.error === 'object' &&
+            'status' in state.error &&
+            state.error.status === 403
```
scarai-sap: magic 403 again.

feature-libs/asm/components/asm-main-ui/asm-main-ui.component.ts
```diff
@@ -165,6 +172,10 @@ export class AsmMainUiComponent implements OnInit, OnDestroy {
                 });
             }
           }
+          if (result === 'forbidden') {
```
scarai-sap: do we have chance to define the forbidden to some macro?

feature-libs/asm/root/model/asm.models.ts
i53577: status comes from HttpErrorModel, it will be set automatically.
I tried to rename it as statusCode, but found it will lose the value in this condition.
https://github.com/SAP/spartacus/blob/develop/projects/core/src/model/misc.model.ts#L102
```
export class HttpErrorModel {
  message?: string;
  status?: number;
  statusText?: string;
  url?: string | null;
  details?: ErrorModel[];
}
```

feature-libs/asm/components/asm-main-ui/asm-main-ui.component.ts
i53577: Because in this case asmUiState will not change. 
If it will change, then no need to have this logic, the following is enough ;)
https://github.com/SAP/spartacus/pull/20347/files#diff-66b2c827a63edda851ae6eb6bd42c439484e7ea914a19447ca249631fa76e9dfR131
```
    this.asmService.getAsmUiState().subscribe((state) => {
      if (state.status === this.forbiddenResponseStatus) {
        this.logout();
        this.hideUi();
        this.asmService.updateAsmUiState({ status: undefined });
      }
    });
```

feature-libs/asm/components/customer-list/customer-list.component.ts
i53577: fixed, please check latest code

feature-libs/asm/components/asm-main-ui/asm-main-ui.component.ts
i53577: What about defining it as
```
export const CLOSE_DIALOG_REASON = {
  FORBIDDEN: 'forbidden'
};

scarai-sap (approved):

PR #20352, opened 2025-05-21
fix: Punchout stuck on login page with Secure Portal (CXSPA-10185)

Pucek9:
closes: [CXSPA-10185](https://jira.tools.sap/browse/CXSPA-10185)


integration-libs/punchout/root/model/punchout.model.ts
```diff
@@ -6,9 +6,9 @@
 
 export const PUNCHOUT_SESSION_KEY = 'sid';
 export const PUNCHOUT_SESSION_ID = 'punchoutSessionId';
-export const PUNCHOUT_SESSION_PAGE_URL = '/punchout/cxml/session';
```
FollowTheFlo, line 9: make sure it does not break the navigation guard when we compare the path

integration-libs/punchout/root/punchout.root.module.ts
```diff
@@ -71,6 +74,17 @@ export function punchoutStatePersistenceFactory(): () => void {
           punchoutComponentsService.init(compRef);
       },
     },
+    provideDefaultConfig({
```
FollowTheFlo: we should move it in separate file like we did in OPF (t/root/config/default-opf-checkout-routing-config.ts), we can have 'root/config/default-punchout-routing-config.ts'

Pucek9: yeah I was considering it, right

integration-libs/punchout/root/model/punchout.model.ts
Pucek9: all good, it checks do wider string includes thin

FollowTheFlo (approved):
good to me

PR #20353, opened 2025-05-21
chore: inject REQUEST optionally in TestConfigServerModule to avoid DI error during prerendering of storefrontapp

Platonn:
It's to avoid Dependency Injection error during the prerendering of storefrontapp, that imports `TestConfigServerModule`.

It's related only to our custom mechanism for injecting config chunks dynamically in SSR e2e tests. It's not affecting customers.

fixes https://jira.tools.sap/browse/CXSPA-10222

---

**QA steps:** 

Check that the SSR e2e tests pass:

- `npm run build:libs`
- `npm run build`
- `npm run build:ssr:local-http-backend`
- `npm run test:ssr`

PR #20356, opened 2025-05-21
fix: inject PROPAGATE_ERROR_TO_SERVER optionally to avoid DI error during prerendering

Platonn:
It's to avoid Dependency Injection error during the prerendering, where `PROPAGATE_ERROR_TO_SERVER` is not provided by default.

It fixes both our storefrontapp and customers' apps

fixes https://jira.tools.sap/browse/CXSPA-10221

---

**QA steps:** 

1. check that the feature of `PROPAGATE_ERROR_TO_SERVER` still works in SSR:

- `npm run build:libs`
- `npm run build`
- `npm run build:ssr:local-http-backend`
- `npm run test:ssr`

2. check that the prerendering starts without issues:
Prerequisite: you might need this PR changes being merged/copy-pasted locally: https://github.com/SAP/spartacus/pull/20353

- in file `projects/storefrontapp/prerender.txt` leave just 1 URL instead of 12 URLS (to speed up testing)
- `npm run build:libs` (if needed)
- `npm run build`
- `SERVER_REQUEST_ORIGIN='http://localhost:4200' npm run prerender:dev`


pawelfras (approved):
LGTM

PR #20360, opened 2025-05-22
fix: (CXSPA-10168) - Future stock - Access Continuum test

Pio-Bar:
Closes: [CXSPA-10168](https://jira.tools.sap/browse/CXSPA-10168)


StanislavSukhanov (changes requested):
image

Hi @Pio-Bar. Tested in b2b - works fine. But fails when I run it in electronics SPA context. 
Maybe it should be run in B2B context?

Zeyber:
> image
> Hi @Pio-Bar. Tested in b2b - works fine. But fails when I run it in electronics SPA context. Maybe it should be run in B2B context?

This should be fine because Future Stock is a b2b-only feature.

ercultimate (approved):

PR #20364, opened 2025-05-22
fix: rename getPunchoutSession to initPunchoutSession

kimhw0630:
Closes: https://jira.tools.sap/browse/CXSPA-9873

we decided to rename only the facade method getPunchoutSession to initPunchoutSession


FollowTheFlo (approved):
lgtm

PR #20365, opened 2025-05-22
fix: improve cart page

FollowTheFlo:


Pucek9:
seems to be duplicate of https://github.com/SAP/spartacus/pull/20366

PR #20366, opened 2025-05-22
fix: add punchout error page

FollowTheFlo:


integration-libs/punchout/components/punchout-error/punchout-error.component.html
```diff
@@ -0,0 +1,9 @@
+<div class="alert">
+  <span class="alert-icon">
+    <cx-icon
+      class="cx-icon fas fa-exclamation-circle"
```
Pucek9:
```suggestion
      [type]="ICON_TYPE.ERROR"
```

integration-libs/punchout/components/punchout-error/punchout-error.component.ts
```diff
@@ -8,8 +8,7 @@ import { Component } from '@angular/core';
 
 @Component({
   selector: 'cx-punchout-error',
-  template: `<h2>Punchout Error</h2>
-    <p>Please navigate back to Ariba</p> `,
+  templateUrl: './punchout-error.component.html',
   standalone: false,
 })
 export class PunchoutErrorComponent {}
```
Pucek9:
```suggestion
export class PunchoutErrorComponent {
  ICON_TYPE = ICON_TYPE;
}
```

integration-libs/punchout/components/punchout-error/punchout-error.component.ts
```diff
@@ -8,8 +8,7 @@ import { Component } from '@angular/core';
```
Pucek9:
```suggestion
import { ICON_TYPE } from '@spartacus/storefront';

```

Pucek9:
Are we able to add pages like this as CMS-paragraph (I mean mainly the icon)? If yes, it's worth considering making it with sampledata only

integration-libs/punchout/components/punchout-error/punchout-error.component.ts
FollowTheFlo: thx!

integration-libs/punchout/components/punchout-components.module.ts
```diff
@@ -9,40 +9,41 @@ import { NgModule } from '@angular/core';
 import { ReactiveFormsModule } from '@angular/forms';
 import { CartSharedModule } from '@spartacus/cart/base/components';
 import { CmsConfig, I18nModule, provideDefaultConfig } from '@spartacus/core';
+import { IconModule } from '@spartacus/storefront';
```
FollowTheFlo, line 12: still needed as punchout error paragrap cpnt contains cx-icon

FollowTheFlo:
> Are we able to add pages like this as CMS-paragraph (I mean mainly the icon)? If yes, it's worth considering making it with sampledata only
consulted GC, he mentioned that any alternative to 'unsanitized data injection' (what CMS Paragraph does) should be used, so we should keep the punchoutErrorCpnt template for this reason.

Pucek9 (approved):

PR #20378, opened 2025-05-27
feat: Make punchout allowed pages configurable (CXSPA-10240)

Pucek9:
Closes [CXSPA-10240](https://jira.tools.sap/browse/CXSPA-10240)


integration-libs/punchout/root/config/punchout-guard-config.ts
```diff
@@ -0,0 +1,38 @@
...
+import { Config } from '@spartacus/core';
+import { LaunchRoute } from '@spartacus/storefront';
+import { PunchOutOperation } from '../model';
+
+@Injectable({
+  providedIn: 'root',
+  useExisting: Config,
+})
+export abstract class PunchoutNavigationGuardConfig {
+  punchoutNavigation?: {
```
FollowTheFlo, line 17: In my understanding, punchoutNavigation property is needed as we are using dynamic keys (eg:[PunchOutOperation.EDIT]) and those cannot be at root level.

integration-libs/punchout/root/config/default-punchout-guard-config.ts
```diff
@@ -0,0 +1,45 @@
+/*
+ * SPDX-FileCopyrightText: 2025 SAP Spartacus team <spartacus-team@sap.com>
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { PunchoutNavigationGuardConfig } from './punchout-guard-config';
+import { PunchOutOperation } from '../model';
+
```
FollowTheFlo, line 9: we should create
punchout-constant.ts in /config folder with those paths to avoid duplication.

integration-libs/punchout/root/config/punchout-guard-config.ts
Pucek9: Generally, feature configs are wrapped to not expose properties directly in the root of the configuration

integration-libs/punchout/root/guards/punchout-navigation.guard.ts
```diff
@@ -30,18 +31,22 @@ export class PunchoutNavigationGuard {
...
 
   canActivate(
     route: CmsActivatedRouteSnapshot,
     _state: RouterStateSnapshot
   ): Observable<GuardResult> {
-    const cxRoute = route.data.cxRoute;
+    const relativeUrl = `${route.url.map((u) => u.path).join('/')}`;
+    const cxRoute =
+      route?.data?.cxRoute ??
+      this.routingConfigService.getRouteName(relativeUrl);
```
Pucek9, line 43: Thanks @Platonn !

FollowTheFlo (dismissed):
good to me.

FollowTheFlo (approved):
good to me

PR #20385, opened 2025-05-28
fix: (CXEC-16834) fix read radio button text

Smudo99:


Smudo99:
This solution will cause other problems, so it is not the best way.

PR #20390, opened 2025-05-28
fix: No Sufficient Permissions message should be a part of Punchout chunk (CXSPA-10255)

Pucek9:
Closes [CXSPA-10255 ](https://jira.tools.sap/browse/CXSPA-10255)


FollowTheFlo (approved):
lgtm

PR #20392, opened 2025-05-29
chore: patch 'node_modules/@angular/build' via "npm postinstall" script to fix angular bug causing 'npm run start' to build app unnecessarily twice

Platonn:
**Context:**
After we merged the PR https://github.com/SAP/spartacus/pull/20242 with the new 'application' builder, the `npm run start` command started to run the build unnecessarily twice. 

It's a known bug in Angular CLI (@angular/build) that was fixed only in v19.2.11 
- see changelong: https://github.com/angular/angular-cli/blob/main/CHANGELOG.md#19211-2025-05-07
- see bugfix: https://github.com/angular/angular-cli/commit/b43da39499ca477a896f7f957debb05ceed1372a

...but we're still on 19.0.4 and we cannot bump a minor without strong justification for our customers - becasue then we would need to increase our minimal required version, which is a breaking change and should be a last resort option used only in very justified cases e.g. crucial security patch.

**Solution:**
Implemented a custom "npm postinstall" script that mimics `patch-package` behavior (but we didn't want to use `patch-package` due to being 2 years old since its last release). The script patches a single line inside the `node_modules/@angular/build` - same line that was fixed by the Angular Team in v19.2.11. Thanks to this workaround, we can benefit from this DX fix in our monorepo, while developing Spartacus. But we're not increasing our minimal required angualr version for customers. And it shouldn't affect our customers.

**NOTE**: This script is just a temporary workaround that should be removed in the future (likely in our 2026 Framework Upgrade Release), when we'll bump Angular version in our monorepo to v21.0.0 (which will already include the bugfix from v19.2.11).

fixes https://jira.tools.sap/browse/CXSPA-10261

---

**QA steps**:
I) OLD behavior:
1. checkout to `develop` branch
2. run `npm install`
3. run `npm run start` and observe carefully the terminal. Please observe an unnecessary automatic Rebuild after the initial Build finishes

II) NEW behavior now:
1. checkout to this PR's branch
2. run `npm install` and observe in the terminal the following message: 
    ```
    > storefrontapp@0.0.0 postinstall
    > ts-node scripts/npm-postinstall/patch-@angular-build-19.0.4.ts
    
    ☑️  Spartacus npm postinstall: Successfully patched file 'node_modules/@angular/build/src/tools/esbuild/bundler-context.js'
    ```
3. run `npm run start` and observe carefully the terminal. Please observe there is NO automatic Rebuild after the initial Build finishes.
4. run `npm install` again and observe in the terminal the following message:
    ```
    > storefrontapp@0.0.0 postinstall
    > ts-node scripts/npm-postinstall/patch-@angular-build-19.0.4.ts
    
    ℹ️  Spartacus npm postinstall: Patch already applied to 'node_modules/@angular/build/src/tools/esbuild/bundler-context.js'
    ```
    
III) NEW behavior, in the future when someone will bump `@angular/build` to version `<19.2.11`, where the patch is still needed:
Simulate a situation in few months from now, when someone bump `@angular/build` installed in our repo, but there is a risk that the patch is no longer applicable and they should verify the patching script still works properly:
1. Force-install newer version of the pacakge: `npm install @angular/build@19.2.10 --force`
2. run `npm install` and observe the following error in the terminal informing about mismatch in the expected @angular/build:
    ```
    > storefrontapp@0.0.0 postinstall
    > ts-node scripts/npm-postinstall/patch-@angular-build-19.0.4.ts
    
    ❌ Spartacus npm postinstall: Error patching file 'node_modules/@angular/build/src/tools/esbuild/bundler-context.js'
    Error: Expected '@angular/build' version 19.0.4, but the installed version is 19.2.10
             🚨 Please update the value of `const EXPECTED_VERSION = '19.0.4'` in the npm "postinstall" script that attempts to patch a bug in the @angular/build package. Then re-run `npm install` and verify the patch applied successfully.
    ```

IV) NEW behavior, in the future when someone will bump `@angular/build` to version `>=19.2.11`, where the patch is not needed anymore:
Simulate a situation in few months from now, when someone bump `@angular/build` installed in our repo to a version that doesn't contain the bug anymore. And that person should remove the script from "npm postinstall":
1. Force-install newer version of the pacakge: `npm install @angular/build@19.2.14 --force`
2. run `npm install` and observe the following error in the terminal informing about mismatch in the expected @angular/build version to be patched:
    ```
    > storefrontapp@0.0.0 postinstall
    > ts-node scripts/npm-postinstall/patch-@angular-build-19.0.4.ts
    
    ❌ Spartacus npm postinstall: Error patching file 'node_modules/@angular/build/src/tools/esbuild/bundler-context.js'
    Error: Expected '@angular/build' version 19.0.4, but the installed version is 19.2.14
             🚨 Please remove the npm "postinstall" script which attempts to patch a bug in the @angular/build package. The installed version 19.2.14 already contains the bugfix, so the patch is not needed anymore.
    ```


rmch91 (approved):
LGTM
QA passed

PR #20396, opened 2025-05-29
fix: Dark theme not applied to the Punchout InspectCart page (CXSPA-10190)

Pucek9:
Technically it is, but the page was not stretched
Closes [CXSPA-10190](https://jira.tools.sap/browse/CXSPA-10190)

### DIFF
--- a/integration-libs/punchout/styles/_punchout-template.scss
+++ b/integration-libs/punchout/styles/_punchout-template.scss
@@ -41,6 +41,7 @@
       min-width: 0;
     }
   }
+  min-height: calc(100vh - 1px);
 }
 
 .cxPunchoutSessionActive {
--- a/integration-libs/punchout/_index.scss
+++ b/integration-libs/punchout/_index.scss
@@ -10,8 +10,8 @@ $punchout-components-allowlist:
 
 $skipComponentStyles: () !default;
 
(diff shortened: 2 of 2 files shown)


integration-libs/punchout/styles/_punchout-template.scss
```diff
@@ -41,6 +41,7 @@
       min-width: 0;
     }
   }
+  min-height: calc(100vh - 1px);
```
kimhw0630, line 44: Doesn't the solution always create a vertical scroll, especially when it's inside an iframe?

FollowTheFlo, line 44: why this value?
is it standard in spartacus?

Pucek9: 100vh in ideal conditions is 100% and does not cause a visible vertical scroll yet. I didnt test it in iframe

I'm not saying it's a perfect solution, I see it works fine on inspect page

Pucek9: Here (in eg, dark theme) we have some weird 1px bottom border of `header` element. So that's why applied `- 1px`

FollowTheFlo (approved):
accepted.
not seeing better solution.

PR #20398, opened 2025-05-29
fix: add pdp url in allowedUrls

FollowTheFlo:


integration-libs/punchout/root/guards/punchout-navigation.guard.ts
```diff
@@ -75,7 +75,7 @@ export class PunchoutNavigationGuard {
     return (
       urls
         .filter((url) => url !== '/')
-        .some((url) => relativeUrl.includes(url)) || isHomePageAllowed
+        .some((url) => `/${relativeUrl}`.includes(url)) || isHomePageAllowed
```
FollowTheFlo, line 78: taking advantage to detect exact segment eg: '/product/' rather than 'product/' which would include larger scope like '/*product/' (eg: /my-product/)

PR #20404, opened 2025-05-30
fix: Punchout permission warning when navigate back on pdp landing page (CXSPA-10254)

Pucek9:
Closes [CXSPA-10254](https://jira.tools.sap/browse/CXSPA-10254)

### DIFF
--- a/projects/storefrontlib/cms-structure/routing/default-routing-config.ts
+++ b/projects/storefrontlib/cms-structure/routing/default-routing-config.ts
@@ -4,75 +4,81 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { RoutesConfig, RoutingConfig } from '@spartacus/core';
+import { FeatureToggles, RoutingConfig } from '@spartacus/core';
+import { inject } from '@angular/core';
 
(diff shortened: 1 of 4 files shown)


projects/storefrontlib/cms-structure/routing/default-routing-config.ts
```diff
@@ -4,75 +4,81 @@
...
  */
 
-import { RoutesConfig, RoutingConfig } from '@spartacus/core';
+import { FeatureToggles, RoutingConfig } from '@spartacus/core';
+import { inject } from '@angular/core';
 
-export const defaultStorefrontRoutesConfig: RoutesConfig = {
-  home: { paths: [''] },
-  notFound: { paths: ['not-found'] },
+export const defaultRoutesConfigFactory: () => RoutingConfig = () => {
```
FollowTheFlo, line 10: if we have agreed that moving form Config to ConfigFactory is not a breaking change, i'm happy with it.

FollowTheFlo (approved):
Tested successfully, see details in jira.
As confirmed there is no breaking change, good to me.

PR #20410, opened 2025-06-03
chore: update bash scripts to run b2b e2e and all unit tests in parallel

rmch91:
Closes: https://jira.tools.sap/browse/CXSPA-10247
This change is needed for azure pipeline.
Should not provide any breaking changes for other envs, like github actions or any other possible automations from other teams


Matejk00 (approved):
LGTM

PR #20418, opened 2025-06-05
chore: Update config:update script to run npm install

rmch91:
Changed config:update script, to run `npm install` after `package.json` changes in order to generate updated `package-lock.json`

> :warning: Developers changing any library's package.json need to do:
> 1. run npm install
> 2. commit any generated changes of the file package-lock.json to your feature branch (!)

When doing releases and bumping versions inside package.json's, the npm install command needs to be run and the package-lock.json should be commited as well.
Without changes in package-lock.json being committed right after changing the package.jsons , every developer who pulls develop now and runs npm install, generates in their machine "unnecessary" changes


tools/config/index.ts
```diff
@@ -294,6 +292,8 @@ if (options.generateDeps) {
  * Format all files.
  */
 if (options.fix) {
+  console.log('\nGenerating updated package-lock file...\n');
+  execSync('npm install', { stdio: 'inherit' });
```
Platonn, line 296: Here we're applying change sin the `config:update`.

IMHO its counterpart, the `config:check` ideally it should fail when the existing-git-comitted `pacakge-lock.json` contents are different than those generated by new run of `npm install`. It would help spot that the author of the changes in the PR forgot to commit their changes in `package-lock.json`. And then such a PR would fail in CI.

In other words, I'd suggest that the `config:check` logic could do the following:
```ts
const gitPackageLock= execSync(`git show HEAD:package-lock.json`, { encoding: 'utf8' });
execSync('npm install', { stdio: 'inherit' });
const newPackageLock = readFileSync('package-lock.json', 'utf8');

if (gitPackageLock !== newPackageLock) {
  // throw error...
}
```

Platonn (approved):

PR #20420, opened 2025-06-06
fix: replace Subject with BehaviorSubject for searchCompleted in…

lasoh:
… SearchBoxComponentService

- Updated searchCompleted to use BehaviorSubject for better state management.
- Added missing import for BehaviorSubject from rxjs.


projects/storefrontlib/cms-components/navigation/search-box/search-box-component.service.ts
```diff
@@ -31,7 +37,7 @@ const HAS_SEARCH_RESULT_CLASS = 'has-searchbox-results';
 export class SearchBoxComponentService {
   chosenWord = new ReplaySubject<string>();
   sharedEvent = new ReplaySubject<KeyboardEvent>();
-  searchCompleted = new Subject<boolean>();
+  searchCompleted = new BehaviorSubject<boolean>(false);
```
rmch91, line 40: Normally we shouldn't change the type of `public` properties, because it can be a breaking change for customers.
But in this case, since `BehaviorSubject` extends `Subject` I would say it is safe. 
At least I don't see any case of usage of old type, where new type will break it.

rmch91 (approved):

PR #20423, opened 2025-06-09
fix: (CXSPA-10264) - Consistent product card sizes

Pio-Bar:
Closes: [CXSPA-10264](https://jira.tools.sap/browse/CXSPA-10264)

Suggestion on how product cards may look like with their size restricted to a hard value. The intention is to keep the sizes consistent on b2b version regardless of the product images provided. 

The numbers follow b2c card dimensions since these appear to be constant. We should rethink this approach.


Zeyber:
There seems to be some overflow text here onto the next product. Do you know if it is related to this?
image

Pio-Bar:
> There seems to be some overflow text here onto the next product. Do you know if it is related to this?

It definitely is, since the hight is now constant It can't expand for the longer description. We can use the line clamp trick to limit its length. I've seen the solution in Spartacus already, but I'm not sure how to feel about it.

Zeyber (approved):

PR #20426, opened 2025-06-09
chore: update expired token detection for auth server rebuild

ercultimate:
Closes: [CXSPA-10394](https://jira.tools.sap/browse/CXSPA-10394)


rmch91 (approved):
LGTM

PR #20428, opened 2025-06-10
perf: support FetchPriority HTML attribute in MediaComponent

kimhw0630:
Closes: https://jira.tools.sap/browse/CXSPA-10297


projects/core/src/features-config/feature-toggles/config/feature-toggles.ts
```diff
@@ -804,6 +804,16 @@ export interface FeatureTogglesInterface {
...
+
+  /**
+   * Enables the `fetchpriority` attribute on images in the `MediaComponent`.
+   * When set, this attribute hints to the browser to prioritize image loading,
+   * improving perceived performance—especially for above-the-fold content.
+   *
+   * Controlled via the new `fetchpriority` input in `media.component.ts`.
+   * See `media.component.ts` for implementation details.
+   */
+  enableMediaFetchPriority?: boolean;
```
Platonn: Thank you for taking care of our public API (not introducing breaking changes) by proactively introducing feature toggles (aka "breaking changes toggles).

IMHO in this specific PR we're not introducing any breaking change. We're just adding a new, optional `@Input fetchPriority` which allows parent components to pass the hint about the priority to the contained  tag's attribute.

But as of now, no parent component uses this new Input. So the behavior of this component occurences is unchanged (no breaking change).

Therefore I suggest to remove this feature toggle as unnecessary (ad to avoid complexity of having 2 branches of HTML templates inside the HTML file, which is already complex).

PS. Only later in later PRs, where we'll be starting to use this new input, those parent components might need to introduce their own feature toggles, to avoid behavioral breaking change.

projects/storefrontlib/shared/components/media/media.model.ts
```diff
@@ -105,3 +105,24 @@ export enum ImageLoadingStrategy {
...
+ */
+export enum ImageFetchPriority {
+  /**
+   * Fetch the image at a high priority relative to other images with the same internal prioritization.
+   */
+  HIGH = 'high',
+  /**
+   * Fetch the image at a low priority relative to other images with the same internal prioritization.
+   */
+  LOW = 'low',
```
Platonn, line 123: I'm sorry. Perhaps it's only me. I'm having problem to understand what did you mean by `with the same internal prioritization.` What's the internal prioritization in this case?

kimhw0630: Context is from  https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/fetchPriority#value

internal prioritization refers to the logic within the application that determines which images should be loaded first, second, etc., based on the rules (explain from Copilot ) 😅

projects/storefrontlib/shared/components/media/media.component.spec.ts
```diff
@@ -371,4 +371,44 @@ describe('MediaComponent', () => {
...
+    const { component } = createComponent();
+    component.fetchPriority = undefined;
+    spyOnProperty(component, 'loadingStrategy', 'get').and.returnValue(
+      ImageLoadingStrategy.LAZY
+    );
+    component.loading = null;
+    expect(component['effectiveLoadingStrategy']).toBe(
+      ImageLoadingStrategy.LAZY
+    );
+  });
```
Platonn: 1. For clarity, can wrap those 3 tests with a `describe('effectiveLoadingStrategy'` block?
2. Did you consider actually testing the _UI behavior_ of the component (instead of - or not only - the implementation of the protected TS methods)? e.g. checking whether the `` element has the proper HTML attributes assigned ("loading" & "fetchPriority" ) in reaction to set `@Input`s (after calling `fixture.detectChanges()`) ?

Platonn:
# QA Part 1 - check the presence of HTML attributes in Elements DevTools and check browser's loading priority in Network DevTools
- opened storefront app `npm run start`

BEFORE:
- visited the homepage in Chrome with an emulation of a mobile device "Moto G Power"
- observed in the Network tab of Chrome DevTools that the request for the main banner image had initially the loading priority Low, and only after some time the browser decided to change the priority to High
    image
- dumped to a file the full HTML structure, using the Elements tab of Chrome DevTools

AFTER:
- for the QA purposes, added an usage of the new feature (fetchPriority): I edited the BannerComponent implementation to pass `@input fetchPriority` with value `'high'` only for cms component with `uuid === 'ElectronicsHompageSplashBannerComponent'`), as in the screenshots below:
    image
    and
    image
- visited the homepage again
- observed in the Network tab of Chrome DevTools that the request for the main banner image had initially the loading priority High, from the beginning (not after a delay)
    image
- dumped to a file the new full HTML structure, using the Elements tab of Chrome DevTools
- checked the `diff` of the dumped HTML structures (BEFORE vs AFTER) and verified that the new `` tag of this banner component was using html attributes `loading=eager` and `fetchpriority=high`. See the screenshot below:
    image


----

# QA part 2 - check when the LCP image is downloaded

To simulate more real conditions, I've:
- setup CDN with optimized (AVIF) images and use it as a base media url
- configured all Spartacus images to be lazy-loaded
    ```ts
    provideConfig({
      imageLoadingStrategy: ImageLoadingStrategy.LAZY,
    });
    ```
- configured the main Banner image to have `@Input fetchPriority` with value `high`
- visited page first (to warm up CDN's cache)
- visited the page again, with recording the Performance recording the Chrome DevTools

BEFORE:
The browser started to download the main Banner image quite late together with all other (less important) images, only after JS files were downloaded. See when starts the horizontal bar highlighter with the red arrow in the screenshot below:
image


AFTER:
The browser started to download the main Banner image immediately after downloading the main HTML document See when starts the horizontal bar highlighter with the red arrow in the screenshot below:
image

Platonn (dismissed):

Platonn (approved):

PR #20432, opened 2025-06-11
test: Include B2B a11y e2e tests in accessibility suite

sdrozdsap:


projects/storefrontapp-e2e-cypress/package.json
```diff
@@ -16,12 +16,14 @@
...
     "cy:run:ci:core": "cypress run --config-file cypress.config.ci.ts --record --key $CYPRESS_KEY --tag \"b2c,core\" --parallel --ci-build-id $BUILD_NUMBER --group B2C --spec \"cypress/e2e/!(vendor|b2b|ssr|cx_ccv2|cx_mcs)/**/*.core-e2e.cy.ts\"",
-    "cy:run:ci:a11y": "cypress run --config-file cypress.config.ci.ts --record --key $CYPRESS_KEY --tag \"a11y,all\" --parallel --ci-build-id $BUILD_NUMBER --group A11Y --spec \"cypress/e2e/**/*.a11y-e2e.cy.ts\" --reporter junit --reporter-options mochaFile=results/e2e-test-a11y-[hash].xml",
+    "cy:run:ci:a11y": "cypress run --config-file cypress.config.ci.ts --record --key $CYPRESS_KEY --tag \"a11y,b2c\" --parallel --ci-build-id $BUILD_NUMBER --group A11Y --spec \"cypress/e2e/a11y/b2c/**/*.a11y-e2e.cy.ts\" --reporter junit --reporter-options mochaFile=results/e2e-test-a11y-[hash].xml",
```
Zeyber, line 25: For the reasons mentioned on [CXSPA-10308](https://jira.tools.sap/browse/CXSPA-10308), lets remove record on these ones. I think we can leave it for the release runs.

Zeyber (changes requested):

projects/storefrontapp-e2e-cypress/package.json
sdrozdsap: I'd suggest to wait with those changes considering that this might conflict with what Stan is working on("might" is wrong word here, rather "will":))
I've synced up with him and we will definetely take care of it in his PR

Zeyber (changes requested):
I really like what you have done with adding helper functions to this script. Makes it easier to digest. 

Do you think we could move out the a11y specific functions into another shell file (is it possible)? I'm thinking the file itself is just getting a little big and we the a11y suite functions are only needed for our teams requirements. So other teams will have less to digest when digging into CI ;)

sdrozdsap:
> I really like what you have done with adding helper functions to this script. Makes it easier to digest.
> 
> Do you think we could move out the a11y specific functions into another shell file (is it possible)? I'm thinking the file itself is just getting a little big and we the a11y suite functions are only needed for our teams requirements. So other teams will have less to digest when digging into CI ;)

Yes! I thought about that too and I think it should be relatively easy to refactor it that way. Let me get to it!

Zeyber (approved):

PR #20433, opened 2025-06-11
feat: enable Angular non-destructive hydration in new apps and storefrontapp

npapp-dev002:
Introduces non-desctructive hydration. Fixes/skips all hydration errors.

### Context
Non-destructive hydration is a feature introduced in Angular 17 (and improved in Angular 18/19) that allows Angular applications to hydrate server-rendered HTML without destroying and recreating the DOM elements. Instead of discarding the server-rendered content and rebuilding it from scratch, Angular attaches event listeners and internal state to the existing DOM, preserving the original HTML and user state.

Positive outcomes for SSR apps:

- Faster startup: The browser does not need to re-render the UI, leading to quicker interactivity.
- Reduced layout shift: Since the DOM is not recreated, there is less visual flicker or content shift.
- Improved Core Web Vitals: Faster hydration and less layout shift improve metrics like LCP and CLS, which are important for SEO and user experience.

`withEventReplay` is an Angular hydration option that enables "event replay" during the hydration process. When using server-side rendering (SSR) with hydration, there can be a gap between when the HTML is rendered and when Angular becomes interactive on the client. During this gap, users might interact with the page (e.g., click buttons, type in inputs).

What event replay does:

Captures user events (like clicks, input, etc.) that occur before Angular is fully bootstrapped on the client.
Once Angular is ready, it replays those captured events so the app responds as if it was interactive the whole time.
Benefits:
- Prevents lost user interactions during hydration.
- Improves user experience by making the app feel more responsive.

Fixes [CXSPA-9841](https://jira.tools.sap/browse/CXSPA-9841)

### Remaining work
- [CXSPA-10422 Breakpoint service causes flickering with SSR](https://jira.tools.sap/browse/CXSPA-10422)
- [CXSPA-10423 Double-loading of Home Page images](https://jira.tools.sap/browse/CXSPA-10423)

Introducing Angular's built-in HTTP Cache Transfer mechanism was descoped to the following ticket: [CXSPA-10430](https://jira.tools.sap/browse/CXSPA-10430)


projects/storefrontlib/layout/a11y/keyboard-focus/lock/lock-focus.directive.ts
```diff
@@ -64,7 +64,10 @@ export class LockFocusDirective
   handleEnter(event: KeyboardEvent) {
     if (this.shouldLock && this.host === (event.target as HTMLElement)) {
       this.unlockFocus(event);
-      event.preventDefault();
+      //prevent errors with event replay
+      if (event.eventPhase !== EventPhase.REPLAY) {
+        event.preventDefault();
+      }
```
Platonn: [QUESTION]
1. Just curious, how did you figure out that we need to check this specific condition `event.eventPhase !== EventPhase.REPLAY` (`import { EventPhase } from '@angular/core/primitives/event-dispatch';`)?
2. How did you test that the LockFocusDirective still works as expected?

projects/storefrontlib/shared/components/carousel/focusable-carousel-item/focusable-carousel-item.directive.ts
```diff
@@ -51,7 +51,7 @@ export class FocusableCarouselItemDirective {
     const focusableTags = ['INPUT', 'SELECT', 'TEXTAREA', 'BUTTON', 'A'];
     const element = this.el.nativeElement;
     const hasTabIndex =
-      element.hasAttribute('tabindex') &&
+      element?.hasAttribute('tabindex') &&
```
Platonn: [QUESTION]
Is this change really needed? Or it's just a leftover from the initial debugging?

projects/storefrontlib/layout/a11y/keyboard-focus/lock/lock-focus.directive.ts
npapp-dev002: 1. Screenshot 2025-06-10 at 9 09 09
2. I assumed that REPLAY events only occur after hydration is done, and Angular just replay events that happened before. However, it is hard to reproduce and therefore I think it can be removed.

projects/storefrontlib/shared/components/carousel/focusable-carousel-item/focusable-carousel-item.directive.ts
npapp-dev002: Leftover. Removed.

projects/storefrontapp/src/app/app.module.ts
```diff
@@ -62,6 +67,7 @@ if (!environment.production) {
   ],
   providers: [
     provideHttpClient(withFetch(), withInterceptorsFromDi()),
+    provideClientHydration(withEventReplay(), withNoHttpTransferCache()),
```
Platonn, line 70: If we decide to put `withNoHttpTransferCache()` in our storefrontapp, we should also:
1. [before mergin this PR] add it to customer's fresh apps (via the logic of our installation schematics)
2. [after merging this PR] document how to enable non-destructive hydration, mentioning that `withNoHttpTransferCache` is necessary and why
   - I guess it would be good to mention it somewhere in our official docs of SSR https://github.com/SAP/spartacus-docs)
   - and maybe also in a possible announcement published by Matthew in SAP notes at the day of the release (to be discussed with Matt)

projects/storefrontlib/cms-components/misc/icon/icon.component.html
```diff
@@ -1 +1,7 @@
-<i [outerHTML]="icon"></i><ng-content></ng-content>
+<ng-container *ngIf="icon; else withoutInternalHtml">
+  <i [outerHTML]="icon"></i>
+</ng-container>
+<ng-template #withoutInternalHtml>
+  <i></i>
+</ng-template>
```
Platonn: I did the QA and compared the production HTML output of homepage SSR.
- before this PR: the IconComponent had no `<i>` element inside, when the property `icon` was undefined
- in this PR: the empty `<i>` element appears inside the IconComponent, when `the property `icon` is undefined

See the diff of the production homepage HTML generated by SSR, focusing on the appeared `<i></i>` in the markup on the right hand side:
image


I'm sorry it's my bad, I remember I might have suggested the `else` template for NgIf, when we discussed it. 
Only now I can see that to preserve the previous behavior, we should remove the "else" template of the `ngIf` directive (in other words, remove the empty `<i></i>`)

projects/storefrontlib/cms-components/product/product-list/product-list-item/product-list-item.component.spec.ts
```diff
@@ -145,6 +145,9 @@ describe('ProductListItemComponent in product-list', () => {
   });
 
   it('should display product name', () => {
+    console.log(
+      fixture.debugElement.nativeElement.querySelector('.cx-product-name')
+    );
```
Platonn: I guess it's just a leftover after debugging and we can remove this console.log

projects/storefrontlib/cms-components/product/product-list/product-list-item/product-list-item.component.html
```diff
@@ -44,7 +44,7 @@ <h2 class="cx-product-name">
 
     <div class="row">
       <div class="col-12 col-md-7 col-xl-7">
-        <p class="cx-product-summary" [innerHtml]="product.summary">
+        <p class="cx-product-summary" [innerHTML]="product.summary">
```
Platonn: just curious: do you think that replacing `innerHtml` to `innerHTML` (uppercase HTML) fixed the hydration error? or flickering? (I don't remember the issue we had here)

projects/storefrontlib/layout/a11y/keyboard-focus/lock/lock-focus.directive.ts
```diff
@@ -64,6 +63,7 @@ export class LockFocusDirective
   handleEnter(event: KeyboardEvent) {
     if (this.shouldLock && this.host === (event.target as HTMLElement)) {
       this.unlockFocus(event);
+      //prevent errors with event replay
```
Platonn: Do we still need this comment or it's a leftover?

projects/storefrontlib/cms-components/product/product-list/product-list-item/product-list-item.component.html
```diff
@@ -44,7 +44,7 @@ <h2 class="cx-product-name">
...
           {{ product.summary }}
```
Platonn: Using our branch, when I've run `dev:ssr` and opened a PLP http://localhost:4001/electronics-spa/en/USD/Brands/all/c/brands, I saw the following hydration error in my console:
image

And I've checked that to fix it it suffices to remove the obsolete `{{ product.summary }}` contents of the `<p>` element. It's redundant anyway, becasue we're controlling the contents with the `[innerHTML]="product.summary"` HTML directive of the on this `<p>` element. After fixing it, I can see no hydration error on PLP:

image

So let's remove this line:
```suggestion
```

projects/storefrontlib/cms-components/product/product-list/product-list-item/product-list-item.component.html
npapp-dev002: Interesting. It seems that we bump into an issue that has nothing to do with hydration. Using the wrong html attribute causes issues here.
image
I used the selector to locate the component, but the error was related to an invalid HTML attribute. Since I was working on hydration at the time, I initially assumed it was connected to that.
I would like to keep this solution here, if you don't mind.

projects/storefrontlib/cms-components/product/product-list/product-list-item/product-list-item.component.spec.ts
npapp-dev002: Leftover.

projects/storefrontlib/layout/a11y/keyboard-focus/lock/lock-focus.directive.ts
npapp-dev002: Leftover.

projects/storefrontlib/cms-components/product/product-list/product-list-item/product-list-item.component.html
npapp-dev002: The error that I couldn't reproduce before is still occuring with innerHTML as well. It is not causing flickering, just logs the error on the browser's console.

Platonn (commented):
✅ QA passed

- I've built libs from this branch and published to verdaccio

SSR app:
- I've created a fresh ng19 app with SSR, with installing Spartacus from verdaccio
- ✅ I confirm the `provideClientHydration(withEventReplay(), withNoHttpTransferCache())` was added to the app - meaning that the non-destructive hydration was enabled and the native httpTransferCache was disabled (as we agreed for the time being, in favor of existing Spartacus TransferState of ngrx)
- ✅ I've opened the homepage and I confirm there is no flickering(no re-appearing) of the main banners when the CSR kicks in after SSR (note: the site has some layout shift, but it's caused by a faulty header component, which is an old known issue and will be handled in other ticket). Below you'll see 2 recordings (with 4x CPU throttling and Slow 4G network throttling):

BEFORE - using Spartacus libs v2211.41 from RBSC:
Please note the main banner re-appears on the page after the JS kicks in (likely due to the destructive hydration):

https://github.com/user-attachments/assets/e93a6418-3d92-49a7-898a-bc2c1a97fadd

AFTER - using Spartacus libs from this branch via verdaccio:
Please note the main banner doesn't re-appear on the page after JS kicks in (likely due to the non-destructive hydration). You'll see a slight layout shift but it's caused by the faulty header component which will be handled in other ticket:

https://github.com/user-attachments/assets/c80a53f7-4661-471d-8186-07723e20a5c2

CSR-only app:
- I've created a fresh ng19 app with ONLY CSR with installing Spartacus from verdaccio
- ✅ I confirm there is no `provideClientHydration(withEventReplay(), withNoHttpTransferCache())` in the `app.module.ts` - as expected

Platonn (approved):

PR #20434, opened 2025-06-11
test: (CXSPA-10031) -  Potential a11y concerns in Cypress log

Pio-Bar:
Closes: [CXSPA-10031](https://jira.tools.sap/browse/CXSPA-10031)

The Continuum is capable of reporting concerns It can not 100% confirm. These include contrast checks and layout problems. These should not be a cause for test failures however, we could still take advantage of them by outputting them in the cypress log. This way a manual assessment can be performed while working on the test.
I imagine in our case most of these to be false positives

I also updated the docs to reflect the slight change of approach we made some time ago.


projects/storefrontapp-e2e-cypress/cypress/support/continuum.commands.ts
```diff
@@ -137,6 +139,26 @@ const isContinuumAvailable = () => {
...
 
+// Some concerns reported by the continuum cannot be verified automatically and require a manual review by the dev.
+// We assume that unconfirmed violations are not critical, so we log them instead of failing the test.
+const getConfirmedConcerns = (
+  accessibilityConcerns,
+  logPossibleConcerns = true
+) => {
+  return accessibilityConcerns.filter((concern) => {
+    if (concern._needsReview && logPossibleConcerns) {
+      cy.get(concern.path, { log: false }).then((node) => {
```
Zeyber: I think we should use the term "potential" throughout the script to keep consistent with the configuration term. I also think we could use a method called something like `displayPotentialConcerns` since its not too clear to me that happens here.

Zeyber (changes requested):
The warnings look very tidy in the log... nice! This is good because they are visible but not obstructive as to create confusion. Its also good that these are not failing the tests.

I do think we need some exception mechanism to suppress a warning that has been validated as not an issue. Perhaps something identifying a node. I don't think it needs to be precise. But I think it would be good for developers to acknowledge they have seen the warning and checked it. 

Regarding site themes, would that be a separate PR?

projects/storefrontapp-e2e-cypress/cypress/support/continuum.commands.ts
Pio-Bar: - I've refactored it for clarity. 
- We now have a way to disable reported best practice violation warnings/errors.
- There's not much we can do about scanning contrast themes. AC can't run a test with increased contrast ratio requirement of 7:1. Would there be other concerns apart from contrast checks?

Zeyber (approved):

PR #20447, opened 2025-06-13
fix: CXSPA-10431 make sonar happy

i53577:
image
https://sonar.tools.sap/dashboard?id=composable-storefront&pullRequest=20447


scarai-sap (approved):
Looks safe

PR #20458, opened 2025-06-17
feat: ASM Agent login adapted for public clients

Melody-zhou-512:
Ticket:https://jira.tools.sap/browse/CXSPA-10428


niehuayang (dismissed):
approved, But please keep using the same JDK21 toggle as the core team, perhaps it is necessary to align with the core team

feature-libs/asm/root/services/csagent-auth.service.ts
```diff
@@ -83,6 +83,43 @@ export class CsAgentAuthService {
...
+      .subscribe((token) => (userToken = token))
+      .unsubscribe();
+
+    this.authStorageService.switchTokenTargetToCSAgent();
+    try {
+      await this.oAuthLibWrapperService.initLoginFlow();
+      this.store.dispatch(new AuthActions.Logout());
+
+      if (customerId !== undefined && userToken !== undefined) {
+        // OCC specific user id handling. Customize when implementing different backend
```
ercultimate, line 109: It might be best to encapsulate this code in a method so customers can override the OCC-specific logic easier.

projects/core/src/features-config/feature-toggles/config/feature-toggles.ts
```diff
@@ -29,6 +29,11 @@ export interface FeatureTogglesInterface {
    */
   formErrorsDescriptiveMessages?: boolean;
 
+  /**
+   * After the feature toggle is enabled, Customer Support Agent logs in ASM with the Authorization Code flow.
+   */
+  showOauth2AsmloginPage?: boolean;
```
ercultimate: Just a warning: feature flags will all be enabled in newly-created Spartacus apps.  This can interfere with E2E testing until a JDK21 backend is used regularly.

ercultimate: I would suggest renaming the flag that is more descriptive of what enabling the flag does.  Something like `agentLoginWithAuthorizationCodeFlow`?

projects/core/src/features-config/feature-toggles/config/feature-toggles.ts
```diff
@@ -29,6 +29,11 @@ export interface FeatureTogglesInterface {
    */
   formErrorsDescriptiveMessages?: boolean;
 
+  /**
+   * After the feature toggle is enabled, Customer Support Agent logs in ASM with the Authorization Code flow.
+   */
```
ercultimate: Please add this note to the flag description.
```
   * NOTE: This flag must be enabled when used with a CCv2 Authorization Server running the
   * September 2025 update or higher. The CCv2 Authorization Server only supports Authorization Code
   * flow for public clients from that version and onwards.
```

projects/core/src/features-config/feature-toggles/config/feature-toggles.ts
ercultimate: Expanding on this warning, the current code is fine if you are planning to wait until August to merge this PR for the september release.  

If you wish to merge is sooner, you will need to do 2 things:
1. Convert the feature flag to a "feature provider" (we have no formal term for this construct), like we did with `provideAuthorizationCodeFlowByDefault()` in develop.
2. Create a PR to convert the "feature provider" back into a feature toggle, with the plan to merge this PR in August so that it is included with the September release.

feature-libs/asm/components/asm-main-ui/asm-main-ui.component.ts
```diff
@@ -86,6 +88,11 @@ export class AsmMainUiComponent implements OnInit, OnDestroy {
   @ViewChild('customerListLink') element: ElementRef;
   @ViewChild('addNewCustomerLink') addNewCustomerLink: ElementRef;
 
+  protected featureConfig = inject(FeatureConfigService);
```
ercultimate: Please use `FeatureToggles`.  

The service `FeatureConfigService` is a legacy service that we should avoid when possible.

niehuayang (dismissed):

feature-libs/asm/components/asm-components.module.ts
```diff
@@ -104,6 +105,7 @@ import { DotSpinnerComponent } from './dot-spinner/dot-spinner.component';
     provideDefaultConfig(defaultCustomerListLayoutConfig),
     provideDefaultConfig(defaultAsmPaginationConfig),
     provideDefaultConfig(defaultAsmCreateCustomerFormLayoutConfig),
+    provideAuthorizationCodeFlowByDefault(),
```
niehuayang: should delete it, code flow is disabled by default

Melody-zhou-512: have been modified.

niehuayang (approved):

rmch91 (approved):

PR #20459, opened 2025-06-17
feat: add media preconnect feature with toggle

npapp-dev002:
closes [CXSPA-9613](https://jira.tools.sap/browse/CXSPA-9613)

The preconnect keyword for the [rel](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/link#rel) attribute of the [<link>](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/link) element is a hint to browsers that the user is likely to need resources from the target resource's origin, and therefore the browser can likely improve the user experience by preemptively initiating a connection to that origin. Preconnecting speeds up future loads from a given origin by preemptively performing part or all of the handshake (DNS+TCP for HTTP, and DNS+TCP+TLS for HTTPS origins). (source:[preconnect documentation]( https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/rel/preconnect))

Server-side rendered and CDN cached sites might benefit the most from the pre-connecting to media sites.

### DIFF
--- a/projects/core/src/features-config/feature-toggles/config/feature-toggles.ts
+++ b/projects/core/src/features-config/feature-toggles/config/feature-toggles.ts
@@ -736,7 +736,7 @@ export interface FeatureTogglesInterface {
    * and that you'd like the process to start as soon as possible. Resources will load more quickly because the setup process
    * has already been completed by the time the browser requests them.
    */
-  createMediaPreconnectLinkInSsr?: boolean;
+  createMediaPreconnectLink?: boolean;
 }
 
 export const defaultFeatureToggles: Required<FeatureTogglesInterface> = {
@@ -843,5 +843,5 @@ export const defaultFeatureToggles: Required<FeatureTogglesInterface> = {
   reserveHorizontalSpaceStarRating: false,
   topProgressBarUseTransformAnimation: false,
   disableCxPageSlotMarginAnimation: false,
-  createMediaPreconnectLinkInSsr: false,
+  createMediaPreconnectLink: false,
 };
--- a/projects/core/src/features-config/feature-toggles/config/feature-toggles.ts
+++ b/projects/core/src/features-config/feature-toggles/config/feature-toggles.ts
@@ -713,6 +713,19 @@ export interface FeatureTogglesInterface {
    * layout-independent operation that improves visual performance and avoids layout shifts.
    */
   topProgressBarUseTransformAnimation?: boolean;
+
+  /**
+   * Feature flag to enable using <link rel=preconnect> in the index.html.
+   *
+   * ## Why this flag exists
+   * 1. **Preconnecting is only effective for domains other than the origin domain, so you shouldn't use it for your site.
+   *
+   * ## When enabled:
+   * Adding rel=preconnect to a <link> informs the browser that your page intends to establish a connection to another domain,
+   * and that you'd like the process to start as soon as possible. Resources will load more quickly because the setup process
+   * has already been completed by the time the browser requests them.
+   */
+  createMediaPreconnectLinkInSsr?: boolean;
 }
 
 export const defaultFeatureToggles: Required<FeatureTogglesInterface> = {
@@ -818,4 +831,5 @@ export const defaultFeatureToggles: Required<FeatureTogglesInterface> = {
   defaultProductPageRouteAllowsNoProductName: false,
   reserveHorizontalSpaceStarRating: false,
   topProgressBarUseTransformAnimation: false,
+  createMediaPreconnectLinkInSsr: false,
 };
--- a/projects/core/src/features-config/feature-toggles/config/feature-toggles.ts
+++ b/projects/core/src/features-config/feature-toggles/config/feature-toggles.ts
@@ -843,5 +843,5 @@ export const defaultFeatureToggles: Required<FeatureTogglesInterface> = {
   reserveHorizontalSpaceStarRating: false,
   topProgressBarUseTransformAnimation: false,
   disableCxPageSlotMarginAnimation: false,
-  createMediaPreconnectLinkInSsr: false,
+  createMediaPreconnectLinkInSsr: false
 };
--- a/projects/core/src/features-config/feature-toggles/config/feature-toggles.ts
+++ b/projects/core/src/features-config/feature-toggles/config/feature-toggles.ts
@@ -729,7 +729,7 @@ export interface FeatureTogglesInterface {
    * Feature flag to enable using <link rel=preconnect> in the index.html.
    *
    * ## Why this flag exists
-   * 1. **Preconnecting is only effective for domains other than the origin domain, so you shouldn't use it for your site.
+* 1. **Preconnecting is not needed (and won't be performed) if the domain of the media base url is the same as the storefront's domain.
    *
    * ## When enabled:
    * Adding rel=preconnect to a <link> informs the browser that your page intends to establish a connection to another domain,
--- a/projects/core/src/features-config/feature-toggles/config/feature-toggles.ts
+++ b/projects/core/src/features-config/feature-toggles/config/feature-toggles.ts
@@ -728,13 +728,12 @@ export interface FeatureTogglesInterface {
   /**
    * Feature flag to enable using <link rel=preconnect> in the index.html.
    *
-   * ## Why this flag exists
-   * 1. **Preconnecting is not needed (and won't be performed) if the domain of the media base url is the same as the storefront's domain.
-   *
    * ## When enabled:
    * Adding rel=preconnect to a <link> informs the browser that your page intends to establish a connection to another domain,
    * and that you'd like the process to start as soon as possible. Resources will load more quickly because the setup process
    * has already been completed by the time the browser requests them.
+   *
+   * Note: Preconnecting is not needed (and won't be performed) if the domain of the media base url is the same as the storefront's domain.
    */
   createMediaPreconnectLink?: boolean;
 }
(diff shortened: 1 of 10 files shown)


projects/core/src/features-config/feature-toggles/config/feature-toggles.ts
```diff
@@ -724,6 +724,19 @@ export interface FeatureTogglesInterface {
    * Enabling this flag removes the margin animation to improve performance and user experience.
    */
   disableCxPageSlotMarginAnimation?: boolean;
+
+  /**
+   * Feature flag to enable using <link rel=preconnect> in the index.html.
+   *
+   * ## Why this flag exists
+   * 1. **Preconnecting is only effective for domains other than the origin domain, so you shouldn't use it for your site.
```
Platonn: Good observation in the JSDoc!

It inspired me to this question:
What about adding OOTB the following logic: let's check in `MediaPreconnectService` if the domain current storefront's domain rendered in the SSR (`windowRef.location.origin`) is the same as the _domain_ of the configured media base url? (note: mediaBaseUrl may contain some _/url/path_ besides the _domain_ too).

And if the domains the same, then the preconnect tag shouldn't be added in SSR (regardless of the toggle).

Then in the toggle's JSDoc we could mention:
```suggestion
* 1. **Preconnecting is not needed (and won't be performed) if the domain of the media base url is the same as the storefront's domain.
```

projects/storefrontlib/cms-structure/seo/page-meta-link.service.ts
```diff
@@ -4,16 +4,18 @@
...
 
 @Injectable({
   providedIn: 'root',
 })
 export class PageMetaLinkService {
   constructor(
     protected winRef: WindowRef,
-    protected rendererFactory: RendererFactory2
+    protected rendererFactory: RendererFactory2,
+    @Inject(DOCUMENT) protected document: Document
```
Platonn: [MEDIUM]
Adding new required parameter of the constructor is a Typescript breaking change.

Let's add new DI dependencies via `inject` function. 

For more, see https://wiki.one.int.sap/wiki/display/spar/Coding+Guidelines#CodingGuidelines-Useinject()functioninsteadofconstructorforDependencyInjection

projects/storefrontlib/cms-structure/seo/page-meta-link.service.ts
```diff
@@ -44,6 +46,13 @@ export class PageMetaLinkService {
     }
   }
 
+  addPreconnectLink(url: string): void {
+    const preconnect = this.renderer.createElement('link');
+    this.renderer.setAttribute(preconnect, 'rel', 'preconnect');
+    this.renderer.setAttribute(preconnect, 'href', url);
+    this.document.head.insertBefore(preconnect, this.document.head.firstChild); // we want the preconnect-link to be at the top of the <head> section
+  }
+
```
Platonn: 1. [MEDIUM]
I realized only now that some customers may have put (hardcoded) already some `<link rel=preconnect` in the <head> of their main `index.html` file in their repo

    So to not possibly duplicate the `<link>` elements for the same `href` for such customers, let's check first in our logic if a `<link rel=preconnect` with the same `href` (as the one we want to add) doesn't exist already.

2. [MINOR]
Let's add unit tests for the new method.

projects/storefrontlib/cms-structure/seo/seo.module.ts
```diff
@@ -29,6 +36,12 @@ export function initSeoService(injector: Injector): () => void {
       deps: [Injector],
       multi: true,
     },
+    {
+      provide: APP_INITIALIZER,
+      useFactory: mediaPreconnectInitializer,
+      deps: [MediaPreconnectService],
+      multi: true,
+    },
```
Platonn: [MEDIUM]
What do you think about putting this app initializer into the [MediaModule.forRoot()](https://github.com/SAP/spartacus/blob/1d5bb08ade8495cf2d779e11f40dd592d0d9a930/projects/storefrontlib/shared/components/media/media.module.ts#L18) ? 

(I mean in the new `providers:[]` array there:
image

projects/storefrontlib/cms-structure/seo/media-preconnect.service.ts
```diff
@@ -0,0 +1,32 @@
...
+  private windowRef = inject(WindowRef);
+
+  addPreconnectLink(): void {
+    if (!this.featureToggles.createMediaPreconnectLinkInSsr) {
+      return;
+    }
+
+    if (!this.windowRef.isBrowser()) {
+      return;
+    }
```
Platonn: [MEDIUM]
My bad, I wrote this condition in the ticket description. But I introduced a bug.
Currently the feature is disabled in SSR. 😅 
But we wanted it to be enabled in SSR:
```suggestion
    if (this.windowRef.isBrowser()) {
      return;
    }
```

Platonn: UPDATE:
I have second thoughts, that this feature might be helpful also in case of CSR only storefront (e.g. CSR fallback). It will help to preconnect to media, in parallel to making the first occ call.

So:
- let's remove the check for `isBrowser` at all. We want the feature to work both for SSR and CSR
- let's update the name of the feature toggle
- let's update JSdocs if needed (if they mentioned that the feature is meant only for SSR)

projects/storefrontlib/cms-structure/seo/media-preconnect.service.ts
```diff
@@ -0,0 +1,38 @@
...
+import { MediaService } from '../../shared/components/media/media.service';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class MediaPreconnectService {
+  protected pageMetaLinkService = inject(PageMetaLinkService);
+  protected mediaService = inject(MediaService);
+  private featureToggles = inject(FeatureToggles);
+  private windowRef = inject(WindowRef);
```
Platonn: Regular dependencies (besides FeatureToggles) should be `protected` so customers can use them in their extended classes:

```suggestion
  protected windowRef = inject(WindowRef);
```

projects/core/src/features-config/feature-toggles/config/feature-toggles.ts
```diff
@@ -724,6 +724,19 @@ export interface FeatureTogglesInterface {
...
+   * Feature flag to enable using <link rel=preconnect> in the index.html.
+   *
+   * ## Why this flag exists
+   * 1. **Preconnecting is not needed (and won't be performed) if the domain of the media base url is the same as the storefront's domain.
+   *
+   * ## When enabled:
+   * Adding rel=preconnect to a <link> informs the browser that your page intends to establish a connection to another domain,
+   * and that you'd like the process to start as soon as possible. Resources will load more quickly because the setup process
+   * has already been completed by the time the browser requests them.
+   */
```
Platonn: This section now looks a bit strange with the title `Why this flag exists`, doesn't it?:
```
   * ## Why this flag exists
   * 1. **Preconnecting is not needed (and won't be performed) if the domain of the media base url is the same as the storefront's domain.
```
I suggest to move it to the bottom Note.

```suggestion
  /**
   * Feature flag to enable using <link rel=preconnect> in the index.html.
   *
   * ## When enabled:
   * Adding rel=preconnect to a <link> informs the browser that your page intends to establish a connection to another domain,
   * and that you'd like the process to start as soon as possible. Resources will load more quickly because the setup process
   * has already been completed by the time the browser requests them.
   *
   * Note: Preconnecting is not needed (and won't be performed) if the domain of the media base url is the same as the storefront's domain.
   */
```

projects/storefrontlib/cms-structure/seo/media-preconnect.service.ts
```diff
@@ -0,0 +1,38 @@
...
+
+    const url = this.mediaService.getBaseUrl();
+    let domain: string | undefined;
+    try {
+      domain = new URL(url).origin;
+    } catch {
+      domain = undefined;
+    }
+
+    if (domain && domain !== this.windowRef.location.origin) {
```
Platonn, line 36: Let's explain the _reason_ of the code also here in the source code (not only in feature toggle's description -> because the feature toggle will be eventually removed in 12months from now)

```suggestion
    // Preconnecting is not needed (and won't be performed) if the domain of the media base url
    // is the same as the storefront's domain
    if (domain && domain !== this.windowRef.location.origin) {
```

projects/storefrontlib/cms-structure/seo/media-preconnect.service.spec.ts
```diff
@@ -0,0 +1,109 @@
...
+      'https://media.example.com'
+    );
+  });
+
+  it('should handle invalid URL gracefully', () => {
+    windowRef.location.origin = 'https://storefront.example.com';
+    (globalThis as any)._originalURL = globalThis.URL;
+    globalThis.URL = function () {
+      throw new Error('Invalid URL');
+    } as any;
```
Platonn: [NITPICK]
Is mocking global `URL` definition really justified? The usage of the `URL` is just an implementation detail.
Can't we simply mock the Spartacus media base url config? (which is the expected dependency of this service)?

Platonn (commented):
QA passed:
- built libs and published to verdaccio. created fresh app with those libs
- opened the url in the browser and inspected the Elements tab in DevTools
  - ✅ SSR (with JS disabled) had the preconnect link
  - ✅ SSR with CSR kicking in had just 1 preconnect link (no duplicate created by CSR)
  - ✅ CSR only (CSR fallback thanks to configuring configured `timeout: 1`) with Network and CPU throttling enabled -> I observed that initially the HTML didn't contain any preconnect link. But when CSR kicked in, the preconnect link was added
   - ✅ SSR (when configured the same media URL as the storefront URL: `provideConfig({backend:{media:{baseUrl: 'http://localhost:4000'}}})`) didn't create the preconnect link.

projects/storefrontlib/cms-structure/seo/media-preconnect.service.spec.ts
npapp-dev002: You are right. Mocking global `URL` might be really a bit of overkill in this case.

Platonn (dismissed):

Platonn (approved):

PR #20465, opened 2025-06-17
fix: ensure carousel updates trigger change detection

npapp-dev002:
closes: [CXSPA-10440](https://jira.tools.sap/browse/CXSPA-10440)


projects/storefrontlib/shared/components/carousel/carousel.component.ts
```diff
@@ -94,7 +96,8 @@ export class CarouselComponent implements OnInit, OnChanges {
 
   constructor(
     protected el: ElementRef,
-    protected service: CarouselService
+    protected service: CarouselService,
+    protected cdr: ChangeDetectorRef
```
kimhw0630, line 100: This is a breaking change..

`protected cdr = inject(ChangeDetectorRef, { optional: true });`

PR #20467, opened 2025-06-18
fix: add copyright footer to Punchout Inspect page

FollowTheFlo:


integration-libs/punchout/styles/_punchout-template.scss
```diff
@@ -41,6 +41,10 @@
       min-width: 0;
     }
   }
+  .BodyContent {
+    min-height: calc(100vh - 200px);
+  }
+
   min-height: calc(100vh - 1px);
```
Pucek9: is this still necessary?

FollowTheFlo: thx, it is not, removing it. I confirm during tests, it has not effect

Pucek9: even with dark mode and zoom out? 😄 btw are we already affected by breaking change?

Pucek9 (approved):

PR #20481, opened 2025-06-24
feat: prioritize LCP image for selected CMS components

Platonn:
**Context:**
In the previous PR https://github.com/SAP/spartacus/pull/20428 we've introduced a new `@Input() fetchPriority` input in the `<cx-media>` component, but didn't use it anywhere.

**Feature description:**
In this PR we introduce a convenient configuration for specifying CMS component IDs that contain the LCP element (and therefore the `fetchPriority=high` should be used inside those components' implementation for their main `<cx-media>` image).

The storefront's owner needs to first determine via experimentation which components contain the LCP element, and then configure Spartacus (and/or adjust the CMS data) to mark specific CMS components as containing the LCP element.

CMS component IDs containing LCP element can be defined in 2 ways via the new Spartacus config:
- `config.lcpCmsComponents.idMarker: string[]` - (recommended) by finding this special substring marker in the CMS component ID (by default it's `__cxLCP__`, so it matches for example the ID `"ProductListComponent__cxLCP__"`). Then the CMS Business User can flexibly change the IDs dynamically in the CMS data to contain this marker.
- `config.lcpCmsComponents.ids: string[]` - static list of CMS component IDs.

**Implementation details:**
- new `CmsLcpService` takes CMS component's data of each CMS component and returns `LcpPresence` enum value
- the information about the `LcpPresence` is provided on the DOM level via Dependency Injection token `LCP_CONTEXT`, with the help of the existing `CmsInjectorService` (which is used by the `[cxComponentWrapper] directive), which provides also `CmsComponentData` at the DOM level.
- descendant components can inject the `LCP_CONTEXT` with the help of the `[cxLcpContext]` directive
  - it's now used in this PR in a few selected components with images: Banner, Product Details image, Product List Item, Product Grid Item, Product Carousel. Those components were chosen in this PR, because they are most likely to contain the LCP image on their relevant sub-pages. But in the future new components can be added when needed. Moreover customers may need to implement in their custom components the usage of `[cxLcpContext]` directive.
  - when the `LcpPresence` is `HAS_LCP`, then `fetchPiority=high` is assigned to the relevant child `<cx-media>` component

By the way, we're introducing new (optional) `@Input`s: `itemIndex` inside the `product-list-item.component` and `product-grid-item.component`. And based on that, if the product listing is marked as containing LCP, then only th item with the `itemIndex === 0` has its image downloaded with a high priority, because only the first listing item is considered as the LCP image. Analogically for the `product-carousel-item.component`.

fixes: https://jira.tools.sap/browse/CXSPA-10304

---

**QA steps**
0) Prerequisites:
- build libs and publish via verdaccio
- create fresh ng19 app with SSR with those Spartacus libs

1) test static config of the ID list:
- provide config in app.module:
```ts
    provideConfig({
      lcpCmsComponents: {
        ids: [
          'ElectronicsHompageSplashBannerComponent',
          // 'ElectronicsHomepageProductCarouselComponent',
          'ProductImagesComponent',
          'ProductListComponent',
        ],
      },
    }),
```

- build the fresh SSR app and run SSR
- disable JS in the browser
- visit homepage in the browser in DESKTOP viewport
- open DevTools Elements tab and search for the `fetchpriority` attribute - it should be present (and have value 'high') only for the image of the Homepage Splash Banner "Save Big"
- visit any PDP
- open DevTools Elements tab and search for the `fetchpriority` attribute - it should be present (and have value 'high') only for the main product image
- visit any PLP
- open DevTools Elements tab and search for the `fetchpriority` attribute - it should be present (and have value 'high') only for the FIRST product on the listing
- change the config provided in the first step - replace `ElectronicsHompageSplashBannerComponent` with `ElectronicsHomepageProductCarouselComponent`
- visit homepage again
- open DevTools Elements tab and search for the `fetchpriority` attribute - it should be present (and have value 'high') only for the image of the FIRST item in the first product carousel

2) test dynamic marker in ID:
- remove the config of `lcpCmsComponents` from the previous (1) phase of QA
- (implicitly the default `lcpCmsComponents.idMarker` is `"__cxLCP__"` but customers can change it to anything)
- open backoffice and apped the marker `__cxLCP__` to the IDs of the components listed below. The ID field can be found in the "Administration" tab after opening specific component's details in pane in the bottom-half of the screen
  - in backoffice go to: WCMS -> Components -> and in the top right corner search the following phrases (but then look only for the ONLINE "SPA Electronics" catalogue):
     - "Electronics Homepage Splash Banner Component" (including doublequotes!)
     - "Electronics Homepage Product Carousel"
     - "Product List Component"
     - "ProductImagesComponent"
   - then apped the suffix marker to those components' IDs 
 - repeat the testing steps from the previous (1) phase of QA 

REMINDER: after the testing please remove the marker suffixes from the edited CMS component IDs


projects/storefrontlib/shared/lcp-context/provide-lcp-presence.directive.ts
Platonn, line 1: Note: this directive is not used anywhere in this PR, but added it for completeness - to allow customers to possibly provide a custom LCP_PRESENCE at any DOM level, if only they want it (without a dependency on the CMS)

projects/storefrontlib/shared/components/media/media.component.ts
```diff
@@ -95,7 +95,7 @@ export class MediaComponent implements OnChanges {
    * and it will automatically use `loading="eager"` for optimal performance.
    * Valid values are: 'low', 'auto', or 'high'.
    */
-  @Input() fetchPriority?: ImageFetchPriority;
+  @Input() fetchPriority?: ImageFetchPriority | null;
```
pawelfras, line 98: `query:` Is the `null` type required here? Since the input is not required, maybe you could check whether value is defined or not.
Btw. I'm wondering if `?`(optional) makes a difference in the case of input.

projects/storefrontlib/shared/lcp-context/lcp-presence.model.ts
```diff
@@ -0,0 +1,13 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * Tells whether the component contains LCP (Largest Contentful Paint) element.
+ */
+export enum LcpPresence {
+  HAS_LCP = 'HAS_LCP',
+  NO_LCP = 'NO_LCP',
+}
```
pawelfras, line 13: `minor:` As we discussed, let's consider transforming it into a type unless you see a possibility that users can extend the enum.

feature-libs/product/image-zoom/components/product-image-zoom/product-image-zoom-product-images/product-image-zoom-product-images.component.html
```diff
@@ -1,5 +1,11 @@
 <ng-container *ngIf="mainImage$ | async as main">
-  <cx-media [container]="main" (click)="triggerZoom(true)"></cx-media>
+  <ng-container *cxLcpContext="let lcpContext">
+    <cx-media
+      [container]="main"
+      (click)="triggerZoom(true)"
+      [fetchPriority]="lcpContext.fetchPriority$ | async"
+    ></cx-media>
+  </ng-container>
```
pawelfras: `query:` Have you perhaps considered using the structural directive directly on the target component?
```suggestion
    <cx-media
      *cxLcpContext="let lcpContext"
      [container]="main"
      (click)="triggerZoom(true)"
      [fetchPriority]="lcpContext.fetchPriority$ | async"
    ></cx-media>
```

Platonn: Yes I considered it. And decided to do it in ng-container for readability - to not mix it with other inputs on the `<cx-media>` element.

projects/storefrontlib/shared/lcp-context/lcp-presence.model.ts
Platonn: As we discussed, I'm having problem to understand the real value of changing it to an union type of string constants.
Using enum for a dictionary of posisble values is a standard approach in Spartacus.

--- 
I'm interested also in your opinion whether the naming isn't too enigmatic: LcpPresence, HAS_LCP, NO_LCP.

projects/storefrontlib/shared/components/media/media.component.ts
Platonn: `?` allows `undefined`

In this PR we're assigning a value from `async` pipe which has `null` in its return union type as a fallback when observable doesn't emit yet. That's why I needed to add `null` here.

Alternatively in every parent component's template I could fallback to `undefined` explicitly:
```diff
-   [fetchPriority]="lcpContext.lcpPresence$ | async"
+   [fetchPriority]="(lcpContext.lcpPresence$ | async) ?? undefined"
```

pawelfras: Understood, thanks. Indeed, union type sounds better than setting an `undefined` in every parent's template

feature-libs/product/image-zoom/components/product-image-zoom/product-image-zoom-product-images/product-image-zoom-product-images.component.html
pawelfras: My take - using the structural directive directly on the target component looks to me more natural, and it's something we used to when using Angular's structural directives. (In addition, it provides less boilerplate)

projects/storefrontlib/shared/lcp-context/lcp-presence.model.ts
pawelfras: You're right, union type does not provide that much value to break the standard for dictionaries in Spartacus. Consistency FTW 💪

pawelfras: About the naming - it's fine for me 👌

pawelfras (dismissed):
LGTM 👌

pawelfras (approved):
`nitpick:` consider putting the structural directive as a first HTML attribute

PR #20485, opened 2025-06-26
test: Split mechanism for a11y e2e

sdrozdsap:
Distribution logic that I used is not very sophisticated, but gives satisfying result 
image


Zeyber (changes requested):
Nice!

I think we could add some comments to explain what is happening with the test-distribution script and something that explains strategy. For us, we know what's happening here. But anybody exploring this area of the code unaware could benefit from further context. It could also encourage them to use it ;)

Zeyber (approved):

PR #20492, opened 2025-06-27
fix: add global function for re-init the payment

Matejk00:
Closes: CXSPA-10052


integration-libs/opf/global-functions/core/facade/opf-global-functions.service.ts
```diff
@@ -372,4 +385,48 @@ export class OpfGlobalFunctionsService implements OpfGlobalFunctionsFacade {
...
+        );
+
+        return Promise.resolve(true);
+      });
+    };
+  }
+
+  protected getPaymentOptionId(providedId?: number): Observable<number> {
+    if (providedId) {
+      return new Observable((observer) => {
```
rmch91: Maybe I'm missing something, but why not to use rx.js `of` for this? `return of(providedId)`? Instead of creating and completing observable manually..

integration-libs/opf/global-functions/core/facade/opf-global-functions.service.ts
```diff
@@ -372,4 +385,48 @@ export class OpfGlobalFunctionsService implements OpfGlobalFunctionsFacade {
...
+        observer.next(providedId);
+        observer.complete();
+      });
+    }
+
+    // Get from metadata store (local storage)
+    return new Observable((observer) => {
```
rmch91: similar concern as above, 
I think you can use rx.js operators and not create nested subscription
```
return this.opfMetadataStoreService
        .getOpfMetadataState().pipe(switchMap((metadata) => {
const storedId =
            metadata.selectedPaymentOptionId ||
            metadata.defaultSelectedPaymentOptionId;

return storedId ? of(storedId) : throwError(...)
}))
```

rmch91: and `take(1)` probably

integration-libs/opf/global-functions/core/facade/opf-global-functions.service.ts
```diff
@@ -41,13 +44,21 @@ export class OpfGlobalFunctionsService implements OpfGlobalFunctionsFacade {
...
+  protected opfPaymentEventsService = inject(OpfPaymentEventsService);
   protected loaderSpinnerCpntRef: void | Observable<
     ComponentRef<any> | undefined
   >;
   protected _readyForScriptEvent: Subject<string> = new Subject();
   readyForScriptEvent$: Observable<string> =
     this._readyForScriptEvent.asObservable();
 
+  // Event for re-initiating payment form
+  protected _reinitiatePaymentEvent: Subject<number | undefined> =
```
rmch91: Probably leftover. I think this is not used any more. There is no `.next()` call inside of this class.
You use `emitReinitiatePaymentEvent` from `OpfPaymentEventsService`.

rmch91 (changes requested):

integration-libs/opf/global-functions/core/facade/opf-global-functions.service.ts
Matejk00: It seems that I pushed old version of the opf-global-functions.service.ts
It's adjusted now

rmch91 (approved):

PR #20501, opened 2025-07-02
chore: Use shared build for cypress jobs

sdrozdsap:


.github/workflows/ci.yml
```diff
@@ -147,14 +252,28 @@ jobs:
             ~/.cache/Cypress
           key: nodemodules-${{ github.event.pull_request.base.sha }}
           restore-keys: nodemodules-${{ github.event.pull_request.base.sha }}
+      - name: Restore built libraries
```
Zeyber: I'm seeing this repeated a few times. It is possible to split actions into other files which could shorten the length of this .yml. Do you think it would make sense for actions in this scenario?
https://docs.github.com/en/actions/tutorials/creating-a-composite-action

Zeyber (changes requested):

.github/workflows/ci.yml
sdrozdsap: Makes sense! Let me investigate that

.github/workflows/ci.yml
```diff
@@ -280,12 +441,19 @@ jobs:
...
+        uses: actions/cache@v4
+        with:
+          path: |
+            dist
+          key: spartacus-libs-${{ github.event.pull_request.head.sha || github.run_id }}
+          fail-on-cache-miss: true
       - name: Build SSR Server
-        run: npm run build:libs && npm run build:ssr:local-http-backend
+        run: npm run build:ssr:local-http-backend
```
rmch91: I'm thinking about `npm scripts` commands. 
Some time ago, because of builder change, guys from Asterix changed `npm run build` to `npm run build:csr`. For automations that means that we have to change it in spartacus repos, in azure pipeline, and other possible automations built around the spartacus. 

That's why, I'm thinking now, maybe it is not worth to create a bash function wrappers inside of the spartacus repo, for each npm script. And call bash functions from automations. 
Thanks to this, if we will need to change for example, `npm run build:ssr:local-http-backend` or any other, for some reason - we will have only one source where we have to do this, and maybe it will not break anything else. What do you think?

.github/workflows/ci.yml
rmch91: Good point

ci-scripts/npm-commands.sh
```diff
@@ -0,0 +1,41 @@
...
+    echo "  build_libs                  - Build Spartacus libraries"
+    echo "  build_csr                   - Build CSR application"
+    echo "  build_ssr                   - Build SSR application for CI"
+    echo "  build_ssr_local_http        - Build SSR with local HTTP backend"
+    echo ""
+}
+
+# If script is executed directly (not sourced), show usage
+if [[ "${BASH_SOURCE[0]}" == "${0}" ]]; then
+    show_usage
```
Zeyber, line 46: Nice to have this fallback!

Zeyber (dismissed):

.github/workflows/ci.yml
```diff
@@ -98,194 +86,209 @@ jobs:
...
-        run: npm ci
+          cache-key: ${{ github.event.pull_request.head.sha || github.run_id }}
+          fail-on-cache-miss: true
       - name: Build SSR Server
-        run: npm run build:libs && npm run build:ssr:local-http-backend
+        run: |
+          source ci-scripts/npm-commands.sh
+          build_ssr_local_http
       - name: Run SSR tests
         run: npm run test:ssr:ci --verbose
```
rmch91: If we are already here, can you please move this also to bash wrapper?

rmch91 (changes requested):

rmch91 (dismissed):

rmch91 (approved):

PR #20507, opened 2025-07-03
fix: same header slots are used to preventing visual shifts after CSR in desktop and mobile (CXSPA-10434)

kimhw0630:
Closes: https://jira.tools.sap/browse/CXSPA-10434


projects/core/src/features-config/feature-toggles/config/feature-toggles.ts
```diff
@@ -760,6 +760,19 @@ export interface FeatureTogglesInterface {
...
+   *
+   * On desktop devices (non-mobile), some header and navigation elements currently render
+   * only after client-side rendering (CSR), leading to a significant layout shift. This
+   * negatively impacts the user experience and results in a poor Cumulative Layout Shift (CLS)
+   *
+   * When enabled:
+   * - Desktop uses the same header slot structure as mobile.
+   * - Reduces layout shift and improves perceived performance and visual stability.
+   */
+  unifiedDefaultHeaderSlotsAcrossBreakpoints?: boolean;
```
Platonn: 1. Let's mention in this JSDocs that for this toggle to fully work, existing customers will need to lso replace `provideConfig(layoutConfig)` in their codebase to `provideConfigFactory(layoutConfigFactory)`
2. Please[ adjust our installation schematics](https://github.com/SAP/spartacus/blob/a3954e1ed8cb86962633f832a1589d61ecd741aa/projects/schematics/src/add-spartacus/configuration.ts#L86-L98) (and its [unit tests](https://github.com/SAP/spartacus/blob/a3954e1ed8cb86962633f832a1589d61ecd741aa/projects/schematics/src/migrations/mechanism/scaffold-app-structure/scaffold-app-structure_spec.ts#L55-L63)) to place in new apps `provideConfigFactory(layoutConfigFactory)`, but no more `provideConfig(layoutConfig)` (note: no feature toggle needed in schematics to control that; schematics are just for creating fresh apps). Then please test it, by creating a fresh app with libs from verdaccio and verify the fresh app contains `provideConfigFactory(layoutConfigFactory)`.
3. I'm having problem to understand this sentence `Feature flag to enable using the same header slots are used to preventing visual shifts after CSR and improving CLS.`. Would you mind rephrasing it?
4. `...some header and navigation elements currently render...` I suggest to use past tense for the previous behavior. e.g. `...some header and navigation elements previously were rendered...`

projects/storefrontapp-e2e-cypress/cypress/e2e/regression/anonymous-consents/anonymous-consents-config-flow.e2e.cy.ts
```diff
@@ -46,7 +46,7 @@ context('Anonymous consents - config flow', () => {
         cy.wait(`@${homePage}`).its('response.statusCode').should('eq', 200);
 
         // Make sure anonymous user is loaded
-        cy.get('cx-login [role="link"]').should('be.visible');
+        cy.loginRegisterLinkSelect(false).should('be.visible');
```
Platonn: [MINOR]
I've read the implementation and understood that there is more going on (also waiting a specific URL when a click is made). Moreover the function sometimes returns an element and sometimes not - depending whether a click was made. Which makes the util harder to reason about.

To reduce the complexity, I suggest to simplify it

leave only a simple util: `cy.getLoginRegisterLink()` -> which returns the link item. and then you can use a native cypress `.click()` on it.

I'm not sure it's worth to extract the waiting logic to an util. If yes, then I'd not mix it with the implementation of the simplified util `cy.getLoginRegisterLink()`

projects/storefrontlib/recipes/config/layout-config.ts
```diff
@@ -126,3 +130,32 @@ export const layoutConfig: LayoutConfig = {
...
+      'PreHeader',
+      'SiteContext',
+      'SiteLinks',
+      'SiteLogo',
+      'SearchBox',
+      'SiteLogin',
+      'MiniCart',
+      'NavigationBar',
+    ];
+  }
```
Platonn: I'm aware we're already after a few rounds of review. Forgive me please that-late observation. I can see that we can also remove the `lg`-breakpoint specific config, because it's identical with the new general `slots` config.

```suggestion
  if (
    unifiedDefaultHeaderSlotsAcrossBreakpoints &&
    config.layoutSlots &&
    config.layoutSlots.header &&
    'slots' in config.layoutSlots.header
  ) {
    if ('lg' in config.layoutSlots.header) {
      delete config.layoutSlots.header.lg;
    }
    
    config.layoutSlots.header.slots = [
      'PreHeader',
      'SiteContext',
      'SiteLinks',
      'SiteLogo',
      'SearchBox',
      'SiteLogin',
      'MiniCart',
      'NavigationBar',
    ];
  }
```

See original config above:
Image

Platonn (dismissed):

Platonn (dismissed):

Platonn (approved):

PR #20516, opened 2025-07-09
refactor: Exported CMS components from OrderRootModule to reuse in other libraries

suprishi:
https://jira.tools.sap/browse/CXSPA-10433

 - Exported the array of CMS components from `OrderRootModule` for re-using in other libraries
 - Imported the above array in `S4ServiceRootModule` to avoid hardcoding duplicate entries


vigneshkrish8691 (approved):

rmch91 (approved):
LGTM

PR #20521, opened 2025-07-11
fix: displaying more than 1 gigya raas component in a page gives error

anjana-bl:
JIRA: https://jira.tools.sap/browse/CXSPA-10586

### DIFF
--- a/integration-libs/cdc/components/gigya-raas/gigya-raas.guard.ts
+++ b/integration-libs/cdc/components/gigya-raas/gigya-raas.guard.ts
@@ -49,8 +49,10 @@ export class GigyaRaasGuard implements CanActivate {
         // Allow activation only if all checks pass, else return first non-true value
         return combineLatest(guardResults$).pipe(
           map((results) => {
-            const firstNonTrue = results.find((result: GuardResult) => result !== true);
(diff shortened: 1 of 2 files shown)


integration-libs/cdc/components/gigya-raas/gigya-raas.guard.ts
```diff
@@ -29,21 +37,45 @@ export class GigyaRaasGuard implements CanActivate {
...
-        }
-        return of(true);
+        // Run guard checks for each component
+        const guardResults$ = componentData.map((data) =>
+          this.checksToProcess(data)
+        );
+
+        // Allow activation only if all checks pass, else return first non-true value
+        return combineLatest(guardResults$).pipe(
+          switchMap((results) => {
```
vigneshkrish8691: Can we use map() instead as we are transforming values ?

Something like this : 
return combineLatest(guardResults$).pipe(
  map((results) => results.find((r) => r !== true) ?? true)
);

anjana-bl: yes, done

vigneshkrish8691 (approved):

PR #20499, opened 2025-07-01
feat: scrolling product carousel 

Platonn:
Introducing new feature controlled with a feature toggle `productCarouselScrolling`.

When enabled, the new carousel component `<cx-carousel-scrolling>` will be used
in the following components instead of the old `<cx-carousel>`:
- `ProductCarouselComponent`
- `ProductReferencesComponent`
- `ProductImagesComponent` and related `ProductImageZoomProductImagesComponent`

The previous carousel had number of issues:
- Caused huge layout shift when transitioning from SSR to CSR on desktop viewport, because in SSR there was rendered just 1 carousel item, but in desktop CSR 4 items appeared after a while (especially noticeable with Chrome DevTools Network throttling)
- Eagerly-loaded images from also from invisible slides, even when Spartacus was configured to lazy load of all images: `provideConfig({ imageLoadingStrategy: ImageLoadingStrategy.LAZY})`)
- Was not swipe-friendly on mobile devices

The new carousel:
- Doesn't suffer from huge layout shifts when transitioning from SSR to CSR anymore, becasue of rendering the same HTML both in SSR and when CSR kicks in after a delay, so the same number of carousel items is visible.
- It's lazy loading invisible images thanks to native horizontal scrolling (when Spartacus is configured to lazy load all images: `provideConfig({ imageLoadingStrategy: ImageLoadingStrategy.LAZY})`) 
- It's swipe-friendly on touch devices thanks to its native horizontal scrolling

Note: a11y review and UX review passed

fixes https://jira.tools.sap/browse/CXSPA-10007

---

**QA steps:**
- checkout to this PR branch and run `npm run dev:ssr`
- in Chrome Dev Tools Network Tab enable Network throttling: Fast 4G
image

The following steps will show comparison between BEFORE (previous behavior) vs AFTER (new behavior):

**BEFORE:**
1. in the file [spartacus-features.module.ts](https://github.com/SAP/spartacus/blob/7fe153a15aa66b1b569d5404c8645a6ff5bd6aea/projects/storefrontapp/src/app/spartacus/spartacus-features.module.ts#L406) disable the new feature toggle: `productCarouselScrolling: false` and wait for the rebuild
  image
2. on desktop viewport visit http://localhost:4000/electronics-spa/en/USD/ (not port 4200! because it's SSR) and zoom out to see the whole page (including carousels)
3. observe the huge one carousel element appearing first from SSR HTML, and after a delay it changes to 4 items

[...]

PR #20525, opened 2025-07-15
fix: carousel-scrolling.component - minor fixes after review #20499

Platonn:
Fixes after review https://github.com/SAP/spartacus/pull/20499
- refactor: rename generic type param
- refactor: move MockFeatureConfigService definition out of the main "describe" block
- refactor: extract keyboard event keys to enum

Additionally:
- fix: observe _scrollingArea_ instead _window_ for resizing events
- refactor: remove unnecessary keyboard event handlers, shareEvent() and keyboardEvent output - as not needed, because previous/next buttons are not focusable
- test: add missing unit tests for keyboard navigation (with and without "cxFocusableCarouselItem" directive) and for default trackByFn
- refactor: refactor tests for readability (e.g. extract common code to beforeEach blocks, rename some variables)

related to https://jira.tools.sap/browse/CXSPA-10007


projects/storefrontlib/shared/directives/horizontal-scrolling-position/horizontal-scrolling-position.directive.ts
```diff
@@ -62,8 +63,13 @@ export class HorizontalScrollingPositionDirective
   protected windowResizeSubscription: Subscription | undefined;
   protected _scrollingAreaWidth$ = new BehaviorSubject<number>(0);
 
-  public readonly scrollingAreaWidth$ =
-    this._scrollingAreaWidth$.asObservable();
+  /**
+   * Width of the scrolling area, debounced every 300ms in case of continuous resizing.
+   */
+  public readonly scrollingAreaWidth$ = this._scrollingAreaWidth$.pipe(
+    debounceTime(300),
```
pawelfras, line 70: `query:` Is there any reason that it has to be 300ms exactly?

pawelfras (approved):

projects/storefrontlib/shared/directives/horizontal-scrolling-position/horizontal-scrolling-position.directive.ts
Platonn: No, it's an arbitrary value. 
I've followed the same value that is used in `WindowRef.resize$`:
https://github.com/SAP/spartacus/blob/29c6a5525ff90a37d48dbc884eab444796523f42/projects/core/src/window/window-ref.ts#L97-L102

PR #20535, opened 2025-07-18
fix: OPF payment wrapper render on the explicit flow

Matejk00:
Closes: CXSPA-10623

### DIFF
--- a/integration-libs/opf/checkout/components/opf-checkout-payments/opf-checkout-payments.component.ts
+++ b/integration-libs/opf/checkout/components/opf-checkout-payments/opf-checkout-payments.component.ts
@@ -88,6 +88,12 @@ export class OpfCheckoutPaymentsComponent implements OnInit, OnDestroy {
   @Input()
   forceDefaultPaymentOptionInputSelection? = false;
 
+  @Input()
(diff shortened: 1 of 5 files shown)


integration-libs/opf/checkout/components/opf-checkout-payments/opf-checkout-payments.component.ts
```diff
@@ -88,6 +88,12 @@ export class OpfCheckoutPaymentsComponent implements OnInit, OnDestroy {
   @Input()
   forceDefaultPaymentOptionInputSelection? = false;
 
+  @Input()
+  renderPaymentWrapper? = true;
```
rmch91, line 92: I think it is not clear naming, maybe shouldDisplayRenderPaymentWrapper? or something that indicates boolean

rmch91 (approved):

integration-libs/opf/checkout/components/opf-checkout-payments/opf-checkout-payments.component.ts
Matejk00: As we discussed on Slack, let's keep the current approach.

PR #20545, opened 2025-07-23
docs: Update a11y-testing.md

Zeyber:
Closes: CXSPA-10616


projects/storefrontapp-e2e-cypress/docs/a11y-testing.md
```diff
@@ -8,6 +8,10 @@ Access Continuum is integrated into our Cypress E2E testing framework to automat
 
 ## Setup
 
+First, we need to install the Access Continuum package. To do this, we need to set the `CONTINUUM_REGISTRY_TOKEN` (see https://wiki.one.int.sap/wiki/display/spar/Spartacus+Accessibility+Feature+Compliance#SpartacusAccessibilityFeatureCompliance-AccessContinuum to acquire this) by using the command `export CONTINUUM_REGISTRY_TOKEN=XXXXXXXXXXXXXXXX` in the terminal. Then we need to install the packages in the `storefrontapp-e2e` directory as usual with `npm install`.
```
sdrozdsap: This is fine, but for the time being we still use `COMMON_SAP_ARTIFACTORY_TOKEN` to authenticate to SAP npm registry. And this token can be obtained by any employee. I've got it from https://common.repositories.cloud.sap/ui/packages (you need to go to Set Me Up there and find Npm -> acservices and then generate token
image


But I believe this token ca be also obtained form https://ui.repositories.cloud.sap/www/webapp/users (usual S-user base64 token should work)

I think its worth mentioning here. 

p.s. We still plan to switch to using levelaccess registry, but they haven't yet published a version with the fix we needed, so we ll need to wait a bit longer

projects/storefrontapp-e2e-cypress/docs/a11y-testing.md
```diff
@@ -8,6 +8,10 @@ Access Continuum is integrated into our Cypress E2E testing framework to automat
 
 ## Setup
 
+First, we need to install the Access Continuum package. To do this, we need to set the `CONTINUUM_REGISTRY_TOKEN` (see https://wiki.one.int.sap/wiki/display/spar/Spartacus+Accessibility+Feature+Compliance#SpartacusAccessibilityFeatureCompliance-AccessContinuum to acquire this) by using the command `export CONTINUUM_REGISTRY_TOKEN=XXXXXXXXXXXXXXXX` in the terminal. Then we need to install the packages in the `storefrontapp-e2e` directory as usual with `npm install`. 
+
+*Note that running `npm install` without setting the registry token will still install the packages to run e2e tests successfully. But any test using continuum libraries (ie. a11y tests) will not perform any accessibity checks.*
```
sdrozdsap: I would suggest to recommend `npm ci` instead of install in the docs

sdrozdsap (changes requested):
Looks great! Could you please just add a note about the COMMON_SAP_ARTIFACTORY_TOKEN token that we use currently

projects/storefrontapp-e2e-cypress/docs/a11y-testing.md
Zeyber: Ah yes, good point I missed. I've added the info. But I wonder if we should instead use the SAP Repository only? Then employees can simply generate their own tokens rather than tracking somebody down.

sdrozdsap: I guess it would make sense, but the downside of this is that in order to update the package we would need to download it from levelaccess registry and upload into ours. However considering that we probably wouldn't update it very often I'd say it should be fine

sdrozdsap (approved):

projects/storefrontapp-e2e-cypress/docs/a11y-testing.md
Zeyber: I guess that would be one man's maintenance vs many tracking down the correct token. But I think the whole reason for doing this is the levelaccess doesn't update their package. I'm a bit confused because we have a more up-to-date version of continuum with the patch and levelaccess doesn't. Maybe not the discussion to have here though.

hackergil (approved):

PR #20553, opened 2025-07-25
fix: should not show the customer360 icon without customer360 module

niehuayang:
for details see: https://jira.tools.sap/browse/CXSPA-10545


i53577 (approved):
LGTM

PR #20556, opened 2025-07-25
fix: remove pageFold from the default layout config

Platonn:
Previously the default Spartacus layout config contained the property `pageFold`
for the following layouts:
- `LandingPage2Template`
- `CategoryPageTemplate`
- `ProductDetailsPageTemplate`

Now introduced a feature toggle for a fix: `defaultLayoutConfigWithoutPageFold`. When this feature toggle is enabled, the `pageFold` property is removed from those layout configs.

It is to improve the CLS (Cumulative Layout Shift) metric. Previously the `pageFold` property
caused the CMS components to be rendered only after a small delay even in SSR pages,
which caused a layout shift.

⚠️ To fully enable this feature toggle, you need to also replace `provideConfig(layoutConfig)`
in your codebase with `provideConfigFactory(layoutConfigFactory)`.

fixes https://jira.tools.sap/browse/CXSPA-10694

## QA steps:

1. Add the following workaround code snippet in app module to provide widht/height to banned images, which are missing in CMS response. If you don't provide it, the Lighthouse CLS will complain about it. And those banners CLS is out of scope for this ticket, so let's use the workaround.

Please scroll down to the very bottom of this PR description to the Appendix to copy-paste the code snippet to your app module. (I didn't want to put it here because it's huge)

2. run Dev SSR server: `npm run dev:ssr`

3. open homepage in chrome http://localhost:4000/electronics-spa/en/USD/

4. zoom out to 75%

5. In chrome devtools open Lighthouse tab and select "desktop". But don't run the Lighthouse audit yet!

**5) HOMEPAGE BEFORE:** 
👉 In `spartacus-features.module.ts` disable the new feature toggle: set `defaultLayoutConfigWithoutPageFold: false` and WAIT FOR THE APP FOR RECOMPILE

and in chrome devtools Lighthouse run "Analyze page load" 
image

Result:
CLS is red

image

please scroll a bit to bottom and select a small button [CLS] which filters only the CLS complaints:

image

see it complains about the Section3

**5) HOMEPAGE AFTER:**
👉 In `spartacus-features.module.ts` enable the new feature toggle: set `defaultLayoutConfigWithoutPageFold: true` and WAIT FOR THE APP FOR RECOMPILE

and in chrome devtools Lighthouse run "Analyze page load" 

Result:
CLS is green

image

scroll a bit to bottom and select a small button [CLS] which filters only the CLS complaints:

image

there is no complaint about the Section3 element 👍 

6. PDP: open PDP in chrome http://localhost:4000/electronics-spa/en/USD/

**6) PDP BEFORE:** 
👉 In `spartacus-features.module.ts` disable the new feature toggle: set `defaultLayoutConfigWithoutPageFold: false` and WAIT FOR THE APP FOR RECOMPILE

and in chrome devtools Lighthouse run "Analyze page load" 
image

Result:
CLS is yellow

image

please scroll a bit to bottom and select a small button [CLS] which filters only the CLS complaints:

image

see it complains about the Tabs element

**6) PDP AFTER:**
👉 In `spartacus-features.module.ts` enable the new feature toggle: set `defaultLayoutConfigWithoutPageFold: true` and WAIT FOR THE APP FOR RECOMPILE

and in chrome devtools Lighthouse run "Analyze page load" 

Result:
CLS is green

image

scroll a bit to bottom and select a small button [CLS] which filters only the CLS complaints:

there is no complaints about the Tabs element 👍 

image

7. There is no difference BEFORE vs AFTER for the Category page. CLS is green regardless of the feature toggle. But I decided to remove `pageFold` from this config for consistency.

## Appendix - workaround snippet to provide dimensions of banners:
```typescript

import { Injectable, Provider } from '@angular/core';
import {
  CmsBannerComponent,
  CmsBannerComponentMedia,
  CmsResponsiveBannerComponentMedia,
  CmsStructureModel,
  Occ,
  OccCmsPageNormalizer,
} from '@spartacus/core';

/**
 * This is a workaround to extract dimensions from a URL of a banner image, based on a filenames convention in our sample data
 * (yes it's a workaround! ideally dimensions should be defined in CMS!)
 */
export class CustomOccCmsPageNormalizer extends OccCmsPageNormalizer {
  override convert(
    source: Occ.CMSPage,
    target: CmsStructureModel = {}
  ): CmsStructureModel {
    let result = super.convert(source, target);

    result = this.populateBannerImagesDimensions(result);

    return result;
  }

  protected populateBannerImagesDimensions(
    result: CmsStructureModel
  ): CmsStructureModel {
    if (!result.components) {
      return result;
    }

    // Iterate through all components
    (result.components || []).forEach((component) => {
      // ignore components that are not banners
      if (
        !component.typeCode ||
        !['SimpleResponsiveBannerComponent', 'SimpleBannerComponent'].includes(
          component.typeCode
        )
      ) {
        return;
      }

      const copyDimensionsFromUrlToSeparateProperties = (
        media: CmsBannerComponentMedia
      ) => {
        if (media?.url) {
          const dimensions = this.extractDimensionsFromUrl(media.url);
          // Add dimensions to media object if found
          if (dimensions.width) {
            (media as any).width = dimensions.width; // although width is not defined in CMS, we add it here
          }
          if (dimensions.height) {
            (media as any).height = dimensions.height; // although height is not defined in CMS, we add it here
          }
          if(media.url.includes('Elec-1400x80-HomeDiscount-EN-01-1400W.jpg')){
            (media as any).height = 70; // there is wrong filename in our sample data
          }
        }
      };

      const bannerComponent = component as CmsBannerComponent;
      // Note:
      // - SimpleBannerComponent has "media" property with a single image
      // - SimpleResponsiveBannerComponent has "media" property with a object containing images
      //    for different media formats (in separate properties)

      if (!bannerComponent.media) {
        return;
      }
      if (bannerComponent.typeCode === 'SimpleBannerComponent') {
        copyDimensionsFromUrlToSeparateProperties(
          bannerComponent.media as CmsBannerComponentMedia
        );
      }
      if (bannerComponent.typeCode === 'SimpleResponsiveBannerComponent') {
        // Process each media format (mobile, tablet, desktop, widescreen)
        Object.values(
          bannerComponent.media as CmsResponsiveBannerComponentMedia
        ).forEach((media) => {
          copyDimensionsFromUrlToSeparateProperties(media);
        });
      }
    });

    return result;
  }

  /**
   * Extracts dimensions from a URL of a banner image, based on a filenames convention in our sample data
   * (yes it's a workaround! ideally dimensions should be defined in CMS!)
   */
  protected extractDimensionsFromUrl(url: string): {
    width?: number;
    height?: number;
  } {
    // Banner images in our sample data happen to follow the pattern `somename-WIDTHxHEIGHT-somename...`
    // so we can leverage it to extract the dimensions
    const pattern = /\/medias\/[^-]+-(\d+)x(\d+)-[^-]+/;
    const match = url.match(pattern);
    if (match) {
      const width = parseInt(match[1], 10);
      const height = parseInt(match[2], 10);
      return { width, height };
    } else {
      return {};
    }
  }
}

export const workaroundExtractBannerDimensionsFromUrl: Provider = {
  provide: OccCmsPageNormalizer,
  useExisting: CustomOccCmsPageNormalizer,
};

```

and then add to app.module providers:
```typescript
providers: [
  /*...*/
  workaroundExtractBannerDimensionsFromUrl
]
```

### DIFF
--- a/projects/core/src/features-config/feature-toggles/config/feature-toggles.ts
+++ b/projects/core/src/features-config/feature-toggles/config/feature-toggles.ts
@@ -959,5 +959,5 @@ export const defaultFeatureToggles: Required<FeatureTogglesInterface> = {
   unifiedDefaultHeaderSlotsAcrossBreakpoints: false,
   reserveSpaceForImagesOnPdpAndPlp: false,
   lazyLoadImagesByDefault: false,
-  defaultLayoutConfigWithoutPageFold: false
+  defaultLayoutConfigWithoutPageFold: false,
 };
--- a/projects/core/src/features-config/feature-toggles/config/feature-toggles.ts
+++ b/projects/core/src/features-config/feature-toggles/config/feature-toggles.ts
@@ -827,6 +827,24 @@ export interface FeatureTogglesInterface {
    * Set to `true` to enable lazy loading by default.
    */
   lazyLoadImagesByDefault?: boolean;
+
+  /**
+   * Previously the default Spartacus layout config contained the property `pageFold`
+   * for the following layouts:
+   * - `LandingPage2Template`
+   * - `CategoryPageTemplate`
+   * - `ProductDetailsPageTemplate`
+   *
+   * When this feature toggle is enabled, the `pageFold` property is removed from those layout configs.
+   *
+   * It is to improve the CLS (Cumulative Layout Shift) metric. Previously the `pageFold` property
+   * caused the CMS components to be rendered only after a small delay even in SSR pages,
+   * which caused a layout shift.
+   *
+   * ⚠️ To fully enable this feature toggle, you need to also replace `provideConfig(layoutConfig)`
+   * in your codebase with `provideConfigFactory(layoutConfigFactory)`.
+   */
+  defaultLayoutConfigWithoutPageFold?: boolean;
 }
 
 export const defaultFeatureToggles: Required<FeatureTogglesInterface> = {
@@ -941,4 +959,5 @@ export const defaultFeatureToggles: Required<FeatureTogglesInterface> = {
   unifiedDefaultHeaderSlotsAcrossBreakpoints: false,
   reserveSpaceForImagesOnPdpAndPlp: false,
   lazyLoadImagesByDefault: false,
+  defaultLayoutConfigWithoutPageFold: false
 };
--- a/projects/storefrontapp/src/app/spartacus/spartacus-features.module.ts
+++ b/projects/storefrontapp/src/app/spartacus/spartacus-features.module.ts
@@ -408,7 +408,7 @@ if (environment.cpq) {
         unifiedDefaultHeaderSlotsAcrossBreakpoints: true,
         reserveSpaceForImagesOnPdpAndPlp: true,
         lazyLoadImagesByDefault: true,
-        defaultLayoutConfigWithoutPageFold: true
+        defaultLayoutConfigWithoutPageFold: true,
       };
       return appFeatureToggles;
     }),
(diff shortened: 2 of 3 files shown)


pawelfras (dismissed):
Reviewed and QAd. LGTM

pawelfras (dismissed):

pawelfras (approved):

PR #20557, opened 2025-07-25
chore: fix bug preventing the resume of checkout flow when returning from login

ercultimate:


pawelfras:
PR left for the reference. Can be closed in favour of https://github.com/SAP/spartacus/pull/20572

PR #20551, opened 2025-07-24
feat: Custom Login Page feature

Pucek9:
Closes [CXSPA-9863](https://jira.tools.sap/browse/CXSPA-9863)

PR #20565, opened 2025-07-30
feat: Custom Login Page feature

Pucek9:
Continuation of https://github.com/SAP/spartacus/pull/20551
Closes [CXSPA-9863](https://jira.tools.sap/browse/CXSPA-9863)


Pucek9:
Unit tests WIP

feature-libs/user/account/components/login-form/login-form-component.service.ts
```diff
@@ -22,14 +26,28 @@ import { tap, withLatestFrom } from 'rxjs/operators';
 
 @Injectable()
 export class LoginFormComponentService {
+  protected authConfigService = inject(AuthConfigService);
+  protected isNewAuthFlow = inject(FeatureConfigService).isEnabled(
+    'authorizationCodeFlowByDefault'
+  );
```
pawelfras: `major:` let's inject the whole `FeatureConfigService` in case we need more feature toggles in this file in the future, and also let's change the access modifier to private so customers can't access this service and build their logic on top of that.


```suggestion
  private featureConfigService = inject(FeatureConfigService);
```

feature-libs/user/account/components/login-form/login-form-component.service.ts
```diff
@@ -22,14 +26,28 @@ import { tap, withLatestFrom } from 'rxjs/operators';
...
   constructor(
     protected auth: AuthService,
     protected globalMessage: GlobalMessageService,
     protected winRef: WindowRef
-  ) {}
+  ) {
+    if (this.isNewAuthFlow) {
+      this.method = 'POST';
+      this.action = this.authConfigService?.getCustomLoginFormEndpoint();
+    }
```
pawelfras: According to the previous comment: 
```suggestion
    if (this.featureConfigService.isEnabled('authorizationCodeFlowByDefault')) {
      this.method = 'POST';
      this.action = this.authConfigService?.getCustomLoginFormEndpoint();
    }
```

It also applies to other places where `isNewAuthFlow` used.

feature-libs/user/account/components/login-form/login-form-component.service.ts
```diff
@@ -22,14 +26,28 @@ import { tap, withLatestFrom } from 'rxjs/operators';
...
+  }
 
   protected busy$ = new BehaviorSubject(false);
 
+  action?: string;
+  method?: string;
+  csrf$ = this.auth.getCsrfToken();
```
pawelfras, line 33: `nitpick:` consider moving these fields before constructor.

feature-libs/user/account/components/login-form/login-form-component.service.ts
```diff
@@ -40,35 +58,60 @@ export class LoginFormComponentService {
...
+        password: new FormControl('', Validators.required),
+        csrf: new FormControl('', Validators.required),
+      })
+    : new UntypedFormGroup({
+        userId: new UntypedFormControl('', [
+          Validators.required,
+          CustomFormValidators.emailValidator,
+        ]),
+        password: new UntypedFormControl('', Validators.required),
+      });
```
pawelfras: `medium:` consider moving this logic to a separate method. You can also use `form.addControl` instead of creating a whole different `UntypedFormGroup` object.

feature-libs/user/account/components/login-form/login-form-component.service.ts
```diff
@@ -40,35 +58,60 @@ export class LoginFormComponentService {
...
-      this.form.markAllAsTouched();
-      return;
-    }
+  login(nativeForm?: HTMLFormElement) {
+    if (this.isNewAuthFlow && nativeForm) {
+      const csrf =
+        [...nativeForm.elements]
+          .find((element) => element.hasAttribute?.('data-csrf'))
+          ?.getAttribute?.('data-csrf') ?? '';
+      this.form.get('csrf')?.setValue(csrf);
```
pawelfras: `query`: what's the purpose of this logic? If I'm not mistaken, the csrf is already part of the form and should have access to the value. Apart from that, the nativeForm do the job here, so I'm missing the reason of explicitly setting the value to the form control.

feature-libs/user/account/components/login-form/login-form.component.ts
```diff
@@ -17,16 +23,21 @@ import { LoginFormComponentService } from './login-form-component.service';
...
+
   constructor(protected service: LoginFormComponentService) {
     useFeatureStyles('a11yPasswordVisibliltyBtnValueOverflow');
   }
 
   form: UntypedFormGroup = this.service.form;
   isUpdating$: Observable<boolean> = this.service.isUpdating$;
+  csrf$ = this.service.csrf$;
+  action = this.service.action;
+  method = this.service.method;
```
pawelfras, line 32: `medium:` let's consider wrapping this with the feature flag, for better safety.

feature-libs/user/account/components/login-form/login-form.component.ts
```diff
@@ -17,16 +23,21 @@ import { LoginFormComponentService } from './login-form-component.service';
...
 
   @HostBinding('class.user-form') style = true;
 
   onSubmit(): void {
-    this.service.login();
+    this.service.login(this.loginForm?.nativeElement);
```
pawelfras, line 39: `medium` let's consider wrapping this with the feature flag so user with the feature flag disabled always call `login` method without any param

feature-libs/user/account/components/login-form/login-form.module.ts
```diff
@@ -58,6 +58,17 @@ import { LoginFormComponent } from './login-form.component';
...
+          component: LoginFormComponent,
+          guards: [NotAuthGuard],
+          providers: [
+            {
+              provide: LoginFormComponentService,
+              useClass: LoginFormComponentService,
+              deps: [AuthService, GlobalMessageService, WindowRef],
+            },
+          ],
+        },
```
pawelfras: `query:` could you help me understand why we are adding the separate configuration, when we agreed on reusing existing CMS component `ReturningCustomerLoginComponent`? I might be missing something obvious.

projects/core/src/auth/user-auth/config/default-auth-config.ts
```diff
@@ -24,6 +24,10 @@ export const defaultAuthConfig: AuthConfig = {
       clearHashAfterLogin: false,
       responseType: 'code',
     },
+    customLoginPage: {
+      csrfEndpoint: '/csrf',
+      loginFormEndpoint: '/login',
```
pawelfras, line 29: `query:` could you help me understand what's the exact purpose of this URL? I'm not sure if I correctly understand the purpose based on property's name.

projects/core/src/auth/user-auth/config/auth-config.ts
```diff
@@ -90,6 +90,11 @@ export abstract class AuthConfig {
      * Config for angular-oauth-oidc library.
      */
     OAuthLibConfig?: AuthLibConfig;
+
+    customLoginPage?: {
+      csrfEndpoint?: string;
+      loginFormEndpoint?: string;
+    };
```
pawelfras, line 97: `medium:` Let's add description for these config params, for customers and for us from the future

projects/core/src/product/store/effects/product-search.effect.ts
pawelfras, line 1: `query:` is this change related to the topic? If not, let's revert it.

projects/storefrontapp-e2e-cypress/cypress/reports/html/.jsons/mochawesome.json
pawelfras, line 1: `query:` is this change related to the topic? If not, let's revert it.

projects/storefrontapp-e2e-cypress/cypress/support/cx-config.commands.ts
pawelfras, line 1: `query:` is this change related to the topic? If not, let's revert it.

projects/storefrontlib/cms-structure/routing/default-routing-config.ts
```diff
@@ -8,18 +8,20 @@ import { FeatureToggles, RoutingConfig } from '@spartacus/core';
 import { inject } from '@angular/core';
 
 export const defaultRoutesConfigFactory: () => RoutingConfig = () => {
+  const featureToggles = inject(FeatureToggles);
   const enableProductPageRouteAllowsNoProductName =
-    inject(FeatureToggles).defaultProductPageRouteAllowsNoProductName;
+    featureToggles.defaultProductPageRouteAllowsNoProductName;
+  const isNewAuthFlow = featureToggles.authorizationCodeFlowByDefault;
```
pawelfras: `medium:` The `isNewAuthFlow` does not fully reflect what the `authorizationCodeFlowByDefault` flag is responsible for. In addition, the name is not quite true, as Code Flow has been available in Spartacus for a long time. Let's consider renaming it or simply using `featureToggles.authorizationCodeFlowByDefault` directly.

```suggestion
  const authorizationCodeFlowByDefault = featureToggles.authorizationCodeFlowByDefault;
```

projects/storefrontlib/cms-structure/routing/default-routing-config.ts
```diff
@@ -8,18 +8,20 @@ import { FeatureToggles, RoutingConfig } from '@spartacus/core';
...
+  const routingConfig = {
     routing: {
       routes: {
         home: { paths: [''] },
         notFound: { paths: ['not-found'] },
 
         // semantic links for login related pages
         login: {
-          paths: ['login'],
+          paths: [isNewAuthFlow ? 'signIn' : 'login'],
```
pawelfras: `major:` for the consistency, let;s change the route name to lowercase
```suggestion
          paths: [isNewAuthFlow ? 'signin' : 'login'],
```

Apart from that, let's add a comment with description why this change is required. It is not obvious and might be difficult for other devs to understand.

projects/storefrontlib/cms-structure/routing/default-routing-config.ts
```diff
@@ -81,4 +83,14 @@ export const defaultRoutesConfigFactory: () => RoutingConfig = () => {
...
     },
   };
+
+  if (isNewAuthFlow) {
+    routingConfig.routing.routes['loginForm'] = {
+      paths: ['login'],
+      protected: false,
+      authFlow: true,
+    };
+  }
```
pawelfras: `query:` could you help me understand what's the purpose of this? IS it related to the offline discussion we had yesterday?

feature-libs/user/account/components/login-form/login-form.module.ts
Pucek9: I forgot to remove it, sorry

projects/core/src/auth/user-auth/config/auth-config.ts
Pucek9: Please see https://github.com/SAP/spartacus/pull/20551#discussion_r2231539008

feature-libs/user/account/components/login-form/login-form-component.service.ts
Pucek9: only those fields? what about others? 
If we want to keep order like:

injects
properties
constructor
methods

i need to reorder whole class

feature-libs/user/account/components/login-form/login-form.component.ts
Pucek9: didn't want to introduce feature flag in component. The same logic is covered in the service. If service returns nothing, there will be undefined as expected and not added attribute to html

projects/storefrontlib/cms-structure/routing/default-routing-config.ts
Pucek9: legacy: `login` route with path `/login`
new: `login` route with path `/signin` and `loginForm` with path `/login`

in new flow we need 2 routes/paths, but one cms page is enough (`/login`) so needn't update sampledata at all

projects/storefrontlib/cms-structure/routing/default-routing-config.ts
Pucek9: are you sure `signin`, not `sign-in`?

projects/storefrontapp-e2e-cypress/cypress/reports/html/.jsons/mochawesome.json
Pucek9: it's prettier stuff my boy

projects/core/src/auth/user-auth/config/default-auth-config.ts
Pucek9: I was wondering why we dont have simple configuration for auth endpoints like in occ:

```ts
export const defaultOccSomeConfig: OccConfig = {
  backend: {
    occ: {
      endpoints: {
        asmCustomerSearch: '/assistedservicewebservices/customers/search',
        asmCustomerLists: '/assistedservicewebservices/customerlists',
        asmBindCart: '/assistedservicewebservices/bind-cart',
        asmCreateCustomer: '/assistedservicewebservices/customers',
        //...
      },
    },
    auth: {
      endpoints: {
        token: '/authorizationserver/oauth/token',
        csrf: '/authorizationserver/csrf'
        //...
      }
    }
  },
};
```
But that's how it was from the beginning. Apparently, there was no need to unify it.

These are paths to the oauth server, which are builder from blocks in auth config service (just re-used existing approach)
`csrfEndpoint` is for receiving expected field form name and token (see type CSRFResponse)
`loginFormEndpoint` is an for submitting target (action)

feature-libs/user/account/components/login-form/login-form.component.ts
Pucek9: https://github.com/SAP/spartacus/pull/20565#discussion_r2245053128

But will check it deeper, as unit tests (eg cdc login form) started failing

feature-libs/user/account/components/login-form/login-form-component.service.ts
Pucek9: I was not supposed to set value of that field direclty in html template (as it was done for `data-csrf`). Just value is not possible to be set this way due to some condtitions (field is type hidden, style hidden etc). 
The second thing is an Angular validation, which we also check below.

Looks ugly, but it's a kind of workaround -  I didn't have a better idea.

projects/storefrontlib/cms-structure/routing/default-routing-config.ts
pawelfras: second thoughts - as we discussed, it indeed might be helpful, e.g. for redirection directly to the `/login` page.

Pucek9: I don't expect anyone to want to go there directly. It's against the flow (and in the best case scenario - if the error handling works - it will be redirected to `signin` and the flow will start).

feature-libs/user/account/components/login-form/login-form-component.service.ts
Pucek9: Yeah, but according to the guideline it is adding new forms as typed `FormGroup` now. Should I add `FormControl` into `UntypedFormGroup` or `UntypedFormControl`?

feature-libs/user/account/root/user-account-root.module.ts
```diff
@@ -25,6 +25,7 @@ export function defaultUserAccountComponentsConfig(): CmsConfig {
           'MyAccountViewUserComponent',
           'ReturningCustomerOTPLoginComponent',
           'RegisterCustomerWithOTPComponent',
+          'CustomLoginComponent',
```
pawelfras: `major:` let's remove this as we don't need additional CMS component anymore.

projects/core/src/auth/client-auth/services/cross-site-request-forgery.service.ts
```diff
@@ -0,0 +1,36 @@
...
+  protected http = inject(HttpClient);
+  protected authConfigService = inject(AuthConfigService);
+
+  /**
+   * Returns CSRF Token
+   */
+  getCsrfToken() {
+    const url = this.authConfigService.getCsrfEndpoint();
+    return this.http.get<CSRFResponse>(url, {
+      withCredentials: true,
```
pawelfras, line 33: `major:` We have an interceptor [WithCredentialsInterceptor](https://github.com/SAP/spartacus/blob/8878b17463bb69b2da5bf9ab88e0d23f963bb88a/projects/core/src/occ/interceptors/with-credentials.interceptor.ts#L21) that takes care of adding this flag to the outgoing requests. Let's remove the flag if redundant.

feature-libs/user/account/components/login-form/login-form-component.service.ts
pawelfras: What about moving this logic to the constructor?
```ts
if (this.featureConfigService.isEnabled('authorizationCodeFlowByDefault')) {
      this.method = 'POST';
      this.action = this.authConfigService?.getCustomLoginFormEndpoin
      this.form.addControl('csrf', new FormControl('', Validators.required));
      this.csrf$.pipe(takeUntilDestroyed()).subscribe((csrf) => {
        this.form.get('csrf')?.setValue(csrf.token);
      });
    }
```

Then you won't need the `attr.data-csrf` attribute and reading the native element to find its value.

feature-libs/user/account/components/login-form/login-form.component.ts
```diff
@@ -17,16 +23,19 @@ import { LoginFormComponentService } from './login-form-component.service';
   standalone: false,
 })
 export class LoginFormComponent {
-  constructor(protected service: LoginFormComponentService) {
-    useFeatureStyles('a11yPasswordVisibliltyBtnValueOverflow');
-  }
-
+  @ViewChild('loginForm') loginForm: ElementRef<HTMLElementTagNameMap['form']>;
+  @HostBinding('class.user-form') style = true;
```
pawelfras, line 27: `query` what's the reason of this line?

projects/core/src/auth/client-auth/services/cross-site-request-forgery.service.ts
Pucek9: Nice catch! But looks like interceptor doesnt handle this particular request.
```ts
  protected requiresWithCredentials(request: HttpRequest<any>): boolean {
    return (
      this.occConfig?.useWithCredentials !== undefined &&
      request.url.indexOf(this.occConfig?.prefix ?? '') > -1
    );
  }
  ```
  seems to be created for occ requests only

File: feature-libs/user/account/components/login-form/login-form.component.ts (hunk 16 above)
Comment (Pucek9) [reply feature-libs/user/account/components/login-form/login-form.component.ts:27]: jsut changed order of properties in that file

File: feature-libs/user/account/components/login-form/login-form-component.service.ts (hunk 5 above)
Comment (Pucek9) [reply feature-libs/user/account/components/login-form/login-form-component.service.ts]: I wanted to avoid 2 subscriptions for the same item, but let me try.
Ah now I remember - there is no native form available in the constructor (to set `name` property), so need to add some `onInit` / `afterViewInit` cycle to handle it

Btw we dont need `takeUntilDestroyed`, `take(1)` or even `unsubscribe` here, as it's a cold observable

File: feature-libs/user/account/components/login-form/login-form.component.html [hunk 17]
```diff
@@ -58,6 +66,14 @@
     ></cx-form-errors>
   </label>
 
+  <input
+    *ngIf="!!form.get('csrf') && (csrf$ | async) as csrf"
+    formControlName="csrf"
+    type="hidden"
+    [name]="csrf.parameterName"
+    [attr.data-csrf]="csrf.token"
+  />
```
pawelfras: Let's use the functionality to handle template parts that depend on feature toggle:
 ```suggestion
   <ng-container *cxFeature="'authorizationCodeFlowByDefault'">
    <input
      *ngIf="csrf$ | async as csrf"
      formControlName="csrf"
      type="hidden"
      [name]="csrf.parameterName"
      [attr.data-csrf]="csrf.token"
    />
  </ng-container>
```

feature-libs/user/account/components/login-form/login-form-component.service.spec.ts
pawelfras, line 1: FOR THE FUTURE REFERENCE:
Let's cover new functionality with unit tests in a separate PR.

pawelfras (approved):
FOR THE FUTURE REFERENCE:
Let's cover new functionality with unit tests in a separate PR.

projects/core/src/auth/user-auth/config/default-auth-config.spec.ts
pawelfras, line 1: FOR THE FUTURE REFERENCE:
As discussed, Let's fix the unit tests scenario in a separate PR.

PR #20572, opened 2025-08-01
fix: CXSPA-10618 Login during Checkout fails to resume checkout flow (cart related)

pawelfras:
This PR provides a fix for the merge cart scenario for the code flow.

QA:

The prerequisites for each test scenario are:
- created user
- starting as an anonynous user with empty cart

1. **Code Flow with the Custom Login Flow:**
  - run app and go to the http://localhost:4200/electronics-spa/en/USD/product/1934794/powershot-a480
  - click `Add to Cart` and the `Proceed to Checkout`
  - after redirection to the custom login page, enter the credentials for your user and log in
  - verify that after successful login, app is redirected to the checkout page
  - go through the checkout and verify that the order has the item added as an anonymous user (successful cart merge)

 2. **Code flow with external auth login page:**
  - in `spartacus-features.module.ts`, set `authorizationCodeFlowByDefault` to false (no custom login page)
  - in `spartacus-features.module.ts`, add the following changes
    - in `defaultAuthConfig`, change `client_id` to `customloginpage_public` (to not change config od client used on the epic branch)
    - change the following condition to return the `defaultAuthConfig`
       ```ts
       if (true) {
            return defaultAuthConfig;
       }
       ```
  - go to the backoffice of the S1 server and make sure there is no custom login page URI (so we are not redirected to the custom login page)
     image
  - make sure the app restarted after provided changes
  - run app and go to the http://localhost:4200/electronics-spa/en/USD/product/1934794/powershot-a480
  - click `Add to Cart` and the `Proceed to Checkout`
  - after redirection to the external login page, enter the credentials for your user and log in
  - verify that after successful login, app is redirected to the checkout page
  - go through the checkout and verify that the order has the item added as an anonymous user (successful cart merge)
  
  3. **Password flow with the default login page:**
  - in `spartacus-features.module.ts`, set `authorizationCodeFlowByDefault` and `incrementProcessesCountForMergeCart`to `false
  - revert previously added changes from the `spartacus-features.module.ts`.
  - in `environment.ts`, change the `occBaseUrl` to the standard dev env:
    ```
     // occBaseUrl: buildProcess.env.CX_BASE_URL,
    occBaseUrl: 'https://40.76.109.9:9002',
    ```
  - make sure the app restarted after provided changes
  - run app and go to the http://localhost:4200/electronics-spa/en/USD/product/1934794/powershot-a480
  - click `Add to Cart` and the `Proceed to Checkout`
  - after redirection to the default login page, enter the credentials for your user and log in
  - verify that after successful login, the app is redirected to the checkout page
  - go through the checkout and verify that the order has the item added as an anonymous user (successful cart merge)
 

closes [CXSPA-10618](https://jira.tools.sap/browse/CXSPA-10618)


feature-libs/cart/base/core/facade/multi-cart.service.spec.ts
```diff
@@ -324,24 +330,131 @@ describe('MultiCartService', () => {
...
-        new CartActions.MergeCart({
+    });
+    describe('feature flag authorizationCodeFlowByDefault is enabled', () => {
+      it('should merge cart', () => {
+        spyOn(service as any, 'generateTempCartId').and.returnValue(
+          'temp-uuid'
+        );
+        const isEnabledSpy = featureConfigService.isEnabled as jasmine.Spy;
+
+        // authorizationCodeFlowByDefault is enabled
```
Pucek9: we have that information in describe block

what are purposes of remaining commented code below?

feature-libs/cart/base/core/facade/multi-cart.service.ts
```diff
@@ -155,14 +157,30 @@ export class MultiCartService implements MultiCartFacade {
...
-        userId,
-        cartId,
-        extraData,
-        tempCartId,
-      })
-    );
+    if (
+      this.featureConfigService.isEnabled(
+        'incrementProcessesCountForMergeCart'
+      ) ||
```
Pucek9, line 163: is not enough to check only `incrementProcessesCountForMergeCart` flag?

projects/storefrontapp/src/app/spartacus/spartacus-features.module.ts
```diff
@@ -408,6 +408,7 @@ if (environment.cpq) {
         unifiedDefaultHeaderSlotsAcrossBreakpoints: true,
         reserveSpaceForImagesOnPdpAndPlp: true,
         lazyLoadImagesByDefault: true,
+        incrementProcessesCountForMergeCart: true,
```
Pucek9, line 411: is it not JDK21 related? Maybe it should be below the comment

Pucek9 (approved):
QA: each cases work as expected

feature-libs/cart/base/core/facade/multi-cart.service.spec.ts
pawelfras: Good point! These are leftovers used before I decidec to move tests to separate blocks.

feature-libs/cart/base/core/facade/multi-cart.service.ts
pawelfras: The `incrementProcessesCountForMergeCart` was introduced to fix the issue for custoemrs that use code flow and don't want to switch to the latest JDK 21 commerce system. Flag `authorizationCodeFlowByDefault` provides much more than only this fix and it won't work JDK 17.

projects/storefrontapp/src/app/spartacus/spartacus-features.module.ts
pawelfras: According to the above comment, it is not related to JDK 21 and it is not necessary to keep it below the comment.

PR #20573, opened 2025-08-01
fix: CXSPA-10632: Auth issue when using CDC library along with JDK21

anjana-bl:
When CDC is enabled in storefront, always show CMS Login Page and not the Authorization server page

### DIFF
--- a/integration-libs/cdc/root/cdc-root.module.ts
+++ b/integration-libs/cdc/root/cdc-root.module.ts
@@ -12,7 +12,7 @@ import {
   provideDefaultConfig,
   provideDefaultConfigFactory,
 } from '@spartacus/core';
-import { LogoutGuard } from '@spartacus/storefront';
+import { LoginGuard, LogoutGuard } from '@spartacus/storefront';
 import { lastValueFrom } from 'rxjs';
 import { tap } from 'rxjs/operators';
 import { CdcConsentManagementModule } from './consent-management/cdc-consent.module';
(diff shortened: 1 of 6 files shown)


vigneshkrish8691 (dismissed):

projects/storefrontlib/cms-components/user/login-route/login.guard.ts
```diff
@@ -35,14 +35,10 @@ export class LoginGuard {
...
-    return this.authService.isUserLoggedIn().pipe(
+    return this.shouldRenderCMSPage().pipe(
       take(1),
-      switchMap((isUserLoggedIn) => {
-        if (
-          this.authConfigService.getOAuthFlow() ===
-            OAuthFlow.ResourceOwnerPasswordFlow ||
-          isUserLoggedIn
-        ) {
+      switchMap((response) => {
```
pawelfras: `minor`:

```suggestion
      switchMap((shouldRenderCMSPage) => {
```

integration-libs/cdc/root/cdc-root.module.ts
```diff
@@ -53,6 +54,7 @@ export function defaultCdcComponentsConfig(): CmsConfig {
   providers: [
     provideDefaultConfigFactory(defaultCdcComponentsConfig),
     { provide: LogoutGuard, useExisting: CdcLogoutGuard },
+    { provide: LoginGuard, useExisting: CdcLoginGuard },
```
pawelfras, line 57: `query:` Would the new implementation work as expected, with the password flow, which used to be the default config for Spartacus apps? If it can affect current users, it should be hidden under the feature flag (see [authorizationCodeFlowByDefault](https://github.com/SAP/spartacus/blob/f576828af1574f8c0a52650868645270ed715c5e/projects/core/src/features-config/feature-toggles/config/feature-toggles.ts#L783C39-L784C1) available on `epic/jdk21auth`)

projects/storefrontlib/cms-components/user/login-route/login.guard.ts
anjana-bl: done

anjana-bl:
closing this PR as changes are followed in https://github.com/SAP/spartacus/pull/20581

PR #20574, opened 2025-08-02
fix: conflcits on epic/jdk21auth

Pucek9:


projects/core/src/auth/user-auth/config/default-auth-config.ts
```diff
@@ -4,40 +4,10 @@
...
- */
-export function provideAuthorizationCodeFlowByDefault(
-  enable = true
-): ValueProvider {
-  return {
-    provide: USE_AUTHORIZATION_CODE_FLOW_BY_DEFAULT,
-    useValue: enable,
-  };
-}
-
```
pawelfras: `major:` Since the injection token and `provideAuthorizationCodeFlowByDefault` are part of the public API, and customers may already use them, we cannot simply remove the logic. Let's mark the mentioned API as deprecated (since 2211.44?), and make its logic dependent on `authorizationCodeFlowByDefault` flag so customers cannot control the same configuration different way.

projects/core/src/auth/user-auth/index.ts
```diff
@@ -18,4 +18,3 @@ export * from './services/index';
 export * from './store/actions';
 export { UserAuthModule } from './user-auth.module';
 export * from './utils/index';
-export { USE_AUTHORIZATION_CODE_FLOW_BY_DEFAULT } from './config/default-auth-config';
```
pawelfras: `major:` according to the previous comment, let's revert this change

feature-libs/asm/components/asm-main-ui/asm-main-ui.component.ts
```diff
@@ -101,6 +100,7 @@ export class AsmMainUiComponent implements OnInit, OnDestroy {
   isAsmCustomer360Configured: boolean | undefined = false;
   isAsmCustomer360Loaded$ = new BehaviorSubject<boolean>(false);
   protected featureModules = inject(FeatureModulesService);
+  protected featureToggles = inject(FeatureToggles);
```
pawelfras: `major:` `FeatureToggles`  is supposed to be used only in factories that provides a `Config` based on the feature flag, to avoid circular dependency. In other places, e.g components, let's use `FeatureConfigService` instead.

Apart from that, let's change the access modifier to `private` so customers cannot build their logic on top of the interim service.

```suggestion
  private featureConfigService = inject(FeatureConfigService);
```

feature-libs/asm/components/asm-main-ui/asm-main-ui.component.ts
```diff
@@ -78,7 +77,7 @@ export class AsmMainUiComponent implements OnInit, OnDestroy {
   isCollapsed$: Observable<boolean> | undefined;
   iconTypes = ICON_TYPE;
   forbiddenResponseStatus = HttpResponseStatus.FORBIDDEN;
-  isShowOauth2AsmloginPage: boolean;
+  isShowOauth2AsmloginPage?: boolean;
```
pawelfras: `major:` let's change the access modifier to `private` or even completely remove the line, for the same reasons as mentioned in the previous comments.

Pucek9: its used in html template, so guess that I should use directive, right?

projects/core/src/auth/user-auth/config/default-auth-config.ts
```diff
@@ -4,10 +4,48 @@
...
  */
 
-import { inject } from '@angular/core';
+import { inject, InjectionToken, ValueProvider } from '@angular/core';
 import { FeatureToggles } from '../../../features-config/feature-toggles';
 import { AuthConfig } from './auth-config';
 
+/**
+ * @deprecated since 2211.44 Please use feature toggle with authorizationCodeFlowByDefault flag
+ */
```
pawelfras:
```suggestion
@deprecated since 2211.44. Please use the `authorizationCodeFlowByDefault` feature toggle instead.
```

projects/core/src/auth/user-auth/config/default-auth-config.ts
```diff
@@ -4,10 +4,48 @@
...
+export const USE_AUTHORIZATION_CODE_FLOW_BY_DEFAULT =
+  new InjectionToken<boolean>('USE_AUTHORIZATION_CODE_FLOW_BY_DEFAULT', {
+    factory: () => false,
+    providedIn: 'root',
+  });
+
+/**
+ * @deprecated since 2211.44 Please use feature toggle with authorizationCodeFlowByDefault flag
```
pawelfras: Apologies, after second thoughts on this, we should not deprecate the factory. Since we updated the logic under the hood to use the feature toggle, the factory remains valid and important.

projects/core/src/auth/user-auth/config/default-auth-config.ts
```diff
@@ -4,10 +4,48 @@
...
+ *
+ * When enabled, sets the default oAuth configuration to use authorization
+ * code flow with PKCE. This results in a more secure authorization scheme
+ * as the default configuration.
```
pawelfras:
```suggestion
* When `authorizationCodeFlowByDefault` feature toggle is enabled, it sets the default oAuth configuration to use authorization.
 * code flow with PKCE. This results in a more secure authorization scheme
 * as the default configuration.
```

pawelfras (approved):

PR #20578, opened 2025-08-06
fix: Customer emulation will be terminated abnormally

Melody-zhou-512:
JIRA ticket:https://jira.tools.sap/browse/CXSPA-10734

### DIFF
--- a/projects/core/src/auth/user-auth/facade/auth.service.spec.ts
+++ b/projects/core/src/auth/user-auth/facade/auth.service.spec.ts
@@ -21,6 +21,9 @@ class MockUserIdService implements Partial<UserIdService> {
   }
   clearUserId() {}
   setUserId() {}
+  isEmulated(): Observable<boolean> {
+    return of();
+  }
 }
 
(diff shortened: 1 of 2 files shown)


niehuayang (dismissed):

projects/core/src/auth/user-auth/facade/auth.service.spec.ts
```diff
@@ -122,6 +125,7 @@ describe('AuthService', () => {
       spyOn(userIdService, 'setUserId').and.callThrough();
       spyOn(store, 'dispatch').and.callThrough();
       spyOn(authStorageService, 'getItem').and.returnValue('token');
+      spyOn(userIdService, 'isEmulated').and.returnValue(of(false));
```
niehuayang, line 128: Can you add a test for isEmulated = true

Melody-zhou-512: have been modified.

niehuayang (approved):

pawelfras (approved):

PR #20579, opened 2025-08-06
fix: login form-service allow overrides

Pucek9:
Clsoes [CXSPA-10737](https://jira.tools.sap/browse/CXSPA-10737)

### DIFF
--- a/feature-libs/user/account/components/login-form/login-form-component.service.ts
+++ b/feature-libs/user/account/components/login-form/login-form-component.service.ts
@@ -53,16 +53,6 @@ export class LoginFormComponentService {
     password: new UntypedFormControl('', Validators.required),
   });
 
-  protected initCustomLogin() {
-    this.method = 'POST';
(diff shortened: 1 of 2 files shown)


feature-libs/user/account/components/login-form/login-form-component.service.ts
```diff
@@ -52,18 +53,23 @@ export class LoginFormComponentService {
...
 
+  protected initCustomLogin() {
+    this.method = 'POST';
+    this.action = this.authConfigService?.getCustomLoginFormEndpoint();
+    this.form.addControl('csrf', new FormControl('', Validators.required));
+    this.csrf$ = this.auth.getCsrfToken();
+    this.csrf$.pipe(take(1)).subscribe((csrf) => {
+      this.form.get('csrf')?.setValue(csrf.token);
+    });
+  }
```
pawelfras: `minor:` let's move the protected method ad the end of class

pawelfras (approved):

PR #20580, opened 2025-08-06
fix: Redundant Login action is dispatched on page refresh

sdrozdsap:
# Fix: Prevent Login action dispatch on page refresh in OAuth code flow

## Problem

When using OAuth authorization code flow (required for JDK 21), the `AuthActions.Login()` action was being dispatched on every page refresh when a user was already logged in. This caused multiple issues:

- **Redundant API requests**: Effects listening to the Login action were triggered unnecessarily (e.g., `/pages` endpoint calls)
- **Tracking events fired incorrectly**: Login tracking events were sent on page refresh instead of actual login
- **Performance impact**: Multiple effects were re-executing business logic meant only for fresh logins
- **Semantic inconsistency**: The Login action should indicate a login event, not that a user is still logged in

## Root Cause

In `AuthService.checkOAuthParamsInUrl()`, the Login action was dispatched whenever both conditions were met:
1. `loginResult.result` was `true` 
2. An access token existed in storage

However, in code flow, `loginResult.result` returns `true` even on page refreshes when the user is already authenticated, leading to the Login action being dispatched incorrectly.

## Solution

The fix leverages the existing `tokenReceived` flag in `OAuthTryLoginResult` which accurately distinguishes between:
- **`tokenReceived: true`** - Fresh token received from auth server (actual login redirect)
- **`tokenReceived: false`** - Token retrieved from storage (page refresh)

This fix resolves the issue described in the team discussion where Login actions were being dispatched on every page refresh in the JDK 21 auth branch, causing unwanted side effects in tracking and other business logic.

### DIFF
--- a/projects/core/src/auth/user-auth/facade/auth.service.spec.ts
+++ b/projects/core/src/auth/user-auth/facade/auth.service.spec.ts
@@ -14,7 +14,10 @@ import { AuthActions } from '../store/actions';
 import { AuthService } from './auth.service';
 import { UserIdService } from './user-id.service';
 import createSpy = jasmine.createSpy;
-import { CrossSiteRequestForgeryService } from '@spartacus/core';
+import {
+  CrossSiteRequestForgeryService,
+  FeatureConfigService,
+} from '@spartacus/core';
 
 class MockUserIdService implements Partial<UserIdService> {
   getUserId(): Observable<string> {
@@ -80,6 +83,10 @@ class MockAuthMultisiteIsolationService {
   }
 }
 
+class MockFeatureConfigService implements Partial<FeatureConfigService> {
+  isEnabled = createSpy().and.returnValue(false);
+}
+
 describe('AuthService', () => {
   let service: AuthService;
   let routingService: RoutingService;
@@ -88,6 +95,7 @@ describe('AuthService', () => {
   let oAuthLibWrapperService: OAuthLibWrapperService;
   let authRedirectService: AuthRedirectService;
   let authMultisiteIsolationService: AuthMultisiteIsolationService;
+  let featureConfigService: FeatureConfigService;
   let store: Store;
 
   beforeEach(() => {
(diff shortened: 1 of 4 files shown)


projects/core/src/auth/user-auth/facade/auth.service.ts
```diff
@@ -61,12 +61,13 @@ export class AuthService {
...
       if (loginResult.result && token) {
         this.userIdService.setUserId(OCC_USER_ID_CURRENT);
-        this.store.dispatch(new AuthActions.Login());
 
         // We check if the token was received during the `tryLogin()` attempt.
         // If so, we will redirect as we can deduce we are returning from the authentication server.
         // Redirection should not be done in cases we get the token from storage (eg. refreshing the page).
+        // In this case we need to dispatch the login action to indicate that the user has just logged in.
         if (loginResult.tokenReceived) {
+          this.store.dispatch(new AuthActions.Login());
```
pawelfras: `major:` I'm wondering if we can consider this change as a breaking, because we're dispatching in the different condition. Even if this is a fix, it might be safer to hide it under the feature flag, to avoid unexpected issues in client's application.

sdrozdsap: Yeah I think it might be safer, considering that we don't know whether someone has business logic dependent on previous behaviour. Ill adjust the changes

projects/core/src/auth/user-auth/facade/auth.service.spec.ts
```diff
@@ -134,39 +147,121 @@ describe('AuthService', () => {
   });
 
   describe('checkOAuthParamsInUrl()', () => {
-    it('should login user when token is present', async () => {
-      spyOn(oAuthLibWrapperService, 'tryLogin').and.callThrough();
-      spyOn(userIdService, 'setUserId').and.callThrough();
-      spyOn(store, 'dispatch').and.callThrough();
-      spyOn(authStorageService, 'getItem').and.returnValue('token');
+    describe('when dispatchLoginActionOnlyWhenTokenReceived feature flag is DISABLED', () => {
```
sdrozdsap, line 150: I'm creating a duplicated version of practically same test cases on purpose, for easier cleanup once feature flag will  be removed

pawelfras (approved):
Review and QA done. LGTM

PR #20581, opened 2025-08-06
fix: CXSPA-10632: Auth issue when using CDC library along with JDK21

anjana-bl:
When CDC is enabled in storefront, always show CMS Login Page and not the Authorization server page

### DIFF
--- a/integration-libs/cdc/root/config/default-cdc-routing-config.ts
+++ b/integration-libs/cdc/root/config/default-cdc-routing-config.ts
@@ -28,7 +28,7 @@ export const cdcRoutesConfig: RoutesConfig = {
     authFlow: true,
   },
   loginForm: {
-    paths: ['login-form'], //overriding to avoid same path in different routes
+    paths: [], //overriding to avoid same path in different routes
     protected: false,
     authFlow: true,
   },
--- a/integration-libs/cdc/root/config/default-cdc-routing-config.ts
+++ b/integration-libs/cdc/root/config/default-cdc-routing-config.ts
@@ -22,6 +22,16 @@ export const cdcRoutesConfig: RoutesConfig = {
     protected: false,
     authFlow: true,
   },
+  login: {
+    paths: ['login'],
+    protected: false,
+    authFlow: true,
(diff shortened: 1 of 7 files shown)


integration-libs/cdc/root/config/default-cdc-routing-config.ts
```diff
@@ -22,6 +22,16 @@ export const cdcRoutesConfig: RoutesConfig = {
     protected: false,
     authFlow: true,
   },
+  login: {
+    paths: ['login'],
+    protected: false,
+    authFlow: true,
+  },
+  loginForm: {
+    paths: ['login-form'], //overriding to avoid same path in different routes
```
Pucek9: fine with override `loginForm`, but I think if you remove paths from here, as it's unnecessary
```suggestion
```

Pucek9: or 
```suggestion
    paths: [], //overriding to avoid same path in different routes
```
If issue will still raise

integration-libs/cdc/user-account/login-form/cdc-login-form-component.service.ts
```diff
@@ -65,6 +65,10 @@ export class CdcLoginFormComponentService
     );
   }
 
+  protected initCustomLogin() {
+    // Intentional empty method , so csrf token is not initialized in case of CDC
```
Pucek9, line 69:
```suggestion
    // Intentional empty method, so csrf token is not initialized in case of CDC
```

Pucek9 (approved):

integration-libs/cdc/user-account/login-form/cdc-login-form-component.service.ts
anjana-bl: done

integration-libs/cdc/root/config/default-cdc-routing-config.ts
anjana-bl: done

PR #20582, opened 2025-08-06
fix: added button for returning physical products in S4ServiceOrderDetailActionsComponent

suprishi:
Return button is not visible for physical products when s4-service library is installed because `S4ServiceOrderDetailActionsComponent` overrides `MyAccountV2OrderDetailsActionsComponent`.

Added return button in `S4ServiceOrderDetailActionsComponent` as a fix.


vigneshkrish8691 (dismissed):

.env-cmdrc
anjana-bl, line 1: MAJOR: remove unnecessary changes..

integration-libs/s4-service/order/components/s4-service-order-detail-actions/s4-service-order-detail-actions.component.ts
```diff
@@ -4,33 +4,39 @@
...
+export class S4ServiceOrderDetailActionsComponent
+  extends OrderDetailActionsComponent
+  implements OnInit
+{
   protected checkoutServiceSchedulePickerService = inject(
     CheckoutServiceSchedulePickerService
   );
   protected globalMessageService = inject(GlobalMessageService);
 
-  displayActions$: Observable<boolean> = this.order$.pipe(
```
anjana-bl, line 26: MAJOR: shouldnt this be considered as a breaking change - removing a property from a publish component? Instead of removing it,  can we mark this deprecated

anjana-bl (changes requested):

integration-libs/s4-service/order/components/s4-service-order-detail-actions/s4-service-order-detail-actions.component.ts
```diff
@@ -4,33 +4,39 @@
...
-  );
+  ngOnInit(): void {
+    this.order$.pipe(tap((order) => this.checkServiceStatus(order)));
+  }
 
-  protected checkServiceStatus(order: Order): boolean {
-    if (order && order.status === 'CANCELLED') {
-      return false;
-    } else if (order && order.servicedAt) {
+  protected checkServiceStatus(order: Order): void {
```
anjana-bl: MAJOR: won't this result in compile-time breaking change for anyone extending the class? can this also be marked deprecated and rewritten instead?

anjana-bl (changes requested):

integration-libs/s4-service/order/components/s4-service-order-detail-actions/s4-service-order-detail-actions.component.html
```diff
@@ -1,10 +1,45 @@
...
+              params: order,
+            } | cxUrl
+          "
+          [attr.aria-label]="
+            'myAccountV2OrderDetails.returnItems' | cxTranslate
+          "
+          class="btn btn-secondary"
+          cxBtnLikeLink
+        >
+          {{ 'myAccountV2OrderDetails.returnItems' | cxTranslate }}
```
anjana-bl, line 39: QUERY:  If in the future, a new button is added to the MyAccountV2OrderDetailsActionsComponent, won’t it become a repetitive task for us to track and update it here each time? How will we ensure such changes are consistently captured?

suprishi: We could add a new CMS component dedicated to S4 service actions and leave `MyAccountV2OrderDetailsActionsComponent` untouched, but that would require inputs from UX as well. 

Could you suggest a better way of resolving the bug which would be favourable in the long run? Also, any idea how frequently do you expect `MyAccountV2OrderDetailsActionsComponent` would be updated?

hackergil (dismissed):

integration-libs/s4-service/order/components/s4-service-order-detail-actions/s4-service-order-detail-actions.component.ts
```diff
@@ -4,29 +4,48 @@
...
   displayActions$: Observable<boolean> = this.order$.pipe(
     map((order) => this.checkServiceStatus(order))
   );
 
+  ngOnInit(): void {
+    this.order$.pipe(tap((order) => this.displayServiceMessage(order)));
+  }
+
+  /**
+   * @deprecated since 2211.42.0-3 - Displaying the individual action buttons will depend on their boolean flags from API response.
```
anjana-bl: there is no offical version as 2211.42.0-3 - it will be only 2211.42 - please cofirm this with UA once and do the necessary change

anjana-bl (changes requested):

.env-cmdrc
anjana-bl: changes are not done, please check again

integration-libs/s4-service/order/components/s4-service-order-detail-actions/s4-service-order-detail-actions.component.html
anjana-bl: Either creating a new CMS component to hold these actions or introducing a generic outlet for adding new buttons would be a cleaner, future-proof approach.
It’s less about how frequently MyAccountV2OrderDetailsActionsComponent changes, and more about ensuring our libraries can adapt smoothly if the core component is updated — since such changes are always a possibility.
General practice: For any class (service or component), we always follow the approach of extending from the parent class. However, since templates cannot be extended in the same way as classes, we should avoid overriding a parent template.

anjana-bl (changes requested):

integration-libs/s4-service/order/components/s4-service-order-detail-actions/s4-service-order-detail-actions.component.ts
anjana-bl: is this done?

integration-libs/s4-service/order/components/s4-service-order-detail-actions/s4-service-order-detail-actions.component.ts
anjana-bl: there is no offical version as 2211.42.0-3 - it will be only 2211.42 - please cofirm this with UA once and do the necessary change

anjana-bl:
please create a PR to release/221121.1.x as mentioned in https://sap-cx.slack.com/archives/CAC8PULNN/p1755293416190099 if you are planning it for 2211.44 release

suprishi:
Received unexpected file changes after changing base branch to `release/221121.1.x`. So closing this one and raised a new PR at https://github.com/SAP/spartacus/pull/20652

PR #20585, opened 2025-08-07
fix: redirect to login cx-route after registration

FollowTheFlo:
In this PR, I am assuming this logic:
SAP custom login --> AuthorizationCode OAothFlow + authorizationCodeFlowByDefault feature toggle true
3rd party vendor -->  AuthorizationCode OAothFlow + authorizationCodeFlowByDefault feature toggle false

Thus the changes made in such way to not break 3rd party vendor.


feature-libs/user/profile/components/register/register.component.ts
```diff
@@ -230,13 +231,12 @@ export class RegisterComponent implements OnInit, OnDestroy {
...
   protected onRegisterUserSuccess(): void {
     if (
-      this.authConfigService.getOAuthFlow() ===
-      OAuthFlow.ResourceOwnerPasswordFlow
+      [
+        OAuthFlow.ResourceOwnerPasswordFlow,
+        OAuthFlow.AuthorizationCode,
+      ].includes(this.authConfigService.getOAuthFlow())
     ) {
-      this.router.go('login');
```
FollowTheFlo, line 236: I don't see the benefits of not using cxRoute, I am assuming it was forgotten when this code was added 4y ago.

feature-libs/user/profile/components/registration-verification-token-form/verify-register-verification-token-form.component.ts
```diff
@@ -241,10 +241,12 @@ export class RegistrationVerificationTokenFormComponent implements OnInit {
...
     if (
-      this.authConfigService.getOAuthFlow() ===
-      OAuthFlow.ResourceOwnerPasswordFlow
+      [
+        OAuthFlow.ResourceOwnerPasswordFlow,
+        OAuthFlow.AuthorizationCode,
+      ].includes(this.authConfigService.getOAuthFlow())
     ) {
-      this.router.go('login');
+      this.router.go({ cxRoute: 'login' });
```
Pucek9, line 255: Yeah, but then still should we not treat it as a breaking change and handle it with a feature flag?
In that file we have also another example of usage it `this.router.go(['/login/register']);`

feature-libs/user/profile/components/register/register.component.ts
```diff
@@ -230,13 +230,12 @@ export class RegisterComponent implements OnInit, OnDestroy {
 
   protected onRegisterUserSuccess(): void {
     if (
-      this.authConfigService.getOAuthFlow() ===
-      OAuthFlow.ResourceOwnerPasswordFlow
+      [
+        OAuthFlow.ResourceOwnerPasswordFlow,
+        OAuthFlow.AuthorizationCode,
+      ].includes(this.authConfigService.getOAuthFlow())
```
pawelfras: `major:` this change might be considered breaking for customers that are using code flow now. I think we should handle it with the feature flag

feature-libs/organization/user-registration/components/form/user-registration-form.service.ts
```diff
@@ -163,12 +163,14 @@ export class UserRegistrationFormService {
...
+   * This only happens in case of the `ResourceOwnerPasswordFlow` & AuthorizationCode OAuth flows.
    */
   protected redirectToLogin(): void {
     if (
-      this.authConfigService.getOAuthFlow() ===
-      OAuthFlow.ResourceOwnerPasswordFlow
+      [
+        OAuthFlow.ResourceOwnerPasswordFlow,
+        OAuthFlow.AuthorizationCode,
+      ].includes(this.authConfigService.getOAuthFlow())
```
pawelfras: `major:` this change might be considered breaking for customers that are using code flow now.. I think we should handle it with the feature flag

feature-libs/organization/user-registration/components/verification-token-form/verification-token-form-component.service.ts
```diff
@@ -67,12 +67,14 @@ export class RegisterVerificationTokenFormComponentService {
...
+   * This only happens in case of the `ResourceOwnerPasswordFlow` & AuthorizationCode OAuth flows.
    */
   protected redirectToLogin(): void {
     if (
-      this.authConfigService.getOAuthFlow() ===
-      OAuthFlow.ResourceOwnerPasswordFlow
+      [
+        OAuthFlow.ResourceOwnerPasswordFlow,
+        OAuthFlow.AuthorizationCode,
+      ].includes(this.authConfigService.getOAuthFlow())
```
pawelfras: `major:` this change might be considered breaking for customers that are using code flow now.. I think we should handle it with the feature flag

feature-libs/user/profile/components/registration-verification-token-form/verify-register-verification-token-form.component.ts
FollowTheFlo: to be safe I added the feature flag condition.

feature-libs/organization/user-registration/components/form/user-registration-form.service.ts
```diff
@@ -163,12 +165,14 @@ export class UserRegistrationFormService {
...
-   * This only happens in case of the `ResourceOwnerPasswordFlow` OAuth flow.
+   * This only happens in case of the `ResourceOwnerPasswordFlow` OAuth flow and authorizationCodeFlowByDefault.
+   *
    */
   protected redirectToLogin(): void {
     if (
       this.authConfigService.getOAuthFlow() ===
-      OAuthFlow.ResourceOwnerPasswordFlow
+        OAuthFlow.ResourceOwnerPasswordFlow ||
+      this.featureConfigService.isEnabled('authorizationCodeFlowByDefault')
```
FollowTheFlo, line 175: I am not adding  OAuthFlow.AuthorizationCode condition as customer using it with a diff vendor might hit breaking change.

feature-libs/user/account/components/verification-token-form/verification-token-form.component.ts
```diff
@@ -90,7 +91,12 @@ export class VerificationTokenFormComponent implements OnInit {
           'verificationTokenForm.needInputCredentials',
           {}
         );
-        this.routingService.go(['/login']);
+
+        this.routingService.go(
+          this.featureConfigService.isEnabled('authorizationCodeFlowByDefault')
```
FollowTheFlo, line 96: avoiding breaking change. There might have been a reason behind the use of  ['/login'] value.

feature-libs/organization/user-registration/components/form/user-registration-form.service.ts
Pucek9, line 175: Is it necessary here? if it was not changed here (unless in epic we introduce that change?)

feature-libs/organization/user-registration/components/verification-token-form/verification-token-form-component.service.ts
```diff
@@ -67,12 +69,13 @@ export class RegisterVerificationTokenFormComponentService {
...
    *
-   * This only happens in case of the `ResourceOwnerPasswordFlow` OAuth flow.
+   * This only happens in case of the `ResourceOwnerPasswordFlow` OAuth flow and authorizationCodeFlowByDefault.
    */
   protected redirectToLogin(): void {
     if (
       this.authConfigService.getOAuthFlow() ===
-      OAuthFlow.ResourceOwnerPasswordFlow
+        OAuthFlow.ResourceOwnerPasswordFlow ||
+      this.featureConfigService.isEnabled('authorizationCodeFlowByDefault')
```
Pucek9, line 78: same here

feature-libs/user/profile/components/register/register.component.ts
```diff
@@ -235,8 +235,8 @@ export class RegisterComponent implements OnInit, OnDestroy {
     ) {
       this.router.go('login');
```
Pucek9, line 238: Why not fix it here as well?

Pucek9:
Could we update them as well? (along with a feature flag there)
By the wayside, we can improve the product we make.

image

feature-libs/user/profile/components/register/register.component.ts
FollowTheFlo: i'd love too, but there might be a valid reason for it?

feature-libs/organization/user-registration/components/form/user-registration-form.service.ts
FollowTheFlo: if we don't make this change, user stays on registration form.
We decided to route to login page instead for b2c and b2b

Pucek9: Ah ok, missed that its not a part of fixing hardcoded routes, but also neccesary change logic for flow

feature-libs/user/profile/components/register/register.component.ts
Pucek9: Because it's just a bug
What about wrapping all these parts in 
`if (!this.featureConfigService.isEnabled('authorizationCodeFlowByDefault')) {...} `?

projects/storefrontapp-e2e-cypress/cypress/helpers/b2b/b2b-user-registration.ts
```diff
@@ -93,11 +93,18 @@ export function submitOrganizationUserRegistrationForm(
...
+  whenJDK17(
+    () => {
+      cy.get('body').within(() => {
+        const alert = alerts.getAlert();
+
+        alert.should('contain', message);
+      });
+    },
+    () => {
+      cy.log('Test not relevant on JDK21');
```
FollowTheFlo, line 105: no global message on login page for jdk21 as we restart the app.

feature-libs/user/profile/components/register/register.component.ts
FollowTheFlo: i was assuming
 if (
      this.authConfigService.getOAuthFlow() ===
      OAuthFlow.ResourceOwnerPasswordFlow
    ) {
      this.router.go('login');
    }
    
would never be in conflict with
if (this.featureConfigService.isEnabled('authorizationCodeFlowByDefault')) {
      this.router.go({ cxRoute: 'login' });
    }
    I'll add && this.featureConfigService.isEnabled('authorizationCodeFlowByDefault' to be safe

PR #20595, opened 2025-08-08
chore: handle manual navigation to custom login page

ercultimate:


projects/core/src/auth/user-auth/facade/auth.service.ts
```diff
@@ -49,6 +47,10 @@ export class AuthService {
    */
   logoutInProgress$: Observable<boolean> = new BehaviorSubject<boolean>(false);
 
+  protected csrfToken$ = this.crossSiteRequestForgeryService
```
Pucek9, line 50: why do we break it down into method and property?

ercultimate: This is so that the shareReplay takes effect for multiple subscribers.  If we create a new observable with shareReplay each time `getCsrfToken()` is called, then each observable instance will make its own request and shareReplay for its response.

Pucek9: ah ok thanks, I was previously convinced that share replay somehow underneath was causing this

projects/core/src/auth/user-auth/guards/custom-login.guard.ts
```diff
@@ -0,0 +1,111 @@
...
+      this.storage as Storage
+    );
+  }
+
+  clearRedirectCount() {
+    persistToStorage(
+      STORAGE_KEY,
+      { t: Date.now(), c: 0 } satisfies CustomLoginGuardMetadata,
+      this.storage as Storage
+    );
```
Pucek9:
```suggestion
    this.setRedirectCount(0);
```

projects/core/src/auth/user-auth/guards/custom-login.guard.ts
```diff
@@ -0,0 +1,111 @@
...
+      STORAGE_KEY
+    );
+    if (countMeta && Date.now() - countMeta.t < timeout) {
+      return countMeta.c;
+    }
+    return 0;
+  }
+
+  setRedirectCount(count: number) {
+    return persistToStorage(
```
Pucek9:
```suggestion
    persistToStorage(
```

projects/core/src/auth/user-auth/guards/custom-login.guard.ts
```diff
@@ -0,0 +1,111 @@
...
+  persistToStorage,
+  readFromStorage,
+} from '../../../state/utils/browser-storage';
+import { WindowRef } from '../../../window/window-ref';
+import { AuthService } from '../facade/auth.service';
+import { CsrfStateService } from '../facade/csrf-state.service';
+
+const STORAGE_KEY = 'login_redirect_count';
+const timeout = 15_000;
+const totalRetries = 2;
```
Pucek9: tested a few scenarios and see that 1 is enough (as expected - 2 failed requests and a redirect)
```suggestion
const totalRetries = 1;
```

projects/core/src/auth/user-auth/guards/custom-login.guard.ts
Pucek9: if we don't use `StatePersistenceService`, our field looks like `login_redirect_count` instead of 
`spartacus⚿login_redirect_count`. Not sure if it's standard

But anyway, I don't see other usages of those functions outside `StatePersistenceService` in the whole project. What benefits do we have over `this.winRef.localStorage?.setItem(STORAGE_KEY, { t: Date.now(), c: count })` ?

projects/core/src/auth/user-auth/guards/custom-login.guard.ts
```diff
@@ -0,0 +1,111 @@
...
+  providedIn: 'root',
+})
+export class CustomLoginGuard implements CanActivate {
+  authService = inject(AuthService);
+  router = inject(Router);
+  semanticPathService = inject(SemanticPathService);
+  windowRef = inject(WindowRef);
+  storage = getStorage(StorageSyncType.LOCAL_STORAGE, this.windowRef);
+  globalMessageService = inject(GlobalMessageService);
+  csrfStateService = inject(CsrfStateService);
```
Pucek9:
```suggestion
  protected authService = inject(AuthService);
  protected router = inject(Router);
  protected semanticPathService = inject(SemanticPathService);
  protected windowRef = inject(WindowRef);
  protected storage = getStorage(StorageSyncType.LOCAL_STORAGE, this.windowRef);
  protected globalMessageService = inject(GlobalMessageService);
  protected csrfStateService = inject(CsrfStateService);
```


and other methods apart from `canActivate` also should be

projects/core/src/auth/user-auth/guards/custom-login.guard.ts
ercultimate: I used the storage methods for simplicity, and the utility methods handle [de]serialization of object values.

I don't know if there is a standard to use StatePersistenceService.  It seems to largely be used for ngrx state persistence.  

StatePersistenceService also requires the value to be passed as an observable.  While this should be safe because there are no async schedulers involved, that could become an issue later if it becomes asynchronous and the write operation is scheduled for a task, then canceled because of browser navigation.

projects/core/src/auth/user-auth/guards/custom-login.guard.ts
ercultimate: Updated, please verify.

projects/core/src/auth/user-auth/guards/custom-login.guard.ts
ercultimate: Updated, please verify.

projects/core/src/auth/user-auth/guards/custom-login.guard.ts
ercultimate: Updated, please verify.

projects/core/src/auth/user-auth/guards/custom-login.guard.ts
ercultimate: Updated, please verify.

projects/core/src/auth/user-auth/facade/csrf-state.service.ts
Pucek9, line 1: Its not a real facade, should it be in the facade folder?

also considering do we really need that

Pucek9 (approved):

Pucek9:
QA done: checked below scenarios:
- with valid jsessionid and not
- opening /sign-in and /login directly
- setting the wrong csrf endpoint in the configuration
works as expected

projects/core/src/auth/user-auth/facade/csrf-state.service.ts
ercultimate: I'm following the conventions from the surrounding code here.  I _believe_ the reason is that the module is not lazy-loaded, so the usual proxy-facades are not necessary.

feature-libs/user/account/components/login-form/login-form.module.ts
```diff
@@ -49,7 +50,7 @@ import { LoginFormComponent } from './login-form.component';
       cmsComponents: {
         ReturningCustomerLoginComponent: {
           component: LoginFormComponent,
-          guards: [NotAuthGuard],
+          guards: [NotAuthGuard, CustomLoginGuard],
```
pawelfras, line 53: `query:` do we want to run this guard even for password flow? I couldn't found any toggle/configuration that hides the csrf call which is redundant for password flow and may cause issues. Apologies if I'm missing something obvious.

ercultimate: No, that's very relevant.  The issue I see now is that we have no way to determine if the customLoginPage feature is enabled.  I can check the config, and use `Boolean(AuthConfig.authentication.customLogin)` as the decider.

ercultimate: I added a flag to the AuthConfigService to indicate when the feature is enabled, and used it to short-circuit the guard when the feature is disabled.  This will allow ROPC grant flow to function with appropriate configuration.

PR #20596, opened 2025-08-11
fix: Track login events for code flow authentication

sdrozdsap:


Pucek9 (approved):

integration-libs/cds/src/profiletag/services/profile-tag-lifecycle.service.ts
```diff
@@ -39,7 +48,20 @@ export class ProfileTagLifecycleService {
...
   loginSuccessful(): Observable<boolean> {
+    const cdsLoginEventsToken = this.featureConfigService.isEnabled(
+      'cdsLoginEventsToken'
+    );
+    if (cdsLoginEventsToken) {
+      return this.loginEnvelopes$.pipe(
+        filter(({ timestamp }) => this.lastProcessedTimestamp !== timestamp),
+        tap(({ timestamp }) => (this.lastProcessedTimestamp = timestamp)),
+        map(() => true)
+      );
```
ercultimate: Would this work?
```suggestion
      return this.loginEnvelopes$.pipe(
        distinctUntilChanged((prev, curr) => prev.timestamp === curr.timestamp),
        map(() => true)
      );
```

sdrozdsap: Hm, that looks super nice, let me test it

PR #20600, opened 2025-08-12
feat: customer login with OTP based on custom login page

niehuayang:


feature-libs/user/account/components/verification-token-form/verification-token-form-component.service.ts
```diff
@@ -41,24 +60,31 @@ export class VerificationTokenFormComponentService {
...
+          this.form.value.tokenCode
+        )
       )
-    )
-      .pipe(
-        withLatestFrom(this.auth.isUserLoggedIn()),
-        tap(([_, isLoggedIn]) => this.onSuccess(isLoggedIn))
-      )
-      .subscribe();
+        .pipe(
```
i53577, line 84: do we need to add take(1) to ensure auto unsubscribe?

feature-libs/user/account/components/verification-token-form/verification-token-form-component.service.ts
```diff
@@ -6,28 +6,47 @@
...
+  private featureConfigService = inject(FeatureConfigService);
+  protected auth: AuthService = inject(AuthService);
+
+  action?: string;
+  method?: string;
+  csrf$ = this.auth.getCsrfToken();
+
+  constructor() {
+    if (this.featureConfigService.isEnabled('authorizationCodeFlowByDefault')) {
+      this.method = 'POST';
```
i53577: what about extract the logic that in `if` block into a function and call it here?

feature-libs/user/account/components/verification-token-form/verification-token-form.component.html
```diff
@@ -28,7 +35,21 @@
...
 
-  <input type="hidden" formControlName="tokenId" [(ngModel)]="tokenId" />
+  <input
+    type="hidden"
+    formControlName="tokenId"
+    name="username"
+    [(ngModel)]="tokenId"
+  />
+
+  <ng-container *cxFeature="'authorizationCodeFlowByDefault'">
```
i53577, line 45: do we have to wrap the <input>?
```suggestion
<input 
  *cxFeature="'authorizationCodeFlowByDefault'"
  *ngIf="csrf$ | async as csrf"
  formControlName="csrf"
  type="hidden"
  [name]="csrf.parameterName"
/>
```

feature-libs/user/account/components/verification-token-form/verification-token-form-component.service.ts
niehuayang: good point, but i had never updated the part of code, just adjust the code format, so I prefer to keep the same as origin

feature-libs/user/account/components/verification-token-form/verification-token-form-component.service.ts
niehuayang: refined code

i53577 (approved):
LGTM

feature-libs/user/account/components/verification-token-form/verification-token-form-component.service.ts
```diff
@@ -90,4 +111,13 @@ export class VerificationTokenFormComponentService {
...
     this.busy$.next(false);
   }
+
+  protected initializeDataforAuthorizationCodeFlow(): void {
+    this.method = 'POST';
+    this.action = this.authConfigService?.getCustomLoginFormEndpoint();
+    this.form.addControl('csrf', new FormControl('', Validators.required));
+    this.csrf$.pipe(take(1)).subscribe((csrf) => {
+      this.form.get('csrf')?.setValue(csrf.token);
+    });
```
ercultimate: The logic for CustomLoginPage csrf token has changed.  The token is now fetched as part of the Route Guard validations, and shared through CsrfStateService.

You may continue to request the csrf token, it would just be duplicating the request.

ercultimate (approved):

feature-libs/user/account/components/verification-token-form/verification-token-form-component.service.ts
niehuayang: well, this is a better way to get csrf token, and I refer to the [PR](https://github.com/SAP/spartacus/pull/20595) to use the CustomLoginGuard, thanks for your comment.

PR #20613, opened 2025-08-13
chore:jdk21 feature flag assessment

ercultimate:


feature-libs/order/components/order-confirmation/order-guest-register-form/order-guest-register-form.component.ts
```diff
@@ -75,7 +78,8 @@ export class OrderGuestRegisterFormComponent implements OnDestroy {
       );
       if (
         !this.subscription &&
-        !this.featureConfigService.isEnabled('authorizationCodeFlowByDefault')
+        this.authConfigService.getOAuthFlow() ===
```
ercultimate, line 81: `!authorizationCodeFlowByDefault` is the same as saying "do this for ROPC Grant only".

feature-libs/organization/user-registration/components/form/user-registration-form.service.ts
```diff
@@ -172,7 +172,9 @@ export class UserRegistrationFormService {
     if (
       this.authConfigService.getOAuthFlow() ===
         OAuthFlow.ResourceOwnerPasswordFlow ||
-      this.featureConfigService.isEnabled('authorizationCodeFlowByDefault')
+      this.featureConfigService.isEnabled(
```
ercultimate, line 175: This change is related to the enhancements we made to support oAuth flows involving a redirect.  It is not related to the change in the default auth config.

feature-libs/organization/user-registration/components/form/user-registration-form.service.ts
```diff
@@ -172,7 +172,9 @@ export class UserRegistrationFormService {
...
+        'nativeSupportForRedirectingOAuthFlows'
+      )
```
ercultimate, line 177: This change is related to the enhancements we made to support oAuth flows involving a redirect.  It is not related to the change in the default auth config.

feature-libs/organization/user-registration/components/verification-token-form/verification-token-form-component.service.ts
```diff
@@ -75,7 +75,9 @@ export class RegisterVerificationTokenFormComponentService {
     if (
       this.authConfigService.getOAuthFlow() ===
         OAuthFlow.ResourceOwnerPasswordFlow ||
-      this.featureConfigService.isEnabled('authorizationCodeFlowByDefault')
+      this.featureConfigService.isEnabled(
+        'nativeSupportForRedirectingOAuthFlows'
```
ercultimate, line 79: This change is related to the enhancements we made to support oAuth flows involving a redirect.  We now want to always redirect to login, not only for ROPC grant.

feature-libs/organization/user-registration/components/verification-token-form/verification-token-form.component.ts
```diff
@@ -93,11 +93,12 @@ export class RegisterVerificationTokenFormComponent implements OnInit {
           {}
         );
         this.routingService.go(
-          this.featureConfigService.isEnabled('authorizationCodeFlowByDefault')
+          this.featureConfigService.isEnabled(
+            'nativeSupportForRedirectingOAuthFlows'
+          )
```
ercultimate, line 98: We could probably create another feature flag to fix string-based router destinations, but if we wanted to piggy-back on another flag, then it should be part of the changes to enhance support for oAuth flows involving a redirect.

feature-libs/user/account/components/guards/login-as-guest.guard.ts
```diff
@@ -22,10 +24,15 @@ export class LoginAsGuestGuard {
...
+  protected authConfigService = inject(AuthConfigService);
 
   canActivate(): Observable<GuardResult> {
     if (
-      this.featureConfigService.isEnabled('authorizationCodeFlowByDefault') &&
+      this.featureConfigService.isEnabled(
+        'nativeSupportForRedirectingOAuthFlows'
+      ) &&
+      this.authConfigService.getOAuthFlow() !==
+        OAuthFlow.ResourceOwnerPasswordFlow &&
```
ercultimate, line 35: This change is related to the enhancements we made to support oAuth flows involving a redirect.  

We want to disable this new guard if the enhanced support is disabled, or if the oauth flow is a redirecting flow.

feature-libs/user/account/components/login-form/login-form-component.service.ts
```diff
@@ -71,10 +69,7 @@ export class LoginFormComponentService {
       this.form.markAllAsTouched();
       return;
     }
-    if (
-      this.featureConfigService.isEnabled('authorizationCodeFlowByDefault') &&
-      nativeForm
-    ) {
+    if (this.authConfigService.customLoginEnabled() && nativeForm) {
```
ercultimate, line 72: This should be verified with the author.  If custom login page is disabled, we wouldn't want to use the native form submission.

feature-libs/user/account/components/login-form/login-form.component.html
```diff
@@ -66,14 +66,12 @@
...
-      type="hidden"
-      [name]="csrf.parameterName"
-    />
-  </ng-container>
+  <input
+    *ngIf="csrf"
+    formControlName="csrf"
+    type="hidden"
+    [name]="csrf.parameterName"
+  />
```
ercultimate, line 74: The feature flag here isn't necessary because the `csrf` property is only filled when custom login page is enabled.  We could add a property to the component that reflects `customLoginEnabled()` for more explicit logic.

feature-libs/user/account/components/verification-token-form/verification-token-form.component.ts
```diff
@@ -93,7 +93,9 @@ export class VerificationTokenFormComponent implements OnInit {
         );
 
         this.routingService.go(
-          this.featureConfigService.isEnabled('authorizationCodeFlowByDefault')
+          this.featureConfigService.isEnabled(
+            'nativeSupportForRedirectingOAuthFlows'
+          )
```
ercultimate, line 98: This change is related to the enhancements we made to support oAuth flows involving a redirect.  Again, could also be another separate flag.

feature-libs/user/profile/components/register/register.component.ts
```diff
@@ -229,13 +229,20 @@ export class RegisterComponent implements OnInit, OnDestroy {
...
+        'nativeSupportForRedirectingOAuthFlows'
+      )
     ) {
-      this.router.go('login');
+      this.router.go(
+        this.featureConfigService.isEnabled(
+          'nativeSupportForRedirectingOAuthFlows'
+        )
+          ? { cxRoute: 'login' }
+          : 'login'
```
ercultimate, line 244: This change is related to the enhancements we made to support oAuth flows involving a redirect.  Alternatively, a separate flag for router destinations.

feature-libs/user/profile/components/register/register.component.ts
```diff
@@ -229,13 +229,20 @@ export class RegisterComponent implements OnInit, OnDestroy {
...
-    if (this.featureConfigService.isEnabled('authorizationCodeFlowByDefault')) {
-      this.router.go({ cxRoute: 'login' });
-    } else if (
+    if (
       this.authConfigService.getOAuthFlow() ===
-      OAuthFlow.ResourceOwnerPasswordFlow
+        OAuthFlow.ResourceOwnerPasswordFlow ||
+      this.featureConfigService.isEnabled(
+        'nativeSupportForRedirectingOAuthFlows'
+      )
```
ercultimate, line 237: This change is related to the enhancements we made to support oAuth flows involving a redirect.  

We always want to redirect to login on register success.

feature-libs/user/profile/components/registration-verification-token-form/verify-register-verification-token-form.component.ts
```diff
@@ -143,7 +143,9 @@ export class RegistrationVerificationTokenFormComponent implements OnInit {
       !this.lastName
     ) {
       this.router.go(
-        this.featureConfigService.isEnabled('authorizationCodeFlowByDefault')
+        this.featureConfigService.isEnabled(
+          'nativeSupportForRedirectingOAuthFlows'
+        )
```
ercultimate, line 148: This change is related to the enhancements we made to support oAuth flows involving a redirect.  Alternatively, add a new separate flag for router destinations.

feature-libs/user/profile/components/registration-verification-token-form/verify-register-verification-token-form.component.ts
```diff
@@ -251,7 +253,11 @@ export class RegistrationVerificationTokenFormComponent implements OnInit {
       this.router.go('login');
     }
 
-    if (this.featureConfigService.isEnabled('authorizationCodeFlowByDefault')) {
+    if (
+      this.featureConfigService.isEnabled(
+        'nativeSupportForRedirectingOAuthFlows'
+      )
+    ) {
```
ercultimate, line 260: This change is related to the enhancements we made to support oAuth flows involving a redirect.  

We always want to redirect to login after successful register.  This should probably be combined with the login route directly above, with a flag used to select between string or cxroute destination.

feature-libs/user/profile/core/facade/user-register.service.ts
```diff
@@ -52,7 +55,8 @@ export class UserRegisterService implements UserRegisterFacade {
     this.userConnector.registerGuest(payload.guid, payload.password).pipe(
       tap((user) => {
         if (
-          !this.featureConfigService.isEnabled('authorizationCodeFlowByDefault')
+          this.authConfigService.getOAuthFlow() ===
+          OAuthFlow.ResourceOwnerPasswordFlow
```
ercultimate, line 59: `!authorizationCodeFlowByDefault` is just "when ROPC"

projects/core/src/auth/user-auth/services/oauth-lib-wrapper.service.ts
```diff
@@ -118,11 +119,13 @@ export class OAuthLibWrapperService {
    */
   initLoginFlow() {
     if (
-      !this.featureConfigService.isEnabled('authorizationCodeFlowByDefault')
+      !this.featureConfigService.isEnabled(
+        'nativeSupportForRedirectingOAuthFlows'
+      ) ||
+      this.authConfigService.getOAuthFlow() ===
+        OAuthFlow.ResourceOwnerPasswordFlow
```
ercultimate, line 126: This change is related to the enhancements we made to support oAuth flows involving a redirect.

When the support is enhanced, we want to set the localstorage item on ROPC flow.  
When enhanced support is off, we want to always set the localstorage item.

projects/storefrontlib/cms-structure/routing/default-routing-config.ts
```diff
@@ -26,7 +26,9 @@ export const defaultRoutesConfigFactory: () => RoutingConfig = () => {
            * Legacy path will stay, new one is updated.           *
            */
           paths: [
-            featureToggles.authorizationCodeFlowByDefault ? 'sign-in' : 'login',
+            featureToggles.nativeSupportForRedirectingOAuthFlows
+              ? 'sign-in'
+              : 'login',
```
ercultimate, line 31: This change is related to the enhancements we made to support oAuth flows involving a redirect.

projects/storefrontlib/cms-structure/routing/default-routing-config.ts
```diff
@@ -93,7 +95,7 @@ export const defaultRoutesConfigFactory: () => RoutingConfig = () => {
    * Configuration necessary to allow customization of login form path,
    * which have to be the same as configured in oauth client
    */
-  if (featureToggles.authorizationCodeFlowByDefault) {
+  if (featureToggles.nativeSupportForRedirectingOAuthFlows) {
```
ercultimate, line 98: This change is related to the enhancements we made to support oAuth flows involving a redirect.

At the very least, it is related to custom login, and should be added when the feature is enabled.

projects/core/src/features-config/services/feature-config.service.ts
```diff
@@ -11,6 +11,11 @@ import {
   isFeatureLevel,
 } from '../utils/feature-config-utils';
 
+/**
+ * This service is not recommended for use.
+ *
+ * Please use `FeatureToggles` instead.
+ */
```
pawelfras, line 18: `major:` `FeatureToggles` was meant to allow creating config factories depending on feature flags without causing a circular dependency issue - `FeatureConfigService` having `FeatureConfig` as dependency does so when injecting to a factory affecting the global `Config` object.
That said, FeatureConfigService is still ok to use, but not in config factories.

projects/core/src/features-config/feature-toggles/config/feature-toggles.ts
```diff
@@ -977,4 +987,5 @@ export const defaultFeatureToggles: Required<FeatureTogglesInterface> = {
   disableClientTokens: false,
   incrementProcessesCountForMergeCart: true,
   dispatchLoginActionOnlyWhenTokenReceived: false,
+  nativeSupportForRedirectingOAuthFlows: true,
```
pawelfras, line 990: `major:` To not provide breaking changes for current customers, let's change its value to `false`
When it comes to fresh apps, there is a [special script](https://github.com/SAP/spartacus/blob/f28330fc71c03d2e7841823e445a99e71b9ebf7d/projects/schematics/src/add-spartacus/spartacus-feature-toggles.ts#L26) that sets all flags to `true` when scaffolding the app
```suggestion
  nativeSupportForRedirectingOAuthFlows: false,
```

projects/storefrontlib/cms-structure/routing/default-routing-config.ts
pawelfras: Ideally, this change should depend on the auth config and info if we use a custom login page. Unfortunately, injecting `AuthConfig` will cause circular dependency. Maybe we should consider the separate feature flag only for the custom login page? 🤔

feature-libs/user/profile/core/facade/user-register.service.ts
pawelfras: That's right, but the main goal of feature toggles is avoiding breaking changes for customers. If someone is using authorisation code flow in JDK 17, then the proposed condition may break their app.

feature-libs/user/account/components/login-form/login-form-component.service.ts
pawelfras: @Pucek9 could you take a look?

feature-libs/order/components/order-confirmation/order-guest-register-form/order-guest-register-form.component.ts
pawelfras: Like in one of the previous comments, this condition might be considered breaking for current JDK 17 customers who use code flow.

projects/storefrontlib/cms-structure/routing/default-routing-config.ts
pawelfras: On the other hand, changes in code that depend on `nativeSupportForRedirectingOAuthFlows` are made because of this change in the default routing config 🤔 Maybe w could consider renaming this feature flag to `routingForCustomLoginPage`?

PR #20621, opened 2025-08-13
fix: Improper Textbox appears for Password in SSR mode

Pucek9:
- add hydration to all components that use `cxPasswordVisibilitySwitch`

Closes [CXSPA-10754](https://jira.tools.sap/browse/CXSPA-10754)


Pucek9:
Steps to reproduce:

0 On develop branch :
1 run `yarn build:libs && yarn build && yarn build:ssr`
2 `NODE_TLS_REJECT_UNAUTHORIZED=0 yarn serve:ssr`
3 Open URL directly/reload page: http://localhost:4000/electronics-spa/en/USD/login
4 See the wrongly rendered password field
image

Expected result:
4 See the login form rendered in the network tab (preview), verify UI that the password field works properly 
5 Navigate by UI between the login and register pages to validate that the CSR render is also properly working 
6 Feel free to verify other affected components (register will work only after merge https://github.com/SAP/spartacus/pull/20624)

projects/storefrontlib/shared/components/captcha/mock-captcha/mock-captcha.service.ts
```diff
@@ -35,39 +34,46 @@ export class MockCaptchaService extends CaptchaService {
...
-    this.label = document.createElement('label');
-    this.label.textContent = "I'm not a robot";
-    this.container.appendChild(this.checkbox);
-    this.container.appendChild(this.label);
-
-    this.spinner = document.createElement('icon');
-    this.spinner.className = 'fa-solid fa-spinner';
-
-    this.checkbox.addEventListener('change', this.onCheckBoxClicked.bind(this));
+    this.container = this.renderer.createElement('div');
```
Pucek9: @pawelfras Is it a breaking change and should a new flag be added?

projects/storefrontlib/shared/components/captcha/captcha.service.ts
pawelfras, line 1: As discussed, let's address the issue caused by the captcha service in a separate PR because it is not related to the issue in the login form.

pawelfras (commented):
This will fix the issue, but, as discussed, let's timebox fixing the root cause in `PasswordVisibilityToggleDirective`.

Pucek9:
I recognized 3 possible solutions:

**1.** Add `host: { ngSkipHydration: 'true' },` for each component that uses  `cxPasswordVisibilitySwitch`
note: adding it to the directive itself doesn't change anything
note2: Using Renderer2 doesn't help as well
\- update necessary in multiple files
\- it's a workaround
\+ but officially guided
\+ doesn't change visual experience

**2.** Execute logic only on CSR, so there will be no confusion with the structure
image

\- there is a small delay between rendering the page and `eye-icon` appearance
\+ update directive only 

**3.** Use a component for wrapping that input over the directive
if there will be no "magic" with changing structure, just static structure with `ng-content`
\- a little bit more effort
\- more changes (breaking)
\+ doesn't change visual experience
\+ seems like only one serious solution in that particular case

FollowTheFlo (approved):
based on offline discussion and urgency, happy with it.

PR #20624, opened 2025-08-14
fix: register page is not rendered by SSR

Pucek9:
- fix ssr on register form page (`mock-captcha.service`)

Closes [CXSPA-10775](https://jira.tools.sap/browse/CXSPA-10775)


Pucek9:
Steps to reproduce:

0 On develop branch :
1 run `yarn build:libs && yarn build && yarn build:ssr`
2 `NODE_TLS_REJECT_UNAUTHORIZED=0 yarn serve:ssr`
3 Open URL directly/reload page: http://localhost:4000/electronics-spa/en/USD/login/register
4 See following error in the network tab:  
image

Expected result:
4 See register form rendered in network tab (preview)

Pucek9:
Hello @scarai-sap @i53577 
Please help with the QA of those changes

projects/storefrontlib/shared/components/captcha/mock-captcha/mock-captcha.service.ts
```diff
@@ -35,39 +34,46 @@ export class MockCaptchaService extends CaptchaService {
 
   protected spinner: HTMLElement;
 
+  protected checkboxListener: () => void;
```
scarai-sap: maybe change to removeEventListener

scarai-sap (dismissed):

projects/storefrontlib/shared/components/captcha/mock-captcha/mock-captcha.service.ts
```diff
@@ -35,39 +34,46 @@ export class MockCaptchaService extends CaptchaService {
...
-    this.label = document.createElement('label');
-    this.label.textContent = "I'm not a robot";
-    this.container.appendChild(this.checkbox);
-    this.container.appendChild(this.label);
-
-    this.spinner = document.createElement('icon');
-    this.spinner.className = 'fa-solid fa-spinner';
-
-    this.checkbox.addEventListener('change', this.onCheckBoxClicked.bind(this));
+    this.container = this.renderer.createElement('div');
```
i53577, line 40: Perhaps these creation logic can be extract into a function to make the `initialize()` function more concise

scarai-sap (approved):

Pucek9:
> Hello @scarai-sap @i53577 Please help with the QA of those changes

>To enable captcha without changing things in backend, you can just change the code  [fetchCaptchaConfigFromServer](https://github.com/SAP/spartacus/blob/83390ad2a6557c6935166991d091085902010e3f/projects/storefrontlib/shared/components/captcha/captcha.service.ts#L68) to below fake implementation
```ts
  fetchCaptchaConfigFromServer(): Observable<CaptchaConfig> {
    return of ( {
      enabled: true,
      publicKey: "123456789abcdef123456789abcdef"
    });
  }
  ```

File: projects/storefrontlib/shared/components/captcha/mock-captcha/mock-captcha.service.ts [hunk 3]
```diff
@@ -78,15 +87,21 @@ export class MockCaptchaService extends CaptchaService {
...
-      renderParams.element.appendChild(this.container);
+      this.renderer.appendChild(renderParams.element, this.container);
     }
 
     return this.retVal.asObservable();
   }
+
+  ngOnDestroy() {
+    if (this.removeCheckboxListener) {
+      this.removeCheckboxListener();
```
FollowTheFlo, line 103: any method needed?, as we bind at initialize() time, I guess we need to unbind the event ?

Pucek9: as discussed, thats how this api works:
image

Additionally, as suggested: considering removal of `ngOnDestroy` as service is provideIn: 'root' anyway. 

Calling removeCheckboxListener, I guess, should be done from outside (in the Adapter and the Component that use it) or it should no longer be in the root
FYI @scarai-sap

Platonn: Reminder: we should cleanup subscriptions also in singleton services in ngOnDestroy - to avoid memory leaks in SSR 

Why:

**One NodeJS process, but many rendered apps**
The NodeJS process of the SSR server on every http request: bootstraps a fresh Angular app, sends the output HTML to the client, and destroys the app. This cycle is repeated for each request inside one shared long-living NodeJS process. So any pending subscription created in any app’s service survives even after the app is destroyed. And it causes memory leaks in the server.

**Unsubscribe in services’ ngOnDestroy**
Angular invokes in SSR the lifecycle method ngOnDestroy not only for components, but also for every service, when the app is destroyed. Therefore it’s a good place to unsubscribe any pending subscriptions created by the service.

FollowTheFlo (approved):
lgtm

PR #20644, opened 2025-08-15
chore: add JDK toggle and client auth within install script

FollowTheFlo:
3 new var in config.default.sh:
- JDK_VERSION -> eaither JDK21 or JDJ17, it toggle authorizationCodeFlowByDefault flag on spartacus-features.module.ts
- ADD_AUTH_CONFIG -> boolean to neable authConfig overwriting with AUTH_CONFIG below
- AUTH_CONFIG -> AuthConfig object that will be added within spartacus-features.module.ts if running on JDK21

Those functions run at 'app starting' time just before compiling.
on command ./run.sh start
No need to reinstall if change is needed.


scripts/install/functions.sh
```diff
@@ -1085,3 +1087,125 @@ function compareSemver() {
...
+    clean_module_file() {
+       #Remove all empty lines then add one above @NgModule
+       echo "Cleaning up empty lines"
+       sed_inplace -e '/^[[:space:]]*$/d' -e '/@NgModule/i\' "$SPARTACUS_CONFIGURATION_MODULE_PATH"
+    }
+
+    if [ "$ADD_AUTH_CONFIG" = false ]; then
+        echo "ADD_AUTH_CONFIG is false -> Do not add AuthConfig."
+        echo "Remove authconfig import and provider if it was previously added."
+        perl -0777 -pi -e 's/provideConfig\(<AuthConfig>.*?\}\),\s*//gs' "$SPARTACUS_CONFIGURATION_MODULE_PATH"
```
FollowTheFlo, line 1170: use case: JDK21 without adding authConfig (ADD_AUTH_CONFIG=false).
the script still open and clean configuration.module.t file.
Reason is that user might have previously run the app with custom authConfig (AUTH_CONFIG with ADD_AUTH_CONFIG=true). we thus need a mechanism to clean previous config even if it is the first attempt.

To avoid this cleaning mechansim, the user can comment the AUTH_CONFIG var from config.default.sh, the function will exit on line 1151.

scripts/install/config.default.sh
```diff
@@ -6,7 +6,7 @@
 # Url of the hybris backend
 # Will replace default host (https://localhost:9002) as a backend endpoint
 # Make sure you specify the full url for the backend (https://[host]:[port]
-BACKEND_URL="https://40.76.109.9:9002"
```
pawelfras: `query:` This is an interim change, right? Maybe it's worth adding a comment above to revert it when the dev env is updated to JDK 21.

scripts/install/config.default.sh
```diff
@@ -108,3 +108,30 @@ CHECK_B2B_AFTER_START=false
 # Forces E2E even if XVFB is not installed
 FORCE_E2E=false
 SKIP_E2E=false
+
+#JDK toggle: JDK21 or JDK17
+JDK_VERSION="JDK21"
```
pawelfras: `query:` is this value type dictated by e.g. Jenkins job? If not, let's consider changing it to a boolean flag with a name like `IS_JDK_21` - it might be more typo-proof. Unless we know there will be a need for JDK23, JDK25 and soon in the future, then this proposal is more future-proof :smile:

scripts/install/config.default.sh
FollowTheFlo: yep, i need to chat with GC about it, I'll revert it for now.

scripts/install/config.default.sh
FollowTheFlo: yep, i was using the ccv2 script logic on:
https://github.tools.sap/cx-commerce/ec-automate-pipelines/blob/master/scripts/deploy-storefront-from-source.sh
update_spartacus_features_module function is similar to our 'function setJdkVersion', I added more checks to see if path/file exists...

pawelfras (commented):
Did a QA and everything as expected.

Performed QA steps:
- go to `scripts/install`
- run 'bash run.sh install'
- run 'bash run.sh start'
- go to http://localhost:4200/electronics-spa/en/USD/login and try to login to verify if authorization code flow is enabled
   - similar test is not possible for SSR app due to different port and missing dedicated auth client or adjustments in existing client
- run `bash run.sh stop`
- change `ADD_AUTH_CONFIG` to true and run `run bash.sh start`
- verify in logs that AuthConfig was added
   image
- run `bash run.sh stop`
- change `JDK_VERSION` to `JDK17`
- run `bash run.sh start` and verify in logs that `authorizationCodeFlowByDefault` is disabled
   image

- go to, try to perform login and verify that the request was sent with `grant_type: password` and `client_id: mobile_android` (so it is old config)
   image

pawelfras (approved):

PR #20682, opened 2025-08-21
test: Fix flaky a11y tests

sdrozdsap:


sdrozdsap:
Fix to flakiness was tested using for loop to execute each spec 20 times. 
image
[Build details](https://cloud.cypress.io/projects/k3nmep/runs/50200/specs?utm_source=github)

projects/storefrontapp-e2e-cypress/cypress/support/continuum.commands.ts
```diff
@@ -48,31 +48,6 @@ const withContinuum = <T extends (...args: any[]) => any>(fn: T): T => {
...
-    /**
-     *  Avoid exposing API key in case of error.
-     */
-    Cypress.on('fail', (error) => {
-      if (error.message.includes(LEVEL_ACCESS_API)) {
-        error.message =
-          'There was an issue submitting accessibility concerns to AMP. Please confirm correct credentials and connection.';
-      }
-      throw error;
-    });
```
sdrozdsap, line 74: Not needed as we don't send any details to AMP server

projects/storefrontapp-e2e-cypress/cypress/support/continuum.commands.ts
```diff
@@ -48,31 +48,6 @@ const withContinuum = <T extends (...args: any[]) => any>(fn: T): T => {
...
-    const origLog = Cypress.log.bind(Cypress);
-    Cypress.log = function (opts, ...other) {
-      if (
-        opts.displayName >= LEVEL_ACCESS_API ||
-        opts.name >= LEVEL_ACCESS_API
-      ) {
-        return;
-      }
-      return origLog(opts, ...other);
-    };
```
sdrozdsap, line 63: This guy is responsible for breaking intercept/wait logic from default cypress behavior. The override caused internal error which was hard to track, but generally it prevented us from using the best practice there is for writing stable cypress which is intercepting the call and waiting for it. 
I've completely removed it also because I don't think we have a valid concern of exposing anything important here, but I might be wrong so Im open to any suggestions

Zeyber: Great job on tracing this! I believe we are keeping the door open on integrating with the level access platform as it might help us with more automated compliance in the future. I'm not really sure how that is handled at this stage. But I think we should leave this code here due to the security concern is resolves with a warning about the potential stability issues. Are we able to hide the overriding code behind the condition of setting the `LEVEL_ACCESS_API` token and produce a warning about the stability issue?

Zeyber (changes requested):

petarmarkov9449:
The `Cypress.log` override was the root cause of the intercept/wait issues. By replacing Cypress’ internal logger, we were unintentionally suppressing the events Cypress relies on to track XHR lifecycle, which explains the instability. The string comparison used to filter Level Access calls also wasn’t reliably targeting the endpoint, so other logs were being swallowed as well.

Removing the override is the right move.
If we still need to hide Level Access traffic, we can safely do so via:

`cy.intercept(`${LEVEL_ACCESS_API}*`, { log: false });`

How about this one?

Zeyber:
> `cy.intercept(`${LEVEL_ACCESS_API}*`, { log: false });`

Could you please try this and confirm it works?

petarmarkov9449:
> > `cy.intercept(`${LEVEL_ACCESS_API}*`, { log: false });`
> 
> Could you please try this and confirm it works?

Can you tell me which e2e shows `LEVEL_ACCESS_API`?

projects/storefrontapp-e2e-cypress/cypress/support/continuum.commands.ts
Zeyber: I forgot we already removed all the integration with the AMP dashboard. So these suppression methods don't actually carry any value at all now. Lets just remove that `const LEVEL_ACCESS_API` and I think we're done.

petarmarkov9449:
https://github.com/SAP/spartacus/pull/20682#discussion_r2444160054
So revert my commits.

projects/storefrontapp-e2e-cypress/cypress/support/continuum.commands.ts
Zeyber: Lets remove this as I don't think we need it anymore

Zeyber (changes requested):

projects/storefrontapp-e2e-cypress/cypress/support/continuum.commands.ts
petarmarkov9449: Just removed.

Zeyber (approved):

PR #20690, opened 2025-08-26
test: Collect a11y e2e failure screenshots

sdrozdsap:
See [this comment](https://github.com/SAP/spartacus/pull/20690#issuecomment-3253207766) as an example


.github/workflows/ci-merge-checks.yml
```diff
@@ -217,9 +217,48 @@ jobs:
...
           GITHUB_MATRIX_CONTAINER: ${{ matrix.containers }}
           GITHUB_MATRIX_TOTAL: ${{ strategy.job-total }}
         run: ci-scripts/e2e-cypress.sh -s a11y --skip-build
+      - name: Collect test screenshots
+        if: failure()
+        env:
+          GITHUB_MATRIX_CONTAINER: ${{ matrix.containers }}
+          GITHUB_PR_NUMBER: ${{ github.event.pull_request.number }}
+        run: |
+          chmod +x ci-scripts/collect-a11y-screenshots.sh
```
Zeyber: Can we push this file with executable permission already set or do we still need to do it on ci?

Zeyber (changes requested):
Nice work! Would it be possible/difficult to decouple the screenshots post exclusively for the a11y tests? We would run this for a11y tests for sure. But it could be useful for other tests perhaps when Cypress dashboard runs out of recordings. It could be scoped into another ticket though. What do you think?

sdrozdsap:
@Zeyber  I've switched to more isolated decoupled solution, which is much simpler and IMO more readable and easier to adopt by other testing suites in the future

Zeyber:
Very nice solution and very easy to follow add suites! Just a couple points:
- Could the comment include the test suites that have failed or some stats to make the link look more recognizable in the pipeline? For example:
```
E2E Tests Failed
- A11y (32/35) (screenshots)
- B2C (130/132) (screenshots) 
// B2B passes so we don't show it here.
```
This is not necessary though. I just think it would be nice. However, we could make a new ticket for this improvement if its complex.
- If we have a new commit, will it add new comments or update existing one? I think it'd be nice to see retriggers in commit history to make sure the intended behaviour works correctly.

Zeyber (approved):

PR #20698, opened 2025-08-29
test: product_configurator_e2e test suite fail on S2 and S3 (CXSPA-10752)

Larisa-Staroverova:
- I've changed the order of the **waits**. For example, it first waits for a **GET** response and then checks whether the configuration has rendered the corresponding elements.
- I've also adjusted most of the **check**..() methods to scroll to the respective element first. This was because the generated video during the test execution might not have clearly indicated which element was currently in focus. Then, the element's visibility is checked.
- For methods that trigger a click, a click event is triggered as soon as the element is visible.
- **viewportContext** is set explicitly for each test suite because there were resolution issues in the pipeline.


Uli-Tiger (approved):
lgtm

PR #20712, opened 2025-09-05
chore: Cleanup avoidCachingErrors feature flag

sdrozdsap:


core-libs/setup/ssr/optimized-engine/optimized-ssr-engine.spec.ts
```diff
@@ -392,14 +391,11 @@ describe('OptimizedSsrEngine', () => {
     }));
   });
 
-  describe('avoidCachingErrors option', () => {
```
pawelfras, line 395: I think we could remove this entire block, as it was dedicated to the `avoidCachingErrors` flag. There is a separate
```ts
 describe('shouldCacheRenderingResult option', () => {})
```
where we could add additional test case checking behaviour for the default `shouldCacheRenderingResult` value.

pawelfras (approved):
Review and QA done. LGTM

PR #20714, opened 2025-09-05
chore: add authConfig for JDK21 ssr app

FollowTheFlo:
context:
On a single installation, CSR and SSR run on different domains, eg for b2c: localhost:4200 and localhost:4100
Which means they must have distinct authConfig to point to distinct cientID.

We know have 2 authConfig objects in the installer speecific to each mode:
AUTH_CONFIG_CSR
with current default value point to client_id: "mobile_android_public",
AUTH_CONFIG_SSR
with current default value point to client_id: "mobile_android_public_ssr",


scripts/install/config.default.sh
```diff
@@ -6,7 +6,7 @@
 # Url of the hybris backend
 # Will replace default host (https://localhost:9002) as a backend endpoint
 # Make sure you specify the full url for the backend (https://[host]:[port]
-BACKEND_URL="https://40.76.109.9:9002"
+BACKEND_URL="https://api.c432wmya2v-teamspart3-s4-public.model-t.myhybris.cloud"
```
FollowTheFlo, line 9: As agreed with Kate
set JDK21 by default, so modified value of keys:
BACKEND_URL, JDK_VERSION, ADD_AUTH_CONFIG

ercultimate (dismissed):

ercultimate (approved):

PR #20715, opened 2025-09-08
chore: remove allPageMetaResolversEnabledInCsr feature flag

ercultimate:


projects/core/src/cms/page/config/page-meta.config.ts
```diff
@@ -26,13 +26,6 @@ export interface PageMetaResolversConfig {
    * Specifies how the canonical url is created.
    */
   canonicalUrl?: CanonicalUrlOptions;
-
-  /**
-   * Enables resolvers in dev mode regardless of the CSR configuration. This
-   * flag will override the disabling in CSR, which can be useful during development
-   * and debugging.
-   */
-  enableInDevMode?: boolean;
```
ercultimate, line 35: This change is the result of a discussion with Kris.  The flag was actually making the resolvers run all the time, ignoring this property.  By removing the flag, we restored a more restrictive application of the resolvers.  

Since this property's behavior was primarily a utility option for development work, it was decided to remove it rather than restore the restrictive behavior.

FollowTheFlo (approved):
lgtm

PR #20726, opened 2025-09-12
fix: auto submit payment form to the iframe

Matejk00:
Closes: CXSPA-10877


integration-libs/opf/checkout/components/opf-checkout-payment-wrapper/opf-checkout-payment-wrapper.component.ts
```diff
@@ -88,6 +95,15 @@ export class OpfCheckoutPaymentWrapperComponent implements OnInit, OnDestroy {
     this.service.reloadPaymentMode();
   }
 
+  protected submitFormToIframe(): void {
+    if (this.isPaymentDataReady && this.formElement?.nativeElement) {
+      const form = this.formElement.nativeElement;
```
FollowTheFlo, line 100: can we move const form inside the 'if' block so won't be created if not IFRAME?

FollowTheFlo (dismissed):
as discussed offline, looks good.

PR #20735, opened 2025-09-16
feat: update Angular minor version from 19.0.4 to 19.2.15

Platonn:
**👉 Context:**

Usually we don't update Angular minor version, because it means also increasing minimal peerDeps versions for customers. (Usually try to update Angular only in a yearly Framework Upgrade Release, where we update 2 _major_ versions at once)

This time we need to make an exception and update Angular minor in the middle of a year, because of a vulnerability found in Angular SSR: https://github.com/angular/angular/security/advisories/GHSA-68x2-mx4q-78m7 that is fixed only in version 19.2.15 (in the version line 19.x):

> The following APIs were vulnerable and required SSR-only breaking changes:
> 
>    - bootstrapApplication: (...) NgModule-based applications are not affected.
>    - getPlatform: (...)
>    - destroyPlatform: (...)

Even though we don't use any of those APIs (so we're likely unaffected by the Angular bug), we decided to upgrade Angular minor version just to stay on the safe side.

**👉 Consequences for customers:**

As as a side-effect we increase minimal peerDeps versions of Angular packages for customers.
Before this change, customers could use any 19.x version of Angular they wanted **but at least higher 19.0.4**. Now they are obliged to use any 19.x version of Angular **but at least higher 19.2.15**.

And we're upgrading TS version to 5.8. But customers need to use at least `~5.7`.

If customers must update their Angular minor version because of us, they shouldn't experience practical breaking changes from Angular, because new Angular minor version they install should contain only new features but not breaking changes (see the [official Angular docs on versioning](https://angular.dev/reference/releases#angular-versioning)). I hope the same holds for new version of TS.

**👉 Implementation notes:**

- the angular and typescript pacakges versions were updated by running `nx migrate` commands
- we no longer need to apply a patch in `@angular/build` for core team's DX in the "npm postinstall" script, so the script is removed: `"postinstall": "ts-node scripts/npm-postinstall/patch-@angular-build-19.0.4.ts"`
- we no longer need to forcefully install any certain version of `vite`  with npm `"overrrides"` in `package.json`. It's because the newest Angular minor 19.2.15 already depends on the nearly-newest version of `vite`.

fixes CXSPA-10974


FollowTheFlo (approved):
LGTM

PR #20756, opened 2025-09-23
fix: Enhance IntersectionService for SSR and browser platform detection

lasoh:
- Added server-side rendering (SSR) support to IntersectionService, allowing immediate intersection results without creating an IntersectionObserver.
- Implemented tests for both SSR and browser platforms to verify expected behavior in each context.
- Updated service constructor to inject PLATFORM_ID for platform detection.

1. SSR Environment Testing ssr
Steps:
1. Navigate to a page containing components that use IntersectionService (fe. MerchandisingCarousel) 
2. Check browser console for errors
3. Verify no ReferenceError: IntersectionObserver is not defined errors
4. Verify components render correctly in SSR
5. Expected Result: ✅ No console errors, components render successfully

### DIFF
--- a/projects/storefrontlib/layout/loading/intersection.service.spec.ts
+++ b/projects/storefrontlib/layout/loading/intersection.service.spec.ts
@@ -6,7 +6,6 @@ import { Observable, of } from 'rxjs';
 import { cold } from 'jasmine-marbles';
 import { PLATFORM_ID } from '@angular/core';
 
-
 const INTERSECTION_MARGIN_GENERAL = '5%';
 const INTERSECTION_MARGIN_SPECIFIC = '4%';
 
@@ -189,7 +188,6 @@ describe('IntersectionService', () => {
       ).toBe(INTERSECTION_MARGIN_SPECIFIC);
     });
   });
-
 });
 
 describe('IntersectionService SSR Platform Detection', () => {
@@ -200,7 +198,7 @@ describe('IntersectionService SSR Platform Detection', () => {
       TestBed.configureTestingModule({
         providers: [
           { provide: LayoutConfig, useValue: MOCK_LAYOUT_CONFIG },
-          { provide: PLATFORM_ID, useValue: 'server' }
+          { provide: PLATFORM_ID, useValue: 'server' },
         ],
       });
       service = TestBed.inject(IntersectionService);
@@ -209,7 +207,8 @@ describe('IntersectionService SSR Platform Detection', () => {
     it('should return true immediately for isIntersected in SSR', (done) => {
       const element: HTMLElement = document.createElement('section');
 
-      service.isIntersected(element)
+      service
+        .isIntersected(element)
         .pipe(take(1))
         .subscribe((isIntersected) => {
           expect(isIntersected).toBe(true);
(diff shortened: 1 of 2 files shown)


projects/storefrontlib/layout/loading/intersection.service.ts
```diff
@@ -23,7 +24,10 @@ export type IntersectingCondition = (
   providedIn: 'root',
 })
 export class IntersectionService {
-  constructor(protected config: LayoutConfig) {}
+  constructor(
+    protected config: LayoutConfig,
+    @Inject(PLATFORM_ID) protected platformId: Object
+  ) {}
```
pawelfras:
```suggestion
  protected platformId = inject(PLATFORM_ID);

  constructor(protected config: LayoutConfig) {}
```

projects/storefrontlib/layout/loading/intersection.service.ts
```diff
@@ -91,6 +95,20 @@ export class IntersectionService {
...
+          target: element,
+          isIntersecting: true,
+          intersectionRatio: 1,
+          boundingClientRect: {} as DOMRectReadOnly,
+          intersectionRect: {} as DOMRectReadOnly,
+          rootBounds: {} as DOMRectReadOnly,
+          time: Date.now(),
+        } as IntersectionObserverEntry,
+      ]);
+    }
```
pawelfras: `major:` in my opinion it is rather healing the symptoms instead of fixing the root cause. Instead of mocking the API not available in the Node.js, let's adjust the logic instead to not run/create intersection observer on a server side (it won't work anyway)

projects/storefrontlib/layout/loading/intersection.service.ts
```diff
@@ -44,6 +47,11 @@ export class IntersectionService {
...
     intersectingCondition?: IntersectingCondition
   ): Observable<boolean> {
+    // Early return for SSR - element is never intersected on server
+    if (isPlatformServer(this.platformId)) {
+      return of(false);
+    }
+
     return this.intersects(element, options, intersectingCondition).pipe(
       first((v) => v === true)
     );
```
pawelfras, line 57:
```suggestion
    return this.intersects(element, options, intersectingCondition).pipe(
      first()
    );
```

projects/storefrontlib/layout/loading/intersection.service.spec.ts
pawelfras, line 1: `major:` `done()` callback needs to be called inside the `subscribe`, otherwise the test case give false results.

projects/storefrontlib/layout/loading/intersection.service.ts
pawelfras: @lasoh Apologies, I have second thoughts on this topic, and I think we cannot change how the `first` operator is used here as it is a breaking change. for the logic which may cause unexpected behaviour on the browser platform, where the IntersectionObserver is used. Instead of that, let's add
```
if (isPlatformServer(this.platformId)) {
   return of(false);
}
```
only to `isIntersected` and `isIntersecting` method. This should solve the issue without providing breaking changes.

pawelfras (approved):
Review and QA done. LGTM

PR #20761, opened 2025-09-29
test: refine  e2e test

niehuayang:


projects/storefrontapp-e2e-cypress/cypress/e2e/regression/asm/asm.deeplink.core-e2e.cy.ts
```diff
@@ -64,8 +67,6 @@ context('Assisted Service Module', () => {
         ).then((customerId) => {
           cy.visit('/assisted-service/emulate?customerId=' + customerId);
 
-          agentLoginForJDK21(b2cAgent.userName, b2cAgent.password);
```
i53577, line 67: I am a little confused, don't we need it any more? how does the agent login without this?

niehuayang: agent already logged in the above 60th line, so no need to log in again

Melody-zhou-512 (approved):
LGTM.

PR #20770, opened 2025-10-01
feat: Product not in catalog

EmilJedrychowski-SAP:
Jira: https://jira.tools.sap/browse/CXSPA-10774


projects/core/src/product/facade/product-catalog.service.ts
```diff
@@ -0,0 +1,24 @@
...
+import { Injectable } from '@angular/core';
+import { Product } from '@spartacus/core';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class ProductCatalogService {
+  /**
+   * Returns whether product is present in commerce product catalog.
+   * Method is meant to be overridden by other integration specific implementations.
```
anjana-bl: NITPICK: you can remove this comment. - Method is meant to be overridden by other integration specific implementations.

projects/core/src/product/facade/product-catalog.service.spec.ts
```diff
@@ -0,0 +1,28 @@
...
+
+    service = TestBed.inject(ProductCatalogService);
+  });
+
+  it('should be created', () => {
+    expect(service).toBeTruthy();
+  });
+
+  it('should return true if product is provided', () => {
+    expect(service.isProductInCatalog(product)).toBeTruthy();
```
anjana-bl: NITPICK: can we use .toEqual(true) instead of .toBeTruthy() as we expect exactly the boolean value true, not just any truthy value.

feature-libs/cart/base/components/add-to-cart/add-to-cart.component.ts
```diff
@@ -128,6 +131,9 @@ export class AddToCartComponent implements OnInit, OnDestroy {
     if (this.product) {
       this.productCode = this.product.code ?? '';
       this.setStockInfo(this.product);
+      this.unavailable = !this.productCatalogService.isProductInCatalog(
+        this.product
+      );
     } else if (this.productCode) {
```
anjana-bl, line 137: MAJOR: in case control goes to `else if (this.productCode) {}`, then what will be the value for `unavailable`

feature-libs/cart/base/components/cart-shared/cart-item-list/cart-item-list.component.spec.ts
```diff
@@ -117,6 +119,11 @@ class MockFeatureConfigService {
   }
 }
 
+const mockProductCatalogService = {
+  // @ts-ignore
```
anjana-bl: NITPICK: any reason why @ts-ignore

EmilJedrychowski-SAP: isProductInCatalog() takes product as an argument. I do not use it anywhere inside this function. This is not allowed and would break 'cart' unit tests without ts-ignore

projects/core/src/product/facade/product-catalog.service.ts
EmilJedrychowski-SAP: Removed

projects/core/src/product/facade/product-catalog.service.spec.ts
EmilJedrychowski-SAP: Replaced

feature-libs/cart/base/components/add-to-cart/add-to-cart.component.ts
EmilJedrychowski-SAP: `unavailable: boolean = false;`
By default flag 'unavailable' is set to false.
If product is not provided to this component there is no way of checking product 'in catalog' availability so it cannot be marked as unavailable.

feature-libs/cart/base/components/cart-shared/cart-item-list/cart-item-list.component.spec.ts
Pucek9: Please use "_" prefix in those cases, which allows for unused parameters

feature-libs/cart/base/components/cart-shared/cart-item-list/cart-item-list.component.spec.ts
```diff
@@ -117,6 +119,11 @@ class MockFeatureConfigService {
...
+  isProductInCatalog: (product?: Product) => true,
```
Pucek9:
```suggestion
  isProductInCatalog: (_product?: Product) => true,
```
etc.

projects/core/src/product/facade/index.ts
```diff
@@ -12,3 +12,4 @@ export * from './product-search-by-category.service';
 export * from './product.service';
 export * from './searchbox.service';
 export * from './product-availability.service';
+export * from './product-catalog.service';
```
Pucek9: I don't think it should be placed here. That's ordinary service, not facade. 
I guess `projects/core/src/product/services` would be more relevant.

feature-libs/cart/base/components/cart-shared/cart-item-list/cart-item-list.component.spec.ts
```diff
@@ -399,6 +410,18 @@ describe('CartItemListComponent', () => {
         })
         .unsubscribe();
     });
+
+    it('should disable item link only if product is not in catalog', () => {
+      component.options = mockContext.options;
+      expect(component.getOptions(mockItem0)).toBe(component.options);
+      // @ts-ignore
+      mockProductCatalogService.isProductInCatalog = (product?: Product) =>
```
Pucek9: That's interesting, it works so simply. In most cases we use `spyOn(...).and.returnValue(...)`, `overrideProvider` etc. I don't remember why, maybe you know?

EmilJedrychowski-SAP: This way is more simple but without spy you cannot check if service method was called, how many times it was called, with what arguments it was called etc.

projects/core/src/product/facade/index.ts
EmilJedrychowski-SAP: Moved

feature-libs/cart/base/components/cart-shared/cart-item-list/cart-item-list.component.spec.ts
EmilJedrychowski-SAP: Didn't know, thanks 👍 
Replaced in each placed I used ts-ignore before

feature-libs/cart/base/components/cart-shared/cart-item-list/cart-item-list.component.spec.ts
EmilJedrychowski-SAP: as per @Pucek9 suggestion I replaced argument name so it starts with '_' that way I could drop ts-ignore

Pucek9 (dismissed):
Approved by the Core Team member. Please note the PR also needs to be tested and approved by one of your teammates

projects/core/src/product/services/product-catalog.service.spec.ts
Pucek9, line 1: please move spec along with service

Pucek9 (dismissed):

Pucek9 (approved):

PR #20772, opened 2025-10-01
fix: fix the issue with Quantity To Return is not being visible

uroslates:
Fixes: [CXSPA-11068](https://jira.tools.sap/browse/CXSPA-11068)


feature-libs/order/components/amend-order/returns/return-order/return-order.component.spec.ts
```diff
@@ -116,16 +139,140 @@ describe('ReturnOrderComponent', () => {
...
     expect(result[1].returnableQuantity).toBe(3);
   });
+
+  it('should initialize form$ and set orderCode', () => {
+    let formValue: UntypedFormGroup = new UntypedFormGroup({});
+    component.form$.subscribe((form) => (formValue = form)).unsubscribe();
+
+    expect(formValue.value.orderCode).toEqual('123');
+    expect(component.orderCode).toEqual('123');
+  });
```
psmul: Since subscription operations are asynchronous, we want to ensure that our tests are as deterministic as possible, and we don't leave any vulnerabilities for bad timing to retrieve the data. We want to ensure that retrieved data is checked properly and we need signal test runner when test case is done, hence I propose to refactor all provided tests to this approach:

```
it('should initialize form$ and set orderCode', (done) => {
  component.form$.pipe(take(1)).subscribe((form: UntypedFormGroup) => {
    expect(form.value.orderCode).toEqual('123');
    expect(component.orderCode).toEqual('123');
    done();
  })
});
```

This way we're passing async callback "done" to let the test runner that it should finish async operations at this step. To be more align with Spartacus test patterns and RxJS practices, it's better to use take(1) as we're sure we're emitting value here, which will cause automatic subscription closing. Alternatively, you can use RxJS method - `await firstValueFrom(component.form$)`.

psmul (changes requested):

feature-libs/order/components/amend-order/returns/return-order/return-order.component.spec.ts
uroslates: Thanks for your feedback @psmul 
Addressed! 👍

psmul (dismissed):
LGTM 👍🏼

uroslates:
@psmul I have hidden everything behind feature toggle (to be on the safe side). Could you please once more approve (if all OK)? Thanks upfront.

feature-libs/order/components/amend-order/returns/return-order/return-order.component.ts
```diff
@@ -47,7 +51,14 @@ export class ReturnOrderComponent {
           return consignmentEntry
             ? {
                 ...entry,
-                returnableQuantity: consignmentEntry.shippedQuantity ?? 0,
+                returnableQuantity:
+                  consignmentEntry.shippedQuantity ??
+                  (this.featureConfigService?.isEnabled(
+                    'enableReturnOrderReturnableQuantityDisplayFix'
```
psmul: I'm not sure if we want to include "Fix" as part of feature toggle. Let's re-think naming for this toggle. 

Additionally, please provide a separate unit test to check if toggle works.

psmul (changes requested):

psmul (approved):

PR #20791, opened 2025-10-10
fix: Delay in subscription details loading on route change

anjana-bl:
Issue: Currently, when navigating between subscription details, the component initially shows stale data from the previously viewed subscription before updating to the correct subscription.

JIRA: https://jira.tools.sap/browse/CXSPA-9994


feature-libs/subscription-billing/core/facade/subscription-billing.service.ts
```diff
@@ -113,7 +113,7 @@ export class SubscriptionBillingService implements SubscriptionBillingFacade {
       this.userIdService.getUserId(),
       this.getSubscriptionCodeFromRoute(),
     ]).pipe(
-      take(1),
+      //take(1),
```
ercultimate: Remove commented code

ercultimate (approved):

PR #20804, opened 2025-10-16
[PARKED] feat: (CXSPA-11133) enable useWithCredentials by default for occ config

psmul:


projects/storefrontapp/src/app/spartacus/spartacus-b2c-configuration.module.ts
```diff
@@ -57,6 +58,7 @@ const baseSite = environment.epdVisualization
         },
       },
     }),
+    provideConfigFactory(defaultOccConfigFactory),
```
Platonn: This file belongs to an app, but not lib. And we'd like to make a behavioral change implicitly.
In other words, we don't want to ask customers to put the `provideConfigFactory(defaultOccConfigFactory),` in their app's code, in order for the feature toggle to be effective.

Note: the same applies to the projects/storefrontapp/src/app/spartacus/spartacus-b2b-configuration.module.ts

How we'd like to do it:
- in BaseOccModule [we have](https://github.com/SAP/spartacus/blob/3d43cba2ffc0ac4c06a9e3fa083c64262d9e2d66/projects/core/src/occ/base-occ.module.ts#L30) `provideDefaultConfig(defaultOccConfig),`
- let's replace that line with `provideDefaultConfigFactory(defaultOccConfigFactory),`

projects/core/src/occ/config/default-occ-config-factory.ts
```diff
@@ -0,0 +1,22 @@
...
+import { FeatureToggles } from '@spartacus/core';
+import { defaultOccConfig } from './default-occ-config';
+import { OccConfig } from './occ-config';
+
+export function defaultOccConfigFactory(): OccConfig {
+  const featureToggles = inject(FeatureToggles);
+  const config = { ...defaultOccConfig };
+
+  if (featureToggles.enableWithCredentialsByDefault) {
+    if (config.backend && config.backend.occ) {
```
Platonn: [nitpick]
```suggestion
    if (config.backend?.occ) {
```

projects/core/src/features-config/feature-toggles/config/feature-toggles.ts
```diff
@@ -669,6 +669,14 @@ export interface FeatureTogglesInterface {
...
    */
   enableReturnOrderReturnableQuantityConsigmentFallback?: boolean;
+  /**
+   * When enabled, all cross-site requests will include credentials (cookies, HTTP authentication entries, and client-side SSL certificates).
+   * This is needed to support scenarios where the backend services are hosted on a different domain than the Spartacus storefront,
+   * and the authentication cookies need to be sent with the requests to maintain the user session.
+   *
+   * Affects: `WithCredentialsInterceptor`
+   */
+  enableWithCredentialsByDefault?: boolean;
```
Platonn: [nitpick]
This change affects only calls to OCC.
let's reflect it in the toggle name, e.g.:
`occWithCredentialsByDefault`

PS. I was never a fan of using the word `enable` in the toggles names. They should describe the change in behavior and that's it. By setting it to `true`, we just _enable_ it 😅

projects/core/src/features-config/feature-toggles/config/feature-toggles.ts
```diff
@@ -759,4 +767,5 @@ export const defaultFeatureToggles: Required<FeatureTogglesInterface> = {
   navigationMenuCloseOnSameLinkClick: false,
   enableQuotePurchaseOrderNumber: false,
   enableReturnOrderReturnableQuantityConsigmentFallback: false,
+  enableWithCredentialsByDefault: true,
```
Platonn: > Add a new property, write JSDocs and **set the default value false** in the core lib there:  projects/core/src/features-config/feature-toggles/config/feature-toggles.ts  (for customers)

source: https://wiki.one.int.sap/wiki/display/spar/Coding+Guidelines#CodingGuidelines-AddaFeatureToggleinthecodesourcecode

We'll enable it by default only in 6 months from now. And after 12 months we'll remove it entirely. It's our default policy for the feature toggles lifecycle.

projects/core/src/features-config/feature-toggles/config/feature-toggles.ts
psmul: Yea, fully agreed with this suggestion, thanks 👍🏼

PR #20808, opened 2025-10-17
fix: CXSPA-10817: Enhance asm vendor's case for JDK21

LindaSAP:
1. Remove accessibility test in vendor folder
2. Enhance asm vendor's case for JDK21
Screenshot 2025-10-17 at 17 13 23
Screenshot 2025-10-17 at 17 15 31


niehuayang (dismissed):

niehuayang (approved):

projects/storefrontapp-e2e-cypress/cypress/e2e/vendor/asm/customer360.e2e.cy.ts
```diff
@@ -18,6 +18,9 @@ import {
 context('Assisted Service Module', () => {
   before(() => {
     clearAllStorage();
+  });
+
+  it('should prapare customer cart/order/interest', () => {
     asmCustomer360.setup();
```
i53577, line 24: I am a little confused, why extract line24-25 into a new test? do not need it in `before` anymore? will this change test result?

i53577: seems `before` in line 19 only run once in the whole context. Then it should make no difference.

PR #20810, opened 2025-10-20
fix: add iframe breakout for the OPF verification component

Matejk00:
Closes: CXSPA-9807


integration-libs/opf/payment/root/components/opf-payment-verification/opf-payment-verification.component.ts
```diff
@@ -63,6 +66,24 @@ export class OpfPaymentVerificationComponent implements OnInit, OnDestroy {
...
   }
 
+  /**
+   * Ensures the verification page is displayed in the top-level window.
+   * Some PSPs render verification/success pages inside an iframe; in such cases
+   * we reload the current URL in the top window to preserve full-page UX.
+   *
+   * Reference: https://docs.saferpay.com/home/integration-guide/general-information/iframe-integration-and-css
+   */
+  protected breakOutOfIframeIfNeeded(): void {
```
FollowTheFlo, line 76: copilot suggest to add a try-catch in case we hit cross origin, would you know if it is necessary or overengineered?
```
protected breakOutOfIframeIfNeeded(): void {
  const win = this.winRef?.nativeWindow;
  if (!win) return;

  try {
    if (win.top && win.top !== win && win.top.location !== win.location) {
      win.top.location.href = win.document.location.href;
    }
  } catch (e) {
    // Cross-origin access error; do nothing or log if needed
  }
}
```

Matejk00: From my point of view, this feels a bit overengineered, since we’re targeting the verification route URL, which will always be on the same domain as the checkout. So I’m not sure there’s a need to handle cross-origin in this case.

FollowTheFlo: ok, make sens, good to me.

FollowTheFlo (approved):
lgtm

PR #20816, opened 2025-10-22
fix: call waitForPage before visit

IustinSER:
CXSPA-11272


projects/storefrontapp-e2e-cypress/cypress/helpers/login.ts
FollowTheFlo, line 1: I checked the occurences of registerUserFromLoginPage() method, it is used 3 times:
early-login
logic.core
navigation-login-a11y
Please remove the snippet:
```
   cy.whenJDK21(() => {
      cy.visit('/login/register');
    });
```
   
in early-login and navigation-login-a11y to avoid visiting login/register page twice, even though it does not make the test failing.

PR #20850, opened 2025-11-04
fix: Fix focus lost in anonymous consent dialog when selecting input

Zeyber:
Closes: https://jira.tools.sap/browse/CXSPA-11392


rmch91 (approved):
LGTM

PR #20855, opened 2025-11-06
feat: Add new builder addressing ng-packagr issues

pawelfras:
This PR provides a temporary workaround for [ng-packagr issue](https://github.com/ng-packagr/ng-packagr/issues/3085) introduced in v20.
The ng-packagr library introduces new functionality: [Flatten type definitions to a single *.d.ts file](https://github.com/ng-packagr/ng-packagr/issues/139#top)
Due to a bug in its dependency, the `rollup-plugin-dts` library, flattened `.d.ts` file still uses relative paths in declaration merging, but after flattening, there are no other directories in the built libs, where types are located.

BEFORE:
image

AFTER:
image

Note: It is just a workaround and hopefully the issue is going to be fixed in the root cause. 
Apart from that, there might be a case the builder was used in libraries which does not require it - let's check it later, to unblock the build.

closes sub-task: [CXSPA-11393](https://jira.tools.sap/browse/CXSPA-11393)

### DIFF
--- a/feature-libs/cart/project.json
+++ b/feature-libs/cart/project.json
@@ -6,7 +6,7 @@
   "prefix": "cx",
   "targets": {
     "build": {
-      "executor": "./tools/build-lib:augmented-types",
+      "executor": "./tools/build-lib:declaration-merging",
       "options": {
         "tsConfig": "feature-libs/cart/tsconfig.lib.json",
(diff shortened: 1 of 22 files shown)


sdrozdsap (approved):
Piece of art ✅

PR #20862, opened 2025-11-12
fix: Workaround for 'ariaLabelDropdown' input signal

pawelfras:
- in ng-select@20 ariaLabelDropdown has been changed to [input signal](https://github.com/ng-select/ng-select/pull/2656/files#diff-4e0502e4dc19d676826005eb8878a23f9daec09195ecd896a44ebf3772f78a63)
- used SIGNAL symbol to get access to change value of the input signal
- known problem: https://github.com/angular/angular/issues/54782


sdrozdsap (approved):
Nice! Thanks for explanation in the comments

PR #20866, opened 2025-11-12
fix: adapt asm test for JDK21

IustinSER:
https://jira.tools.sap/browse/CXSPA-11416


projects/storefrontapp-e2e-cypress/cypress/e2e/a11y/b2c/asm.a11y-e2e.cy.ts
```diff
@@ -17,9 +19,16 @@ describe('ASM Continuum tests', { testIsolation: false }, () => {
...
-    cy.get('input').first().type('brandon.leclair@acme.com');
-    cy.get('input[type="password"').type('pw4all{enter}');
-    cy.get('.searchLabel');
+    cy.whenJDK21(() => {
+      cy.get('.cx-asm-customer-list .cx-asm-customer-list-link').click();
+      agentLoginForJDK21('brandon.leclair@acme.com', 'pw4all');
+    });
+
+    cy.whenJDK17(() => {
+      asm.agentLogin('brandon.leclair@acme.com', 'pw4all');
```
FollowTheFlo: I think it would fail wehn using 'asm.' as `import * as asm from '../../../../helpers/asm';` as not been imported.
See other files using asm.agentlogin() as references.

FollowTheFlo: I would advice to use previous code for JDK17 as issue is only about JDK21.

FollowTheFlo (approved):
good to me.

PR #20870, opened 2025-11-13
fix: add debouncing and fix stale results in onEnter handler

darvsorceix:
Add search input debouncing to reduce canceled requests and fix onEnter to properly wait for search completion using searchCompleted observable


projects/storefrontlib/cms-components/navigation/search-box/search-box.component.ts
```diff
@@ -149,6 +152,9 @@ export class SearchBoxComponent implements OnInit, OnDestroy {
 
   protected subscriptions = new Subscription();
 
+  // Subject for debounced search input
+  private searchInput$ = new Subject<string>();
```
Pucek9:
```suggestion
  prrotected searchInput$ = new Subject<string>();
```
Please see https://wiki.one.int.sap/wiki/spaces/spar/pages/3564966608/Coding+Guidelines#CodingGuidelines-Usepublic%2Fprotectedinsteadofprivateinclasses

projects/storefrontlib/cms-components/navigation/search-box/search-box.component.ts
```diff
@@ -245,15 +271,26 @@ export class SearchBoxComponent implements OnInit, OnDestroy {
...
     this.checkOuterResults();
     // force the searchBox to open
     this.open();
   }
 
+  /**
+   * Performs the actual search operation.
+   * This is called after debouncing to reduce HTTP requests.
+   */
+  private performSearch(query: string): void {
```
Pucek9:
```suggestion
  protected performSearch(query: string): void {
```

Pucek9:
@darvsorceix, please provide QA steps so I can verify the behavior change

c5346163:
> @darvsorceix, please provide QA steps so I can verify the behavior change

image

You need to check the Network tab in DevTools. If you type quickly, there used to be a lot of canceled requests, but that should be mitigated now (screenshot from the develop branch).

Pucek9:
> image
> You need to check the Network tab in DevTools. If you type quickly, there used to be a lot of canceled requests, but that should be mitigated now (screenshot from the develop branch).

@darvsorceix It works much better, previously it happened for every sign, but still not ideal (it depends on the writing pace)
image
image

PR #20900, opened 2025-11-24
feat: CXSPA-11294 - add SSL option to install script for CSR mode

Uli-Tiger:
Adds the option to provide a path to an SSL CERT/Key combo in order to start start spartacus in SSL mode.


rmch91 (approved):
LGTM

PR #20901, opened 2025-11-24
chore: remove deprecated epd-visualization integration library

sam-garland:
CXSPA-11549


projects/storefrontapp/tsconfig.app.json
```diff
@@ -5,7 +5,7 @@
     "strict": false, // TODO: Fix strict errors and remove this option
     "target": "ES2022",
     "module": "ES2022",
-    "skipLibCheck": true // Needed due to @sapui5/ts-types-esm dependency (used by @spartacus/epd-visualization)
+    "skipLibCheck": true
```
theta682, line 8: Probably
```suggestion
    "module": "ES2022"
```

integration-libs/opf/tsconfig.lib.json
```diff
@@ -12,7 +12,7 @@
     "experimentalDecorators": true,
     "allowSyntheticDefaultImports": true,
     "importHelpers": true,
-    "skipLibCheck": true, // Needed due to @sapui5/ts-types-esm
+    "skipLibCheck": true,
```
theta682, line 15:
```suggestion
```

sam-garland: Someone in the Spartacus team can decide whether`skipLibCheck": true,` should be kept.

The code compiles without it now, but:
- `.d.ts` files would normally be used for third party JavaScript libraries (meaning the .`d.ts` files are not under the control of the Spartacus team)
- There is a performance benefit to keeping `skipLibCheck": true` (though this will be reduced with the UI5 types removed).

PR #20910, opened 2025-11-26
fix: add login page missing translations

uroslates:
Closes: [CXSPA-10923](https://jira.tools.sap/browse/CXSPA-10923).


uroslates:
This PR introduces translations for Language & Currency selectors Labels (which were hard-coded in English before.
The feature is not hidden behind PR since its modifying local (non-public) [LABEL](https://github.com/SAP/spartacus/pull/20910/files#diff-731fa7cb693e6e09bcd073b94c91303161d7a1442bfc21ef75e47c2b2e7c0db8R22) constant values within [SiteContextComponentService](https://github.com/SAP/spartacus/blob/a08b780872f67a5a41be25ac897407149dcccc5d/projects/storefrontlib/cms-components/misc/site-context-selector/site-context-component.service.ts#L28). But I leave it to reviewer to say if they thing it needs to be feature-toggle driven.

projects/assets/src/translations/en/common.json
```diff
@@ -35,7 +35,9 @@
     "expand": "Expand",
     "collapse": "Collapse",
     "readMore": "Read More",
-    "readLess": "Read Less"
+    "readLess": "Read Less",
+    "language": "Language",
+    "currency": "Currency"
```
FollowTheFlo, line 40: I tested ands seeing 'Theme' is not translated, we can create a separate jira for it.

FollowTheFlo:
To know if it requires feature flag or introducing breaking change, i went through list:
breaking change:
Removing or renaming public APIs, services, components, or methods.
Changing the signature of public methods, services, or components.
Modifying or removing configuration options or default values.
Changing the structure of data models, interfaces, or expected input/output.
Altering the behavior of core features in a way that is not backward compatible.
Removing or renaming CSS classes or selectors used in templates.
Changing routes, route parameters, or route guards in a way that breaks navigation.

feature toggle:
Introduce a new feature gradually, allowing it to be enabled or disabled without code changes.
Test or preview features with selected users before a full rollout.
Maintain backward compatibility while refactoring or replacing existing functionality.
Control experimental or beta features in production environments.
Allow customers or integrators to opt-in or opt-out of specific features.

imho we don't hit any of those condition with this fix so no featureToggle needed.

FollowTheFlo (approved):
tested, assessed if feature toggle was needed. lgtm

uroslates:
**DO NOT MERGE BEFORE FEATURE FREEZE IS OVER (post February 2026.)**

wangzixi-diablo:
Dear @uroslates 

Thanks a lot for your time and effort on this i18n issue!

I am running with latest code in develop branch locally in just the same way as you did on 2025 November. When I switch language from English to Chinese, the page title is still displayed as "Login". It should be "登录" in Chinese.

Would you please kindly check again? Thanks a lot!
language

PR #20911, opened 2025-11-27
fix: noise in logs about not-found static files

Pucek9:


projects/storefrontapp/src/server.ts
```diff
@@ -68,6 +69,7 @@ export function app(): express.Express {
   });
 
   server.use(defaultExpressErrorHandlers(indexHtmlContent));
+  server.use(getChromeDevtoolsExpressMiddleware());
```
Platonn: [MEDIUM]
1. Unfortunately now it doesn't work as expected. Let's move it above line 64. Otherwise the `*` route matches and our new middleware doesn't kick in.
2. If possible, let's also add this for our customers **in fresh-generated apps** via the installation schematics in https://github.com/SAP/spartacus/blob/develop/projects/schematics/src/add-ssr/files/server.__typescriptExt__#L46-L48 + let's fix schematics unit tests then (i.e. the snapshots of the generated server.ts file)
3. I suggest to tell our customers in the Major release migration docs to also add this line to their **existing apps.**
   - so let's change the target branch of this PR to `develop-next-major`
   - please ask to Pawel and Siarhei, where in the migration docs, we could add a note for existing customers to add this new middleware in their `server.ts` files

Pucek9: @Platonn
1 Thanks, my bad!
2 Are you sure we want to expose this solution to our customers?
- It's generally a workaround to avoid noise in SSR logs, it doesn't add static configuration for Chrome DevTools (at least it's not provided in CSR mode), so those middleware will possibly override customers' configuration and do that only in SSR mode. 
- Introducing it causes us should take the responsibility of maintaining of handling 3rd part plugins, which can be simply done by clients, only if they want it.
- There will be a plugin for [vite](https://github.com/ChromeDevTools/vite-plugin-devtools-json/) available for us soon, do you think it is worth spending more effort now on introducing a temporary solution?

PS. Additionally, I can see those warnings appear also in prod mode, which was not expected. What should we do?

Platonn: > There will be a plugin for [vite](https://github.com/ChromeDevTools/vite-plugin-devtools-json/) available for us soon

I believe we won't add any custom Vite plugins. But instead we'll first switch to Angular modern SSR, which uses Angular Vite SSR dev server, which under the hood has the [OOTB logic for handling the chrome dev tools .well-known route](https://github.com/angular/angular-cli/blob/23ff4e98a480a788985f6de4ff1204f0c938fd65/packages/angular/build/src/tools/vite/middlewares/chrome-devtools-middleware.ts#L22-L65).

> Introducing it causes us should take the responsibility of maintaining of handling 3rd part plugins, which can be simply done by clients, only if they want it.

Thank you for taking care of our Public API and the perspective of the long-term maintenance!
The issue with .well-known ChromeDevTools route was frustrating since a year already. And I cannot promise when we'll adopt the new Angular SSR. So it's a solution at least for few months I guess (perhaps more), which brings a value for all those months. Eventually when we'll switch to modern Angular SSR, we can deprecate our middleware

> It's generally a workaround to avoid noise in SSR logs, it doesn't add static configuration for Chrome DevTools

I'm having problem to understand that. And perhaps I'm missing something important that you're aware of. I'd be happy to have a quick call with you explaining me why you perceive it as a workaround but not a solution.

Pucek9:
Moved to https://github.com/SAP/spartacus/pull/21025

PR #20914, opened 2025-11-27
chore: Angular 20 upgrade

sdrozdsap:
## Angular 20 Upgrade

**Nx migration** (all packages and their migrations) ran `npx nx migrate <package>@<version>` and `npx nx migrate --run-migrations` (run packages' migrations, all except optional ones like control-flow-migration).

**Bumped packages:**
- `@angular/core` → `^20.3.9`
- `@angular/animations` → `^20.3.9`
- `@angular/common` → `^20.3.9`
- `@angular/compiler` → `^20.3.9`
- `@angular/forms` → `^20.3.9`
- `@angular/platform-browser` → `^20.3.9`
- `@angular/platform-browser-dynamic` → `^20.3.9`
- `@angular/platform-server` → `^20.3.9`
- `@angular/router` → `^20.3.9`
- `@angular/service-worker` → `^20.3.9`
- `@angular/cli` → `^20.3.9`
- `@angular/compiler-cli` → `^20.3.9`
- `@angular/language-service` → `^20.3.9`
- `@angular/ssr` → `^20.3.9`
- `@angular/pwa` → `^20.3.9`
- `@angular/build` → `^20.3.9` (new package)
- `@angular-devkit/build-angular` → `^20.3.9`
- `@angular-devkit/core` → `^20.3.9`
- `@angular-devkit/schematics` → `^20.3.9`
- `@schematics/angular` → `^20.3.9`
- `@ngrx/effects` → `^20.1.0`
- `@ngrx/operators` → `^20.1.0`
- `@ngrx/router-store` → `^20.1.0`
- `@ngrx/store` → `^20.1.0`
- `@ngrx/store-devtools` → `^20.1.0`
- `@angular-eslint/builder` → `^20.7.0`
- `@angular-eslint/eslint-plugin` → `^20.7.0`
- `@angular-eslint/eslint-plugin-template` → `^20.7.0`
- `@angular-eslint/schematics` → `^20.7.0`
- `@angular-eslint/template-parser` → `^20.7.0`
- `@angular-eslint/test-utils` → `^20.7.0`
- `@angular-eslint/utils` → `^20.7.0`
- `@ng-select/ng-select` → `^20.4.3`
- `angular-oauth2-oidc` → `^20.0.2`
- `ngx-infinite-scroll` → `^20.0.0`
- `@angular-builders/custom-esbuild` → `^20.0.0`
- `ng-packagr` → `^20.3.2`
- `jest-preset-angular` → `^15.0.3`
- `jest` → `^30.2.0`
- `jest-circus` → `^30.2.0`
- `jest-environment-jsdom` → `^30.2.0` (new package)
- `jest-environment-node` → `^30.2.0`
- `jest-util` → `^30.2.0`
- `ts-jest` → `^29.4.5`
- `@nx/angular` → `^22.1.1`
- `@nx/devkit` → `^22.1.1`
- `@nx/eslint-plugin` → `^22.1.1`
- `@nx/jest` → `^22.1.1`
- `@nx/workspace` → `^22.1.1`
- `nx` → `^22.1.1`
- `zone.js` → `^0.15.1`
- `typescript` → `^5.9.3`
- `express` → `^5.1.0`
- `eslint` → `^9.28.0`
- `@types/node` → `^22.19.1`
- `glob` → `^11.1.0`
- Removed `eslint-plugin-deprecation` (incompatible with ESLint 9)

**Unit tests:**
- Refactored tests to avoid relying on `ng-reflect-*` attributes (removed in Angular 20 dev mode by default)
- Updated snapshots containing Angular version references
- Upgraded Jest and related packages to v30
- Added `jest-environment-jsdom` package
- Applied changes due to Jest API changes
- Fixed tests across multiple libraries (organization, quote, product, cds, pickup-in-store, product-configurator, storefrontlib, storefinder)

**Build issues:**
- Resolved router guards typing issues
- [fix: Adjust type in private wrapCmsGuard method in CmsRoutesImplService (#20863)](https://github.com/SAP/spartacus/pull/20863)
- [fix: Adjust templates using `<ng-select>` component to satisfy 'items' input (#20864)](https://github.com/SAP/spartacus/pull/20864)
- [fix: Workaround for 'ariaLabelDropdown' input signal (#20862)](https://github.com/SAP/spartacus/pull/20862)
- Use `outputToObservable` to align with signal outputs used in new ng-select
- [feat: Add new builder addressing ng-packagr issues (#20855)](https://github.com/SAP/spartacus/pull/20855)
- Fixed ESLint configuration (disabled prefer-inject rule)

**Schematics:**
- [fix: CXSPA-11164 Fix schematics (#20895)](https://github.com/SAP/spartacus/pull/20895)
- [fix: CXSPA-11400 Fix schematics unit tests (#20896)](https://github.com/SAP/spartacus/pull/20896)
- [fix: CXSPA-11548 Installation script (#20909)](https://github.com/SAP/spartacus/pull/20909)


feature-libs/cart/project.json
```diff
@@ -6,7 +6,7 @@
   "prefix": "cx",
   "targets": {
     "build": {
-      "executor": "./tools/build-lib:augmented-types",
+      "executor": "./tools/build-lib:declaration-merging",
```
pawelfras, line 9: `nitpick:` Recently was released the [new version](https://github.com/Swatinem/rollup-plugin-dts/commit/771110cad6fa8cfc84b9ccd71c29c21095aedfc8) of `rollup-plugin-dts` and it contains a fox for the bug because ow which we introduced new custom builder. The 6.3.0 version is in range of versions supported by the ng-packagr.
Let's update version of this peer dependency and verify if we can use `ng-packagr` builder directly. 
Let's do it in a separate PR - leaving comment just for reference.

.github/actions/setup-node-and-cache/action.yml
```diff
@@ -24,8 +24,8 @@ runs:
           node_modules
           projects/storefrontapp-e2e-cypress/node_modules
           ~/.cache/Cypress
-        key: nodemodules-${{ inputs.cache-key }}
-        restore-keys: nodemodules-${{ inputs.cache-key }}
+        key: nodemodules-${{ inputs.cache-key }}-${{ hashFiles('**/package-lock.json') }}
```
pawelfras, line 27: `query:` Could you help me understand what was the exact reason of changing that? I'm wondering how was it working before vs now and why it was not needed in previous version.

.github/actions/setup-node-and-cache/action.yml
```diff
@@ -24,8 +24,8 @@ runs:
...
+        restore-keys: nodemodules-${{ inputs.cache-key }}-
```
pawelfras, line 28: `query:` is the `-` sign at the end by purpose?

feature-libs/product-configurator/rulebased/components/attribute/types/single-selection-image/configurator-attribute-single-selection-image.component.spec.ts
```diff
@@ -188,18 +192,17 @@ describe('ConfiguratorAttributeSingleSelectionImageComponent', () => {
...
     fixture.detectChanges();
-    CommonConfiguratorTestUtilsService.expectElementPresent(
-      expect,
-      htmlElem,
-      "cx-icon[ng-reflect-type='INFO']"
+    const icons = fixture.debugElement.queryAll(By.css('cx-icon'));
+    const infoIcon = icons.find(
+      (icon) => icon.componentInstance.type === ICON_TYPE.INFO
     );
+    expect(infoIcon).toBeTruthy();
```
pawelfras, line 199: `query:` out of curiosity, what was the reason that only this test case stopped working with `CommonConfiguratorTestUtilsService` ?

feature-libs/organization/administration/components/shared/card/card.component.spec.ts
```diff
@@ -142,8 +142,8 @@ describe('CardComponent', () => {
 
     it('should display hint after click info button', () => {
       const infoButton = fixture.debugElement.query(
-        By.css('button[ng-reflect-cx-popover]')
-      ).nativeElement;
+        By.css('.title h3 button cx-icon')
```
pawelfras, line 145: `query:` could you help me understand why the change was required?

I noticed there are more similar changes in files with unit tests and I assume the reason was the same for all of them.

feature-libs/quote/components/list/quote-list.component.ts
```diff
@@ -34,7 +34,7 @@ export class QuoteListComponent implements OnInit {
   sorts = this.quoteListComponentService.sortOptions;
   sortLabels$ = this.quoteListComponentService.sortLabels$;
   quotesState$ = this.quoteListComponentService.quotesState$;
-  dateFormat: string = 'MMMM d, YYYY h:mm aa';
+  dateFormat: string = 'MMMM d, yyyy h:mm aa';
```
pawelfras, line 37: `query:` Could you help me understand the reason behind changing the date format in a commit with fixes for unit tests? If the format changed, would be required to confirm if the new on fits business needs.

feature-libs/storefinder/components/store-finder-list-item/store-finder-list-item.component.spec.ts
```diff
@@ -143,8 +143,7 @@ describe('StoreFinderListItemComponent', () => {
     const encodedName = name.replace(' ', '%20');
     const link = fixture.debugElement
       .queryAll(By.css('.cx-store-name'))
-      .find((el) => el.nativeElement.innerText === displayName).nativeElement;
+      .find((el) => el.nativeElement.innerText === displayName)?.nativeElement;
     expect(link.getAttribute('href')).toEqual(`/${encodedName}`);
-    expect(link.getAttribute('ng-reflect-router-link')).toEqual(name);
```
pawelfras, line 148: `query:` does the `ng-reflect-router-link` not exist anymore?

integration-libs/cds/src/merchandising/cms-components/directives/attributes/attributes.directive.spec.ts
```diff
@@ -39,9 +39,7 @@ describe('cxAttributes directive', () => {
   });
 
   it('should not add any custom attributes if none are provided', () => {
-    const expectedAttributes = {
-      'ng-reflect-cx-attributes-name-prefix': attributeNamePrefix,
```
pawelfras, line 43: `query:` I see another change in tests related to `ng-reflect` attributes. Are they all due to the same reason?

.eslintrc.json
```diff
@@ -30,6 +29,7 @@
         "prefer-const": "off",
         "@angular-eslint/use-lifecycle-interface": "error",
         "@angular-eslint/prefer-standalone": "off",
+        "@angular-eslint/prefer-inject": "off",
```
pawelfras, line 32: `query:` Is this rule now enabled by default?
In general, we'd like to use `inject()` in a new code, but I'm aware it might be difficult/impossible to keep both - "error" for new code and "off" for the existing one.
Leaving this comment just for a reference, no action required.

.eslintrc.json
```diff
@@ -69,6 +69,7 @@
             "accessibility": "explicit"
           }
         ],
+        "@typescript-eslint/no-deprecated": "warn",
```
pawelfras, line 72: `query:` If I'm not mistaken, you mentioned that in a sense, this rule is more strict. Were any other code changes required as a result of this change?

migrations.json
pawelfras, line 1: `minor:` W could remove this file because it is no longer needed. If any migrations pops up during Angular 21 upgrade, new file will be generated.

feature-libs/cart/project.json
sdrozdsap: I've created this [Jira ticket](https://jira.tools.sap/browse/CXSPA-11591) for tracking any post-merge tasks and included this comment in the description

.github/actions/setup-node-and-cache/action.yml
sdrozdsap: Previously the cache key didn’t include the `package-lock.json` hash, so for a given `cache-key` (in our case PR base SHA) we always restored the same `node_modules` even when dependencies changed. That caused stale installs and skipped `npm install`, whenever we would adjust packages in the scope of same PR. I found out about it because I had an issue in the build job, that TS couldn't find `DOCUMENT` exported from `@angular/core`, because it started exporting it in v20, however old version of angular was being used

By adding `hashFiles('**/package-lock.json')`, the cache now invalidates correctly whenever dependencies change. And with this my issue got resolved

_Example from [GH docs ](https://docs.github.com/en/actions/reference/workflows-and-actions/dependency-caching#input-parameters-for-the-cache-action)_
```yaml
restore-keys: |
  npm-feature-${{ hashFiles('package-lock.json') }}
  npm-feature-
  npm-```

File: .github/actions/setup-node-and-cache/action.yml (hunk 3 above)
Comment (sdrozdsap) [reply .github/actions/setup-node-and-cache/action.yml:28]: Yes, it’s intentional. Restore keys use prefix-matching

Example from [GH docs](https://docs.github.com/en/actions/reference/workflows-and-actions/dependency-caching#input-parameters-for-the-cache-action)
```yaml
restore-keys: |
  npm-feature-${{ hashFiles('package-lock.json') }}
  npm-feature-
  npm-
```

feature-libs/organization/administration/components/shared/card/card.component.spec.ts
sdrozdsap: According to [Angular upgrade to 20 guide](https://angular.dev/update-guide?v=19.0-20.0&l=3) 
> Refactor application and test code to avoid relying on ng-reflect-* attributes. If needed temporarily for migration, use provideNgReflectAttributes() from @angular/core in bootstrap providers to re-enable them in dev mode only.

Since we no longer have `ng-reflect-*` attributes I had to refactor several test cases to not rely on it

feature-libs/storefinder/components/store-finder-list-item/store-finder-list-item.component.spec.ts
sdrozdsap: See https://github.com/SAP/spartacus/pull/20914#discussion_r2584735207

integration-libs/cds/src/merchandising/cms-components/directives/attributes/attributes.directive.spec.ts
sdrozdsap: See https://github.com/SAP/spartacus/pull/20914#discussion_r2584735207

.eslintrc.json
sdrozdsap: Good question! There are indeed much more lint warnings due to this change, however I didn't address them, as they are just warnings(especially considering that we didn't have them before). Some of them might be easy to fix, but for some cases more advanced refactor might be needed. And this felt like out of the upgrade scope. Please let me know if you think any further actions should be taken in this direction

.eslintrc.json
sdrozdsap: > Is this rule now enabled by default?

Yes it is. I can set it to any value we want, but considering that we already are in progress of migrating to `inject` as part of one of the spikes scope, I think we should be okay with not having this lint rule enabled

feature-libs/quote/components/list/quote-list.component.ts
sdrozdsap: According to [Angular 20 upgrade guide](https://angular.dev/update-guide?v=19.0-20.0&l=3):

> Review DatePipe usages. Using the Y (week-numbering year) formatter without also including w (week number) is now detected as suspicious. Use y (year) if that was the intent, or include w alongside Y.

Error from console:
> Error: NG02100: InvalidPipeArgument: 'NG02300: Suspicious use of week-based year "Y" in date pattern "MMMM d, YYYY h:mm aa". Did you mean to use calendar year "y" instead?' for pipe '_DatePipe'


The change is most probably also a fix, since `YYYY`(ISO Week-numbering year) is for very special cases and I don't think it was ever meant to be used in this context. `yyyy` is just ordinary year format which in most cases looks exactly like `YYYY` with the exception of dates around New Year, for example:

Date | yyyy | YYYY | Why
-- | -- | -- | --
2020-12-31 | 2020 | 2020 | Same year.
2021-01-01 | 2021 | 2020 | Jan 1, 2021 still belongs to the last ISO week of 2020.

feature-libs/product-configurator/rulebased/components/attribute/types/single-selection-image/configurator-attribute-single-selection-image.component.spec.ts
sdrozdsap: The problem actually is not with `CommonConfiguratorTestUtilsService`, rather with the same `ng-reflect-*` issue(see https://github.com/SAP/spartacus/pull/20914#discussion_r2584735207). Since we no longer can use it, I had to find elements with icon type INFO by using separate query

pawelfras (approved):
LGTM 🔥 
I also performed testing based on [CXSPA-11434](https://jira.tools.sap/browse/CXSPA-11434)

PR #20915, opened 2025-11-28
fix: add high contrast theme support for the OPF

Matejk00:
Closes: CXSPA-11569


uroslates (approved):
1. I started the app locally using the `npm run start:opf:ssl` command
2. Set authorizationCodeFlowByDefault: true (only option app works for me, for false I get 400 error on login)
3. Navigatd to https:localhost:4200 
4. Registered new user & logged in (redirected back to localhost).

Everything seems to look fine in Dark Mode. (on develop its reproducible).

👍 looks good!

PR #20930, opened 2025-12-05
fix: Use caret unicode characters for dropdowns over border shapes

Zeyber:
Closes: https://jira.tools.sap/browse/CXSPA-11554


uroslates (approved):
👍

PR #20935, opened 2025-12-05
fix: add proxy config for ccv2 server

FollowTheFlo:


projects/ssr-tests/src/utils/proxy.utils.ts
```diff
@@ -7,7 +7,15 @@
...
 import httpProxy from 'http-proxy';
+import * as https from 'https';
+import * as zlib from 'zlib';
 
+/**
+ * Set to `true` if the target server is a CCv2 server.
+ *
+ * This is needed to set up a custom HTTPS agent with the correct SNI servername.
+ */
+const IS_CCV2_SERVER = false;
```
FollowTheFlo, line 18: I couldn't find a way to automatically detect a CCV2 url VS Azure Dev url without speculation (detect ipv4 format or presence of 'myhybris'...), so decided to implement manual bool flag.
I'm checking with GC, if we can add SNI in our Azure Dev VM in futur so we'll be align on CCV2 and this flag won't be needed.

projects/ssr-tests/src/utils/proxy.utils.ts
```diff
@@ -42,17 +50,33 @@ interface ProxyOptions {
...
+  if (IS_CCV2_SERVER) {
+    proxyOptions.agent = new HttpsAgent({
+      servername: extractUrlWithoutProtocol(options.target),
+    });
+  }
+
+  const proxy = httpProxy.createProxyServer(proxyOptions);
+
+  proxy.on('proxyReq', (proxyReq) => {
+    proxyReq.setHeader('host', extractUrlWithoutProtocol(options.target));
```
FollowTheFlo, line 77: azure dev VM is not sensitive to header host key/value, it's ok to add it.

projects/ssr-tests/src/utils/proxy.utils.ts
Platonn: [QUESTION]

Do I can understand correctly that we're hardcoding here a value for IS_CCV2_SERVER to `false` only on the `develop` branch. And in the JDK21 branch we plan to hardcode it to `true`?

If yes, then how I can QA the CCV2 case? (besides enabling `IS_CCV2_SERVER = true;`)
What I should configure locally or what command to run, to test it for the CCV2 JDK21 server? I've tried setting locally temporarily dev CX_BASE_URL in `.env-cmdrc` to p5 ccv2 url https://api.cg79x9wuu9-eccommerc1-p5-public.model-t.myhybris.cloud . But then `npm run test:ssr` didn't pass due to backend returning 400 Bad Request. See logs from ssr tests:

```
{
        message: 'Ql [Error]: Outbound HTTP Error\n' +
...
          '  [cause]: je {\n' +
      "    headers: t { headers: Map(6) { 'connection' => [ 'close' ], 'content-length' => [ '226' ], 'content-type' => [ 'text/html; charset=iso-8859-1' ], 'date' => [ 'Tue, 09 Dec 2025 09:19:46 GMT' ], 'server' => [ '*' ], 'strict-transport-security' => [ 'max-age=15724800' ] }, normalizedNames: Map(6) { 'connection' => 'connection', 'content-length' => 'content-length', 'content-type' => 'content-type', 'date' => 'date', 'server' => 'server', 'strict-transport-security' => 'strict-transport-security' }, lazyInit: undefined, lazyUpdate: null },\n" +
          '    status: 400,\n' +
          "    statusText: 'Bad Request',\n" +
          "    url: 'http://localhost:9002/occ/v2/basesites?fields=FULL',\n" +
          '    ok: false,\n' +
          '    type: undefined,\n' +
          "    name: 'HttpErrorResponse',\n" +
          "    message: 'Http failure response for http://localhost:9002/occ/v2/basesites?fields=FULL: 400 Bad Request',\n" +
          `    error: SyntaxError: Unexpected token '<', "<!DOCTYPE "... is not valid JSON\n` +
          '        at JSON.parse (<anonymous>)\n' +
...
          '}',
```

FollowTheFlo: To QA with ccv2:
- in .env-cmdrc, set dev with s1 or p5:
  "dev": {
    "CX_BASE_URL": "https://api.c432wmya2v-teamspart3-s1-public.model-t.myhybris.cloud"
  },
 - set ->  IS_CCV2_SERVER = true; (in proxy.utils.ts)
 - nx reset
 - npm run test:ssr
 
 
 I just tried with p5 ccv2 env, tests are passing.

Platonn (approved):

PR #20936, opened 2025-12-08
chore: update Node.js engine version in package.json to ^22.22.0

sdrozdsap:
We set minimum Node version as `^22.22.0` however Angular 21 itself works with `22.12.0`. The reason for that is security vulnerability that was fixed only in `22.22.0` (for more details see: https://nodejs.org/en/blog/vulnerability/december-2025-security-releases)


pawelfras (dismissed):
LGTM

pawelfras (approved):
LGTM.

In the merge commit, let's add a note what's the reason of setting this exact version even if Angular 21 itself works with ^22.12:
https://nodejs.org/en/blog/vulnerability/december-2025-security-releases

PR #20941, opened 2025-12-09
feat: CXSPA-11615 Schematics for removing server routes config in fresh SPA application

pawelfras:
This PR introduces schematics for removing `app.routes.server.ts` with config for server routes, that are not supported by Spartacus SSR.

QA steps:
- checkout the PR's branch
- build libs:
  ```bash
  npm run build:libs
  ```
  ```bash
  ```
- generate fresh Angular 20 app
  ```bash
  npx @angular/cli@20 new my-app --style=scss --ssr=false --zoneless=false --standalone=false
  ```
- go to new app and install Spartacus with debug flag:
  ```bash
  npx ng add @spartacus/schematics --baseUrl=https://40.76.109.9:9002 --ssr --debug 
  ```
- verify if logs contain information about server routes removal:
  Screenshot 2025-12-09 at 17 18 05

- open the project and see if the following code was removed from `app.module.server.ts` file
   ```diff
    import { NgModule } from '@angular/core';
   - import { provideServerRendering, withRoutes } from '@angular/ssr';
    import { App } from './app.component';
    import { AppModule } from './app.module';
   - import { serverRoutes } from './app.routes.server';
    import { provideServer } from '@spartacus/setup/ssr';
    
    @NgModule({
      imports: [AppModule],
   - providers: [provideServerRendering(withRoutes(serverRoutes)), 
   + providers: [
         ...provideServer({
            serverRequestOrigin: process.env['SERVER_REQUEST_ORIGIN'],
          }),],
      bootstrap: [App],
    })
    export class AppServerModule {}
   ```
- verify if there is no `app.routes.server.ts` file in the project
- build the app and run the prod version to verify if it works:
   ```bash
   npm run build
   NODE_TLS_REJECT_UNAUTHORIZED=0 run serve:ssr:my-app
   ```
   
closes [CXSPA-11615](https://jira.tools.sap/browse/CXSPA-11615)


projects/schematics/src/add-ssr/index.ts
```diff
@@ -82,7 +94,7 @@ export function modifyAppServerModuleFile(): Rule {
     const changes = [importChange, ...providerChanges];
     commitChanges(tree, appServerModulePath, changes);
 
-    context.logger.log('info', `✅️ Modified app.server.module.ts file.`);
+    context.logger.log('info', `✅️ Modified app.module.server.ts file.`);
```
sdrozdsap, line 97: `nitpick`: In the same file `context.logger.info` is used multiple times, is there any difference between `context.logger.info` vs `context.logger.log('info',...)`?
If there isn't any difference I suggest to use the same method for consistency reasons

sdrozdsap (approved):
LGTM 👍 

Review and QA ✅

projects/schematics/src/add-ssr/index.ts
pawelfras: Hi, 
thank you for this question! The [logger.info](https://github.com/angular/angular-cli/blob/1f3387a016a473776b35ae99fd7753ee2d2179f2/packages/angular_devkit/core/src/logger/logger.ts#L127) uses under the hood the `logger.log('info', ...)`, so there is no difference.
When it comes to this particular line of code, logger.log was used 6 years ago, but only now I noticed the message was not adjusted after changes in Angualr 17, where name of file containing AngularServerModule was changed from `app.server.module.ts` to `app.module.server.ts`.
If you don't mind, I'll keep it that way in this PR. Nevertheless the consistency is important and we could addres this in a separate PR across whe whole schematic

PR #20943, opened 2025-12-10
feat: CXSPA-11633: add asm config to disable create customer

i333055:


feature-libs/asm/root/config/default-asm-config.ts
```diff
@@ -77,5 +77,9 @@ export const defaultAsmConfig: AsmConfig = {
     userIdHttpHeader: {
       enable: true,
     },
+    createCustomer:
+    {
+      enable: false,
```
scarai-sap: maybe set default value to true?

scarai-sap (approved):
LGTM

PR #20948, opened 2025-12-12
fix: enhance organization conflict handler

niehuayang:
https://jira.tools.sap/browse/CXEC-57988


i53577 (approved):
LGTM

Zeyber (approved):

Pucek9:
@Zeyber I'm not sure if I understand the [point](https://sap.github.io/spartacus-docs/breaking-changes/#avoiding-breaking-changes) properly: 
> Changing function behavior where the function returns different results. In other words, when changing function behavior, the function should return the same results for the same provided input as it did before the change.

I think that theoretically this is the case here

Pucek9:
@niehuayang you need to change (at least the last one) your commiter email, to be from @sap.com domain
You can simply do that in your console 
`git config --global user.email "your.email@sap.com"`
and
`git commit --allow-empty -m "Trigger Build"`
and push

Pucek9 (changes requested):
Update: We discussed it in the core team, and we agreed this example meets the criteria for a breaking change and should be nested in the appropriate featureToggle. 

The point is that there may be customers who have already handled this in their own way, which means that this update will affect the operation of their applications in a way that is unexpected for them.

Of course, we are open to arguments; perhaps there are circumstances we are unaware of.

niehuayang:
> @Zeyber I'm not sure if I understand the [point](https://sap.github.io/spartacus-docs/breaking-changes/#avoiding-breaking-changes) properly:
> 
> > Changing function behavior where the function returns different results. In other words, when changing function behavior, the function should return the same results for the same provided input as it did before the change.
> 
> I think that theoretically this is the case here

Hello @Pucek9 , thanks for your review, 
I would like to explain more to you. Basically, this is a bug from backend side on commerce, related to API POST/{baseSiteId}/costcenters. When creating a cost center using existing cost center code, the API returns a status of 400 with the error body ("message" : "The application has encountered an error",       "type" : "ModelSavingError"). 
However, Spartacus does not handle this 400 error and there is no error prompt on the page. For more details, please refer to the ticket: https://jira.tools.sap/browse/CXEC-57988

Now our team has fixed the bugs related to the API from backend and are planning to enhance the organizational conflict handling logic in Spartacus

> The point is that there may be customers who have already handled this in their own way, which means that this update will affect the operation of their applications in a way that is unexpected for them.

I agree with your concern above, but I believe that customer never handled this in their own way, because the root cause of bug is from backend. The prerequisite for customers themselves to fix the Spartacus issue is to first fix it on the backend side(I think it's impossible for the customer to fix this issue on their own in the backend..)

If you have any further concerns, please let me know. Thank you

Pucek9 (approved):
In fact, the only scenario that would be possible is a scenario where they fixed both the backend and the frontend. This sounds unlikely, but still possible.
I approve, but the responsibility lies with you 😁

PR #20951, opened 2025-12-12
feat: The site-context getLabel has been made multilingual

MehmetSert:
The Language and Currency label values ​​used in `site-context-selector.component` are fixed values ​​within `site-context-component.service`. I've changed this to be compatible with multi-language structures.

The Language and Currency label values ​​used in `site-context-selector.component` are constant values ​​within `site-context-component.service`. I've changed this to be compatible with multi-language structures.

Now, instead of reading the label value from a constant variable, it reads it from the `language` and `currency` parameters in `common.json`. Also, the `getLabel` function used to return `Observable<any>`; I've changed this to `Observable<string>` for type safety.

### DIFF
--- a/projects/assets/src/translations/cs/common.json
+++ b/projects/assets/src/translations/cs/common.json
@@ -33,7 +33,9 @@
     "zoomOut": "Zmenšit",
     "selected": "Vybráno",
     "expand": "Rozbalit",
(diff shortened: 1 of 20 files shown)


MehmetSert:
https://github.com/SAP/spartacus/commit/14166c011f

PR #20954, opened 2025-12-15
feat: CXSPA-11613 Migration - removal of obsolete 'index' property from 'angular.json'

pawelfras:
This PR contains a migration schematics for removing obsolete `index` property from the `angular.json` to makethe migrated app closer to the fresh Angular 21 app.

**QA steps:**

**Prerequisites:**
- checkout the PR's branch
- build libs:
  ```bash
  npm run build:libs
  ```
  ```bash
  ```

**CSR:**
1. generate fresh Angular 19 app
   ```bash
   npx @angular/cli@19 new my-app --style=scss --ssr=false --zoneless=false --standalone=false
   ```
2. go to app's location and add `.npmrc` file (`touch .npmrc`) with the following snippet (to install packages from RBSC):
    ```bash
    always-auth=true
    @spartacus:registry=https://73554900100900004337.dev.npmsrv.base.repositories.cloud.sap/
    //73554900100900004337.dev.npmsrv.base.repositories.cloud.sap/:_auth=<PUT_HERE_YOUR_TOKEN>
    ```
3. commit changes
4. install from RBSC the Spartacus in version 221121.5
    ```bash
    npx ng add @spartacus/schematics@221121.5 --baseUrl=https://40.76.109.9:9002
    ```
5. bump Angular libs to v20 and relatet libs to version that supports new Angular:
     ```bash
     ng update @angular/core@20 @angular/cli@20 @ngrx/store@20 angular-oauth2-oidc@20 @ng-select/ng-select@20 ngx-infinite-scroll@20 --force
     git add .
     git commit -m "update angular 20 and 3rd party deps angular 20 compatible"
     ```
6. comment or remove the `.npmrc` from the project (to get packages from Verdaccio) and commit changes
7. update Spartacus packages:
   ```bash
   ng update @spartacus/schematics@latest
   ```
8. verify if the following logs are present in the terminal:
   image
9. verify, if the `index` property has been removed from `angular.json`
10. build the project and run the prod version to verify if it works
     ```bash
     run build
     npx lite-server --baseDir="dist/my-app/browser"
     ```
   
**SSR:**
1. Repeat steps 1-9 from instructions for CSR, but add `--ssr` flag when installing Spartacus in step 4.
2. build project and run the prod version to verify if it works
    ```bash
    run build
    NODE_TLS_REJECT_UNAUTHORIZED=0 run serve:ssr:my-app
    ```

closes [CXSPA-11613](https://jira.tools.sap/browse/CXSPA-11613)


projects/schematics/src/migrations/221121_8/angular-config/index.ts
```diff
@@ -0,0 +1,71 @@
...
+  return (tree: Tree, context: SchematicContext): Tree => {
+    context.logger.info(
+      '\n⌛️ Removing obsolete "index" property from angular.json...'
+    );
+
+    const { path, workspace: angularJson } = getWorkspace(tree);
+    const projectName = getDefaultProjectNameFromWorkspace(tree);
+
+    const project = angularJson.projects[projectName];
+    const architect = project.architect;
```
sdrozdsap, line 30: `query`: Is `project` would always be non-nullable here?

projects/schematics/src/migrations/221121_8/angular-config/index.ts
```diff
@@ -0,0 +1,71 @@
...
+    const build = architect?.build;
+    const options = build?.options;
+
+    // To avoid introducing to the @spartacus/schematics a new library `@angular-devkit/build-angular` where the `BrowserBuilderOptions` is located,
+    //`as any` was used to satisfy TS compiler.
+    if (!(options as any)?.index) {
+      context.logger.info('  ↳ No "index" property found to remove');
+      return tree;
+    }
```
sdrozdsap, line 39: `major`: The migration only handles the default build.options, but we can have multiple configurations (production, development, etc.) under `build.configurations`. The index property might exist in those configurations too. I suggest to add something like:
```ts
if (build?.configurations) {
  Object.keys(build.configurations).forEach(configName => {
    const config = build.configurations[configName];
    if ((config as any)?.index !== undefined) {
      delete (config as any).index;
    }
  });
}
```

projects/schematics/src/migrations/221121_8/angular-config/index_spec.ts
```diff
@@ -0,0 +1,121 @@
...
+      const newTree = await runner.runSchematic(
+        MIGRATION_SCRIPT_NAME,
+        {},
+        tree
+      );
+
+      const updatedContent = JSON.parse(newTree.readText('/angular.json'));
+      expect(updatedContent).toEqual(workspaceContent);
+    });
+  });
```
sdrozdsap, line 87: `minor`: Consider adding a test case for projects with multiple build configurations (production, development) that might have the index property

projects/schematics/src/migrations/migrations.json
```diff
@@ -254,6 +254,11 @@
       "version": "2211.36.0",
       "factory": "./2211_36/silence-sass-deprecations/silence-sass-deprecations#migrate",
       "description": "Silence Sass deprecations about the `@import` directive"
+    },
+    "00-migration-v221121_6-angular-config": {
+      "version": "221121.5",
```
sdrozdsap: `query`: Is this version not supposed to match the folder path name which is `221121_6` in this case?

sdrozdsap (commented):
Generally looks good! 
The PR title contains a typo: "obsolate" (probably should be "obsolete")

projects/schematics/src/migrations/221121_8/angular-config/index.ts
pawelfras: Good point with multiple configurations! However, skipping such a check was done by purpose. From my point of view, if customer's app has multiple configs with separate `index` properties, then we should not touch them as here might be there for a reason and it is difficult to predict if they're not required. We'd like to handle only default scenario, where the `index` is obsolete in the `angular.json` structure we expect to be based on our past installation and migration mechanisms.

I'm happy to discuss if you see it differently.

projects/schematics/src/migrations/221121_8/angular-config/index_spec.ts
pawelfras: see: https://github.com/SAP/spartacus/pull/20954#discussion_r2698856510

projects/schematics/src/migrations/221121_8/angular-config/index.ts
pawelfras: Yes, it is enforced by Angular schema and I think it does not require additional check.

projects/schematics/src/migrations/221121_8/angular-config/index.ts
sdrozdsap: Thanks for the explanation, that makes sense to me!

sdrozdsap (dismissed):
LGTM, thanks!

sdrozdsap (dismissed):

sdrozdsap (approved):

PR #20960, opened 2025-12-17
fix: make ng-select ariaLabelDropdown translateable

uroslates:
Closes: https://jira.tools.sap/browse/CXSPA-11443


uroslates:
This is one approach to make ng-select ariaLabelDropdown value translateable.
Cons of this approach is that we now depend on Dom structure (selector) encasulated in ng-select library implementation. Maning we're prone to it breaking our code W/o us noticing it (even with Unit test in place).

Alternative second approach (maybe cleaner but way more descriptive) is to modify all templates using ng-select and manually append **ariaLabelDropdown** translated value. This would mean also adding featureToggles everywhere. This is better from perspective of us not depending that much on breaking changes coming from the encapsulated implementation of ng-select component (which this PRs approach is leaning on).

Third approach is to completely wrap ngSelect into custom component. This is even harder to implement.

projects/storefrontlib/shared/directives/horizontal-scrolling-position/horizontal-scrolling-position.directive.spec.ts
```diff
@@ -113,10 +113,7 @@ describe('HorizontalScrollingPositionDirective', () => {
         );
         expect(
           directive['scrollingAreaIntersectionObserver'].takeRecords()
-        ).toEqual([
-          jasmine.objectContaining({ target: mockScrollingAreaStart }),
-          jasmine.objectContaining({ target: mockScrollingAreaEnd }),
-        ]);
+        ).toEqual([]);
```
Pucek9: If you modified this file due to failing unit tests, please check how it was done in [fix: failing-unit-test horizontal-scrolling-position.directive#20953](https://github.com/SAP/spartacus/pull/20953/changes)

Pucek9 (dismissed):
LGTM besides https://github.com/SAP/spartacus/pull/20960#discussion_r2641123852

projects/storefrontlib/shared/components/ng-select-a11y/ng-select-a11y.directive.spec.ts
```diff
@@ -32,9 +32,11 @@ class MockFeatureConfigService {
   }
 }
 
+const EXPECTED_TRANSLATION = 'of';
```
pawelfras: `minor:` Is the change required? If `EXPECTED_TRANSLATION` is not used in other places, let's revert the change.

projects/storefrontlib/shared/components/ng-select-a11y/ng-select-a11y.directive.ts
```diff
@@ -35,7 +34,9 @@ import { filter, merge, take } from 'rxjs';
 import { map } from 'rxjs/operators';
 import { BREAKPOINT, BreakpointService } from '../../../layout';
 
-const ARIA_LABEL = 'aria-label';
+export const NG_SELECT_DROPDOWN_PANEL_ITEMS_SELECTOR =
```
pawelfras: `major:` looks like this const is not used anywhere. Let's remove it.

projects/storefrontlib/shared/components/ng-select-a11y/ng-select-a11y.directive.ts
```diff
@@ -35,7 +34,9 @@ import { filter, merge, take } from 'rxjs';
...
+  '.ng-dropdown-panel-items[role="listbox"]';
+export const ARIA_LABEL = 'aria-label';
```
pawelfras: `query:` could you help me understand the purpose of making this label available outside the file?

projects/storefrontlib/shared/components/ng-select-a11y/ng-select-a11y.directive.ts
```diff
@@ -221,22 +215,4 @@ export class NgSelectA11yDirective implements AfterViewInit {
...
-        // NOTE: SIGNAL is not a part of the public API of @angular/core and may change without notice
-        // TODO: CXSPA-11443 find a way to apply customizeNgSelectAriaLabelDropdown in a different way
-        const ariaLabelDropdownSignal =
-          this.selectComponent.ariaLabelDropdown[SIGNAL];
-        ariaLabelDropdownSignal.applyValueToInputSignal(
-          ariaLabelDropdownSignal,
-          translation
-        );
-      });
-  }
```
pawelfras, line 235: `major:` since the method was public, removing it will cause breaking change. Let's mark it as deprecated since 221121.8 (version of upcoming major release) with a proper note, that it should not be used and the `ariaLabelDropdown` needs to me set explicitly in the template each time `ng-select` is used.

projects/storefrontlib/shared/directives/horizontal-scrolling-position/horizontal-scrolling-position.directive.spec.ts
uroslates: Applied! Thanks for the heads up. 👍

projects/storefrontlib/shared/components/ng-select-a11y/ng-select-a11y.directive.spec.ts
uroslates: 👍 Applied!

projects/storefrontlib/shared/components/ng-select-a11y/ng-select-a11y.directive.ts
uroslates: Applied! 👍

projects/storefrontlib/shared/components/ng-select-a11y/ng-select-a11y.directive.ts
uroslates: Removed export. Applied! 👍

projects/storefrontlib/shared/components/ng-select-a11y/ng-select-a11y.directive.ts
uroslates: 👍 Applied

projects/storefrontlib/shared/components/ng-select-a11y/ng-select-a11y.directive.ts
```diff
@@ -216,6 +216,11 @@ export class NgSelectA11yDirective implements AfterViewInit {
     observer.disconnect();
   }
 
+  /**
+   * @deprecated since 2211.33
+   * Use ariaLabelDropdown input of ng-select component to set the label for dropdown options list.
+   * E.g.: <ng-select [ariaLabelDropdown]="'custom.translation.key' | translate"></ng-select>
+   */
   customizeNgSelectAriaLabelDropdown() {
```
pawelfras: Thanks! Actually we don't need to call it anywhere (so the call in line 106 can be removed). It's only about keeping the method and marking as deprecated in case if any customer extended this function - it might be a thing due to fact the method was public.
By the way, only now I realised we should revert the body of the method to the original one, from `develop`, because the workaround with `SIGNAL` usage was only for the interim period. Would you mind reverting the shape of the method the original shape from the `develop` branch?

uroslates: Applied! 👍!

pawelfras: ... but since the method in original shape won't work in the original shape due to changes in the API of `ng-select` lib, we cannot keep it that way. That said, there is no other way than providing a breaking change by removing it. Apologies for the confusion!

Note: let's keep that in mind and ask Matthew what would be the best approach to inform customers about this breaking change.

pawelfras (approved):

PR #20963, opened 2025-12-18
feat: CXSPA-11690 Enable `typeCheckHostBindings`

pawelfras:
The `typeCheckHostBindings` is enabled by defult starting from Angualr 21. Unfortunately, enabling the flag caused type issues due to known Angular issue with types when binding to specific key, like `keydown.Enter`:
```bash
TS2345: Argument of type 'Event' is not assignable to parameter of type 'KeyboardEvent'.
  Type 'Event' is missing the following properties from type 'KeyboardEvent': altKey, charCode, code, ctrlKey, and 17 more.
```
This required type augmentation.

- enabled `typeCheckHostBindings` in main tsconfig file
- provided `types.d.ts` with augmentation for `GlobalEventHandlersEventMap`
- unified `keydown.*` event names across the app to use W3 standard to reduce number of possible event names in type augmentation (also in templates, for consistency)
   - the change was made only for required places in the code, but in the followup ticket we could convert all events to W3-standard names, most likely in any minor release.
- used triple-slash directive to inject `types.d.ts` in failing libraries
- fixed issues with @HostBinding and @HostListener decorators not related to type augmentation

for more, see: 
https://github.com/angular/angular/issues/63170
https://github.com/angular/angular/issues/40778
https://www.typescriptlang.org/docs/handbook/triple-slash-directives.html
https://www.w3.org/TR/DOM-Level-3-Events-key/#named-key-attribute-values

closes [CXSPA-11690](https://jira.tools.sap/browse/CXSPA-11690)


Platonn (approved):

pawelfras:
closed in favour of https://github.com/SAP/spartacus/pull/20967

PR #20970, opened 2025-12-23
feat: CXSPA-11727 Installation and migration docs adjustments

pawelfras:
- adjusted installation and migration docs with Angular 21 requirements
- adjusted version references to 221121.8 - the one we will use for major release

QA steps:

1. installation
   - follow instructions from installation.md file
   - test if installed app works
2. migration
    - build libs:
      ```bash
      npm run build:libs
      ```
      ```bash
      ```
   - generate fresh Angular 19 app
     ```bash
     npx @angular/cli@19 new my-app --style=scss --ssr=false --zoneless=false --standalone=false
     ```
   - go to app's location and add `.npmrc` file (`touch .npmrc`) with the following snippet (to install packages from RBSC):
      ```bash
      always-auth=true
      @spartacus:registry=https://73554900100900004337.dev.npmsrv.base.repositories.cloud.sap/
      //73554900100900004337.dev.npmsrv.base.repositories.cloud.sap/:_auth=<PUT_HERE_YOUR_TOKEN>
      ```
   - commit changes
   - install from RBSC the Spartacus in version 221121.4
      ```bash
      npx ng add @spartacus/schematics@221121.4 --baseUrl=https://40.76.109.9:9002
      ```
    - comment/remove `.npmrc` file after installing Spartacus
    - follow migration.md steps
    - verify if app works
    - repeat above steps for SSR app (install Saprtacus with `--ssr` flag in app that supposed to be migrated) and include steps required for SSR

### DIFF
--- a/docs/migration/221121_8/installation.md
+++ b/docs/migration/221121_8/installation.md
@@ -1,6 +1,6 @@
-# Installing Spartacus 2211.21 with Angular 20
+# Installing Spartacus 221121.8 with Angular 21
 
-This guide provides step-by-step instructions for creating a fresh Angular 20 application and installing Spartacus 2211.21.
+This guide provides step-by-step instructions for creating a fresh Angular 20 application and installing Spartacus 221121.8
 
 ## Prerequisites
 
@@ -8,20 +8,20 @@ Before starting, ensure you have the following installed:
 
 - **Node.js**: Version 22 or higher
 - **npm**: Version 10 or higher
-- **Angular CLI**: Version 20
+- **Angular CLI**: Version 21
 
 Install or update Angular CLI globally:
 
 ```bash
-npm install -g @angular/cli@20
+npm install -g @angular/cli@21
 ```
 
-## Step 1: Create a New Angular 20 Application
+## Step 1: Create a New Angular 21 Application
 
-Create a new Angular 20 application:
+Create a new Angular 21 application:
 
 ```bash
-ng new my-spartacus-app --style=scss --ssr=false --zoneless=false --standalone=false
+ng new my-spartacus-app --style=scss --ssr=false --zoneless=false --standalone=false --file-name-style-guide=2016
 cd my-spartacus-app
 ```
 
@@ -37,6 +37,21 @@ The schematics will:
 - rename files in the project to use classic file name style guide (e.g., `app.component.ts` instead of `app.ts`)
 - Install required Spartacus libraries
 - Configure your application for Spartacus
+
+## Manual Changes
+
+//TODO: if Angular installation schematics are updated to include these changes, we can remove this section.
+In `main.ts`, update the replace deprecated BootstrapOptions object property `ngZoneEventCoalescing` with the new `provideZoneChangeDetection` function:
+
+```diff
++ import { provideZoneChangeDetection } from '@angular/core';
+
+ platformBrowser().bootstrapModule(AppModule, {
+-  ngZoneEventCoalescing: true,
++  provideZoneChangeDetection({ eventCoalescing: true }),
+})
+  .catch(err => console.error(err));
+```
 
 ## Step 3: SSR-Specific Configuration (If Using SSR)
 
--- a/docs/migration/221121_8/installation.md
+++ b/docs/migration/221121_8/installation.md
@@ -8,12 +8,12 @@ Before starting, ensure you have the following installed:
 
 - **Node.js**: Version 22 or higher
 - **npm**: Version 10 or higher
-- **Angular CLI**: Version 21
+- **Angular CLI**: Version 21.0.5
 
 Install or update Angular CLI globally:
 
 ```bash
-npm install -g @angular/cli@21
+npm install -g @angular/cli@21.0.5
 ```
 
 ## Step 1: Create a New Angular 21 Application
@@ -38,21 +38,6 @@ The schematics will:
 - Install required Spartacus libraries
 - Configure your application for Spartacus
 
-## Manual Changes
-
-//TODO: if Angular installation schematics are updated to include these changes, we can remove this section.
-In `main.ts`, update the replace deprecated BootstrapOptions object property `ngZoneEventCoalescing` with the new `provideZoneChangeDetection` function:
-
-```diff
-+ import { provideZoneChangeDetection } from '@angular/core';
-
- platformBrowser().bootstrapModule(AppModule, {
--  ngZoneEventCoalescing: true,
-+  provideZoneChangeDetection({ eventCoalescing: true }),
-})
-  .catch(err => console.error(err));
-```
-
 ## Step 3: SSR-Specific Configuration (If Using SSR)
 
 For Spartacus with Server-Side Rendering (SSR), run the following command:
@@ -62,30 +47,3 @@ ng add @spartacus/schematics@221121.<latest> --ssr
 ```
 This will set up SSR-specific configurations.
 
-### Manual Changes for SSR with `application` builder
-
-If you enabled SSR during project creation, you need to make an additional manual changes in the project files:
-
-1. remove `app.routes.server.ts` file as it is not supported by Spartacus SSR.
-2. remove `provideServerRendering` from `app.server.module.ts` file:
-
-```diff
-import { NgModule } from '@angular/core';
-import { provideServerRendering, withRoutes } from '@angular/ssr';
-import { provideServer } from '@spartacus/setup/ssr';
-import { App } from './app.component';
-import { AppModule } from './app.module';
-import { serverRoutes } from './app.routes.server';
-
-@NgModule({
-  imports: [AppModule],
-  providers: [
--   provideServerRendering(withRoutes(serverRoutes)),
-    ...provideServer({
-      serverRequestOrigin: process.env['SERVER_REQUEST_ORIGIN'],
-    }),
-  ],
-  bootstrap: [App],
-})
-export class AppServerModule {}
-```
--- a/docs/migration/221121_8/installation.md
+++ b/docs/migration/221121_8/installation.md
@@ -30,11 +30,10 @@ cd my-spartacus-app
 Run the Spartacus schematics to add Spartacus to your project:
 
 ```bash
-ng add @spartacus/schematics@221121.<latest>
+ng add @spartacus/schematics@221121.8
 ```
 
 The schematics will:
-- rename files in the project to use classic file name style guide (e.g., `app.component.ts` instead of `app.ts`)
 - Install required Spartacus libraries
 - Configure your application for Spartacus
 
@@ -43,7 +42,7 @@ The schematics will:
 For Spartacus with Server-Side Rendering (SSR), run the following command:
 
 ```bash
-ng add @spartacus/schematics@221121.<latest> --ssr
+ng add @spartacus/schematics@221121.8 --ssr
 ```
 This will set up SSR-specific configurations.
 
--- a/docs/migration/221121_8/migration.md
+++ b/docs/migration/221121_8/migration.md
@@ -139,7 +139,31 @@ Let's make following manual changes to modernize so it's similar to a new Angula
     }
 ```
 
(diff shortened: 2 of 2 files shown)


docs/migration/221121_8/migration.md
```diff
@@ -150,7 +162,6 @@ Let's make following manual changes to modernize so it's similar to a new Angula
     "enableI18nLegacyMessageIdFormat": false,
     "strictInjectionParameters": true,
     "strictInputAccessModifiers": true,
-+    "typeCheckHostBindings": true,
```
Platonn, line 153: [QUESTION]
Why are we no longer recommending to add this manually to a migrated app?
I guess because it will be cumbersome for customers to comply with those new type checks and would require [a workaround like we did](https://github.com/SAP/spartacus/pull/20967) (with global keyboard event types)? That said, won't our `declare global` workaround implicitly fix all the issues for customers (so they could safely use `"typeCheckHostBindings": true,` in their migrated apps)?

docs/migration/221121_8/migration.md
```diff
@@ -199,16 +210,36 @@ import { SpartacusModule } from './spartacus/spartacus.module';
...
 ```
 
-## If using Server Side Rendering (SSR)
+Note: The following migration should also take care of adding `provideZoneChangeDetection` to `main.ts` together with migrating deprecated bootstrap options:
+
+```
+❯ Migrates deprecated bootstrap options to providers.
+UPDATE src/main.ts (334 bytes)
+(1 file modified)
+```
```
Platonn: [MINOR]
While reading the doc top-down, I'm confused in this moment of reading. When this migration runs? There is no migration command to be run here.

I guess this paragraph is about a migration that has been run automatically earlier in the process. In that case, let's make it clear it was already run in the past (so me as a reader I'm not uncertain whether I maybe missed running some command). For example:
"The following migration was likely already executed earlier by the `ng update` command and it should already have taken care of adding..."

docs/migration/221121_8/migration.md
```diff
@@ -199,16 +210,36 @@ import { SpartacusModule } from './spartacus/spartacus.module';
...
++  provideZoneChangeDetection({ eventCoalescing: true }),
+})
+  .catch(err => console.error(err));
+```
+
+## Additional migration steps if using Server Side Rendering (SSR)
 
 ### Upgrade Express to Version 5
 
 Spartacus 221121.<latest> requires Express 5.x. Upgrade Express:
```
Platonn: [MINOR]
PR description says everything is meant to be updated to `.8`. Is this a left over? if yes, there are more places like this.

docs/migration/221121_8/installation.md
Platonn, line 1: [NITPICK]
Not related to this file, but to the PR description QA steps migration guide:

1. I needed to remove --zoneless flag, not supporeted in ng new v19
```diff
- npx @angular/cli@19 new my-app --style=scss --ssr=false --zoneless=false --standalone=false
+ npx @angular/cli@19 new my-app --style=scss --ssr=false --standalone=false
```

2. I needed to use .5 version for ng19:
```diff
- npx ng add @spartacus/schematics@221121.4 --baseUrl=https://40.76.109.9:9002
+ npx ng add @spartacus/schematics@221121.5 --baseUrl=https://40.76.109.9:9002
```
(but then I needed to first change all our libs' package.jsons "version" and "peerDeps" to be changed to .8

docs/migration/221121_8/migration.md
```diff
@@ -86,12 +83,27 @@ The result of migration should be following:
```
Platonn, line 81: According to Matt's last year's comments, let's mention also that customer might be faced with other optional migration:

```
Select the migrations that you'd like to run
❯◯ [control-flow-migration] Converts the entire application to block control flow syntax.
 ◯ [router-current-navigation] Replaces usages of the deprecated Router.getCurrentNavigation method with the Router.currentNavigation signal.
```

And let's inform that from Spartacus perspective those Angular migrations are not required, but customer may decide to opt-in to them

docs/migration/221121_8/migration.md
```diff
@@ -86,12 +83,27 @@ The result of migration should be following:
 
 Angular migration also takes care of adding `"typeSeparator": "."` to and proper type (suffix) to relevant schematics.
 
+### Update Angular to 21 and 3rd party deps to be compatible with Angular 21
+
+Follow the [Angular guidelines for upgrading from v20 to v21](https://angular.dev/update-guide?v=20.0-21.0&l=3) and bump the Angular version locally, and update other 3rd party dependencies from Angular ecosystem to versions compatible with Angular 21 (e.g. `@ng-select/ng-select@latest`, `@ngrx/store@21`, `angular-oauth2-oidc@21`, `ngx-infinite-scroll@latest`):
+
+```bash
+ng update @angular/core@21 @angular/cli@21 @ngrx/store@21 angular-oauth2-oidc@21 @ng-select/ng-select@21 ngx-infinite-scroll@21 --force
```
Platonn: [INFO]
I got an error
`Package specified by 'angular-oauth2-oidc@21' does not exist within the registry.`
But I understand it's temporary error until the release. I'm testing with angular-oauth2-oidc@20 then

docs/migration/221121_8/migration.md
```diff
@@ -199,16 +210,36 @@ import { SpartacusModule } from './spartacus/spartacus.module';
 export class AppModule { }
 ```
 
-## If using Server Side Rendering (SSR)
+Note: The following migration should also take care of adding `provideZoneChangeDetection` to `main.ts` together with migrating deprecated bootstrap options:
```
Platonn: [MEDIUM]
To modernize the app truly (i.e to look like a fresh angular app), I'd suggest in Manual Steps to move this provider directly manually to app.module.

docs/migration/221121_8/installation.md
```diff
@@ -0,0 +1,49 @@
...
+## Step 2: Install Spartacus
+
+Run the Spartacus schematics to add Spartacus to your project:
+
+```bash
+ng add @spartacus/schematics@221121.<latest>
+```
+
+The schematics will:
+- rename files in the project to use classic file name style guide (e.g., `app.component.ts` instead of `app.ts`)
```
Platonn: [QUESTION]


I guess it's no longer the responsibility of the installation schematics, since Angular CLI was fixed by your PR there?

docs/migration/221121_8/migration.md
Platonn, line 1: [I could not comment on exact line, so commenting here]

1. 

> The result of migration should be following:
> (...)
> ```
> +      "root": "",
> +      "sourceRoot": "src",
> ```

I already had it in my ng19 app

2. 

> ```diff
>             "outputPath": "dist/<your-project-name>",
> -            "index": "src/index.html",
>             "browser": "src/main.ts",
> ```

When I compare a migrated app vs new app, I can see one more property removed: outputPath. Is it correct?

docs/migration/221121_8/installation.md
pawelfras: True, apologies. My way of thinking was to suggest installing N-1 version from the RBSC, to not change version of libraries in package.json files. Apologies for not highlighting that

docs/migration/221121_8/migration.md
pawelfras: Good point. That was my intention, however the v21 is not yet released and it still to be decided whether we want wait for it or we will go with version 20 (works as expected also with Angular 21). For the safety reasons, I'll change the version here to v20 (even if duplicated with the previous step) and we'll track it and adjust, if needed.

docs/migration/221121_8/migration.md
pawelfras: The `typeCheckHostBindings` is enabled by default starting from version 21 and does not need to be explicitly set, that's why it was removed. We still recommend to do it and thank you for this comment - most likely we'll have to let our customers know about the workaround we did, so they can apply it in their projects.

pawelfras: Alternatively, we could initially suggest/automate disabling this flag and describe how to enable it and what's te workaround for missing types. Which approach would you prefer? In my opinion we should require this flag to keep migrated app in line with fresh installation, what we want to achieve. In addition, having it enabled now make's our customers' apps future proof (in case Angular Team decide to remove option for disabling such a type check).

I pushed update to the doc describing the issue and how to fix it. Let me know if such description works for you.

docs/migration/221121_8/migration.md
pawelfras: 1. Good point, I removed it from the doc. Thanks!
2. Yes, starting from v20, the `outputPath` defaults to `dist/<project_name>` (for more, see: https://github.com/angular/angular-cli/pull/29905). Unfortunately, it is difficult to determine whether customers modified this value in their projects and what their motivation was, so I did not want to force the removal of this value from migrated applications, even at the cost of inconsistency between new and migrated. 
However, we can suggest to customers that they can remove it from their projects if the value is the same as the default, i.e.

docs/migration/221121_8/migration.md
```diff
@@ -176,11 +181,30 @@ Let's make following manual changes to modernize so it's similar to a new Angula
...
+
+While it's not recommended, you can still disable the flag by adding the following configuration to your `tsconfig.json`:
+
+```diff
+{
+  "angularCompilerOptions": {
+    "enableI18nLegacyMessageIdFormat": false,
+    "strictInjectionParameters": true,
+    "strictInputAccessModifiers": true,
+    "strictTemplates": true,
```
Platonn: [NITPICK]
For brevity, I'd remove those guys, and leave only what has changed

docs/migration/221121_8/migration.md
```diff
@@ -176,11 +181,30 @@ Let's make following manual changes to modernize so it's similar to a new Angula
 +}
 ```
 
-1. In `app.module.ts`, add the `provideBrowserGlobalErrorListeners` to the `providers` array. For more, see: https://angular.dev/best-practices/error-handling#client-side-rendering
+It's worth noting that starting from Angular 21, flag `typeCheckHostBindings` is enabled by default. Due to its srtict type checking, it cause a [known issue](https://github.com/angular/angular/issues/63170) if specific `keydown` bindings are used in `@HostListener` decorators. To solve the problem in Spartacus repo, we introduced [type augmentation](https://github.com/SAP/spartacus/blob/ac651f413f44345bf8519391789c4f47c8ed02b0/types.d.ts#L1) for `global` interface. If you encounter similar issues in your application, it is recommended to apply the same type augmentation approach in your project.
```
Platonn: [MEDIUM]
> starting from Angular 21, flag `typeCheckHostBindings` is enabled by default.

I understood in the past that only fresh apps get this config enabled by default. (Maybe I missunderstood it?) But this paragraph sounds like all ng21 apps (even migrated ones) get it enabled by default.

If I understood correctly, I'd suggest the following edits:

```suggestion
Note: In fresh apps generated with Angular 21 CLI, the flag `typeCheckHostBindings` is enabled by default, so we suggest adding it also in migrated apps from Angular 19 to 21. However in apps migrated from Angular 19 to 21 beware it might cause issues. Due to its strict type checking, it causes a [known Angular issue](https://github.com/angular/angular/issues/63170) if specific `keydown` bindings are used in `@HostListener` decorators. To solve the problem in Spartacus repo, we introduced [type augmentation](https://github.com/SAP/spartacus/blob/ac651f413f44345bf8519391789c4f47c8ed02b0/types.d.ts#L1) for `global` interface. If you encounter similar issues in your application, we recommend you to apply analogical type augmentation solution in your project like we did in Spartacus repo.
```

docs/migration/221121_8/migration.md
```diff
@@ -139,6 +139,8 @@ Let's make following manual changes to modernize so it's similar to a new Angula
     }
 ```
 
+Starting from Angular 20, the `outputPath` option defaults to `dist/<project-name>`, what removest the need to explicitly setting it within a freshly generated app. If your migrated app has `outputPath` set to `dist/<project-name>`, we recommend removing it to keep it consistent with a new app structure.
```
Platonn:
```suggestion
Note: In fresh apps generated with Angular 21, the `outputPath` option is skipped and implicitly defaults to `dist/<project-name>`. If your migrated app has `outputPath` set to `dist/<project-name>`, we recommend removing it as not necessary.
```

Platonn (approved):

PR #21002, opened 2026-01-13
fix: remove pickup in store dependency from OPF

Matejk00:
Closes: CXSPA-11808


integration-libs/opf/checkout/components/opf-checkout-billing-address-form/opf-checkout-billing-address-form.service.ts
```diff
@@ -57,7 +56,6 @@ export class OpfCheckoutBillingAddressFormService {
   );
   protected userAddressService = inject(UserAddressService);
   protected _noDefaultAddressFoundForPickupMode$ = new Subject<void>();
-  protected pickupOptionFacade = inject(PickupOptionFacade);
```
FollowTheFlo, line 60: I guess it was a left over as pickupOptionFacade is declred but not used in this service.

FollowTheFlo (approved):
nice solution, lgtm

integration-libs/opf/checkout/components/opf-checkout-billing-address-form/opf-checkout-billing-address-form.service.ts
Matejk00: Yes.

PR #21009, opened 2026-01-14
chore: bump-schematics-deps

Pucek9:
Closes [CXSPA-11874](https://jira.tools.sap/browse/CXSPA-11874), [CXSPA-11875](https://jira.tools.sap/browse/CXSPA-11875), [CXSPA-11876](https://jira.tools.sap/browse/CXSPA-11876), [CXSPA-11877](https://jira.tools.sap/browse/CXSPA-11877)

"jsonc-parser": "^3.2.0" -> "^3.3.1",
[Changelog](https://github.com/microsoft/node-jsonc-parser/releases)

"parse5": "^7.2.1" ->  "^8.0.0", 
[Changelog](https://github.com/inikulin/parse5/releases)

"semver":  "^7.6.3" -> "^7.7.3",
[Changelog](https://github.com/npm/node-semver/releases)

"ts-morph": "^26.0.6" -> "^27.0.2",
[Changelog](https://github.com/dsherret/ts-morph/releases/tag/27.0.0)


Pucek9:
☑ tests `npm run test:all-schematics` 
☑ building and publishing `yarn build:libs && ./node_modules/ts-node/dist/bin.js ./tools/schematics/testing.ts`
☑ creating and running new app
```
npx @angular/cli@20 new appname --standalone=false --ssr=false --style scss
cd appname
npx @angular/cli@20 add @spartacus/schematics@v221121.5.0 --base-url=https://40.76.109.9:9002/
yarn start
```

feature-libs/asm/jest.schematics.config.js
```diff
@@ -19,6 +19,7 @@ module.exports = {
       },
     ],
   },
+  transformIgnorePatterns: ['node_modules/(?!parse5|@angular)'],
```
Platonn: [QUESTION]
I can understand this line allows to transform just `parse5` and `@angular` inside `node_modules` (rest of node_modules is not transformed).

Why do we need to add this config property at all? 
1. why do we need to transform parse5? I can see in major version 8.0 they changed to ESM-only https://github.com/inikulin/parse5/pull/1411 . But I'm still lacking understanding why would it (or something else?) imply that we have to transform it?
2. Why did you add `@angular` to the transform allow-list? To my limited understanding, before this PR, `@angular` was ignored from transforming (as part of [implicit default](https://jestjs.io/docs/configuration#transformignorepatterns-arraystring) ignoring of all `node_modules`.

Platonn: I've investigated answers to those questions and eventually I'd suggest the following code snippet:
```suggestion
  transformIgnorePatterns: [
    // Because ESM support is experimental in Jest, we don't enable it
    // So Jest has to transform ESM files inside node_modules.
    // For more see Jest docs: https://jestjs.io/docs/ecmascript-modules
    //
    // Due to a negative lookahead, we need to compose 2 patterns into one:
    // -  Original pattern `node_modules/(?!(.*\\.mjs$|@angular/common/locales/.*\\.js$))` - copied from jest-preset-angular sources
    //   see https://github.com/thymikee/jest-preset-angular/blob/8b7673ee739919433b0834fe4a3f55862801bd40/src/presets/create-cjs-preset.ts#L22
    // - Custom pattern `node_modules/(?!parse5)` - because in parse5@18.0.0 became an ESM-only package
    `node_modules/(?!(.*\\.mjs$|@angular/common/locales/.*\\.js$)|parse5)`,
  ],
```

My reasoning below:

I've tried removing this config on local in this PR branch and I can see the error:
```

    Jest encountered an unexpected token

    Jest failed to parse a file. This happens e.g. when your code or its dependencies use non-standard JavaScript syntax, or when Jest is not configured to support such syntax.

    Out of the box Jest supports Babel, which will be used to transform your files into valid JS based on your Babel configuration.

    By default "node_modules" folder is ignored by transformers.

    Here's what you can do:
     • If you are trying to use ECMAScript Modules, see https://jestjs.io/docs/ecmascript-modules for how to enable it.
     • If you are trying to use TypeScript, see https://jestjs.io/docs/getting-started#using-typescript
     • To have some of your "node_modules" files transformed, you can specify a custom "transformIgnorePatterns" in your config.
     • If you need a custom transformation, specify a "transform" option in your config.
     • If you simply want to mock your non-JS modules (e.g. binary assets) you can stub them out with the "moduleNameMapper" config option.

    You'll find more details and examples of these config options in the docs:
    https://jestjs.io/docs/configuration
    For information about custom transformations, see:
    https://jestjs.io/docs/code-transformation

    Details:

    /Users/krzysztof/code/classic-help-2/node_modules/parse5/dist/index.js:1
    import { Parser } from './parser/index.js';
    ^^^^^^

    SyntaxError: Cannot use import statement outside a module

      1 | import { SchematicsException, Tree } from '@angular-devkit/schematics';
    > 2 | import { DefaultTreeAdapterMap, parse as parseHtml } from 'parse5';
        | ^
```

I'd go for enabling enable ESM support in Jest, instead of transforming it, but only then I [learned in the Jest docs](https://jestjs.io/docs/ecmascript-modules) ESM support is experimental. So better not use it for testing. 👍 

Regarding 1) I find it justified then to transform `parse5`
Regarding 2) I find it justified to ignore also `@angular`, but not all of it. Learned that by overriding the whole property `transformIgnorePatterns,` we've overriden the default of the `jest-preset-angular` which was defined as [this](https://github.com/thymikee/jest-preset-angular/blob/8b7673ee739919433b0834fe4a3f55862801bd40/src/presets/create-cjs-preset.ts#L22). The original pattern is more sophisticated. I wondered maybe we could compose both patterns as an array, but due to negative lookahead with AND operator it didn't work as expected. I mean I tried:
```
const { createCjsPreset } = require('jest-preset-angular/presets');
const cjsAngularPreset = createCjsPreset();
/*...*/
transformIgnorePatterns: [
    ...cjsAngularPreset.transformIgnorePatterns,
    'node_modules/(?!parse5)',
  ],
```
and then one of those folders was ignored (while it shouldn't). This gotcha is described also in [Jest docs](https://jestjs.io/docs/configuration#transformignorepatterns-arraystring)

So eventually we need to compose "manually" those 2 patterns (the original from jest-preset-angualr and our custom one for parse5).
And IMHO we should add a code comment with explanation, as it's not obvious for a future newcomer developer why such a pattern.

Pucek9: Thanks @Platonn for the deeper analysis! 
Indeed, the pattern `transformIgnorePatterns: ['node_modules/(?!parse5|.*\\.mjs$|@angular/common/locales/.*\\.js$)'],` also works

FollowTheFlo (dismissed):

Platonn (approved):

PR #20991, opened 2026-01-08
fix: (CXSPA-11706) Deprecate enableCarouselCategoryProducts feature toggle

StanislavSukhanov:
closes: https://jira.tools.sap/browse/CXSPA-11706

PR #21013, opened 2026-01-14
fix: (CXSPA-11706) Deprecate enableCarouselCategoryProducts feature toggle.

StanislavSukhanov:
This was merged to develop. Needs to be merged here as well. 

https://github.com/SAP/spartacus/pull/20991


StanislavSukhanov:
not relevant

PR #21017, opened 2026-01-14
refactor: make StorefrontComponent a Standalone Component, but use new AppComponent in storefrontapp

Platonn:


Pucek9:
Nice! Btw it remind me this topic: https://github.com/SAP/spartacus/pull/20598/changes
Will we have similar effect? If yes I could close it.
But worth to inspire eslint rule

docs/migration/221121_8/standalone-components-in-spartacus.md
```diff
@@ -2,18 +2,26 @@
...
 
 ## Migrating to 221121.8.0
 
 ### Migrating to use OOTB Spartacus Standalone Components
-Existing OOTB Spartacus components can be used in your custom code as before, even after they became Standalone Components. You should be able to use them as you did before, even in your custom non-Standalone components.
+Existing OOTB Spartacus components can be used in your custom code as before, even after they became Standalone Components. You should be able to use them even in your custom non-Standalone components.
+
+Note: Very likely you have been using your custom root `AppComponent` in the `bootstrap` array of your `AppModule`. But if by some chance, you were using directly the Spartacus OOTB `StorefrontComponent` in the `bootstrap` array, then you should replace it with your custom non-Standalone `AppComponent`. It's because Angular's `bootstrap` array of `NgModule` accepts only non-Standalone components, but `StorefrontComponent` became now a Standalone Component.
```
pawelfras, line 12: `nitpick:` If not exist yet, let's create a ticket for automating this migration step (for the standard scenario)

pawelfras (approved):
LGTM

docs/migration/221121_8/standalone-components-in-spartacus.md
Platonn: I perceive it as an extreme edge case. Our installation schematics always create apps with a custom AppComponent as the root component. There is no reason/motivation for customers in practice to have changed their root component to StorefrontComponent. So this doc is just for mentioning it. I'm not convinced yet it's worth automating.

PR #21025, opened 2026-01-16
fix: noise in logs about not-found static files

Pucek9:


Pucek9:
Closed due to https://github.com/SAP/spartacus/pull/21078

PR #21032, opened 2026-01-19
fix: fix pages title (breadcrumb section) on languge switch

uroslates:
Closes: https://jira.tools.sap/browse/CXSPA-10923


projects/core/src/cms/facade/page-meta.service.ts
```diff
@@ -22,12 +23,19 @@ import { CmsService } from './cms.service';
   providedIn: 'root',
 })
 export class PageMetaService {
+  protected languageService: LanguageService = inject(LanguageService);
+
   constructor(
     protected cms: CmsService,
     protected unifiedInjector: UnifiedInjector,
-    protected pageMetaConfig: PageMetaConfig,
-    @Inject(PLATFORM_ID) protected platformId: string
```
rmch91, line 29: I don't think we can just like this remove the dependency from constructor.
Any change in constructor is a breaking change for customers that extend our PageMetaService.
I guess you removed it because it is not used anywhere, but it is protected, can be used in customers implementation.

In this case we should deprecate it and wait 12 months.

rmch91 (changes requested):

projects/core/src/cms/facade/page-meta.service.ts
uroslates: @rmch91 OK thanks I added it back with the depreciation comment.

projects/core/src/cms/facade/page-meta.service.ts
```diff
@@ -22,12 +23,21 @@ import { CmsService } from './cms.service';
   providedIn: 'root',
 })
 export class PageMetaService {
+  protected languageService: LanguageService = inject(LanguageService);
+
   constructor(
     protected cms: CmsService,
     protected unifiedInjector: UnifiedInjector,
     protected pageMetaConfig: PageMetaConfig,
+    // @deprecated since 4.3.8 - to be removed in 5.0
```
rmch91: Sorry, what is 4.3.8 and 5.0?
The next planned spa release is 221121.8.0

rmch91 (changes requested):

projects/core/src/cms/facade/page-meta.service.ts
rmch91: Also you can tell only that will be removed in 12 months, we don't know the version which will be in 12 months yet :)

uroslates: Applied!

rmch91 (approved):

PR #21039, opened 2026-01-20
docs: add migration instructions for enabling non-destructive hydration in SSR applications

npapp-dev002:


docs/migration/221121_8/migration.md
```diff
@@ -273,6 +273,76 @@ platformBrowser().bootstrapModule(AppModule, {
 
 ## Additional migration steps if using Server Side Rendering (SSR)
 
+### Enable Non-Destructive Hydration (Required for SSR)
+
+If your application uses Server-Side Rendering (SSR), you **must** enable Angular's non-destructive hydration. This is now a requirement for Spartacus 221121_8.
```
Platonn: [minor]
`must` vs `should`

I'm aware I said we'd like to _force_ customers to start using this feature, so the usage of word `must` feels right from this point of view.
In practice the truth is that their app won't break if they don't enable this feature.
So I think I'd use a word `should` in this case.

We may also add a reason/motivation behind it:
"This aligns your app with current Angular best practices, making maintenance and upgrades easier."

```suggestion
If your application uses Server-Side Rendering (SSR), you should enable Angular's non-destructive hydration. This is now a requirement for Spartacus 221121_8. This aligns your app with current Angular best practices, making maintenance and upgrades easier.
```

docs/migration/221121_8/migration.md
```diff
@@ -273,6 +273,76 @@ platformBrowser().bootstrapModule(AppModule, {
...
+
+#### Why are these options required?
+
+- **`withEventReplay()`**: Ensures that user interactions that occur before the application is fully hydrated are captured and replayed. This provides a seamless user experience even during the hydration process.
+
+- **`withNoHttpTransferCache()`**: Disables the HTTP transfer cache for hydration. This is required for Spartacus because Spartacus uses its own state transfer mechanism. Without this option, there could be conflicts between Angular's built-in HTTP transfer cache and Spartacus's custom implementation.
+
+#### Important Notes
+
+- Non-destructive hydration is **only required for SSR applications**. If your application does not use SSR, you can skip this step.
```
Platonn: [nitpick]
It's already inside SSR-only-migrations section of the doc, so I'm not sure we need to remind it again here (for brevity)

npapp-dev002: Removed line.

PR #21046, opened 2026-01-21
chore: remove old major-version migration schematics for versions v3-v6 (but leave v2211+)

Platonn:
Removing old migration, as they have been already executed in customers' projects.

Why:
The old migrations are disturbing the development of new ones. In particular, unit tests of old migrations are failing when modernizing the codebase in future migrations (e.g. due to enabling standalone:true in created apps in tests, or improving the shared util functions in schematics).

Scope of changes:
- migrations.json - removed migrations for /3_0 - /6_0
- removed folders for migrations /3_0 - /6_0

Note: We don't delete migration schematics for versions 2211+ (i.e last 2 previous majors), because they may be useful for ourselves. In particular 2211.19 (2 years ago) migration was very rich migration, including many steps that we might want to reuse with in the future (as a context/inspiration for future schematics implementation).

PS. Angular core team also removes old migration schematics, when they release a new major version, see examples:
- https://github.com/angular/angular/pull/60304
- https://github.com/angular/angular/pull/63207


pawelfras (approved):
LGTM

PR #21050, opened 2026-01-21
fix: Analyze git history for feature toggles properly

rmch91:
Closes: https://jira.tools.sap/browse/CXSPA-11954


ci-scripts/check-feature-toggles.sh
```diff
@@ -82,18 +83,19 @@ toggle_overrides=$(load_overrides)
 # Extract current toggles
 echo "📊 Analyzing current feature toggles..."
 
-# Function to extract feature toggles and their line numbers
-extract_toggles() {
+# Function to extract csr Spartacus feature toggles
```
Platonn: [NITPICK]
1. The output format is not obvious. Let's document it in a comment
```suggestion
# Function to extract csr Spartacus feature toggles
# It looks for a code block:
# ```
# export const defaultFeatureToggles = ... {
#   ...
# }
# ```
# Returns multiple lines string, each line separating 3 values with colons:
# `regular:TOGGLE_NAME:TOGGLE_VALUE`
```

Analogically, let's add docs comment for the SSR counterpart function

ci-scripts/check-feature-toggles.sh
```diff
@@ -102,39 +104,86 @@ extract_toggles() {
...
-echo "Found $(echo "$current_toggles" | wc -l | tr -d ' ') feature toggles"
+# Function to extract SSR feature toggles  
+extract_ssr_toggles() {
+    local file="$1"
+    # Extract feature toggles from ssrFeatureToggles object in defaultSsrOptimizationOptions
+    # Find the start of ssrFeatureToggles
+    local start_line=$(grep -n "ssrFeatureToggles: {" "$file" | cut -d: -f1)
+    if [ -z "$start_line" ]; then
+        echo "Warning: Could not find ssrFeatureToggles object in $file" >&2
+        return 0  # Not an error, just no SSR toggles
```
Platonn: [MEDIUM]
I suggest to return error.

Otherwise, if we ever change in the feature the way of defining ssrFeatureToggles, or refactor it anyhow, then this script will silently accept it, while the refactoring developer should update this script then.

As a result, we'll enforce that the wrapper object key `ssrFeatureToggles` must be always present, even if it's `ssrFeatureToggles: {}` (with empty object inside).
We just need to make sure the `end_line` is calculated properly and script works in such an edge case.

ci-scripts/check-feature-toggles.sh
```diff
@@ -275,6 +323,7 @@ if [[ ${#TOGGLES_TO_REMOVE[@]} -gt 0 ]] || [[ ${#TOGGLES_TO_ENABLE[@]} -gt 0 ]];
     
     echo "💡 To apply changes, manually edit:"
     echo "   projects/core/src/features-config/feature-toggles/config/feature-toggles.ts"
+    echo "   core-libs/setup/ssr/optimized-engine/ssr-optimization-options.ts"
```
Platonn: [MINOR]
In 99% cases developers change only Regular feature toggles.
SSR toggles are changed very rarely by few Asterix core team member.
Let's make it clear for 99% devs where they need to make changes (and when):
```suggestion
    echo "   Regular Feature Toggles:"
    echo "      projects/core/src/features-config/feature-toggles/config/feature-toggles.ts"
    echo "   SSR Feature Toggles:"
    echo "      core-libs/setup/ssr/optimized-engine/ssr-optimization-options.ts"

File: ci-scripts/check-feature-toggles.sh [hunk 4]
```diff
@@ -5,8 +5,9 @@
 set -e
 
 # Configuration
-# File containing feature toggles
-TOGGLES_FILE="projects/core/src/features-config/feature-toggles/config/feature-toggles.ts"
+# Files containing feature toggles
+CSR_TOGGLES_FILE="projects/core/src/features-config/feature-toggles/config/feature-toggles.ts"
```
Platonn: [MEDIUM]

Conceptually there are no _CSR_ feature toggles. Those just _Regular_ feature toggles. And I suggest to rename them so  (e.g. REGULAR_TOGGLES_FILE) as well as in other places in this script rename `csr`/`CSR` to `regular`/`Regular`.

Why: those toggles can control behavior both in CSR and in SSR runtime

The exceptional nature of "SSR feature toggles" is that they control the SsrOptimizationEngine, which is not a part of Spartacus product , but a "co-product" which wraps Spartacus (therefore it cannot access Spartacus internals, inlcuding Spartacus toggles config). A shorthand name for those feature toggles of SsrOptimizationEngine is "SSR feature toggles".

Summing up, I suggest renaming stuff in this script to:
- "csr" to "regular"
- keep "ssr"

ci-scripts/check-feature-toggles.sh
rmch91: changed

ci-scripts/check-feature-toggles.sh
rmch91: Changed

ci-scripts/check-feature-toggles.sh
rmch91: Changed

ci-scripts/check-feature-toggles.sh
rmch91: Changed

ci-scripts/check-feature-toggles.sh
```diff
@@ -102,39 +112,101 @@ extract_toggles() {
...
+# Function to extract ssr Spartacus feature toggles
+# It looks for a code block:
+# ```
+# ssrFeatureToggles: {
+#   ...
+# },
+# ```
+# Returns multiple lines string, each line separating 3 values with colons:
+# `ssr:TOGGLE_NAME:TOGGLE_VALUE`
+extract_ssr_toggles() {
```
Pucek9, line 132: If both scripts are to be used in the future, it's worth separating functions (maybe not only) into another file and reusing them; it should be possible:
Image

.github/workflows/ci.yml
```diff
@@ -95,6 +95,7 @@ jobs:
     needs: no_retries
     name: Check for new feature toggles
     runs-on: ubuntu-latest
+    if: github.ref_name == 'develop'
```
Platonn: [QUESTION]
Should we also check release/* branches?
I can imagine an edge-case situation where:
- a dev introduced a breaking change
- FF happened, release branch was cut off
- lately someone discovers that the already introduced breaking change wasn't covered with a toggle
- a toggle is added only after FF, directly to the release branch
- then a Doc Writer (Matt) needs to know we have to document one more new last-minute toggle

My understanding is that this script runs on every change in develop branch (not periodically once a month), right? Please correct me if I'm wrong.

Platonn (dismissed):

ci-scripts/check-feature-toggles.sh
Pucek9: I mean I had see 2 or 3 identical functions in both files

.github/workflows/ci.yml
rmch91: You are right, added release branch to triggers

ci-scripts/check-feature-toggles.sh
rmch91: So, that's not fully truth.
I agree that functions are similar, but one is parsing changes in the commit, other parsing file in the repo.
The thing is that when I tried to refactor it to have some common methods in separate file, it appeared that it is bigger ammount of code, but not really more readable. 
So I'd keep it as it is, because if there will be no bugs more, I don't see the reason to change this scripts in the future.

Platonn (approved):

ci-scripts/check-feature-toggles.sh
Pucek9: Yeah, indeed, there are even more lines after the refactor, but they include large comments
I got your arguments and I'm fine with the existing state, anyway please take a look if you like [fix/CXSPA-11954-separate-common](https://github.com/SAP/spartacus/pull/21113/changes)

PR #21043, opened 2026-01-20
feat: migration schematics for adopting bootstrapApplication

Platonn:
related to https://jira.tools.sap/browse/CXSPA-11890

**QA steps:**
- create app with ng19 and 221121.5
- migrate it to 221121.8 (and ng21) according to our migration guide in migration.md
- run `ng g @spartacus/schematics:modernize-app-standalone-bootstrap-application`
- test CSR and SSR

Btw.
- compare app using bootstrapApplicate vs app using bootstrapModule, and check whether we didn't remove too much with our migration

PR #21024, opened 2026-01-16
(CLOSED IN FAVOR OF #21059) feat: use standalone-friendly bootstrapApplication Angular API instead of bootstrapModule

Platonn:
(CLOSED IN FAVOR OF #21059)

PR #21059, opened 2026-01-21
feat: use standalone-friendly bootstrapApplication Angular API instead of bootstrapModule

Platonn:
Detailed scope:
1. use bootstrapApplication instead of bootstrapModule **in our storefrontapp** and mimic app structure of a fresh Standalone app
   - use `bootstrapApplication` in `main.ts` and `main.server.ts`
   - change `AppComponent` to standalone
   - create `app.config.ts` with Angular-speciifc providers (+ `importProvidersFrom(AppModule)`)
   - create `app.config.server.ts` Angular-specific SSR providers (+ `importsProvidersFrom(AppServerModule)`)
   - renamed default JS import in `server.ts` from `AppServerModule` to `bootstrap`
2. changed the installation schematics
   - angular-specific providers are placed in `app.config.ts` and `app.config.server.ts`
   - various pre-existing modules (ngrx modules, `SpartacusModule` and Spartacus `AppRoutingModule`) are kept in `AppModule`
   - Spartacus `provideServer()` is kept in `AppServerModule`
   - default JS import in `server.ts` is now `bootstrap` (instead of `AppServerModule`)
   - removed old unnecessary logic for installing non-standalone SSR (i.e. logic that cleaned up `app.module.server.ts`)
3. No longer use `--standalone=false`:
   - in unit tests of schematics (which create fresh apps)
   - in `221121_08/installation.md`
   - in `scripts/install/.run.sh` script
4. Update unit tests of installation schematics
   - update snapshots of generated `angular.json` - to include standalone:true config for generator schematics
   - update assertions of positions of code changes - for tests that implicitly relied on the non-standalone app strucutre
5. Update unit tests of add-cms-component generator schematics. But the schematics seem to have been broken and unmaintained even before this PR on develop. This PR doesn't address pre-existing issues with this generator.
   
By the way, moved out our private repo's stuff from  `AppModule` and `SpartacusConfigurationModule` - to align it 1:1 with fresh customer's apps:
   - Move out stuff from `AppModule`:
     - i18n config moved from `AppModule` to `SpartacusConfigurationModule`
     - configs specific to only our example storefront app and e2e testing moved from `AppModule` to `privateProviders` in separate file in storefrontapp
   - Created file `SpartacusConfigurationModule` with things common for b2b and b2c (like in customer apps)
     - `SpartacusB2bConfigurationModule` (and `B2c`) renamed to `spartacusB2bConfigurationProviders` (and `B2c` respectively) and moved there only stuff related to b2b (and b2c respectively)
     - used conditionally `spartacusB2bConfigurationProviders` (and `B2c`) in `SpartacusConfigurationModule`, based on the `environment.b2b`
     
     
By the way:
- added missing command to package.json: `test:schematics` - which runs tests just for the the `projects/schematics/`

Note: in another partial PR there are modernization/migration schematics for existing projects: https://github.com/SAP/spartacus/pull/21043

closes https://jira.tools.sap/browse/CXSPA-11890

---

**Things to QA:**
- our storefrontapp: check CSR and SSR
- installation: create fresh angular+spartacus app, then check CSR and SSR


docs/migration/221121_7/installation.md
Platonn, line 1: TODO: Address PR comments from the old PR:
https://github.com/SAP/spartacus/pull/21024

cc @pawelfras

projects/schematics/src/add-cms-component/index.ts
```diff
@@ -148,30 +139,29 @@ function updateModule(options: CxCmsComponentSchema): Rule {
...
+      componentPath
+    );
+    const componentImport = insertImport(
+      moduleTs,
+      modulePath,
+      componentName,
+      stripTsFromImport(componentRelativeImportPath),
+      false
+    );
+    changes.push(componentImport);
```
Platonn, line 163: Note: For standalone components, component is never added to `declarations` array, but we still want to add it to exports array of the module, therefore we need to add an import statement of it

projects/schematics/src/shared/utils/import-utils_spec.ts
```diff
@@ -210,8 +207,8 @@ describe('Import utils', () => {
         moduleSpecifier: SPARTACUS_CORE,
         namedImports: ['xxx'],
       });
-      expect(results[0].print()).toEqual(
-        `import { xxx } from "@spartacus/core";`
+      expect(results[0].print()).toMatch(
+        /import\s*\{[^}]*\bxxx\b[^}]*\}\s*from\s*['"]@spartacus\/core['"];?/
```
Platonn, line 211: Note: now the util adds import to existing import statement.

BEFORE:
```
import {x} from 'y';
import {z} from 'y'; // <--- import we added (z from y)
```

NOW
```
import { x, z } from 'y'; // <--- we add z from y to existing import block
```

projects/schematics/src/shared/utils/import-utils_spec.ts
```diff
@@ -285,17 +282,25 @@ describe('Import utils', () => {
       const appModule = program.getSourceFileOrThrow(appModulePath);
 
       expect(
-        staticImportExists(appModule, ANGULAR_PLATFORM_BROWSER, BROWSER_MODULE)
```
Platonn, line 288: Note: Standalone apps don't have `BrowserModule` anymore in the AppModule. So we need to use a different module as an example to test the util of NGModule.import removal

(side note: BROWSER_MODULE_PROVIDERS [are implicitly provided](https://github.com/angular/angular/blob/d9c980a958961629649769613adfd40a9cb0d505/packages/platform-browser/src/browser.ts#L173) by [bootstrapApplication](https://github.com/angular/angular/blob/d9c980a958961629649769613adfd40a9cb0d505/packages/platform-browser/src/browser.ts#L127-L134))

projects/schematics/src/shared/utils/new-module-utils_spec.ts
```diff
@@ -101,42 +101,52 @@ describe('New Module utils', () => {
...
      * ],
-     * providers:
+     * providers: (or })
      * ```
      */
     function doesNgModuleImportsContainModule(moduleName: string): RegExp {
-      return new RegExp(`imports:[\\s\\S]*${moduleName}[\\s\\S]*providers:`);
+      return new RegExp(
+        `imports:[\\s\\S]*${moduleName}[\\s\\S]*(providers:|\\}\\))`
+      );
```
Platonn, line 118: Note: The util was not working correctly before, when there was no `providers` array following the `imports` array, but just the ending `}`. Because of it, our unit tests failed (because AppModule now has no `providers:[]` array).
Now the util is more robust to account for this special case of absent `providers`

projects/schematics/src/add-cms-component/index_spec.ts
```diff
@@ -281,13 +279,7 @@ describe('add-cms-component', () => {
       );
       assertContentExists(
         appTree,
-        [`declarations: [`, `DummyComponent,`, `MyAwesomeCmsComponent`],
-        existingModulePath
-      );
-      assertContentExists(appTree, [], existingModulePath);
-      assertContentExists(
-        appTree,
-        [`exports: [`, `DummyComponent,`, `MyAwesomeCmsComponent`],
```
pawelfras, line 290: `query:` isn't the case that only `MyAwesomeCmsComponent` should be removed from `declarations` array`?

projects/schematics/src/add-spartacus/index.ts
```diff
@@ -339,22 +335,25 @@ function verifyAppModuleExists(options: SpartacusOptions): Rule {
...
+Please re-create your application:
 1. remove your application code
-2. make sure to pass the flag "--standalone=false" to the command "ng new". For more, see https://angular.io/cli/new#options
-3. try again installing Spartacus with a command "ng add @spartacus/schematics" ...
-
-Note: Since version 17, Angular's command "ng new" by default creates an app without a file "app.module.ts" (in a so-called "standalone" mode). But Spartacus installer requires this file to be present.
-`
+2. make sure to pass the flag "--standalone=false --file-name-style-guide=2016" to the command "ng new". For more, see https://angular.io/cli/new#options
+3. try again installing Spartacus with a command "ng add @spartacus/schematics" ...`
       );
```
pawelfras: `major:` if I'm not mistaken, we should not check at this stage if `app.module.ts` exists, because we want from customers to generate Angular apps without `--standalone=false`. The `app.module.ts` is somethint that will be added later, by our schematics.

That said, I think we should also adjust error messsage to not contain info about `app.module.ts` and `--standalone=false`.

projects/schematics/src/add-spartacus/update-app-config.ts
```diff
@@ -0,0 +1,110 @@
...
+      providersArray.addElement(providerContent);
+    }
+  }
+}
+
+function addProvideHttpClient(sourceFile: SourceFile): void {
+  // Add imports for provideHttpClient, withFetch, withInterceptorsFromDi
+  createImports(sourceFile, [
+    {
+      moduleSpecifier: '@angular/common/http',
```
pawelfras: `minor:` consider using `ANGULAR_HTTP` from `constants.ts`

projects/schematics/src/add-ssr/update-app-config-server.ts
```diff
@@ -0,0 +1,140 @@
...
+  // Remove withRoutes argument from provideServerRendering
+  const providersArray = getAppConfigServerProviders(sourceFile);
+
+  if (providersArray && Node.isArrayLiteralExpression(providersArray)) {
+    const elements = providersArray.getElements();
+
+    elements.forEach((element) => {
+      if (Node.isCallExpression(element)) {
+        const expression = element.getExpression();
+        if (expression.getText() === 'provideServerRendering') {
```
pawelfras, line 125: Note there is a function that doid the same for module-based apps:
https://github.com/SAP/spartacus/blob/30f81e5be6126b651fe885474e73eef6a68f1b12/projects/schematics/src/add-ssr/index.ts#L535

Let's remove it if no longer needed.

projects/schematics/src/add-cms-component/index_spec.ts
Platonn: DummyComponent doeesn't exist in declarations array ,according to the new tests, setup. In new tests setup even the DummyComponent has been created as standalone, therefore it doesn't pre-exist in the declarations array

projects/schematics/src/add-ssr/update-app-config-server.ts
Platonn: Oh, thanks for letting me know.
Important thing:
 - previously we removed whole `provideServerRendering(withRoutes(serverRoutes))`
 - now we're removing only the argument `withRoutes(serverRoutes))`, but `provideServerRendering()` stays, which is justified. [It provides Angular's PLATFORM_SERVER_PROVIDERS`](https://github.com/angular/angular/blob/d9c980a958961629649769613adfd40a9cb0d505/packages/platform-server/src/provide_server.ts#L13-L34), which is an equivalent to the dropped [ServerModule](https://github.com/angular/angular/blob/d9c980a958961629649769613adfd40a9cb0d505/packages/platform-server/src/server.ts#L83-L92).

projects/schematics/src/add-spartacus/index.ts
Platonn: Oh, good catch. Thank you!
Now changed the check to make more sense:
- Checks if the app is standalone (by checking existence of app.config.ts).
- Checks if the app uses Angular 2016 filename convention (by checking `.component.ts` suffix in app.component.ts)

projects/schematics/src/add-spartacus/update-app-config.ts
Platonn: done

projects/schematics/src/add-ssr/update-app-config-server.ts
Platonn: fixed. now removed all the now-unnecessary logic that previously editing app.module.server.ts

projects/schematics/src/ng-add/__snapshots__/index_spec.ts.snap
pawelfras, line 1: `minor:` let's add additional test cases that checks also body od `app.config.ts` and `app.server.config.ts`, since after changes most of the configuration is placed there instead of modules and having it might be helpful in future Angular upgrades.

Platonn: I'm not convinced yet that we need new snapshot tests for this file.

`ng-add` calls `add-spartacus` (and then also `add-ssr`, if `--ssr flag` was passed), see source code:
https://github.com/SAP/spartacus/blob/8bb7b838618176038a16c4bdf7fed7f52f022d7b/projects/schematics/src/ng-add/index.ts#L19-L21

And those nested schematics are already covered with tests:
- add-spartacus spec: https://github.com/SAP/spartacus/blob/8bb7b838618176038a16c4bdf7fed7f52f022d7b/projects/schematics/src/add-spartacus/index_spec.ts#L683-L736
- add-ssr spec: https://github.com/SAP/spartacus/blob/8bb7b838618176038a16c4bdf7fed7f52f022d7b/projects/schematics/src/add-ssr/index_spec.ts#L77-L143

pawelfras: Thanks for the answer! Indeed, all these files are covered by tests for `add-spartacus` and `add-ssr`. Apologies for the confusion.

pawelfras (approved):
Review and QA done.
What was tested
- [x] installation in fresh Angular app, with CSR  
- [x] installation in fresh Angular app, with SSR
- [x] migration of Spartacus app, with CSR
- [x] migration of Spartacus app, with SSR
- [x] build of storefrontapp in various combination
- [x] serving storefrontapp in various combination

LGTM!
Before merging, please run the branch against [Spartacus build pipeline](https://github.com/SAP/spartacus/actions/workflows/ci.yml) to run all e2es

PR #21078, opened 2026-01-27
fix: get rid of noise in logs

Pucek9:


projects/storefrontapp/src/app/private/ssr/get-chrome-devtools-express-middleware.ts
Platonn, line 1: 1. Let's align the filename with the exported funciton:
`get-chrome-devtools-express-middleware.ts`

2. And let's move it to the recently-added  `/private` folder, e.g. nested in `/private/ssr`:

Image

Platonn:
Do we still need this PR or we can close it without merging?
I don't remember, did Angular 20/21 internals already handle this case and we already benefit form it since the Feb 2026 Framework Update release?

Pucek9:
@Platonn I don't think the issue has been fixed on our side. We still don't use a modern SSR (Vite), so eg. [this](https://github.com/ChromeDevTools/vite-plugin-devtools-json) plugin can't work.

PR #21081, opened 2026-01-27
fix: Refresh neasted component when update in smart edit

rmch91:
Closes: CXSPA-11046


Pucek9 (dismissed):

Pucek9:
QA to do after deploy only, right?

projects/core/src/cms/facade/cms.service.ts
```diff
@@ -291,4 +291,20 @@ export class CmsService {
...
+
+  /**
+   * Clear the cached component data.
+   * This is useful when components need to be forcefully reloaded,
+   * for example in SmartEdit context when child components need to be refreshed.
+   *
+   * @param uids Optional array of component UIDs to clear. If not provided, clears all cached components.
+   */
+  clearComponentState(uids?: string[]): void {
+    const componentUids = uids || Object.keys(this.components);
```
StanislavSukhanov: Since we check against `undefined` would it be more appropriate to use Nullish Coalescing Operator here?

rmch91: changed

StanislavSukhanov (approved):

PR #21096, opened 2026-01-30
feat: add mediaApiPrefix env var and backend.media.prefix config

uroslates:
Closes: https://jira.tools.sap/browse/CXSPA-12005

Adding 
- new Spartacus **backend.media.prefix** (hidden behind **enableMediaPrefix** feature flag) - similar to **backend.occ.prefix**.  The new config is used when constructing the media baseUrl (added to the configured host as suffix === uri).
- new Spartacus **environment(-prod).ts** file via  **mediaBaseUrl** and  **mediaApiPrefix**  used for configuring **backend.media.baseUrl** and **backend.media.prefix**  respectively.


projects/storefrontapp/src/environments/environment.ts
```diff
@@ -23,6 +23,8 @@ export const environment: Environment = {
   production: false,
   occBaseUrl: buildProcess.env.CX_BASE_URL,
   occApiPrefix: '/occ/v2/',
+  // mediaBaseUrl: buildProcess.env.CX_BASE_URL,
+  // mediaApiPrefix: '',
```
StanislavSukhanov: is it commented intentionally?

projects/core/src/occ/config/default-occ-config.ts
```diff
@@ -11,6 +11,8 @@ export const defaultOccConfig: OccConfig = {
     occ: {
       prefix: '/occ/v2/',
     },
```
StanislavSukhanov: media is `optional` do wen need to explicitly have it in config?

projects/storefrontlib/shared/components/media/media.service.ts
```diff
@@ -401,10 +407,42 @@ export class MediaService {
...
+          );
+        }
+        return baseUrl;
+      }
+    } else {
+      return baseUrl;
+    }
+  }
+
+  private getPrefix(): string {
```
StanislavSukhanov: let's remove this it method since `new URL(prefix, baseUrl)` will do it for us

projects/storefrontlib/shared/components/media/media.service.ts
```diff
@@ -401,10 +407,42 @@ export class MediaService {
...
+        } else {
+          this.logger.warn(
+            'Invalid media/occ baseUrl and prefix configuration'
+          );
+        }
+        return baseUrl;
+      }
+    } else {
+      return baseUrl;
+    }
```
StanislavSukhanov, line 417:
```suggestion
   const baseUrl =
      this.config.backend?.media?.baseUrl ??
      this.config.backend?.occ?.baseUrl ??
      '';

    if (!this.featureConfigService.isEnabled('enableMediaPrefix')) {
      return baseUrl;
    }

    try {
      // Construct URL (with Validation)
      const mediaPrefix = this.config?.backend?.media?.prefix ?? '';
      const url = new URL(mediaPrefix, baseUrl);
      return url.toString();
    } catch (error) {
      if (isDevMode()) {
        this.logger.error(
          'Invalid media/occ baseUrl and prefix configuration'
        );
      } else {
        this.logger.warn(
          'Invalid media/occ baseUrl and prefix configuration'
        );
      }
      return baseUrl;
    }
    }
```

StanislavSukhanov (commented):
Hey @uroslates great job! Left some comments/suggestions

projects/core/src/occ/config/default-occ-config.ts
```diff
@@ -11,6 +11,8 @@ export const defaultOccConfig: OccConfig = {
...
-    media: {},
+    media: {
+      prefix: '',
+    },
```
StanislavSukhanov: `media` seems to be optional. Do we need to explicitly put it here? 🤔

StanislavSukhanov (changes requested):

projects/storefrontapp/src/environments/environment.ts
uroslates: Yes just because I wanted to make awareness that these configs exist. But most probably it should be deleted from here (since its visible in model definition (should be documented as well) 
cc Just for awareness @matt-williams8

projects/storefrontlib/shared/components/media/media.service.ts
uroslates: 👍 nice catch! Applied!

projects/storefrontlib/shared/components/media/media.service.ts
uroslates: 👍  Applied

projects/core/src/occ/config/default-occ-config.ts
uroslates: 👍 Applied.

projects/storefrontapp/src/environments/environment.ts
StanislavSukhanov: let's remove it. 🙏

StanislavSukhanov (changes requested):

StanislavSukhanov (dismissed):
Legit.

StanislavSukhanov (approved):

PR #21104, opened 2026-02-02
chore: Remove deprecated media token

rmch91:
Closes: https://jira.tools.sap/browse/CXSPA-11241


docs/migration/221121_7/221121_7-typescript-manual.doc.md
```diff
@@ -106,6 +106,25 @@
...
+  - If you need to use `img` HTML element instead of `picture`, pass `[elementType]="'img'"` as an input to the `MediaComponent`.
+  - Remove any providers for this token from your module or component configurations.
+
+### `useLegacyMediaComponent` Config Property
+- **Removed From:** `MediaConfig` in `projects/storefrontlib/shared/components/media/media.config.ts`
+- **Reason:** This config property was deprecated since version 2211.31 and has now been removed.
+- **Impact:** The config property was used to globally force the `MediaComponent` to use the legacy `img` element instead of the `picture` element.
+- **Action Required:**
+  - Remove `useLegacyMediaComponent` from your config if you were using it.
+  - If you need to use `img` HTML element instead of `picture`, pass `[elementType]="'img'"` as an input to the `MediaComponent` where needed.
```
StanislavSukhanov, line 126: Hey Roman. 
I believe this can be misleading, because the `MediaComponent`'s `elementType` input defaults to 'img', 
hence it will render an `` instead of `picture` by default

StanislavSukhanov (changes requested):

docs/migration/221121_7/221121_7-typescript-manual.doc.md
rmch91: Not really, with this config, user be default received pictures tags everywhere. 
Imagine that user want img, he set useLegacyMediaComponent: true. Now he upgrade, but we removed this config, but user still want img everywhere, in banner also, and we are recommending him `If you need to use `img` HTML element instead of `picture`, pass `[elementType]="'img'"` as an input to the `MediaComponent` where needed.`

StanislavSukhanov (approved):

PR #21109, opened 2026-02-03
fix: ASM load issue when early login enabled.

i336588:


feature-libs/asm/root/services/asm-protected-routes.service.ts
```diff
@@ -0,0 +1,49 @@
...
+ * allowing access to CMS pages without customer emulation.
+ *
+ * Overrides `ProtectedRoutesService` when ASM module is enabled.
+ */
+@Injectable({ providedIn: 'root' })
+export class AsmProtectedRoutesService extends ProtectedRoutesService {
+  protected csAgentAuthService = inject(CsAgentAuthService);
+
+  constructor(
+    protected override config: RoutingConfig,
```
i53577: let's use inject() instead of constructor

i53577 (approved):
LGTM

PR #21113, opened 2026-02-03
chore/CXSPA-11238: refactor feature toggle scripts to use common func…

Pucek9:
…tions


Pucek9:
Closed because there will be no reason to change these scripts in the future.

PR #21117, opened 2026-02-04
chore: removed the deprecated 'OptimizedSsrEngine.log' method in favor of 'this.logger'

Platonn:
Before:
- method `OptimizedSsrEngine.log` was deprecated since more than year ago
- the deprecated method contained a legacy 2nd argument (which was useless and ignored, which disturbed readability, but we couldn't remove it until the method was `protected`, i.e. in public APi)

After:
- now removed the deprecated method and replaced its usages `this.log()` with `this.logger.log()`
- changed `this.logger` from `private` to `protected`
- updated unit tests which expected `this.log` to be called. Now they expect `this.logger.log` to be called instead
- updated migration doc on manual Typescript breaking changes with advice to use `this.logger` object instead

fixes https://jira.tools.sap/browse/CXSPA-12038

### DIFF
--- a/docs/migration/221121_7/221121_7-typescript-manual.doc.md
+++ b/docs/migration/221121_7/221121_7-typescript-manual.doc.md
@@ -106,10 +106,10 @@
 - **Reason:** Deprecated in favor of `CdcPreferenceSerializer` service class.
 - **Action Required:** Use the `CdcPreferenceSerializer` class methods.
 
-### `OptimizedSsrEngine` Logger Method
+### `OptimizedSsrEngine`
 - **Removed Method:** `log(message: string, _ignoredLegacyDebugParameter: boolean, context: ExpressServerLoggerContext)`
 - **Reason:** Deprecated since **2211.27**
-- **Action Required:** If you extended the `OptimizedSsrEngine` and used called yourself the removed method `log()`, please update your code to use the property `this.logger` instead, e.g. `this.logger.log(message, context);`
+- **Action Required:** Use `this.logger` instead, e.g. `this.logger.log(message, context);`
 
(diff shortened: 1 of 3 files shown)


Pucek9 (dismissed):

docs/migration/221121_7/221121_7-typescript-manual.doc.md
```diff
@@ -106,6 +106,11 @@
 - **Reason:** Deprecated in favor of `CdcPreferenceSerializer` service class.
 - **Action Required:** Use the `CdcPreferenceSerializer` class methods.
 
+### `OptimizedSsrEngine`
+- **Removed Method:** `log(message: string, _ignoredLegacyDebugParameter: boolean, context: ExpressServerLoggerContext)`
+- **Reason:** Deprecated since **2211.27**
+- **Action Required:** Use `this.logger` instead, e.g. `this.logger.log(message, context);`
```
pawelfras, line 112: currently `logger` is a private member of the `OptimizedSsrEngine` without possibility of using it in a customer's class extending our engine. If it is desired  to expose the logger, let's change access modifier to `protected` or let's remove `action required` section.

Platonn: good catch, thanks!
fixed: changed logger to protected

pawelfras (approved):
Review and QA done. LGTM

docs/migration/221121_7/221121_7-typescript-manual.doc.md
Pucek9: Nice catch, Paweł! 
Sorry, I didn't expand the unchanged code part

PR #21129, opened 2026-02-09
feat: Implement Tokenization for Saved Payment Details (My Account & Checkout)

Diksha-28:
This PR focuses on implementing tokenisation. The following references are provided to help understand the scope, flow, and implementation details of the items included in this epic.

Epic : [CXOPF-4039](https://jira.tools.sap/browse/CXOPF-4039) Tokenization: Saved Payment Details on Account from PSP payment form during checkout.

Design Document : https://wiki.one.int.sap/wiki/pages/viewpage.action?pageId=5748553978&spaceKey=SAPCCWiki&title=CXOPF-4039%2BTokenization%2BSaved%2BPayment%2BDetails%2Bon%2BAccount%2Bfrom%2Bcheckout.

This PR includes changes for the following Jira tickets :
[CXSPA-11820](https://jira.tools.sap/browse/CXSPA-11820) List saved cards on My Accounts page.
[CXSPA-11821](https://jira.tools.sap/browse/CXSPA-11821) Delete saved cards on My Accounts.
[CXSPA-11822](https://jira.tools.sap/browse/CXSPA-11822) List saved cards on checkout page.
[CXSPA-12077](https://jira.tools.sap/browse/CXSPA-12077) Use saved card to place order.

I've added screen recordings to each Jira ticket for reference, in case you'd like to see how it works in the UI.


integration-libs/opf/tokenisation/components/opf-tokenisation-payment-methods/opf-tokenisation-payment-methods.module.ts
```diff
@@ -0,0 +1,50 @@
...
+  ],
+  providers: [
+    // Register a tokenisation-specific CMS component type.
+    // NOTE: This is a new CMS type and intentionally does NOT override
+    // the existing `AccountPaymentDetailsComponent`. The backend must
+    // place `OpfTokenisationPaymentMethodsComponent` on pages where
+    // tokenisation behavior is desired.
+    provideDefaultConfig(<CmsConfig>{
+      cmsComponents: {
+        OpfTokenisationPaymentMethodsComponent: {
```
Matejk00: As we talked, you can re-use existing CMS Component (AccountPaymentDetailsComponent) for this feature.

Matejk00 (changes requested):
In general looks good, waiting for the final implementation.

integration-libs/opf/checkout/components/opf-checkout-payments/opf-checkout-payments.component.html
```diff
@@ -9,6 +9,85 @@ <h3 *ngIf="isHeadingDisplayed">
     <ng-container *ngIf="!activeConfigurationsState?.loading; else loading">
       <div class="cx-payment-options-list">
         <ng-content></ng-content>
+
+        <!-- Saved Payment Options Section (only shown if saved cards exist) -->
+        <ng-container *ngIf="hasSavedCards$ | async">
```
Matejk00: This all additional logic should come from the tokenisation library and be injected via an outlet, for example using a dedicated opf-saved-cards-like component (name for illustration only). This ensures that if the tokenisation library is not installed, the checkout flow remains unchanged and behaves as it did before.

integration-libs/opf/checkout/components/opf-checkout-payments/opf-checkout-payments.component.html
```diff
@@ -9,6 +9,85 @@ <h3 *ngIf="isHeadingDisplayed">
...
+                <span>{{
+                  savedCardsHeadingKey || 'opfCheckout.cards' | cxTranslate
+                }}</span>
+              </div>
+            </label>
+          </div>
+
+          <!-- Saved Cards List Container (shown when saved cards radio is selected) -->
+          <ng-container *ngIf="selectedPaymentId === SAVED_CARDS_ID">
+            <ng-template [cxOutlet]="opfCheckoutOutlets.OPF_TOKENISATION">
```
Matejk00: This outlet is currently named `OPF_TOKENISATION`, but it’s positioned before the payment options, so the name shouldn’t be feature-specific. I’d suggest renaming it to a position-based name such as `opfCheckoutOutlets.OPF_CHECKOUT_BEFORE_PAYMENT_OPTIONS`, so the checkout template defines the slot and any feature (including tokenisation) can inject content there.

integration-libs/opf/checkout/components/opf-checkout-payments/opf-checkout-payments.component.html
```diff
@@ -22,6 +101,7 @@ <h3 *ngIf="isHeadingDisplayed">
                   class="form-check-input"
                   role="radio"
                   type="radio"
+                  name="paymentId"
```
Matejk00, line 32: Do we need to provide this `name` attribute?

integration-libs/opf/checkout/components/opf-checkout-payments/opf-checkout-payments.component.ts
```diff
@@ -116,16 +132,26 @@ export class OpfCheckoutPaymentsComponent implements OnInit, OnDestroy {
   @Input()
   noRenderPaymentWrapperMessage?: string;
 
+  readonly SAVED_CARDS_ID = -1;
```
Matejk00: So I believe these properties should also be part of the tokenisation component or its service.

integration-libs/opf/tokenisation/components/opf-tokenisation-checkout-payment-method/opf-tokenisation-checkout-payment-method.component.ts
```diff
@@ -0,0 +1,330 @@
...
+  );
+
+  get existingPaymentMethods$(): Observable<PaymentDetails[]> {
+    return this.userPaymentService.getPaymentMethods();
+  }
+  // instead of reading from checkoutPaymentFacade (which depends on the backend returning paymentInfo), we are using the local subject
+  get selectedMethod$(): Observable<PaymentDetails | undefined> {
+    return this.selectedPaymentMethod$.asObservable();
+  }
+  constructor(
```
Matejk00: Would it be possible to use the Angular inject() function here instead of constructor injection (for example: `protected userPaymentService = inject(UserPaymentService);`)?

integration-libs/opf/tokenisation/components/opf-tokenisation-checkout-payment-method/opf-tokenisation-checkout-payment-method.component.ts
```diff
@@ -0,0 +1,330 @@
...
+      text: [paymentDetails.cardNumber ?? '', cardLabels.textExpires],
+      actions: isSelected
+        ? []
+        : [{ name: cardLabels.textUseThisPayment, event: 'send' }],
+      header: isSelected ? cardLabels.textSelected : undefined,
+      label: paymentDetails.defaultPayment
+        ? 'paymentCard.defaultPaymentLabel'
+        : 'paymentCard.additionalPaymentLabel',
+    };
+  }
```
Matejk00, line 284: There’s quite a lot of business logic and many methods in this component. Consider extracting this logic into a dedicated component service, so that the component itself focuses only on calling methods from the service and handling the view.

integration-libs/opf/tokenisation/core/tokens/index.ts
```diff
@@ -0,0 +1,7 @@
+/*
+ * SPDX-FileCopyrightText: 2026 SAP Spartacus team <spartacus-team@sap.com>
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+// export * from './tokens';
```
Matejk00: Should this be commented out?

Matejk00 (changes requested):

integration-libs/opf/checkout/components/opf-checkout-payments/opf-checkout-payments.component.html
Diksha-28: First, thanks for the feedback — the direction you're suggesting makes good architectural sense, and I agree with the principle. I actually wanted to take the same approach. Let me explain why I ended up going with the current one and the constraints that led to that decision.
 
Context
 
I introduced the saved cards section to support the new Figma designs, which require showing saved payment methods alongside active PSP options in the same payment selection flow. Because of that UX requirement, I had to make a deliberate decision about where this logic should live.
 
Why I kept it in the checkout component
 
I kept this logic in the checkout component mainly because of the radio button mutual exclusivity. The saved cards radio button and all PSP radio buttons share the same name="paymentId" and formControlName="paymentId" within the same Angular FormGroup. This ensures they remain mutually exclusive both at the browser level and within Angular reactive forms.
If I moved the saved cards radio into a separate outlet-injected component, it wouldn’t have access to the parent’s FormGroup, which would break that mutual exclusivity.
 
When a user selects a saved card and then switches to a PSP, it’s not just a radio button change. We also trigger checkoutPaymentFacade.deletePaymentDetails() to remove the existing payment info from the cart before the new PSP wrapper loads.
This side effect needs close coordination between both parts of the flow, which a simple outlet-injected component wouldn’t handle out of the box.
 
Non-trivial design work
Doing this properly would require a dedicated service contract interface , a concrete implementation in the tokenisation library, and an agreed outlet context API for two-way state communication. I didn’t want to rush that and risk introducing subtle bugs.

For instance, imagine the page has a list of payment options — credit card, PayPal, whatever — and right at the top there's also a 'Saved Cards' option. All of these are radio buttons, and only one can be selected at a time. Right now they all live in the same component, so that just works automatically.
If I move the saved cards part into a separate injected component, they're no longer in the same place — so I need to manually wire up the communication between them. For example, if I select PayPal, the saved cards component needs to know to deselect itself. And if I had a saved card selected and then switch to PayPal, I also need to make an API call to remove that saved card from the cart — otherwise the order could get placed with the wrong payment.
 
No regression guarantee
 
When the tokenisation library isn’t installed, getPaymentMethods() returns [], hasSavedCards$ emits false, and the saved cards section is never rendered. So the checkout flow behaves exactly as it did before — this is already covered. You can check the Jira tickets where I’ve attached the testing recordings.
 
But yes, we can schedule a call to align on the appropriate approach for a proper follow-up.

integration-libs/opf/tokenisation/core/tokens/index.ts
Diksha-28: fixed

integration-libs/opf/checkout/components/opf-checkout-payments/opf-checkout-payments.component.html
Diksha-28: No, it was redundant. Removed it since we already have formControlName="paymentId"

integration-libs/opf/tokenisation/components/opf-tokenisation-checkout-payment-method/opf-tokenisation-checkout-payment-method.component.ts
Diksha-28: yes, done.

integration-libs/opf/checkout/components/opf-checkout-payments/opf-checkout-payments.component.html
Diksha-28: done

integration-libs/opf/tokenisation/components/opf-tokenisation-checkout-payment-method/opf-tokenisation-checkout-payment-method.component.ts
Diksha-28: done.

integration-libs/opf/checkout/components/opf-checkout-payments/opf-checkout-payments.component.html
```diff
@@ -9,6 +9,36 @@ <h3 *ngIf="isHeadingDisplayed">
...
+            savedCardsId: SAVED_CARDS_ID,
+            disabled:
+              (paymentDisabled$ | async) ||
+              (disabled && explicitTermsAndConditions),
+            savedCardsSelected: onSavedCardsSelected.bind(this),
+          }"
+        >
+        </ng-template>
+
+        <!-- New Payment Options Heading -->
```
Matejk00: Should this also be part of the tokenization component? I imagine that if customers do not want to use tokenization, they also do not need the additional “New Payment Methods” header. So it should probably be included together with the tokenization component and loaded from the module only when it is enabled.

integration-libs/opf/checkout/components/opf-checkout-payments/opf-checkout-payments.component.ts
```diff
@@ -116,16 +132,22 @@ export class OpfCheckoutPaymentsComponent implements OnInit, OnDestroy {
...
+
+  readonly opfCheckoutOutlets = OpfCheckoutOutlets;
+
+  readonly iconTypes = ICON_TYPE;
```
Matejk00: Let’s move as many of these properties as possible to the tokenization component and pass through the outlet context only the data that is truly needed and cannot be handled independently or fetched by the tokenization component.

integration-libs/opf/checkout/components/opf-checkout-payments/opf-checkout-payments.component.ts
```diff
@@ -271,12 +293,35 @@ export class OpfCheckoutPaymentsComponent implements OnInit, OnDestroy {
...
+    this.isSavedCardsSelected = true;
+  }
+
   changePayment(payment: OpfActiveConfiguration): void {
-    this.selectedPaymentId = payment.id;
-    this.opfMetadataStoreService.updateOpfMetadata({
-      selectedPaymentOptionId: this.selectedPaymentId,
-    });
-    this.paymentChange.emit(payment);
+    if (this.isSavedCardsSelected) {
```
Matejk00: > Let’s move as many of these properties as possible to the tokenization component and pass through the outlet context only the data that is truly needed and cannot be handled independently or fetched by the tokenization component.

The same applies to the logic. Ideally, `OpfCheckoutPaymentsComponent` should not introduce any changes other than an additional outlet position template and the data required for the context. Everything else should be moved to a tokenization service and become part of the tokenization module. We do not need tokenization-related code in the customer JavaScript bundle if they are not using this feature.

integration-libs/opf/checkout/styles/components/_opf-checkout-payments.scss
```diff
@@ -1,4 +1,12 @@
 %cx-opf-checkout-payments {
+  .cx-payment-section-heading {
```
Matejk00: Can we create a separate SCSS file, for example `_opf-tokenisation.scss`, and move the styling there, as we do for other sub-features?

integration-libs/opf/tokenisation/components/opf-tokenisation-payment-method/opf-tokenisation-payment-method.component.spec.ts
```diff
@@ -0,0 +1,586 @@
+// import { Component, Input, Type } from '@angular/core';
```
Matejk00: Please provide unit tests for this component and for everything under the tokenization library.

integration-libs/opf/tokenisation/components/opf-tokenisation-checkout-payment-method/opf-tokenisation-checkout-payment-method.component.ts
```diff
@@ -0,0 +1,87 @@
...
+    NgIf,
+    NgFor,
+    CardComponent,
+    SpinnerComponent,
+    AsyncPipe,
+    TranslatePipe,
+  ],
+  providers: [OpfTokenisationCheckoutPaymentMethodService],
+})
+export class OpfTokenisationCheckoutPaymentMethodComponent
```
Matejk00: I think we can remove the the "checkout" part from the name.

integration-libs/opf/tokenisation/components/opf-tokenisation-payment-methods/opf-tokenisation-payment-methods.component.ts
```diff
@@ -0,0 +1,93 @@
...
+export class OpfTokenisationPaymentMethodsComponent implements OnInit {
+  paymentMethods$: Observable<PaymentDetails[]>;
+  editCard: string | undefined;
+  loading$: Observable<boolean>;
+  @Input() showHeader = true;
+
+  constructor(
+    private tokenisationFacade: OpfTokenisationFacade,
+    private translation: TranslationService,
+    protected globalMessageService?: GlobalMessageService
```
Matejk00: Can we use an `inject` method here?

Matejk00 (changes requested):

integration-libs/opf/tokenisation/components/opf-tokenisation-payment-methods/opf-tokenisation-payment-methods.component.ts
Diksha-28: done

integration-libs/opf/checkout/components/opf-checkout-payments/opf-checkout-payments.component.html
Matejk00: Isn't this `name="paymentId"` attribute still in PR?

integration-libs/opf/checkout/components/opf-checkout-payments/opf-checkout-payments.component.ts
```diff
@@ -253,12 +263,19 @@ export class OpfCheckoutPaymentsComponent implements OnInit, OnDestroy {
...
+            this.emitOutletContext();
           } else if (
             !state.termsAndConditionsChecked &&
             this.explicitTermsAndConditions
           ) {
             isPreselected = false;
             this.selectedPaymentId = undefined;
+            this.emitOutletContext();
+          } else if (isPreselected) {
+            // After initial preselection, continue listening for payment option changes
```
Matejk00: I think we can remove those comments.

integration-libs/opf/tokenisation/components/public_api.ts
```diff
@@ -0,0 +1,9 @@
+/*
+ * SPDX-FileCopyrightText: 2026 SAP Spartacus team <spartacus-team@sap.com>
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+export * from './opf-tokenisation-components.module';
+export * from './opf-tokenisation-account-payment-methods/opf-tokenisation-account-payment-methods.component';
```
Matejk00: I think we can rename the folders from `opf-tokenisation-account-payment-methods` to `account-payment-methods` and from `opf-tokenisation-payment-method` to `payment-method`. We do not need to duplicate the name, since these components are already inside the opf-tokenisation folder.

Matejk00: Please also do the same for all tokenisation components folders. For example `opf-tokenisation-saved-cards-toggle` -> `saved-cards-toggle`

integration-libs/opf/tokenisation/core/services/opf-tokenisation-saved-cards.service.ts
```diff
@@ -0,0 +1,105 @@
...
+          distinctUntilChanged(),
+          pairwise()
+        )
+        .subscribe(([prev, curr]) => {
+          // Only delete if:
+          // 1. We're transitioning away from SAVED_CARDS_ID
+          // 2. A card was actually selected from the saved cards list
+          // 3. We're not going to an undefined state
+          if (
+            prev === SAVED_CARDS_ID &&
```
Matejk00: Can this statement logic be inside a function? It looks quite complex.

Matejk00 (changes requested):

Matejk00:
Please also make the pipeline checks green.

integration-libs/opf/checkout/components/opf-checkout-payments/opf-checkout-payments.component.html
Diksha-28: Yes, we need it. The name attribute groups these radio buttons at the HTML level, so only one can be selected at a time. Plus, it's needed for proper accessibility.

integration-libs/opf/schematics/add-opf/schema.json
```diff
@@ -82,6 +84,11 @@
             "value": "OPF-Quick-Buy",
             "label": "OPF Quick Buy",
             "checked": true
+          },
+          {
+            "value": "OPF-Tokenisation",
+            "label": "OPF Tokenisation",
+            "checked": true
```
Matejk00: Should the OPF tokenisation feature be enabled by default for the customers during the schematics installation?

integration-libs/opf/tokenisation/components/opf-tokenisation-payment-method/opf-tokenisation-payment-method.service.ts
```diff
@@ -0,0 +1,343 @@
...
+  OpfTokenisationSavedCardsService,
+} from '@spartacus/opf/tokenisation/core';
+
+@Injectable()
+export class OpfTokenisationPaymentMethodService {
+  protected userPaymentService = inject(UserPaymentService);
+  protected checkoutPaymentFacade = inject(CheckoutPaymentFacade);
+  protected opfMetadataStoreService = inject(OpfMetadataStoreService);
+  protected savedCardsService = inject(OpfTokenisationSavedCardsService);
+  protected busy$ = new BehaviorSubject<boolean>(false);
```
Matejk00: Could we avoid mixing `inject` properties with the other variable types? Let’s keep them always at the top of the class declaration.

Matejk00 (dismissed):

Matejk00 (approved):

Matejk00:
TME and TM status is available under the epic ticket: https://jira.tools.sap/browse/CXOPF-4039

PR #21141, opened 2026-02-10
fix: fix searchbox carousel navigation behaviour

c5346163:
Fix Safari searchbox carousel clicks closing results panel by preventing mousedown blur and moving next-slide logic into a reusable handler.


Pucek9 (dismissed):
LGTM

QA steps I did:
On Mac, opened the homepage and decreased the window height, typed "cam" in searchbox and saw carousel arrows; both work fine and no longer close the search results
image

rmch91 (approved):

projects/storefrontlib/cms-components/navigation/search-box/search-box.component.html
```diff
@@ -150,6 +150,7 @@ <h3 *ngIf="result.products?.length">
         [template]="carouselItem"
         itemWidth="160px"
         (keybordEvent)="carouselEventPropagator($event)"
+        (mousedown)="preventDefault($event)"
```
rmch91: Actually this can be breaking change for customers, example:
Customer registered event listener, something like: `this.el.nativeElement.addEventListener('focus', this.onFocus, true);` and implement some custom logic on focus. 
When you are adding `(mousedown)="preventDefault($event)"` it will not be triggered.

rmch91 (dismissed):

projects/storefrontlib/cms-components/navigation/search-box/search-box.component.html
Pucek9: @rmch91 That's true, but the question that comes to mind is why would he do it in such a complicated way? If he extends the class, that forces him to have its own HTML structure (which will not be automatically updated from here). The fact that we cannot fix the behavior in his code - it's beyond our control if there is any customization, the question is only whether we are able to break something in his code/ui behavior this way?

rmch91: User can do it without any customization of the component. just register an event listener on cx-carousel in app.component and apply his own logic. 
After this change it will not work. So seems that it is breaking

c5346163: I've made the fix opt-in via a preventNavigationFocus input flag (defaults to false), so it's only enabled in search-box and doesn't affect existing carousels or customer event listeners.

projects/core/src/features-config/feature-toggles/config/feature-toggles.ts
```diff
@@ -475,6 +485,7 @@ export const defaultFeatureToggles: Required<FeatureTogglesInterface> = {
   a11yFacetFilterByLabel: false,
   removeDuplicatedOrderHistoryHeader: false,
   a11yCardNotificationMessage: false,
+  a11yCarouselPreventNavigationFocus: true,
```
Pucek9: As far as I know, for the introduction of toggles to make sense, they should be introduced with a false value. Some time after their introduction, they remain in place: first, they are changed to true, then, before the release of the next major version, the flags and legacy code are cleared.
```suggestion
  a11yCarouselPreventNavigationFocus: false,
```

projects/storefrontlib/shared/components/carousel/carousel.component.html
```diff
@@ -2,23 +2,37 @@
...
+        (keydown.arrowright)="shareEvent($any($event))"
+        (keydown.arrowup)="shareEvent($any($event))"
+        (keydown.arrowdown)="shareEvent($any($event))"
+      >
+        <cx-icon [type]="previousIcon"></cx-icon>
+      </button>
+      <button
+        *cxFeature="'!a11yCarouselPreventNavigationFocus'"
+        class="previous"
+        (click)="onPreviousClick($event, size)"
```
Pucek9: Old block should remain unchanged
```suggestion
        (click)="
          $event.stopPropagation();
          activeSlide === 0 ? null : (activeSlide = activeSlide - size)
        "
```

projects/storefrontlib/shared/components/carousel/carousel.component.html
```diff
@@ -54,25 +68,37 @@ <h2 *ngIf="title">{{ title }}</h2>
...
+        (keydown.arrowright)="shareEvent($any($event))"
+        (keydown.arrowup)="shareEvent($any($event))"
+        (keydown.arrowdown)="shareEvent($any($event))"
+      >
+        <cx-icon [type]="nextIcon"></cx-icon>
+      </button>
+      <button
+        *cxFeature="'!a11yCarouselPreventNavigationFocus'"
+        class="next"
+        (click)="onNextClick($event, size)"
```
Pucek9: Here as well
```suggestion
        (click)="
          $event.stopPropagation();
          activeSlide > items.length - size - 1
            ? null
            : (activeSlide = activeSlide + size)
        "
```

projects/storefrontlib/shared/components/carousel/carousel.component.ts
```diff
@@ -115,6 +117,13 @@ export class CarouselComponent implements OnInit, OnChanges {
   @Input() previousIcon = ICON_TYPE.CARET_LEFT;
   @Input() nextIcon = ICON_TYPE.CARET_RIGHT;
 
+  /**
+   * When true, prevents default mousedown behavior on navigation buttons to avoid
+   * unwanted blur events (e.g., in Safari when carousel is used inside modals or search boxes).
+   * Defaults to false to maintain backward compatibility.
+   */
+  @Input() preventNavigationFocus = false;
```
Pucek9: Instead of input, you can move inject and logic from parent component here

projects/storefrontlib/shared/components/carousel/carousel.component.html
```diff
@@ -2,23 +2,40 @@
...
-      [attr.aria-label]="'carousel.previousSlide' | cxTranslate"
-      [attr.title]="'carousel.previousSlide' | cxTranslate"
-      (keydown.arrowleft)="shareEvent($any($event))"
-      (keydown.arrowright)="shareEvent($any($event))"
-      (keydown.arrowup)="shareEvent($any($event))"
-      (keydown.arrowdown)="shareEvent($any($event))"
-    >
-      <cx-icon [type]="previousIcon"></cx-icon>
-    </button>
+    <ng-container *ngIf="size < items.length">
```
Pucek9, line 5: [MINOR]: Having a single condition instead of 2 is indeed cleaner, but IMO if we treat it as temporary changes, its worth to have as least changes as possible to clean/change, that's why I'm more fan of leaving:
```
<button *ngIf="size < items.length" *cxFeature="'a11yCarouselPreventNavigationFocus'"...></button>
<button *ngIf="size < items.length" *cxFeature="'!a11yCarouselPreventNavigationFocus'"...></button>
```
instead of introducing wrapper in this particular case (here and below)

c5346163: Angular does not allow two structural directives on the same element

Pucek9 (approved):

projects/core/src/features-config/feature-toggles/config/feature-toggles.ts
```diff
@@ -396,6 +396,16 @@ export interface FeatureTogglesInterface {
...
 
+  /**
+   * When enabled (default: true), carousel navigation buttons call preventDefault on mousedown
+   * to fix unwanted blur in Safari when the carousel is inside modals or search boxes (broken by default in Safari).
+   *
+   * Set to `false` if you rely on custom focus listeners (e.g. addEventListener('focus', ...)) on elements
+   * that contain or interact with the carousel, since preventing mousedown default can affect focus behavior.
+   * Affects: `CarouselComponent` (when preventNavigationFocus input is true, e.g. in SearchBoxComponent)
+   */
+  a11yCarouselPreventNavigationFocus?: boolean;
```
Pucek9, line 407: is it really a11y related flag?

Pucek9 (changes requested):
Plese check https://github.com/SAP/spartacus/pull/21141/changes#r2923890653

projects/core/src/features-config/feature-toggles/config/feature-toggles.ts
c5346163: Focus management and predictable keyboard navigation are core a11y concerns, so gating this Safari-specific focus fix behind an a11y* flag is reasonable.

Pucek9 (approved):

PR #21143, opened 2026-02-10
feat: CXSPA-11888: Handle non navigable conflicts

steinsebastian:


feature-libs/product-configurator/rulebased/occ/variant/converters/occ-configurator-variant-normalizer.ts
```diff
@@ -125,6 +125,7 @@ export class OccConfiguratorVariantNormalizer
       );
     }
     const uiType = this.convertAttributeType(sourceAttribute);
+    const firstConflict = sourceAttribute.conflicts?.[0] as unknown as { text?: string } | undefined;
```
Uli-Tiger: if we expect as conflicts an array of objects with property text, so `{text: string}[]` why do we define the conflict property in the occ-model as array of simple strings `string[]`.
To me it seems that there is  mismatch between occ-model and API and we should fix this mismatch.
@ChristophHi: what do you think?

Uli-Tiger (changes requested):

feature-libs/product-configurator/rulebased/occ/variant/converters/occ-configurator-variant-normalizer.ts
steinsebastian: As discussed, this will be changed with the next major release. A task has been created: https://jira.tools.sap/browse/CXSPA-12133

Uli-Tiger (approved):
lgtm

PR #21148, opened 2026-02-11
fix: missing peerDependency '@angular/forms' in 'subscription-billing' lib

StanislavSukhanov:
Checking for missing peerDependencies

--- feature-libs/subscription-billing/package.json -----------------------------
 ✖  Missing `@angular/forms` dependency that is directly referenced in library.

 i  All dependencies that are directly referenced should be specified as `dependencies` or `peerDependencies`.
    Adding new `peerDependency` might be a breaking change!
    This can be automatically fixed by running `npm run config:update`.

### DIFF
--- a/feature-libs/subscription-billing/package.json
+++ b/feature-libs/subscription-billing/package.json
@@ -29,6 +29,7 @@
     "@angular-devkit/schematics": "^21.1.0",
     "@angular/common": "^21.1.0",
     "@angular/core": "^21.1.0",
+    "@angular/forms": "^21.1.0",
     "@angular/router": "^21.1.0",
     "@ng-select/ng-select": "^21.1.4",
     "@spartacus/cart": "221121.7.0",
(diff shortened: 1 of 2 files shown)


feature-libs/subscription-billing/package.json
```diff
@@ -29,6 +29,7 @@
     "@angular-devkit/schematics": "^21.1.0",
     "@angular/common": "^21.1.0",
     "@angular/core": "^21.1.0",
+    "@angular/forms": "^21.1.0",
```
Platonn, line 32: Please run also `npm run generate:deps` locally and commit changes from `dependencies.json`. Otherwise our Installation Schematics (which read `dependencies.json`) might be misconfigured. (and the PR pipeline won't pass anyway)

StanislavSukhanov: done

Platonn (approved):

PR #21149, opened 2026-02-11
fix: Fix tabIndex typo in pagination to restore keyboard a11y

Zeyber:
Closes: https://jira.tools.sap/browse/CXSPA-11695


uroslates (approved):
Hi @Zeyber I spotted that there are few more files having the same [tabIndex] issue. Maybe we could also address those in the cox of this one as well?

Zeyber:
Hey @uroslates, Which files are you referring to for example? I only see `tabIndex` in some specs and the keyboard focus utility. But not in any templates.

Issue #12351, opened 2021-05-07
Call SiteContextRoutesHandler.init method in context initializers


Platonn:
It's a follow up for https://github.com/SAP/spartacus/pull/12340. Related to https://github.com/SAP/spartacus/issues/8368

Currently `SiteContextRoutesHandler` remains in the separate `APP_INITIALIZER` and we _assume_ it's provided and executed earlier than `LanguageInitializer` (or any other context initializers).

As an improvement, we can call the method `SiteContextRoutesHandler.init()` from `LanguageInitializer` (and all other specific context initializers), so we no longer depend on the order of `APP_INITIALIZER`s. But as prerequisite, we need to make the method `SiteContextRoutesHandler.init(contextId: string)` parametrizable with a specific site context ID.

### Consequences for customers using custom site context
Some customers are using custom site contexts and storing them in the URL. See [docs for custom site contexts](https://sap.github.io/spartacus-docs/automatic-context-configuration/#adding-a-custom-context) and the use case https://github.com/SAP/spartacus/issues/11549.

Currently the `SiteContextRoutesHandler.init()` is called once and handles all defined site contexts (even custom ones). But after this refactor we'll invoke it with 3 times, only with explicit parameters: for lang, curr and base site, so custom context won't be synchronized with URL anymore. So it will be a breaking change for customers - they'll need to implement their own `CustomSiteContextInitializer` (similar to LanguageInitializer) and invoke there the routes handler with the param being the custom context ID.

PR #21151, opened 2026-02-12
feat: add site context route handler to initialization to currency, language and basesite initialization

npapp-dev002:
#closes [CXSPA-12011](https://jira.tools.sap/browse/CXSPA-12011) 

**Summary of changes:** 
- TODO (#12351) comments were replaced by site-context-initialization. `initOnce()` method is used without a parameter because it is possible to handle URL values globally.
- `site-context-routes-handler` initialization was remade to initialize once and reuse the initialized value. Language, country, currency initializaters uses the same value. There is no race condition anymore.
- spec files were updated.

**QA:**
- Language should be read from the url (highest priority) and should stay the same through the customer's journey
- Check the example implementation about adding a [custom context](https://github.com/SAP/spartacus/pull/21152)


projects/core/src/site-context/providers/context-initializer-providers.ts
```diff
@@ -12,27 +12,25 @@ import { LanguageInitializer } from '../services/language-initializer';
 export function initializeCurrency(
   currencyInitializer: CurrencyInitializer
 ): () => void {
-  const result = () => {
+  return () => {
```
uroslates, line 15: Does it make sense to replace deprecated APP_INITIALIZER (with recommended provideAppInitializer) in the context of this file as well?

projects/core/src/site-context/services/base-site-initializer.ts
```diff
@@ -24,15 +27,17 @@ export class BaseSiteInitializer implements OnDestroy {
...
-  initialize(): void {
-    this.subscription = this.configInit
-      .getStable('context')
-      .pipe(
-        // TODO(#12351): <--- plug here explicitly SiteContextRoutesHandler
-        switchMap(() => this.setFallbackValue())
-      )
-      .subscribe();
+  initialize(): Observable<unknown> {
+    const init$ = this.configInit.getStable('context').pipe(
```
uroslates: Good work! 👍 

The only recommendation for all these initialisers (base-site/currency/language) from my side (but I leave it to you to decide if it makes sense to do it here or maybe better to just create an improvement ticket for that) is to somehow get rid of explicit/manual subscription management. Having said that I can think of 2 possible options:
1. (if we still continue using APP_INITIALIZER in context-initialiser-providers.ts) - using `firstValueFrom()` something like below might work (and eliminate subscription and `ngOnDestroy`):
```
initialize(): Promise<unknown> {
  return firstValueFrom(
    this.configInit.getStable('context').pipe(
      switchMap(() => this.siteContextRoutesHandler.initOnce()),
      switchMap(() => this.setFallbackValue())
    )
  );
}
```
 2. Or if decided to replace APP_INITIALIZER (as mentioned in previous comment for `context-initialiser-providers.ts`) with `provideAppInitializer` moving the logic to provider itself might help getting rid of those (by avoiding class method pattern). Maybe something similar to:
```
// In context-initializer-providers.ts
provideAppInitializer(() => {
  const configInit = inject(ConfigInitializerService);
  const routesHandler = inject(SiteContextRoutesHandler);
  const baseSiteService = inject(BaseSiteService);
  
  return firstValueFrom(
    configInit.getStable('context').pipe(
      switchMap(() => routesHandler.initOnce()),
      tap((config) => /* set fallback value */)
    )
  );
})
```

uroslates (dismissed):
👍

projects/core/src/site-context/services/base-site-initializer.ts
npapp-dev002: Thank you very much for the recommendations. I believe the first one can be implemented. However, the scope of the second option might be outside the current ticket. 
I created a ticket specifically for replacing `APP_INITIALIZER` with `provideAppInitializer`: [CXSPA-12150](https://jira.tools.sap/browse/CXSPA-12150)

uroslates (approved):

projects/core/src/site-context/services/base-site-initializer.ts
uroslates: _(for tracking purposes)_ Created follow-up improvement ticket for APP_INITIALIZER replacement here [CXSPA-12192](https://jira.tools.sap/browse/CXSPA-12192).

PR #21154, opened 2026-02-12
fix: remove redundant heading for orders in order history

npapp-dev002:
closes #CXSPA-12087

The feature flag of `removeDuplicatedOrderHistoryHeader` was added.

Before changes:
image
After changes: 
image


feature-libs/order/components/order-history/order-history.component.html
```diff
@@ -17,9 +17,6 @@
         <h4 *ngIf="type.replenishmentOrder">
           {{ 'orderHistory.replenishmentHistory' | cxTranslate }}
         </h4>
-        <h2 *ngIf="!type.replenishmentOrder">
```
Zeyber, line 20: Thanks! Just fyi we'll need to put this behind a feature flag since its a DOM change (its why we have so many a11y feature flags ;) ).

Zeyber (changes requested):

feature-libs/order/components/order-history/order-history.component.html
npapp-dev002: Thank you very much for pointing this out. I added `removeDuplicatedOrderHistoryHeader` using the (https://wiki.one.int.sap/wiki/spaces/spar/pages/3564966608/Coding+Guidelines#CodingGuidelines-UseFeatureToggleswhenchangingbehaviorofanexistingfeature) documentation. Please check.

Zeyber (dismissed):

Zeyber (approved):

PR #21158, opened 2026-02-16
feat: (CXSPA-7459) - Organization table notification readout fix

Pio-Bar:
Closes: [CXSPA-7459](https://jira.tools.sap/browse/CXSPA-7459)

The fix prevents the table component from re-rendering its rows each time cell data is updated. 
This wold cause the screen reader to quickly jump between elements and skip some messages.

Please see the ticket to recreate the issue for QA, the solution need to work for both Voice Over and JAWS.

### DIFF
--- a/feature-libs/organization/administration/components/shared/message/notification/notification-message.component.html
+++ b/feature-libs/organization/administration/components/shared/message/notification/notification-message.component.html
@@ -1,6 +1,10 @@
 <div
   class="inner"
-  [cxFocus]="featureConfigService.isEnabled('a11yCardNotificationMessage') ? {} : { autofocus: true, focusOnEscape: true }"
+  [cxFocus]="
+    isA11yCardNotificationMessageEnabled
+      ? {}
+      : { autofocus: true, focusOnEscape: true }
+  "
   (esc)="close()"
 >
   <cx-icon *ngIf="messageIcon" [type]="messageIcon"></cx-icon>
--- a/feature-libs/organization/administration/components/shared/message/notification/notification-message.component.html
+++ b/feature-libs/organization/administration/components/shared/message/notification/notification-message.component.html
@@ -1,6 +1,6 @@
 <div
   class="inner"
-  [cxFocus]="{ autofocus: true, focusOnEscape: true }"
+  [cxFocus]="featureConfigService.isEnabled('a11yCardNotificationMessage') ? {} : { autofocus: true, focusOnEscape: true }"
   (esc)="close()"
 >
   <cx-icon *ngIf="messageIcon" [type]="messageIcon"></cx-icon>
--- a/feature-libs/organization/administration/components/shared/message/notification/notification-message.component.html
+++ b/feature-libs/organization/administration/components/shared/message/notification/notification-message.component.html
@@ -1,7 +1,7 @@
 <div
   class="inner"
   [cxFocus]="
-    isA11yCardNotificationMessageEnabled
+    isA11yCardNotificationMessageFeatureEnabled
       ? {}
       : { autofocus: true, focusOnEscape: true }
   "
--- a/feature-libs/organization/administration/components/shared/table/cell.component.ts
+++ b/feature-libs/organization/administration/components/shared/table/cell.component.ts
@@ -33,16 +33,15 @@ export class CellComponent implements OnInit, OnDestroy {
   changeDetectorRef = inject(ChangeDetectorRef);
   private featureConfigService = inject(FeatureConfigService);
 
-  constructor(
-    protected outlet: OutletContextData<TableDataOutletContext>
-  ) {}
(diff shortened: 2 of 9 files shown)


feature-libs/organization/administration/components/shared/message/notification/notification-message.component.html
```diff
@@ -1,6 +1,10 @@
 <div
   class="inner"
-  [cxFocus]="{ autofocus: true, focusOnEscape: true }"
+  [cxFocus]="
+    isA11yCardNotificationMessageEnabled
```
uroslates: The name of getter seems a bit counterintuitive - `isA11yCardNotificationMessageEnabled` suggests that enabling it should activate accessibility features, but currently, enabling it disables the focus behaviour (`autofocus` and `focusOnEscape`)

feature-libs/organization/administration/components/shared/table/cell.component.ts
```diff
@@ -5,24 +5,49 @@
...
+    if (this.featureConfigService.isEnabled('a11yCardNotificationMessage')) {
+      this.contextSubscription = this.outlet.context$.subscribe((context) => {
+        this.outlet.context = context;
+        this.changeDetectorRef.markForCheck();
+      });
+    }
+  }
+
+  ngOnDestroy(): void {
+    this.contextSubscription?.unsubscribe();
```
uroslates: We might have a potential uninitialised Subscription bug since `contextSubscription` (could be `undefined` in `ngOnDestroy`) when feature flag is disabled.

uroslates: One possibility here could be using `takeUntilDestroyed()`  (which auto-unsubscribes) would eliminate the need for having `ngOnDestroy` method.


```
private destroyRef = inject(DestroyRef);

ngOnInit(): void {
  if (this.featureConfigService.isEnabled('a11yCardNotificationMessage')) {
    this.outlet.context$.pipe(
      takeUntilDestroyed(this.destroyRef)
    ).subscribe((context) => {
      this.outlet.context = context;
      this.changeDetectorRef.markForCheck();
    });
  }
}
```

projects/core/src/features-config/feature-toggles/config/feature-toggles.ts
```diff
@@ -379,6 +379,12 @@ export interface FeatureTogglesInterface {
    * on Customer Ticketing dialog.
    */
   a11yCustomerTicketingVisualFocusFix?: boolean;
+
+  /**
+   * Improves the screen reader experience of the notification message component.
```
uroslates: Would recommend adding here more context for the consumers to understand, something like:
> When enabled, the `NotificationMessageComponent` will use .... to ensure that screen readers announce the messages...

uroslates (changes requested):

uroslates:
@Pio-Bar Tested locally using JAWS and the message when unassigning Approver is being vocalised! Example "User Carla Torres unassigned successfully." is vocalised! Was that all that was needed checking or should I test something more?

feature-libs/organization/administration/components/shared/message/notification/notification-message.component.html
Pio-Bar: I've improved the name slightly. However, enabling it does improve the a11y disabling the focus directive is an improvement and stops the abrupt focus shift.

feature-libs/organization/administration/components/shared/table/cell.component.ts
Pio-Bar: I like your suggestion better, thanks!

projects/core/src/features-config/feature-toggles/config/feature-toggles.ts
Pio-Bar: Message improved

Pio-Bar:
> @Pio-Bar Tested locally using JAWS and the message when unassigning Approver is being vocalised! Example "User Carla Torres unassigned successfully." is vocalised! Was that all that was needed checking or should I test something more?

The original issue was present on VoiceOver but the fix must work well on JAWS as well. Please make sure it's fine on both. The fix also prevents the assign button from losing focus. It's possible to test this in the `MANAGE` tab above the table as it gives you persistent buttons to work with. Thanks for testing this!

uroslates:
> The original issue was present on VoiceOver but the fix must work well on JAWS as well. Please make sure it's fine on both. The fix also prevents the assign button from losing focus. It's possible to test this in the MANAGE tab above the table as it gives you persistent buttons to work with. Thanks for testing this!

Tired locally with the latest code updates (on JAWS) and seems to be working fine. Assign/unassign don't lose focus and message conveyed is read out.

uroslates (approved):
👍

PR #21183, opened 2026-02-23
fix: Fix background style not showing in high contrast themes for tab accordian component

Zeyber:
Closes: https://jira.tools.sap/browse/CXSPA-12191


uroslates (approved):
👍 Works fine!

Without Fix:
Image

With Fix:
Image

PR #21184, opened 2026-02-23
fix: notification added to announce when item is successfully removed from cart

npapp-dev002:
Changes: 

- Notification added for successful card removal
- Feature toggle added to preserve the old behavior

QA steps:

- Select a product
- Click on **Add to Cart**
- Click on Remove
- A notification about successful removal should appear

image


projects/storefrontlib/cms-components/misc/global-message/global-message.component.html
```diff
@@ -1,10 +1,20 @@
 <div *ngIf="messages$ | async as messages">
-  <div
-    class="cx-visually-hidden"
-    *ngFor="let assistiveMsg of messages[messageType.MSG_TYPE_ASSISTIVE]"
-  >
-    <span>{{ assistiveMsg | cxTranslate }}</span>
-  </div>
+  <ng-container *cxFeature="'globalMessageCartRemovalNotification'">
+    <div class="cx-visually-hidden" aria-live="assertive" aria-atomic="true">
```
Zeyber: You don't need to hide this behind a feature flag because we don't treat aria attribute changes as breaking changes ;)

feature-libs/cart/base/components/cart-shared/cart-item/cart-item.component.ts
```diff
@@ -58,6 +69,8 @@ export class CartItemComponent implements OnChanges {
 
   @Input() promotionLocation: PromotionLocation = PromotionLocation.ActiveCart;
 
+  protected globalMessageService = inject(GlobalMessageService);
```
Zeyber: We should inject this as an optional service. So that the call in remove item can also be optional (ie. `this.globalMessageService?.add(...`. This is because we are introducing the global message service as a new service that might not necessarily be injected already. This is how we can handle the breaking change.

Zeyber (changes requested):

projects/storefrontlib/cms-components/misc/global-message/global-message.component.html
```diff
@@ -1,10 +1,12 @@
 <div *ngIf="messages$ | async as messages">
-  <div
-    class="cx-visually-hidden"
-    *ngFor="let assistiveMsg of messages[messageType.MSG_TYPE_ASSISTIVE]"
-  >
-    <span>{{ assistiveMsg | cxTranslate }}</span>
-  </div>
+  <ng-container>
```
Zeyber: Do we need this ng-container?

feature-libs/cart/base/components/cart-shared/cart-item/cart-item.component.ts
```diff
@@ -103,5 +116,9 @@ export class CartItemComponent implements OnChanges {
   removeItem() {
     this.quantityControl.setValue(0);
     this.quantityControl.markAsDirty();
+    this.globalMessageService?.add(
```
Zeyber: I'm not sure this is the correct place to announce the item has been removed. Because at this point, only the request to remove the item is made.

Instead, we should announce the removed item based on the successful response from the server. You can check out how the global message interceptors work for hints on this. Please let me know if you need any help with this.

Zeyber (changes requested):

projects/storefrontlib/cms-components/misc/global-message/global-message.component.html
npapp-dev002: Nope. It's just a remnant of the previous solution.

feature-libs/cart/base/components/cart-shared/cart-item/cart-item.component.ts
npapp-dev002: You are right.

feature-libs/cart/base/core/store/effects/cart-entry.effect.ts
```diff
@@ -85,6 +88,10 @@ export class CartEntryEffects {
           .remove(payload.userId, payload.cartId, payload.entryNumber)
           .pipe(
             map(() => {
+              this.globalMessageService.add(
```
Zeyber: This method is better. Just remember to inject the service optionally otherwise the method could fail since this is a breaking change.

Zeyber (changes requested):

projects/storefrontlib/cms-components/misc/global-message/global-message.component.html
```diff
@@ -1,9 +1,9 @@
 <div *ngIf="messages$ | async as messages">
-  <div
-    class="cx-visually-hidden"
-    *ngFor="let assistiveMsg of messages[messageType.MSG_TYPE_ASSISTIVE]"
-  >
-    <span>{{ assistiveMsg | cxTranslate }}</span>
+  <div class="cx-visually-hidden" aria-live="assertive" aria-atomic="true">
+    <span
+      *ngFor="let assistiveMsg of messages[messageType.MSG_TYPE_ASSISTIVE]"
```
Zeyber: Is there any reason *ngFor has moved from the `div` to the `span` tag?

Zeyber (changes requested):

projects/storefrontlib/cms-components/misc/global-message/global-message.component.html
npapp-dev002: Nope. Thanks for noticing. 👍

Zeyber (approved):

PR #21186, opened 2026-02-23
chore: add initial AGENTS.md and few SKILL.md

Platonn:
fixes https://jira.tools.sap/browse/CXSPA-12195


.claude/CLAUDE.md
```diff
@@ -0,0 +1 @@
+@AGENTS.md
```
Platonn, line 1: As of today, Claude Code doesn't pick AGENTS.md. This is a documented workaround on Claude docs:
https://code.claude.com/docs/en/claude-code-on-the-web#best-practices

AGENTS.md
```diff
@@ -0,0 +1,82 @@
...
+
+| Path | What belongs there |
+|------|--------------------|
+| `feature-libs/` | Optional features for standard SAP Commerce backend (cart, checkout, order, etc.) |
+| `integration-libs/` | Features requiring special backend addons (cdc, cds, digital-payments, opf) |
+| `core-libs/setup/` | Core setup utilities, SSR engine |
+| `projects/core/` | Core non-UI library |
+| `projects/storefrontlib/` | UI-related core, CMS engine |
+| `projects/schematics/` | Installation and migration schematics |
+| `projects/storefrontapp/` | Demo application |
```
pawelfras: I would also add information to the file about where the custom eslint rules and custom builders are located, so it's easier for agent to propose new rule or go through our custom build process for easier debugging

AGENTS.md
```diff
@@ -0,0 +1,82 @@
...
+
+Angular meta-framework for SAP Commerce Cloud e-commerce storefronts. Monorepo using npm workspaces + Nx.
+
+## Critical Rules
+
+- **NEVER** add/change `peerDependencies` in library `package.json` files — causes breaking changes for customers. Add `devDependencies` to root `package.json` only if truly needed for development.
+- **NEVER** introduce breaking changes without feature toggles (since v2211.19, no SemVer). If you must change behavior, styling, or public API, add a feature toggle.
+- **ALWAYS** export public APIs through `public_api.ts` files via barrel files. Never export APIs explicitly marked as internal.
+
+## Directory Quick Reference
```
pawelfras: also, maybe it's worth mentioning about where e2es are located, if we mention the command about how to run e2es.

pawelfras: another thing that came to my mind is to introduce separate info about GH actions and scripts and also tools - can be added in the future

AGENTS.md
```diff
@@ -0,0 +1,82 @@
...
+nx test <library-name>         # Single library (e.g., nx test storefrontlib)
+npm run test:all-schematics    # All schematics tests
+
+# Lint & Format
+npm run lint
+npm run prettier:fix
+
+# E2E
+npm run e2e:run
+npm run e2e:run:b2b
```
pawelfras, line 57: Also, regarding e2e, might be nice to introduce later an example of fine-grained command that runs only affected e2es (to avoid situation where agent runs all tests, which might be redundant and heavy). But it might be a different story, and should be leveraged first to Nx to use its affected functionality 🤔

AGENTS.md
Platonn: done

Platonn: done

AGENTS.md
Platonn: done

AGENTS.md
Platonn: good point. i found AI agents (and myelf too 😅 ) often struggle to remember how to run tests for a specific file - in Jest vs Jasmine

done

AGENTS.md
```diff
@@ -0,0 +1,105 @@
...
+| `projects/core/` | Core non-UI library |
+| `projects/storefrontlib/` | UI-related core, CMS engine |
+| `projects/schematics/` | Installation and migration schematics |
+| `projects/storefrontapp/` | Demo application |
+| `projects/storefrontapp-e2e-cypress/` | E2E Browser tests (Cypress) |
+| `projects/ssr-tests/` | E2E SSR tests (Node) |
+| `tools/eslint-rules/` | Custom ESLint rules |
+| `tools/build-lib/` | Custom Angular builders for library builds |
+| `ci-scripts/` | Scripts used by CI to check peer deps, feature toggles, unit tests, E2E |
+| `.github/workflows/` | GitHub Actions CI/CD pipelines |
```
npapp-dev002: Shouldn’t we add the path to where our styles are located?
| `projects/storefrontstyles/scss/` | Styles |

AGENTS.md
```diff
@@ -0,0 +1,105 @@
...
+npm run lint
+npm run prettier:fix
+
+# E2E
+npm run e2e:run
+npm run e2e:run:b2b
+
+# SSR E2E
+npm run build:ssr:local-http-backend # special build prerequisite before SSR E2E
+npm run test:sst
```
npapp-dev002: Shouldn't be `npm run test:ssr`?

AGENTS.md
```diff
@@ -0,0 +1,105 @@
...
+
+# Jasmine Test Angular libs
+npm run test:libs              # All library tests
+nx run <library-name>:test         # Single library (e.g., nx run storefrontlib:test)
+nx run <library-name>:test --include="**/<spec filename>" # Specific test file
+
+# Jest Test Schematics libs
+npm run test:all-schematics    # All schematics tests
+nx run <library-name>:test-jest # Single library
+nx run <library-name>:test-jest --testPathPatterns="<spec filename>" # Specific file
```
ercultimate: Should we use kebab-case for `<spec filename>`?

pawelfras (commented):
After second thoughts and discussion offline, let's revisit what is really unique/opinionated in our project, that whould worth to let agent know about and also, what could be extracted to separate skills, so that we do not clutter the context with redundant information. 
Food for thought (from Addy Osmani's [post](https://www.linkedin.com/feed/update/urn:li:activity:7431963663441285120/?originTrackingId=iyv5FKX%2Fw1PAuaBSOF%2FA7Q%3D%3D))

> Treat your [AGENTS.md](http://agents.md/) as a living list of codebase smells you haven't fixed yet. If an agent keeps reaching for the wrong dependency or putting utilities in the wrong folder, don't just add a prose instruction to the context file. Restructure the code, add a linter rule, or improve the test coverage. Treat it as a diagnostic tool, fix the underlying friction, and then delete the line.

AGENTS.md
Platonn: thanks. fixed

AGENTS.md
Platonn: thanks. fixed

AGENTS.md
```diff
@@ -0,0 +1,67 @@
+# Spartacus
+
+Angular meta-framework for SAP Commerce Cloud e-commerce storefronts. Monorepo using npm workspaces + Nx.
+
+## Critical Rules
+
+- **NEVER** add/change `peerDependencies` in library `package.json` files — causes breaking changes for customers. Add `devDependencies` to root `package.json` only if truly needed for development.
+- **NEVER** introduce breaking changes without feature toggles (since v2211.19, no SemVer). If you must change behavior, styling, or public API, add a feature toggle.
+- **ALWAYS** export public APIs through `public_api.ts` files via barrel files. Never export APIs explicitly marked as internal.
```
rmch91: Maybe we can add also information to not use private access modifiers?

rmch91 (dismissed):

pawelfras (approved):
LGTM. Thank you for preparing these docs. I propose to merge them in that state, see how it works with our individual workflows and adjust, if needed 👍

PR #21192, opened 2026-02-24
fix: skip first active language emission

FollowTheFlo:
symptoms: duplicate occ call when starting the app
solution: skip the first language getActive() emission. It is still uncleared why it this initial call occurs.
we see this pattern in other services
weg:
spartacus/integration-libs/opf/checkout/components/opf-checkout-payment-wrapper/opf-checkout-payment-wrapper.component.ts
This fix has been tested on e2e and manually. No side effects to report.


uroslates (approved):
👍 
Looks good. Thanks for working on this one @FollowTheFlo!
I have added just one more test.

Tested locally and seems to be working fine.
Maybe we should also have @KateChuen QA it (try it out) prior to merging!?

PR #21197, opened 2026-02-26
fix: enhance accessibility of Added to Cart dialog with semantic heading

npapp-dev002:


uroslates (approved):
Code changes looks good and works fine for Add to Card dialog.
But the issue mentiones subscriptions. So i'm not sure if it has to be applied there as well. Best to dbl check with person requesting it.

Zeyber (approved):
Thanks for the PR @npapp-dev002 and @uroslates for your review. For me, the fix solves the issue as the reporters specifically mention the add-to-cart dialogue text as being the issue. 

They comment that the issue is across the application. But this comment is too generalized in my opinion to be able to address it. So I think this fix is the best we can do.

PR #21202, opened 2026-02-27
fix: Cart not refreshed after re-order

rmch91:
Closes: https://jira.tools.sap/browse/CXSPA-11929

QA steps:

Can be reproduced locally with SPA from develop-next-major and backend 22.11jdk21. ("CX_BASE_URL": "https://spartacus-devci7677.eastus.cloudapp.azure.com:8443/")

Submit an order in B2B site. View order history and navigate to order details. Re-order

Message 'Products have been sucessfully added to the cart' appears.

Navigate to cart -> no cart entries visible. 

Only when you hit F5 or add another product to the cart, cart becomes updated with the entry from re-order.


feature-libs/order/core/facade/reorder-order.service.ts
```diff
@@ -59,16 +59,23 @@ export class ReorderOrderService implements ReorderOrderFacade {
...
+          // before proceeding with the reorder.
+          return this.multiCartFacade.getCartEntity(cartId).pipe(
+            filter((state) => !state.value && !state.loading),
+            take(1),
+            map(() => userId)
+          );
         }
 
-        return userId;
+        return of(userId);
```
uroslates, line 71:
```
// ...
    switchMap(([userId, cartId]) => {
      if (!userId) {
        return throwError(() => new Error('Must be logged in to reorder'));
      }

      if (cartId) {
        return this.deleteCartAndWait(cartId, userId);
      }

      return of(userId);
    })
  );
// ...

protected deleteCartAndWait(cartId: string, userId: string): Observable<string> {
 this.multiCartFacade.deleteCart(cartId, userId);
  
  return this.multiCartFacade.getCartEntity(cartId).pipe(
    filter((state) => !state.value && !state.loading || !!state.error),
    take(1),
    switchMap((state) => {
      if (state.error) {
        // Log W/O blocking reorder
        console.warn('Cart deletion may have failed:', state.error);
      }
      return of(userId);
    })
}
```

1. [neat pick]  Would add proper RxJs error propagation (`throwError` instead of `throw` for consistency with reactive nature & better integration with `catchError` and `retry`/... RxJs primitives)
2. [neat pick]  For the sake of better readability and testability I would suggest extracting `deleteCartAndWait` into separate `protected` method.
3. [neat pick]  handle failed cases (`!state.value && !state.loading || !!state.error`) - deletion failed but we continue
4. [neat pick] `this.multiCartFacade.getCartEntity` -> `switchMap` block - we log error as warning and continue (not sure if necessary - leave it to you to decide)

I am not sure if in addition to above - in the context of returned `deleteCartAndWait` method observable - we should also handle adding rxjs `timeout` (for preventing indefinite hanging) and `catchError` (just logging and proceeding event if it times out). I guess it would be too much.

uroslates (dismissed):
Looks good! 👍 Great job!

Added a comment with some neat-peek suggestions but not required, leave it to you to decide if they make sense in this case.

Would suggest we assign it to @KateChuen for QA!

uroslates (approved):

PR #21206, opened 2026-03-03
feat: add b2b cost center search

i333055:


projects/core/src/features-config/feature-toggles/config/feature-toggles.ts
```diff
@@ -399,6 +399,12 @@ export interface FeatureTogglesInterface {
   /**
    * When enabled, allows searching B2B customers by name in the organization.
    */
+  enableB2BAdminCustomerSearch?: boolean;
```
Melody-zhou-512: This one has been modified in parent branch. please update branch.

i333055: fixed

PR #21207, opened 2026-03-03
fix: filter inactive currencies from the currency service

npapp-dev002:
closes [CXSPA-11636](https://jira.tools.sap/browse/CXSPA-11636)

QA:
- Go to backoffice 
image
- Select japanese currency and set its active field to false
image
- It takes time for the changes to be reflected in the output (5-15 min or sometimes even more)

- Start the app with `npm start`

- There are currently two currencies supported (USD, JPY). If only one is available, the currency field cannot be seen.
image

- Check the response, JPY should have its active field set to false.
image

**Note**: Be aware that setting JPY to false might interfere with running e2e tests.


uroslates (approved):
👍 Gr8 job!
Tested locally and works as described!

Image

PR #21208, opened 2026-03-03
feat: (CXSPA-12100) - PasswordVisibilityToggle doubled screen reader output

Pio-Bar:
Closes: [CXSPA-12100](https://jira.tools.sap/browse/CXSPA-12100)

Resolves the issue of the screen reader repeating the message twice while interacting with the button. This is achieved by providing a static `title` and `aria-label`. It's best practice not to modify those and provide the dynamic information with `aria-pressed` instead.

The fix needs to work well on both VoiceOver and JAWS. The now static aria-label/title needs to be read out as well as it's status (pressed/not pressed). The output should not be repeated more than once.


uroslates (dismissed):
Checked and seems to be working fine. Here is attached screenshot of current behaviour (with the changes):

Image


The only minor remarks I have (thinking out loud) are:
1. not sure if `title` attributes are required here (not relevant for screen reader but yes for visual tooltip). **Maybe best to leave it as is** unless explictily requested to be remove by someone.
2. `[attr.aria-controls]="inputId"` is recommended (but not required) - to link button to the input it controls. But same as above - **Maybe best to leave it as is** unless explictily requested to be remove by someone.

uroslates (dismissed):

uroslates (approved):

PR #21209, opened 2026-03-03
fix: fix a11y issue with reading selected address

uroslates:
Closes: https://jira.tools.sap/browse/CXSPA-12107.

QA: Start the app and use JAWS to verify that upon adding any product to the cart and Proceeding to the checkout any selected address is read out completely (previously only 'Selected Address' was read out by the screen reader).


feature-libs/checkout/base/components/checkout-delivery-address/checkout-delivery-address.component.ts
```diff
@@ -376,10 +376,8 @@ export class CheckoutDeliveryAddressComponent implements OnInit {
     return !!this.checkoutConfigService?.shouldUseAddressSavedInCart();
   }
 
-  protected getCardRole(isCardSelected: boolean): 'button' | 'application' {
-    const role: 'button' | 'application' = !isCardSelected
-      ? 'button'
-      : 'application';
+  protected getCardRole(isCardSelected: boolean): 'group' | 'region' {
+    const role: 'group' | 'region' = !isCardSelected ? 'group' : 'region';
     return role;
```
StanislavSukhanov: Let's remove redundand `const` 
e.g: `return isCardSelected? 'region' : 'group'`

StanislavSukhanov (changes requested):

feature-libs/checkout/base/components/checkout-delivery-address/checkout-delivery-address.component.ts
uroslates: 👍 Applied!

StanislavSukhanov (dismissed):
legit 👍

StanislavSukhanov (approved):

PR #21211, opened 2026-03-03
feat: (CXSPA-12108) - Ng select value readout on collapsed dropdown

Pio-Bar:
Closes: [CXSPA-12108](https://jira.tools.sap/browse/CXSPA-12108)

After the fix the currently selected option should be announced alongside the label.
This change also allows us to clean up a previous workaround we used for mobiles.

The solution has to work for both VoiceOver and JAWS.


uroslates (dismissed):
👍 Checked using JAWS - selection is read out correctly. Focus is set to facets upon setting but navigating back to the selectbox reads out correct option! Attached screenshot:

Image

uroslates (dismissed):

uroslates (approved):

PR #21220, opened 2026-03-05
fix: (CXSPA-12100) - Password visibility e2e

Pio-Bar:
Updates selectors to fix failing `/regression/password-visibility/password-visibility.e2e.cy.ts`


uroslates (approved):
Tested locally on develop and test does fail.

The changes fix the issue! 👍

PR #21224, opened 2026-03-09
fix: (CXSPA-11223) Maintain site context changes on login page

BenjaminEmiliani:


projects/core/src/auth/user-auth/services/auth-redirect.service.ts
```diff
@@ -67,7 +68,20 @@ export class AuthRedirectService implements OnDestroy {
...
+          const redirectPageContextUrl =
+            this.siteContextUrlSerializer.urlExtractContextParameters(
+              redirectUrl
+            ).url;
+          this.routing.goByUrl(
+            this.siteContextUrlSerializer.combineUrlAndSiteContextUrlParams(
+              redirectPageContextUrl,
+              actualParams
+            )
+          );
```
Platonn: [MINOR]
Currently we're replacing the possibly-outdated site-context params URL prefix explicitly with the up-to-date site-context params URL prefix at the time "just before the redirect". And by the way we need to expose a new public method `SiteContextUrlSerializer.combineUrlAndSiteContextUrlParams`

What do you think about alternative approach: to avoid saving the site-context params URL prefix in the first place, during the time of "saving the redirect url"? 
As a consequence, the up-to-date values will be applied implicitly during the time of the redirect. For example like in the code snippet below, changing just the `setRedirectUrl` method, and using the already public method `siteContextUrlSerializer.urlExtractContextParameters`:

```ts
  setRedirectUrl(url: string): void {
    if (!this.authFlowRoutesService.isAuthFlow(url)) {
      // Remove site-context params from the redirect URL to avoid using outdated values (e.g., language) after authentication.
      // Without these params in the Redirect URL, up-to-date site-context values will be applied implicitly on redirect.
      const urlWithoutSiteContextParams =
        this.getUrlWithoutSiteContextParams(url);
      this.authRedirectStorageService.setRedirectUrl(
        urlWithoutSiteContextParams
      );
    }
  }

  /**
   * For a given URL, removes the prefix with site-context params (e.g., baseSite, language, currency)
   *
   * @example
   * Supposing the configured site-context URL params are [baseSite, language, currency],
   * and given the url '/electronics-spa/en/USD/product/123',
   * it returns '/product/123'.
   */
  protected getUrlWithoutSiteContextParams(url: string): string {
    const { url: urlWithoutSiteContextParams } =
      this.siteContextUrlSerializer.urlExtractContextParameters(url);
    return urlWithoutSiteContextParams;
  }
```

projects/storefrontapp-e2e-cypress/cypress/e2e/regression/site-context/site-context.e2e.cy.ts
```diff
@@ -0,0 +1,69 @@
...
+ */
+import { login } from '../../../helpers/auth-forms';
+import * as siteContextSelector from '../../../helpers/site-context-selector';
+import * as merchandisingCarousel from '../../../helpers/vendor/cds/merchandising-carousel';
+import { myCompanyAdminUser } from '../../../sample-data/shared-users';
+import { whenJDK17, whenJDK21 } from '../../../support/utils/jdk-versions';
+import { switchSiteContext } from '../../../support/utils/switch-site-context';
+import { isolateTestsBefore } from '../../../support/utils/test-isolation';
+
+context('Site Context on redirect', { testIsolation: false }, () => {
```
Platonn, line 12: I really appreciate you've covered the bugfix with new e2e test to prevent future regressions 🙌

projects/core/src/auth/user-auth/services/auth-redirect.service.ts
BenjaminEmiliani: Firstly, thanks for feedback, I will look into the alternative approach. Second, I have a question, in the first paragraph, you mention "And by the way we need to expose a new public method SiteContextUrlSerializer.combineUrlAndSiteContextUrlParams". In your alternative solution this new method is no longer used, so are you saying that we need to expose it even if not used in bug fix implementation?

Platonn (approved):

PR #21229, opened 2026-03-10
feat: (CXSPA-12121) - Fix screen reader output of cart's item list table

Pio-Bar:
Closes: https://jira.tools.sap/browse/CXSPA-12121

Improves the readout of cart's item list table by ensuring proper semantic HTML structure.

JAWS is required for testing, use the table controls to make sure cells are recognized as cells and the count of rows and columns is accurate.


uroslates (commented):
- Tested on win / JAWS and table as well as selected row/column is vocalised (as well as quantity input value)! Check attached JAWS `View History` screenshot:

Image

feature-libs/cart/base/components/cart-shared/cart-item-list-row/cart-item-list-row.component.html
uroslates, line 120: `nitpick` I believe role is not required for td's here (unless display prop is explicitly changed). But lets leave it for now.

uroslates (approved):
👍

PR #21231, opened 2026-03-10
chore: make AGENTS.md more concise, remove rarely used skill (create-library) and gitignore explicitly '.claude/settings.local.json'

Platonn:
followup of https://github.com/SAP/spartacus/pull/21186/

related to https://jira.tools.sap/browse/CXSPA-12195

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -27,7 +27,7 @@ Monorepo: `npm workspaces` and `Nx`
 See `ci-scripts/unit-tests.sh`
 
 ### Specific tests
-```
+```bash
 ## Add `--no-watch --source-map --code-coverage --browsers ChromeHeadless`
 nx run <library-name>:test         # Single library (e.g., nx run storefrontlib:test)
 nx run <library-name>:test --include="**/<spec-filename>" # Specific test file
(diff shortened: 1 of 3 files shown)


AGENTS.md
```diff
@@ -20,43 +22,13 @@ Spartacus is an Angular meta-framework for SAP Commerce Cloud e-commerce storefr
...
-npm run prettier:fix
-
-# E2E
-npm run e2e:run
-npm run e2e:run:b2b
-```
-
-## Critical Rules
-
-- **DO** export members in public API and use `protected`/`public` modifiers, because everything should be extendable by customers (except when explicitly marked as "internal API")
```
pawelfras, line 62: IMO we could keep this until we address it with linter rule 👍

pawelfras: As discussed offline, in most of cases access modifiers are set properly, so there is a high possibility coding agent will use the proper pattern. Apart from that, the idea is to start with minimal doc and see how agents will behave. Thanks! Marking this comment as resolved.

pawelfras (approved):
LGTM

PR #21236, opened 2026-03-11
fix: fix issue with adding products to cart on slow networks

uroslates:
Closes: https://jira.tools.sap/browse/CXSPA-10582.

fix: (CXSPA-10582) prevent lost cart entries and phantom carts on slow networks

**Context:** When users rapidly click "Add to Cart" for multiple products (on slow 3G networks), concurrent `requireLoadedCart()` calls each detect an empty cart and dispatch parallel `CREATE_CART` actions. The cart creation effect uses `mergeMap` which processes these in parallel, causing the backend to create multiple carts (race conditions). Products then get added to different carts, with only the last-created cart becoming active. Optimally they get added to 1 cart. After fixing issues with card creations - rapid multi-product add-to-cart on slow networks lost line items in the UI and could place an order against a half loaded cart, creating a phantom cart server-side. Three coordinated fixes:

- multi-cart reducer: optimistically merge the entry returned by CART_ADD_ENTRY_SUCCESS into the cart entity so the UI reflects each POST before the trailing GET reconcile lands.
- cart effect: rewrite refreshWithoutProcesses$ to wait for the falling edge of processesCount per cart (groupBy + concatMap + take(1)) before dispatching a single LoadCart, so the downstream "drop LoadCart while pending" filter no longer swallows every refresh during a burst.
- Place Order button: gate on ActiveCartFacade.isStable() so the order cannot be submitted while cart writes are still in flight; submitForm() also re-checks isStable() before launching the order. Mirrored in the scheduled-replenishment subclass.

Adds unit coverage for the reducer merge path, the falling-edge effect behaviour, and the Place Order gate. New Cypress regression in `cart-race-condition.e2e.cy.ts` simulates slow POST .../entries and asserts all entries land and the Place Order button stays disabled until a state of quietness is reached.

**QA Steps:**
1. Run the application with the flag `enableCartSlowNetworkResilience` set to true.
2. Go to PLP page (with more than 1/2 products).
3. Turn slow 3g network on (in your Network tab)
4. Try quickly adding multiple products to the cart (add some multiple times eg 2/3 times, by clicking esc when confirmation popup shows).
5. While requests are being performed verify that the card button has loading indicator)
6. Click on the cart and verify it is pending/loading and once finished verify correct items (and correct quantities are shown in the cart).

Note: Above when FT `enableCartSlowNetworkResilience` is `false` is actually falling back to the bug with the cart and nr. of items being reproducable (attached video [url](https://sap.enterprise.slack.com/files/W8TPWKE65/F0952BTAE2Y/grabaci__n_de_pantalla_2025-07-02_a_las_16.30.25_3.mov)).


feature-libs/cart/base/core/store/effects/cart.effect.ts
```diff
@@ -146,7 +147,9 @@ export class CartEffects {
     this.actions$.pipe(
       ofType(CartActions.CREATE_CART),
       map((action: CartActions.CreateCart) => action.payload),
-      mergeMap((payload) => {
+      // Use exhaustMap to prevent parallel cart creation - ignores new CREATE_CART
+      // actions while one is in-flight, preventing race conditions on slow networks
+      exhaustMap((payload) => {
```
sdrozdsap: `major`: This change seems very dangerous to me. In my understanding with `exhaustMap` being used here we will be silently dropping legitimate cart creations. This might cause bugs in guest cart merging, or saved carts creation. Generally I haven't checked all code paths that might be affected by this, but I would flag this change as critical and if you are 100% this doesn't cause any regression, I would appreciate if you could provide some analysis that you did in order to make sure this change is safe.

It appears to me that caching might be enough to solve this issue, without using exhaustMap here

feature-libs/cart/base/core/facade/active-cart.service.ts
```diff
@@ -43,6 +45,12 @@ import {
   isTempCartId,
 } from '../utils/utils';
 
+/**
+ * Timeout for cart loading/creation operations in milliseconds.
+ * Protects against hung requests keeping the cache indefinitely.
+ */
+const CART_LOAD_TIMEOUT_MS = 15000;
```
sdrozdsap: `major`: This doesn't seem as the best possible solution, as it might cause issues if request takes longer that 15 seconds for some reason. And generally using timeouts with magic numbers can cause non-stability and potential bugs which might be hard to debug.

feature-libs/cart/base/core/facade/active-cart.service.ts
```diff
@@ -353,7 +361,48 @@ export class ActiveCartService implements ActiveCartFacade, OnDestroy {
...
+
   requireLoadedCart(forGuestMerge = false): Observable<Cart> {
+    // For guest merge, don't use caching as it has special filtering requirements
+    // For normal flow, return cached observable if cart creation is in-flight
+    if (!forGuestMerge && this.loadedCart$) {
+      return this.loadedCart$;
+    }
+
+    const cart$ = this.buildRequireLoadedCartPipeline(forGuestMerge).pipe(
+      // Timeout protection: clear cache if cart creation hangs (triggers finalize to clear the cache allowing susequent calls to retry/start a new cart creation flow)
```
sdrozdsap: `minor`: typo "susequent" -> "subsequent"

feature-libs/cart/base/core/facade/active-cart.service.ts
```diff
@@ -353,7 +361,48 @@ export class ActiveCartService implements ActiveCartFacade, OnDestroy {
...
+      timeout(CART_LOAD_TIMEOUT_MS),
```
sdrozdsap: `query`: Is this really necessary? HTTP layer has its own handling of request timeout which will provide proper request failure handling.
`shareReplay({ refCount: true })` + `finalize` already clears the cache on completion

Considering that timeout might generally not work if requests are just taking longer I'd suggest removing it entirely. 

Another point here is that we would need to handle `TimeoutError` now, which is not handled in subscriber of this method, e.g. `addEntry` method

Isn't issue solved without the timeout? Please let me know if I'm missing something here

sdrozdsap (commented):
I'm a bit worried about these changes, I've added couple comments - however I realize I might have missed something, so please fell free to provide some additional details to me where suitable

feature-libs/cart/base/core/facade/active-cart.service.ts
uroslates: 👍 makes sense - Instead of using timeout I have now implemented cache clearance when the cart is successfully emitted which should add simpler and more predictable behaviour. New changes are:
1. removing magic timeout numbers
2. working on any network speed and clear the cache deterministically (on success/error).

feature-libs/cart/base/core/facade/active-cart.service.ts
uroslates: 👍 code is removed as part of previous comment work

feature-libs/cart/base/core/facade/active-cart.service.ts
uroslates: `timeout` and magic number timeout logic is removed as part of previous PR comment. So I believe this is hence not relevant any more? Please correct me if I'm wrong (after my latest commit).

feature-libs/cart/base/core/store/effects/cart.effect.ts
uroslates: 👍 makes sense. I have now reverted this change which makes race condition fix now to rely solely on `shareReplay` caching in `active-cart.service.ts`.
The behaviour can now be described as following: Multiple rapid addEntry() calls -> requireLoadedCart() returns SAME cached observable -> Only ONE createCart action dispatched -> Effect processes it normally with mergeMap.

projects/storefrontapp-e2e-cypress/cypress/e2e/regression/cart/cart-race-condition.e2e.cy.ts
sdrozdsap, line 1: This files still references exhaustMap which was reverted. Maybe we need to revisit this test

sdrozdsap:
Also can you please update PR description according to latest changes?

projects/storefrontapp-e2e-cypress/cypress/e2e/regression/cart/cart-race-condition.e2e.cy.ts
uroslates: 👍 Applied

sdrozdsap (dismissed):
Im adding approve from my end, however I don't have a 100% confidence in this changes just because of potential risks I see whenever active-cart service has changes, and I'd appreciate if we could involve a second pair of eyes to look at the changes as well, so Ill unassign myself from the task leaving it in the code review column.

Also a question: shouldn't we use a feature flag for this changes as well? I'm wondering if this was considered and if there is a reason for not adding it?

p.s. PR description is not up to date anymore

feature-libs/cart/base/core/facade/active-cart.service.ts
```diff
@@ -353,7 +354,54 @@ export class ActiveCartService implements ActiveCartFacade, OnDestroy {
...
 
+  /**
+   * Cached observable for in-flight cart loading/creation.
+   * Prevents race conditions when multiple addEntry() calls happen rapidly
+   * before the cart is created - they all share the same cart creation flow.
+   */
+  private loadedCart$: Observable<Cart> | null = null;
+
   requireLoadedCart(forGuestMerge = false): Observable<Cart> {
+    // For guest merge, don't use caching as it has special filtering requirements
```
npapp-dev002, line 431: The error described in this video still can be reproduced even with these changes: [url](https://sap.enterprise.slack.com/files/W8TPWKE65/F0952BTAE2Y/grabaci__n_de_pantalla_2025-07-02_a_las_16.30.25_3.mov)

feature-libs/cart/base/components/cart-details/cart-details.component.ts
```diff
@@ -73,6 +100,15 @@ export class CartDetailsComponent implements OnInit {
           : cartLoaded
       )
     );
+
+    this.updating$ = this.featureToggles.enableCartSlowNetworkResilience
```
Pucek9: Given how easy it is to customize, it's a good idea to move this into a separate protected method and call it here.

feature-libs/cart/base/components/cart-proceed-to-checkout/cart-proceed-to-checkout.component.ts
```diff
@@ -18,23 +20,59 @@ import {
...
+   * `isStable()` selector cannot lock the user out indefinitely.
+   */
+  cartUpdating$: Observable<boolean>;
+
   protected subscription = new Subscription();
+  private featureToggles = inject(FeatureToggles);
 
   constructor(
     protected router: Router,
+    protected activeCartFacade: ActiveCartFacade,
```
Pucek9: Oh man! Broken rules here 😅
- Adding a new parameter is allowed only at the end, and only optional.
- Prefer inject instead of increasing the constructor members

feature-libs/cart/base/core/facade/active-cart.service.ts
```diff
@@ -353,7 +355,65 @@ export class ActiveCartService implements ActiveCartFacade, OnDestroy {
...
 
+  private featureToggles = inject(FeatureToggles);
+
+  /**
+   * Cached observable for in-flight cart loading/creation.
+   * Prevents race conditions when multiple addEntry() calls happen rapidly
+   * before the cart is created - they all share the same cart creation flow.
+   * Only used when `enableCartSlowNetworkResilience` is enabled.
+   */
+  private loadedCart$: Observable<Cart> | null = null;
```
Pucek9:
```suggestion
  protected loadedCart$: Observable<Cart> | null = null;
```

feature-libs/cart/base/core/facade/active-cart.service.ts
```diff
@@ -353,7 +355,65 @@ export class ActiveCartService implements ActiveCartFacade, OnDestroy {
...
+    }
+
+    return cart$;
+  }
+
+  /**
+   * Internal method containing the cart loading/creation pipeline logic.
+   * Extracted to support caching in requireLoadedCart().
+   */
+  private buildRequireLoadedCartPipeline(
```
Pucek9:
```suggestion
  protected buildRequireLoadedCartPipeline(
```

feature-libs/checkout/base/components/checkout-place-order/checkout-place-order.component.ts
```diff
@@ -76,7 +117,8 @@ export class CheckoutPlaceOrderComponent implements OnDestroy, OnInit {
     protected routingService: RoutingService,
     protected fb: UntypedFormBuilder,
     protected launchDialogService: LaunchDialogService,
-    protected vcr: ViewContainerRef
+    protected vcr: ViewContainerRef,
+    protected activeCartFacade: ActiveCartFacade
```
Pucek9:
```suggestion
```
Please use inject instead

Pucek9:
Please add *cxFeature="'enableCartSlowNetworkResilience'" into **every** added/changed html block

Pucek9:
Please don't complicate the logic for checking featureToggles by using shorthand notation, etc. Use simple if/else statements—they're easier for whoever has to remove them later.

feature-libs/cart/base/components/cart-details/cart-details.component.ts
uroslates: Since **CartDetailsComponent** has no subclasses anywhere in the repo, I would leave it as is - mostly because **CartDetailsComponent**'s `updating$` is assigned in `ngOnInit` and that pattern already allows a subclass to override `ngOnInit` and reassign it, which is the natural Angular extension point, no protected factory needed.

feature-libs/cart/base/components/cart-proceed-to-checkout/cart-proceed-to-checkout.component.ts
uroslates: Good catch! Thanks will apply the change! Applied 👍 

Speaking of which we should probably create a linter ticket enforcing this rule in the project!

feature-libs/cart/base/core/facade/active-cart.service.ts
uroslates: Applied 👍

feature-libs/cart/base/core/facade/active-cart.service.ts
uroslates: Applied 👍

feature-libs/checkout/base/components/checkout-place-order/checkout-place-order.component.ts
uroslates: Applied 👍

feature-libs/cart/base/components/cart-details/cart-details.component.ts
Pucek9: I don't understand the argument that we don't have subclasses in the repository. Our clients will have subclasses, and that's what I'm worried about here.

Yes, you can override onInit, but unfortunately, you have to override the entire onInit. Granularity allows you to override only the parts of the logic that clients actually need to change. I proposed a change to the coding guidelines.

Update: this point has been approved and merged
https://github.tools.sap/cx-commerce-storefronts/spartacus-requirements/pull/75/files

feature-libs/cart/base/components/cart-proceed-to-checkout/cart-proceed-to-checkout.component.ts
```diff
@@ -50,12 +90,38 @@ export class CartProceedToCheckoutComponent implements OnInit, OnDestroy {
...
   }
 
   disableButtonWhileNavigation(): void {
     this.cartValidationInProgress = true;
   }
 
+  protected isSlowNetworkResilienceEnabled(): boolean {
+    return !!this.featureToggles.enableCartSlowNetworkResilience;
+  }
+
```
Pucek9: Please check why its antipattern:
https://github.tools.sap/cx-commerce-storefronts/spartacus-requirements/blob/main/spartacus-classic/coding-guidelines.md#using-feature-toggles-in-html

```suggestion
```

feature-libs/cart/base/components/mini-cart/mini-cart-component.service.ts
```diff
@@ -76,6 +89,40 @@ export class MiniCartComponentService {
...
+   * required (anonymous, no cart in storage, no CreateCart event) we don't
+   * touch ActiveCartFacade — keeps the cart code chunk out of the initial
+   * load.
+   *
+   * Gated by `enableCartSlowNetworkResilience`; emits constant `false` when
+   * the toggle is OFF so an extending client sees pre-CXSPA-10582 behaviour
+   * (no indicator).
+   */
+  getUpdating(): Observable<boolean> {
+    if (!this.featureToggles.enableCartSlowNetworkResilience) {
```
Pucek9, line 108: if new code changes behavior of only new html blocks, which are already covered by cxFeature, we don't need to handle it twice

feature-libs/cart/base/components/cart-details/cart-details.component.ts
uroslates: Applied! 👍

feature-libs/cart/base/components/cart-details/cart-details.component.ts
```diff
@@ -73,6 +102,19 @@ export class CartDetailsComponent implements OnInit {
...
+
+  protected buildUpdating$(): Observable<boolean> {
+    return this.featureToggles.enableCartSlowNetworkResilience
+      ? this.activeCartService.isStable().pipe(
+          map((stable) => !stable),
+          debounceTime(CART_DETAILS_UPDATING_DEBOUNCE_MS),
+          startWith(false),
+          distinctUntilChanged()
+        )
+      : of(false);
```
Pucek9:
```suggestion
    this.buildUpdating();
  }

  protected buildUpdating$(): Observable<boolean> {
    this.updating$ = this.activeCartService.isStable().pipe(
          map((stable) => !stable),
          debounceTime(CART_DETAILS_UPDATING_DEBOUNCE_MS),
          startWith(false),
          distinctUntilChanged()
        )
```

feature-libs/cart/base/components/cart-proceed-to-checkout/cart-proceed-to-checkout.component.ts
```diff
@@ -50,6 +94,28 @@ export class CartProceedToCheckoutComponent implements OnInit, OnDestroy {
         }
       })
     );
+
+    if (!this.featureToggles.enableCartSlowNetworkResilience) {
+      this.cartUpdating$ = of(false);
```
Pucek9: same antipattern

feature-libs/cart/base/components/cart-proceed-to-checkout/cart-proceed-to-checkout.component.ts
```diff
@@ -50,6 +94,28 @@ export class CartProceedToCheckoutComponent implements OnInit, OnDestroy {
...
+    this.cartUpdating$ = combineLatest([
+      this.activeCartFacade.isStable(),
+      safetyValveExpired$,
+    ]).pipe(
+      map(([stable, expired]) => !stable && !expired),
+      debounceTime(PROCEED_TO_CHECKOUT_UPDATING_DEBOUNCE_MS),
+      startWith(false),
+      distinctUntilChanged()
+    );
   }
```
Pucek9, line 112:
```suggestion
    this.initializeCartUpdating();
  }  

protected initializeCartUpdating() {
    const safetyValveExpired$ = timer(
      PROCEED_TO_CHECKOUT_GATE_SAFETY_VALVE_MS
    ).pipe(
      map(() => true),
      startWith(false)
    );

    this.cartUpdating$ = combineLatest([
      this.activeCartFacade.isStable(),
      safetyValveExpired$,
    ]).pipe(
      map(([stable, expired]) => !stable && !expired),
      debounceTime(PROCEED_TO_CHECKOUT_UPDATING_DEBOUNCE_MS),
      startWith(false),
      distinctUntilChanged()
    );
}
```

uroslates: Addressed. 👍

feature-libs/cart/base/components/cart-proceed-to-checkout/cart-proceed-to-checkout.component.ts
uroslates: Addressed!  👍

feature-libs/cart/base/components/cart-details/cart-details.component.ts
uroslates: Addressed!

feature-libs/checkout/scheduled-replenishment/components/checkout-place-order/checkout-place-order.component.html
```diff
@@ -20,12 +20,37 @@
...
-  <button
-    (click)="submitForm()"
-    class="btn btn-primary btn-block"
-    [disabled]="termsAndConditionInvalid || (daysOfWeekNotChecked$ | async)"
-    [cxAtMessage]="'checkoutReview.orderInProcess' | cxTranslate"
-  >
-    {{ 'checkoutReview.placeOrder' | cxTranslate }}
-  </button>
+  <ng-container *cxFeature="'enableCartSlowNetworkResilience'">
+    <button
```
Pucek9: [NITPICK]
If there is no other structural directive on the element (eg. *ngIf) you can use *cxFeature directly on the element (no ng-container needed)
```suggestion
  <button *cxFeature="'enableCartSlowNetworkResilience'">
```
same for belows

feature-libs/cart/base/core/store/reducers/multi-cart.reducer.ts
```diff
@@ -39,7 +39,46 @@ export function cartTypeIndexReducer(
...
+ * Reducers cannot use Angular DI, so the value is captured at app bootstrap
+ * by `getMultiCartReducers()` (see `./index.ts`) and read here at action
+ * dispatch time. When false, the reducer skips the optimistic
+ * `CART_ADD_ENTRY_SUCCESS` merge and behaves identically to pre-CXSPA-10582.
+ */
+let cartSlowNetworkResilienceEnabled = false;
+
+export function setCartSlowNetworkResilienceEnabled(enabled: boolean): void {
+  cartSlowNetworkResilienceEnabled = enabled;
+}
```
Pucek9: This is very fancy and dangerous - module-level mutable state in a reducer file is a purity violation.

feature-libs/checkout/scheduled-replenishment/components/checkout-place-order/checkout-place-order.component.html
uroslates: Addressed! 👍

feature-libs/cart/base/core/store/reducers/multi-cart.reducer.ts
uroslates: Addressed! 👍

Pucek9 (dismissed):

rmch91 (approved):

feature-libs/cart/base/components/mini-cart/mini-cart-component.service.ts
Pucek9: ah, ok you based it in uncovered by cxFeature element
```[class.is-updating]="data.updating"```
but it seems safe

PR #21249, opened 2026-03-16
feat: (CXSPA-12331) - Listbox semantics for facet component

Pio-Bar:
Closes: https://jira.tools.sap/browse/CXSPA-12331

The links in facets should be contained in a listbox. This way we ensure JAWS does not override keyboard controls and `visible-focus` outline will be visible. PR Includes attribute changes to improve JAWS output as well.

This issue needs a screen reader to be reviewed. Please, ensure the outline is visible while JAWS is on and the reader output is satisfactory.


uroslates (approved):
👍 Tested locally using JAWS and selecting options using arrows now works fine and outlines what is highlighted in the list. Check the attached screenshot:

Image

PR #21258, opened 2026-03-17
fix: (CXSPA-12321) Add quick order error msg for empty product ID

BenjaminEmiliani:
Closes [CXSPA-12321](https://jira.tools.sap/browse/CXSPA-12321)


Pucek9 (changes requested):
There are 2 ways:
- one is a disabled button for an invalid form (afaik we got rid of it in most of the cases)
- always enabled button - clicking will trigger validation, only a valid form will be sent (a11y friendly)

I can see that the 1st approach was applied. Anyway, clicking on a disabled button shouldn't trigger any event

Seems you need to switch from way 1 to way 2 for a11y improvements:
- create a new a11y feature toggle flag
- handle old button variant with flag negation
- handle new button without disabled attribute with flag
- add validation (cx-form-errors)

feature-libs/cart/quick-order/components/cart-quick-order-form/cart-quick-order-form.component.html
```diff
@@ -37,14 +42,22 @@
...
+          [control]="quickOrderForm.get('productCode')"
+        ></cx-form-errors>
         <button
           [attr.aria-label]="'quickOrderCartForm.addToCart' | cxTranslate"
           [class.disabled]="cartIsLoading$ | async"
           [disabled]="
             (cartIsLoading$ | async) ||
-            !quickOrderForm?.get('productCode')?.value ||
-            !quickOrderForm?.get('quantity')?.value
+            (!useFormValidation &&
```
Pucek9: nice try, but why I don't like that result:
-  you already have `FeatureDirective` imported into the component anyway, so it won't give you any benefits (if your goal was to reduce its weight)
- our target with introducing feature toggle wrappers is also to make them easy to remove in the future. Adding logic (logical expressions) will make it harder for someone who remove them in the future. 
In this case, it is hard to guess at first glance (or maybe even after thinking for a while) what the final result after removing the flag is supposed to be. 
I mean, for us it's clear that it should remain `[disabled]="cartIsLoading$ | async"` because we know the context of the introduced change. For a person who is doing the cleanup of 10 random feature toggle wrappers, it could cause unnecessary confusion

Just sometimes [KISS](https://en.wikipedia.org/wiki/KISS_principle) is better than [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself)

BenjaminEmiliani: I thought of this too but was not sure what takes priority. Since removing the featureConfig TS logic results in static error in HTML file, I thought it would be easy to maintain in future - but ultimately I agree.

feature-libs/cart/quick-order/components/cart-quick-order-form/cart-quick-order-form.component.html
```diff
@@ -12,6 +12,11 @@
           </label>
 
           <input
+            [attr.aria-invalid]="
```
Pucek9, line 14: Considering it as a breaking change as well 🤔

Pucek9: Update:
Form is created in a component, not in a component.service, so if overriding the component will also require introducing its own HTML.
IMO not a breaking change

projects/core/src/features-config/feature-toggles/config/feature-toggles.ts
```diff
@@ -437,6 +437,14 @@ export interface FeatureTogglesInterface {
...
    */
   a11yReviewsKeyboardControls?: boolean;
+
+  /**
+   * Use on existing form buttons that are programatically disabled/enabled.
+   * To use, duplicate button and use false in original and true in duplicate. The duplicated button
+   * should be initialized as enabled, clickable and use cx-form-errors in outcomes where original button
+   * is in disabled state.
+   */
+  a11yEnableButtonAndUseCxFormErrors?: boolean;
```
Pucek9: Name sounds very generic, are we going to use is in other places/tickets?
If not, it's worth specifying that it's for a quick order form only

Pucek9 (approved):

PR #21260, opened 2026-03-18
fix: suppress i18next ad in logs

Platonn:
Note: this PR fixes an issue that happens in customers' apps that use higher minor version of i18next than we in our repo. So you won't see the ad in our storeforntapp in our repo, but only in fresh apps using latest i18next minor.

fixes https://jira.tools.sap/browse/CXSPA-12274

-- 

QA steps:
- build libs and publish to verdaccio
- create fresh app with angular and spartacus (from verdaccio)
- start the app
- check the browser's console 
- verify there is no i18next ad in the console


npapp-dev002 (approved):
LGTM

PR #21268, opened 2026-03-19
fix: handle full page redirection within smartedit

FollowTheFlo:


feature-libs/smartedit/root/services/smart-edit-launcher.service.ts
```diff
@@ -53,16 +59,61 @@ export class SmartEditLauncherService {
...
    */
   isLaunchedInSmartEdit(): boolean {
-    const path = this.location.path().split('?')[0];
-    const params = this.location.path().split('?')[1];
+    const [path, params] = this.location.path().split('?');
     const cmsToken = params
       ?.split('&')
       .find((param) => param.startsWith('cmsTicketId='));
-    this._cmsTicketId = cmsToken?.split('=')[1];
+    const cmsTicketId = cmsToken?.split('=')[1] ?? undefined;
```
uroslates: [very minor]  I believe it already returns undefined if no match. Maybe we should ommit `?? undefined`.

feature-libs/smartedit/root/services/smart-edit-launcher.service.spec.ts
```diff
@@ -26,21 +36,25 @@ describe('SmartEditLauncherService', () => {
   let location: Location;
   let scriptLoader: ScriptLoader;
   let featureModules: FeatureModulesService;
+  let windowRef: WindowRef;
 
   beforeEach(() => {
+    sessionStorage.clear();
```
uroslates, line 42: Should we also cleanup after test to ensure cleanup?
```
afterEach(() => {
  sessionStorage.clear();
});
```

feature-libs/smartedit/root/services/smart-edit-launcher.service.ts
```diff
@@ -53,16 +59,61 @@ export class SmartEditLauncherService {
...
+    cmsTicketId: string | undefined
+  ): boolean {
+    // When both the SmartEdit cmsTicketId and the storefrontPreviewRoute values are found in the URL, store the cmsTicketId in sessionStorage.
+    // so it survives full-page navigation (e.g. CDC OIDC redirect flow).
+    if (
       path.split('/').pop() === this.config.smartEdit?.storefrontPreviewRoute &&
-      !!this._cmsTicketId
+      !!cmsTicketId
+    ) {
+      this.persistCmsTicketId(cmsTicketId);
```
uroslates, line 84: Question: I might be wrong but looks like `cmsTicketId` is never cleared from `sessionStorage` when leaving SmartEdit. Do you think we should add the  cleanup mechanism (is it relevant)?

feature-libs/smartedit/root/services/smart-edit-launcher.service.ts
```diff
@@ -53,16 +59,61 @@ export class SmartEditLauncherService {
...
+    const [path, params] = this.location.path().split('?');
     const cmsToken = params
       ?.split('&')
       .find((param) => param.startsWith('cmsTicketId='));
-    this._cmsTicketId = cmsToken?.split('=')[1];
+    const cmsTicketId = cmsToken?.split('=')[1] ?? undefined;
 
     return (
+      this.isInitialSmartEditPage(path, cmsTicketId) ||
+      this.isFullPageRedirectInSmartEdit()
```
uroslates, line 70: in SmartEdit detection it seems to me that there might be potential false positives - in `isFullPageRedirectInSmartEdit()` the `cmsTicketId` is restored from `sessionStorage` unconditionally on every page (method returns true if ANY old ticket exists). 
I am not sure **how relevant this use case might be** but if a user was in SmartEdit yesterday and closed the browser tab (sessionStorage has persisted), than opens the regular storefront today - this will incorectly report `isLaunchedInSmartEdit() === true`.
Suggestion: Maybe we should add validation checking if we're still in an iframe or SmartEdit context (in which case we return false) - something like following (again not sure if relevant - leave it to you to decide):
```ts
private isFullPageRedirectInSmartEdit(): boolean {
  if (this.windowRef.nativeWindow?.self === this.windowRef.nativeWindow?.top) {
    return false; // Not in iframe, not SmartEdit redirect
  }
  this._cmsTicketId = this.restoreCmsTicketId();
  return !!this._cmsTicketId;
}
```

uroslates (dismissed):
Great work! Changes look really good. 👍 

I have only few questions/remark, leave it to you to reply and decide on if those are relevant.

feature-libs/smartedit/root/services/smart-edit-launcher.service.spec.ts
FollowTheFlo: thx, sounds good and safer

feature-libs/smartedit/root/services/smart-edit-launcher.service.ts
FollowTheFlo: i followed fix made on spa-bff by Roman, i believed it is to avoid throwing an null object
https://github.tools.sap/cx-commerce-storefronts/spartacus-bff/pull/1179 (line 45)

feature-libs/smartedit/root/services/smart-edit-launcher.service.ts
FollowTheFlo: Thx, i already tested this scenario (launch smartedit, then launch sparatcus in same tab), It is considered as a new session because smartedit launch spartacus in an iframe. So, session storage get cleared automatically. We are good.

feature-libs/smartedit/root/services/smart-edit-launcher.service.ts
FollowTheFlo: second thought, it is indeed not needed in current logic.

uroslates (approved):
👍

PR #21272, opened 2026-03-23
fix: (CXSPA-9504) support Chinese address in address book

espada945:
## Summary
- Add cascading city/district dropdowns when China is selected as country
- Change "State" label to "Province" for Chinese addresses
- Make cellphone required for Chinese addresses
- Add district field for Chinese addresses
## Test plan
- [ ] Select China as country → verify city/district dropdowns and Province label
- [ ] Add complete Chinese address → verify successful save
- [ ] Switch to non-China country → verify form reverts to default


feature-libs/user/profile/components/address-book/address-form/address-form.component.ts
```diff
@@ -130,7 +144,9 @@ export class AddressFormComponent implements OnInit, OnDestroy {
     protected globalMessageService: GlobalMessageService,
     protected translation: TranslationService,
     protected launchDialogService: LaunchDialogService,
-    protected userProfileFacade: UserProfileFacade
+    protected userProfileFacade: UserProfileFacade,
```
scarai-sap: use inject instead of change constructor

feature-libs/user/profile/components/address-book/address-form/address-form.component.ts
```diff
@@ -72,6 +80,11 @@ export class AddressFormComponent implements OnInit, OnDestroy {
   titles$: Observable<Title[]>;
   regions$: Observable<Region[]>;
   selectedCountry$: BehaviorSubject<string> = new BehaviorSubject<string>('');
+  selectedRegion$: BehaviorSubject<string> = new BehaviorSubject<string>('');
+  selectedCity$: BehaviorSubject<string> = new BehaviorSubject<string>('');
+  isChinaAddress = false;
```
scarai-sap: Should isChineseAddress?

feature-libs/user/profile/components/address-book/address-form/address-form.component.ts
```diff
@@ -171,6 +187,34 @@ export class AddressFormComponent implements OnInit, OnDestroy {
...
+      switchMap((regionIsocode) => {
+        if (!regionIsocode) {
+          return of([]);
+        }
+        const baseUrl = this.occEndpointsService.getBaseUrl();
+        return this.http
+          .get<any>(
+            `${baseUrl}/regions/${regionIsocode}/cities?fields=cities(name,isocode)`
+          )
+          .pipe(map((res) => res.cities ?? []));
```
scarai-sap: where is error handling?

feature-libs/user/profile/components/address-book/address-form/address-form.component.ts
```diff
@@ -171,6 +187,34 @@ export class AddressFormComponent implements OnInit, OnDestroy {
...
+      switchMap((cityIsocode) => {
+        if (!cityIsocode) {
+          return of([]);
+        }
+        const baseUrl = this.occEndpointsService.getBaseUrl();
+        return this.http
+          .get<any>(
+            `${baseUrl}/cities/${cityIsocode}/districts?fields=districts(name,isocode)`
+          )
+          .pipe(map((res) => res.districts ?? []));
```
scarai-sap: where is error handling?

feature-libs/user/profile/components/address-book/address-form/address-form.component.ts
```diff
@@ -214,10 +258,37 @@ export class AddressFormComponent implements OnInit, OnDestroy {
...
     this.selectedCountry$.next(country?.isocode ?? '');
+    this.isChinaAddress = country?.isocode === 'CN';
+
+    const cellphoneControl = this.addressForm.get('cellphone');
+    const districtControl = this.addressForm.get('district');
+    if (this.isChinaAddress) {
+      cellphoneControl?.setValidators([Validators.required]);
+      districtControl?.setValidators([Validators.required]);
+    } else {
+      cellphoneControl?.clearValidators();
```
scarai-sap, line 297: adding reset data as  this.selectedRegion$.next('');this.selectedCity$.next('');

feature-libs/user/profile/components/address-book/address-form/address-form.component.ts
```diff
@@ -72,6 +80,11 @@ export class AddressFormComponent implements OnInit, OnDestroy {
   titles$: Observable<Title[]>;
   regions$: Observable<Region[]>;
   selectedCountry$: BehaviorSubject<string> = new BehaviorSubject<string>('');
+  selectedRegion$: BehaviorSubject<string> = new BehaviorSubject<string>('');
+  selectedCity$: BehaviorSubject<string> = new BehaviorSubject<string>('');
+  isChinaAddress = false;
+  cities$: Observable<any[]>;
```
scarai-sap: try to avoid using any

feature-libs/user/profile/components/address-book/address-form/address-form.component.ts
```diff
@@ -171,6 +187,34 @@ export class AddressFormComponent implements OnInit, OnDestroy {
...
     this.addresses$ = this.userAddressService.getAddresses();
+
+    this.cities$ = this.selectedRegion$.pipe(
+      switchMap((regionIsocode) => {
+        if (!regionIsocode) {
+          return of([]);
+        }
+        const baseUrl = this.occEndpointsService.getBaseUrl();
+        return this.http
+          .get<any>(
```
scarai-sap: try to avoid using any

feature-libs/user/profile/components/address-book/address-form/address-form.component.ts
```diff
@@ -171,6 +187,34 @@ export class AddressFormComponent implements OnInit, OnDestroy {
...
+            `${baseUrl}/regions/${regionIsocode}/cities?fields=cities(name,isocode)`
```
scarai-sap: Don't hardcode url in component

feature-libs/user/profile/components/address-book/address-form/address-form.component.html
```diff
@@ -247,6 +272,35 @@
...
+        >
+        </ng-select>
+        <cx-form-errors
+          [translationParams]="{
+            label: 'addressForm.district' | cxTranslate,
+          }"
+          [control]="addressForm.get('district')"
+        ></cx-form-errors>
+      </label>
+    </div>
```
i336588, line 285: The display order of the province and city should be adapted as the city is only available after selecting one province for a Chinese address. Besides, the place holder for city and district is empty when province is not selected. You'd better ask the UX's advice how to display these fields for Chinese address.

Image

feature-libs/user/profile/components/address-book/address-form/address-form.component.ts
```diff
@@ -171,6 +187,34 @@ export class AddressFormComponent implements OnInit, OnDestroy {
...
+
+    this.districts$ = this.selectedCity$.pipe(
+      switchMap((cityIsocode) => {
+        if (!cityIsocode) {
+          return of([]);
+        }
+        const baseUrl = this.occEndpointsService.getBaseUrl();
+        return this.http
+          .get<any>(
+            `${baseUrl}/cities/${cityIsocode}/districts?fields=districts(name,isocode)`
```
i336588: Do not call the http request directly here. The cities and districts should be get through userAddressService. You can refer to the implementation of countries and regions for details.

feature-libs/user/profile/components/address-book/address-form/address-form.component.ts
```diff
@@ -171,6 +187,34 @@ export class AddressFormComponent implements OnInit, OnDestroy {
...
+          )
+          .pipe(map((res) => res.districts ?? []));
+      })
+    );
```
i336588, line 242: There's a issue: When switching language to Chinese, the cities and districts list is not refreshed and keep in previous language.

Image

feature-libs/user/profile/components/address-book/address-form/address-form.component.spec.ts
```diff
@@ -14,6 +15,7 @@ import {
   Country,
   GlobalMessageService,
   I18nTestingModule,
+  OccEndpointsService,
```
i336588: The address book should also be adapted for Chinese address. The countries, cities and districts should use names instead of isocode.

Image

feature-libs/user/profile/components/address-book/address-form/address-form.component.html
espada945: Fixed, Louis said LGTM.

feature-libs/user/profile/components/address-book/address-form/address-form.component.ts
```diff
@@ -214,10 +280,70 @@ export class AddressFormComponent implements OnInit, OnDestroy {
...
+      .subscribe((text) => (this.cityPlaceholder = text));
+
+    this.translation
+      .translate(
+        this.selectedCity$.value
+          ? 'addressForm.selectOne'
+          : 'addressForm.selectCityFirst'
+      )
+      .pipe(take(1))
+      .subscribe((text) => (this.districtPlaceholder = text));
```
i336588: The place holders sometimes are not displayed.

Image

Image

feature-libs/user/profile/components/address-book/address-book.component.ts
```diff
@@ -125,16 +125,21 @@ export class AddressBookComponent implements OnInit {
...
+
           return {
             role: 'application',
             textBold: address.firstName + ' ' + address.lastName,
             text: [
               address.line1,
               address.line2,
-              address.town + ', ' + region + address.country?.isocode,
+              locationParts,
+              districtName,
```
i336588, line 139: The country and province is not translated in address book:

Image

feature-libs/user/profile/components/address-book/address-form/address-form.component.ts
```diff
@@ -214,10 +280,70 @@ export class AddressFormComponent implements OnInit, OnDestroy {
...
+  }
+
+  citySelected(city: City | undefined): void {
+    if (city?.isocode) {
+      this.selectedCity$.next(city.isocode);
+      this.addressForm.get('district')?.reset();
+      this.addressForm.get('district')?.enable();
+      this.updateChinesePlaceholders();
+    }
+  }
```
i336588, line 324: When click to edit an existing Chinese address, the city and district are empty. It should show the saved city and district.

Image

feature-libs/user/profile/components/address-book/address-form/chinese-address.service.ts
```diff
@@ -0,0 +1,72 @@
...
+
+@Injectable({
+  providedIn: 'root',
+})
+export class ChineseAddressService {
+  protected http = inject(HttpClient);
+  protected occEndpoints = inject(OccEndpointsService);
+  protected logger = inject(LoggerService);
+
+  getCities(regionIsocode: string): Observable<City[]> {
```
i336588: Should reuse UserAddressService and get cities and districts via OccSiteAdapter. You can refer to the implementation of getCountries and getRegions for details.

feature-libs/user/profile/components/address-book/address-form/address-form.component.ts
```diff
@@ -214,12 +283,48 @@ export class AddressFormComponent implements OnInit, OnDestroy {
...
   }
 
+  citySelected(city: { isocode?: string; name?: string } | undefined): void {
+    if (city?.isocode) {
+      this.selectedCity$.next(city.isocode);
+      this.addressForm.get('district')?.reset();
+    }
+  }
+
+  protected updateChinesePlaceholders(): void {}
```
i336588: This empty method is never used.

projects/core/src/user/facade/user-address.service.ts
```diff
@@ -197,6 +199,19 @@ export class UserAddressService {
       })
     );
   }
+
+  getCities(
+    regionIsocode: string
+  ): Observable<{ isocode?: string; name?: string }[]> {
+    return this.siteConnector.getCities(regionIsocode);
```
i336588: Using NgRx store pattern instead of direct API call would be better for consistency and performance, as getDeliveryCountries() and getRegions() already use.
1. Performance Issues with Current Implementation

  Problem: Multiple API calls for same data
  // In address-form.component.ts:194-218
  combineLatest([this.selectedRegion$, this.languageService.getActive()])
    .pipe(switchMap(([regionIsocode]) => {
      return this.userAddressService.getCities(regionIsocode); // 🔴 API call every time!
    }))

  Every time language changes → new API call
  User switches back to previous region → API call again
  Multiple address forms on screen → duplicate API calls

  2. Consistency with Existing Architecture

  The codebase follows a clear pattern for location data:

  Countries → Store (user/store/reducers/delivery-countries.reducer.ts)
  Regions   → Store (user/store/reducers/regions.reducer.ts)
  Cities    → Direct API ❌ (breaks pattern)
  Districts → Direct API ❌ (breaks pattern)

core-libs/setup/ssr/engine/ng-express-engine.ts
```diff
@@ -84,7 +84,7 @@ export function ngExpressEngine(setupOptions: NgSetupOptions) {
     bootstrap: setupOptions.bootstrap,
     providers: setupOptions.providers,
     enablePerformanceProfiler: setupOptions.enablePerformanceProfiler,
-    allowedHosts: setupOptions.allowedHosts,
+    // allowedHosts: setupOptions.allowedHosts,
```
i336588: Why change this setting?

feature-libs/user/profile/components/address-book/address-book.component.ts
```diff
@@ -110,11 +110,11 @@ export class AddressBookComponent implements OnInit {
           textPhone,
           textMobile,
         ]) => {
-          let region = '';
-
-          if (address.region && address.region.isocode) {
-            region = address.region.isocode + ', ';
-          }
+          const region = this.buildRegion(address);
+          const countryName = address.country?.isocode || '';
```
i336588:
```suggestion
          const countryName = address.country?.name || address.country?.isocode || '';
```

feature-libs/user/profile/components/address-book/address-form/address-form.component.ts
```diff
@@ -171,6 +189,57 @@ export class AddressFormComponent implements OnInit, OnDestroy {
     }
 
     this.addresses$ = this.userAddressService.getAddresses();
+
+    this.subscription.add(
+      this.selectedRegion$
+        .pipe(
+          switchMap((regionIsocode) => {
```
i336588, line 222: When switching language to Chinese, the country/province/city/district does not show in Chinese:

Image

projects/core/src/model/address.model.ts
```diff
@@ -38,8 +38,10 @@ export interface Address {
   line2?: string;
   postalCode?: string;
   town?: string;
+  city?: { isocode?: string; name?: string };
```
i336588: It's better to extract a City interface as below:
```
export interface City {
  regionIso?: string;
  isocode?: string;
  name?: string;
}
```

projects/core/src/model/address.model.ts
```diff
@@ -38,8 +38,10 @@ export interface Address {
...
   region?: Region;
   district?: string;
+  cityDistrict?: { isocode?: string; name?: string };
```
i336588: It's better to extract a CityDistrict interface.
```
export interface CityDistrict {
  cityIso?: string;
  isocode?: string;
  name?: string;
}
```

core-libs/core/src/user/facade/user-address.service.ts
```diff
@@ -197,6 +197,73 @@ export class UserAddressService {
...
   }
+
+  getCities(
+    regionIsocode: string
+  ): Observable<{ isocode?: string; name?: string }[]> {
+    return this.store.pipe(
+      select(UsersSelectors.getCitiesDataAndLoading),
+      map(({ cities, regionIsocode: storedRegion, loading, loaded }) => {
+        if (!regionIsocode && (loading || loaded)) {
+          this.clearCities();
```
i336588, line 220: Not only city, but also district needs to be cleared when changing to another province.
Image

core-libs/core/src/user/facade/user-address.service.ts
```diff
@@ -197,6 +197,73 @@ export class UserAddressService {
...
+          return [];
```
i336588, line 222: When changing Country, the city also need to be cleared.

Image

core-libs/setup/ssr/engine/ng-express-engine.ts
espada945: > Why change this setting?

This was temporarily commented out during local testing and was mistakenly uploaded.

espada945:
@giancorderoortiz 

This PR adds Chinese address support (province/city/district cascading dropdowns). As part of this, we changed the address card to display country.name instead of country.isocode for better UX.

This requires updating a few existing E2E assertions in files you authored:
checkout-variants.ts: 'GB' → 'United Kingdom'
checkout-as-guest.ts: 'US-CA' → 'California'
etc.

Could you take a quick look and let us know if this change is acceptable? Happy to discuss if there are concerns.

feature-libs/user/profile/components/address-book/address-form/address-form.component.ts
```diff
@@ -214,10 +306,46 @@ export class AddressFormComponent implements OnInit, OnDestroy {
...
     this.addressForm.get('region')?.get('isocode')?.setValue(region.isocode);
+    if (this.isChineseAddress) {
+      this.selectedRegion$.next(region.isocode ?? '');
+      this.addressForm.get('town')?.reset();
+      this.selectedCity$.next('');
+      this.addressForm.get('district')?.reset();
+    }
+  }
+
+  citySelected(city: { isocode?: string; name?: string } | undefined): void {
```
i336588:
```suggestion
  citySelected(city: City | undefined): void {
```

feature-libs/user/profile/root/model/user-profile.model.ts
```diff
@@ -30,6 +30,16 @@ export interface Title {
   name?: string;
 }
 
+export interface ChineseCity {
+  name?: string;
+  isocode?: string;
+}
+
+export interface ChineseDistrict {
```
i336588: ChineseCity and ChineseDistrict are dead code which are not used elsewhere at all. Can be removed.

feature-libs/user/profile/components/address-book/address-form/address-form.component.ts
```diff
@@ -171,6 +191,78 @@ export class AddressFormComponent implements OnInit, OnDestroy {
     }
 
     this.addresses$ = this.userAddressService.getAddresses();
+
+    this.subscription.add(
+      this.selectedRegion$
+        .pipe(
+          switchMap((regionIsocode) => {
+            if (!this.isChineseAddress) {
```
i336588: There's one minor issue:
1. Select Canada as country which has state and city, and fill in the required data without saving.
2. Switch to China as country, the place holder in province is gone. The place holder "Select one..." should exist.

Image

Image

feature-libs/user/profile/components/address-book/address-form/address-form.component.ts
```diff
@@ -171,6 +191,78 @@ export class AddressFormComponent implements OnInit, OnDestroy {
...
+              return of([]);
```
i336588, line 224: Another issue:
When editing an existing Chinese address.
1. Switch to another country such as Canada.
2. Switch country back to China. The previous province is filled, and the city field is empty.
3. Click the drop down list to select a city, but find the city list is empty.

Image

feature-libs/user/profile/assets/translations/en/address.json
```diff
@@ -36,6 +38,8 @@
     "streetAddress": "Street Address",
     "aptSuite": "Apartment Number",
     "selectOne": "Select One...",
+    "selectProvinceFirst": "Please select a province first",
+    "selectCityFirst": "Please select a city first",
     "setAsDefault": "Set as default delivery address",
     "titleRequired": "Title is required.",
     "userAddressAddSuccess": "New address was added successfully!",
```
i336588, line 46: When adding a Chinese address, and the mobile number is invalid, the address is failed to create with error message not translated:

Image

feature-libs/user/profile/components/address-book/address-form/address-form.component.ts
```diff
@@ -114,10 +134,11 @@ export class AddressFormComponent implements OnInit, OnDestroy {
     lastName: ['', Validators.required],
     line1: ['', Validators.required],
     line2: [''],
-    town: ['', Validators.required],
+    town: [null, Validators.required],
     region: this.fb.group({
       isocode: [null, Validators.required],
     }),
+    district: [null],
```
i336588, line 147: For Chinese address, validation error for district is missing. And the error message for Province should be refined to "Field Province is required".

Image

feature-libs/user/profile/components/address-book/address-form/address-form.component.ts
```diff
@@ -171,6 +191,78 @@ export class AddressFormComponent implements OnInit, OnDestroy {
...
+          }
+          this.cdr.markForCheck();
+        })
+    );
+
+    this.subscription.add(
+      this.selectedCity$
+        .pipe(
+          switchMap((cityIsocode) => {
+            if (!this.isChineseAddress) {
```
i336588: When editing a non-Chinese address, the value for city is missing.

Image

feature-libs/user/profile/components/address-book/address-form/address-form.component.ts
```diff
@@ -162,15 +182,82 @@ export class AddressFormComponent implements OnInit, OnDestroy {
...
         this.regionSelected(this.addressData.region);
       }
     }
 
     this.addresses$ = this.userAddressService.getAddresses();
+
+    this.subscription.add(
+      this.selectedRegion$
+        .pipe(
+          switchMap((regionIsocode) => {
```
i336588, line 222: When creating a Chinese address, and filled in a invalid mobile number, and submit. 
Expected: Stay in the current address form page and give error message on the top as this:

Image
Actual: Redirect to address book page, and give error message.

Image

feature-libs/user/profile/assets/translations/cs/address.json
```diff
@@ -29,7 +29,8 @@
     },
     "cellphone": {
       "label": "Číslo mobilního telefonu (volitelné)",
-      "placeholder": "např. (###) - ### - ####"
+      "placeholder": "např. (###) - ### - ####",
+      "requiredLabel": "Číslo mobilního telefonu"
```
i336588: We should only add properties in en/address.json.

feature-libs/user/profile/components/address-book/address-form/address-form.component.html
```diff
@@ -267,15 +339,23 @@
...
-          'addressForm.cellphone.label' | cxTranslate
-        }}</span>
+        <span class="label-content" [class.required]="isChineseAddress"
+          >{{ 'addressForm.cellphone.label' | cxTranslate }}
+          <cx-form-required-asterisks *ngIf="isChineseAddress" />
+        </span>
         <input
           type="tel"
           class="form-control"
           placeholder="{{ 'addressForm.cellphone.placeholder' | cxTranslate }}"
```
i336588: mobile number is required for Chinese address but optional for other countries. The (optional) should be removed when it's Chinese address.

Image

feature-libs/user/profile/assets/translations/zh/address.json
```diff
@@ -19,6 +19,8 @@
       "placeholder": "城市"
     },
     "state": "州/省",
+    "province": "省份",
```
i336588: address.json under zh and zh_TW should not be changed.

i336588 (dismissed):

feature-libs/user/profile/components/address-book/address-book.component.spec.ts
```diff
@@ -257,13 +269,101 @@ describe('AddressBookComponent', () => {
...
+
+  describe('addAddressSubmit', () => {
+    beforeEach(() => {
+      isLoading.next(false);
+      isError.next(false);
+    });
+
+    it('should close the form when addUserAddress succeeds', () => {
+      component.showAddAddressForm = true;
+      component.addAddressSubmit(mockAddress);
```
Pucek9, line 312: It would be nice to spy methods called in `addAddressSubmit` like `getAddressesStateLoading`, `getAddressesError` and verify if they were called:`toHaveBeenCalledWith(...)` / `not.toHaveBeenCalled();`

It's just an example, but it's worth keeping better test quality as documentation of code.
My recommendation: Just ask AI agent to verify your spec files from this perspective and add specs which are checking it.

feature-libs/user/profile/components/address-book/address-book.component.ts
```diff
@@ -71,18 +93,54 @@ export class AddressBookComponent implements OnInit {
   }
 
   addAddressSubmit(address: Address): void {
-    this.showAddAddressForm = false;
+    if (!address) {
+      this.showAddAddressForm = false;
+      return;
+    }
```
Pucek9, line 115: Is this an intentional change?
It changes the behavior of that method and perhaps needs to be handled by a feature toggle.
Otherwise you can safely change into:
```suggestion
    this.showAddAddressForm = false;
    if (!address) {
      return;
    }
```
here and below (editAddressSubmit)

projects/core/src/occ/adapters/site-context/default-occ-site-context-config.ts
```diff
@@ -15,6 +15,8 @@ export const defaultOccSiteContextConfig: OccConfig = {
         countries: 'countries',
         regions:
           'countries/${isoCode}/regions?fields=regions(name,isocode,isocodeShort)',
+        chineseAddressCities: 'regions/${regionId}/cities',
+        chineseAddressDistricts: 'cities/${cityId}/districts',
```
Pucek9: Are you sure this will only be used in the Chinese configuration? Or is there another country with a similar setup? In any case, I think it’s better to avoid hard-coding this into the props name and code (variables). Is there another common factor?
```suggestion
        cties: 'regions/${regionId}/cities',
        cityDistricts: 'cities/${cityId}/districts',
```

projects/core/src/occ/occ-models/occ-endpoints.model.ts
```diff
@@ -97,6 +97,14 @@ export interface OccEndpoints {
...
    */
   regions?: string | OccEndpoint;
+  /**
+   * Fetch cities for a Chinese address region
+   */
+  chineseAddressCities?: string | OccEndpoint;
+  /**
+   * Fetch districts for a Chinese address city
+   */
+  chineseAddressDistricts?: string | OccEndpoint;
```
Pucek9:
```suggestion
  /**
   * Fetch cities for an address region
   */
  addressCities?: string | OccEndpoint;
  /**
   * Fetch districts for an address city
   */
  addressDistricts?: string | OccEndpoint;
```

feature-libs/user/profile/components/address-book/address-form/address-form.component.html
```diff
@@ -213,40 +319,6 @@
         ></cx-form-errors>
       </label>
     </div>
-    <ng-container *ngIf="regions$ | async as regions" formGroupName="region">
```
Pucek9, line 216: Seems like breaking change (needs *cxFeature directive to handle this order)

feature-libs/user/profile/components/address-book/address-form/address-form.component.ts
```diff
@@ -114,10 +134,11 @@ export class AddressFormComponent implements OnInit, OnDestroy {
     lastName: ['', Validators.required],
     line1: ['', Validators.required],
     line2: [''],
-    town: ['', Validators.required],
+    town: [null, Validators.required],
```
Pucek9: breaking change I guess

feature-libs/user/profile/components/address-book/address-form/address-form.component.ts
```diff
@@ -162,15 +182,82 @@ export class AddressFormComponent implements OnInit, OnDestroy {
...
 
-      this.countrySelected(this.addressData.country);
       if (this.addressData.region) {
         this.regionSelected(this.addressData.region);
       }
     }
 
     this.addresses$ = this.userAddressService.getAddresses();
+
+    this.subscription.add(
```
Pucek9, line 219: AFAIK we separate similar code to methods to make the constructor clean and easy to override. I guess for ngOnInit, it's worth keeping the same approach

feature-libs/user/profile/components/address-book/address-book.component.ts
```diff
@@ -125,16 +184,21 @@ export class AddressBookComponent implements OnInit {
...
+            .join(', ');
+
           return {
             role: 'application',
             textBold: address.firstName + ' ' + address.lastName,
             text: [
               address.line1,
               address.line2,
-              address.town + ', ' + region + address.country?.isocode,
+              locationParts,
```
Pucek9: is using `region?.name` with higher priority than `region?.isocode` not a breaking change?

feature-libs/user/profile/components/address-book/address-form/address-form.component.html
espada945: Fixed. The regions block is now conditionally rendered based on isChineseAddress — for Chinese addresses, province appears before city (required for the cascading selection); for all other countries, the original order (state after zip) is preserved unchanged.

feature-libs/user/profile/components/address-book/address-form/address-form.component.ts
espada945: Reverted to '' as the initial value — null for the Chinese city dropdown is handled at runtime via townControl.reset() inside countrySelected().

feature-libs/user/profile/components/address-book/address-form/address-form.component.ts
espada945: Extracted the three subscription blocks from ngOnInit into separate protected methods: initCitiesSubscription(), initDistrictsSubscription(), and initLanguageSubscription().

projects/core/src/occ/adapters/site-context/default-occ-site-context-config.ts
espada945: Renamed to addressCities and addressDistricts as suggested.

feature-libs/user/profile/components/address-book/address-book.component.ts
espada945: This is intentional. In this PR, we added cellphone as a required field for Chinese addresses. We want the form to stay open when submission fails (e.g. invalid phone number), so the user can correct it and resubmit.

feature-libs/user/profile/components/address-book/address-book.component.spec.ts
espada945: Added toHaveBeenCalledWith assertions for addUserAddress and updateUserAddress in the success/failure cases as well, so the tests now document the full flow.

feature-libs/user/profile/components/address-book/address-book.component.ts
espada945: `region?.name` is now only preferred for Chinese addresses. Other countries continue to use `region?.isocode` as before.

core-libs/core/src/occ/adapters/site-context/default-occ-site-context-config.ts
```diff
@@ -15,6 +15,8 @@ export const defaultOccSiteContextConfig: OccConfig = {
         countries: 'countries',
         regions:
           'countries/${isoCode}/regions?fields=regions(name,isocode,isocodeShort)',
+        addressCities: 'regions/${regionId}/cities',
+        addressDistricts: 'cities/${cityId}/districts',
```
Pucek9: Commonly, we add "fields" params directly in occ config. I can see only one exception in 'storeDetails', which is probably a mistake.
```suggestion
        addressCities: 'regions/${regionId}/cities?fields=cities(name,isocode)',
        addressDistricts: 'cities/${cityId}/districts?fields=districts(name,isocode)',
```

core-libs/core/src/occ/adapters/site-context/occ-site.adapter.ts
```diff
@@ -110,4 +116,26 @@ export class OccSiteAdapter implements SiteAdapter {
...
   }
+
+  loadCities(regionIsocode: string): Observable<City[]> {
+    const url = this.occEndpointsService.buildUrl('addressCities', {
+      urlParams: { regionId: regionIsocode },
+    });
+    return this.http
+      .get<{ cities: City[] }>(url, {
+        params: { fields: 'cities(name,isocode)' },
+      })
```
Pucek9:
```suggestion
      .get<{ cities: City[] }>(url)
```

core-libs/core/src/occ/adapters/site-context/occ-site.adapter.ts
```diff
@@ -110,4 +116,26 @@ export class OccSiteAdapter implements SiteAdapter {
...
+  }
+
+  loadDistricts(cityIsocode: string): Observable<CityDistrict[]> {
+    const url = this.occEndpointsService.buildUrl('addressDistricts', {
+      urlParams: { cityId: cityIsocode },
+    });
+    return this.http
+      .get<{ districts: CityDistrict[] }>(url, {
+        params: { fields: 'districts(name,isocode)' },
+      })
```
Pucek9:
```suggestion
      .get<{ districts: CityDistrict[] }>(url)
```

core-libs/core/src/occ/adapters/site-context/occ-site.adapter.ts
```diff
@@ -110,4 +116,26 @@ export class OccSiteAdapter implements SiteAdapter {
...
+
+  loadCities(regionIsocode: string): Observable<City[]> {
+    const url = this.occEndpointsService.buildUrl('addressCities', {
+      urlParams: { regionId: regionIsocode },
+    });
+    return this.http
+      .get<{ cities: City[] }>(url, {
+        params: { fields: 'cities(name,isocode)' },
+      })
+      .pipe(map((res) => res.cities ?? []));
```
Pucek9: What about adding a normalizer? Even without any mapping logic, just for customers for easier customization. 

Here and below

feature-libs/user/profile/components/address-book/address-form/address-form.component.ts
```diff
@@ -68,10 +79,19 @@ import { filter, map, switchMap, take, tap } from 'rxjs/operators';
...
+  protected cdr = inject(ChangeDetectorRef);
+  protected userProfileFacade = inject(UserProfileFacade);
+
   countries$: Observable<Country[]>;
   titles$: Observable<Title[]>;
   regions$: Observable<Region[]>;
   selectedCountry$: BehaviorSubject<string> = new BehaviorSubject<string>('');
+  selectedRegion$: BehaviorSubject<string> = new BehaviorSubject<string>('');
+  selectedCity$: BehaviorSubject<string> = new BehaviorSubject<string>('');
+  isChineseAddress = false;
```
Pucek9:
```suggestion
  isHierarchicalAddressFormat = false;
```
or
```suggestion
  isMultiLevelAddressFormat = false;
```

feature-libs/user/profile/components/address-book/address-form/address-form.component.ts
```diff
@@ -214,10 +311,49 @@ export class AddressFormComponent implements OnInit, OnDestroy {
   countrySelected(country: Country | undefined): void {
     this.addressForm.get('country')?.get('isocode')?.setValue(country?.isocode);
     this.selectedCountry$.next(country?.isocode ?? '');
+    this.isChineseAddress = country?.isocode === 'CN';
```
Pucek9:
```suggestion
    this.isHierarchicalAddressFormat = countriesUsesHierarchicalAddressFormat.includes(country?.isocode);
```

where some address model file should:
```
export const countriesUsesHierarchicalAddressFormat = ['CN', 'JP', 'KR', 'VN', 'TW'];
```

Pucek9: OR (if there is no plan to use it in other countries now)


```suggestion
    this.isHierarchicalAddressFormat = supportedCountriesUsesHierarchicalAddressFormat.includes(country?.isocode);
```

where some address model file should:
``
export const supportedCountriesUsesHierarchicalAddressFormat = ['CN'];
``

feature-libs/user/profile/components/address-book/address-book.component.ts
Pucek9: Okay, but it still change of current customer's application behavior, which is a breaking change and requires introducing a feature toggle check. 
Here and below

feature-libs/user/profile/components/address-book/address-book.component.ts
```diff
@@ -147,6 +205,21 @@ export class AddressBookComponent implements OnInit {
     );
   }
 
+  protected buildLocationLine(address: Address): string {
+    if (address.country?.isocode === 'CN') {
```
Pucek9: Please follow the same approach as mentioned here: https://github.com/SAP/spartacus/pull/21272/changes#r3307543250

feature-libs/user/profile/components/address-book/address-book.component.ts
```diff
@@ -124,17 +176,23 @@ export class AddressBookComponent implements OnInit {
           actions.push({ name: textDelete, event: 'delete' });
 
           const numbers = getAddressNumbers(address, textPhone, textMobile);
+          const locationLine = this.buildLocationLine(address);
+          const districtName =
+            address.country?.isocode === 'CN'
```
Pucek9: Please follow the same approach as mentioned here: https://github.com/SAP/spartacus/pull/21272/changes#r3307543250

feature-libs/user/profile/components/address-book/address-book.component.ts
Pucek9: By the way, what about customers who previously used CN and customized code in their own way? 
Wrapping cases like here by the condition of the country isocode is not preventing breaking changes.

core-libs/core/src/occ/adapters/user/occ-user-address.adapter.spec.ts
```diff
@@ -31,11 +32,17 @@ const address: Address = {
 const suggestedAddresses: AddressValidation = {
   suggestedAddresses: [address],
 };
+
+class MockFeatureConfigService {
```
Pucek9:
```suggestion
class MockFeatureConfigService implements Partial<FeatureConfigService> {
```

core-libs/core/src/occ/adapters/user/occ-user-address.adapter.spec.ts
```diff
@@ -55,6 +66,9 @@ describe('OccUserAddressAdapter', () => {
     httpMock = TestBed.inject(HttpTestingController);
     converter = TestBed.inject(ConverterService);
     occEnpointsService = TestBed.inject(OccEndpointsService);
+    featureConfigService = TestBed.inject(
+      FeatureConfigService
+    ) as unknown as MockFeatureConfigService;
```
Pucek9:
```suggestion
    featureConfigService = TestBed.inject(FeatureConfigService);
```
```

File: core-libs/core/src/occ/adapters/user/occ-user-address.adapter.ts [hunk 55]
```diff
@@ -31,6 +32,7 @@ const CONTENT_TYPE_JSON_HEADER = { 'Content-Type': 'application/json' };
 @Injectable()
 export class OccUserAddressAdapter implements UserAddressAdapter {
   protected logger = inject(LoggerService);
+  protected featureConfigService = inject(FeatureConfigService);
```
Pucek9: `FeatureConfigService` and `FeatureToggles` are exceptions: its useles for our customers

```suggestion
  private featureConfigService = inject(FeatureConfigService);
```

feature-libs/user/profile/components/address-book/address-form/address-form.component.ts
```diff
@@ -68,10 +81,19 @@ import { filter, map, switchMap, take, tap } from 'rxjs/operators';
   ],
 })
 export class AddressFormComponent implements OnInit, OnDestroy {
+  protected languageService = inject(LanguageService);
+  protected cdr = inject(ChangeDetectorRef);
+  protected featureConfigService = inject(FeatureConfigService);
```
Pucek9:
```suggestion
  private featureConfigService = inject(FeatureConfigService);
```

feature-libs/user/profile/components/address-book/address-book.component.ts
```diff
@@ -47,6 +57,11 @@ export class AddressBookComponent implements OnInit {
   showEditAddressForm = false;
   editCard: string | null;
 
+  private subscription = new Subscription();
+
+  protected languageService = inject(LanguageService);
+  protected featureConfigService = inject(FeatureConfigService);
```
Pucek9:
```suggestion
  private featureConfigService = inject(FeatureConfigService);
```

core-libs/core/src/model/address.model.ts
```diff
@@ -23,6 +23,25 @@ export interface Region {
...
+  isocode?: string;
+  name?: string;
+}
+
+/**
+ * Country isocodes whose addresses use a hierarchical (multi-level)
+ * format: country → region → city → district. For these countries the
+ * address form shows chained dropdowns and skips OCC address verification.
+ */
+export const supportedCountriesUsesHierarchicalAddressFormat = ['CN'];
```
Pucek9: [MAJOR] Terribly sorry for suggesting placing it in the model!
A single source of truth is definitely better. Looking at it again, I was concerned about the lack of an override option. This collection looks like a candidate for a new member of the configuration. Please discuss this with your team if not sure. I believe it's a good direction

eg.
```ts
@Injectable({
  providedIn: 'root',
  useExisting: Config,
})
export abstract class HierarchicalAddressConfig { // or just AddressConfig
  countriesUsingHierarchicalAddressFormat?: string[];
}
```
[MINOR]
instead of `string` we could have some type that is more strict for isocodes (2 letters?)

feature-libs/user/profile/components/address-book/address-form/address-form.component.html
```diff
@@ -158,24 +158,89 @@
     </label>
   </div>
   <div class="row">
+    <ng-container *ngIf="isChineseAddress">
```
Pucek9: Forgot to rename & handle toggle in this component
```suggestion
    <ng-container *cxFeature="enableHierarchicalAddressFormat">
      <ng-container *ngIf="isHierarchicalAddressFormat">
```

feature-libs/user/profile/components/address-book/address-form/address-form.component.html
```diff
@@ -158,24 +158,89 @@
...
           >{{ 'addressForm.city.label' | cxTranslate }}
 
           <cx-form-required-asterisks />
         </span>
-        <input
-          required="true"
-          type="text"
-          class="form-control"
-          placeholder="{{ 'addressForm.city.placeholder' | cxTranslate }}"
+        <ng-select
```
Pucek9: I think it should work fine:
```suggestion
    <ng-container *cxFeature="enableHierarchicalAddressFormat">
       <ng-container *ngIf="isHierarchicalAddressFormat; else cityTextInput">
         <ng-select ...>
       </ng-container> 
     </ng-container>       
     <ng-container *cxFeature="!enableHierarchicalAddressFormat">   
       <ng-container *ngTemplateOutlet="cityTextInput"></ng-container>     
     </ng-container>
     <ng-template #cityTextInput>
```

feature-libs/user/profile/components/address-book/address-form/address-form.component.html
```diff
@@ -185,6 +250,39 @@
         ></cx-form-errors>
       </label>
     </div>
+    <div class="form-group col-md-6" *ngIf="isChineseAddress">
```
Pucek9:
```suggestion
    <ng-container *cxFeature="enableHierarchicalAddressFormat">
      <div class="form-group col-md-6" *ngIf="isHierarchicalAddressFormat">
```

core-libs/core/src/user/store/user-state.ts
```diff
@@ -48,6 +56,8 @@ export interface UserState {
   countries: DeliveryCountriesState;
   payments: StateUtils.LoaderState<PaymentDetails[]>;
   regions: StateUtils.LoaderState<RegionsState>;
+  cities: StateUtils.LoaderState<CitiesState>;
+  districts: StateUtils.LoaderState<DistrictsState>;
```
Pucek9: Not sure if adding new properties for the existing interface should be with an optional type

espada945: Actually those selector signatures are added in this PR too, so we could relax them. Should I switch to optional + LoaderState<...> | undefined selector signatures?

Pucek9: I'm afraid we'll have to do it.

That is mentioned [here](https://sap.github.io/spartacus-docs/breaking-changes/#avoiding-breaking-changes):
>Adding new, required properties to interfaces and types. However, you can add optional properties with ?.

I guess if someone implements this interface in their app, and if it will be not optional, his typescript code will break.


```suggestion
  cities?: StateUtils.LoaderState<CitiesState>;
  districts?: StateUtils.LoaderState<DistrictsState>;
```

feature-libs/user/profile/components/address-book/address-book.component.ts
```diff
@@ -47,6 +57,12 @@ export class AddressBookComponent implements OnInit {
   showEditAddressForm = false;
   editCard: string | null;
 
+  private subscription = new Subscription();
```
Pucek9:
```suggestion
  prototected subscription = new Subscription();
```

feature-libs/user/profile/components/address-book/address-form/address-form.component.html
```diff
@@ -267,15 +380,30 @@
     </div>
     <div class="form-group col-md-6">
       <label>
-        <span class="label-content">{{
-          'addressForm.cellphone.label' | cxTranslate
-        }}</span>
+        <span
```
Pucek9: one more *cxFeature here

feature-libs/user/profile/components/address-book/address-form/address-form.component.html
```diff
@@ -267,15 +380,30 @@
...
+          }}
+          <cx-form-required-asterisks *ngIf="isHierarchicalAddressFormat" />
+        </span>
         <input
           type="tel"
           class="form-control"
           placeholder="{{ 'addressForm.cellphone.placeholder' | cxTranslate }}"
           formControlName="cellphone"
         />
+        <cx-form-errors
```
Pucek9: and here

Pucek9 (dismissed):

Pucek9 (approved):

PR #21281, opened 2026-03-24
fix: add removeVoucher endpoint

FollowTheFlo:


projects/core/src/features-config/feature-toggles/config/feature-toggles.ts
```diff
@@ -504,4 +511,5 @@ export const defaultFeatureToggles: Required<FeatureTogglesInterface> = {
   a11yReviewsKeyboardControls: false,
   a11yCartQuickOrderFormEnableSubmitAndAddValidation: false,
   a11yVocalizeDropdownItemCount: false,
+  enableRemoveVoucherEndpoint: false,
```
FollowTheFlo, line 514: about the naming, maybe _useRemoveVoucherEndpoint_ is better? but 'enable' might fit more the context of toogle.

ercultimate (approved):

PR #21285, opened 2026-03-26
feat: epic-giftCard

ijitendrasap:
This PR focuses on implementing giftcard. The following references are provided to help understand the scope, flow, and implementation details of the items included in this epic.

Epic : [CXOPF-4421](https://jira.tools.sap/browse/CXOPF-4421)  OPF: Support Gift Card as a payment option+MVP Split Payment(Gift Card+ Digital Payments) _Q2

figma : https://www.figma.com/design/LYg8naFyqTHdwJMMh3senh/OPF-2025?node-id=8720-29036&t=bBiF7YyQWoLN6bYk-0

Design Document : https://wiki.one.int.sap/wiki/spaces/SAPCCWiki/pages/5766229671/Gift+Card+Authorization
https://wiki.one.int.sap/wiki/spaces/SAPCCWiki/pages/5766222178/OPF+Gift+Card


integration-libs/opf/gift-card/assets/translations/en/giftCard.json
```diff
@@ -0,0 +1,38 @@
+{
+  "giftCard": {
```
Matejk00: Let's name a key `opfGiftCard` for consistency. 
Also the filename to `opfGiftCard.json`

integration-libs/opf/checkout/components/opf-checkout-payments/opf-checkout-payments.component.html
```diff
@@ -9,6 +9,8 @@ <h3 *ngIf="isHeadingDisplayed">
     <ng-container *ngIf="!activeConfigurationsState?.loading; else loading">
       <div class="cx-payment-options-list">
         <ng-content></ng-content>
+        <div *ngIf="!(opfPaymentEventsService.isGiftCardCoveredTotalAmountEvent$ | async)"
```
Matejk00: Do we need this if statement? I think the custom logic should be injected using outlets in the template. We should not modify the existing logic, as it may introduce breaking changes.

integration-libs/opf/base/root/model/opf-base.model.ts
```diff
@@ -84,4 +84,5 @@ export enum OpfDynamicScriptResourceType {
 export enum OpfPaymentProviderType {
   PAYMENT_GATEWAY = 'PAYMENT_GATEWAY',
   PAYMENT_METHOD = 'PAYMENT_METHOD',
+  STORED_VALUE_PAYMENT = 'STORED_VALUE_PAYMENT',
```
Matejk00, line 87: I think this property can be augmented from the gift-card root library. So, if the gift-card feature is not installed, customers can only use the two default values.

Augmentation example can be found in: `integration-libs/opf/quick-buy/root/model/augmented-opf-quick-buy.model.ts`

feature-libs/cart/base/root/models/cart.model.ts
```diff
@@ -102,6 +102,14 @@ export interface Cart {
   user?: Principal;
   quoteCode?: string;
   quotePurchaseOrderNumber?: string;
+  _availableOperations?: SAPAvailableOperation[];
```
Matejk00: Could this also be done through model augmentation from the gift-card library?

feature-libs/cart/base/root/models/cart-outlets.model.ts
```diff
@@ -26,4 +26,5 @@ export enum CartOutlets {
   CPQ_QUOTE = 'cx-cpq-quote-offer',
   SUBSCRIPTION_PRICE_HEADING = 'cx-subscription-cart-price-heading',
   SUBSCRIPTION_PRICE_BODY = 'cx-subscription-cart-price-body',
+  ORDER_DETAILS_PAYMENT_METHOD_DETAILS = 'cx-order-details-payment-method-details',
```
Matejk00: Just to confirm, isn’t this duplicating the OPF Outlets positions defined in `integration-libs/opf/checkout/root/model/opf-checkout-outlets.model.ts`?

feature-libs/cart/base/root/models/cart.model.ts
```diff
@@ -102,6 +102,14 @@ export interface Cart {
...
+}
+export interface SAPAvailableOperation {
```
Matejk00: Can you please provide some JS doc, not sure about the purpose of this interface. The name is a quite generic.

integration-libs/opf/gift-card/root/model/opf-gift-card.model.ts
```diff
@@ -0,0 +1,27 @@
+/*
+ * SPDX-FileCopyrightText: 2026 SAP Spartacus team <spartacus-team@sap.com>
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { Price } from '@spartacus/core';
+
+export interface SAPGiftCards {
```
Matejk00: I suggest small fix here. Let’s name it with the Opf prefix, so it could be `OpfGiftCards`, `OpfGiftCardSummary`, and so on as they are part of the OPF lib. Or maybe `OpfCartGiftCards`?

Matejk00 (changes requested):

feature-libs/cart/base/root/models/cart.model.ts
ijitendrasap: I kept this in the cart model because it’s generic and not specific to gift card only.
discussion : https://sap-cx.slack.com/archives/CF4TZ6EBB/p1768985829997259?thread_ts=1768985771.957929&cid=CF4TZ6EBB

integration-libs/opf/gift-card/assets/translations/en/giftCard.json
ijitendrasap: done.

integration-libs/opf/checkout/components/opf-checkout-payments/opf-checkout-payments.component.html
ijitendrasap: The gift card component dispatches an event (via opfPaymentEventsService) with a flag indicating whether the gift card fully covers the total amount. The payment options component subscribes to this event stream and uses a boolean value to conditionally render the available payment options.

This approach keeps the logic centralized within the event service and avoids modifying existing logic, reducing the risk of breaking changes.
If we use outlets instead, it would lead to duplication of the payment options component.

integration-libs/opf/base/root/model/opf-base.model.ts
ijitendrasap: STORED_VALUE_PAYMENT this value is also used in method (OpfCheckoutPaymentsComponent)
getActiveConfigurations(): Observable<QueryState<OpfActiveConfigurationsResponse | undefined>> 
to filter the gift card as payment options because this is coming with other payment options

integration-libs/opf/gift-card/root/model/opf-gift-card.model.ts
ijitendrasap: done

feature-libs/cart/base/root/models/cart-outlets.model.ts
ijitendrasap: done. 
In opf-checkout-outlets.model.ts, ORDER_DETAILS_AFTER_PAYMENT_OPTIONS is used to display the different payment options applied to the order.

ORDER_DETAILS_PAYMENT_METHOD_DETAILS, on the other hand, is used to show the payment method details, such as masked card numbers, and also includes gift card information in cases of split payments.

feature-libs/cart/base/root/models/cart.model.ts
ijitendrasap: _availableOperations is a generic field and not specific to gift cards alone.

integration-libs/opf/checkout/components/opf-checkout-payment-and-review/opf-checkout-payment-and-review.component.html
```diff
@@ -60,6 +60,14 @@ <h3>{{ 'opfCheckout.termsAndConditions' | cxTranslate }}</h3>
       [isExplicit]="explicitTermsAndConditions.value"
       [ngClass]="['for-desktop']"
     ></cx-opf-checkout-terms-and-conditions-alert>
+    <ng-container>
```
Matejk00: I think that this `ng-container` tag is unnecessary.
Would you mind to remove it?

integration-libs/opf/schematics/add-opf/schema.json
```diff
@@ -89,6 +91,11 @@
             "value": "OPF-Tokenisation",
             "label": "OPF Tokenisation",
             "checked": false
+          },
+          {
+            "value": "OPF-Gift-Card",
+            "label": "OPF Gift Card",
+            "checked": true
```
Matejk00: Should this feature be enabled by default for customers during installation? If not, we need to set this flag to false, and it will be installed only if the customer selects it during the installation in schematics.

feature-libs/cart/base/root/models/cart.model.ts
Matejk00: Thank you for sharing: https://github.com/SAP/spartacus/pull/21285/changes#r3015031392

Please let’s avoid underscore-prefixed fields in the public Cart model and expose `availableOperations?: CartAvailableOperation[]` instead if possible (We should also rename SAPAvailableOperation to a cart-scoped name (e.g. CartAvailableOperation) to reflect its actual usage and avoid implying a broader SAP-wide contract)

I understand that might be a global change, but If OPF Gift Card is the only consumer at this moment, better would be declaring this via OPF Gift Card Cart augmentation rather than adding OPF-specific fields to the base @spartacus/cart types and touching another lib structure.

Matejk00 (changes requested):
Please check the pipeline issues and try to make them green. For example `npm run prettier:fix` needs to be run:

```
Validating code formatting (using prettier)

> storefrontapp@0.0.0 prettier
> prettier --config ./.prettierrc --list-different "{projects,feature-libs,core-libs,integration-libs}/**/*{.ts,.js,.json,.scss,.html}"

integration-libs/opf/checkout/components/opf-checkout-payments/opf-checkout-payments.component.html
```

Also the PR title should be more changelog friendly and descriptive.

integration-libs/opf/schematics/add-opf/schema.json
ijitendrasap: this wont be enabled by default. changed to false

integration-libs/opf/checkout/components/opf-checkout-payment-and-review/opf-checkout-payment-and-review.component.html
ijitendrasap: done

feature-libs/cart/base/root/models/cart.model.ts
ijitendrasap: Done. You’re correct currently the availableOperations field is only used by the gift card. I’ve renamed the field and moved its definition from cart.model to augmented-opf-gift-card.model.ts.

Matejk00 (approved):

PR #21286, opened 2026-03-26
feat: Add vocalization of dropdown item count

BenjaminEmiliani:
Tested and validated on iOS 26.4 - iPhone 17. Behaviour is almost identical to what is observed on mac Voice Over.


projects/storefrontlib/shared/components/ng-select-a11y/ng-select-a11y.directive.ts
```diff
@@ -76,6 +78,46 @@ export class NgSelectA11yDirective implements AfterViewInit {
     private elementRef: ElementRef
   ) {
     useFeatureStyles('a11yNgSelectUnicodeCarets');
+    if (this.featureConfigService.isEnabled('a11yVocalizeDropdownItemCount')) {
+      effect(() => {
```
Pucek9: Please add a new method and move this logic into it, and call it in the constructor. 
It will allow overraidabilty

projects/storefrontlib/shared/components/ng-select-a11y/ng-select-a11y.directive.ts
```diff
@@ -76,6 +78,46 @@ export class NgSelectA11yDirective implements AfterViewInit {
...
+                );
+              if (inputCombobox) {
+                this.renderer.setAttribute(
+                  inputCombobox,
+                  'aria-describedby',
+                  countId
+                );
+              }
+              this.destroyRef.onDestroy(() => itemCountSpan.remove());
+            }
```
Pucek9: Imo woth to separate into another method:
```ts
  protected createItemCountSpan() {
    const itemCountSpan = this.renderer.createElement('span');
    this.renderer.addClass(itemCountSpan, 'cx-ng-select-count');
    this.renderer.addClass(itemCountSpan, 'cx-visually-hidden');
    this.renderer.setAttribute(itemCountSpan, ARIA_HIDDEN, 'true');
    this.renderer.appendChild(this.elementRef.nativeElement, itemCountSpan);
    const countId =
      (this.elementRef.nativeElement.id || 'ng-select') + '-count';
    this.renderer.setAttribute(itemCountSpan, 'id', countId);
    const inputCombobox =
      this.elementRef.nativeElement.querySelector('[role="combobox"]');
    if (inputCombobox) {
      this.renderer.setAttribute(inputCombobox, 'aria-describedby', countId);
    }
    this.destroyRef.onDestroy(() => itemCountSpan.remove());
    return itemCountSpan;
  }
```

projects/storefrontlib/shared/components/ng-select-a11y/ng-select-a11y.directive.ts
```diff
@@ -76,6 +78,46 @@ export class NgSelectA11yDirective implements AfterViewInit {
...
+      effect(() => {
+        this.translationService
+          .translate('assistiveMessage.ngSelectDropdownCount', {
+            count: this.selectComponent.items()?.length ?? 0,
+          })
+          .pipe(take(1), takeUntilDestroyed(this.destroyRef))
+          .subscribe((countText) => {
+            let itemCountSpan = this.elementRef.nativeElement.querySelector(
+              '.cx-ng-select-count'
+            );
```
Pucek9: Then we can simplify 
```suggestion
            const itemCountSpan = this.elementRef.nativeElement.querySelector(
              '.cx-ng-select-count'
            ) ?? createItemCountSpan();
```

nx.json
```diff
@@ -17,5 +17,6 @@
       "cache": true
     }
   },
-  "useInferencePlugins": false
+  "useInferencePlugins": false,
+  "analytics": false
```
Pucek9, line 21: What is the reason for that change?

projects/assets/src/translations/en/common.json
```diff
@@ -247,7 +247,9 @@
     "refreshThePage": "Please refresh the page."
   },
   "assistiveMessage": {
-    "actionCancelled": "Action cancelled, nothing changed"
+    "actionCancelled": "Action cancelled, nothing changed",
+    "ngSelectDropdownCount": "{{count}} option available",
+    "ngSelectDropdownCount_other": "{{count}} options available"
```
Pucek9: > ngSelectDropdownOptionsList

@uroslates I can see you introduced it year ago and Ben would like to continue this semantic.
Is it some new customary to include the library name in the translation key? Is there a reason for that?
```suggestion
    "dropdownCount": "{{count}} option available",
    "dropdownCount_other": "{{count}} options available"
```

nx.json
BenjaminEmiliani: it is always prompted by nx if user wants to share analytics when using the CLI. By committing this, no one will be asked - and usually it is better to share as little as possible for analytics.

projects/assets/src/translations/en/common.json
BenjaminEmiliani: The only reason I can think of is clarity

projects/storefrontlib/shared/components/ng-select-a11y/ng-select-a11y.directive.ts
BenjaminEmiliani: done

projects/storefrontlib/shared/components/ng-select-a11y/ng-select-a11y.directive.ts
BenjaminEmiliani: Done

projects/storefrontlib/shared/components/ng-select-a11y/ng-select-a11y.directive.ts
BenjaminEmiliani: Done

projects/storefrontlib/shared/components/ng-select-a11y/ng-select-a11y.directive.ts
```diff
@@ -76,6 +78,7 @@ export class NgSelectA11yDirective implements AfterViewInit {
     private elementRef: ElementRef
   ) {
     useFeatureStyles('a11yNgSelectUnicodeCarets');
+    this.vocalizeItemCount();
```
Pucek9: Would prefer ratcher

```suggestion
  if (this.featureConfigService.isEnabled('a11yVocalizeDropdownItemCount')) {
    this.vocalizeItemCount();
  }  
```
what do you think?

nx.json
Pucek9: Does this only affect our project or our clients as well? If you agreed to change with someone with more context, then fine

projects/storefrontlib/shared/components/ng-select-a11y/ng-select-a11y.directive.ts
BenjaminEmiliani: I agree - Done

projects/assets/src/translations/en/common.json
BenjaminEmiliani: Renamed to `dropdownItemCount`

Pucek9 (approved):

projects/assets/src/translations/en/common.json
uroslates: Yes the reason was clarity because at that time the lib did not support customisation for these kind of customization. Later they introduced the config for controlling this. I think its good that you've renamed it.

PR #21309, opened 2026-03-31
fix: persist focus on consent

rmch91:
Closes: https://jira.tools.sap/browse/CXSPA-11392

QA steps:
**Checkbox focus retained on toggle**

1. Open the storefront as an anonymous (not logged in) user
2. Click the "Consent management" link in the footer to open the Anonymous Consent Dialog
3. Tab to a consent checkbox
4. Press Space to toggle the checkbox

Expected: Focus remains on the same checkbox. A success message appears but does not steal focus.
Press Space again to un-toggle
**Expected:** Focus still remains on the same checkbox.

**Message close restores focus**

1. Open the Anonymous Consent Dialog
2. Toggle a checkbox (success message appears)
3. Tab to the message close button and press Space
**Expected:** Focus returns to the previously toggled checkbox


Pio-Bar (commented):
Changes look good to me. I've tested the ConsentDialogComponent and the focus stays on checkboxes after toggling. It is not lost after the success message is closed. The Screen reader output seems consistent as well.

However, I see the original ticket concerned the `ConsentManagementComponent` not the dialog. When testing on `/my-account/consents` the focus is lost after toggling the checkboxes. The component re-renders causing the focus to default to the `body` element. Is this in the scope of the ticket?

Pio-Bar (approved):
The focus is now preserved after in both the dialog and `/my-account/consents`. LGTM!

PR #21312, opened 2026-03-31
fix: handle full page redirection within smartedit (#21268)

FollowTheFlo:
carry 'develop' fix into release branch with cherry-pick, see original PR:
https://github.com/SAP/spartacus/pull/21268

CXSPA-11484

SessionStorage key/value was added to be able to retrieve cmsTicketId after a full page redirection that requires spartacus to restart within SmartEdit.

### DIFF
--- a/feature-libs/smartedit/root/services/smart-edit-launcher.service.ts
+++ b/feature-libs/smartedit/root/services/smart-edit-launcher.service.ts
@@ -6,7 +6,11 @@
 
 import { Location } from '@angular/common';
 import { inject, Injectable } from '@angular/core';
-import { FeatureModulesService, ScriptLoader } from '@spartacus/core';
+import {
+  FeatureModulesService,
+  ScriptLoader,
+  WindowRef,
+} from '@spartacus/core';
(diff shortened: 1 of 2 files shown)


feature-libs/smartedit/root/services/smart-edit-launcher.service.ts
```diff
@@ -53,16 +59,61 @@ export class SmartEditLauncherService {
    * Indicates whether Spartacus is launched in SmartEdit
    */
   isLaunchedInSmartEdit(): boolean {
-    const path = this.location.path().split('?')[0];
-    const params = this.location.path().split('?')[1];
+    const [path, params] = this.location.path().split('?');
```
ercultimate, line 62: Please use `new URL()` to handle the parsing and `URL.searchParams` to access the values.

ercultimate (approved):

feature-libs/smartedit/root/services/smart-edit-launcher.service.ts
FollowTheFlo: as discussed, this logic was already there before, it will be handled separately

PR #21319, opened 2026-04-01
fix: update undici version to 6.24.1 in package json files

npapp-dev002:


giancorderoortiz (approved):
ok, for compliance purposes

PR #21321, opened 2026-04-01
feat: add new validation to secure password validator

BenjaminEmiliani:
Here we can see the form validation on register page when password ends with '&' (not allowed char)
image


projects/storefrontlib/shared/utils/validators/custom-form-validators.ts
```diff
@@ -213,6 +214,13 @@ export class CustomFormValidators {
     this.noConsecutiveCharacters,
   ];
 
+  // TODO: Delete this array and move mustEndWithLegalCharacter validator to securePasswordValidators
```
Pucek9: we can't remove enhancedSecurePasswordValidators since it will be part of public api.
Also, I dont think you can use featureToggle here to add mustEndWithLegalCharacter into securePasswordValidators optionally 🤔

BenjaminEmiliani: Good point, let me revisit this

Pucek9 (approved):
Great solution

